PR #1, opened 2015-08-19
[TASK] Include Exception class in ExceptionHandler with full path

kitsunet:
The ExceptionHandler includes the Flow Exception class directly by using
the relative path. This might not work out if at some point we decide to
combine autoloaded classes as this class is autoloaded and the relative
path will be wrong then. Using the FLOW_PATH_FLOW constant we can easily
construct a full path to the file.

Change-Id: I1154677458afc1633994158479fe8b3f18763d90
Releases: master


robertlemke:
:+1:

PR #4, opened 2015-08-19
[TASK] Use more sophisticated UUID generator

kitsunet:
Uses php-uuid extension if available and otherwise the
``https://github.com/ramsey/uuid`` library which is a complete
solution for generating and working with UUIDs.

Change-Id: Ide03010ffb3ded05c9d8d646e80c544a411b6e40
Releases: master


dfeyer:
Should we add the php ext php-uuid in the composer.json, it's a nice way to help user improve their setup.

kitsunet:
Reopened as #71 because I couldn't get travis to test this one... php-uuid recommendation was added.

PR #5, opened 2015-08-19
[TASK] Replace Security_Randomizer with RandomLib

kitsunet:
Replaces the old ``Security_Randomizer`` that we shipped with Flow
by ``ircmaxell/random-lib`` which is able to generate highly
secure random data and is actively maintained.

The static methods in ``Utilities\Algorithms`` are kept for
backwards compatibility but Flow doesn't use them anymore.


kitsunet:
Note that the random data generation is much slower as before and as we rely on the same random data for UUIDs we should decouple that for performance reasons. Therefore this change should be merged together with https://github.com/neos/flow-development-collection/pull/4

aertmann:
:+1: by reading

albe:
:+1: from reading

dfeyer:
:+1: by reading, but need a rebase

kitsunet:
Merge conflict fixed, rebased, should be good to go.

kitsunet:
reopening new PR, travis doesn't correctly do it's job

PR #7, opened 2015-08-19
[TASK] Compile StreamWrapper implementations statically

kitsunet:
Reduce use of ReflectionService by compiling implementations of
``TYPO3\Flow\Resource\Streams\StreamWrapperInterface`` statically into
the StreamWrapperAdapter. This in itself wil not improve performance
because it was done in the ``ResourceManager`` before, but makes the
``ResourceManager`` somewhat cleaner and allows further refactorings
of that.

Change-Id: I98b259ea15a424c426d2831f34df16c98e18a7a8
Releases: master
Relates: NEOS-1294


kitsunet:
See https://github.com/neos/flow-development-collection/pull/43 makes sense to do both together...

PR #8, opened 2015-08-19
[TASK] Make Http\Request easier to use for outbound Requests

kitsunet:
Currently the ``Http\Request`` defaults to php://input for
content. This is good for a request coming in to Flow, but rather
error prone for outbound requests as it can lead to different behavior
depending if the Request is created during a GET, POST or even CLI
call. So sometimes outbound requests might work and other times they
won't work. This change adjusts the defaul to have the Request an
empty ``Content`` (body) and sets it to NULL (which will trigger the
fallback to php://input) just if the Request is created by using
``Request::createFromEnvironment()``

Change-Id: I091033527fcd47f3bbaf4fbbd0ac9405195dd8d3
Releases: master
Resolves: FLOW-235

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/AbstractMessage.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/AbstractMessage.php
@@ -38,7 +38,7 @@ abstract class AbstractMessage
      *
      * @var string
      */
-    protected $content;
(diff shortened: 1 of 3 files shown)


dfeyer:
:+1: by reading

kitsunet:
Again too early PR with travis not reporting... Reopening...

PR #9, opened 2015-08-19
[FEATURE] Refactor Files::readDirectoryRecursively

kitsunet:
This implements ``Files::readDirectoryRecursively`` in a way
that is using a stack instead of recursive calls. This prevents
``too many files open errors`` as only one directory will be
opened at a time.

Additionally adds a new method ``getRecursiveDirectoryGenerator()``
if you don't need an array of files to manipulate but just want to
iterate the results. This reduces memory consumption.

Usage of ``readDirectoryRecursively`` was changed to
``getRecursiveDirectoryGenerator`` were possible.

Speed should also slightly improve in both cases.

Change-Id: I8b8928e7e5c7449c38d3cff4121fc682e1b4c5a9
Releases: master


dfeyer:
:+1:

aertmann:
:+1: by reading

kitsunet:
Just fixed a small oversight that lead to a (correctly) breaking unit test. Should be good to go.

PR #10, opened 2015-08-19
[BUGFIX] FlockLockStrategy should clean up lock files

kitsunet:
The FlockLockStrategy creates files to apply the lock on.
These files reside in the temporary folder but are never cleaned
on releasing the Lock that means the amount of files in this folder
will increase over time unless the folder is cleared manually.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
@@ -99,6 +99,7 @@ class FlockLockStrategy implements LockStrategyInterface {
 				$success = FALSE;
 			}
 			fclose($this->filePointer);
+			Files::unlink($this->lockFileName);
 		}
 
 		return $success;


TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
```diff
@@ -99,6 +99,7 @@ public function release() {
 				$success = FALSE;
 			}
 			fclose($this->filePointer);
+			Files::unlink($this->lockFileName);
```
albe: This creates a small race condition, for when a second process has just opened/created the file and not yet got a lock on it (being in between Ln 74 and 79). The file would then possibly be deleted before the other process gets ahold of the lock and hence fails.

dfeyer:
Why not a sort of Garbage collection based on file creation time ? Like on every X request, clear all file older than 2 minutes ? This will avoid the race condition.

albe:
Yeah, that's probably the safest way to go.

kitsunet:
I don't see that, what about long running requests? I mean yes maybe we could assume after 1 or 2 hours the file could be removed but still....

dfeyer:
Maybe appending the PID to the lock file, can also avoid the race condition and keep the code simple.

dfeyer:
I don't remember the history, but why did we put the lock on the cache file and not the real file, only to support thing like NFS right ? To support network share or windows, we should implement a other lock strategy, like based on redis with a nice garbage collector. But maybe I miss something ...

For the long running process, if a process need a lock on a file for a really long time ... it's not really a good design I think.

I just try to refresh the home page of the demo site, and I have more than 1'000 files in the Lock directory ... sounds crazy to me ;)

albe:
@kitsunet The longer the process is the running, the less likely it is that it will be triggered multiple times per second, hence the probability for the race condition occuring is less likely too. But generally yes, a GC also is not completely race condition free, plus it's relatively complex to implement and creates unpredictable spikes in IO, which may even be worse than a possible lock fail for some use cases.

So there's those options:
- flock the actual file that is being worked on, hence provide a special FileLock implementation which can be used for single server setups, is fast and creates no garbage. This only works for locking file accesses on non-NFS of course

- unlink the file after release, creating the small race condition, which might be acceptable for most low to mid-traffic sites

- do a gc every X requests, avoiding the race condition in most cases (a lock file that hasn't been accessed for X minutes is not accessed more than once per X minutes, hence the chance for the race condition is low enough to realistically never occur). This needs a change to actually touch the lock file after the lock was acquired, so that it's mtime matches last locking time

- provide a different lock strategy based on shared memory (see http://redis.io/topics/distlock, or http://php.net/manual/en/memcached.cas.php + https://en.wikipedia.org/wiki/Compare-and-swap)

So after all, there is again no silver bullet solution to the locking problem (in php), hence the chosen implementation of the lock should be configurable and properly documented for when to use which strategy

albe:
After digging around a bit on the net, interestingly enough I could not find any existing implementations of Lock strategies other than flock (or using database in drupal) in any of the famous PHP frameworks (symfony, zend, drupal, laravel)... could be worth providing something solid here

albe:
A few additional informations:

According to manpage and NFS FAQ flock **does** work on NFS since Linux **2.6.12** (http://man7.org/linux/man-pages/man2/flock.2.html#NOTES & http://nfs.sourceforge.net/#faq_d10).
Linux 2.6.12 was released **2005**, so I'd say it's safe to assume that every current server is running a Linux kernel that supports flock on NFS.
This means it's safe to use flock directly on the cache file in cache *FileBackend.

After more reading on the core problem, I found this stackoverflow thread again, which provides a race condition free implementation of locking a file that is unlinked after unlock:
http://stackoverflow.com/questions/17708885/flock-removing-locked-file-without-race-condition
This even works on windows, because there files can not be unlinked when they are opened by another process (see also http://blog.httrack.com/blog/2013/10/05/creating-deletable-and-movable-files-on-windows/), hence that race condition doesn't even exist on windows.
Btw: Is there still any valid reason why the FlockLockStrategy is a NOP on windows? It doesn't even use the supposedly non-working LOCK_NB flag.

Also, since Lock is supposed to be a generic Lock mechanism, supposed to be extended with different strategies with different multi-server locking behaviour (shared/non-shared lock), following must be assured for the cache *FileBackends:

- if a shared storage is used (NFS, etc.), the lock strategy *must* be shared the same way too

- if only local filesystem is used, the lock strategy *should not* be shared, or else writing behaviour is unnecessarily degraded to sequential writes among all servers

Given this, it would be necessary that *FileBackend is sure to use the FlockLockStrategy, while other application processes may require a different locking strategy (shared/non-shared). Therefore, the Lock class is no longer generically useable or the strategy pattern is irrelevant.

I would therefore suggest to decouple cache *FileBackend locks from the generic Lock class, meaning that flocks can directly be used on the cache file, avoiding additional lock files for that use case alltogether (less IOps) and making sure the lock will be shared exactly as the file is shared.

The Lock class could then be further extended to support different use cases based on the chosen strategy, like memcached/redis strategy for shared locks and flock/sema for non-shared locks, though that belongs to a different PR.

dfeyer:
Thanks @albe, I agree with your points, should work fine.

albe:
PS: See https://github.com/kitsunet/flow-development-collection/pull/1 for an implementation of the race condition free unlink (trying to get into the working-together on change workflow, so it's a PR on @kitsunet's fork)

kitsunet:
See #138

PR #11, opened 2015-08-19
[BUGFIX] Improve error message when running Flow with PHP < 5.5

bwaidelich:
This makes the ``Bootstrap`` compatible with earlier PHP version in
order to prevent the meaningless fatal error::

 PHP Parse error:  syntax error, unexpected T_CLASS, expecting T_STRING or T_VARIABLE or '$'

when running it with a PHP version lower than 5.5.
This also moves importing of other core classes (``PackageManager``,`
``Scripts``, ...) behind ``ensureRequiredEnvironment()`` so they won't
have to be adjusted.

Fixes: FLOW-370
Releases: master, 3.0

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
@@ -11,15 +11,8 @@ namespace TYPO3\Flow\Core;
  * The TYPO3 project - inspiring people to share!                         *
  *                                                                        */
 
-// Those are needed before the autoloader is active
-require_once(__DIR__ . '/ApplicationContext.php');
-require_once(__DIR__ . '/../Exception.php');
+// This is needed for the initial environment check and before the autoloader is active
 require_once(__DIR__ . '/../Utility/Files.php');
-require_once(__DIR__ . '/../Package/PackageInterface.php');
-require_once(__DIR__ . '/../Package/Package.php');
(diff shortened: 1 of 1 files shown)


robertlemke:
I wonder if we really need to keep our Bootstrap.php compatible with PHP versions lower than 5.5. I guess there will be another commit at some point which breaks compatibility again, and nobody will really notice. And since we officially only support PHP >= 5.5 (and declare it so in our Composer.json), I think it's just fine to not care about earlier versions.

As an alternativ I could imagine adding a check for PHP versions in our index.php.

:-1:

bwaidelich:
@robertlemke this is not about keeping the Bootstrap B/C but about making it easier for users to find out what it's wrong if they use a (CLI) PHP < 5.5.
But, yeah, checking in index.php would also be a solution ofc

kitsunet:
Having a check in index.php means we need to duplicate the php version (constant in bootstrap) as we cannot access it from there if the Bootstrap is not PHP 5.5 compatible.
I see no problem in keeping the bootstrap compatible, we should thin it out anyway ;)

albe:
+1 for thinned out bootstrap and +1 for informative errors in non-supported environments (composer isn't the only way to deploy a flow installation, it's just the most convenient one)

kitsunet:
Why did you close the PR? :(

bwaidelich:
Oops, didn't mean to - and now I can't reopen it b/c I deleted my fork. But I'll re-submit it to the right branch

PR #13, opened 2015-08-27
[TASK] Allow InstallerScripts to work with collection repository

kitsunet:
The InstallerScripts taking care of copying distribution resources
after composer operations would assume package paths that are
incompatible with the joined repositories. This will result in
missing Neos Routes on installing from the collection repository.

This is a first step after which the joined composer.json can be
used fully.

This deprecates the ["extras"]["typo3/flow"]["manage-resources"]
configuration for copying distribution resources. The deprecation
is in effect from 3.1 onwards and this option will be removed
three versions later. The new
["extras"]["neos"]["installer-resource-folders"] configuration
which is an array of directories containing installer resources
superseeds it with this changes and takes precedence. A typical
replacement would be:

  "installer-resource-folders": ["Resources/Private/Installer/"]

Releases: master, 3.0, 2.3

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
@@ -24,7 +24,7 @@ class InstallerScripts {
 
 	/**
 	 * Make sure required paths and files are available outside of Package
-	 * Run on every Composer install or update - most be configured in root manifest
+	 * Run on every Composer install or update - must be configured in root manifest
 	 *
 	 * @param CommandEvent $event
 	 * @return void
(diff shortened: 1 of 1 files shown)


kitsunet:
@kdambekalns this is the modified installer. Should be merged ASAP and also backported to the other branches.

TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
```diff
@@ -62,18 +62,40 @@ static public function postPackageUpdateAndInstall(PackageEvent $event) {
...
+
+			$evaluatedInstallerResources = FALSE;
+			if (isset($packageExtraConfig['neos']['installer-resource-folders'])) {
+				foreach ($packageExtraConfig['neos']['installer-resource-folders'] as $installerResourceDirectory) {
+					static::copyDistributionFiles($installPath . $installerResourceDirectory);
 				}
+				$evaluatedInstallerResources = TRUE;
+			}
+
+			if (!$evaluatedInstallerResources && isset($packageExtraConfig['typo3/flow']['manage-resources']) && $packageExtraConfig['typo3/flow']['manage-resources'] === TRUE) {
```
kdambekalns: Why not just use an else block instead of the "evaluated" flag?

TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
```diff
@@ -62,18 +62,40 @@ static public function postPackageUpdateAndInstall(PackageEvent $event) {
...
 			$installPath = $event->getComposer()->getInstallationManager()->getInstallPath($package);
-			if (isset($packageExtraConfig['typo3/flow']['manage-resources']) && $packageExtraConfig['typo3/flow']['manage-resources'] === TRUE) {
-				if (is_dir($installPath . 'Resources/Private/Installer/Distribution/Essentials')) {
-					Files::copyDirectoryRecursively($installPath . 'Resources/Private/Installer/Distribution/Essentials', getcwd() . '/', FALSE, TRUE);
-				}
-				if (is_dir($installPath . 'Resources/Private/Installer/Distribution/Defaults')) {
-					Files::copyDirectoryRecursively($installPath . 'Resources/Private/Installer/Distribution/Defaults', getcwd() . '/', TRUE, TRUE);
+
+			$evaluatedInstallerResources = FALSE;
+			if (isset($packageExtraConfig['neos']['installer-resource-folders'])) {
```
kdambekalns: This only happens `if (isset($packageExtraConfig['typo3/flow']))` which kind of defeats the purpose of using "neos" as namespace, no?

TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
kitsunet: would be a else if or a nested if in the else, seemed to be more complicate to read.

TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
kitsunet: ups, I missed the if above... Will fix that ;)

kitsunet:
See changed code. :)

kdambekalns:
:+1: by reading

TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
```diff
@@ -24,7 +24,7 @@ class InstallerScripts {
 
 	/**
 	 * Make sure required paths and files are available outside of Package
-	 * Run on every Composer install or update - most be configured in root manifest
+	 * Run on every Composer install or update - must be configured in root manifest
```
robertlemke: I somewhat liked "most" ;-)

robertlemke:
:+1:

PR #17, opened 2015-08-28
[TASK] Use yaml_parse instead of yaml_parse_file

kdambekalns:
If the PECL yaml extension is used, it was not possible to load YAML
configuration files from 'resource://some/path' while this worked for
the Symfony YAML parser.

To align the functionality, the file content is now fetched using
file_get_contents() (which uses stream wrappers) and then passed to
yaml_parse().


dfeyer:
:+1:

aertmann:
Is this as performant as the direct call? Could imagine this uses more memory. An alternative is to check for "resource://" before doing the workaround.

kdambekalns:
TBH I didn't check performance, and I guess it's about the same, since it needs to be loaded anyway.

So the question is, do we measure this now, or do we assume the gain (of making `resource://` work) is wirth it? The check you mention would be doable, but then again: is the check worth it, it might slow down the thing even more…

kdambekalns:
@aertmann added a check for `resource://` as suggested

aertmann:
:+1:  thanks

kdambekalns:
@dfeyer fine with you this way?

skurfuerst:
:+1: by reading. makes sense! Merging with previous votes, assuming @dfeyer has nothing against it ;)

PR #25, opened 2015-09-04
[!!!][BUGFIX] Store site lock in Flow temporary base path

bwaidelich:
Storing the site lock files in the system temporary directory
could lead to endless locks on some file systems.
This changes the site locks to be stored in the Flow temporary base
path again.

This is a breaking change because it removes the setting
``TYPO3.Flow.utility.environment.temporaryDirectoryBase`` in favor of
a new environment variable ``FLOW_PATH_TEMPORARY_BASE`` that allows for
changing the path if needed.

Background:

FLOW-348 introduced a new locking mechanism that stored lock files
in the systems default temporary folder determined via
``sys_get_temp_dir()``. On some systems files created there by the
PHP process could not be removed afterwards.

Related: FLOW-348
Fixes: FLOW-381
Releases: master, 3.0, 2.3

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
@@ -129,7 +129,7 @@ class SimpleFileBackend extends AbstractBackend implements PhpCapableBackendInte
 		$this->cacheEntryFileExtension = ($cache instanceof PhpFrontend) ? '.php' : '';
 
 		if ((strlen($this->cacheDirectory) + 23) > $this->environment->getMaximumPathLength()) {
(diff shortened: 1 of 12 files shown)


bwaidelich:
I failed to backport this one (for 2.3.x and 3.0.x) in a fully backwards compatible manner.
My idea was to replace
```
return rtrim(sys_get_temp_dir(), '/') . '/';
```
by something like
```
return FLOW_PATH_DATA . 'Temporary/' . str_replace('/', '/SubContext', (string)Bootstrap::$staticObjectManager->getContext()) . '/';
```

in `LockManager::getLockPath()` but `Bootstrap::$staticObjectManager` is not available during compile time and I can't find a way to determine the current context from within the LockManager.

albe:
Why not just inject the Context into the LockManager on construction? The Context is directly available through the Bootstrap itself.

hlubek:
This PR needs to target branch "2.3" by looking at the Releases footer

robertlemke:
:+1: in general for changing this back. As far as I can see, the `sys_get_temp_dir()` approach also would have bad conflicts when there are multiple website on a single server!

kitsunet:
:+1: By reading. 

@robertlemke The LockManager is fine with temp dir because the lock filename includes the hashed FLOW_PATH_ROOT

bwaidelich:
Closed in favor of #97 which applies this to the lowest affected branch (2.3)

PR #15, opened 2015-08-27
[TASK] Remove composer/installers requirement

kitsunet:
Removes the leftover composer/installers requirement from
Flow and replaces it with the neos/composer-plugin.

Releases: master, 3.0, 2.3

PR #30, opened 2015-09-08
[TASK] Add composer.json and template

kitsunet:


composer.json
```diff
@@ -0,0 +1,44 @@
...
+        "ext-SPL": "*",
+        "ext-json": ">=1.2.0",
+        "ext-mbstring": "*",
+        "ext-reflection": "*",
+        "doctrine/orm": "~2.4.0",
+        "doctrine/migrations": "~1.0.0",
+        "symfony/yaml": "~2.5.0",
+        "symfony/dom-crawler": "~2.5.0",
+        "symfony/console": "~2.0",
+        "composer/installers": "~1.0.2"
```
aertmann: should this be replaced with "neos/composer-plugin": "*" as in master?

related to https://github.com/neos/flow-development-collection/pull/15

aertmann:
Do you plan to synchronize this one with the master one? both have something the other doesn't..

composer.json
kitsunet: Good point, that was a separate change. Ok let me go though that and push everything needed to make this work...

kitsunet:
re composer plugin and installers you mean? Yes that should be synchronised, I'll take care.
This commit was a test of how this could go automatically. I basically didn't touch the composer.json manually.

aertmann:
:+1:

dfeyer:
Yesterday I install the neos dev collection, but the composer.json does not require neos/composer-plugin ... maybe i'm wrong but should we add this in the require section ?

kitsunet:
no, neos shouldn't require it. It's fine if only flow and the build essentials do. pushing a new chnage in a few minutes...

kitsunet:
There we go, lets merge this... Hard to test anyway

dfeyer:
:+1: by reading

PR #33, opened 2015-09-09
[TASK] Run only tests for Flow dev collection packages on Travis CI

hlubek:
Releases: master


hlubek:
Runs 414 functional tests as before, so nothing changes for now

PR #36, opened 2015-09-10
[TASK] Allow InstallerScripts to work with collection repository

kitsunet:
The InstallerScripts taking care of copying distribution resources
after composer operations would assume package paths that are
incompatible with the joined repositories. This will result in
missing Neos Routes on installing from the collection repository.

This is a first step after which the joined composer.json can be
used fully.

This deprecates the ["extras"]["typo3/flow"]["manage-resources"]
configuration for copying distribution resources. The deprecation
is in effect from 3.1 onwards and this option will be removed
three versions later. The new
["extras"]["neos"]["installer-resource-folders"] configuration
which is an array of directories containing installer resources
superseeds it with this changes and takes precedence. A typical
replacement would be:

  "installer-resource-folders": ["Resources/Private/Installer/"]

Releases: master, 3.0, 2.3

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
@@ -24,7 +24,7 @@ class InstallerScripts {
 
 	/**
 	 * Make sure required paths and files are available outside of Package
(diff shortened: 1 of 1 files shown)


kdambekalns:
Looks like the same change as for master, so :+1: by reading.

dfeyer:
:+1:

PR #38, opened 2015-09-10
[BUGFIX] Class loader includes some files twice

martin-helmich:
For some reason, the FLOW class loader includes some files twice.

This especially occurs when a Composer dependency declares an autoload file that contains function definitions (one example for a package like that being `guzzlehttp/promises`).

This commit fixes this issue by replacing `include` with `include_once`.

Fixes: FLOW-362


dfeyer:
:+1:

aertmann:
:+1:

martin-helmich:
Any updates on this? I'm starting to feel a little bit frustrated about this...

kitsunet:
@martin-helmich travis CI correctly says there are failing unit tests. I have the same locally. The three failing tests need to be adjusted because they use the same class file name which obviously won't work anymore with require_once (which is fine, just the test needs to respect that)

martin-helmich:
@kitsunet, the tests should be passing now. I changed the class names in the test case to be unique. Does that sound OK?

kitsunet:
Great, thanks! :)

kitsunet:
:+1:

PR #39, opened 2015-09-11
[BUGFIX] Remove neos composer plugin from reflection

kitsunet:
As Flow 2.3 still uses a blacklist approach for reflection and proxybuilding the neos composer plugin needs to be excluded instead of the composer installers package.


robertlemke:
:+1:

PR #40, opened 2015-09-13
[TASK] Improve exception output for subprocesses failing with fatal error

aertmann:
Improves the exception output in development context when a command executed
in a subprocess fails with fatal error. Previously the actual error could only
be found by looking in the system log or in some cases by running a CLI command.

Resolves: FLOW-382


albe:
+1 works nicely on windows and is just so convenient

dfeyer:
:+1:

kdambekalns:
:+1: by reading

hlubek:
:+1: by reading

bwaidelich:
:+1: tested on mac osx

aertmann:
:+1: tested on Debian w/ php-fpm (5.5), Apache 2.2

PR #41, opened 2015-09-13
TASK: Refactor the resource management to use PHP Generator

dfeyer:
Currently the Resource Management does not scale with a lots of resources.

This change introduces some API refactoring to use PHP Generators to
iterate during resource cleanup and publishing (affected are ``StorageInterface``
and ``TargetInterface``).

In ``CollectionInterface`` the return type of ``getObjects()`` changes from
``array`` to ``Generator`` which should be the same for practical purposes.

Some methods can now receive a callback function, the function will receive
the current iteration, as integer, and the currently processed object, as
TYPO3\Flow\Resource\Storage\Object. The callback is executed after each
object was processed.

Other interfaces (``StorageInterface`` and ``TargetInterface``) are not
changed to stay BC.

FLOW-294 #close


hlubek:
Hi Dominique,

I fully agree that a scaleable implementation is very important! But I'd really like to not change the public interfaces if possible (we need to take semver seriously). So by looking at the code the new argument in the interface could be fetched from `func_get_args()` in implementations and just documented in the interface (and included in the next major version). Not the nicest thing but maybe a compromise to change this while not being forced to break code.

TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
```diff
@@ -184,4 +190,15 @@ public function cleanCommand() {
 		}
 	}
 
+	/**
+	 * This method is used internal as a callback method to clear doctrine states
+	 *
+	 * @param integer $iteration
```
aertmann: @return void

aertmann:
Good point @hlubek makes, it would probably be pretty nasty code though :/

dfeyer:
Let's discuss this next week ... I think the func_get_args is an option, but @aertmann say ... nasty code ...

dfeyer:
Last push with:

- PSR-2 compliance
- Some bugfix in the callback execution parts
- More testing and improve the PR message to give more information on the change

So this change is only breaking because of:

> StorageInterface::getObjectsByCollection() can now receive an optional callback function. This method will not return an array but a \Generator

Has the generator can be use more or less like an array in PHP, it's not that breaking, and big improvement for scalability of Neos. I agree that we need to be serious about semantic versionning, but this change is breaking for really small side case.

:+1:

bwaidelich:
FYI: Not arguing against this change, but it is not only breaking because the return type changed but because of the modified method signatures of the interfaces.
This will lead to Fatal Errors for every package that implements `CollectionInterface`, `StorageInterface` or `TargetInterface`.

bwaidelich:
BTW: We could increase the backwards compatibility by omitting the optional callback parameter from the interfaces or by adding more specialized interfaces, but then again: it is quite easy to adjust the code that uses this and as long as it doesn't go into 3.0.x I think it's OK!?

dfeyer:
Close this one in favor or #59 for Flow 3.0.x as this change is not fully not breaking

TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
```diff
@@ -185,4 +191,16 @@ public function cleanCommand()
             }
         }
     }
+
+    /**
+     * This method is used internal as a callback method to clear doctrine states
+     *
+     * @param integer $iteration
+     */
```
kdambekalns: `@return void`

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceRepository.php
```diff
@@ -106,6 +122,62 @@ public function findByIdentifier($identifier)
...
+    public function iterate(IterableResult $iterator, callable $callback = null)
+    {
+        $iteration = 0;
+        foreach ($iterator as $object) {
+            $object = current($object);
+            yield $object;
+            if ($callback !== null) {
+                call_user_func($callback, $iteration, $object);
+            }
+            ++$iteration;
```
kdambekalns: `$iteration++` is more "common" and has the same effect here…

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Storage/FileSystemStorage.php
```diff
@@ -128,38 +128,43 @@ public function getStreamByResourcePath($relativePath)
...
-            $object->setStream(function () use ($that, $resource) { return $that->getStreamByResource($resource); });
-            $objects[] = $object;
+            $object->setStream(function () use ($resource) {
+                return $this->getStreamByResource($resource);
+            });
+            yield $object;
+            if (is_callable($callback)) {
+                call_user_func($callback, $iteration, $object);
+            }
+            ++$iteration;
```
kdambekalns: Same as above, `$iteration++` is more "common"

TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
```diff
@@ -84,7 +84,10 @@ public function publishCommand($collection = null)
             foreach ($collections as $collection) {
                 /** @var CollectionInterface $collection */
                 $this->outputLine('Publishing resources of collection "%s"', array($collection->getName()));
-                $collection->publish();
+                $target = $collection->getTarget();
```
kdambekalns: If userland code calls `$collection->publish()` will there be any optimization? I guess not, would that be possible? Or is the code (in Neos, first and foremost) publishing full collections just this code here?

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceRepository.php
```diff
@@ -106,6 +122,62 @@ public function findByIdentifier($identifier)
     }
 
     /**
+     * Iterator over an IterableResult and return a Generator
+     *
+     * This methos is useful for batch processing huge result set as it clear the object
+     * manager and detach the current object on each iteration.
```
kdambekalns: This is a false promise, no? The code here doesn't do any of that! It just makes it possible…

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Storage/FileSystemStorage.php
```diff
@@ -128,38 +128,43 @@ public function getStreamByResourcePath($relativePath)
...
+     * @param callable $callback Function called after each iteration
+     * @return \Generator<\TYPO3\Flow\Resource\Storage\Object>
      */
-    public function getObjects()
+    public function getObjects(callable $callback = null)
     {
-        $objects = array();
         foreach ($this->resourceManager->getCollectionsByStorage($this) as $collection) {
-            $objects = array_merge($objects, $this->getObjectsByCollection($collection));
+            return $this->getObjectsByCollection($collection, $callback);
```
kdambekalns: Maybe I do not understand generators well enough, but how can a `foreach` loop with a return statement work over all entries?

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Storage/PackageStorage.php
```diff
@@ -91,13 +93,17 @@ public function getObjectsByPathPattern($pattern)
...
-                    $object->setStream(function () use ($resourcePathAndFilename) { return fopen($resourcePathAndFilename, 'r'); });
-                    $objects[] = $object;
+                    $object->setStream(function () use ($resourcePathAndFilename) {
+                        return fopen($resourcePathAndFilename, 'r');
+                    });
+                    yield $object;
+                    if (is_callable($callback)) {
+                        call_user_func($callback, $iteration, $object);
+                    }
+                    ++$iteration;
```
kdambekalns: `$iteration++`

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Target/FileSystemSymlinkTarget.php
```diff
@@ -24,10 +25,11 @@ class FileSystemSymlinkTarget extends FileSystemTarget
     /**
      * Publishes the whole collection to this target
      *
-     * @param \TYPO3\Flow\Resource\Collection $collection The collection to publish
+     * @param CollectionInterface $collection The collection to publish
+     * @param callable $callback Function called after each resource publishing
```
kdambekalns: The callback parameters are undocumented!

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Target/FileSystemTarget.php
```diff
@@ -153,13 +153,14 @@ public function getName()
     /**
      * Publishes the whole collection to this target
      *
-     * @param \TYPO3\Flow\Resource\Collection $collection The collection to publish
+     * @param CollectionInterface $collection The collection to publish
+     * @param callable $callback Function called after each resource publishing
```
kdambekalns: The callback parameters are undocumented!

TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
dfeyer: Clearing the state, need to be done carefully, all object are detatched in doctrine, during a web request it's pretty dangereous (and even impossible currently). Check the code of the resource clean command, to see how I restore object to do the cleanup.

dfeyer:
Rebase on top of master, fix some bug in the resource:clean command (and add support to clearing broken thumbnail from this command, unfriendly for the user to play between the thumb CLI and the resource CLI to clean up resource stuff).

Currently running a last test session with ~500'000 resources.

dfeyer:
Successfully tested again. Can we spend some time to discuss the impact of the "breaking" change. Basically this change require some modification on cloud storage driver, but has I don't code any of those driver, maybe @robertlemke has more informations ?

kdambekalns:
This is still "current", right? So maybe @dfeyer and @robertlemke could get their heads together at some point?

dfeyer:
@kdambekalns Will try to get @robertlemke attention this afternoon ;)

dfeyer:
@robertlemke Tested with https://github.com/Flownative/flow-google-cloudstorage work fine, can you test with S3 ? And merge this one ;)

robertlemke:
Thanks a lot! So, I tested this with the AWS S3 plugin and it works great so far!

Just one thing I stumbled upon, which is not related to S3 but maybe to this change: when I forcefully delete resource data from the storage (directly from the S3 bucket) and run `resource:clean` afterwards, the cleanup might miss missing resources on the first pass, supposedly Image Variants etc. - depending on which resources I trashed, I had to run `resource:clean` a second time.

:fist:

robertlemke:
@dfeyer I tested `resource:clean` again based on master and could not reproduce the "needs second pass" bug. I tried 10 full import / randomly deleting files / clean cycles using the Filesystem Storage.

Do you have a chance to take a closer look at this?

kitsunet:
Couldn't reproduce the "second pass" problem so far. I had one strange occurance where I removed 1 resource file, the command said it would remove two resources and in the end 3 where removed. But maybe that was some relation (thumbnail of variant) that was not listed. So works fine for me.
:+1: 

Two side notes on the whole topic not necessarily related to the change:
- Removing Media files that are related to Nodes is seeminly horribly breaking, so maybe we should find another solution or make that someone less breaking.
- I definitely want those mentions of Media gone from the command. We need a strategy interface for relations to check and remove as well, so packages can plug to the resource:clean command.

robertlemke:
@dfeyer @kitsunet if we can't reproduce it anymore and @dfeyer still feels confident about the implementation I think we can merge this (gave my +1 as reaction). I assume that it's backwards compatible now, contrary to the earlier change in this thread, right?

dfeyer:
@robertlemke @kitsunet Try to reproduce the cleanup problem, but no way, every single run clean the correct resource. Try with and without the PR applied, with no difference. So I think we have a side case not correctly handled ... but nothing that we can improve in the PR. 

So :+1: for merging this one and I agree with the comment from @kitsunet, issue are now on Jira:

- https://jira.neos.io/browse/NEOS-1794
- https://jira.neos.io/browse/FLOW-442

Both are added the Unicorn sprint board

dfeyer:
Last push just merge master on top of this PR

kitsunet:
Will merge when travis is done :+1:

PR #42, opened 2015-09-21
[TASK] Compile list of properties information into proxy classes

kitsunet:
Reduce usage of ``ReflectionService`` during Runtime by compiling a list
of transient properties and variable type declarations into proxy classes
to be used during serialisation of objects.

Relates: NEOS-1294


kdambekalns:
:+1: Tests pass, looks good.

PR #43, opened 2015-09-21
[TASK] Lazily initialize ResourceManager

kitsunet:
The ``ResourceManager`` creates a bunch of objects and iterates over
the whole resource configuration on initialization. In some cases this
initialisation might not even be necessary so with this change we defer
it until the information is really needed.


robertlemke:
The additional `this->initialize()` calls are a bit ugly, but if that really speeds up Flow, it's probably worth it.

:+1: verified with Neos (using resources in all kinds of ways)

kitsunet:
Yeah, if you have a cached Neos page the resource urls are already in the cached HTML so there for example the thing doesn't need to be initialized which definitely saves time.

PR #46, opened 2015-09-22
[BUGFIX] Prevent overwriting source files

hhoechtl:
After switching from FileSystemSymlinkTarget to FileSystemTarget
the symlinks are still present in the targetPath and therefore Flow
tries to copy the stream content in the symlink (original file, same
stream) which results in a 0 byte file.
We handle this issue by removing the symlinks if present.

Fixes: FLOW-384


aertmann:
needs to be rebased

hhoechtl:
Done

kitsunet:
lol, we are to slow... Can we agree on the usefulness of this and then rebase and merge?
I totally see the problem and think we need it, even though I don't like that there is so much new code to fix it. I also don't see another way. So for now :+1:

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Target/FileSystemTarget.php
```diff
@@ -151,6 +151,25 @@ public function getName()
     }
 
     /**
+     * Checks if the PackageStorage has been previously initialized with symlinks
+     * and clears them. Otherwise the original sources would be overwritten.
+     *
+     * @param \TYPO3\Flow\Resource\Storage\StorageInterface $storage
```
aertmann: @return void

albe:
:+1: but what about the other way round? Switching from FileSystemTarget to FileSystemSymlinkTarget? Would that just work?

hhoechtl:
Sorry for the long delay. Yes, the switch from FileSystemTarget to FileSystemSymlinkTarget worked as far as I can remember.

hhoechtl:
Reopened pull request https://github.com/neos/flow-development-collection/pull/196

PR #50, opened 2015-09-23
[TASK] use CollectionInterface in resource management

hhoechtl:
Currently some methods use the Collection class instead of the
CollectionInterface.


kitsunet:
Seems like a a bugfix that could go into 3.0 to me

aertmann:
I agree, should go into 3.0

:-1:

aertmann:
Moved to #81

PR #54, opened 2015-09-23
TASK: Update CGL to match the PSR-2 standard

radmiraal:


aertmann:
Shouldn't new one page documents be created instead of deleted? There are quite some rules things PSR-2 doesn't cover.

bwaidelich:
@radmiraal ping ;)

kdambekalns:
The change changes the "one page" documents, both the ODT and the PDF. Seems all good.

PR #56, opened 2015-09-24
FEATURE: .env file can configure application environment variables

hhoechtl:
You can put a .env file (INI-format) into the root directory of your
Flow application to configure global variables like FLOW_CONTEXT.

FLOW-389 #close

### DIFF
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartII/Installation.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartII/Installation.rst
@@ -217,6 +217,15 @@ machine. Add the following line to your */etc/hosts* file
 	# You can specify a default context by activating this option:
 	SetEnv FLOW_CONTEXT Production
 
+Or if you are not using Apache as webserver, you can simply add a ``.env`` file
+to the root directory of your Flow application with the following content:
+
+*.env*:
+
+.. code-block:: none
+
+	FLOW_CONTEXT=Production
+
 Welcome to Flow
 ---------------
 
--- a/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/.gitignore
+++ b/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/.gitignore
@@ -8,3 +8,4 @@
 /Upgrading.rst
 /flow
 /flow.bat
+/.env
--- a/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/Web/index.php
+++ b/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/Web/index.php
@@ -20,6 +20,14 @@ if ($rootPath === false) {
     $rootPath .= '/';
 }
 
+if (file_exists($rootPath . '.env')) {
+    require($rootPath . 'Packages/Libraries/vlucas/phpdotenv/src/Dotenv.php');
+    require($rootPath . 'Packages/Libraries/vlucas/phpdotenv/src/Loader.php');
+    require($rootPath . 'Packages/Libraries/vlucas/phpdotenv/src/Validator.php');
+    $dotenv = new \Dotenv\Dotenv($rootPath);
+    $dotenv->overload();
+}
+
 require($rootPath . 'Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php');
 
 $context = \TYPO3\Flow\Core\Bootstrap::getEnvironmentConfigurationSetting('FLOW_CONTEXT') ?: 'Development';
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
         "symfony/yaml": "2.5.*",
         "symfony/dom-crawler": "2.5.*",
         "symfony/console": "2.*",
-        "neos/composer-plugin": "^1.0.1"
+        "neos/composer-plugin": "^1.0.1",
+        "vlucas/phpdotenv": "^2.0"
     },
     "replace": {
         "typo3/eel": "self.version",


robertlemke:
To be honest, I don't think that a new dependencies and so much additional code which is executed on every request is justified in this case.

My suggestion for those who can't use environment variables in their environment is to add a `putenv()` into `index.php` – it's rather uncritical to touch that file because it hardly ever chances over the releases. If in doubt, you can still `include()` or `require()` an additional PHP file.

:-1:

bwaidelich:
For the record: I came to the same conclusion as @robertlemke (see https://discuss.neos.io/t/easier-handling-of-context-variables/492/6). I could imagine a middle-way (e.g. a `.context` file) but phpdotenv seems to me like an overkill for that very case

kitsunet:
I will check the stuff code wise, but to me having this makes a lot of sense, especially for some (hosting) environments and requirements. I will write some more in discuss. I just would like to avoid to roll our own solution for this stuff.

robertlemke:
@bwaidelich something like the .context file would be a good compromise, but then again, there will be further environment variables one would like to set.

So, if we really need all these features and need an external file in some format, the library @hhoechtl might be a good choice. My point was rather if we want to open that box, add that (little) overhead, or if it could simply be solved by modifying index.php.

How about:

```php
if (file_exists($rootPath . 'env.php')) {
   include_once($rootPath . 'env.php');
}
```

robertlemke:
Just following up on this change - how about the code I suggested on Sep 24?

bwaidelich:
@robertlemke I think it is pragmatic enough, but it would need to be included for CLI requests, too. I'm still a bit concerned to introduce an alternative mechanism for all env variables where we really only need it for the Application context.. But probably we should continue discussing that on https://discuss.neos.io/t/easier-handling-of-context-variables/492/

kitsunet:
why would you only need it for the context? I still think it has value for the temp folder and other env variables.

bwaidelich:
repeating history? ;)
https://discuss.neos.io/t/easier-handling-of-context-variables/492/16?u=bwaidelich
My thinking was: We decided to go for environment variables for those "core options" and I think that works pretty nicely for most cases. Now we want to allow to use Flow/Neos in environments that don't allow to set the env variables. But is it really necessary to set "advanced options" like a different root path (btw: FLOW_ROOTPATH and FLOW_CONTEXT are currently the only env variables passed to the sub request).
Of course if we add a mechanism like this it will be easy to set all env variables but I'd say YAGNI - and with every flag you can set via two ways complexity and error-proneness increases.
But having written all this, it's probably not a biggy (even though setting env variables from a PHP script could be considered bad practice and even dangerous, but that's in the users hand anyways)

kitsunet:
haha ;) yeah I really don't mind that much. I guess we can also have a cookbook entry ala "you can install the env package yourself and include it in index.php and the flow.php script"....
If we did it in the core I would actually add an additional env variable that if set to true disables the loading of the .env file if you don't need it (for performance reasons).

bwaidelich:
hehe, I don't really care that much either - even if it might seem like it ;)
I wouldn't worry too much about performance though, the `file_exists` check should be very fast and I would expect such a file to *overrule* any env variable.. but that's exactly one of the questions that arises.. Anyways, in my opinion: :+1:  for a file that is included during bootstrap, :-1: for an additional library as suggested in the original PR ;)

kitsunet:
I would like to have a way to use ENV (and maybe such an .env file) for configuration of stuff in Flow please. I don't mind the library but could do without. Can we move forward towards a solution that is good for everyone?

hhoechtl:
Yes especially for docker we would need to think of a solution to use ENV variables.
Maybe I'll take care at Neos DevCamp

dfeyer:
> Yes especially for docker

It's not only about docker many cloud provider user ENV variable to provide you DB name / credentials, ... @hhoechtl I think that this can be quiet important for the next LTS, has it make Flow more easy to use in those modern hosting environnement.

dfeyer:
Currently we are evaluation Plateform.sh to host most of our project, and check how they handle ENV variables: https://docs.platform.sh/user_guide/reference/services/redis.html (down in the page, basically the env variable is a JSON BASE64 encoded, so it require processing and not just accessing a simple ENV live in docker.

radmiraal:
Just some small comments:
1) I'd be in favor of using .env (and follow common usage, so name it .env too)
2) Wouldn't it be enough (at least in case of database credentials with linked containers for example) to only parse the file before caching the config?

albe:
So is this still something to go for?

radmiraal:
imho this is obsolete as it can easily be accomplished by requiring helhum/dotenv-connector which works like a charm and hooks into composer

bwaidelich:
https://packagist.org/packages/helhum/dotenv-connector nice one

albe:
So I'll close this for now, if the desire comes up to get this back in, feel free to reopen.

helhum:
My package however only works, when at some point the vendor/autoload.php is included in the code, which Flow still does not, does it?

bwaidelich:
@helhum Hey there! Long time no see :)

> My package [...] only works, when [...] vendor/autoload.php is included

Yes, that's the case and we want to get rid of our custom loader for most parts in the long run

albe:
You can set the environment variable `FLOW_ONLY_COMPOSER_LOADER=1` to enable composer only autoloading since 4.3 https://github.com/neos/flow-development-collection/pull/925

But I see how this is a bad solution for enabling a package that allows to set environment variables in the first place :D this should be resolved with the upcoming major where this will be default behavior

kitsunet:
either way the composer autoloader is used. The only things that won't work because they are already read in index.php or the command line flow script are `FLOW_ROOTPATH` and `FLOW_CONTEXT`. Maybe something to move over to the bootstrap at some point...

helhum:
Nice and thanks for the quick feedback!

> The only things that won't work because they are already read in index.php or the command line flow script are FLOW_ROOTPATH and FLOW_CONTEXT

Ah, `FLOW_CONTEXT` is what I tried to set in the .env. Now I know why it didn't work :)

PR #57, opened 2015-09-24
[TASK] rename "TYPO3 Flow" to "Flow" and update URLs

hhoechtl:
Updated the documentation and remove the "TYPO3" prefix from "TYPO3 Flow".
Also the support channels, JIRA, Slack, etc. have been updated.


TYPO3.Flow/Documentation/TheDefinitiveGuide/PartV/CommandReference.rst
```diff
@@ -28,10 +28,10 @@ Package *TYPO3.FLOW*
 --------------------
 
 
-.. _`Flow Command Reference: TYPO3.FLOW typo3.flow:cache:flush`:
+.. _`Flow Command Reference: TYPO3.FLOW flow:cache:flush`:
 
-``typo3.flow:cache:flush``
-**************************
+``flow:cache:flush``
```
bwaidelich: that won't work as long as the package is not renamed

bwaidelich:
great initiative, but I guess you changed a little too much ;)
The command and ViewHelper references used to be created automatically anyways

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartV/CommandReference.rst
bwaidelich: ah, nevermind, the short form will work already.. Still, the references will probably be overridden by the automatic generation (triggered by https://github.com/neos/doctools ?)

robertlemke:
@bwaidelich @hhoechtl how do we proceed with this – can it be merged like this or do we need to remove the parts which are auto-rendered by the doc tools?

kitsunet:
If it is auto generated anyway I wouldn't mind merging it as is. But by now it has conflicts anyway. So I would say solve conflicts remove autogenerated stuff and merge.

kdambekalns:
But shouldn't we fix that in 2.3 and up, not only in master?

albe:
+1 for retargeting 2.3, this is some due task ;)

kdambekalns:
See #244 for the replacement.

PR #59, opened 2015-09-24
[TASK] Refactor the resource management to use PHP Generator

dfeyer:
Currently the Resource Management does not scale with a lots of
resources. This change introduce some API refactoring to use PHP
Generator to iterate during resource cleanup and publishing.

Some method can now receive a callback function, the function will
receive the current iteration, as integer, and the currently
processed object, as TYPO3\Flow\Resource\Storage\Object. The callback
is executed after each object processing.

All interfaces (``CollectionInterface``, ``StorageInterface`` and
``TargetInterface``) are not changed to stay BC.

Resolves: FLOW-294

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceRepository.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceRepository.php
@@ -11,7 +11,11 @@ namespace TYPO3\Flow\Resource;
  * The TYPO3 project - inspiring people to share!                         *
  *                                                                        */
 
+use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\ORM\Internal\Hydration\IterableResult;
+use Doctrine\ORM\QueryBuilder;
 use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Persistence\PersistenceManagerInterface;
 use TYPO3\Flow\Persistence\QueryResultInterface;
 use TYPO3\Flow\Persistence\Repository;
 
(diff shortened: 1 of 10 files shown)


dfeyer:
:+1:

hhoechtl:
The interface changes are also in https://github.com/neos/flow-development-collection/pull/50

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceRepository.php
```diff
@@ -106,6 +122,62 @@ public function findByIdentifier($identifier)
     }
 
     /**
+     * Iterator over an IterableResult and return a Generator
+     *
+     * This methos is useful for batch processing huge result set as it clear the object
```
Akii: typo

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Storage/FileSystemStorage.php
```diff
@@ -128,38 +128,43 @@ public function getStreamByResourcePath($relativePath)
...
             $object->setSha1($resource->getSha1());
             $object->setMd5($resource->getMd5());
             $object->setFileSize($resource->getFileSize());
-            $object->setStream(function () use ($that, $resource) { return $that->getStreamByResource($resource); });
-            $objects[] = $object;
+            $object->setStream(function () use ($resource) {
+                return $this->getStreamByResource($resource);
+            });
+            yield $object;
+            if (is_callable($callback)) {
```
Akii: NULL check is enough

Akii:
-1 Looks pretty breaking to me, 3.1 please.

bwaidelich:
I agree to @Akii - even if it can be considered a bugfix, the risk of regressions is high.
For example: Now we adjust the implementations but not the interfaces and *especially the invoking code*. So if someone writes a custom `CollectionInterface` implementation suddenly the `publish()` method is no longer called etc.
A way to work around this was to add a new interface that introduces a new method `getObjectsWithCallback()` or similar and then check it via `instanceof`..
But I agree that this would be an overkill and make things more complex - Let's get your initial version into 3.1 and try to release that as soon as possible instead

aertmann:
seems like it's too breaking, re-open #41 then?

dfeyer:
Close this one and reopon #41

PR #62, opened 2015-09-24
[TASK] Generate PSR-2 based code and some cleanup

dfeyer:
This change generate PSR-2 code, remove the TYPO3 brand name and update some invalid link.

### DIFF
--- a/TYPO3.Kickstart/Resources/Private/Generator/Controller/CrudControllerTemplate.php.tmpl
+++ b/TYPO3.Kickstart/Resources/Private/Generator/Controller/CrudControllerTemplate.php.tmpl
@@ -10,71 +10,79 @@ use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Mvc\Controller\ActionController;
 use {modelFullClassName -> k:format.ltrim(charlist: '\')};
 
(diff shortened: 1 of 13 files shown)


TYPO3.Kickstart/Resources/Private/Generator/Controller/CrudControllerTemplate.php.tmpl
```diff
@@ -10,71 +10,79 @@ use TYPO3\Flow\Annotations as Flow;
...
-	/**
-	 * @Flow\Inject
-	 * @var {repositoryClassName}
-	 */
-	protected ${modelName}Repository;
+    /**
+     * @Flow\Inject
+     * @var {repositoryClassName}
+     */
+    protected ${modelNam  e}Repository;
```
albe: two spaces inside the variable

albe:
Other than that little typo looks good

kdambekalns:
I think we should fix this in 2.3 an onwards, since we do PSR-2 on all branches. I tookthe freedom to create a new PR (https://github.com/neos/flow-development-collection/pull/175) targeting the 2.3 branch, fixing the two spaces @albe found and also updating the file-level docblocks that are generated.

PR #68, opened 2015-09-30
[TASK] Add example for if view helper inline syntax

robertlemke:
This adds another example for the if view helper inline syntax,
explaining the format for conditions when using comparisons.

Releases: 2.3, 3.0, master


kdambekalns:
:+1:

dfeyer:
:+1:

PR #71, opened 2015-10-02
[FEATURE] Use more sophisticated UUID generator

kitsunet:
Uses php-uuid extension if available and otherwise the
https://github.com/ramsey/uuid library which is a complete
solution for generating and working with UUIDs.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Algorithms.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Algorithms.php
@@ -10,6 +10,7 @@ namespace TYPO3\Flow\Utility;
 
 require_once(FLOW_PATH_FLOW . 'Resources/PHP/iSecurity/Security_Randomizer.php');
 
+use Ramsey\Uuid\Uuid;
 use TYPO3\Flow\Annotations as Flow;
 
 /**
@@ -23,12 +24,18 @@ class Algorithms
      * Generates a universally unique identifier (UUID) according to RFC 4122.
      * The algorithm used here, might not be completely random.
      *
+     * If php-uuid was installed it will be used instead to speed up the process.
+     *
      * @return string The universally unique id
-     * @todo check for randomness, optionally generate type 1 and type 5 UUIDs, use php5-uuid extension if available
+     * @todo Optionally generate type 1 and type 5 UUIDs.
      */
     public static function generateUUID()
     {
-        return strtolower(\Security_Randomizer::getRandomGUID());
+        if (is_callable('uuid_create')) {
+            return strtolower(uuid_create(UUID_TYPE_RANDOM));
+        }
+
+        return (string)Uuid::uuid4();
     }
 
     /**
--- a/TYPO3.Flow/composer.json
+++ b/TYPO3.Flow/composer.json
@@ -17,6 +17,8 @@
         "typo3/fluid": "*",
         "typo3/eel": "*",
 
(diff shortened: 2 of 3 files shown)


kitsunet:
@hlubek How is the travis config supposed to work? The flow development collection seems never to be instaleld via composer. That means any change to dependencies (like in this change) will not be tested correctly and fail...

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Algorithms.php
```diff
@@ -10,6 +10,7 @@
 
 require_once(FLOW_PATH_FLOW . 'Resources/PHP/iSecurity/Security_Randomizer.php');
```
aertmann: can't this be removed then? also should the file be kept?

kitsunet: Nope if you watch the full file you notice it is used in many more places (for actual random data). To remove that there is a separate pull request: #5
The thing is that the secure random number generator is WAY slower than the one we use currently (but also more secure and up to the current knowledge) so generating UUIDs with it is not feasible (functional tests take WAY longer then) so I figured it would make sense to use something else for the UUID generation and found this library. As it supports all types of UUIDs we can fullfil the todo for other types with it as well but in a first step I just wanted to replace it so the secure randomizer change can happen as well.

kitsunet:
Just to make this clear, this pull request works just fine, all tests pass, the problem is our travis setup that doesn't pull in the new dependency and therefore breaks.

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Algorithms.php
aertmann: Thanks for explaining, makes sense.

aertmann:
:+1: by reading

kitsunet:
Ok see also my explanation in slack #team I think it will take a bit until we can solve the dependency issue with travis, so this needs some local testing pls :)

albe:
+1 from reading, can't test the php-uuid part (windows)

bwaidelich:
:+1: by reading

kitsunet:
Can someone maybe test this locally and then merge it? Would be great :)

hlubek:
> the problem is our travis setup that doesn't pull in the new dependency and therefore breaks.

How to change this? Can we somehow use the composer.json merging tools for the CI to solve this issue?

kitsunet:
No, it's not about the merged composer json (that merge is already included in the change). It's because composer pulls the original collection and doesn't see the dependencies. What we would need is to clone the dev distribution and modify the composer.json in there so that it points to the pull request branch and then run composer install. That way composer should pick up the new dependency.

hlubek:
> What we would need is to clone the dev distribution and modify the composer.json in there so that it points to the pull request branch

I couldn't find a way to have a reference to the pull request from the composer.json version. We also need to make sure we keep the Travis build working for post-merge tests without a PR. But well, it's not directly related to this change here so we can continue that discussion somewhere else.

kitsunet:
Ok, but then please could someone manually test this change and get it merged, please :)

hlubek:
With `ramsey/uuid` package:

* Unit Tests :ok:
* Functional Tests :ok:

With `php-uuid` extension (removing `ramsey/uuid` to be sure it's used):

* Unit Tests :ok:
* Functional Tests :ok:

PR #75, opened 2015-10-04
TASK: Use symfony process for subrequests

kitsunet:
Utilizes the symfony process component to create subrequests.
That means Flow no longer requires `exec` to be available.

Additionally fixes a bug with statcaches that would try to
delete configuration files that no longer existed.


kitsunet:
Again the dependency problem...

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -541,22 +546,25 @@ public static function initializeResources(Bootstrap $bootstrap)
...
-            } else {
-                $exceptionMessage = sprintf('Execution of subprocess failed with exit code %d without any further output. (Please check your PHP error log for possible Fatal errors)', $result);
+        $environmentVariables = self::buildFlowEnvironmentVariables($settings);
+        $process = new Process($command, FLOW_PATH_ROOT, $environmentVariables);
+        try {
+            $process->mustRun();
+            $output = $process->getOutput();
+        } catch (ProcessFailedException $processException) {
+            $exceptionMessage = $process->getOutput() . PHP_EOL . $process->getErrorOutput();
+            if ($exceptionMessage === PHP_EOL) {
```
aertmann: wouldn't a trim and strlen count be better here? in case a command exits with error code but no output? no entirely sure that can happen, but this check seems rather flaky

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -252,6 +255,8 @@ public static function initializeProxyClasses(Bootstrap $bootstrap)
 
         // The compile sub command will only be run if the code cache is completely empty:
         if ($objectConfigurationCache->has('allCompiledCodeUpToDate') === false) {
+            OpcodeCacheHelper::clearAllActive(FLOW_PATH_CONFIGURATION);
+            OpcodeCacheHelper::clearAllActive(FLOW_PATH_DATA);
```
aertmann: can you explain why is this now necessary?

aertmann: shouldn't this be a separate commit and applied to earlier releases?

aertmann:
Nice one. Would have thought there were more exec code in Flow actually.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
kitsunet: Well either the process ended correctly then we wouldn't be in this catch block or it had an error. Now there could be the case that no output was provided, then this if is true and we throw a general message, otherwise we throw whateve the child process returned...

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
dfeyer: I agree that this need to be pushed as a bugfix for 3.0

kdambekalns:
Merged master and extracted the opcode cache bugfix.

kdambekalns:
To be honest, pulling in yet another dependency just to replace `exec` with `$process->run` is… a tad bit lame?

So deep down internally we switch `exec` against `proc_open`, which is not an advantage per se, right? But we gain more compatibility and error handling, at least the code in `Process.start` looks complex enough to hint at that. Is that the deal?

kitsunet:
@kdambekalns I thought about compatibility and error handling making this, it's debatable if it's worth to add another dependency... I had high hopes that process also allowed to "stream" input / output similar as we do in behat tests but that didn't quite work. If that worked it would be nice also for future possibilities. Maybe I will have a look at other CLI exec abstractions...

PR #76, opened 2015-10-04
[BUGFIX] Prevent filenames longer than 255 characters

kitsunet:
The temporary file created in the (Simple)FileBackend should not
exceed 255 characters which happened with the addition of
microtime and process id. Therefore the cache identifier is now
hashed in the temporary file name.

Fixes: #55306

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
@@ -430,6 +430,6 @@ class SimpleFileBackend extends AbstractBackend implements PhpCapableBackendInte
      */
     protected function generateTemporaryPathAndFilename($entryIdentifier)
     {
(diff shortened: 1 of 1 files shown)


kitsunet:
See https://review.typo3.org/#/c/27043/ for original discussion. @kdambekalns maybe you could have a look also in regards to the original discussion.

aertmann:
Should this only be targeted at master?

kitsunet:
@aertmann we can discuss that. If we decide it is relevant it should probably go into 2.3 and up...

aertmann:
@kitsunet would suggest just to close and push for 2.3, don't see why it wouldn't be relevant?

kitsunet:
Great, it's semi irrelevant because this method is actually no longer called...
Will provide a change to remove it. Still the basic problem persists.
We throw an exception
`  if ((strlen($this->cacheDirectory) + 23) > $this->environment->getMaximumPathLength()) {`
but I am not sure what this magic 23 was meant to be. Currently we just use the given cache identifier, so that might be way longer than 23 chars and even if using MD5 it would still be longer...

aertmann:
Replaced by #152 right?

kitsunet:
Correct

PR #77, opened 2015-10-04
[TASK] Defer initialization of PersistenceManager

kitsunet:
Removes initialization of the PersistenceManager from the bootstrap as
that can be done lazily. For backwards compatibility the
``PersistenceManagerInterface::initialize`` method was removed but is
still called when available. All existing code should work as expected
but it is advisable to switch to using ``initializeObject`` instead. As
the ``initialize`` method will no longer be called starting from the next
major Flow version.
A separate breaking change will remove the backwards compatibility layer.

Before ``persistAll`` is called after a runtime run finished it is first
checked if there is a ``PersistenceManagerInterface`` instance registered
because otherwise  an unnecessary initialization of the PersistenceManager
would be triggered.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
@@ -492,18 +492,6 @@ class Scripts
     }
 
     /**
-     * Initializes the persistence framework
-     *
-     * @param Bootstrap $bootstrap
-     * @return void
-     */
-    public static function initializePersistence(Bootstrap $bootstrap)
-    {
-        $persistenceManager = $bootstrap->getObjectManager()->get(\TYPO3\Flow\Persistence\PersistenceManagerInterface::class);
-        $persistenceManager->initialize();
-    }
-
-    /**
      * Initializes the session framework
      *
      * @param Bootstrap $bootstrap
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/PersistenceManagerInterface.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/PersistenceManagerInterface.php
@@ -27,13 +27,12 @@ interface PersistenceManagerInterface
      */
     public function injectSettings(array $settings);
 
-    /**
-     * Initializes the persistence manager, called by Flow.
-     *
-     * @return void
-     * @api
+    /*
+     * The "initialize" method has been deprecated and was already removed from the interface.
+     * It is still called until the next major Flow version, but you should replace it by
+     * using "initializeObject" instead.
+     * TODO: Remove for next major Flow version.
      */
-    public function initialize();
 
     /**
      * Commits new objects and changes to objects in the current persistence session into the backend.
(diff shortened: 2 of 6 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -490,18 +490,6 @@ protected static function compileDoctrineProxies(Bootstrap $bootstrap)
...
     /**
-     * Initializes the persistence framework
-     *
-     * @param Bootstrap $bootstrap
-     * @return void
-     */
-    public static function initializePersistence(Bootstrap $bootstrap)
-    {
-        $persistenceManager = $bootstrap->getObjectManager()->get(\TYPO3\Flow\Persistence\PersistenceManagerInterface::class);
-        $persistenceManager->initialize();
```
kdambekalns: This is no longer called with this change, so the call needs to be called elsewhere. `initializeObject()` might be enough, even…

kitsunet:
The new commit should work with the CI BUT :-1: for now as that would be a breaking change (we remove the call to PersistenceManagerInterface->initialize()). Will follow up with a BC version.

robertlemke:
Thank you for giving it a try to make it backwards compatible!

kitsunet:
This version is backwards compatible and at the same time allows for easy migration to ``inititializeObject`` without leaving unnecessary code by removing initialize from the PersistenceManagerInterface.

So :+1: please review and merge.

TYPO3.Flow/Classes/TYPO3/Flow/Persistence/PersistenceManagerInterface.php
```diff
@@ -27,13 +27,12 @@
...
-     * Initializes the persistence manager, called by Flow.
-     *
-     * @return void
-     * @api
+    /*
+     * The "initialize" method has been deprecated and was already removed from the interface.
+     * It is still called until the next major Flow version, but you should replace it by
+     * using "initializeObject" instead.
+     * TODO: Remove for next major Flow version.
      */
```
robertlemke: I think we shouldn't add that "TODO Remove for next major Flow version" tag, but instead add a `@deprecated` annotation to the `initialize()` method which allows us to automatically search for deprecated methods.

kitsunet: So you mean leaving the initialize() in the interface? I wanted to avoid that to make migration more meaningful, otherwise you have to keep an empty initialize until we finally remove it...

robertlemke: no, removing it from the interface is fine – I meant adding `@deprecated` to the method in the class, so we can include it into the list of deprecated methods.

kitsunet: But it's gone from the class an replaced by initializeObject... See the change in 
TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php

PR #78, opened 2015-10-07
[BUGFIX] Fix functional test by explicitly naming sequence

kdambekalns:
The auto-generated name of a sequence exceeds the maximum length, is
truncated and thus duplicates an already existing name in the schema.
This is solved by manually giving a name to the sequence.

This bug affects only PostgreSQL and is triggered by a functional test
fixture.


aertmann:
:+1: don't know much about this, but seems okay :)

PR #79, opened 2015-10-07
[BUGFIX] Skip commonObjectIsPersistedAndIsReconstituted() on PgSQL

kdambekalns:
The object we assign in the test is `Persistence\Fixtures\CommonObject`,
containing a protected property. Doctrine stores `object` as
`serialize()`d PHP data in a text column. Which doesn't work on
PostgreSQL, since the string is truncated at the first `null` byte,
used in the serialised data to mark the protected property.

The official fix is to use a custom datatype if you need it, for the
test I decided to skip it if on PostgreSQL.

Related: FLOW-396

### DIFF
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
@@ -1006,6 +1006,15 @@ possibly inconsistent query results.
 	reading queries. However, this might lead to higher load on the master database and should be
 	well observed.
 
+Known issues
+------------
+
+* When using PostgreSQL the use of the ``object``, and ``array`` mapping types is not possible, this is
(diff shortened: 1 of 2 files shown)


aertmann:
:+1: okay with me, if you don't think Flow should handle it automatically

kdambekalns:
Well, that "automatic handling" would be needed in the Doctrine type, but since that would not be BC it won't be accepted. So what this changes does is: it skips a test for something that is "officially acknowledged as being broken".

What we could do, though, is add "known issues" with PostgreSQL to the Flow documentation.

kdambekalns:
Just rebased that on top of the current 3.0 branch and force-pushed. The change looks way less intimidating now. :)

aertmann:
I see, documentation would be good

kdambekalns:
Well, I amended the documentation…

PR #80, opened 2015-10-07
[BUGFIX] Use jsonb in JsonArrayType on PostgreSQL

kdambekalns:
The `JsonArrayType` in Flow inherits from the same type in Doctrine DBAL.

That type uses the `json` format, which is not comparable in PostgreSQL,
something that leads to issues if you want to use `DISTINCT` in a query.
Starting with PostgreSQL 9.4 the `jsonb` type is available, and the DB
knows how to compare it, making distinct queries possible.

Neos uses that, so the easiest way to fix this is to always use `jsonb`
for our custom type. The downside: the minimum supported version is
raised to 9.4.

Related: FLOW-396
Related: NEOS-1627

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 language: php
+addons:
+  postgresql: "9.4"
+services:
+  - postgresql
 matrix:
   fast_finish: true
   include:
@@ -8,8 +12,10 @@ matrix:
       env: DB=mysql
     - php: 5.6
       env: DB=mysql BEHAT=true
-#    - php: 5.6
-#      env: DB=pgsql
+    - php: 5.6
+      env: DB=pgsql
+    - php: 5.6
+      env: DB=pgsql BEHAT=true
     - php: 5.5
       env: DB=sqlite
 cache:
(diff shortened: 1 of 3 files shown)


kdambekalns:
See https://discuss.neos.io/t/supporting-postgresql-again/536

dfeyer:
I'm in favor to rise the version number to 9.4, it open a lots of improvement in json handling (indexes, ...)

Does this change need some DB migration ?

kdambekalns:
Rebased.

This doesn't need a DB migration on itself. If this used in code that actually runs on PostgreSQL a migration might be needed, but a search for `type="flow_json_array"` results in a single page of results, and all relate to our own use of that or the "invention" of the type as such. So IMO the risk of breaking anything with this for anyone is very low.

aertmann:
still need to decide on @dfeyer's point about raising the minimum version to 9.4 in a patch version..

kdambekalns:
As I wrote, I see the risk as very low and would consider the current behavior in 3.0 as broken.

kdambekalns:
Please see https://discuss.neos.io/t/supporting-postgresql-again/536/4?u=kdambekalns for some background on this.

kitsunet:
:+1: Fine for me, AFAIR @aertmann you had a server back then that had postgresql but not a version that supported jsonb which was the cause I didn't make that the default.

aertmann:
good point @kitsunet, still do actually (9.1 on debian wheezy, will update soon though).. but if it's necessary then we have to and considering that postgres is broken for 2.0 and has been all along there's low risk of affecting anybody so :+1: from my side

PR #83, opened 2015-10-09
[TASK] Improve readability of ReflectionService

kitsunet:
This is a mild refactoring of the ``ReflectionService`` which
doesn't change any functionality but reduces method complexity
and size by splitting and reordering of code.


aertmann:
bit hard to review in GH :/

kitsunet:
What can I say... Lot to refactor there...

albe:
:+1: from reading, some new method names might be debatable, but I don't really mind them as they are clear enough IMO. All in all a very good improvement for code clearity

PR #84, opened 2015-10-09
[TASK] Test Flow 2.3 on PHP 5.3

kdambekalns:
This change sets the lowest tested PHP version to 5.3 for Flow 2.3


kdambekalns:
Hey, @aertmann and @kitsunet - PHP 5.3 for Flow 2.3 to be tested. :)

kdambekalns:
Hm, it seems on 5.3 "Antarctica/Troll" is not valid…

> Warning: date(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier.

aertmann:
Why not add all versions? If we run into problems with unavailable VMs, we can change it..

kdambekalns:
I would check the lowest and highest supported (by us), adding all raises the time taken for the tests massively…

neos-project:
Tests run in parallel, so it won't be slower by testing more environments unless there are no VMs available. And more tests are better than fewer. Something could be broken in between.

kdambekalns:
Hi neos-project! :)

Ok, so maybe I got fooled by the output at Travis CI, but it says

>  ran for 12 min 18 sec

for https://travis-ci.org/neos/flow-development-collection/builds/84506936 - which is the tum of the times the individual tests took.

aertmann:
lol, wrong window ^^

bwaidelich:
Any Idea on how to solve this?

aertmann:
The problem is that the Behat package is using traits, which is not supported on 5.3 right?

bwaidelich:
Ah, I see.. Maybe we could at least activate unit and functional tests for 5.3!?
Then again, EOL for PHP 5.4 has been [announced](http://php.net/supported-versions.php) a month ago

kitsunet:
Still we need to ensure it works with 5.3 until we put 2.3 to EOL

kdambekalns:
Right, as long as we support a version that is supposed to run on PHP 5.3, we should make sure it does. The solution would be to use a Behat package that does not (yet) use traits, I guess we could tag some older state?

aertmann:
Right good idea, the traits are only necessary for Neos actually. So it's easy for Flow, but a bit harder to solve for Neos where we should do it as well.

kdambekalns:
Ok, added all supported PHP versions as suggested by @aertmann, and removed Behat completely, since we don't use that for Flow. Hm. Now that I think of it, we *do* run Behat for Neos 1.2, do we? Bummer…

kdambekalns:
Hm, and the build fails due to the segmentation fault? The tests do pass, eh?

aertmann:
@kdambekalns hey actually not sure about that now, considering we'll test 5.3, 5.4, 5.5, 5.6 for 2.3/1.2 and 5.5, 5.6, 7.0 for 3.0/2.0.. although I recommend we start with it and then change it if it proves to be an issue

aertmann:
we only run behat in the latest php version since it's super slow and the risk for php conflicts is very low so that's fine.. the only problem might be that when behat is installed it causes issues, but we can probably uninstall it for 5.3 specifically

kdambekalns:
I'll close this and open a new PR, since the branch name is so wrong by now…

PR #87, opened 2015-10-10
[BUGFIX] Ignore race condition on cached Configuration include

kitsunet:
It can happen that the cache include file was already removed
by a subrequest before getting to the unlink. The warning that
would follow can safely be ignored.


aertmann:
a test needs to be adjusted

kitsunet:
Fixed it :) Didn't expect a test that checks the generated code, but well...

kitsunet:
Btw. this race condition happens for me if I delete Data/Temporary and do a web request.

albe:
:+1: from reading, makes sense

aertmann:
should this go into earlier versions?

dfeyer:
:+1: but agree with @aertmann that should be backported

kitsunet:
Replaced by #135

PR #88, opened 2015-10-11
[TASK] Properly encode Uris

kitsunet:
Encode uri path and arguments according to HTTP specifications
on casting a `Uri` object to string.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Uri.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Uri.php
@@ -390,8 +390,17 @@ class Uri
                     $uriString .= (isset($this->port) ? ':' . $this->port : '');
             }
         }
-        $uriString .= isset($this->path) ? $this->path : '';
-        $uriString .= isset($this->query) ? '?' . $this->query : '';
+
+        if (isset($this->path)) {
+            $pathSegments = explode('/', $this->path);
+            $pathSegments = array_map('rawurlencode', $pathSegments);
+            $uriString .= implode('/', $pathSegments);
+        }
+
+        if (isset($this->arguments) && $this->arguments !== array()) {
+            $uriString .= '?' . http_build_query($this->arguments, null, '&', PHP_QUERY_RFC3986);
+        }
+
         $uriString .= isset($this->fragment) ? '#' . $this->fragment : '';
         return $uriString;
     }


kitsunet:
We need to discuss this. But IMHO this is needed to create fully correct URLs.

bwaidelich:
I think my comments in https://review.typo3.org/40547 still apply

kitsunet:
Right, they do, I put this up as WIP to not loose the general idea. IMHO it's still a point to discuss if the URL object shouldn't take care of encoding in general because it's the only place that can responsibly do so. Maybe not in this way but we have no place to get a properly encoded url from.

kitsunet:
So I would like to hear opinions here @kdambekalns @robertlemke @bwaidelich IMHO the Uri object should have some build-in way to encode the resulting string properly. Basically having a way to put unencoded url in and get encoded out.
The implemention in here is just a sketch, can be a separate method as well...
PSR-7 implementations usually do this. I just want to know if you agree that the Uri object should have this capability or not.

bwaidelich:
@kitsunet In the meantime my comments on review.typo3.org have been vanished, but maybe that's a good thing so I can judge without prejudice ;)
To keep it short: Yes, I think it's fine that the `Uri` class knows about encoding/decoding but it should be done explicitly to avoid side effects (like double encoding)

aertmann:
@bwaidelich: hmm doesn't seem like the original change in Gerrit had any comments http://gerrit.neos.io/#/c/40863/ though

bwaidelich:
mh, I'm pretty sure it had a whole thread of discussions in the original PS.. anyways, not so important IMO

mficzel:
I fear this will cause trouble if the path already has url-encoded segments in wich case they will be encoded twice. The PSR-7 interface states 

>   The value returned MUST be percent-encoded, but MUST NOT double-encode any characters. To determine what characters to encode, please refer to RFC 3986, Sections 2 and 3.3.

I am not sure wether we can safely assume that pathes or query arguments are encoded or not.

kitsunet:
Currently there is no correct way to de/encode urls, but this is old and would be superseded by the psr-7 change anyway.

mficzel:
I checked and those lines are not (yet) changed in the psr7 pr ... but you are right that it should be adjusted there.

bwaidelich:
Mh, need to get into this one again.. Can anyone come up with a good expected vs actual end-to-end behavior so we can reason about this?

bwaidelich:
Mh, need to get into this one again.. Can anyone come up with a good expected vs actual end-to-end behavior so we can reason about this?

kitsunet:
My wish would be -> incoming urls are properly decoded, inside an app I don't need to care about encoding, displaying urls is per default done properly encoded but if I need to, I can get the unendcoded value.

bwaidelich:
> My wish would be -> incoming urls are properly decoded
Sorry if I'm thick but I don't really understand what you mean with "incoming URLs"?
Can you describe a scenario that is currently impossible or hard to do? (just to find out if there may be other ways around)

PR #90, opened 2015-10-13
[BUGFIX] transmitting value objects by identifier

kaystrobach:
given:

```
/**
	 * @param Article\Keyword $keyword
	 */
	public function showKeywordAction(Article\Keyword $keyword) {
….
```

```
<f:link.action action="showKeyword" arguments="{keyword:keyword}" class="btn btn-default btn-sm“>

...
```

```
/**
 * @Flow\ValueObject
 */
class Keyword
{
    /**
     * @var string
     */
    protected $keyword;

    public function __construct($keyword) {
        $this->keyword = $keyword;
    }
...
```

url looks like: `/article/showkeyword?keyword%5B__identity%5D=761842cbf8e2c16abac74a60aaeada1b6203f4a9`

and the exceptions are:

#1297759968: Exception while property mapping for target type "SBS\Materialdatenbank\Domain\Model\Article\Keyword", at property path "": Missing constructor argument "keyword" for object of type "SBS\Materialdatenbank\Domain\Model\Article\Keyword“.

#1268734872: Missing constructor argument "keyword" for object of type "SBS\Materialdatenbank\Domain\Model\Article\Keyword".

I do use FLOW 3.0.0 in that project

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/PersistentObjectConverter.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/PersistentObjectConverter.php
@@ -159,9 +159,12 @@ class PersistentObjectConverter extends ObjectConverter
     {
         if (is_array($source)) {
             if ($this->reflectionService->isClassAnnotatedWith($targetType, \TYPO3\Flow\Annotations\ValueObject::class)) {
-                // Unset identity for value objects to use constructor mapping, since the identity is determined from
-                // property values after construction
-                unset($source['__identity']);
+                if (isset($source['__identity']) && (count($source) > 1)) {
+                    // @TODO fix that in the URI building and transfer VOs as values instead as with their identities
+                    // Unset identity for value objects to use constructor mapping, since the identity is determined from
+                    // property values after construction
+                    unset($source['__identity']);
+                }
             }
             $object = $this->handleArrayData($source, $targetType, $convertedChildProperties, $configuration);
             if ($object instanceof TargetNotFoundError) {


kitsunet:
The thing is, that I am not sure this is the best way to fix it. Actually the other side should be fixed. A value object should be represented by value(s) and not by identity. In this case the identity is even more a necessary technical construct that should be kept internally.

IMHO the url should look like:
article/showkeyword?keyword%5Bkeyword%5D=something
I guess that makes a change in the argument mapping for links necessary. Maybe @bwaidelich or @skurfuerst have an opinion on that?

kaystrobach:
Problem is that the current state is really a blocker not just for one project. Using the identities like for entities also makes the routing easy to understand. I personally prefer the identity, ... But I'm open for any kind of fix ... At least the identity should work ...

----- Originale Nachricht -----
Von: "Christian Müller" <notifications@github.com>
Gesendet: 15.10.2015 - 21:05
An: neos/flow-development-collection <flow-development-collection@noreply.github.com>
Betreff: Re: [flow-development-collection] [BUGFIX] transmitting value objects by identifier (#90)

Akii:
-1 for using identity hash.

If at all, you need to pass its values. Doing that automatically is probably not possible due to the constructor argument / properties mismatch.

The fix here could be rather easy: arguments="{keyword: keyword.keyword}" (with a custom TC).
Or implement a TC that converts from __identity to value object manually. I'd prefer to not see this in core though, for the reasons @kitsunet already stated.

albe:
This

    arguments="{keyword: keyword.keyword}"

is IMO a bad idea, because then you need to properly type ALL attributes of a VO whenever you want to submit it via URL. Now guess what happens when you change the VO and add a new property. Not fun. Also, from the Dev POV you actually want to submit the VO and not it's properties.

I see it that way:
Technically, from pure DDD, it is correct that the VO should be submitted/referenced by it's properties. But, since we have it and the identifier is a hash of the properties (i.e. a different representation of them), it is ok to just submit that to reference the specific VO (think of a VO where there are a multitude of properties and/or a property contains a lot of text, in which case you probably don't want to submit all those bytes every time just because it is more correct).
Yes, it is a technical implementation detail that we even have the identifier and that should be transparent to the user, but that does not necessarily mean that we must translate all VOs to their properties for form submission/URI generation. After all, those uri parameters are to be considered transparent to the user, too.
IMO in fact both should be possible, but the default should be to use the identifier, because it is more versatile.

This in turn leads back to the original problem: If a VO is submitted with just it's identifier, that must be correctly translated back to the original VO. Basically, that is what the commit does.

I would say this should be merged as is, because it fixes a critical misbehaviour on our own technical construct. We can then still create a new PR/issue that targets to streamline VO handling in request arguments in a more clean way.

Akii:
`arguments="{keyword: keyword.keyword}"` was a suggestion for this particular case.

The VO here is simply misused and can be worked around by a simple type converter. There is no need to change the core for that. Another approach would be to use an Entity here.

Furthermore, the change would encourage others to use VOs like Entities, even more so than that is already the case. Again, the issue is not that it's not impossible to implement (that's like tripple negative right here :o). It violates the idea behind VOs. Depending on a technical necessity will bring you doom once they're e. g. embedded (as Doctrine embeddable). Then you run out of value hashes. This leads to inconsistent behaviour, and so on.

Should the team decide to merge this, I'd like to nitpick on `(count($source) > 1)` which is a weird thing to do, and probably better replaced by either `strlen === 0` or a simple `!== ''`.

```
Now guess what happens when you change the VO and add a new property. Not fun. Also, from the Dev POV you actually want to submit the VO and not it's properties.
```

I'd actually like to see a string cast when attempting to "link" a VO. Or some sort of serialization. Not always feasible, as you said, but a good indicator that you might not want to stick VOs into an url.

kaystrobach:
@Akii so you would just transfer the string and then cast that string (and maybe other properties) into the value object again? Isn't that a lot more expensive in the database instead of simply asking the DB for that simply primary key uuid? especially as doctrine does not add indexes over all fields here automatically (i know i can do that manually ...).

So searching for the UUID is a lot cheaper then searching for a bunch of objects. and the idea in that case is not so uncommon it is simply show me all the articles containing a relation to a given keyword / tag, whatever you call it ... and from the definition of a tag, it should never change, so it should be a value object ...

kaystrobach:
well one option would be to not have a uuid in the keyword table at all and use all fields together as primary key, but this will create a lot more overhead in the database.

regarding the ```count() > 1``` i added this one to ensure, if there are other properties present in the VO it won't have the uuid present, as it was before and will get a new one (or an existing one)

Akii:
There is no need to go to the database, it's just values. In your case, all you have to do is to get the property mapper to take your string and create a new Keyword instance. `new Keyword('thing from the url');`

Ah good point with the count, overlooked that it's the $source not the identifier (I thought you counted the identifier).

afoeder:
@kaystrobach: 
>  Isn't that a lot more expensive in the database instead of simply asking the DB for that simply primary key uuid?

It's not about making it simple or "neat", it's about making it *correct* IMO. That's why rules and definitions like Value Objects exist: following them prevents you inherently from architectorial mistakes. If you break these rules, you will get into trouble, maybe not now, maybe never, but maybe later.
A value object *must not* have any kind of identity, but define itself via its properties. This pattern must be respected all over the place wherever value objects are touched.

albe:
> Furthermore, the change would encourage others to use VOs like Entities, even more so than that is already the case.

Well, as said, IMO it's supposedly transparent to the developer how the VO gets submitted. So unless he gets that exception he probably won't even notice that the identifier is used.
But yes, I see that, but I value fixing the current misbehaviour on top of the already flawed implementation more.

> I'd actually like to see a string cast when attempting to "link" a VO. Or some sort of serialization.

+1 that, but I've yet to see a perfectly transparent and flawless way to serialize objects. It would work if you restrict VOs to not contain relations, but ...

> Isn't that a lot more expensive in the database instead of simply asking the DB for that simply primary key uuid?

No, because in the first case you don't have to go to the database at all.

kaystrobach:
BTW. the suggested create Value Object via constructor fails sometimes with the following exception:

image

The keyword used in the image is definitely existing, so this problem is getting even worse :cry:

kaystrobach:
btw. this is the RFC which introduced the problem https://forge.typo3.org/issues/29905

albe:
See #107 regarding that last exception.

albe:
:+1: from me - anyone still against merging this bugfix as is? Maybe retarget to 2.3, since it's a pretty essential bugfix?

kaystrobach:
Would liek to see it in 2.3 ;)

albe:
@kaystrobach Can you recreate the PR targeting the 2.3 branch then? Then we can close this one and upmerge to other releases and master from there.

albe:
Closing in favor of #205 which targets 2.3

PR #91, opened 2015-10-13
[TASK] Use DateTimeInterface in Flow

kitsunet:
Since Flow requires PHP 5.5 it should use and support the ``\DateTimeInterface``.
With support for it the ``Now`` class can extend from ``DateTimeImmutable`` to
prevent changes to the singleton throughout a request.

### DIFF
--- a/TYPO3.Eel/Classes/TYPO3/Eel/Helper/DateHelper.php
+++ b/TYPO3.Eel/Classes/TYPO3/Eel/Helper/DateHelper.php
@@ -44,7 +44,7 @@ class DateHelper implements ProtectedContextAwareInterface
      */
     public function format($date, $format)
     {
-        if ($date instanceof \DateTime) {
+        if ($date instanceof \DateTimeInterface) {
             return $date->format($format);
         } elseif ($date instanceof \DateInterval) {
             return $date->format($format);
@@ -127,10 +127,10 @@ class DateHelper implements ProtectedContextAwareInterface
     /**
      * Get the day of month of a date
      *
-     * @param \DateTime $dateTime
+     * @param \DateTimeInterface $dateTime
      * @return integer The day of month of the given date
      */
-    public function dayOfMonth(\DateTime $dateTime)
+    public function dayOfMonth(\DateTimeInterface $dateTime)
     {
         return (integer)$dateTime->format('d');
     }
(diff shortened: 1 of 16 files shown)


kitsunet:
Note that this would break Neos if support for DateTimeInterface is not implemented there first, so do not merge yet.

kdambekalns:
From the top of my head: 

* why not for 3.0, it should be BC
* wasn't there something about DateTimeInterface and HHVM (and/or PHP7)?

kitsunet:
That's the matching Neos change: https://github.com/neos/neos-development-collection/pull/142
We could put it in 3.0 but I don't see a strong reason for it. Currently there is no bug that this would solve.
So just tell me if I should rebase to 3.0.

albe:
:+1: from reading, couldn't find anything problematic about DateTimeInterface, but I'll be sure to check it out on our HHVM machine

skurfuerst:
:+1: by reading; I merged Neos change beforehand so I am also merging this one. good work!

bwaidelich:
I'm late to the party (again), but the `DateTimeConverter` should probably be adjusted, too.
Also `\TYPO3\Flow\Utility\Now` does not seem to play nice with PHP7 because it is not compatible with `\DateTime` any longer

bwaidelich:
Ah, that hasn't got anything to to do with PHP7 I suppose but with this very change:
Code like
```
public function setDate(\DateTime $date) {
// ...
}
```
now needs to be adjusted to `\DateTimeInterface` but that breaks support for the `PropertyMapper`

kitsunet:
Mmm, yes that is true, but then we need to fix that I suppose. A workaorund for now is to cast now to a DateTime

kitsunet:
But that just shows we need more tests.

albe:
Wait, what? DateTimeImmutable is not an instanceof DateTime? oh... PHP.... doing it wrong again :(

kitsunet:
Nope and that is a good thing. You would expect something mutable if you expect DateTime. In that regard DateTimeImmutable IS different, so it would be possibly dangerous to make it an instance of DateTime.

kitsunet:
https://github.com/neos/flow-development-collection/pull/167 for the partial revert.

PR #93, opened 2015-10-19
[TASK] Add apigen.yml

kdambekalns:
This adds an apigen.yml file for use when generating API docs.


kitsunet:
:+1: I guess we need to merge and see...

kdambekalns:
That's what I'd think as well :)

aertmann:
awesome thanks for taking care

PR #26, opened 2015-09-04
[BUGFIX] Fix lookup path for ``AvailableProxyClasses.php``

bwaidelich:
Since 2787b2a3216deb188c4cd1c9b2b823e6e3a10da3 Flow creates a
map for available proxy classes stored in ````AvailableProxyClasses.php``
within the temporary directory.

The ``ClassLoader`` failed to include that file though when run in
a nested Application Context. That is fixed with this change.

Releases: master, 3.0, 2.3, 2.2

PR #96, opened 2015-10-20
[BUGFIX] Fix lookup path for ``AvailableProxyClasses.php``

bwaidelich:
Since 2787b2a3216deb188c4cd1c9b2b823e6e3a10da3 Flow creates a
map for available proxy classes stored in ``AvailableProxyClasses.php``
within the temporary directory.

The ``ClassLoader`` failed to include that file though when run in
a nested Application Context. That is fixed with this change.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php
@@ -440,12 +440,12 @@ class ClassLoader
         }
 
         if ($context !== null) {
-			// see Environment::createTemporaryDirectory()
-			$temporaryDirectory = FLOW_PATH_DATA . 'Temporary/' . str_replace('/', '/SubContext', (string)$context) . '/';
-			$proxyClasses = @include($temporaryDirectory . '/AvailableProxyClasses.php');
-			if ($proxyClasses !== false) {
-				$this->availableProxyClasses = $proxyClasses;
-			}
+            // see Environment::createTemporaryDirectory()
+            $temporaryDirectory = FLOW_PATH_DATA . 'Temporary/' . str_replace('/', '/SubContext', (string)$context) . '/';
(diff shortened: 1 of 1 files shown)


bwaidelich:
This is an updated version of #26

dfeyer:
:+1:

dfeyer:
Based on this discussion: https://discuss.neos.io/t/release-cycle-version-numbers-breaking-changes/538/18 we should try to avoid / discuss breaking change. What about this one @aertmann, @robertlemke, @kitsunet ?

albe:
:+1: from reading

TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php
```diff
@@ -440,7 +440,9 @@ protected function initializeAutoloadInformation($composerPath, ApplicationConte
         }
 
         if ($context !== null) {
-            $proxyClasses = @include(FLOW_PATH_DATA . 'Temporary/' . (string)$context . '/AvailableProxyClasses.php');
+            // see Environment::createTemporaryDirectory()
+            $temporaryDirectory = FLOW_PATH_DATA . 'Temporary/' . str_replace('/', '/SubContext', (string)$context) . '/';
```
kitsunet: Not necessarily something to fix in this change but as I see it, shouldn't Context have a getConfigurationPath() method or something like that... This string manipulation stuff with hardcoded strings makes me edgy ;)

kitsunet:
:+1:

TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php
albe: +1

bwaidelich: Absolutely. I just copied the code from `Environment::createTemporaryDirectory()` for now

PR #97, opened 2015-10-20
[!!!][BUGFIX] Store site lock in Flow temporary base path

bwaidelich:
Storing the site lock files in the system temporary directory
could lead to endless locks on some file systems.
This changes the site locks to be stored in the Flow temporary base
path again.

This is a breaking change because it removes the setting
``TYPO3.Flow.utility.environment.temporaryDirectoryBase`` in favor of
a new environment variable ``FLOW_PATH_TEMPORARY_BASE`` that allows for
changing the path if needed.

Background:

FLOW-348 introduced a new locking mechanism that stored lock files
in the systems default temporary folder determined via
``sys_get_temp_dir()``. On some systems files created there by the
PHP process could not be removed afterwards.

Related: FLOW-348
Resolves: FLOW-381


bwaidelich:
Closing in favor of #97 which applies this to the lowest affected branch (2.3)

dfeyer:
Based on this discussion: https://discuss.neos.io/t/release-cycle-version-numbers-breaking-changes/538/18 we should try to avoid / discuss breaking change. What about this one @aertmann, @robertlemke, @kitsunet ?

bwaidelich:
@dfeyer I agree, it's a tricky one.. The problem is: This was broken by design (see https://jira.neos.io/browse/FLOW-348 and https://jira.neos.io/browse/FLOW-381)  and I see no other easy way to fix it.
Besides: It is breaking only in "theory" because the removed `TYPO3.Flow.utility.environment.temporaryDirectoryBase` setting wasn't evaluated consistently anyways

aertmann:
It's a breaking bugfix. The current implementation is broken and should be fixed, regardless of the change being breaking IMO. The bug is much worse than the fix being breaking IMO. Unfortunately it's a regression introduced in a bugfix during late stages of the 3.0 release (beta -> RC).

aertmann:
:+1: by reading

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
```diff
@@ -590,6 +602,12 @@ protected function ensureRequiredEnvironment()
                 exit(1);
             }
         }
+        if (!is_dir(FLOW_PATH_TEMPORARY) && !is_link(FLOW_PATH_TEMPORARY)) {
+            if (!@mkdir(FLOW_PATH_TEMPORARY, 0777, true)) {
```
aertmann: wait is 0777 a good idea?

kitsunet:
I am all for getting this in :+1:

albe:
:+1: and +1 in what aske said.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
bwaidelich: I just copied the code from `Files::createDirectoryRecursively()`. We can't use that method from here because the exception handler is not registered at this point

bwaidelich:
See my comment above. Can I somehow rebase this or should I just merged from CLI?

kitsunet:
you can either do a rebase and force push your PR branch or do a merge and push that. Both work equally well

bwaidelich:
rebased (solving conflicts due to #96)

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
aertmann: I see, seems okay then. However this might not work as expected since it doesn't do the umask workaround like ``createDirectoryRecursively `` does, which could lead to a different file permission being set. Why not use a try catch statement?

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
```diff
@@ -86,11 +87,11 @@ class Bootstrap
...
     public function __construct($context)
     {
-        $this->defineConstants();
-        $this->ensureRequiredEnvironment();
-
         $this->context = new ApplicationContext($context);
         $this->earlyInstances[__CLASS__] = $this;
+
+        $this->defineConstants();
+        $this->ensureRequiredEnvironment();
```
kdambekalns: Why has this block been moved down? A specific reason?

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
kdambekalns: I agree, 0777 without any further adjustments is not good.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
```diff
@@ -526,6 +527,17 @@ protected function defineConstants()
         define('FLOW_PATH_DATA', FLOW_PATH_ROOT . 'Data/');
         define('FLOW_PATH_PACKAGES', FLOW_PATH_ROOT . 'Packages/');
 
+        if (!defined('FLOW_PATH_TEMPORARY')) {
+            $temporaryDirectoryPath = self::getEnvironmentConfigurationSetting('FLOW_PATH_TEMPORARY');
+            if ($temporaryDirectoryPath !== null) {
+                $temporaryDirectoryPath = rtrim($temporaryDirectoryPath, '/') . '/';
+            } else {
+                $temporaryDirectoryBasePath = self::getEnvironmentConfigurationSetting('FLOW_PATH_TEMPORARY_BASE') ?: FLOW_PATH_DATA . '/Temporary';
+                $temporaryDirectoryPath = Files::concatenatePaths(array($temporaryDirectoryBasePath, str_replace('/', '/SubContext', (string)$this->context))) . '/';
```
kdambekalns: Adding the context is only done here, so this requires to set a different temporary path along with the context. Why? I think this will surprise people.

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Environment.php
```diff
@@ -58,28 +47,22 @@ public function __construct(\TYPO3\Flow\Core\ApplicationContext $context)
...
-     * @api
+     * @return string
      */
     public function getPathToTemporaryDirectory()
     {
-        if ($this->temporaryDirectory !== null) {
-            return $this->temporaryDirectory;
-        }
-
-        $this->temporaryDirectory = $this->createTemporaryDirectory($this->temporaryDirectoryBase);
```
kdambekalns: Whoa, I don't like this… Before, when the temporary directory could not be created a helpful message was thrown, not Flow will probably just die in weird ways. What about checking the needed checks in `defineConstants()` at least?

kdambekalns: Ah, well. There is an error message output in `ensureRequiredEnvironment()`. Ok. I'd still rather call some method there to create the directory, and have more/better error and permission handling there.

kdambekalns:
Left a bunch of inline comments, for me this looks rather like :-1: TBH. No matter how bad the current behviour is, adding even more (potential) regressions and loopholes seems not worth it.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
kitsunet: You need to have the context ready to define the constants now.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
kitsunet: I agree, that could probably change.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
bwaidelich: Right, `$this->context` is required in `defineConstants()` (line 536)

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
bwaidelich: I don't really understand the objection. The context was always part of the temporary path (see Environment:: createTemporaryDirectory()).
That's why I introduced a new env variable FLOW_PATH_TEMPORARY_BASE because the context sub folders always need to be added or you end up using the same temporary files for multiple contexts

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
bwaidelich: You're right - I'll add the umask magic from `createDirectoryRecursively()`. Using that method with a `try/catch` won't work because it does not use the shutup-operator leading to a fatal error

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Environment.php
bwaidelich: You mean adding the umask magic to
```php
if (!is_dir(FLOW_PATH_TEMPORARY) && !is_link(FLOW_PATH_TEMPORARY)) {
    if (!@mkdir(FLOW_PATH_TEMPORARY, 0777, true)) {
        echo('Flow could not create the directory "' . FLOW_PATH_TEMPORARY . '". Please check the file permissions manually or run "sudo ./flow flow:core:setfilepermissions" to fix the problem. (Error #1441354578)');
        exit(1);
    }
}
```
and extract it to a method??

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
kitsunet: How I understood it (and that is what I would agree to) is that having base path AND full path settable via env might be unnecesary and possibly confusing and just the base bath would be enough.

bwaidelich:
@kdambekalns thanks for your feedback. I agree that adding regressions is not a good idea and that wasn't my intention ;)
I'll push a follow-up adding the umask stuff as discussed - is there anything else you would consider blocking here?

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
bwaidelich: Ah I see - right, I agree

bwaidelich: The reason for this was to be able to pass `FLOW_PATH_TEMPORARY` to sub processes (see `CoreCommandController::launchSubProcess()` for example)

bwaidelich: I guess I could just pass `FLOW_PATH_TEMPORARY_BASE` instead.. I'll try

kitsunet: Thanks :)

bwaidelich: not so easy unfortunately.. :-/

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
```diff
@@ -526,6 +527,12 @@ protected function defineConstants()
         define('FLOW_PATH_DATA', FLOW_PATH_ROOT . 'Data/');
         define('FLOW_PATH_PACKAGES', FLOW_PATH_ROOT . 'Packages/');
 
+        if (!defined('FLOW_PATH_TEMPORARY')) {
+            define('FLOW_PATH_TEMPORARY_BASE', self::getEnvironmentConfigurationSetting('FLOW_PATH_TEMPORARY_BASE') ?: FLOW_PATH_DATA . '/Temporary');
+            $temporaryDirectoryPath = Files::concatenatePaths(array(FLOW_PATH_TEMPORARY_BASE, str_replace('/', '/SubContext', (string)$this->context))) . '/';
+            define('FLOW_PATH_TEMPORARY', $temporaryDirectoryPath);
```
bwaidelich: This is now simplified as follows:
* We only evaluate the `FLOW_PATH_TEMPORARY_BASE` env variable (as documented)
* Accordingly only the `FLOW_PATH_TEMPORARY_BASE` is now passed to the sub requests (therefore we need to set the `FLOW_PATH_TEMPORARY_BASE` *constant*, too)
* The `FLOW_PATH_TEMPORARY` constant is set for convenience, so that it can be used instead of having to inject `Environment`

kitsunet:
I like, thanks for the effort :) @kdambekalns could you check, agree and merge please? ;)

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Environment.php
kdambekalns: That's what I meant :)

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Environment.php
```diff
@@ -58,28 +47,22 @@ public function __construct(\TYPO3\Flow\Core\ApplicationContext $context)
...
-        $this->temporaryDirectoryBase = $temporaryDirectoryBase;
-        $this->temporaryDirectory = null;
+        throw new Exception('Changing the temporary directory path during runtime is no longer supported. Set the environment variable FLOW_PATH_TEMPORARY_BASE to change the temporary directory base', 1441355116);
     }
 
     /**
      * Returns the full path to Flow's temporary directory.
      *
-     * @return string Path to PHP's temporary directory
-     * @api
```
kdambekalns: Well, this could be deprecated or left as api, no? Just making it non-api seems useless.

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Environment.php
bwaidelich: OK, I see. But, honestly, do you really think it is worth extracting those for lines https://github.com/bwaidelich/flow-development-collection/blob/NEOS-1333/TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php#L530-L534  - especially in light of the surrounding code

bwaidelich: -> https://github.com/bwaidelich/flow-development-collection/blob/NEOS-1333/TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php#L600-L608  these ones, sorry

kdambekalns:
Ok, I still think the various checks and according messages in the former `createTemporaryDirectory()` should be brought back, but aside from that it looks fine.

Question: does this really lead to lock files being removed again in a reliable way? Because right now it seems leftover lock files are really a PITA.

Also, @robertlemke might want to have a close look on this one…

kitsunet:
The cause for the leftover files is usually permissions, in this directory we will have more influence so yes should help it.

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Environment.php
bwaidelich: I didn't deprecated it because it is used in the core quite a lot - and replacing it with the constant is not a (viable) option because many tests that mock the environment would fail..
I removed the api annotation because it was the last method with it and I think that we should get rid of the whole Environment class at some point.. But, sure, we could deprecated it in a follow-up targeted for 3.1+ - agree?

kitsunet: I would like that a lot (getting rid of Environment in the mid term).

kdambekalns: We should probaboly discuss that on discuss, but looking at the amount of code that is left there, it seems doable. Most of the code already has replacements or is just syntactic sugar (like the max path length getter).

bwaidelich: Yes, it's totally doable. But this is meant to be a fix for 3.0.1 (and 2.3.x+1) mainly with little as possible impact.
So I'd suggest to re-add the api annotation as you suggested, get this one merged as patchlevel release and then have `Environment` deprecated for 3.1..
BTW: Is it possible to target a PR for 2.3.x+1 but *not* for 3.0.x+1?

bwaidelich: I can take care of removing `Environment`-calls from the core as follow-up (for 3.1)

robertlemke:
Regarding the "breaking-change" notion of this change: I'm using the setting in all my hosting setups at the moment, so even it it might be broken for some cases, it worked generally and in my case won't work anymore if we remove the setting.

I'll take a closer look at the code as soon as possible.

bwaidelich:
@robertlemke oh really? I honestly expected noone to actually change this setting because it is not evaluated consistently (for example in `ClassLoader` and `setfilepermissions.sh`.. well, that's about all I could find atm..). Stupid hen & eggs everywhere

dfeyer:
We use it too in development inside Vagrant / Docker to avoid having the temp files on shared FS, but @bwaidelich if right this parameter is not evaluated correctly and some temp files stay in the original folder.

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Environment.php
kdambekalns: @bwaidelich a PR that should not go into the upper branch(es) should be handled by doing the upmerge right after merging the PR and making sure the unwanted changed are not carried over by the merge. That should be done by whoever merges the PR immediately after merging.

kdambekalns:
So, I assume using the setting is not possible (anymore) because configuration is not available realy enough? Or is there a way to still use the setting if defined, to keep BC? The settings could then be deprecated, of course…

kitsunet:
How about this: We deprecate the setting, the getter will still return the directory defined by the setting but we define the constants based on data/env if it is set. Then people could slowly migrate

kitsunet:
@kdambekalns the site lock uses sys_temp_dir only because the temp dir couldn't be fetched that early because no settings. and IMHO a temp dir is so essential it shouldn't depend on soo many things as our configuration does

bwaidelich:
How shall we proceed with this?
I see three options:
1.) Keep it a breaking change for those that actually used the setting
2.) Go the middle way @kitsunet suggested for 3.0.1 and fix it properly for 3.1 - I do wonder if having two ways to set the temp path improves the situation though
3.) Skip the fix for 3.0.x and go for the proper fix for 3.1 directly (possibly with a migration that warns the user if the setting was changed)

kitsunet:
Well, the thing is, according to all we discuss right now such a breaking change would only be allowed in 4.0, so 3.1 isn't even an option. I guess the middle way is the only possibility right now, should be very clearly documented, but should work for all people that use the setting, and those that do not use it are fine anyway.

bwaidelich:
But even if documented clearly, what *would* be the behavior if you use the setting and the env variable?
The site lock would use the constant and all other parts the setting or what?
I tend to give up on that one for a patchlevel release and have it for the next release including core migration.  Now, whether that can only be 4.0.. I can't judge. I don't even know how to target a branch that does not exist yet

kitsunet:
IMHO if you set the env AND the setting the env should win for everything. That can be documented. As the env is new people won't have it. So only if you choose to use the env you will only be able to use the env.

bwaidelich:
OK, that makes sense I guess. So the `FLOW_PATH_TEMPORARY` would *not* be set if the env variable is not set? Or only later when the setting is available?
But that would mean that this hotfix only works if you actually set the env variable, because the setting is not accessible to the site lock..
Another option might be to revert https://github.com/neos/flow-development-collection/commit/907440254fc52aaac252929bfcaf391d183107af#diff-63767a35bc6cfd2334a794e844a9bde2L72 partly so that the site lock is stored in the default temp directory.. that might be less harmful, but I'm not sure

kitsunet:
Ok, my idea:

setTemporaryDirectoryBase will still work as before, but be deprecated, same for getTemporaryDirectoryBase.
We remove the setting from Flow, so we can do an isset() check on it to know if someone set it in their project.
In either case we set the constant as now done in this change. For the time being as a deprecation helper we set an additional constant ala `FLOW_PATH_TEMPORARY_BASE_WAS_SET` to true/false.
If `FLOW_PATH_TEMPORARY_BASE_WAS_SET` then `FLOW_PATH_TEMPORARY_BASE`  will be given to setTemporaryDirectoryBase in any case (basically ignoring the setting), else if the setting was set we use that or otherwise fallback to (the default) FLOW_PATH_TEMPORARY_BASE in Data/Temporary.

For 4.0 we remove setTemporaryDirectoryBase, getTemporaryDirectoryBase, the setting and the helper constant.

kitsunet:
Note, that means that the constant and getTemporaryDirectoryBase can potentially have different values when you don't use the env but the setting. I guess that's OK because the results will be about the same as we have currently. So if it worked before with the setting for you, it will still.
You just need to know that unless you use the env, the constant will be different if you refactor your code to it.

bwaidelich:
I'm a bit puzzled, what `getTemporaryDirectoryBase` are you referring to?

bwaidelich:
And where should the `isset` check be added to?

bwaidelich:
FYI: I just had a quick chat with Christian and I'm still not sure whether it's worth the hassle for the patchlevel release.. However, a pragmatic approach we could go is:

As soon as the settings are available (in `Bootstrap\Scripts`) check the `FLOW_PATH_TEMPORARY_BASE` constant.
If it is equal to `FLOW_PATH_ROOT . 'Temporary/'` (default), set the directory base as before:
```
$environment->setTemporaryDirectoryBase($settings['utility']['environment']['temporaryDirectoryBase']);
```
If it is *not* the default (i.e. it has been defined by the user) *and* the `temporaryDirectoryBase` setting is specified, throw an exception during bootstrap.

If we didn't forget anything this should be 100% backwards-compatible with the exception of the site lock which will be stored in the configured/default temp directory instead of using `sys_get_temp_dir()`.

For 3.1 we could deprecate the `Environment` class and for 4.0 we could remove it completely.

bwaidelich:
With d341e26e I now reverted all changes to `Environment` and added the check to the bootstrap as announced..
This makes the PR backwards compatible with the two (desired) exceptions:
The site lock is now stored in `FLOW_PATH_TEMPORARY` instead of `sys_get_temp_dir()` - *even if specified otherwise in the Settings*.
So for those cases where this setting was actually used, the site lock moves from the systems temp directory to flow's default temp directory (being `FLOW_PATH_DATA . '/Temporary/<Context>..`).

albe:
+1 by reading, thx for the work bastian

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -159,7 +160,15 @@ public static function initializeConfiguration(Bootstrap $bootstrap)
         $settings = $configurationManager->getConfiguration(\TYPO3\Flow\Configuration\ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'TYPO3.Flow');
 
         $environment = new \TYPO3\Flow\Utility\Environment($context);
-        $environment->setTemporaryDirectoryBase($settings['utility']['environment']['temporaryDirectoryBase']);
+        if (isset($settings['utility']['environment']['temporaryDirectoryBase'])) {
+            $defaultTemporaryDirectoryBase = FLOW_PATH_DATA . '/Temporary';
+            if (FLOW_PATH_TEMPORARY_BASE !== $defaultTemporaryDirectoryBase) {
+                throw new \TYPO3\Flow\Exception(sprintf('It seems like the PHP default temporary base path has been changed from "%s" to "%s" via the FLOW_PATH_TEMPORARY_BASE environment variable. If that variable is present the TYPO3.Flow.utility.environment.temporaryDirectoryBase setting must not be specified!', $defaultTemporaryDirectoryBase, FLOW_PATH_TEMPORARY_BASE), 1447707261);
```
robertlemke: "If that variable is present" -> there's a comma missing after "present", which would help a bit understanding it at first read.

robertlemke:
Can't test it right now, but thanks a lot and :+1: by reading

dfeyer:
:+1: by reading

bwaidelich:
FYI the LockManager still creates far too many lock files with this patch applied (1564 for Development context, 2262 for Production context), but I think that's a different issue

kitsunet:
See https://github.com/neos/flow-development-collection/pull/10

Totally unrelated to the site locking

aertmann:
:+1: tested with default settings only on apache + php-fpm setup

bwaidelich:
Thanks for the active feedback!

PR #98, opened 2015-10-21
[BUGFIX] Initialize Router lazily

aertmann:
The Router should be able to initialize configured routes lazily
except when explicitly told not to do so.
This change allows the Router to get the configuration directly
from the ``ConfigurationManager`` if no other routing configuration
exists. If some routing configuration was set, this is used.

Fixes: FLOW-192
Fixes: FLOW-205


aertmann:
:+1: 

This bugfix should have been cherry-picked into 2.3 as well since the bug exists there as well, but wasn't. The change is identical to 493b04c2b1e58b96a8e494f4db568af920b8d335, except for the PSR-2 adjustments.

Confirmed this fixed an issue for client.

bwaidelich:
:+1: by reading

bwaidelich:
@aertmann I guess this one can be merged?

aertmann:
Yeah, if nobody else feels like reviewing it.

kitsunet:
I mean :+1: from me, but I basically created the original change so...

aertmann:
thanks

PR #102, opened 2015-10-22
[FEATURE] Add permissions check for setfilepermissions.sh

wexi79:
Using the setfilepermissions.sh script fails due to wrong file permissions left by the composer installation (missing executable bit). Added the checking of the file's permissions, and correcting the file's permissions if needed. Mode 0700 is set as the setfilepermissions.sh script sets it's own permissions to that mode.

### DIFF
--- a/TYPO3.Flow/Scripts/flow.php
+++ b/TYPO3.Flow/Scripts/flow.php
@@ -20,6 +20,10 @@ if (PHP_SAPI !== 'cli') {
 }
 
 if (isset($argv[1]) && ($argv[1] === 'typo3.flow:core:setfilepermissions' || $argv[1] === 'flow:core:setfilepermissions' || $argv[1] === 'core:setfilepermissions')) {
+    $perms = decoct(fileperms(__DIR__. '/setfilepermissions.sh') & 0777);
+    if ($perms !== '700'){
+        chmod(__DIR__. '/setfilepermissions.sh', 0700);
+    }
     if (DIRECTORY_SEPARATOR !== '/') {
         exit('The core:setfilepermissions command is only available on UNIX platforms.' . PHP_EOL);
     }


TYPO3.Flow/Scripts/flow.php
```diff
@@ -20,6 +20,10 @@
 }
 
 if (isset($argv[1]) && ($argv[1] === 'typo3.flow:core:setfilepermissions' || $argv[1] === 'flow:core:setfilepermissions' || $argv[1] === 'core:setfilepermissions')) {
+    $perms = decoct(fileperms(__DIR__. '/setfilepermissions.sh') & 0777);
```
bwaidelich: we usually try to avoid abbrevations and there needs to be a space between concatenated strings. So this should be sth like
```
$filePermissions = decoct(fileperms(__DIR__ . '/setfilepermissions.sh') & 0777);
```

TYPO3.Flow/Scripts/flow.php
```diff
@@ -20,6 +20,10 @@
...
+    if ($perms !== '700'){
+        chmod(__DIR__. '/setfilepermissions.sh', 0700);
```
bwaidelich:
```
chmod(__DIR__ . '/setfilepermissions.sh', 0700);
```
(whitespacing)

bwaidelich:
I wonder why styleCI didn't detect it, but there are some minor cosmetic code style nit picks.
Apart from that this change makes a lot of sense IMO

bwaidelich:
Ah, but you pushed this to *master" I just realized.. I think this should go into 3.0.1 so the PR should be done against branch 3.0

aertmann:
why only 3.0? isn't it relevant for older branches as well?

PR #104, opened 2015-10-25
[!!!][TASK] Refactor package management

kitsunet:
This is a major refactoring of the package management.
With this change the structure of the ``PackageStates`` file
is overhauled now using the composer key for consistency to
avoid problems if namespace or autoload information changes
as that would change the (Flow) package name. Additionally
makes the resolution mechanism for Flow package names more
robust and stores additional metadata in the PackageStates to
avoid loading all composer manifests at runtime.
Furthermore the package loading order generation was improved
by using a more sophisticated algorithm to get the order right.
All the changes together lead to a noticeable speed
improvement in runtime situations.
Deactivating packages now results in the package being moved out
of regular package paths into an ``Inactive`` directory.

The Following changes make this change a breaking change:
New packages are only automatically picked up if they were
installed via composer or created through the Flow package
management. In all other cases you need to run the:
``package:rescan`` command to pick new package up.

Some @api methods in the PackageManager (and Interface) have
a modified structure due to deprecation of the package
metadata classes. Specifically the ``createPackage`` method has
changed.

Some @api classes and methods were deprecated.

The whole MetaData part of the package management is no longer
used and will be removed in the next major version.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/PackageCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/PackageCommandController.php
@@ -13,6 +13,7 @@ namespace TYPO3\Flow\Command;
 
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Cli\CommandController;
+use TYPO3\Flow\Composer\Utility as ComposerUtility;
 use TYPO3\Flow\Core\Booting\Scripts;
 use TYPO3\Flow\Core\Bootstrap;
 use TYPO3\Flow\Package\PackageInterface;
@@ -63,21 +64,21 @@ class PackageCommandController extends CommandController
      * @return string
      * @see typo3.kickstart:kickstart:package
      */
-    public function createCommand($packageKey, $packageType = 'typo3-flow-package')
+    public function createCommand($packageKey, $packageType = PackageInterface::DEFAULT_COMPOSER_TYPE)
     {
         if (!$this->packageManager->isPackageKeyValid($packageKey)) {
-            $this->outputLine('The package key "%s" is not valid.', array($packageKey));
+            $this->outputLine('The package key "%s" is not valid.', [$packageKey]);
             $this->quit(1);
         }
         if ($this->packageManager->isPackageAvailable($packageKey)) {
-            $this->outputLine('The package "%s" already exists.', array($packageKey));
+            $this->outputLine('The package "%s" already exists.', [$packageKey]);
             $this->quit(1);
         }
-        if (substr($packageType, 0, 11) !== 'typo3-flow-') {
-            $this->outputLine('The package must be a Flow package, but "%s" is not a valid Flow package type.', array($packageKey));
+        if (ComposerUtility::isFlowPackageType($packageType)) {
+            $this->outputLine('The package must be a Flow package, but "%s" is not a valid Flow package type.', [$packageKey]);
             $this->quit(1);
         }
-        $package = $this->packageManager->createPackage($packageKey, null, null, $packageType);
+        $package = $this->packageManager->createPackage($packageKey, ['type' => $packageType], null);
         $this->outputLine('Created new package "' . $packageKey . '" at "' . $package->getPackagePath() . '".');
     }
 
@@ -93,11 +94,11 @@ class PackageCommandController extends CommandController
     public function deleteCommand($packageKey)
     {
         if (!$this->packageManager->isPackageAvailable($packageKey)) {
-            $this->outputLine('The package "%s" does not exist.', array($packageKey));
+            $this->outputLine('The package "%s" does not exist.', [$packageKey]);
             $this->quit(1);
         }
         $this->packageManager->deletePackage($packageKey);
-        $this->outputLine('Deleted package "%s".', array($packageKey));
+        $this->outputLine('Deleted package "%s".', [$packageKey]);
         Scripts::executeCommand('typo3.flow:cache:flush', $this->settings, false);
         $this->sendAndExit(0);
     }
@@ -115,17 +116,17 @@ class PackageCommandController extends CommandController
     public function activateCommand($packageKey)
     {
         if (!$this->packageManager->isPackageAvailable($packageKey)) {
-            $this->outputLine('The package "%s" does not exist.', array($packageKey));
+            $this->outputLine('The package "%s" does not exist.', [$packageKey]);
             $this->quit(1);
         }
 
         if ($this->packageManager->isPackageActive($packageKey)) {
-            $this->outputLine('Package "%s" is already active.', array($packageKey));
+            $this->outputLine('Package "%s" is already active.', [$packageKey]);
             $this->quit(1);
         }
 
         $this->packageManager->activatePackage($packageKey);
-        $this->outputLine('Activated package "%s".', array($packageKey));
+        $this->outputLine('Activated package "%s".', [$packageKey]);
         Scripts::executeCommand('typo3.flow:cache:flush', $this->settings, false);
         $this->sendAndExit(0);
     }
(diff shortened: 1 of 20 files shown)


kitsunet:
So there we go, finally the PackageManager change is here. I guess it could need some more refinement, I was happy to have it rebased and all conflicts resolved...
A few points: I guess I could separate the `@api` method changes out to make this less breaking. I would still mark it breaking as it is such a major change and could have influence on things possibly.
If we would decide that the `@api` changes are not so bad I would suggest that I (re)do the remval of the MetaData classes as well, because they are dead code now and have no value. I doubt that someone actually uses them.

TYPO3.Flow/Classes/TYPO3/Flow/Command/PackageCommandController.php
```diff
@@ -63,21 +64,21 @@ public function injectSettings(array $settings)
...
             $this->quit(1);
         }
         if ($this->packageManager->isPackageAvailable($packageKey)) {
-            $this->outputLine('The package "%s" already exists.', array($packageKey));
+            $this->outputLine('The package "%s" already exists.', [$packageKey]);
             $this->quit(1);
         }
-        if (substr($packageType, 0, 11) !== 'typo3-flow-') {
-            $this->outputLine('The package must be a Flow package, but "%s" is not a valid Flow package type.', array($packageKey));
+        if (ComposerUtility::isFlowPackageType($packageType)) {
```
radmiraal: Can we make 'isFlowPackageType' a  bit more clear? In CR we use NodeType->isOfType, something similar here ends up in a long name so I don't really have an awesome suggestion coming up, but maybe someone has a brilliant idea? Otherwise leave it as is ;)

TYPO3.Flow/Classes/TYPO3/Flow/Command/PackageCommandController.php
```diff
@@ -384,22 +383,38 @@ public function refreezeCommand($packageKey = 'all')
...
-            $this->outputLine('Refroze package "%s".', array($packageKey));
+            $this->outputLine('Refroze package "%s".', [$packageKey]);
         }
 
         Scripts::executeCommand('typo3.flow:cache:flush', $this->settings, false);
         $this->sendAndExit(0);
     }
+
+    /**
+     * Rescan package availability and recreates the PackageStates configuration.
```
radmiraal: do we still require a @return void?

TYPO3.Flow/Classes/TYPO3/Flow/Composer/Utility.php
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * Read the content of the composer.lock
+     *
+     * @return array
+     */
+    public static function readComposerLock()
+    {
+        if (self::$composerLockCache === null) {
+            if (!file_exists(FLOW_PATH_ROOT . 'composer.lock')) {
+                return array();
```
radmiraal: return []; would be more consistent with the further boyscouting in the change ;)

TYPO3.Flow/Classes/TYPO3/Flow/Composer/Utility.php
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Checks if the given (composer) package type is a type native to the neos project.
+     *
+     * @param string $packageType
+     * @return boolean
+     */
+    public static function isFlowPackageType($packageType)
+    {
+        foreach (['typo3-flow-', 'neos-'] as $allowedPackageTypePrefix) {
```
radmiraal: nice prep ;) (neos-)

TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php
```diff
@@ -26,12 +27,22 @@ class ClassLoader
...
      * @var string
      */
-    const MAPPING_TYPE_PSR0 = 'Psr0';
+    const MAPPING_TYPE_PSR0 = 'psr-0';
 
     /**
      * @var string
      */
-    const MAPPING_TYPE_PSR4 = 'Psr4';
+     const MAPPING_TYPE_PSR4 = 'psr-4';
```
radmiraal: major indentation error :D

TYPO3.Flow/Classes/TYPO3/Flow/Command/PackageCommandController.php
kitsunet: No good idea how to make that clearer...

TYPO3.Flow/Classes/TYPO3/Flow/Command/PackageCommandController.php
kitsunet: no idea ;) We should clarify that...

TYPO3.Flow/Classes/TYPO3/Flow/Composer/Utility.php
kitsunet: Mercí :)

TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php
```diff
@@ -234,33 +245,19 @@ protected function loadClassFromPossiblePaths(array $possiblePaths, array $names
...
-                // Remove entries coming from composer for inactive packages.
-                if ($package->getAutoloadType() === Package::AUTOLOADER_TYPE_PSR4) {
-                    $this->removeNamespaceMapEntry($package->getNamespace(), $package->getClassesPath(), self::MAPPING_TYPE_PSR4);
-                } else {
-                    $this->removeNamespaceMapEntry($package->getNamespace(), $package->getClassesPath());
-                }
-                if ($this->considerTestsNamespace) {
-                    $this->removeNamespaceMapEntry($package->getNamespace(), $package->getPackagePath(), self::MAPPING_TYPE_PSR4);
-                }
+        foreach ($activePackages as $packageKey => $package) {
```
skurfuerst: I'd only do `$activePackages as $package`, you do not use `$packageKey` AFAIK :)

TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
```diff
@@ -206,17 +207,20 @@ protected function registerClassFiles(array $packages)
...
         /** @var \TYPO3\Flow\Package\Package $package */
         foreach ($packages as $packageKey => $package) {
-            if ($package->isObjectManagementEnabled() && (strpos($package->getComposerManifest('type'), 'typo3-flow') === 0 || isset($includeClassesConfiguration[$packageKey]))) {
-                $classFiles = $package->getClassFiles();
-                if (count($classFiles) > 0) {
-                    if (isset($this->allSettings['TYPO3']['Flow']['object']['registerFunctionalTestClasses']) && $this->allSettings['TYPO3']['Flow']['object']['registerFunctionalTestClasses'] === true) {
-                        $classFiles = array_merge($classFiles, $package->getFunctionalTestsClassFiles());
+            if ($package->isObjectManagementEnabled() && (ComposerUtility::isFlowPackageType($package->getComposerManifest('type')) || isset($includeClassesConfiguration[$packageKey]))) {
+                foreach ($package->getClassFiles() as $fullClassName => $path) {
+                    if (substr($fullClassName, -9, 9) !== 'Exception') {
```
skurfuerst: add some comment on why "Exception" is ignored here maybe? :)

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
```diff
@@ -336,37 +279,55 @@ public function getPackagePath()
     }
 
     /**
-     * Returns the full path to the packages Composer manifest
+     * Returns the full path to this package's Classes directory
      *
-     * @return string
+     * @return string Path to this package's Classes directory
+     * @api
+     * @deprecated
```
skurfuerst: add a note when the deprecation has happened :)

skurfuerst:
Hey Christian,

I just read over the code and found some small nitpicks. Generally, I like the code a lot, and it feels a lot cleaner to me than the old one.

I know that the "rescan" is really needed as command; although it would be *really* cool if it was only needed in Production. This, for my side, is the biggest breaking change I am seeing.

All the best,
Sebastian

kitsunet:
Thanks for going through it :) Will take care of the nits. The rescan brings a really nice performance boost for development and IMHO there are not many use cases left where you will actually need it. Rescan is automatically triggered on composer update/install (see InstallerScript) and also on package create/delete. So the only case would be if you manually create a package folder and composer.json to start a package, and I would discourage that anyway in favor of using package:create. So I would really like to keep it at that.

TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
kitsunet: If I would know...
See further down in the original code. It happens there already, I just copied it over...

kitsunet:
Note, I made a new PR here: https://github.com/neos/flow-development-collection/pull/131
The change there is essentially the same, I just tried to avoid any Interface / @api changes to make the change less/not breaking. All needed breaking changes are documented so for Flow 4.0 we just need to go in and do what the comments say ;)

PR #105, opened 2015-10-26
[TASK] Add permissions check for setfilepermissions.sh

wexi79:
Using the setfilepermissions.sh script fails due to wrong file permissions left by the composer installation (missing executable bit). Added the checking of the file's permissions, and correcting the file's permissions if needed. Mode 0700 is set as the setfilepermissions.sh script sets it's own permissions to that mode.

### DIFF
--- a/TYPO3.Flow/Scripts/flow.php
+++ b/TYPO3.Flow/Scripts/flow.php
@@ -20,6 +20,11 @@ if (PHP_SAPI !== 'cli') {
 }
 
 if (isset($argv[1]) && ($argv[1] === 'typo3.flow:core:setfilepermissions' || $argv[1] === 'flow:core:setfilepermissions' || $argv[1] === 'core:setfilepermissions')) {
+       $filePermissions = decoct(fileperms(__DIR__ . '/setfilepermissions.sh') & 0777);
+    if ($filePermissions !== '700'){
+        chmod(__DIR__ . '/setfilepermissions.sh', 0700);
+    }
+    
     if (DIRECTORY_SEPARATOR !== '/') {
         exit('The core:setfilepermissions command is only available on UNIX platforms.' . PHP_EOL);
     }


bwaidelich:
:+1: by reading
Awesome, thanks ;)

dfeyer:
:+1: nice one

dfeyer:
Maybe it's not a FEATURE but a TASK, please change commit message

TYPO3.Flow/Scripts/flow.php
```diff
@@ -20,6 +20,11 @@
 }
 
 if (isset($argv[1]) && ($argv[1] === 'typo3.flow:core:setfilepermissions' || $argv[1] === 'flow:core:setfilepermissions' || $argv[1] === 'core:setfilepermissions')) {
+       $filePermissions = decoct(fileperms(__DIR__ . '/setfilepermissions.sh') & 0777);
+    if ($filePermissions !== '700'){
```
dfeyer: Maybe this can be moved after line 30 (windows check)

TYPO3.Flow/Scripts/flow.php
```diff
@@ -20,6 +20,11 @@
 }
 
 if (isset($argv[1]) && ($argv[1] === 'typo3.flow:core:setfilepermissions' || $argv[1] === 'flow:core:setfilepermissions' || $argv[1] === 'core:setfilepermissions')) {
+       $filePermissions = decoct(fileperms(__DIR__ . '/setfilepermissions.sh') & 0777);
```
aertmann: something seems wrong with indentation?

aertmann:
agree it's a task, I wonder why the styleci test passed since it has codestyle issues

bwaidelich:
> I wonder why the styleci test passed

yep, wondered about the same in #102

PR #106, opened 2015-10-26
TASK: Add permissions check for setfilepermissions.sh

wexi79:
Using the ``setfilepermissions.sh`` script fails due to wrong file permissions left by the composer installation (missing executable bit). Added the checking of the file's permissions, and correcting the file's permissions if needed. Mode 0700 is set as the ``setfilepermissions.sh`` script sets it's own permissions to that mode.

FLOW-400 #close

### DIFF
--- a/TYPO3.Flow/Scripts/flow.php
+++ b/TYPO3.Flow/Scripts/flow.php
@@ -23,6 +23,12 @@ if (isset($argv[1]) && ($argv[1] === 'typo3.flow:core:setfilepermissions' || $ar
     if (DIRECTORY_SEPARATOR !== '/') {
         exit('The core:setfilepermissions command is only available on UNIX platforms.' . PHP_EOL);
     }
+    
+    $filePermissions = decoct(fileperms(__DIR__ . '/setfilepermissions.sh') & 0777);
+    if ($filePermissions !== '700'){
+        chmod(__DIR__ . '/setfilepermissions.sh', 0700);
+    }
+    
     array_shift($argv);
     array_shift($argv);
     $returnValue = 0;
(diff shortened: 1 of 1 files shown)


TYPO3.Flow/Scripts/flow.php
```diff
@@ -23,6 +23,12 @@
     if (DIRECTORY_SEPARATOR !== '/') {
         exit('The core:setfilepermissions command is only available on UNIX platforms.' . PHP_EOL);
     }
+    
+    $filePermissions = decoct(fileperms(__DIR__ . '/setfilepermissions.sh') & 0777);
+    if ($filePermissions !== '700'){
```
aertmann: hey thanks for keeping updating this one, there one tiny nitpick left.. should be a space between ``)`` and ``{`` as in ``if ($filePermissions !== '700') {``.. you can just go to https://github.com/wexi79/flow-development-collection/edit/patch-4/TYPO3.Flow/Scripts/flow.php, fix it and submit an additional commit to the branch and it will automatically update this PR.. no need to close the pr and such

aertmann:
:+1:

kitsunet:
Will this work with windows or shouldn't we skip that there?

kdambekalns:
@kitsunet - on windows the script exits just before the new block, so that's not going to be an issue.

kdambekalns:
I'm not sure about this. Never had an issue so far, and doing the check on every call of the `flow` script… I don't know…

kitsunet:
Yeah, I missed the fact that it will run on every CLI request. Maybe before setfilepermissions is run? So when the setfilepermissions command was triggered via ./flow

wexi79:
Umm guys, the permission check is run only when flow:core:setfilepermissions is called ;)

kitsunet:
my fault ;) :+1:

PR #107, opened 2015-10-26
[BUGFIX] Ignore injected properties for value hash generation

albe:
When injecting properties into a value object, an exception was thrown that
a closure is tried to be serialized. This was due to the hash generation
in PersistenceMagicAspect only skipping properties that are annotated as
transient.

This change makes the value hash generation resort to the class schema instead
of directly iterating all properties, since the class schema is already focused
on persistence relevant properties and hence contains no injected properties.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Aspect/PersistenceMagicAspect.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Aspect/PersistenceMagicAspect.php
@@ -106,11 +106,11 @@ class PersistenceMagicAspect
         $proxyClassName = get_class($proxy);
         $hashSourceParts = array();
 
-        $properties = $this->reflectionService->getClassPropertyNames($proxyClassName);
-        foreach ($properties as $property) {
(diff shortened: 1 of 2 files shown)


kaystrobach:
seems to work properly

Akii:
+1 nice fix

Akii:
On the other hand: does this alter the value hash in any way? If so, this is pretty breaking.

albe:
No, because previously the hash generation would just crash when you used @Injected dependencies without explicitly making them @Transient also. So not breaking, but fixing :)

kaystrobach:
yes exactly :D

kitsunet:
:+1: by reading

Akii:
@albe it's not about the fix but more about the hash. If it would change for existing VOs you'd e. g. lose the ability to perform queries. I took a brief look at the generation of the class schema and didn't see anything suspicious.

To ensure keeping the value hash the same, tests could be added.

Other than that, looks good to me.

afoeder:
:+1: by reading, too.

dfeyer:
:+1:

PR #108, opened 2015-10-27
FEATURE: Allow usage of is*/has* accessors in Fluid templates directly

albe:
This changeset adds support for accessor methods is* and has* to be
used directly for property access.
This allows to use such accessors in Fluid templates, which makes
the template code more readable and avoids getIs* and getHas* methods
in domain models.

Example::

    <f:if condition="{someObject.isSomething}"></f:if>

This will call someObject->isSomething() method.

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
@@ -76,6 +76,8 @@ class ObjectAccessorNode extends AbstractNode
      *
      * If propertyPath is "bla.blubb", then we first call getProperty($object, 'bla'),
      * and on the resulting object we call getProperty(..., 'blubb').
+     * If propertyPath is "bla.isBlubb" or "bla.hasBlubb", and calling getProperty(..., 'isBlubb'/'hasBlubb')
+     * fails, we directly call $bla->isBlubb()/$bla->hasBlubb().
      *
      * For arrays the keys are checked likewise.
      *
@@ -87,21 +89,17 @@ class ObjectAccessorNode extends AbstractNode
     public static function getPropertyPath($subject, $propertyPath, RenderingContextInterface $renderingContext)
     {
         $propertyPathSegments = explode('.', $propertyPath);
-        foreach ($propertyPathSegments as $pathSegment) {
-            $isHasDirectAccess = false;
+        foreach ($propertyPathSegments as $propertyName) {
             try {
-                if (preg_match('/^(is|has)([A-Z].*)/', $pathSegment, $matches) > 0) {
-                    $subject = ObjectAccess::getProperty($subject, lcfirst($matches[2]));
-                    $isHasDirectAccess = true;
-                }
+                $subject = ObjectAccess::getProperty($subject, $propertyName);
             } catch (PropertyNotAccessibleException $exception) {
-                // Don't do anything here for backwards compatiblity
-            }
-
-            if (!$isHasDirectAccess) {
-                try {
-                    $subject = ObjectAccess::getProperty($subject, $pathSegment);
-                } catch (PropertyNotAccessibleException $exception) {
+                if (is_object($subject) && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) > 0) {
+                    try {
+                        $subject = $subject->$propertyName();
+                    } catch (\Exception $exception) {
+                        $subject = null;
+                    }
+                } else {
                     $subject = null;
                 }
             }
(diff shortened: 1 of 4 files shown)


aertmann:
Not sure why this is necessary, doesn't https://github.com/neos/flow/blob/master/Classes/TYPO3/Flow/Reflection/ObjectAccess.php#L166 already do this? AFAIR ``object.hidden`` calls ``isHidden`` if available.

albe:
Yes, but the use case is to use the more explicit and expressive "object.isHidden" syntax inside Fluid. This would currently create the necessary getter "getIsHidden()" on the object, which is suboptimal for code readability.

dfeyer:
:+1: 

Maybe we need to check with @bwaidelich if that be in sync with the standalone fluid ?

TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
```diff
@@ -88,10 +88,22 @@ public static function getPropertyPath($subject, $propertyPath, RenderingContext
     {
         $propertyPathSegments = explode('.', $propertyPath);
         foreach ($propertyPathSegments as $pathSegment) {
+            $isHasDirectAccess = false;
             try {
-                $subject = ObjectAccess::getProperty($subject, $pathSegment);
+                if (preg_match('/^(is|has)([A-Z].*)/', $pathSegment, $matches) > 0) {
```
bwaidelich: more a theoretical issue probably, but this won't match all allowed php method names starting with "is" or "has". Checking with ``substring`` like it's done in ``ObjectAccess`` would be safer probably

bwaidelich: Nevermind - as you're only comparing the first letter after "is"/"has" this is not a real problem!

TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
```diff
@@ -88,10 +88,22 @@ public static function getPropertyPath($subject, $propertyPath, RenderingContext
...
+                    $subject = ObjectAccess::getProperty($subject, lcfirst($matches[2]));
```
bwaidelich: The readability can be increased by changing the ``pathSegment`` instead of repeating the whole ``try/catch`` block (maybe also rename "pathSegment" in this case):
```php
foreach ($propertyPathSegments as $propertyName) {
  if (preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) === 1) {
    $propertyName = lcfirst($matches[2]);
  }
  try {
    $subject = ObjectAccess::getProperty($subject, $propertyName);
  } catch (PropertyNotAccessibleException $exception) {
    $subject = null;
  }
}
```

TYPO3.Fluid/Tests/Functional/Core/ParserIntegrationTest.php
```diff
@@ -94,4 +95,27 @@ public function layoutViewHelperCanContainIfCondition()
...
+        $standaloneView->assignMultiple(array('post' => $post));
+        $standaloneView->setTemplateSource('<f:if condition="{post.isPrivate}">Private!</f:if>');
+
+        $actual = $standaloneView->render();
+        $this->assertSame('Private!', $actual);
+
+        $standaloneView->setTemplateSource('<f:if condition="{post.isprivate}">Private!</f:if>');
+
+        $actual = $standaloneView->render();
+        $this->assertSame('', $actual);
```
bwaidelich: Great, that you added a test, even a functional one!
Just a mini nit-pick: If you would extract the second case ("post.isprivate") into its own test method it would be easier to read & debug in case of errors

TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
bwaidelich: And checking again, I realize that you kept the duplication for greater B/C..
But maybe we can still keep the readability by doing sth like:

```
if (ObjectAccess::isPropertyGettable($$subject, lcfirst($matches[2]))) {
  $propertyName = lcfirst($matches[2]);
}
```

then instead

bwaidelich:
Nice one, thanks for this.
Apart from the nit picks I still think this is a breaking change currently if you *did* go the extra round and implemented a method ``getIsFoo()`` which won't be called any longer with this change. So maybe we should try first whether the unmodified property name is gettable:
```php
if (ObjectAccess::isPropertyGettable($subject, $propertyName)) {
```
and only change it if that's not the case

albe:
Thanks for the feedback, will push new commits soonish

albe:
@bwaidelich I think it's more readable now and fully b/c. I also split the tests and added another one for the getIs/Has* case.

albe:
Fixed the StyleCI issue with a rebase

bwaidelich:
:+1: by reading. Thanks for taking care!!

kitsunet:
:+1: looks good

albe:
Just added a small documentation reference to the new behaviour. Will merge with previous votes.

albe:
Rebased, maybe this helps the Travis hickup...

albe:
Ok, after the Travis issue is fixed, will merge with previous votes now.

PR #109, opened 2015-10-27
[BUGFIX] Validator chain is built correctly for cyclic relations

albe:
For cases where an entity A references entity B and B in turn
references A (as first property), the validator chain would not
validate the reference to A inside B, even though those might be
different instances.
This is due to the newly created ObjectValidator for B not being added
to the CollectionValidator inside A - so when B again visits A, the
returned validator collection is empty and hence not added to the
validator chain.

This change fixes that, by adding the new ObjectValidator to the chain
early, and only removing it later on, if it is still empty after all
properties being checked. This could potentially lead to a deeper
validator chain than would be necessary (cycle of empty ObjectValidators),
but should not break anything or degrade performance.

Fixes: FLOW-352

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Validation/ValidatorResolver.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Validation/ValidatorResolver.php
@@ -275,6 +275,7 @@ class ValidatorResolver
         if (!TypeHandling::isSimpleType($targetClassName) && class_exists($targetClassName)) {
             // Model based validator
             $objectValidator = new GenericObjectValidator(array());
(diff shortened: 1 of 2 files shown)


albe:
Should be fine now, test fails without change. Please review.

kitsunet:
Looks good to me, but I would like to see one more review. @bwaidelich or @skurfuerst maybe?

kitsunet:
forgot :+1: by reading.

skurfuerst:
Absolutely makes sense. It's such an obvious small change once you see it. Really good catch :-D

:+1: by reading! Merging with previous votes

PR #110, opened 2015-10-28
[BUGFIX] Make sure expired cache entries get deleted in PdoBackend

dimaip:
`$this->has($entryIdentifier)` returns false for expired cache entries
which leads to duplicate key violations. Remove existing entries,
including expired ones, before creating new cache entry.

Fixes: FLOW-193
Releases: master, 3.0, 2.3


bwaidelich:
:+1: by reading. Thanks, Dmitry ;)

kitsunet:
:+1: by reading. Go for it @dimaip  :)

PR #111, opened 2015-10-28
[BUGFIX] ResourceTypeConverter should return null for empty source

kitsunet:
The given source can either be an array or a string and in both cases
an empty value would signify a value that cannot be converted and
probably stems from an empty input. In this case the converter
should return null immediately, otherwise it will go on with processing
the empty value and eventually ends up in returning a conversion error
which would be wrong.


albe:
:+1: by reading

dfeyer:
:+1: by reading

PR #113, opened 2015-10-28
Backprot/flow 301

fanat98:


kitsunet:
:-1: Could you explain this pull request and the intention? Also it seems it was made based on a different branch or on a master that was not current. Would be nice to know what this is about. The "check if migrations exist" change seems interesting.

fanat98:
Hi @kitsunet 
We use Neos 1.2 with Flow 2.3 Package and get Exeption when trying to run doctrine:migrate or migrationstatus.
This Bug was Fixed in Flow 3.0.
https://jira.neos.io/browse/FLOW-301
https://github.com/fanat98/flow-development-collection/commit/581c94d0a84cb0d837cbc253bb99b14097a652cd

kdambekalns:
Replaced by https://github.com/neos/flow-development-collection/pull/115

PR #114, opened 2015-10-29
[TASK] More flexible controllerObjectName in Request Patterns settings

dfeyer:
This change make the controllerObjectName extensible in the Request Patterns configuration. It's backward compatible with the current single string configuration, but now you should use this synthax:

```yaml
'MyLDAPProvider':
  provider: 'TYPO3\MyCoolPackage\Security\Authentication\MyLDAPProvider'
  providerOptions: 'Some LDAP configuration options'
  requestPatterns:
    controllerObjectName:
      'TYPO3\MyApplication\AdministrationArea\.*': TRUE
```

This change also update the documentation to match the new syntax.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Security/RequestPattern/ControllerObjectName.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Security/RequestPattern/ControllerObjectName.php
@@ -37,11 +37,14 @@ class ControllerObjectName implements \TYPO3\Flow\Security\RequestPatternInterfa
     /**
      * Sets an controller object name pattern (preg_match() syntax)
      *
-     * @param string $controllerObjectNamePattern The preg_match() styled controller object name pattern
+     * @param string|array $controllerObjectNamePattern The preg_match() styled controller object name pattern
      * @return void
      */
     public function setPattern($controllerObjectNamePattern)
     {
+        if (is_array($controllerObjectNamePattern)) {
+            $controllerObjectNamePattern = implode('|', array_keys(array_filter($controllerObjectNamePattern)));
+        }
         $this->controllerObjectNamePattern = $controllerObjectNamePattern;
     }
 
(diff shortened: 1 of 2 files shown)


bwaidelich:
Hi Dominique,
thanks for taking care of this!
I think there was a little misunderstanding though. The syntax I had in mind originally would be s.th. like:
```yaml
'MyLDAPProvider':
  provider: 'TYPO3\MyCoolPackage\Security\Authentication\MyLDAPProvider'
  providerOptions: 'Some LDAP configuration options'
  requestPatterns:
    'someRequestPattern':
      pattern: 'ControllerObjectName'
      patternOptions:
        'regularExpressions':
          'TYPO3.MyApplication:Administration': 'TYPO3\MyApplication\AdministrationArea\.*'
```
This might seem a bit cumbersome, but it would be in sync with the other configurations for authentication providers, loggers, resources, ... and it would be very flexible.
Keep in mind that "ControllerObjectName" is just one class, so s.th. like this should be possible too:
```yaml
  # ...
  requestPatterns:
    'someOtherRequestPattern':
      pattern: 'TYPO3\MyApplication\Security\RequestPatterns\FooPattern'
      patternOptions:
        'foo': 'bar'
```

Of course I don't ask you to implement it that way, but it might be worth considering it before we change the format twice (and have to support 3 formats in the end), what do you think?

dfeyer:
If I can have some pointer to try to implement it, I can try, but not enough in the security framework to do it right now.

bwaidelich:
I'll try to look into it.
Note: This is not just nitpicking on the syntax. But it should be possible to *replace* existing patterns from a 3rd party package IMO (so that you can replace the `controllerObjectName` pattern by your custom implementation for example)

foerthner:
Yes, we definitely should stay with the possibility to combine multiple patterns of arbitrary types. I like Bastis proposal. In the end the method buildProvidersAndTokensFromConfiguration() in the authentication provider manager has to be changed to work with the new syntax. I guess it's possible there to support both variants, to old and the new one...

bwaidelich:
@dfeyer @foerthner I didn't know how to take over existing PR, but I gave it a try in a separate branch for now: https://github.com/bwaidelich/flow-development-collection/commit/12cc0916cde2880000cd678f47b8f677a4147190

It's not exactly the syntax I suggested before but it allows for the same flexibility..
The change is still missing (adjusted) tests and maybe a core migration to adjust existing settings to the new format but I wanted to collect your feedback first

dfeyer:
@bwaidelich I closed this one, can you send a PR from your branch ?

PR #115, opened 2015-10-29
[BUGFIX] Check if migrations exist before trying to register them

fanat98:
Backport from 9823ae980e803 (Author: Laurent Cherpit @lcherpit )

Currently the check for the existence of migration files is delegated
to the doctrine/migration third party library.

The behavior has changed and an exception is thrown if the folder doesn't
exist. To not let the third party library handle that and to prevent this,
the check is done upstream in Flow.


kitsunet:
:+1: Will merge in 24h if no one objects. IMHO quite obvious and good fix :)

dfeyer:
:+1:

bwaidelich:
:+1: by reading

PR #118, opened 2015-11-04
TASK: Do not instantiate all available TypeConverters

kdambekalns:
Previously the ``PropertyMapper`` would instantiate all available
TypeConverter implementations to build a map of possible sources
and targets. This is now done during compiletime so that only
instances of TypeConverters are created that have the required
source and target types.

NEOS-1288 #close


kdambekalns:
@kitsunet did you forget to commit some part here?

kitsunet:
yep seems pretty much so. I will think about it when my head is clear again. I am sure I can reproduce what I originally had there.

kdambekalns:
@kitsunet Could you have a look at this one (again)? Thanks!

kitsunet:
https://github.com/neos/flow-development-collection/pull/318
Is a working replacement that improves things in production.

PR #119, opened 2015-11-10
[BUGFIX] Set correct request port if X-Forwarded-Proto is set

robertlemke:
This fixes an issue resulting in wrong rendered URLs if Flow is accepting
request from a load balancer or proxy which is accessed via https
externally, sends the X-Forwarded-Proto header to Nginx but does not
specify the X-Forwarded-Port header.

For example, a load balancer (for example Google HTTP/HTTPS load
balancer) is accessible via https and terminates SSL. The load balancer
communicates with Nginx via http on port 80. Google only sends the
X-Forwarded-Proto header ("https") but not the port. URLs, for example
in an action URI of a form, are rendered wrongly.

An expected URL would be https://example.com/foo.html, however, the
rendered URL is https://example.com:80/foo.html

This change changes the behavior in `Request` so that if the
X-Forwarded-Proto header is set, but the X-Forwarded-Port header isn't,
the port is set to the standard port of the given protocol (80 / 443).

Resolves: FLOW-409

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
@@ -95,6 +95,8 @@ class Request extends Message
 
         if ($this->headers->has('X-Forwarded-Port')) {
             $this->uri->setPort($this->headers->get('X-Forwarded-Port'));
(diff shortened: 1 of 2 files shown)


kdambekalns:
:+1:

radmiraal:
:+1:

robertlemke:
okay, wrong branch, need to push that for 2.3 ...

PR #120, opened 2015-11-10
[BUGFIX] Set correct request port if X-Forwarded-Proto is set

robertlemke:
This fixes an issue resulting in wrong rendered URLs if Flow is accepting
request from a load balancer or proxy which is accessed via https
externally, sends the X-Forwarded-Proto header to Nginx but does not
specify the X-Forwarded-Port header.

For example, a load balancer (for example Google HTTP/HTTPS load
balancer) is accessible via https and terminates SSL. The load balancer
communicates with Nginx via http on port 80. Google only sends the
X-Forwarded-Proto header ("https") but not the port. URLs, for example
in an action URI of a form, are rendered wrongly.

An expected URL would be https://example.com/foo.html, however, the
rendered URL is https://example.com:80/foo.html

This change changes the behavior in `Request` so that if the
X-Forwarded-Proto header is set, but the X-Forwarded-Port header isn't,
the port is set to the standard port of the given protocol (80 / 443).

Resolves: FLOW-409

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
@@ -95,8 +95,8 @@ class Request extends Message
 
         if ($this->headers->has('X-Forwarded-Port')) {
             $this->uri->setPort($this->headers->get('X-Forwarded-Port'));
-        } elseif ($this->headers->has('X-Forwarded-Proto') && $protocol === 'https') {
-            $this->uri->setPort(443);
+        } elseif ($this->headers->has('X-Forwarded-Proto')) {
+            $this->uri->setPort($protocol === 'https' ? 443 : 80);
         } elseif (isset($server['SERVER_PORT'])) {
             $this->uri->setPort($server['SERVER_PORT']);
         }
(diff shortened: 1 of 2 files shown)


robertlemke:
Previously pushed as PR #119 but to the wrong branch ...

kdambekalns:
:+1: by reading (including the slight difference to #119)

bwaidelich:
:+1:  by reading

kitsunet:
:+1: by reading. I feel the code is becoming slightly confusing and complex there, but that can be fixed with a future refactoring ;)

kitsunet:
Already wondered why the change makes a modification of the test necessary. Seems the modification is not necessary?

robertlemke:
... because the test was wrong – clearified that with Bastian.

kitsunet:
travis says now it's wrong

robertlemke:
@kitsunet Okay, I didn't completely understand what you meant, now I see. Well, the tests were wrong, so they had to be changed. But the change I pushed for this PR also was wrong (copied the lines from a wrong editor window), it was meant to be identical with the original PR #119. So, here's the fixed version now ...

kdambekalns:
Ah, better now. :+1:

kitsunet:
YAY

bwaidelich:
For the record: I don't really remember why the tests were like they were. But probably they just documented the previous *behavior* rather than testing the *specification*.

PR #121, opened 2015-11-10
TASK: Update Flow to use Doctrine ORM 2.5

kitsunet:
With this we update Flow to use Doctrine ORM 2.5. It should allow new features
in Doctrine ORM (like embedded objects and the cache annotation) to be used without
restrictions.

This also removes some special proxy methods to initialise doctrine proxies.

FLOW-260 #close


robertlemke:
@kitsunet you tagged this change with Discussion - what would you like to discuss? And does it make sense to review it even though Travis reports failing tests?

kitsunet:
It's just a move from gerrit. I want to discuss with @kdambekalns how to further proceed with this :)
Totally WIP right now.

Akii:
What's the status here, what's missing?

kdambekalns:
See comments on the original change (at http://gerrit.neos.io/#/c/38465/ now)

kdambekalns:
Note: ignore the Travis CI tests (for now), since they install ORM 2.4 but then try to run the code targeted at 2.5 on top.

kdambekalns:
So, with this it's down to 2 errors and 3 failures _if_ the `relatedEntitiesCanBePersistedWhenFetchedAsDoctrineProxy()` test is disabled. If not, this test fails and hundreds of tests fail afterwards due to the EntityManager being closed.

kitsunet:
\o/ cool, do you have any ideas about this remaining thing yet? Or is is worth to have a look as well?

kdambekalns:
Ok, two tests in `TYPO3\Flow\Tests\Functional\Security\Authorization\Privilege\Entity\Doctrine\ContentSecurityTest` cause an error:

- `customersCannotSeeTestEntityAAssociatedToATestEntityBWithValueAdmin`
- `customersCannotSeeTestEntityAAssociatedToATestEntityBSomeoneElsesAccount`

Both lead to `SQLSTATE[HY000]: General error: 1 no such column: subselect1bf7080995e80724e88d98d3868e344e.persistence_object_identifier0`

Three tests fail, don't have details at hand.

If the test `relatedEntitiesCanBePersistedWhenFetchedAsDoctrineProxy` is enabled, the first error is

    1) TYPO3\Flow\Tests\Functional\Persistence\Doctrine\PersistClonedRelatedEntitiesTest::relatedEntitiesCanBePersistedWhenFetchedAsDoctrineProxy
    Doctrine\DBAL\Exception\UniqueConstraintViolationException: An exception occurred while executing 'INSERT INTO persistence_testentity (persistence_object_identifier, name, description, arrayproperty, relatedentity, relatedvalueobject) VALUES (?, ?, ?, ?, ?, ?)' with params ["35425c9f-049d-4851-9415-4811b43131d4", "Robert", "This is some text", "a:0:{}", null, null]:
    
    SQLSTATE[23000]: Integrity constraint violation: 19 UNIQUE constraint failed: persistence_testentity.persistence_object_identifier`

albe:
The issue with the security tests is following SQL query:

```SQL
SELECT t0_.persistence_object_identifier AS persistence_object_identifier_0, t0_.relatedentityb AS relatedentityb_1
FROM typo3_flow_tests_functional_security_fixtures_testentitya t0_
WHERE (
	( NOT (( (t0_.relatedentityb IN (
		SELECT subselect1bf7080995e80724e88d98d3868e344e.persistence_object_identifier0
		FROM (
			SELECT t0_.persistence_object_identifier AS persistence_object_identifier_0
			FROM typo3_flow_tests_functional_security_fixtures_testentityb t0_
			WHERE t0_.stringvalue = 'Admin'
		) AS subselect1bf7080995e80724e88d98d3868e344e
	) ) )))
AND ( NOT ((( (t0_.relatedentityb IN (
		SELECT subselect0334fa2e33f3004f8c573d79fb3b8f25.persistence_object_identifier0
		FROM (
			SELECT t0_.persistence_object_identifier AS persistence_object_identifier_0
			FROM typo3_flow_tests_functional_security_fixtures_testentityb t0_
			WHERE NOT (t0_.owneraccount = 'b2a7fe47-8c63-4626-8af0-c4133d371600')
		) AS subselect0334fa2e33f3004f8c573d79fb3b8f25
	) ) ) AND (t0_.relatedentityb IN (
		SELECT subselect6fbdea564b41bf771c0f30ec684a9f9c.persistence_object_identifier0
		FROM (
			SELECT t0_.persistence_object_identifier AS persistence_object_identifier_0
			FROM typo3_flow_tests_functional_security_fixtures_testentityb t0_
			WHERE NOT (t0_.owneraccount IS NULL)
		) AS subselect6fbdea564b41bf771c0f30ec684a9f9c 
	) ) ))) )
```

In each sub-sub-select, `t0_.persistence_object_identifier` is selected, but `*.persistence_object_identifier0` is tried to be accessed, ie. the alias index is appended, which is wrong. This happens somewhere in the custom `SqlFilter` queries in `Security\Authorization\Privilege\Entity\Doctrine`. Will have to dig a bit deeper, but maybe @foerthner can help.

The other test fails because a Doctrine Proxy is cloned, which will (no longer?) reset the persistence_object_identifier, leading to a duplicate primary key error.
This can be worked around by
```PHP
        $relatedEntity = $loadedEntity->getRelatedEntity();
        if ($relatedEntity instanceof Proxy) {
            $relatedEntity->__load();
        }
        $clonedRelatedEntity = clone $relatedEntity;
```
in the test, but of course that's not the solution.

albe:
Note: The relatedEntity test also fails even if ORM 2.4.* is still installed, so it's solely dependent on the changes in this PR.

albe:
Ok, the core issue is, that the AOP methodAndAdvices array isn't built for the invocation of __clone on a doctrine Proxy, hence the generateUuid advice isn't invoked and the identifier not updated leading to the primary key problem.
So basically a lazy loaded relation entity will not invoke the __wakeup postLoad, which makes sense, but in turn __clone on the doctrine Proxy also doesn't invoke the initializer of the object, so we end up with the above.

The easy solution would be to always build the advices array on __clone call, but of course that isn't optimal, because __clone could be called multiple times, so there should at least be a safeguard check.

MarkusRodler:
Doctrine ORM 2.5 is released quite some time ago. Is there a chance to get this task done in the next time? It would be great to see the embedded value object feature in Flow.

albe:
ping @kdambekalns @foerthner @kitsunet - I'd like to get this done for 3.3 LTS. I'll take a closer look at the security SqlFilter again this week.

Any comments on the advices inside `__clone`?

kdambekalns:
I have to admit I don't really get the `__clone` issue…

> always build the advices array on __clone call

We do not build advices on calling `__clone`–do you mean to build advices _for_ `__clone`?

albe:
I think i worded that badly, let me try to make it more clear.

Currently, the generated proxy for an entity looks something like this:

	/**
	 * Autogenerated Proxy Method
	 */
	 public function __wakeup() {
		$this->Flow_Aop_Proxy_buildMethodsAndAdvicesArray();

	        ....
	}

	/**
	 * Autogenerated Proxy Method
	 */
	 public function __clone() {
				// FIXME this can be removed again once Doctrine is fixed (see fixMethodsAndAdvicesArrayForDoctrineProxiesCode())
			$this->Flow_Aop_Proxy_fixMethodsAndAdvicesArrayForDoctrineProxies();

	    ....
	}

With the `FIXME` removed in this change, `$this->Flow_Aop_Proxy_buildMethodsAndAdvicesArray();` is no longer called because the lazy Doctrine Proxy prevents calling `__wakeup`, like any other method invocation would. Hence the generateUuid advice for the `__clone` method is not invoked and the identifier stays the same as the uncloned entity, leading to the primary key issue.

The key point is, that the AOP (for persistence identifier) depends on the methodAndAdvicesArray being built before method invocation, but Doctrine Proxy doesn't initialize the Proxy on `__clone`, which would trigger `__wakeup` - it handles that differently (see `createCloner` vs. `createInitializer` in Doctrine\ORM\Proxy\ProxyFactory).

So I guess we need to manually trigger buildMethodsAndAdvicesArray in `__clone` like we do in `__wakeup`.

kdambekalns:
To me
> manually trigger buildMethodsAndAdvicesArray in __clone like we do in __wakeup

sounds good, did/could you try this?

albe:
Will give it a try

albe:
Not sure what happened, but tests now *all* pass, even the security tests that failed before... @kdambekalns could that really have been connected somehow?

In any case, please everyone check this out and test heavily!

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -444,12 +444,13 @@ public function buildProxyClass($targetClassName, array &$aspectContainers)
...
         $proxyClass->getConstructor()->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
         $proxyClass->getMethod('__wakeup')->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
+        $proxyClass->getMethod('__clone')->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
 
         if (!$this->reflectionService->hasMethod($targetClassName, '__wakeup')) {
             $proxyClass->getMethod('__wakeup')->addPostParentCallCode("        if (method_exists(get_parent_class(), '__wakeup') && is_callable('parent::__wakeup')) parent::__wakeup();\n");
         }
 
-        $proxyClass->addTraits(['\TYPO3\Flow\Object\Proxy\DoctrineProxyFixingTrait', '\TYPO3\Flow\Aop\AdvicesTrait']);
+        $proxyClass->addTraits(['\TYPO3\Flow\Aop\AdvicesTrait']);
```
dfeyer: Can we use ::class here ? All those classname as string is bad smell for me

    $proxyClass->addTraits([AdvicesTrait::class]);

dfeyer:
To test this PR, I just need to pull the branch and composer update, right ?

albe:
Yeah. I'll quickly rebase on latest master.

albe:
Now really latest master and added the AdvicesTrait::class

albe:
Ok, I guess there was a reason for the FQCN string... seems `addTraits()` expects relative class names

kitsunet:
Anything missing that deserves the WIP label? Otherwise I would review tomorrow in detail.

albe:
Hm, seems there are still the failing security tests. Travis just doesn't correctly load the dependency it seems. I was too eager, but my local setup also played me. Will take a look this weekend where this malformed query comes from.

albe:
Got it. There is some bad hardcoded knowledge about Doctrine internals (i.e. default select alias postfixing) inside the `PropertyConditionGenerator`. This should solve it for now. Should definitely be reconsidered though.

albe:
PS: Don't mind Travis failing now, that's due to the wrong Doctrine ORM dependency

albe:
That should pretty much be it AFAIC. So test to your heart's liking and as said, don't mind Travis for now, test locally with correct dependency.

albe:
Ok, so here are advices to test this:

Add 

```
    "repositories": [
		{
			"type": "git",
			"url": "https://github.com/kitsunet/flow-development-collection.git"
		}
    ],
```

to your composer.json and change the require for flow like that:
``"neos/flow-development-collection": "dev-gerrit-38465 as 3.3.0",``

Now composer update, flush caches, and go. So far, everything worked fine for me.

albe:
Works fine so far on our REST API project. No problems.

kitsunet:
Neos fails with 
`Exception while property mapping for target type "TYPO3\TYPO3CR\Domain\Model\NodeInterface", at property path "": Notice: Undefined index: TYPO3\Flow\Aop\Advice\AroundAdvice in ...` (NodeConverter).

You can see that there we expect `$adviceChains['TYPO3\Flow\Aop\Advice\AroundAdvice']` which is not set in the array (supposedly because the fix advices stuff is no longer there). Didn't really look into it yet, but blocking this for now...

albe:
Argh :(

kitsunet:
LOL, fun fact... The object which has broken Advices is not even an Entity. It has nothing to do with persistence...

kitsunet:
Ok, no idea for fixing yet, but this is what happens (luckily I choose to try it on exactly this project):

I have extended `\TYPO3\Neos\TypeConverter\NodeConverter` in my project. But the `prepareContextProperties` in there is subject to a security policy. Now my extended converter is called, who calls `prepareContextProperties` in the parent class. Now the proxy kicks in and wants to execute the advice. Unfortunately the advices array is never filled because my extended class is not subject to any AOP and therefore the buildAdvices thingy is not inserted into the proxy clsss.. -> BAM
It worked before because the doctirne fix method and advices thingy would just be triggered beforehand and fill the array in the correct class.

I might have an idea though... We need to figure out if we can know that a class extends an AOP'ed class and then just call it there on the parent, otherwise we need to put it in all classes proxies and check at runtime if the method is callable.

kitsunet:
Ok, see #418 for fix.

kdambekalns:
To install, we need to adjust the dev dependency of neos/utility-objecthandling to doctrine/orm 2.4.0 as well.

But I still have three failing tests (`Undefined index: TYPO3\Flow\Aop\Advice\AroundAdvice`), even with #418 applied. :( Any hints?

kitsunet:
good catch, actually I just tried something ala `$a instanceof \Foo\Bar\Baz` which simply returns false, so IMHO we could even drop the dependency from objecthandling as that is the only way a doctrine class is used in there.

kitsunet:
(separate change ofc)

kitsunet:
-> #424

kitsunet:
OK, so finally it works with Flow, Neos etc. but I still found a case were it breaks without the doctrine fix in functional tests. Will push a separate change again to fix that.

kitsunet:
See #425

PR #123, opened 2015-11-10
[!!!][TASK] Defer initialisation of View

kitsunet:
Resolving and initialisation of a view are now deferred to after
the action was called so that it only happens if the view is
actually needed.

This can be breaking if you rely on instanceof checks on the view,
because the view will always be a DependencyProxy. You need to call
`_activateDependency()` on the view to actually resolve it, before
checking the object type.


robertlemke:
How much speed improvement will this give, roughly speaking? Because I guess it might be a breaking change for (a very few) some.

kitsunet:
Yeah, this is also a quick take over from gerrit. I will brush it up and add some numbers...

kitsunet:
And TBH, with this I am also not sure if the breakiness isn't too much :)

kdambekalns:
Let's bury this for good…

PR #27, opened 2015-09-07
[TASK] Behat: ensure that "Given I have the following policies" is executed first

skurfuerst:
If this is violated, very funny error will appear because proxy
classes have already been built without taking the new Policy.yaml
into account.

Change-Id: Ic8f1165852898ef0afe295c61c6a69858de1622a
Releases: master, 3.0

PR #124, opened 2015-11-10
[TASK] Behat: ensure that "Given I have the following policies" is executed first

skurfuerst:
If this is violated, very funny error will appear because proxy
classes have already been built without taking the new Policy.yaml
into account.

This change has been re-targeted for 3.0 from https://github.com/neos/flow-development-collection/pull/27


kitsunet:
Could you quickly fix the style issue?
Generally makes sense and :+1: by reading.

skurfuerst:
force-pushed the change; waiting for StyleCI to run through.

PR #125, opened 2015-11-10
[TASK] Fix misleading log message in logException()

kdambekalns:
When logException() is used to log an exception that has been caught,
the log will still say "Uncaught exception". This is misleading.


skurfuerst:
Agreed. :+1: merging as this is a trivial change.

PR #126, opened 2015-11-11
[TASK] Improve variable access in compiled templates

kitsunet:
This is a slight improvement for variable access in Fluid
that reduces recursions of ``getPropertyPath`` by one level
by pre evaluating the property path and fetching the first level
directly from the variable container in compiled templates.


kitsunet:
Rather wait for standalone fluid which does that anyway.

PR #128, opened 2015-11-11
[FEATURE] Setting for application name

robertlemke:
This change introduces a new setting which allows developers to display
a custom application name and version in the ./flow help commands and
potentially elsewhere. For example, ./flow will display "Neos" and the
version number of the Neos package, when the command is run in a Neos
distribution.

The setting does not refer to the application name directly, but to a
package key. The specified package's meta data (Composer manifest) is
used to determine the application name and version. Since it is best
practice to use the "description" property of the Composer manifest for
specifying the application name, that field is used as the application
name in ./flow too (see also comments by Jordi at
https://github.com/composer/composer/issues/1140).

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/HelpCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/HelpCommandController.php
@@ -41,6 +41,12 @@ class HelpCommandController extends CommandController
     protected $bootstrap;
 
     /**
+     * @Flow\InjectConfiguration(path = "core.applicationPackageKey")
+     * @var string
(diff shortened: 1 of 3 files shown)


skurfuerst:
For the feature, a +-0 from my side. For the code, a slight :+1: by reading.

I am not sure why this feature is actually needed but I won't block it ;)

All the best, Sebastian

dfeyer:
:+1: by reading

robertlemke:
I mean, yeah, it's not soooo important. But I think it doesn't really hurt. And it replace "TYPO3 Flow" by "Flow" ...

PR #129, opened 2015-11-13
[FEATURE] Routes via Settings

bwaidelich:
This change allows routes to be included via `Settings.yaml` like follows::

    TYPO3:
      Flow:
        mvc:
          routes:
            'Some.Package': TRUE

This would append the main `Routes.yaml` from the package
``Some.Package`` (including its subroutes, if any) to existing
route definitions.

It's also possible to specify the order in relation to other packages::

    TYPO3:
      Flow:
        mvc:
          routes:
            'SomeOther.Package':
              position: 'before Some.Package'

This will add the ``SomeOther.Package`` routes *before* the routes
of ``Some.Package`` (if those are included).

*NOTE:* Routes included via main `Configuration/Routes.yaml` will
*always* be included first, it's not possible to add routes before or
in between those via the new setting.
Thus it's recommended to remove the main `Routes.yaml` and include
routes via `Settings.yaml` instead.

For the Flow Routes this could look like::

    TYPO3:
      Flow:
        mvc:
          routes:
            'TYPO3.Flow':
              position: 'start'

Resolves: FLOW-411


kitsunet:
nice idea

TYPO3.Flow/Classes/TYPO3/Flow/Cache/CacheManager.php
```diff
@@ -358,7 +358,7 @@ protected function flushConfigurationCachesByChangedFiles(array $changedFiles)
         $objectConfigurationCache = $this->getCache('Flow_Object_Configuration');
         $caches = array(
             '/Policy\.yaml/' => array('Flow_Security_Authorization_Privilege_Method', 'Flow_Persistence_Doctrine', 'Flow_Persistence_Doctrine_Results', 'Flow_Aop_RuntimeExpressions'),
-            '/Routes([^\/]*)\.yaml/' => array('Flow_Mvc_Routing_Route', 'Flow_Mvc_Routing_Resolve'),
+            '/Routes|Settings([^\/]*)\.yaml/' => array('Flow_Mvc_Routing_Route', 'Flow_Mvc_Routing_Resolve'),
```
bwaidelich: This could be a dealbreaker because it will flush the routing cache whenever a setting changes..
The alternative would be that the cache has to be flushed manually if routes are added/changed in Settings.yaml.. Or we make this clevererer so that it only triggers this when a setting in `TYPO3.Flow.mvc.routes` changed..!?

bwaidelich:
This is more a RFC yet. It works but currently leads to flushing of all route caches whenever a setting changed (see inline comment). If someone has an idea of how to make that less intrusive, please share ;)

kdambekalns:
A bit misleading: "define routes" is not what this does. Other than that, very nice idea. That way the welcome package could include it's routes when installed, and if not, then not. :wind_chime:

bwaidelich:
good point, adjusted

bwaidelich:
any idea on what to do about the routing cache?

TYPO3.Flow/Classes/TYPO3/Flow/Cache/CacheManager.php
albe: "Or we make this clevererer so that it only triggers this when a setting in TYPO3.Flow.mvc.routes changed..!?" - I see not much other than that. Could be a cool feature anyway to only flush specific caches depending on specific settings.

skurfuerst: @albe that's probably difficult to implement, as the file monitor only tells us "which files have changes" and not "what has changed in them"...

@bwaidelich to me, this cache flushing is not a big problem; as it won't occur in production.

However, a comment would be really good at this point explaining why Routing also depends on Settings.yaml. This would be at least a question I'd have when I just see this code standalone.

TYPO3.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
```diff
@@ -1388,20 +1459,15 @@ public function packageViewConfigurationsCallback($filenameAndPath)
...
 
-    /**
-     * @param string $configurationSourceCallbackName
-     * @param string $contextName
-     * @return ConfigurationManager
-     */
     protected function getConfigurationManagerWithFlowPackage($configurationSourceCallbackName, $contextName)
     {
-        $mockConfigurationSource = $this->getMock(\TYPO3\Flow\Configuration\Source\YamlSource::class, array('load', 'save'));
+        $mockConfigurationSource = $this->getMock('TYPO3\Flow\Configuration\Source\YamlSource', array('load', 'save'));
```
skurfuerst: why did you change this?

TYPO3.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
```diff
@@ -1388,20 +1459,15 @@ public function packageViewConfigurationsCallback($filenameAndPath)
...
         $mockConfigurationSource->expects($this->any())->method('load')->will($this->returnCallback(array($this, $configurationSourceCallbackName)));
 
-        $configurationManager = $this->getAccessibleMock(\TYPO3\Flow\Configuration\ConfigurationManager::class, array('postProcessConfiguration'), array(new ApplicationContext($contextName)));
+        $configurationManager = $this->getAccessibleMock('TYPO3\Flow\Configuration\ConfigurationManager', array('postProcessConfiguration', 'includeSubRoutesFromSettings'), array(new ApplicationContext($contextName)));
```
skurfuerst: why did you change this?

TYPO3.Flow/Tests/Unit/Utility/PositionalArraySorterTest.php
```diff
@@ -41,7 +42,6 @@ public function invalidPositions()
             array('subject' => array('foo' => array('position' => 'start123'), 'first' => array())),
             array('subject' => array('foo' => array('position' => 'start 12 34'), 'first' => array())),
             array('subject' => array('foo' => array('position' => 'after 12 34 56'), 'first' => array())),
-            array('subject' => array('foo' => array('position' => 'after foo-bar'), 'first' => array())),
```
skurfuerst: why is this testcase removed?

skurfuerst:
Hey @bwaidelich,

I just left some nitpicks -- like this change very much :-)

Thanks a lot,
Sebastian

TYPO3.Flow/Classes/TYPO3/Flow/Cache/CacheManager.php
albe: Yeah, but the settings is just an assoc array at that point, no? So a simple recursive array_diff_assoc could reveal the changed settings. At least worth a second thought IMHO - as said, it could maybe be a nice general feature.

skurfuerst: We can try :) But I think we do not have the settings even loaded at this point yet... But maybe we find a clever way :)

All the best,
Sebastian

aertmann:
:+1: for this change as well and would make it easier to introduce the feature "Asynchronous thumbnail generation in media browser", which I'm currently working on. Additionally we could adopt the SEO package to install the sitemap route automatically with this. Thanks for taking care of it :)

bwaidelich:
Little update: I gave it a quick shot but you're right, we never have both old and new settings loaded in one request..

> to me, this cache flushing is not a big problem; as it won't occur in production.

That's right, but it could make development context painful slow if all routes need to be re-created whenever you change any setting of any package. The router is not that slow, but it's quite common to have landing-pages with a couple of hundred or thousand links and the (file) cache seems to be a bottleneck there.
I'm thinking about caching the "settings-based" routes in the cache too and compare them upon settings-flushing via some signal

bwaidelich:
I checked a bit further. The ConfigurationManager is quite limited due to the early stage in which it is required. So using signal/slots there is not an option really.

But I just gave my last idea a quick shot and it seems to do the trick: 4ed61be

This can be improved a little so that it is not done in production context for example - and this needs tests still, but what do you think of this rather "pragmatic" approach?

skurfuerst:
Hey, @bwaidelich ,

really cool idea :) I like it a lot; if we only do it in Dev context that should be good to go!

Awesome!
All the best,
Sebastian

aertmann:
Second what @skurfuerst said

albe:
Nice idea

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
```diff
@@ -526,6 +527,7 @@ protected function loadConfiguration($configurationType, array $packages)
                 $this->configurations[$configurationType] = array_merge($this->configurations[$configurationType], $this->configurationSource->load(FLOW_PATH_CONFIGURATION . $configurationType));
 
                 // Merge routes with SubRoutes recursively
```
aertmann: this comment belongs to the next line..

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Routing/RouterCachingService.php
```diff
@@ -49,6 +49,23 @@ class RouterCachingService
...
+     * @var array
+     */
+    protected $routeSettings;
+
+    /**
+     * @return void
+     */
+    public function initializeObject()
+    {
+        if ($this->routeCache->get('routingSettings') !== $this->routeSettings) {
```
aertmann: add check for development context here as suggested by @skurfuerst or?

aertmann:
Tested and works. However for this to be useful for something like the SEO package, it needs to support ``variables`` so pushed a PR to support that https://github.com/bwaidelich/flow-development-collection/pull/7. Additionally we'd have to include the Neos routes using settings, which requires manual changes to all distributions. Although not a breaking change since it would work anyway, not sure if they would be duplicated though.

Last I find it a bit odd with the ``'Package.Key': TRUE`` or array if it has options which means it's enabled, but maybe that's the best solution? To disable one with options you set it to false or?

aertmann:
needs to be documented as well

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
bwaidelich: I added it there by intention because `includeSubRoutesFromSettings()` also just loads sub routes, but I'll rephrase it a bit to make that clearer

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Routing/RouterCachingService.php
bwaidelich: yes, I'll adjust that

TYPO3.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
bwaidelich: no idea, I probably didn't have the latest version. good catch, I'll revert it

TYPO3.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
bwaidelich: good point

bwaidelich:
Adjusted the code according to your valuable feedback, merged Aske's patch and added tests & documentation.
:+1:  from my side

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
```diff
@@ -549,7 +550,8 @@ This will load the SubRoutes from a file ``Routes.Foo.yaml`` in the ``My.Demo``
 With that feature you can include multiple Routes with your package (for example providing different URI styles).
 Furthermore you can nest routes in order to minimize duplication in your configuration. You nest SubRoutes by including
 different SubRoutes from within a SubRoute, using the same syntax as before.
-Additionally you can specify a set of ``variables`` that will be replaced in ``name`` and ``uriPattern`` of merged routes:
+Additionally you can specify a set of ``variables`` that will be replaced in ``name``, ``uriPattern`` and ``defaults``
+of merged routes:
```
aertmann: should also go to 3.0 right?

TYPO3.Flow/Classes/TYPO3/Flow/Utility/PositionalArraySorter.php
```diff
@@ -247,7 +247,7 @@ protected function extractAfterKeys(array &$arrayKeysWithPosition)
     {
         $this->afterKeys = array();
         foreach ($arrayKeysWithPosition as $key => $position) {
-            if (preg_match('/^after ([a-zA-Z0-9]+)(?: ([0-9]+))?$/', $position, $matches) < 1) {
+            if (preg_match('/^after (\S+)(?: ([0-9]+))?$/', $position, $matches) < 1) {
```
aertmann: is this a bugfix that should go to stable releases?

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
```diff
@@ -669,6 +671,71 @@ This will result in a merged configuration like this:
...
+you to set order depending on other SubRoutes.
+
+.. note::
+
+	SubRoutes that are loaded via Settings will always be appended **after** Routes loaded via ``Routes.yaml``
+	Therefore you should consider getting rid of the main ``Routes.yaml`` and only use settings to include routes
+	for greater flexibility.
+
+It's not possible to adjust route defaults or the ``UriPattern`` when including SubRoutes via Settings, but there are
+two more options you can use:
```
aertmann: needs to be two colons ``::``

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Routing/RouterCachingService.php
```diff
@@ -49,6 +50,30 @@ class RouterCachingService
...
+     */
+    protected $routingSettings;
+
+    /**
+     * @return void
+     */
+    public function initializeObject()
+    {
+        // flush routing caches if in Development context & routing settings changed
+        if ($this->objectManager->getContext()->isDevelopment() && $this->routeCache->get('routingSettings') !== $this->routingSettings) {
```
aertmann: not sure if it's any problem, but it will flush the empty caches on first request

aertmann:
looks pretty good, left a couple of comments though

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
kdambekalns: even better: use one colon and the `.. code-block:: yaml` directive…

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Routing/RouterCachingService.php
bwaidelich: good point.. Not sure how to prevent this - I'll check whether it's a potential issue

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
bwaidelich: True, I'll extract it to a separate PR

TYPO3.Flow/Classes/TYPO3/Flow/Utility/PositionalArraySorter.php
bwaidelich: it allows positional keys to contain special characters (such as "." when referring to package keys). One could consider it a bugfix, I'll push it to a separate PR

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
bwaidelich: Good catch, thx

kitsunet:
Would it be very hard to allow inclusion of specific sub routes instead of the full set for one package?

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
bwaidelich: Done with #170

TYPO3.Flow/Classes/TYPO3/Flow/Utility/PositionalArraySorter.php
bwaidelich: Done with #171

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Routing/RouterCachingService.php
bwaidelich: This will only happen in Dev context upon initial request - but it will just flush an empty cache, which is not an issue as far as I can tell (and I just tested it as well)

bwaidelich:
Addressed the comments from @aertmann and simplified the change by extracting the routing docs fix to #170  as well as the ``PositionalArraySorter`` fix (which is required by this change!) #171 .

@kitsunet you can use the ``suffix`` keyword to only include a certain sub route file

kdambekalns:
:+1: by reading

bwaidelich:
:+1: tested with flow master

aertmann:
:+1: verified, but needs Flow to be upmerged for the #171 dependency

can be merged when that is done IMO

bwaidelich:
@kdambekalns thanks for merging - but has the required #171 already been applied to master?

kdambekalns:
Yes, 18 hours ago, says GH. Ah, that's to another branch… Ok, pushed upmerges 2.3 -> 3.0 -> master

bwaidelich:
Awesome, thanks!!

PR #131, opened 2015-11-16
!!!TASK: Refactor package management

kitsunet:
This is a major refactoring of the package management.
With this change the structure of the ``PackageStates`` file
is overhauled now using the composer key for consistency to
avoid problems if namespace or autoload information changes
as that would change the (Flow) package key.

Additionally makes the resolution mechanism for Flow package
keys more robust and stores additional metadata in the
``PackageStates.php`` to avoid loading all composer manifests
at runtime.
Furthermore the package loading order generation was improved by
using a more sophisticated algorithm to get the order right.

All the changes together lead to a noticeable speed improvement
in runtime situations.

Deactivating packages now results in the package being moved out
of regular package paths into an ``Inactive`` directory.

This change is marked breaking due to the following:

New packages are only automatically picked up if they were
installed via composer or created through the Flow package
management. In all other cases you need to run the:
``package:rescan`` command to pick new package up.

Some @api classes and methods were deprecated and will be removed
or changed in the next major Flow version.

Some newly added methods of PackageManager and Package are used
in the Flow core now, so if someone would reimplement both
according to interface it would not work with Flow, but that
is already the case without this change, so it shouldn't be an issue.

The whole MetaData part of the package management is no longer
used and will be removed in the next major version.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Composer/Utility.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Composer/Utility.php
@@ -1,0 +1,175 @@
+<?php
+namespace TYPO3\Flow\Composer;
+
+/*                                                                        *
+ * This script belongs to the TYPO3 Flow framework.                       *
+ *                                                                        *
+ * It is free software; you can redistribute it and/or modify it under    *
+ * the terms of the GNU Lesser General Public License, either version 3   *
+ * of the License, or (at your option) any later version.                 *
+ *                                                                        *
+ * The TYPO3 project - inspiring people to share!                         *
+ *                                                                        */
+
+use TYPO3\Flow\Package\PackageInterface;
+use TYPO3\Flow\Reflection\ObjectAccess;
+use TYPO3\Flow\Utility\Files;
+
+/**
+ * Utility to access composer information like composer manifests (composer.json) and the lock file.
+ */
+class Utility
+{
+    /**
+     * Runtime cache for composer.json data
+     *
+     * @var array
+     */
+    protected static $composerManifestCache;
+
+    /**
+     * Runtime cache for composer.lock data
+     *
+     * @var array
+     */
+    protected static $composerLockCache;
+
+    /**
+     * Returns contents of Composer manifest - or part there of.
+     *
+     * @param string $manifestPath
+     * @param string $configurationPath Optional. Only return the part of the manifest indexed by configurationPath
+     * @return array|mixed
+     */
+    public static function getComposerManifest($manifestPath, $configurationPath = null)
+    {
+        $composerManifest = static::readComposerManifest($manifestPath);
+        if ($composerManifest === null) {
+            return null;
+        }
+
+        if ($configurationPath !== null) {
+            return ObjectAccess::getPropertyPath($composerManifest, $configurationPath);
+        } else {
+            return $composerManifest;
+        }
+    }
+
+    /**
+     * Read the content of the composer.lock
+     *
+     * @return array
+     */
+    public static function readComposerLock()
+    {
+        if (self::$composerLockCache !== null) {
+            return self::$composerLockCache;
+        }
+
+        if (!file_exists(FLOW_PATH_ROOT . 'composer.lock')) {
+            return [];
+        }
+
+        $json = file_get_contents(FLOW_PATH_ROOT . 'composer.lock');
+        $composerLock = json_decode($json, true);
+        $composerPackageVersions = isset($composerLock['packages']) ? $composerLock['packages'] : [];
+        $composerPackageDevVersions = isset($composerLock['packages-dev']) ? $composerLock['packages-dev'] : [];
+        self::$composerLockCache = array_merge($composerPackageVersions, $composerPackageDevVersions);
+
+        return self::$composerLockCache;
+    }
+
+    /**
+     * Read the content of composer.json in the given path
+     *
+     * @param string $manifestPath
+     * @return array
+     * @throws Exception\MissingPackageManifestException
+     */
+    protected static function readComposerManifest($manifestPath)
+    {
+        $manifestPathAndFilename = $manifestPath . 'composer.json';
+        if (isset(self::$composerManifestCache[$manifestPathAndFilename])) {
+            return self::$composerManifestCache[$manifestPathAndFilename];
+        }
+
+        if (!is_file($manifestPathAndFilename)) {
+            throw new Exception\MissingPackageManifestException(sprintf('No composer manifest file found at "%s".', $manifestPathAndFilename), 1349868540);
+        }
+        $json = file_get_contents($manifestPathAndFilename);
+        $composerManifest = json_decode($json, true);
+
+        self::$composerManifestCache[$manifestPathAndFilename] = $composerManifest;
+        return $composerManifest;
+    }
+
+    /**
+     * Checks if the given (composer) package type is a type native to the neos project.
+     *
+     * @param string $packageType
+     * @return boolean
+     */
+    public static function isFlowPackageType($packageType)
+    {
+        foreach (['typo3-flow-', 'neos-'] as $allowedPackageTypePrefix) {
+            if (strpos($packageType, $allowedPackageTypePrefix) === 0) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Determines the composer package name ("vendor/foo-bar") from the Flow package key ("Vendor.Foo.Bar")
+     *
+     * @param string $packageKey
+     * @return string
+     */
+    public static function getComposerPackageNameFromPackageKey($packageKey)
+    {
+        $nameParts = explode('.', $packageKey);
+        $vendor = array_shift($nameParts);
+        return strtolower($vendor . '/' . implode('-', $nameParts));
+    }
+
+    /**
+     * Write a composer manifest for the package.
+     *
+     * @param string $manifestPath
+     * @param string $packageKey
+     * @param array $composerManifestData
+     * @return array the manifest data written
+     */
+    public static function writeComposerManifest($manifestPath, $packageKey, array $composerManifestData = [])
+    {
+        $manifest = [
+            'description' => ''
+        ];
+
+        if ($composerManifestData !== null) {
+            $manifest = array_merge($manifest, $composerManifestData);
+        }
+        if (!isset($manifest['name']) || empty($manifest['name'])) {
+            $manifest['name'] = static::getComposerPackageNameFromPackageKey($packageKey);
+        }
+
+        if (!isset($manifest['require']) || empty($manifest['require'])) {
+            $manifest['require'] = array('typo3/flow' => '*');
+        }
+
+        if (!isset($manifest['autoload'])) {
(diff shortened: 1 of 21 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Composer/Utility.php
```diff
@@ -0,0 +1,172 @@
...
+     * Read the content of composer.json in the given path
+     *
+     * @param string $manifestPath
+     * @return array
+     * @throws Exception\MissingPackageManifestException
+     */
+    protected static function readComposerManifest($manifestPath)
+    {
+        $manifestPathAndFilename = $manifestPath . 'composer.json';
+        if (isset(self::$composerManifestCache[$manifestPathAndFilename])) {
```
albe: Can we somehow streamline the code for the cache check? Here it's quick return and above the loading logic is wrapped in the if

TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php
```diff
@@ -109,11 +120,16 @@ class ClassLoader
     /**
      * @param ApplicationContext $context
      */
```
albe: Missing @param annotation

albe: missing @param annotation

TYPO3.Flow/Classes/TYPO3/Flow/Composer/Utility.php
```diff
@@ -0,0 +1,172 @@
...
+     * @var array
+     */
+    protected static $composerLockCache;
+
+    /**
+     * Returns contents of Composer manifest - or part there of.
+     *
+     * @param string $manifestPath
+     * @param string $configurationPath Optional. Only return the part of the manifest indexed by configurationPath
+     * @return array|mixed
```
albe: "array|mixed" looks strange - is there a specific reason why it shouldn't just be "mixed"?

kitsunet: Idea was that in most cases you fetch the full composer manifest, thus being an array, but in case you request a part of the manifest that is mixed then.

TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php
kitsunet: fixed

TYPO3.Flow/Classes/TYPO3/Flow/Composer/Utility.php
kitsunet: done

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
```diff
@@ -425,97 +417,159 @@ public function getDocumentationPath()
...
+        return $this->flattenedAutoloadConfiguration;
     }
 
     /**
      * Returns contents of Composer manifest - or part there of.
      *
      * @param string $key Optional. Only return the part of the manifest indexed by 'key'
-     * @return mixed|NULL
-     * @see json_decode for return values
+     * @return array|mixed
```
albe: See above on "array|mixed"

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
```diff
@@ -425,97 +417,159 @@ public function getDocumentationPath()
...
+    protected function getClassesInNormalizedAutoloadPath($baseAutoloadPath, $autoloadNamespace)
+    {
+        $autoloadNamespace = trim($autoloadNamespace, '\\') . '\\';
+        $directories = [''];
+        while ($directories !== []) {
+            $currentRelativeDirectory = array_pop($directories);
+            $currentAbsoluteDirectory = $baseAutoloadPath . $currentRelativeDirectory;
+            if ($handle = opendir($currentAbsoluteDirectory)) {
+                while (false !== ($filename = readdir($handle))) {
+                    if ($filename[0] === '.') {
```
albe: What about '..'?

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
```diff
@@ -425,97 +417,159 @@ public function getDocumentationPath()
...
+     * @return \Generator
+     */
+    protected function getClassesInNormalizedAutoloadPath($baseAutoloadPath, $autoloadNamespace)
+    {
+        $autoloadNamespace = trim($autoloadNamespace, '\\') . '\\';
+        $directories = [''];
+        while ($directories !== []) {
+            $currentRelativeDirectory = array_pop($directories);
+            $currentAbsoluteDirectory = $baseAutoloadPath . $currentRelativeDirectory;
+            if ($handle = opendir($currentAbsoluteDirectory)) {
```
albe: a specific reason why you went back to opendir/readdir/closedir over DirectoryIterator?

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
```diff
@@ -425,97 +417,159 @@ public function getDocumentationPath()
...
+                while (false !== ($filename = readdir($handle))) {
+                    if ($filename[0] === '.') {
+                        continue;
+                    }
+                    $pathAndFilename = $currentAbsoluteDirectory . $filename;
+                    if (is_dir($pathAndFilename)) {
+                        $directories[] = $currentRelativeDirectory . $filename . '/';
+                        continue;
+                    }
+                    if (strpos(strrev($filename), 'php.') === 0) {
```
albe: Isn't that slower apart from being harder to read than "substr($filename, -4[, 4]) === '.php'"?

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageInterface.php
```diff
@@ -152,7 +154,7 @@ public function getDocumentationPath();
      * Returns the available documentations for this package
      *
      * @return array Array of \TYPO3\Flow\Package\Documentation
-     * @api
+     * @deprecated To be removed in Flow 4.0
```
albe: For this and all above (except getNamespace): what are the replacements to get the same information? Since those functions where @api there might be code around that depends on retrieving that information in some way.

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
```diff
@@ -375,61 +351,15 @@ protected function filterPackagesByPath(&$packages, $filterPath)
...
-        }
-        return $this->composerNameToPackageKeyMap[$lowercasedComposerName];
-    }
-
-    /**
-     * Check the conformance of the given package key
-     *
-     * @param string $packageKey The package key to validate
-     * @return boolean If the package key is valid, returns TRUE otherwise FALSE
-     * @api
```
albe: What about this function, it's @api but is just removed. Would need a @deprecated b/c dummy at least

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
```diff
@@ -375,61 +351,15 @@ protected function filterPackagesByPath(&$packages, $filterPath)
...
-        return $filteredPackages;
-    }
-
-    /**
-     * Returns the upper camel cased version of the given package key or FALSE
-     * if no such package is available.
-     *
-     * @param string $unknownCasedPackageKey The package key to convert
-     * @return mixed The upper camel cased package key or FALSE if no such package exists
-     * @api
```
albe: This method is @api but simply removed. Needs at least a deprecated b/c dummy

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
kitsunet: Way faster, methods in File Utility and FileMonitor were also changed to do that.

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
kitsunet: same reasoning as in the Utility.

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
kitsunet: It was just reordered, see further below. It still exists.

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
kitsunet: Again, just reordered, see below.

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
kitsunet: I think substr is slower, but I am fine with changing that.

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageInterface.php
kitsunet: IMHO that's up to your application to decide. It's pretty hardcoded in here, so if you want to keep these paths you can hardcode it in your app. Alternatively I would suggest to extend the composer.json with the information (extra.paths.documentation eg.)

skurfuerst:
:+1: after a thorough code review with Christian; after Travis and StyleCI run through, let's merge this!

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageOrderResolver.php
```diff
@@ -0,0 +1,136 @@
...
+        $packageState = $this->packageStates[$packageKey];
+
+        // $iteationForPackage will be -1 if the package is already worked on in a stack, in that case we will return instantly.
+        $iterationForPackage = $unsortedPackages[$packageKey];
+
+        if ($iterationForPackage === -1) {
+            return false;
+        }
+
+        if (!isset($unsortedPackages[$packageKey])) {
```
albe: should that isset() check come before accessing $unsortedPackages[packageKey]? Else that might give a fatal error

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
```diff
@@ -741,419 +693,451 @@ public function registerPackage(PackageInterface $package, $sortAndSave = true)
...
     /**
-     * Recursively sort dependencies of a package. This is a depth-first approach that recursively
-     * adds all dependent packages to the sorted list before adding the given package. Visited
-     * packages are flagged to break up cyclic dependencies.
+     * Derive a flow package key from the given information.
+     * The order of importance is:
+     *
+     * - package install path
+     * - first found autoload namespace
+     * - composer name
```
albe: Just a nitpick, but would be cool to have order of arguments match the order of importance, wdyt?

kitsunet:
Note this had problems in the IntallerScripts triggered by composer, as this never happens for the chnaged code in our travis pre merge tests the errors only happened after merging, but I merged a fix for those.

PR #132, opened 2015-11-19
[BUGFIX] Refresh first level roles cache on authentication

foerthner:
As soon as an authentication process completes, roles might
have changed. Therefore we have to reinitialize the roles
first level cache in the security context.

Fixes: FLOW-415

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
@@ -131,7 +131,7 @@ class SimpleFileBackend extends AbstractBackend implements PhpCapableBackendInte
         $this->cacheEntryFileExtension = ($cache instanceof PhpFrontend) ? '.php' : '';
 
         if ((strlen($this->cacheDirectory) + 23) > $this->environment->getMaximumPathLength()) {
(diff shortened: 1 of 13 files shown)


bwaidelich:
Thanks a lot for migrating this, @foerthner!  Is 3.0 the right branch?
It's a shame the bug is so hard to reproduce via tests apparently.. I think I stumbled upon it when the `CsrfProtection` filter called `authenticate()` before everything else was completely initialized.. But that might have been another issue as well

robertlemke:
Did you try to apply it on 2.3? If it's easy to merge into 2.3, I suggest that we just do that, otherwise it's fine to just release it in 3.0.x.

foerthner:
The fix also works fine with 2.3. How to proceed? Close thie pull request and open a new one for 2.3?

bwaidelich:
> How to proceed? Close thie pull request and open a new one for 2.3?

Yes, that seems to be the only way unfortunately

foerthner:
I opened a new PR for 2.3: https://github.com/neos/flow-development-collection/pull/136

Will close this one hereby.

foerthner:
I reopen this PR, as we have to change some more stuff in 3.0 than in 2.3, where the bugfix is already merged...

kitsunet:
Please first do an upmerge then apply remaining changes via this (or a new PR)

bwaidelich:
:+1: by reading. Looks good, thanks a lot for taking care!

kitsunet:
Thanks a lot for the upmerge. All seems well, usually we make the upmerges separate from anything else, but I think this is cool. Seems to work out, so :+1: by reading and trying. Merging.

PR #133, opened 2015-11-19
TASK: Add support for X-Original-Host HTTP header

dfeyer:
If Neos is behind a proxy, and the public domain name is not the same as the internal domain name, flow will redirect request from the public domain to the internal domain. This change add support for X-Original-Host HTTP header to get the public domain correctly.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
@@ -89,7 +89,11 @@ class Request extends AbstractMessage
         } else {
             $protocol = isset($server['SSL_SESSION_ID']) || (isset($server['HTTPS']) && ($server['HTTPS'] === 'on' || strcmp($server['HTTPS'], '1') === 0)) ? 'https' : 'http';
         }
-        $host = isset($server['HTTP_HOST']) ? $server['HTTP_HOST'] : 'localhost';
+        if ($this->headers->has('X-Original-Host')) {
+            $host = $this->headers->get('X-Original-Host');
+        } else {
+            $host = isset($server['HTTP_HOST']) ? $server['HTTP_HOST'] : 'localhost';
+        }
         $requestUri = isset($server['REQUEST_URI']) ? $server['REQUEST_URI'] : '/';
         if (substr($requestUri, 0, 10) === '/index.php') {
             $requestUri = '/' . ltrim(substr($requestUri, 10), '/');


robertlemke:
Thanks for the patch. I generally agree that we want support for this (and also for `X-Forwarded-For`), but while we are at it, I think that we really should implement a setting which allows us to specify whom to trust when using these headers.

I like the approach of Symonfy with their "Trusted Proxies" feature: http://symfony.com/doc/current/components/http_foundation/trusting_proxies.html

We could simply implement that via settings and disable proxy header acceptance by default. What do you think?

dfeyer:
So the setting should allow to configure:

- A list of IP Range
- And even "*" to allow everything (pretty useful when using service like ngrok, and you don't know the IP of the proxy)

A proxy MUST match a leat 1 IP or IP RANGE to allow the Trusted Proxy feature ?

If we I'm on the correct track, I can work on this one, one a issue on jira, ...

robertlemke:
Ideally you would be able to specify an array of ip adresses annotated as CIDR (https://en.m.wikipedia.org/wiki/Classless_Inter-Domain_Routing).

I don't know how much work that would be .... 

Oh and do we need to consider IPV6? I guess its okay without forbthe time being, since its usually about private network.



-- 
Sent from a mobile device.

bwaidelich:
FYI Regarding CIDR you can checkout the existing `\TYPO3\Flow\Security\RequestPattern\Ip` - maybe we can even reuse that here

albe:
+1 for the change together with trusted proxies. Regarding IP check, see the Ip RequestPattern as Bastian already stated. It fully supports IPv4 and IPv6 CIDR notations and should be easy to reuse here.

The only loophole I see is, that Ip RequestPattern depends on other HTTP Headers  to retrieve the client IP (Request::getClientIpAddress()), in order to check if the HTTP Headers are to be trusted. Classic hen-egg problem.

robertlemke:
@dfeyer would you be in to enhance this PR by including support for trusted proxies or should someone else try to do this?

dfeyer:
@robertlemke If somebody else would take the cookie, it's fine for me

kdambekalns:
So it seems no one took the cookie. What do we do about this?

albe:
I played around with this a little bit yesterday and I found following minor issues to tackle:

* since the http classes are non-proxied, settings can not be injected. My solution was to statically access the ConfigurationManager through the Bootstrap::$staticObjectManager and load the settings into a protected static variable if it was not set yet

* the Ip RequestPattern currently only matches against an ActionRequest and it checks against the request getClientIpAddress(), which is not reliable. The TrustedProxies check should only check against the REMOTE_ADDR though, because only that is reliable. My solution was to make the cidrMatch method on the pattern public - not nice, but gets the job done for now.

* symphony currently only checks a single configured header (e.g. "X-Forwarded-For") - would there also be a valid use-case to have a cascade of proxies that make use of different headers? e.g. proxy 1 uses "X-Client-Ip", proxy 2 "X-Forwarded-For"... ? That would highly complicate the trusted proxies configuration and the implementation, so maybe it's just too much for a next to non-existant use-case

albe:
Ok, so I pushed a new PR for the trusted proxies feature. Please take a look everyone and tell what you think.

PR #136, opened 2015-11-23
[BUGFIX] Roles are refreshed after setting authentication status

foerthner:
Otherwise getRoles() might act on the wrong value of the overall
authentication status stored in the authentication manager.


aertmann:
:+1: by reading, although not deep into the subject

albe:
Makes sense and should have no side-effects, :+1: by reading

kitsunet:
👍 by reading

bwaidelich:
:+1:  by reading.
A test to reproduce the broken behavior would be great, but it's probably not easy to do. IIRC I had the issue when the CsrfProtection called `authenticate` very early

bwaidelich:
FYI: I also tested this patch, verifying that it doesn't break an existing application. However I couldn't reproduce the bug - it probably only occurs with multiple concurring authentication tokens

PR #137, opened 2015-11-24
[BUGFIX] FastCGI compatible resources ``.htaccess`` file

aertmann:
The ``.htaccess`` file in ``Web/_Resources`` contained php_flag,
which requires the ``mod_php`` module to be installed. Since FastCGI
setups don't have this module, an invalid command error is thrown.

In this change the flag is wrapped in ``IfModule`` tags to avoid that error,
and the ``SetHandler`` statement is added a second time in a ``Files`` tag
to avoid it being overwritten in certain cases.

### DIFF
--- a/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/Web/_Resources/.htaccess
+++ b/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/Web/_Resources/.htaccess
@@ -1,2 +1,16 @@
+# Disable unwanted options
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Prevent execution of script files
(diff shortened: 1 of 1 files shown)


bwaidelich:
:+1: I can't test it with FastCGI atm, but I can confirm that this still works with PHP 5.6 & Apache 2.0 Handler

bwaidelich:
:+1: still works with mod_php (uploaded image can be accessed, script is blocked)

kdambekalns:
:+1: by reading…

albe:
:+1: verified, works on ZendServer with CGI/FastCGI but fails with old .htaccess

kitsunet:
:+1: By reading

PR #138, opened 2015-11-24
[BUGFIX] FlockLockStrategy should clean up lock files

kitsunet:
The FlockLockStrategy creates files to apply the lock on.
These files reside in the temporary folder but are never cleaned
on releasing the Lock that means the amount of files in this folder
will increase over time unless the folder is cleared manually.

Additionally cleans the code a bit and reduces chance of race
conditions while creating the lock.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
@@ -19,12 +19,13 @@ use TYPO3\Flow\Utility\Files;
 /**
  * A flock based lock strategy.
  *
- * This lock strategy is based on Flock. This strategy is currently disabled on Windows.
+ * This lock strategy is based on Flock.
  *
  * @Flow\Scope("prototype")
  */
 class FlockLockStrategy implements LockStrategyInterface
 {
+
     /**
      * @var string
      */
@@ -32,18 +33,21 @@ class FlockLockStrategy implements LockStrategyInterface
 
     /**
      * Identifier used for this lock
+     *
      * @var string
      */
     protected $id;
 
     /**
      * File name used for this lock
+     *
      * @var string
      */
     protected $lockFileName;
 
     /**
      * File pointer if using flock method
+     *
      * @var resource
      */
     protected $filePointer;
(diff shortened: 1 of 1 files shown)


kitsunet:
Replaces #10 
@albe I integrated your changes, was easier that way. Maybe you can check this as well?

aertmann:
:+1: works, from 2154 lock files to 1 :)

kitsunet:
👍

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
```diff
@@ -51,56 +55,97 @@ class FlockLockStrategy implements LockStrategyInterface
...
+        // This will always be TRUE on windows, because 'ino' stat will always be 0, but unlink is not possible on opened files anyway
+        if ($stat !== false && $stat['ino'] === $fstat['ino']) {
+            return false;
+        }
+
+        flock($this->filePointer, LOCK_UN);
+        fclose($this->filePointer);
+        $this->filePointer = null;
+
+        usleep(100 + rand(0, 100));
```
albe: This usleep can be moved up before the return false, since it's only needed if locking failed and will be reattempted, in order to prevent deadlocks due to two processes synchronously retrying to acquire the same lock.

albe:
Just move the usleep, other than that looks good from reading

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
kitsunet: Uhm nope, see docblock comment and how the return value is used. If this returns true then another lock attempt should be made, so we need to sleep beforehand.

albe: Ah sorry, my bad. Inverse logic there

andigehle:
+1 Works with Flow 2.3.8 and PHP<5 when replacing ::class constants

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
```diff
@@ -51,56 +55,97 @@ class FlockLockStrategy implements LockStrategyInterface
...
+        $temporaryDirectory = Files::concatenatePaths([$environment->getPathToTemporaryDirectory(), 'Lock']);
+        Files::createDirectoryRecursively($temporaryDirectory);
+        self::$temporaryDirectory = $temporaryDirectory;
+    }
+
+    /**
+     * Tries to open a lock file and apply the lock to it.
+     *
+     * @param boolean $exclusiveLock
+     * @return boolean True if a lock could not be aquired, false if the lock was successful.
```
albe: After thinking about that again, that doc reads really strange: tryLock -> "True if .. not, false if ... successfull."
Would maybe make sense to invert the return value.

kdambekalns: Agreed. Even if where this is used it sounds logical (`while ($continueTrying)`, it would be easier to understand here. And then when using it just do `while ($lockAcquired !== true)`.

kitsunet:
There we go, this now has nicely forward logic, class constants are replaced for 2.3, all tests work, manual testing works, checked that lock files are removed... merge, merge pls :)

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
kitsunet: changed the logic now

kdambekalns:
Looks good.

albe:
:+1:

PR #139, opened 2015-11-25
[TASK] Replace Security_Randomizer with PHP7 random functions

kitsunet:
Replaces the old Security_Randomizer that we shipped with Flow
by random_bytes and random_int that come with PHP7 and adds a
polyfill library for lower versions.


kitsunet:
Replaces #5

kitsunet:
Travis doesn't checkout dependencies, so tests won't work correctly.

foerthner:
Looks really good to me. So +1 by reading! Also the functional tests run fine. I just had problems running unit tests, as the library seems not to be loaded in the testing context. As soon as this is fixed, I guess this one can go in.

kitsunet:
Sorry @foerthner I will probably rework this change again because I just was remembered that PHP7 brings new methods for secure random data and there is a polyfill for lower versions. I think that makes even more sense to use because then we can just remove the dependency when we support only PHP7.

So :-1: please wait for new change

Note to self: https://github.com/paragonie/random_compat

kitsunet:
There we go... :+1: again

kdambekalns:
Looks good, but… is the polyfill as good as the real thing?

kitsunet:
It can be considere quite good. Depends on your system obviously as it uses the usual sources like openssl, urandom etc. etc.
But the company seems to have a solid understanding of security, also it supports and suggests libsodium php extension (https://download.libsodium.org/doc/) which seems to be pretty good.
Additionally they list a handful of people that reviewed it one of them being the author of the first library (ircmaxwell) that I suggested.

kdambekalns:
Ok, sounds good.

kdambekalns:
Just merged master into this, to resolve the conflict (`composer.json`, trivial one).

kitsunet:
thanks

kdambekalns:
Ran unit and functional tests locally with PHP 7: :ok: 
Ran unit and functional tests locally with PHP 5.6: :ok: 

:+1:

kitsunet:
Yay, a +1 from Andi would be great, then this should be good to go IMHO

foerthner:
yep, I'm at it. you already can have :+1: by reading. Tests are currently running...

foerthner:
Ok, another :+1: for running tests on PHP 5.5. I guess this one can be merged now!

PR #140, opened 2015-11-25
[DOCS] Fix typo in security code examples

grebaldi:


albe:
The change is good, but there is no "[DOCS]" tag, it should rather be "[TASK]". That can be fixed on merge though. :+1:

kitsunet:
:+1: merging and fixing the tag

PR #141, opened 2015-11-26
[TASK] Add missing compression support in RedisBackend

dfeyer:
Resolves: FLOW-420

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/RedisBackend.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/RedisBackend.php
@@ -78,6 +78,11 @@ class RedisBackend extends AbstractBackend implements TaggableBackendInterface, 
     protected $database = 0;
 
     /**
+     * @var integer
+     */
+    protected $compressionLevel = 0;
+
+    /**
      * @param \TYPO3\Flow\Core\ApplicationContext $context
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/RedisBackend.php
```diff
@@ -415,6 +416,43 @@ public function setDatabase($database)
     }
 
     /**
+     * @param integer $compressionLevel
```
aertmann: @return void

aertmann:
:+1: by reading aside from missing return

robertlemke:
@dfeyer good catch!

albe:
:+1: by reading, but would it maybe make sense to add a compressionThreshold option like in memcache, so that only data larger than X bytes is compressed?

robertlemke:
@albe yes, good idea. In that case we might also enable compression by default, because we'll have to unzip the data when reading any time then.

kitsunet:
IMHO we could make that two separate commits, first introducing compression with this commit and then adding a threshold and automation in a separate commit as that is more complicated and IMHO needs careful measurements of the speed benefit vs. overhead.

kitsunet:
that said :+1: for this commit besides askes comment which is valid but wouldn't be a blocker for merge for me :)

PR #142, opened 2015-11-26
[TASK] Adjust exception handling for compatibility with PHP 7

robertlemke:
This change adjusts the non-public API of Flow's exception handling to
fit the new exception types in PHP 7 (`\Throwable`). It does not yet
take advantage of the new possibilites, but rather makes the existing
code compatible with both, PHP 5 and PHP 7.


kdambekalns:
Looks good, didn't test (sorry).

dfeyer:
:+1: make sense and work fine, next step add Logger::logThrowable currently it's a bit to silent

robertlemke:
Closing this PR in favour of #143 because I think that Neos 2.0 / Flow 3.0 should also be compatible with PHP 7.

PR #143, opened 2015-11-27
[TASK] Adjust exception handling for compatibility with PHP 7

robertlemke:
This change adjusts the non-public API of Flow's exception handling to
fit the new exception types in PHP 7 (`\Throwable`). It does not yet
take advantage of the new possibilities, but rather makes the existing
code compatible with both, PHP 5 and PHP 7.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Error/AbstractExceptionHandler.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Error/AbstractExceptionHandler.php
@@ -75,10 +75,10 @@ abstract class AbstractExceptionHandler implements ExceptionHandlerInterface
     /**
      * Handles the given exception
      *
-     * @param \Exception $exception The exception object
+     * @param object $exception The exception object - can be \Exception, or some type of \Throwable in PHP 7
      * @return void
      */
-    public function handleException(\Exception $exception)
+    public function handleException($exception)
     {
         // Ignore if the error is suppressed by using the shut-up operator @
         if (error_reporting() === 0) {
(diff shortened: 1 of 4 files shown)


robertlemke:
:+1: 

Considering the :+1: from @kdambekalns and @dfeyer in #142 I'd say it's safe to merge this one, unless anyone objects today.

kdambekalns:
Travis passes, I tested locally (unit/functional) and looked at the code again: :+1:

bwaidelich:
FYI: This is a breaking change for packages implementing `ExceptionHandlerInterface` / extending one of the existing exception handlers.
As mentioned in the commit message those classes are not part of the public API but I wonder if there's a way to make those packages work with Flow 2.x *and* 3.0.x

kitsunet:
dang :( good catch

kdambekalns:
No _dang_ at all, it's non public API.

bwaidelich:
not public API but very likely to be used - at least in larger projects you'll need custom error handling (as described in http://flowframework.readthedocs.org/en/3.0/TheDefinitiveGuide/PartIII/ErrorAndExceptionHandling.html#exception-handlers) and for custom logger packages like https://github.com/yeebase/Yeebase.Graylog/

kdambekalns:
… and there is no way to solve this for both versions, except using `versions_compare()` and doing ugly things.

kdambekalns:
Sonthis is actually used? Ok… my ignorance then. But then all existing implementations would need to have the same switch on both versions.

bwaidelich:
> and there is no way to solve this for both versions

Right, but we could at least remove the type hint for all supported branches to increase compatibility

kdambekalns:
Well, see discussion of today - PHP7 for 2.3 is in the works, so it has the same change…

bwaidelich:
FYI: It has been merged with #160 - which means it's a possible breaking change (not in the strict sense of modified API) for upgrades from 2.3.8 to 2.3.9, but I think that's the lesser of two evils

aertmann:
Related #303

PR #144, opened 2015-11-30
Prevent invalid Cookie pairs from raising errors

kitsunet:
Formally a Cookie header should consist of semi-colon separated pairs
of ``key=value`` but some clients might sent invalid cookie headers
resulting in a notice raised when there was no equals sign to split a
pair on.


kitsunet:
Corrected branch, replaces #6

aertmann:
:+1:

PR #145, opened 2015-12-01
TASK: Compile CommandController arguments statically

kitsunet:
Instead of using the ReflectionService at runtime to determine
arguments for commands they will be compiled statically and just
read from the array. Additionally this was centralized into the
``CommandManager`` as ``CommandController``, ``RequestBuilder`` and
``CommandManager`` all fetched similar information from the
``ReflectionService``.

NEOS-1294

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cli/CommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cli/CommandController.php
@@ -13,8 +13,8 @@ namespace TYPO3\Flow\Cli;
 
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Mvc\Controller\Argument;
-use TYPO3\Flow\Mvc\Controller\ControllerInterface;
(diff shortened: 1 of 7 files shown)


kitsunet:
This is a reopened #2 because travis didn't properly connect in the old PR...

kitsunet:
YAY, and I fixed the tests, so :+1: from my side.

aertmann:
:+1: by reading

nice cleanup and performance improvement

robertlemke:
:+1: verified

But why? I like the clean up part, but I can't measure any performance difference with and without the patch. I tried 6 different command with and without arguments in Development Context and without. In Development Context I measure 21 ms for each command (before it actually begins the task) and in Production Context 14ms.

kitsunet:
The main thing is that I want to see if we can actually come to a point where the ReflectionService is no longer needed for most runtime actions.

skurfuerst:
:+1: by reading; I agree with the idea of getting rid of dynamic ReflectionService calls!

PR #148, opened 2015-12-01
[TASK] Adjust generated proxy code to PHP 7 uniform variable syntax

robertlemke:
This change contains a fix in the "related entities" proxy class code
which makes the generated code PHP 7 compatible.

Related: NEOS-1608


kitsunet:
:+1: The relevant code in the trait in master already contains this fix, so merging this.

PR #149, opened 2015-12-01
TASK: Make Http\Request easier to use for outbound Requests

kitsunet:
Currently the ``Http\Request`` defaults to php://input for
content. This is good for a request coming in to Flow, but rather
error prone for outbound requests as it can lead to different behavior
depending if the Request is created during a GET, POST or even CLI
call. So sometimes outbound requests might work and other times they
won't work. This change adjusts the defaul to have the Request an
empty ``Content`` (body) and sets it to NULL (which will trigger the
fallback to php://input) just if the Request is created by using
``Request::createFromEnvironment()``

Resolves: FLOW-235

### DIFF
--- a/TYPO3.Flow/Tests/Unit/Http/RequestTest.php
+++ b/TYPO3.Flow/Tests/Unit/Http/RequestTest.php
@@ -311,6 +311,7 @@ class RequestTest extends UnitTestCase
         file_put_contents('vfs://Foo/content.txt', $expectedContent);
 
         $request = Request::create(new Uri('http://flow.typo3.org'));
+        $request->setContent(null);
         $this->inject($request, 'inputStreamUri', 'vfs://Foo/content.txt');
 
         $actualContent = $request->getContent();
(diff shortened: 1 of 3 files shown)


kitsunet:
Replacement for #8 where Traivs didn't work...

albe:
Makes sense, :+1: by reading

kdambekalns:
Looks good, anything else, @robertlemke ?

TYPO3.Flow/Tests/Unit/Http/RequestTest.php
```diff
@@ -311,6 +311,7 @@ public function getContentReturnsTheRequestBodyContent()
         file_put_contents('vfs://Foo/content.txt', $expectedContent);
 
         $request = Request::create(new Uri('http://flow.typo3.org'));
```
robertlemke: hey, you could have changed the test URI to something non-TYPO3 ;-)

robertlemke:
The change makes sense, but I can't assess if this will break someone's code. I'd guess that the risk is minimal.

:+1: but can't test it right now. Did anyone?

kitsunet:
Back then when I had the use case I tested it. I think the risk is minimal, basically people who had the problem I am trying to fix will already do a setContent('') or didn't use our Request at all and createFromEnvironment() should be the only source of truth if you want a request build from incoming values.

PR #150, opened 2015-12-02
[TASK] Remove typo3.org SSO provider from Flow

kdambekalns:
This is no longer used and should have never been part of the framework
itself, to begin with.


robertlemke:
:+1:

kitsunet:
:+1:

kitsunet:
Ok I missed that this is for 2.3, I wouldn't have :+1: in that case. It didn't even occur to me to remove a class from a released version....

robertlemke:
@kitsunet but I think there's really no one who can use that feature, because you would need root access to typo3.org in order to do so.

kitsunet:
I am not sure if maybe http://forger.typo3.org/ uses it^^

PR #151, opened 2015-12-02
[TASK] Remove unused method in SimpleFileBackend

kitsunet:
The ``generateTemporaryPathAndFilename`` is no longer in use
since the introduction of locking last year, it can therefore
be removed.


albe:
:+1:

PR #152, opened 2015-12-03
[TASK] Adjust check for maximum path length

kitsunet:
The SimpleFileBackend did a check for maximum path length
on construction but reserved an arbitrary length of 23 for cache
entry identifiers. Many identifiers are longer though and so even
if the exception was not triggered cache entries could fail to be
written.
This change moves the check to after a failed cache writing attempt
to check against the actual cache entry path.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
@@ -11,6 +11,7 @@ namespace TYPO3\Flow\Cache\Backend;
  * source code.
  */
 
+use TYPO3\Flow\Cache\Exception;
 use TYPO3\Flow\Cache\Frontend\PhpFrontend;
 use TYPO3\Flow\Cache\Frontend\FrontendInterface;
 use TYPO3\Flow\Annotations as Flow;
@@ -85,7 +86,7 @@ class SimpleFileBackend extends AbstractBackend implements PhpCapableBackendInte
      *
      * @param \TYPO3\Flow\Cache\Frontend\FrontendInterface $cache The cache frontend
      * @return void
-     * @throws \TYPO3\Flow\Cache\Exception
+     * @throws Exception
      */
     public function setCache(FrontendInterface $cache)
     {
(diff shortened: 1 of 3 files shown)


kdambekalns:
Well, the test needs to be adjusted, since the check is triggered differently now.

kitsunet:
Tada! Cleanly refactored :)

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
```diff
@@ -396,4 +395,31 @@ public function rewind()
             $this->cacheFilesIterator->next();
         }
     }
+
+    /**
+     * @param string $cacheEntryPathAndFilename
```
aertmann: @return void

aertmann:
:+1: by reading

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
kitsunet: you and your return voids.... ;)

Can we remove that from the coding guideline? Or put it as CAN HAVE instead of MUST

kdambekalns: > Am 03.12.2015 um 13:26 schrieb Christian Müller <notifications@github.com>:
> 
> In TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php:
> 
> > @@ -396,4 +395,31 @@ public function rewind()
> >              $this->cacheFilesIterator->next();
> >          }
> >      }
> > +
> > +    /**
> > +     * @param string $cacheEntryPathAndFilename
> you and your return voids.... ;)
> 
> Can we remove that from the coding guideline? Or put it as CAN HAVE instead of MUST
> 

Feeding trolls here, but: No, because then you no longer know if someone has forgotten something or it was intended.

kitsunet: ok, ok, I find that is a weak argument but it is one, so lets keep it.

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
```diff
@@ -396,4 +395,32 @@ public function rewind()
...
+     * @return void
```
kitsunet: There we go :)

albe:
So, what exactly is the reason to postpone the length check until after the write attempt?

kdambekalns:
@albe Because the identifier can be of variable length, so one could fit, another not. Checking every time is a waste of resources, checking in case of failure helps with debugging…

albe:
@kdambekalns ah ok, makes sense then :+1: by reading from me

PR #153, opened 2015-12-03
[BUGFIX] Fix PHP 5.3 compatibility issues

kdambekalns:
The 2.3 branch should be compatible to PHP 5.3, but since we did not have automated
testing in place on that version of PHP, some incompatibilties have crept in.

This fixes those issues and enables the unit and functional tests to be run on PHP 5.3
on Travis CI.

Also the Behat tests for Flow are run with this, since there is no reason they shouldn't.


kdambekalns:
This depends on https://github.com/neos/flow-development-distribution/pull/3

kitsunet:
aren't the traits used in neos? respectively the typo3cr behat tests?

kitsunet:
yep see `/Volumes/CaseSensitive/Work/PhpstormProjects/neos-dev-master/Packages/Neos/TYPO3.TYPO3CR/Tests/Behavior/Features/Bootstrap/FeatureContext.php`
which doesn't even have a namespace....

kdambekalns:
And the counterpart for Neos: https://github.com/neos/neos-development-collection/pull/236 /cc @kitsunet

kitsunet:
Great work :) thanksa lot

kdambekalns:
Ok, looks good to me now. _sigh_

PR #154, opened 2015-12-03
BUGFIX: Allow composite keys over foreign entities

albe:
Currently, the implementation of the FlowAnnotationDriver prevents
composite primary keys including an foreign entity reference to
work, as in the example of doctrine:
http://docs.doctrine-project.org/en/latest/tutorials/composite-primary-keys.html#identity-through-foreign-entities

This change adds the required mapping informations, which is only
a first step towards full composite key support.

FLOW-259 #close


albe:
There was already a discussion with Jacob about the further consequences of supporting composite keys on gerrit: https://review.typo3.org/#/c/30659/
I'd still like to get this merged as is ASAP, because it basically just fixes a missing feature from Doctrine in the AnnotationDriver and is really simple, but helps already.

kitsunet:
Looks good, will take a closer look

albe:
@kitsunet ping ;) maybe @kdambekalns might also want to take a quick look

kitsunet:
I like :+1:

kdambekalns:
:+1:

PR #155, opened 2015-12-04
[BUGFIX] Lock files shouldn't be opened twice

kitsunet:
Prevents opening lock files twice, first in read then in
write mode. Only one open should happen if successful.
The change fixes that.


kdambekalns:
Same effect, still tries to open a non-existig file for reading. Why is that needed at all? Why not open for writing right away?

kitsunet:
Good question, that comes from the original code...

kitsunet:
Oh right, because before the lock file could exist or not. Now we can be pretty sure it doesn't exist... So yes, will change it to write

albe:
Looks good :+1:

daniellienert:
:+1: by reading

PR #156, opened 2015-12-04
[TASK] Use wrapper for phpunit to catch segfaults

kdambekalns:
This uses a wrapper around the unit tests to handle an exit code of 139
as a non-error.

This happens when testing on Travis CI and while those error happen, we
cannot do much else about it.


kdambekalns:
:+1: – I did some testing locally, and a real failing tests leads to an exit code of 1 as expected.

PR #159, opened 2015-12-04
[TASK] Add DBAL minimum version to make JSON type work

kitsunet:
The Json column type is support in doctrine DBAL only from
version 2.5.0 onwards. The current dependencies allow
installing it but it won't happen automatically, this
change makes sure that dbal is installed in the right
version.


kitsunet:
ok, seems this alone won't work but I am not yet sure why....

PR #160, opened 2015-12-05
[TASK] PHP 7 compatibility

aertmann:


kitsunet:
same comment as for #143 goes for this too :(

bwaidelich:
FYI, re #143 : With this change the interface is in sync with 3.0 and master but people might have to adjust their code when upgrading from 2.3.8 to 2.3.9.  Maybe we should mention that in the release notes!?

aertmann:
Related #303

PR #164, opened 2015-12-07
[BUGFIX] Documentation example for translating error messages uses wrong parameter name 'key' instead of 'id'

dlubitz:


aertmann:
:+1: should also be fixed in 2.3 though, but there there's two occurrences that need to be fixed so needs a separate change anyway

kdambekalns:
@aertmann where in 2.3 does this need to be fixed? I find only the equivalent to this one…

aertmann:
https://github.com/neos/flow-development-collection/blob/2.3/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Internationalization.rst

actually only one occurrence

kdambekalns:
@dlubitz would you close this an create a new one against the 2.3 branch? thanks!

dlubitz:
@kdambekalns Sure. Is there an easy way to do this? Or do I have to create a new branch against 2.3 and commit the change again?

kdambekalns:
No easier way, no… Thanks!

dlubitz:
New PR: https://github.com/neos/flow-development-collection/pull/180

PR #166, opened 2015-12-07
[BUGFIX] Uniform variable syntax for proxy trait

kitsunet:
Fixes a leftover place which didn't use the uniform variable syntax
and therefore caused notices with PHP7.


bwaidelich:
:+1: Thanks!

PR #167, opened 2015-12-08
[BUGFIX] Now extends DateTime for backwards compatibility

kitsunet:
This partially reverts ``6b1448d2c540d79e882f527449e05660bc5095e3``
which changed ``\TYPO3\Flow\Utility\Now`` to extend the
``DateTimeImmutable`` class introduced with PHP 5.5.
As this is potentially breaking this part of the change is reverted
and will be reintroduced as breaking change in the next major
version.


albe:
:+1:

bwaidelich:
:+1: Thanks a lot!

PR #169, opened 2015-12-08
[BUGFIX] Use instanceof instead of relying on typed parameter

kitsunet:
The Configuration class relied on type hints to enforce the validity
of user-supplied data. This was explicitly tested, so the assumption is,
this was supposed to be tested properly. An instanceof check is now used
and a proper exception is thrown (instead of a PHP error).

Along the way this fixes some failing unit tests on PHP 7, since the old
assertion in the tests no longer holds.


bwaidelich:
:+1: by reading

albe:
:+1: by reading

kdambekalns:
Re-open, please: `setProperties()` vs `setArguments()`, this is indeed still needed.

kdambekalns:
:+1: - I am pretty sure I had the exact same change, so that's why I got confused.

kdambekalns:
The StyleCI check is utterly unrelated to this change. :)

PR #171, opened 2015-12-08
[BUGFIX] Support for special characters in ``PositionalArraySorter`` keys

bwaidelich:
When using the ``PositionalArraySorter`` one can position keys
relative to other keys with ``before/after <key>``.
But for the ``<key>`` only ``[a-zA-Z0-9]`` were allowed limiting the
functionality especially when dealing with package keys that contain
a dot.

This change adjusts the regular expression to allow any string to be
referenced.

Fixes: FLOW-422


bwaidelich:
:+1: This is required and tested with #129

kdambekalns:
:+1:

PR #172, opened 2015-12-10
BUGFIX: Set the subpackage key when setting the package key in forward()

cerlestes:
There's no logical case where you'd want to forward to a different package but stay within the same subpackage key. That's why the subpackage key should be (re)set whenever the package is changed when forwarding.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/AbstractController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/AbstractController.php
@@ -196,8 +196,6 @@ abstract class AbstractController implements ControllerInterface
         }
         if ($packageKey !== null) {
             $nextRequest->setControllerPackageKey($packageKey);
-        }
-        if ($subpackageKey !== null) {
             $nextRequest->setControllerSubpackageKey($subpackageKey);
         }
 


TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/AbstractController.php
```diff
@@ -196,8 +196,6 @@ protected function forward($actionName, $controllerName = null, $packageKey = nu
         }
         if ($packageKey !== null) {
             $nextRequest->setControllerPackageKey($packageKey);
-        }
```
kitsunet: This basically solves the case I guess but opens another bug if you don't set $packageKey but $subpackagekey because that won't be picked up then

cerlestes: You cannot set the subpackageKey without setting the packageKey though.

albe: Right, subpackageKey is not a method argument, but is directly derived from packageKey.

Test needs to be adjusted though, and there seems to be an unrelated CGL error - should maybe be rebased

Also, does this maybe make sense for 2.3/3.0 too?

albe:
Closing in favor of #319

PR #173, opened 2015-12-11
[BUGFIX] Ignore warning on failed stat

kitsunet:
We check if the stat call in ``FlockLockStrategy`` returned false anyway and
can safely suppress the generated warning.


kdambekalns:
:+1:

PR #174, opened 2015-12-11
[TASK] Break after 10000 lock attempts

kitsunet:
The ``FlockLockStrategy`` should at some point always be able
to aquire a lock even if it takes a while. But to prevent really
long running processes that interlock each other a natural
boundary to the amount of lock attempts is introduced.
The strategy will throw an exception after 10000 tries to aquire
the lock file. Depending on the results of random the maximum time
to reach this upper boundary is about 2 seconds.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
@@ -56,6 +56,7 @@ class FlockLockStrategy implements LockStrategyInterface
      * @param string $subject
      * @param boolean $exclusiveLock TRUE to, acquire an exclusive (write) lock, FALSE for a shared (read) lock.
      * @return void
+     * @throws LockNotAcquiredException
      */
     public function acquire($subject, $exclusiveLock)
     {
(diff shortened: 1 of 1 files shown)


kitsunet:
@kdambekalns @albe We talked about this. Might make sense....

kdambekalns:
Just killed two stray httpd processes that were fighting over some lock. Might be good to get this in… :)

kdambekalns:
Looks solid, and concurrently calling a Neos site in dev context after emptying the caches works fine.

kdambekalns:
Production context also works for me.

aertmann:
:+1: by reading, makes sense a very helpful

albe:
:+1: by reading

robertlemke:
Sounds helpful! 

Just a question (sorry ...) - anything speaking against using `sleep()` or `time_nanosleep()` instead and just making 2 iterations? 10000 iterations will unnecessarily put load on the CPU and the time PHP needs to run these iterations varies from system to system ...

kitsunet:
Two iterations?

kitsunet:
Currently we loop until we aquired a lock file. If aquiring the lock file failed we usleep and try again. This just additonally adds an exception if the loop was run 10000 times to prevent infinite loops if a lock could never be aquired.

aertmann:
it already uses usleep between the tries https://github.com/neos/flow-development-collection/blob/lock-strategy-break/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php#L125, so if you want less tries and more sleep that one can be increased instead

robertlemke:
Okay, thanks for the pointer. That didn't get clear from the code changed in this PR. 

To me, the `usleep()` in the last line of `tryToAquireLock()` is a bit unexpected, I think it should rather be in the while loop in `acquire`.

By the way, just saw that `tryToAquireLock` is misspelled – should be `tryToAcquireLock`.

kitsunet:
The method encapsulates everythign that happens in one loop wi regardless of the cycle (that's why I kept the if $i in the while loop. For me that makes a lot of sense to see what happens in one cycle in one method. So IMHO the usleep is correct there.

The misspelling can surely be fixed in a separate commit.

robertlemke:
@kitsunet I disagree about the placement of `usleep()`, but that's fine, doesn't make sense to argue on that detail.

:+1:

PR #178, opened 2015-12-11
[TASK] Object shutdown should be run without security checks

kitsunet:
The ``ObjectManager`` will shutdown configured objects when
shutdown is called on it. As the session is shutdown as one
of those configured objects reliance on it and therefore on the
Security\Context is not possible. To prevent any errors stemming
from the fact that the security couldn't be initalized while in
shutdown we execute the whole shutdown without security.

Due to the special nature of the ``CompileTimeObjectManager``
we need to call the shutdown as before, but security will not
be used during compile time.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
@@ -458,4 +458,15 @@ class CompileTimeObjectManager extends ObjectManager
         array_pop($this->objectNameBuildStack);
         return $object;
     }
+
+    /**
+     * Shuts down this Object Container by calling the shutdown methods of all
+     * object instances which were configured to be shut down.
+     *
+     * @return void
+     */
+    public function shutdown()
+    {
+        $this->shutdownLifecycleObjects();
+    }
 }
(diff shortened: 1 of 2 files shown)


kitsunet:
@robertlemke @kdambekalns @albe This is the result of the discussion in Slack.

@bwaidelich @foerthner The problem this solves is that the ResourceManager sometimes removes unnecessary resources in shutdown, because that uses the DB the Context is triggered which tries to use the Session but it is already closed due to shutdown.... So this fix could also be locally applied to the ResourceManager (you can see the issue if you try to crop in Neos master) but it seems to make sense for all things that happen during shutdown.

kdambekalns:
Looks good, but let's see what the _security experts_ have to say… :)

robertlemke:
@bwaidelich / @foerthner are you up for a quick assessment? Can't imagine any drawback with this solution?

Either way, I suggest to merge this today, because currently `master` is broken (cropping doesn't work) and this seems to fix the issue.

foerthner:
Hi everyone! Thanks for finding this one! In general this fix will work of course and probably is the only way around. At least I don't see a better solution either, currently. However, if we merge this we have to make sure to add a big red warning sign to the documentation of lifecycle shutdown methods. As these methods might be implemented in application code as well, developers have to be aware that their code will be executed without any security checks at this point.

robertlemke:
@foerthner spontaneous idea: if we would make sure that the shutdown objects of the Flow package are the last ones called, we could only call those without authorization checks.

foerthner:
yes, that sounds like a good option, if we can make sure that application code ist always executed with security checks. I guess it's a good idea to shutdown the framework at the end anyways.

robertlemke:
FYI, pushing a follow up

robertlemke:
Closing this one in favour of #181

PR #179, opened 2015-12-13
[BUGFIX] Don't unlink lock file

kitsunet:
Unlinking the lock file again causes race conditions which the
locking should prevent in the first place. Until we found out a
solid way to do this we need to keep those files even though
this is not optimal due to the high amount of inodes those lock
files use up.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
@@ -68,7 +68,7 @@ class FlockLockStrategy implements LockStrategyInterface
         $aquiredLock = false;
         $i = 0;
         while ($aquiredLock === false) {
-            $aquiredLock = $this->tryToAquireLock($exclusiveLock);
+            $aquiredLock = $this->tryToAcquireLock($exclusiveLock);
             $i++;
             if ($i > 10000) {
                 throw new LockNotAcquiredException(sprintf('After 10000 attempts a lock could not be aquired for subject "%s".', $subject), 1449829188);
@@ -101,7 +101,7 @@ class FlockLockStrategy implements LockStrategyInterface
      * @return boolean Was a lock aquired?
      * @throws LockNotAcquiredException
      */
-    protected function tryToAquireLock($exclusiveLock)
+    protected function tryToAcquireLock($exclusiveLock)
     {
         $this->filePointer = @fopen($this->lockFileName, 'w');
         if ($this->filePointer === false) {
@@ -150,11 +150,13 @@ class FlockLockStrategy implements LockStrategyInterface
     {
         $success = true;
         if (is_resource($this->filePointer)) {
+            // FIXME: The lockfile should be unlocked at this point but this will again lead to race conditions,
+            // so we need to find out how to do this in a safe way. Keeping the lock files is very inode intensive
+            // and should therefore change ASAP.
             if (flock($this->filePointer, LOCK_UN) === false) {
                 $success = false;
             }
             fclose($this->filePointer);
-            Files::unlink($this->lockFileName);
         }
 
         return $success;


kitsunet:
@kdambekalns @albe @robertlemke That safest for now. It's really bad to leave those files there but until we found a solid way to do this we should keep it at this. Probably a test would be awesome but that's hard due to the nature of the bug....

robertlemke:
Thanks a lot for diving into this issue!

If we aim for a shippable master, I'd give :-1: for this solution, as we can't leave it like that when we release the next Flow version. I know, 2.3+ are broken at the moment anyway, but let's still aim for a solution which we can theoretically leave as is for a release.

Is it feasible to add a garbage collection mechanism, similar to the one in `shutdownObject()`in https://github.com/neos/flow-development-collection/blob/master/TYPO3.Flow/Classes/TYPO3/Flow/Session/Session.php ?

What could be done, I hope, is unlinking all lock files whose `mtime` is older than a day.

kdambekalns:
I dislike this as well. We have so many issues with locking, I start to wonder why we started to work on it…!? What was it? Race conditions? Not much gained, it seems.

kdambekalns:
Ok. The removal of the file exactly like this seems to be the right way (I faintly remember discussing this in Nuremberg). If that doesn't work, what about unlinking the file after we remove the lock on it? It is done like it is done currently to avoid the lockfile being used again before deletion is attempted, but as we accumulate huge amounts of lockfiles, we do not seem to use the same file anyway?!

kitsunet:
Doesn't matter where we remove the lock file, I even tried in __destruct() (and everywhere around the unlock) the problem seems to be the time consumed in deletion and recreation which makes concurrent lock attempts to run out of tries.

kitsunet:
The main "problem" is that we non-exclusively lock for reading cache files, which means these locks are called for on ALL requests. That also means we rarely have expiring lock files especially over a longer timeframe. If we check mtime for lock files it's very likely that every single one was used in a 24h timespan (besides maybe stuff like content cache locks for pages that are seldom openened).
In theory the current concept works, we could crank up the amount of retries to something like 100000 which should reduce the probabillity of not being able to acquire the lock greatly. It could still happen on heavily loaded sites though.

albe:
I'd really like to dig into the real cause of this lock running out of tries. I really can't follow why even a very slow unlink should prevent the lock from getting acquired (unless unlink itself isn't atomic).

So just to get everyone on track of the current implementation details:
- we need to delete the lock file, because else we quickly fill inodes (really really quickly)
- unlink **before** unlock seems to work on *nix, but isn't possible on windows (can't delete an opened file)
- unlink **after** unlock creates a new race condition, where the file is unlinked while another process just fopened/flocked the file (at least on *nix, on windows that isn't possible as above)
- to work around this, we compare the inode of the fopened file against the inode of the filename after acquiring the lock - if they don't match, the file was unlinked and we just retry the lock (again, on windows this never happens)

as it seems, we somehow currently still run into this scenario with an infinite loop:
   process X (px): fopen, [lock, ..., unlock]
   p1: fopen, try lock
   px: fclose, unlink
   p1: fail inode check, reattempt
   px: fopen, [lock, ..., unlock]
   p1: try lock
   px: unlink
   *repeat at* p1: fail inode check, reattempt

But I can not see any code that could constantly cause one single lock file to be recreated and deleted within a timeframe that fits into the locking reattempt (unlock, fclose, a usleep(~200µs), an fopen, flock and two stat calls).

Only other loophole I can imagine happening: if the stat() call fails for any non-temporary reason, we currently run into an infinite loop (broken only by the reattempt limit). But what are the reasons for stat to fail (other than the file not existing anymore)?

Maybe we should also check for file_exists and if file_exists === true and stat === false, we throw an exception immediately instead, because there's something odd?

kitsunet:
I also want to find the root cause and fix it, but actually it is more stable with the files than now. 2.0.5 contained the changes and people have problems with it due to the exception. So I suggest to merge this, make a patch level release and then find a way to remove those files.

albe:
Sure sure, a quick solution is due, but let's maybe invest a bit more into finding a solution first before quickly reverting to old "not exactly as bad" behaviour.

kdambekalns:
So I gave in and actually applied this, and it did indeed fix the exceptions (retry count exceeded) for me.

The downside is, that around 4600 lock files are in my Lock folder now. But it seems that at least that number is not growing (or only slowly).

kitsunet:
It's simply the total amount of cache files you have. So the bigger the project the more lock files because more cache files.

kitsunet:
I have something in the pipe I can push tomorrow

kitsunet:
Ok, so first try I made was to lock the actual subject (again), that works fine because the lock itself no longer needs to take care of deleting, BUT I think I remember the reason again why we introduced the lock files, because then then the cache files (that are usually locked) can reside on filesystems that do not properly support flock like NFS.

robertlemke:
@kitsunet history repeating ... shared volumes (NFS ...) was the original reason why we didn't use real file locking in the first place when we created the very first file backend.

But let's imagine someone would point his file backend to a shared volume – how could we guarantee locking across multiple webserver sharing that volume? I think we can't. Putting an extra lock file onto the local filesystem makes it even worse.

kitsunet:
Exactly, so do we want to switch back to direct locking? That would make things much easier of course. And IMHO the only realistic way for read/write locking.

kitsunet:
Would be a bad idea to store the cache files via NFS anyway...

robertlemke:
@kitsunet I'd say yes, let's go lock the actual cache files. The question really is, why we introduced that in the first place - was it really because we want to enable people to put cache files on a shared volume? Or do we miss something now?

From my superficial view at the moment, I think that it will only (well, at least something) fix the issue with concurrency regarding the cache entries. But I don't think it will fix all possible race conditions related to proxy building in the Development context, because we'd need to limit file monitoring and re-compilation of proxy classes to a single process. So, I still expect weird things to happen when you are in Development context, change some code and then, for example, generate 20 thumbnails asynchronously.

kitsunet:
Sure, but that is a different issue. I would not mix that. It definitely also needs to be tackled of course.

kdambekalns:
That "compile step locking" is the site locking we do, maybe that needs improvement, too. But, a different issue…

robertlemke:
How about merging this one and then @kitsunet or someone else can come up with an improvement? The locking current master is really a problem.

:+1:

albe:
See also https://github.com/neos/flow-development-collection/pull/10#issuecomment-135978839 ff. - I already found that the "NFS" argument is no longer valid.

My final suggestion was to decouple FileBackend locking from the generic Lock class, because in FileBackend we can just lock the actual cache file directly. The generic Lock class can not do that and hence needs a race condition free implementation still, which could eventually also be supplied with other implementations like redis or memcache.

kitsunet:
@albe see #183 The DirectFlockLockStrategy should work just fine. IMHO locking inside the Backend is not a good idea.

kitsunet:
See #183

kitsunet:
Everyone is complaining about this issue. So better keep the huge amount of files for now and search a better solution with a working system....

Just two notes:

- The file is reused as the name is based on the subject. It just shows how many cache files we load...
- I have no satisfactory idea about cleanup for now besides avoiding the lock file in the first place as the lock file is potentially in use all the time on a loaded server.... See #183 and discussion there as well
- I will think further about possible cleanup scenarios
- as mentioned in #183 MUTEX would be an option and for that multiple implementations are available but that means no concurrent reads are possible, which we probably all can agree is a bad idea....

PR #180, opened 2015-12-14
BUGFIX: Fix documentation example

dlubitz:
The documentation example for translating error messages has to use parameter 'id' instead of 'key'.


dlubitz:
@kdambekalns

kdambekalns:
:+1:

PR #181, opened 2015-12-14
BUGFIX: Security context not available in shutdown lifecycle objects

robertlemke:
Because the order of shutdown methods being executed by the Object
Manager is undetermined, it may happen that `shutdownObject()` methods
relying on an initialized security context will fail because that
context is not available anymore. Additionally, Flow's own shutdown
object methods might fail because security checks are still active
even though the security framework is not available anymore.

This change makes sure that

1. shutdown methods of any other than the Flow package are executed first
2. Flow's own shutdown methods are called last, and without security checks

Due to the special nature of the CompileTimeObjectManager
we need to call the shutdown as before, but security will not
be used during compile time.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
@@ -458,4 +458,16 @@ class CompileTimeObjectManager extends ObjectManager
         array_pop($this->objectNameBuildStack);
         return $object;
     }
+
+    /**
+     * Shuts down this Object Container by calling the shutdown methods of all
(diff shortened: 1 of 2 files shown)


kitsunet:
I like! :+1:

foerthner:
:+1: by reading! Thanks a lot!

robertlemke:
Okay, I tried this locally by triggering a few parts in Neos which I remember caused problems during shutdown, among them also removing assets and cropping images.

:+1: let's merge this

kdambekalns:
:+1: by reading - if you wait with the merge a bit, I could actually test it ;)

robertlemke:
@kdambekalns sure, thanks!

kdambekalns:
I can haz cropped images again. Nice.

PR #182, opened 2015-12-14
BUGFIX: Properly escape sub process variables on windows

albe:
Windows SET command does not parse out quotes of the variable value but rather treats them as part of the value, which currently results in an error on windows since the fix for FLOW-381:

    Flow could not create the directory
    ""C:/workspace/Flow/Data/Temporary"/Development/".

Note the extra quotes around the temporary base path.

This change fixes that by properly escaping the SET command arguments on windows by using escapeshellcmd instead of escapeshellarg.

FLOW-381 #comment Regression fix for compilation on Windows

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
@@ -622,7 +622,8 @@ class Scripts
             if (DIRECTORY_SEPARATOR === '/') {
                 $command .= sprintf('%s=%s ', $argumentKey, escapeshellarg($argumentValue));
             } else {
-                $command .= sprintf('SET %s=%s&', $argumentKey, escapeshellarg($argumentValue));
+                // SET does not parse out quotes, hence we need escapeshellcmd here instead
+                $command .= sprintf('SET %s=%s&', $argumentKey, escapeshellcmd($argumentValue));
             }
         }
         if (DIRECTORY_SEPARATOR === '/') {


dlubitz:
:+1: by reading and testing on local windows 8/php5.5

kdambekalns:
No need to test this on non-windows, because it doesn't affect that setup. And no way to test this for me, since I don't have Windows. But I guess @albe and @dlubitz can be trusted, so… :+1:

bwaidelich:
Thanks a lot for this!!

aertmann:
https://jira.neos.io/browse/FLOW-425 was just reported, this change fixes that right?

dlubitz:
Exactly. We should create tickets for all bugs, even if we get it fixed directly. Jira should be the first point to figure out if a bug is already known to prevent multiple efforts.

----- Reply message -----
Von: "Aske Ertmann" <notifications@github.com>
An: "neos/flow-development-collection" <flow-development-collection@noreply.github.com>
Cc: "dlubitz" <lubitz@vivomedia.de>
Betreff: [flow-development-collection] BUGFIX: Properly escape sub process variables on windows (#182)
Datum: Do., Dez. 17, 2015 06:52
https://jira.neos.io/browse/FLOW-425 was just reported, this change fixes that right?

&mdash;

simonschaufi:
i spend quite a lot of time debugging this since the error output form subrequests are not very clear (and i figured out that flow calls many times itself, maybe thats the reason why its so slow on windows...)

Is it possible to do such tests with travis or does travis only support linux?

kitsunet:
No Windows Support in Travis.

albe:
Ah, my fault, should have checked jira before creating the PR so quickly. Thx @kdambekalns for adding the missing jira link

PR #183, opened 2015-12-16
BUGFIX: Lock strategy should not delete lockfile

kitsunet:
This splits the current lock strategy into two slightly different
strategies. The ``FlockLockStrategy`` will remain the default for
backwards compatibility and still create lock files and acquire locks
on those instead of the real subject. It doesn't try to delete those
lock files though.

Additionally the ``DirectFlockLockStrategy`` will acquire locks on the
subject itself thus avoiding the problem of leftover files, but this
works only if the cache files are stored on a filesystem that supports
locking, which is not the case for NFS.

If you store the caches on a regular filesystem you can switch to the
``DirectFlockLockStrategy`` by configuring the settings path:

TYPO3.Flow.utility.lockStrategyClassName

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/DirectFlockLockStrategy.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Lock/DirectFlockLockStrategy.php
@@ -1,0 +1,158 @@
+<?php
+namespace TYPO3\Flow\Utility\Lock;
+
+/*
+ * This file is part of the TYPO3.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Utility\Exception\LockNotAcquiredException;
+
+/**
+ * A flock based lock strategy.
+ *
+ * This lock strategy is based on Flock and will directly lock the subject file.
+ *
+ * @Flow\Scope("prototype")
+ */
+class DirectFlockLockStrategy implements LockStrategyInterface
+{
+    /**
+     * File name used for this lock
+     *
+     * @var string
+     */
+    protected $lockFilename;
+
+    /**
+     * File pointer if using flock method
+     *
+     * @var resource
+     */
+    protected $filePointer;
+
+    /**
+     * The subject under lock (usually a file path)
+     *
+     * @var string
+     */
+    protected $subject;
+
+    /**
+     * Is this lock an exclusive lock
+     *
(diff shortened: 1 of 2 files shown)


kitsunet:
This is made with full backwards compatibility in mind. The FlockLockStrategy behavior isn't changed (beside the fact that it doesn't try to delete the files anymore but that is the state it was in for quite some time now). Also the FlockLockStrategy is still the default. In master we could change the configuration to use the DirectFlockLockStrategy

kitsunet:
This is the better alternative to #179

albe:
1) that NFS not supporting flock no longer holds, it's an outdated myth
2) how does this fit with the Lock class being generically usable? If DirectFlockStrategy will be the default implementation, the Lock class can no longer be used on anything else but full filenames and will even error out or, in the non-fatal case, create lock files with fancy names in the current working dir

kitsunet:
The lock class can only be generally useable if we only lock exclusively or live with left over files, unless you have some super clever idea. I don't see it at the moment. But it's true, maybe that needs to be considered.

Any implementations out there are MUTEX implementations that only support exclusive locks. While that is fine it doesn't help us much...

kitsunet:
Do we state that this is generally usable somewhere?

albe:
> Do we state that this is generally usable somewhere?

Not that I know, but the way the code is designed, it just screams to be generically useable (strategy pattern and residing in Utility etc.)

albe:
> The lock class can only be generally useable if we only lock exclusively or live with left over files

Why? I mean, the lock type (exclusive/shared) could be an parameter of the lock constructor and the left over files is just an internal specific of the FlockLockStrategy. Either we get the race condition free unlinking working, or we need to provide an alternative LockStrategy that works generically. It could even be a mutex implementation, because having shared locking fall back to exclusive locking is not perfect performancewise, but the algorithms depending on the lock will still work correctly.

kitsunet:
Well the bad thing is that we rely on a global implementation, otherwise we could just define this as a file only implementation and use it in the file cache.
I must say I would like to avoid EX locks for everything as I think it's a major slow down, but surely that can be done as well. IMHO the Direct strategy is a valid way but I agree that it screams for generally usable so the Direct strategy would need to be "selected" from the code that wants the lock (like the file backend) if it knows that it has a valid file path to lock.

kitsunet:
ok, so we merge the other change for now and think this through some more?

albe:
I don't like it, but that's probably still the best thing to do. But we should really tackle this asap

> Direct strategy would need to be "selected" from the code that wants the lock
Could be a way

kdambekalns:
Is this going to be continued? When? By whom? I'm lost by now, when it comes to locking. Maybe @dfeyer and @kitsunet can decide the fate of this one…

kitsunet:
Probably not, we are currently talking about using ex locking everywhere because that would make the whole case much easier to handle.

PR #184, opened 2015-12-17
BUGFIX: Properly escape sub process variables on windows

albe:
Windows SET command does not parse out quotes of the variable value but rather treats them as part of the value,
which currently results in an error on windows since the fix for FLOW-381:

  Flow could not create the directory
  ""C:/workspace/Flow/Data/Temporary"/Development/".

Note the extra quotes around the temporary base path.

This change fixes that by properly escaping the SET command arguments on windows by using escapeshellcmd instead
of escapeshellarg.

FLOW-425 #comment Fix for 2.3.x


albe:
This is a PR to apply the bugfix from #182, which was wrongly targeted at 3.0 only, onto 2.3

kitsunet:
Thanks! Should be merged as soon as travis is done.

albe:
Merged with previous votes

PR #185, opened 2015-12-17
TASK: Do not parse request body for GET and HEAD requests

kitsunet:
A HTTP request with GET or HEAD method should not carry any body data.
Initialisation for the MediaTypeConverter and running it costs
unnecessary time in such requests and therefore the process is skipped
now for GET and HEAD requests.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/DispatchComponent.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/DispatchComponent.php
@@ -28,19 +28,19 @@ use TYPO3\Flow\Utility\Arrays;
 class DispatchComponent implements ComponentInterface
 {
     /**
-     * @Flow\Inject
+     * @Flow\Inject(lazy=false)
      * @var Dispatcher
      */
     protected $dispatcher;
 
     /**
-     * @Flow\Inject
+     * @Flow\Inject(lazy=false)
      * @var Context
      */
     protected $securityContext;
 
     /**
-     * @Flow\Inject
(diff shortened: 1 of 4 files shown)


kitsunet:
Note that @robertlemke mentioned it might be valid to have a body in a GET or HEAD request, so the condition might need to become a bit more sophisticated but not instanciating the media converter and running the body through it seemed to improve performance when I created this change this change back then.

Taken over from Gerrit 40510

robertlemke:
I don't remember in which context I said that, but yes, there are some odd situations in which you need to send a body along with a GET request (maybe even with a HEAD request). It's somewhat against the HTTP 1.1 spec, but it does happen.

For example, Elasticsearch offers such a feature - it doesn't require you to send a GET request though: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html

In general, as per the HTTP spec, it is generally allowed to send bodies with any kind of request:

> Roy T. Fielding Dec 12 2:52 PM
Yes. In other words, any HTTP request message is allowed to contain
a message body, and thus must parse messages with that in mind.
Server semantics for GET, however, are restricted such that a body,
if any, has no semantic meaning to the request. The requirements
on parsing are separate from the requirements on method semantics.

> So, yes, you can send a body with GET, and no, it is never useful
to do so.

> This is part of the layered design of HTTP/1.1 that will become
clear again once the spec is partitioned (work in progress).

https://groups.yahoo.com/neo/groups/rest-discuss/conversations/messages/9962

kitsunet:
closing as new commit has nothign to do with this old commit...

PR #186, opened 2015-12-17
FEATURE: Trait introduction via AOP

kitsunet:
This allows to introduce traits in generated proxy classes
via AOP. You can use the ``Introduce`` annotation with the
argument ``traitName`` to introdue the given trait into
generated proxy classes matching the pointcut.

For example::

  @Flow\Introduce(
    traitName="Vendor\Package\Service\Traits\ExampleTrait"
    pointcutExpression="class(TYPO3\Flow\Mvc\Controller\ActionController)"
  )

Would introduce the trait ``\Vendor\Package\Service\Traits\ExampleTrait``
into all proxies that either are the class or extend the class
``TYPO3\Flow\Mvc\Controller\ActionController``.

Resolves: FLOW-123

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
@@ -12,9 +12,11 @@ namespace TYPO3\Flow\Aop\Builder;
  */
 
 use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Aop\AspectContainer;
 use TYPO3\Flow\Aop\PropertyIntroduction;
 use TYPO3\Flow\Reflection\ClassReflection;
 use TYPO3\Flow\Reflection\PropertyReflection;
+use TYPO3\Flow\Aop\TraitIntroduction;
 
 /**
  * The main class of the AOP (Aspect Oriented Programming) framework.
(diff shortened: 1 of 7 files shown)


kitsunet:
Note that this change originally in Gerrit (34903) received a rather big discussion centering around the question if this feature is needed at all and if the implementation needs to be improved because properties from traits won't be picked up for persistence this way as the AOP building comes after reflection.

skurfuerst:
Hey,

I actually really like this feature/idea! To me, this could be a good and clean idea for extending classes... Of course; best would be to also update the Reflection information along the way if that was possible :-)

I didn't check the implementation yet.

All the best,
Sebastian

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -360,13 +362,21 @@ protected function buildAspectContainer($aspectClassName)
         }
         $introduceAnnotation = $this->reflectionService->getClassAnnotation($aspectClassName, \TYPO3\Flow\Annotations\Introduce::class);
         if ($introduceAnnotation !== null) {
-            if ($introduceAnnotation->interfaceName === null) {
-                throw new \TYPO3\Flow\Aop\Exception('The interface introduction in class "' . $aspectClassName . '" does not contain the required interface name).', 1172694761);
+            if ($introduceAnnotation->interfaceName === null && $introduceAnnotation->traitName === null) {
+                throw new \TYPO3\Flow\Aop\Exception('The introduction in class "' . $aspectClassName . '" does neither contain an interface name nor a trait name, at least one is required.', 1172694761);
```
gerhard-boden: maybe update error code?

kitsunet: good catch, will do!

kitsunet: ah, sorry, I missed that I just extended the old extension instead of adding another one. I think I would leave with with the old code in that case. I thought I duplicated a code there.

gerhard-boden: ah okay, thought the error could be updated since the error description changed

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -664,6 +677,34 @@ protected function getMatchingPropertyIntroductions(array &$aspectContainers, $t
     }
 
     /**
+     * Traverses all aspect containers and returns an array of trait
```
gerhard-boden: trait`s`s?

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -664,6 +677,34 @@ protected function getMatchingPropertyIntroductions(array &$aspectContainers, $t
...
+     * introductions which match the target class.
+     *
+     * @param array &$aspectContainers All aspects to take into consideration
+     * @param string $targetClassName Name of the class the pointcut should match with
+     * @return array array of trait names
+     */
+    protected function getMatchingTraitNamesFromIntroductions(array &$aspectContainers, $targetClassName)
+    {
+        $introductions = [];
+        /** @var AspectContainer $aspectContainer */
```
gerhard-boden: do we leave this comments for type hinting in?

gerhard-boden:
:+1: by reading
Makes sense as far as I can tell with my rather limited expertise how aspects are processed in Flow ;)

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
skurfuerst: yep :)

skurfuerst:
awesome change :)

PR #187, opened 2015-12-17
FEATURE: Resources can be published with relative symlinks

kitsunet:
Adds a new option to the ``FileSystemSymlinkTarget`` named
``relativeSymlinks`` which will calculate a relative path from
source to target and symlink to that relative path instead of an
absolute path.


kitsunet:
Moved over from Gerrit 30519, needs good testing...

kdambekalns:
@kitsunet so this is ready but needs thorough testing, right?

kdambekalns:
Merged master into the branch…

kitsunet:
Can we gt this done?

PR #188, opened 2015-12-18
TASK: Add method to get public URI by path

dfeyer:
This change introduces two new methods in the ResourceManager.

The method ``getPackageAndPathByPublicPath`` returns an array with package and
path. This method is useful to avoid duplication of this logic. The second method ``getPublicPackageResourceUriByPath`` is syntactic sugar.

This change also adapts the ``ResourceViewHelper`` to use the new methods.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
@@ -36,6 +36,8 @@ class ResourceManager
     const DEFAULT_STATIC_COLLECTION_NAME = 'static';
     const DEFAULT_PERSISTENT_COLLECTION_NAME = 'persistent';
 
+    const PUBLIC_RESSOURCE_REGEXP = '#^resource://(?<packageKey>[^/]+)/Public/(?<relativePathAndFilename>.*)#';
+
     /**
      * @Flow\Inject
      * @var ObjectManagerInterface
@@ -408,6 +410,39 @@ class ResourceManager
         /** @var TargetInterface $target */
         $target = $this->collections[self::DEFAULT_STATIC_COLLECTION_NAME]->getTarget();
         return $target->getPublicStaticResourceUri($packageKey . '/' . $relativePathAndFilename);
+    }
+
+    /**
+     * Returns the public URI for a static resource provided by the public package
+     *
+     * @param string $path The ressource path, like resource://Your.Package/Public/Image/Dummy.png
+     * @return string
+     * @api
+     */
+    public function getPublicPackageResourceUriByPath($path)
+    {
+        $this->initialize();
+        list($packageKey, $relativePathAndFilename) = $this->getPackageAndPathByPublicPath($path);
+        return $this->getPublicPackageResourceUri($packageKey, $relativePathAndFilename);
+    }
+
+    /**
+     * Return the package key and the relative path and filename from the given resource path
+     *
+     * @param string $path The ressource path, like resource://Your.Package/Public/Image/Dummy.png
+     * @return array The array contains two value, first the packageKey followed by the relativePathAndFilename
+     * @throws Exception
+     * @api
+     */
+    public function getPackageAndPathByPublicPath($path)
+    {
+        if (preg_match(self::PUBLIC_RESSOURCE_REGEXP, $path, $matches) !== 1) {
+            throw new Exception(sprintf('The path "%s" which was given must point to a public resource.', $path), 1450358448);
+        }
+        return [
+            0 => $matches['packageKey'],
+            1 => $matches['relativePathAndFilename']
+        ];
     }
 
     /**
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Uri/ResourceViewHelper.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Uri/ResourceViewHelper.php
@@ -13,6 +13,7 @@ namespace TYPO3\Fluid\ViewHelpers\Uri;
 
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\I18n\Service;
+use TYPO3\Flow\Resource\Exception;
 use TYPO3\Flow\Resource\ResourceManager;
 use TYPO3\Flow\Resource\Resource;
 use TYPO3\Fluid\Core\ViewHelper\AbstractViewHelper;
@@ -97,7 +98,11 @@ class ResourceViewHelper extends AbstractViewHelper
                 $package = $this->controllerContext->getRequest()->getControllerPackageKey();
             }
             if (strpos($path, 'resource://') === 0) {
-                list($package, $path) = $this->resourceManager->getPackageAndPathByPublicPath($path);
+                try {
+                    list($package, $path) = $this->resourceManager->getPackageAndPathByPublicPath($path);
+                } catch (Exception $exception) {
+                    throw new InvalidVariableException(sprintf('The specified path "%s" does not point to a public resource.', $path), 1386458851);
+                }
             }
             if ($localize === true) {
                 $resourcePath = 'resource://' . $package . '/Public/' . $path;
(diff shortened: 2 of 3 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,39 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
     }
 
     /**
+     * Returns the public URI for a static resource provided by the public package
+     *
+     * @param string $path
+     * @return string
+     * @throws Exception
```
aertmann: no exception here..

kitsunet:
Isn't that the task of the resource stream wrapper? I don't get the use case yet.

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,39 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
...
+     * @api
+     */
+    public function getPublicPackageResourceUriByPath($path)
+    {
+        $this->initialize();
+        list($package, $path) = $this->getPackageAndPathByPublicPath($path);
```
aertmann: packageKey and relativePathAndFilename would be more fitting variable names

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,39 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
...
+    public function getPackageAndPathByPublicPath($path)
+    {
+        $matches = array();
+        if (preg_match(self::PUBLIC_RESSOURCE_REGEXP, $path, $matches) === 1) {
+            $package = $matches[1];
+            $path = $matches[2];
+        } else {
+            throw new Exception(sprintf('The path "%s" which was given must point to a public resource.', $path), 1450358448);
+        }
+        return [$package, $path];
```
aertmann: could be shortened by !== 1 and then throw the exception and ``return [$matches[1], $matches[2]];``

could even make sense to use named groups in the regex to replace 1 and 2 with ``packageKey`` and ``relativePathAndFilename ``

dfeyer:
@kitsunet We use ResourceManager::getPublicPackageResourceUri() in Viewhelper and some other places, where we duplicated the logic to extract the package / path from the resource string, With the new thumbnail I need to duplicate this logic in some more place, so the syntactic sugar is welcome ```list($package, $path) = $this->resourceManager->getPackageAndPathByPublicPath($path);``` is a lots more nice than a regexp + matches and test everywhere :)

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
robertlemke: Agree with @aertmann - it's important that an API user knows exactly what to pass to that method and what will be returned. We have to be picky when its an API method, so please also add documentation to the params.

aertmann:
like the change, it's odd having no API for it and only support it in the view helper.. left some comments for code improvements though

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
robertlemke: Also, this method does not have any documentation in the doc comment, which I would require for `@api` methods. Regarding the exception message: just imagine you would see that exception without further context - something like "Could not determine the package key from the given public resource path '%s'".

kitsunet:
right, true enough. I need a bit closer look, generally agreed that we need to centralize the stuff. If nothing better comes up then probably here, I feel it could have a better place but well... Maybe it doesn't need to be `@api`?

robertlemke:
Also, please only merge this if it's urgently needed for a bugfix which needs to go into 3.1 (which may be the case of course). We are so close to the release and we should invest in our time into getting the release done instead.

dfeyer:
@robertlemke It will be used in bugfix ;) I just want to avoid adding more code duplication than we currently have

:+1: from my side, last version fix all comment from @aertmann

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,35 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
...
+
+    /**
+     * @param string $path
+     * @return array
+     * @throws Exception
+     * @api
+     */
+    public function getPackageAndPathByPublicPath($path)
+    {
+        $matches = array();
```
aertmann: this is unnecessary right?

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,35 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
     }
 
     /**
+     * Returns the public URI for a static resource provided by the public package
+     *
+     * @param string $path
```
kdambekalns: Please add some hint on what $path exactly is

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,35 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
...
+        $this->initialize();
+        list($packageKey, $relativePathAndFilename) = $this->getPackageAndPathByPublicPath($path);
+        return $this->getPublicPackageResourceUri($packageKey, $relativePathAndFilename);
+    }
+
+    /**
+     * @param string $path
+     * @return array
+     * @throws Exception
+     * @api
```
aertmann: description missing

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,35 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
     }
 
     /**
+     * Returns the public URI for a static resource provided by the public package
```
aertmann: mentioning the supported format here wouldn't hurt

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,35 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
...
+    public function getPublicPackageResourceUriByPath($path)
+    {
+        $this->initialize();
+        list($packageKey, $relativePathAndFilename) = $this->getPackageAndPathByPublicPath($path);
+        return $this->getPublicPackageResourceUri($packageKey, $relativePathAndFilename);
+    }
+
+    /**
+     * @param string $path
+     * @return array
```
kdambekalns: Please document the return value (what keys)

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
```diff
@@ -411,6 +413,35 @@ public function getPublicPackageResourceUri($packageKey, $relativePathAndFilenam
...
+     */
+    public function getPublicPackageResourceUriByPath($path)
+    {
+        $this->initialize();
+        list($packageKey, $relativePathAndFilename) = $this->getPackageAndPathByPublicPath($path);
+        return $this->getPublicPackageResourceUri($packageKey, $relativePathAndFilename);
+    }
+
+    /**
+     * @param string $path
```
aertmann: same with mentioning format here

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
aertmann: and should be [] if not unneccessary

kdambekalns:
So, please fix the test(s)… and my comments.

Also I'd like to see the fix this is needed for, so I can asses if it's urgent enough to be merged…

TYPO3.Flow/Classes/TYPO3/Flow/Resource/ResourceManager.php
dfeyer: It's documented in the @param

dfeyer:
Fix comments, but test are not broken by this change, locally I have a lots of failure for SimpleFileBackendTest, but if I try on 3.1 branch, I have the same error.

@kdambekalns What test did break with this PR for you ?

dfeyer:
@kdambekalns Forget my comment, I don't run the test for Fluid

kdambekalns:
I just looked at the Travis CI results… ;)

kdambekalns:
:+1:

PR #189, opened 2015-12-18
TASK: Use ReflectionObject in the Debugger to dump dynamic properties

radmiraal:
This change replaces the usage of ReflectionClass in the Debugger
class by ReflectionObject. This allows for debugging properties which
are set on the object during runtime but are not defined as property
on the class. Without this change those properties are ignored in the
var_dump command.


kdambekalns:
:+1: by reading. And I learned something new, thanks! :)

skurfuerst:
:+1: by reading a well. Also learned something new :)

PR #190, opened 2015-12-18
TASK: Remove typo3.org reference for exception codes

kitsunet:
The ``DebugExceptionHandler`` renders "more information" links
pointing to  wiki.typo3.org. This was never really used to document
exceptions and with the departure from the TYPO3 project this should
be removed. A possible replacement can be readded later.

FLOW-421 #close


dlubitz:
:+1: by reading

robertlemke:
:+1:

PR #195, opened 2016-01-04
FEATURE: Configurable Redirects

dfeyer:
Adds a new entity "Redirection" and a RedirectionService that enables
basic administration of redirects.

The RedirectionService will be triggered *before* the Routing Framework
kicks in and redirects to any configured target URI if applicable.
It however is only invoked when no cached route matches the current
request. Thus, if a redirection is added, depending route cache entries
are flushed automatically.

Redirects can be managed via new commands::

    ./flow redirection:list
    ./flow redirection:add source/path target/path 307
    ./flow redirection:remove source/path
    ./flow redirection:removeall
    ./flow redirection:removebyhost --host host.tld

In addition redirects can be imported and exported using::

    ./flow redirection:export
    ./flow redirection:import

The plan is, to create redirects automatically as soon as related
routes become obsolete due to changed entities for example.

NEOS-721 #comment basis for fixing the issue


dfeyer:
Import from Gerrit: https://review.typo3.org/#/c/24223

dfeyer:
By reading comments on Gerrit, I think the current status of this PR go in the correct direction:

* The Storage is now based on a Interface (so external package can implement more performant storage, async redirection generation, ...)
* The performance of the DB backend as been greatly improved by adding index on URI hash in the DB
* Redirection are now generated recursively (in the Neos PR, https://github.com/neos/neos-development-collection/pull/323)
* An aspect trigger an Signal on redirection creation for external package to listen and do some specific stuff

I think that basically we need to add domain support in the Neos PR, and write a solid test suite for this feature. But first I need some preliminary review to be sure that we are on the same track for the next steps

dfeyer:
All the logic as been remove from Flow core an moved to two new package:
- Neos.RedirectHandler (API)
- Neos.RedirectHandler.DatabaseStorage (Default storage implement, database)

dfeyer:
Basic support for multi domain nearly done.

Missing feature are:

* [ ] Test Refactoring
* [ ] Listen to domain add/update/remove in SiteController to cleanup redirect

Currently if you move one domain from site A to site B, redirect stay in site A, in the final version redirection in site A matching the domain should be deleted when the domain is attached to site B.

Some cleanup in the command controller. 

We should add a export / import for the redirect, but maybe not in this initial version (export/import of the raw table should work)

gerhard-boden:
hi @dfeyer 
again ran some tests today.
After adding a route with hostURI my test instance broke with the exception `Call to a member function getHostPattern() on string`. Can't really test any further because of this now, but this is what my http chain settings look like, since the redirects did not fire yesterday:
```
preprocess:
    position: 'before process'
    chain: {  }
process:
    chain:
        routing:
            position: start
            component: TYPO3\Flow\Mvc\Routing\RoutingComponent
        dispatching:
            component: TYPO3\Flow\Mvc\DispatchComponent
        ajaxWidget:
            position: 'before routing'
            component: TYPO3\Fluid\Core\Widget\AjaxWidgetComponent
        redirection:
            position: 'before routing'
            component: Neos\RedirectHandler\RedirectionComponent
postprocess:
    chain:
        standardsCompliance:
            position: end
            component: TYPO3\Flow\Http\Component\StandardsComplianceComponent
```

When writing the redirects to the database the `sourceuripath` contains a full URL including the protocol. By default this seems to be `http://`. When a site is running with TLS this may cause problems or additional redirects. (very often all http://some.thing requests are redirected to https://some.thing). Can't we just use the same protocol that the incoming request used?

dfeyer:
Hi @gerhard-boden,

Thanks for try to review ;)

In your database you should have something like this (so protocol agnostic):
https://dl.dropboxusercontent.com/s/5wfj6al5x55tykj/2016-01-10%20at%2022.20.png?dl=0

About your exception, sounds pretty weird, your HTTP components chain look good to me. Can you give me the stack trace of this exception (gist) ?

What did you mean by "After adding a route with host pattern my test instance broke", did you edit your Routes.yaml ? Or just changing the uri path segment of a node from the Neos backend ?

gerhard-boden:
So here is the exception I'm getting:
https://rawgit.com/gerhard-boden/1ffad905650e932997d6/raw/a036199f34e625c8b3329b3bbe67757188339110/redirect-error.html

(just tried again with the latest version of this PR)

This is what my DB looks like:
image

By "adding a route" I actually meant "adding a redirect" with `redirection:add`.
Sorry, wrong terminology.

dfeyer:
@gerhard-boden Did you have the Neos PR https://github.com/neos/neos-development-collection/pull/323, look like you have this exception because of the missing PR for Neos.

dfeyer:
Rebase on master and implement Hit Counter for the Redirection Storage

dfeyer:
Implement Loading from a CSV and Saving to a CSV/Standart output, warning you need league/csv composer package, updated composer.json

gerhard-boden:
checked out #323 again, now i instantly get a 404 message for the home page, but with this on top of the page (1st request)
```PHP Notice: Undefined property: stdClass::$config in /var/www/test-instances/test-master/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php on line 115 PHP Notice: Trying to get property of non-object in /var/www/test-instances/test-master/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php on line 115PHP Notice: Undefined property: stdClass::$config in /var/www/test-instances/test-master/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php on line 115 PHP Notice: Trying to get property of non-object in /var/www/test-instances/test-master/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php on line 115```

At first i thought it was a problem with my test installation, but the same thing happened on a fresh installation also. As soon as i check out master again everything is fine again. 

Am i missing something? I also checked out this PR in the `Packages/Framework` at the same time. I also ran `doctrine:migrationgenerate` and `doctrine:migrate` so the DB should be okay too. 

Any ideas? I'm really eager to test the import/export functionality :smiley:

File: Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php [hunk 1]
```diff
@@ -0,0 +1,239 @@
...
+     * MD5 hash of the Source Uri Path
+     *
+     * @var string
+     * @ORM\Column(length=32)
+     * @Flow\Identity
+     */
+    protected $sourceUriPathHash;
+
+    /**
+     * Target URI path to which a redirect should be pointed
```
skurfuerst: relative or not relative?

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
```diff
@@ -0,0 +1,239 @@
...
+    /**
+     * Status code to be send with the redirect header
+     *
+     * @var integer
+     * @Flow\Validate(type="NumberRange", options={ "minimum"=100, "maximum"=599 })
+     */
+    protected $statusCode;
+
+    /**
+     * Host Pattern
```
skurfuerst: could you add a small comment what the host pattern is / is for?

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
```diff
@@ -0,0 +1,231 @@
...
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Mvc\Routing\RouterCachingService;
+use TYPO3\Flow\Persistence\PersistenceManagerInterface;
+
+/**
+ * Database Storage for the Redirections
+ *
+ * @Flow\Scope("singleton")
+ */
+class RedirectionStorage implements RedirectionStorageInterface
```
skurfuerst: should we rename this to "DatabaseRedirectionStorage"? :)

skurfuerst:
@dfeyer I really like this change :) Just added some nitpicks / hints...

All the best,
Sebastian

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
dfeyer: Uri Path, so just the path (relative)

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
dfeyer: I search if this is currently documented somewhere in Neos, but found nothing. Look like the strategy allow something like www.*.domain.com in the Site host pattern, but the validator does not allow "*" in the host pattern :)

If we have a proper hostpattern documentation somewhere I can add a @see annotation

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
dfeyer: The package is name ... DatabaseStorage, did you think it's not enough ? Each storage should have a dedicated package. But I'm fine for renaming too DatabaseRedirectionStorage, but don't found this really useful.

dfeyer:
@skurfuerst I need to write some documentation for this feature too, but first let's have a working test suite ;)

dfeyer:
@gerhard-boden Your last error is pretty strange. 

Try:

    flow doctrine:migrationexecute --version 20130826225205 --direction down
    flow doctrine:migrationexecute --version 20130826225205 --direction up

Should destroy the storage table a create a new one. I did some change in the DB structure, and update the migration (i know it's bad). And as the PR use an HTTP component, sometimes exception are badly handled (too early in the request ... or I don't know why).

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
```diff
@@ -0,0 +1,239 @@
...
+class Redirection
+{
+    /**
+     * @var \DateTime
+     */
+    protected $creationDateTime;
+
+    /**
+     * @var \DateTime
+     * @Gedmo\Timestampable(on="update", field={"sourceUriPath", "targetUriPath", "statusCode", "hostPattern"})
```
bwaidelich: The dependency to `gedmo/doctrine-extensions` is not specified in a composer.json. This is a package with quite a lot of other dependencies. Do we really need that last modification TS?

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
```diff
@@ -0,0 +1,239 @@
...
+     * @param integer $statusCode
+     */
+    public function update($targetUriPath, $statusCode)
+    {
+        $this->setTargetUriPath($targetUriPath);
+        $this->statusCode = $statusCode;
+    }
+
+    /**
+     * @return mixed
```
bwaidelich: integer

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
```diff
@@ -0,0 +1,239 @@
...
+        $this->setTargetUriPath($targetUriPath);
+        $this->statusCode = (integer)$statusCode;
+        $this->hostPattern = $hostPattern ? trim($hostPattern) : null;
+
+        $this->hitCounter = 0;
+    }
+
+    /**
+     * @param string $targetUriPath
+     * @param integer $statusCode
```
bwaidelich: missing @return

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Repository/RedirectionRepository.php
```diff
@@ -0,0 +1,153 @@
...
+
+use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\ORM\Internal\Hydration\IterableResult;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\QueryBuilder;
+use Neos\RedirectHandler\DatabaseStorage\Domain\Model\Redirection;
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Persistence\QueryInterface;
+use TYPO3\Flow\Persistence\Repository;
+use TYPO3\Neos\Domain\Service\DomainMatchingStrategy;
```
bwaidelich: This adds a dependency to the TYPO3.Neos package!

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
bwaidelich: I agree it can probably save some confusion to name the implementations differently, even if they reside in a custom package.. But not a big deal IMO

Neos.RedirectHandler/Classes/Aspect/RedirectionStorageAspect.php
```diff
@@ -0,0 +1,58 @@
...
+use Neos\RedirectHandler\DatabaseStorage\Domain\Model\Redirection;
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Aop\JoinPointInterface;
+use TYPO3\Flow\Log\SystemLoggerInterface;
+use \Neos\RedirectHandler\Redirection as RedirectionDto;
+
+/**
+ * Redirection Storage Aspect
+ *
+ * This aspect is responsible to emit a signal just after a redirection has been added.
```
bwaidelich: Why do we need this aspect?
Couldn't the `RedirectionService` trigger it instead?  Adding an aspect to some interface implementation seems quite dodgy and implicit to me

Neos.RedirectHandler/Classes/Service/SettingsService.php
```diff
@@ -0,0 +1,55 @@
...
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+use TYPO3\Flow\Annotations as Flow;
+
+/**
+ * Settings Service
```
bwaidelich: What is this?

Neos.RedirectHandler/Classes/Storage/RedirectionStorageInterface.php
```diff
@@ -0,0 +1,78 @@
...
+     * Adds a redirection to the repository and updates related redirection instances accordingly
+     *
+     * @param string $sourceUriPath the relative URI path that should trigger a redirect
+     * @param string $targetUriPath the relative URI path the redirect should point to
+     * @param integer $statusCode the status code of the redirect header
+     * @param array $hostPatterns the list of host patterns
+     * @return array<Redirection> the freshly generated redirection instance
+     * @api
+     */
+    public function addRedirection($sourceUriPath, $targetUriPath, $statusCode = null, array $hostPatterns = []);
```
bwaidelich: Why not expect a single `Redirection` DTO parameter?

Neos.RedirectHandler/Configuration/Settings.yaml
```diff
@@ -0,0 +1,17 @@
...
+    'gone': 410
+
+TYPO3:
+  Flow:
+    http:
+      chain:
+        'process':
+          chain:
+            'redirection':
+              position: 'before routing'
```
bwaidelich: Adding the redirection *before* the routing means that it will be triggered almost for every request.
In some (most?) scenarios it would be enough to trigger the redirect only if the existing routes didn't match a request - for those cases the redirection http component could be triggered only if the routing component didn't succeed:
```
$routingMatchResults = $componentContext->getParameter(RoutingComponent::class, 'matchResults');
if ($routingMatchResults !== NULL) {
	return;
}
// ..
```

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use League\Csv\Reader;
+use League\Csv\Writer;
```
bwaidelich: adds dependency to league/csv

bwaidelich:
Hi and thanks again for digging into this, I'd really love to get this to work finally!!

I tried to test this PR in a flow setup, but hit some errors due to missing dependencies:
`TYPO3.Neos`, `gedmo/doctrine-extensions` & `league/csv`.  At least the neos package shouldn't be a requirement for this I think.
The CSV-Export is a nice idea, but it could be done in a separate package (getting rid of the `league/csv` dependency). Besides: Is there a particular reason why you removed the `redirection:list` command when adding support for CSV?

When I fix the dependencies I still get an error when trying the `redirection:add` command:
```Integrity constraint violation: 1048 Column 'lasthitcount' cannot be null```

I left some more inline comments.

Just looking at the code I really like your approach of separating the storage into its own package.
But personally I'd prefer not to add these packages to the base distribution of Flow as it's an *optional* package

**Comment by skurfuerst** [comment]
@bwaidelich I like that this package is part of the "Flow Development Collection", but does not need to be part of "Flow Distribution"...

All the best,
Sebastian

File: Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php (hunk 3 above)
Comment (kitsunet) [inline Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php]: I am against the duplication, Neos\RedirectHandler\DatabaseStorage\RedirectionStorage is already pretty much duplicate everything. Custom "use" statements are the solution if you feel it is ambigous in your code. IMHO the FQCN should make sense, everything else is to be solved in code using it.

File: Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php (hunk 4 above)
Comment (kitsunet) [inline Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php]: IF we do this, then please move this package to the neos side of things, because there we require gedmo anyway. I don't really like the gedmo package and wouldn't want to introduce a dependency on it for Flow.

File: Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Repository/RedirectionRepository.php (hunk 7 above)
Comment (kitsunet) [inline Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Repository/RedirectionRepository.php]: Yep, IMHO must move over to neos collection

File: Neos.RedirectHandler/Classes/Service/SettingsService.php (hunk 9 above)
Comment (kitsunet) [inline Neos.RedirectHandler/Classes/Service/SettingsService.php]: agree

File: Neos.RedirectHandler/Classes/Storage/RedirectionStorageInterface.php (hunk 10 above)
Comment (kitsunet) [inline Neos.RedirectHandler/Classes/Storage/RedirectionStorageInterface.php]: :+1:

File: Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php (hunk 12 above)
Comment (kitsunet) [inline Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php]: Mmm, again something I would rather see in Neos then.

**Comment by kitsunet** [comment]
@skurfuerst but if the dependency is missing that might even result in Reflection errors. IMHO this needs to be either separated from Neos or move to the neos collection.
An independent redirection package that works for flow as well could make a lot of sense though.

**Comment by kitsunet** [comment]
We could also think about moving the DomainMatchingStrategy out of Neos to some better place (like this package).

File: Neos.RedirectHandler.DatabaseStorage/composer.json [hunk 13]
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "neos/redirecthandler-databasestorage",
+    "type": "typo3-flow-package",
+    "description": "A plugin for neos/redirecthandler to store redirection in the database",
+    "license": "GPL-3.0+",
+    "require": {
+        "php": ">=5.5.0",
```
aertmann: why this requirement? it's defined by flow..

Neos.RedirectHandler/composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "neos/redirecthandler",
+    "type": "typo3-flow-package",
+    "description": "Basic API to handle http redirect with Flow Framework",
+    "license": "GPL-3.0+",
+    "require": {
+        "php": ">=5.5.0",
```
aertmann: same here

Neos.RedirectHandler/Classes/Aspect/RedirectionStorageAspect.php
dfeyer: @bwaidelich  The RedirectionService is used to generation the redirection, not for storage management. So use can inject the StorageInterface, and use the Storage directly. This aspect simplify additional storage coding and enforce the call of the signal emitter

Neos.RedirectHandler/Classes/Storage/RedirectionStorageInterface.php
dfeyer: Because the DTO can not contains multiple hostPatterns, so we can use to simplify $sourceUriPath, $targetUriPath, $statusCode, with the DTO but need to keep the $hostPatterns array, did you think it's important ? Because the DTO contains hostPattern ... should be a bit confusing, no ?

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
dfeyer: Keep in mind that this package is in the dev collection, but optional in the Flow distribution. So dependency is not really important here. 

I will try to do my best to have this package with no Neos dependencies, so a generic redirection handling for Flow. But people a free to use it. The Gedmo and League package are quiet common in the PHP community (even if we can "not like" some parts).

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
dfeyer: Also against the duplication

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
dfeyer: fixed

dfeyer:
> We could also think about moving the DomainMatchingStrategy out of Neos to some better place (like this package).

@kitsunet Moving it this this package will create a dependencies between Neos and this package, no go

Need to think a bit more about how to remove this dependency

> I tried to test this PR in a flow setup, but hit some errors due to missing dependencies:
TYPO3.Neos, gedmo/doctrine-extensions & league/csv. At least the neos package shouldn't be a requirement for this I think.

See my comment https://github.com/neos/flow-development-collection/pull/195#discussion_r49901564 so from my POV the dependency to Neos MUST be solved. But the other are fine.

> When I fix the dependencies I still get an error when trying the redirection:add command:
Integrity constraint violation: 1048 Column 'lasthitcount' cannot be null

@bwaidelich Strange this value is set to 0 in the constructor, did you have the last version ?

About where to store those package, what's the best / easy way currently ? Move every thing to the Neos Collection ? Should be easy (lost history but it's not an issue for this PR).

gerhard-boden:
@dfeyer the PRs are working now on my test instance :) 
Will do some intense testing over the weekend and report all issues i run into

gerhard-boden:
Here is the bugreport from my first testing:
* Updating a row in the CSV does not update the entry in the database when reimported
* Allow only 3xx and 418 status codes, everything else makes no sense in my opinion
* Duplicate `sourceuripath` entries break the CSV import
* The redirect chain prevention is not working as expected when importing additional redirects via CSV (the redirect is not imported at all then)

Really like this plugin, thanks for all your work so far!

dfeyer:
> Duplicate sourceuripath entries break the CSV import

Need to cleanup the CSV and output warning if duplicate are found.

@gerhard-boden Thanks for the review.

I will push a new version this week-end to remove the Neos coupling and fix some of the issues

bwaidelich:
Sorry for being stubborn about it, but I feel that this introduces a lot of dependencies.
Are creation timestamps and CSV exports really a must-have feature for this and couldn't they be added through a separate (optional) package?
Also it might be possible to add a simple "hook" mechanism that allows to intercept the URI matching (i.e. the Neos-redirect package could inject its DomainMatchingStrategy there)..

I really don't want to cause additional work and I'm happy to help if possible. I'll need a redirect feature for a Flow application very soon so I'll create a very simple external package for that for now, but maybe we can merge ideas if it works out

dfeyer:
> Sorry for being stubborn about it, but I feel that this introduces a lot of dependencies.

Gedmo is pretty common package, used by Neos per ex. Maybe we can remove the League CSV dependencies, but we need parts of this logic ... so it's dependency vs more code. 

About the dependencies, in the final version Neos will not be a dependency for the Flow package. My current personal situation make it impossible to work on the PR, before next week.

I'm fine to extract the import / export stuff to an external package ... but look like overly complex. I think people will need import / export stuffgoo

dfeyer:
Last version decouple the Flow package and the default storage from Neos, please update your DB:

    flow doctrine:migrationexecute --version 20130826225205 --direction down
    flow doctrine:migrationexecute --version 20130826225205 --direction up

Now the DomainMatching need to me moved to the redirect component. So we can include a new Redirect component for the the Neos package, that support domain matching based on host pattern.

@bwaidelich @kitsunet Your input is highly welcome on the decoupling.

Now open topics:

- [ ] Test suite, maybe some behat stuff
- [ ] Did we need to decouple the Gedmo (last usage timestamp for each redirect)
- [ ] Did we need to decouple the Import/Export stuff 

I miss something ping me and I update this list. During the last few days, I'm mostly offline, feel free to work on some of those point if you have time, just add a comment here, to avoid double work.

Thanks all for the reviews

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
kdambekalns: That pattern is not a pattern… Since 2013, when we added that HostnameValidator there. See this commit in my pipeline: https://github.com/kdambekalns/neos-development-collection/commit/79465245aab2561630febcf7f72e37c558f3f9ed

albe:
I'd also prefer to avoid the Gedmo dependency if it is only needed for the Timestampable feature, which can easily be implemented otherwise (explicitly change the modificationTimestamp in the setters, use an doctrine event listener directly or even AOP).

Regarding the CSV I'm not sure. I also don't see that as a must-have feature for this package, but I also see the use for it, that would make creating a seperate package just for that feel awkward. Maybe this can be a soft-dependency (composer recommended)? As in if the csv library is installed/available, the import/export  feature is working, otherwise it will just error out, suggesting to do a simple "composer require leage/csv".

dfeyer:
@albe Thanks for the review, I will push a new version during the week-end. Nice idea to have the "soft dependency for the CSV package.

bwaidelich:
Just one more thought regarding the CSV feature:
Making it a soft-dependency is a good idea, but might be difficult because you can't use the classes directly then. An alternative solution could be a slim 3rd package that adds the `save` and `load` commands and has that dependency.

Or we might be able to create a little `Cli2Csv`-Package that captures the `ConsoleOutput` and transform `outputTable`-calls to a CSV output.. It could be used like this:
```
./flow csv:export redirection:list other parameters
```

albe:
> because you can't use the classes directly then.

A simple class_exists() check should do, no?

kitsunet:
Yes exactly, as long as we don't extend from non existing classes (or implement non existing interfaces) all should be finde.

dfeyer:
- [x] league/csv is not optional
- [x] remove dependency to the gedmo package
- [ ] Add an HTTP component in the Neos package to support host pattern
- [ ] Test suite

dfeyer:
Add basic test suite for Neos.RedirectHandler.DatabaseStorage and fix MySQL migration, please run the following command to avoid some error with ```lastHitCount``` that can not be null. 

    flow doctrine:migrationexecute --version 20130826225205 --direction down
    flow doctrine:migrationexecute --version 20130826225205 --direction up

dfeyer:
Currently a bit broken after the master merge in this PR, will try to solve the issue tomorrow

dfeyer:
Fix most unit test, based on the test suite created by @bwaidelich with some addition to test new feature. 

Just one unit test is broken:

    1) TYPO3\Flow\Tests\Unit\Http\Component\ComponentChainTest::handleResetsTheCancelParameterIfItW  asTrue
    Expectation failed for method name is equal to <string:getParameter> when invoked at sequence index 0.
    Mocked method does not exist.

Not sure why this test is broken, so don't want to fix the test without an external review. @bwaidelich maybe ? It's maybe related to this commit: https://github.com/dfeyer/flow-development-collection/commit/d764289482be450aa21446e7914dc898e317029a

I switch the default status code to 307.

Will fix the functional test tomorrow, and add support for host pattern in the Neos redirect component. 

Feel free to test / review, comment, ...

albe:
Nice, really like your work on that :) will check it out again when I find time

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
dfeyer: @kdambekalns Fine one less thing to support ;) did you think it can be merged asap ? Like this I can test the PR on top of this change

kdambekalns: No, that is my "next-major" branch, Flow 4.0 so to speak. And that particular change is breaking, because it changes the arguments of some commands, people using that in automation would need to adjust their tooling.

Anyway, what I wanted to point out: pattern is not supported, so no problem :)

kitsunet:
Just had a read through the code, looks pretty good to me so far!

dfeyer:
Fix a small conflict in the composer.json caused by the splits branch merging.

dfeyer:
The tests are currently broken, but the feature is ready for testing / review.

I will work on the tests this week.

gerhard-boden:
did some more testing today, thanks for all the work so far!

- [ ] Automatic redirects to not consider alternative URLs
2.1 makes the whole story a little more complex since /en/download.html and /download.html are allowed now.
I'm not sure how to handle this, but maybe the user must be aware of this and add the correct redirects for both urls.
The automatic redirects should probably add redirects for both versions. 
If there is any way to disable this new behaviour via configuration, it should be recommended to disable it when using redirects, since this makes redirect management less confusing.
- [ ]  While you get an error when importing a redirect with the `add` command if the target already exists as a source uri, the redirect is ignored without notice when using csv import.
- [ ]  The target uri for an existing souce uri is not updated when reimported via CSV

dfeyer:
@gerhard-boden Thanks for the review, I will try to take care of this next week.

> Automatic redirects do not consider alternative URLs

This one is a bit of a HELL ;) But I think it's not a Flow issue, but a Neos one, so this need to be solved in the Neos adapter, not in the Flow package.

kitsunet:
Yep definitely ignore that problem for now. I am currently thinking into the neos/node routing a lot and intent to change it.

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Repository/RedirectionRepository.php
```diff
@@ -0,0 +1,164 @@
...
+    {
+        $iteration = 0;
+        foreach ($iterator as $object) {
+            /** @var Redirection $object */
+            $object = current($object);
+            yield $object;
+            if ($callback !== null) {
+                call_user_func($callback, $iteration, $object);
+            }
+            ++$iteration;
```
aertmann: while there are occurrences of `++$var`, the more common is `$var++;`

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
```diff
@@ -0,0 +1,240 @@
...
+    {
+        foreach ($this->redirectionRepository->findAll($host) as $redirection) {
+            yield new RedirectionDto($redirection->getSourceUriPath(), $redirection->getTargetUriPath(), $redirection->getStatusCode(), $redirection->getHost());
+        }
+    }
+
+    /**
+     * Return a list of all host patterns
+     *
+     * @return array
```
aertmann: @api?

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
```diff
@@ -0,0 +1,240 @@
...
+     * @var array
+     */
+    protected $defaultStatusCode;
+
+    /**
+     * Returns one redirection for the given $sourceUriPath or NULL if it doesn't exist
+     *
+     * @param string $sourceUriPath
+     * @param string $host Full qualified hostname or host pattern
+     * @return RedirectionDto or NULL if no redirection exists for the given $sourceUriPath
```
aertmann: RedirectionDto|null

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Model/Redirection.php
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * @var integer
+     */
+    protected $hitCounter;
+
+    /**
+     * @var \DateTime
+     * @ORM\Column(nullable=true)
+     */
+    protected $lastHitCount;
```
aertmann: this naming is confusing, it's not a count but a timestamp

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,165 @@
...
+            $writer->output();
+        } else {
+            file_put_contents($filename, (string)$writer);
+        }
+    }
+
+    /**
+     * Load redirection from a CSV file
+     *
+     * @param string $filename CSV file path
```
aertmann: @return void

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,165 @@
...
+     * @Flow\Inject
+     * @var PersistenceManagerInterface
+     */
+    protected $persistenceManager;
+
+    /**
+     * Save all redirectection to a CSV file
+     *
+     * @param string $filename CSV file path, if null use standard output
+     * @param string $host Optional host pattern
```
aertmann: @return void

Neos.RedirectHandler/Classes/Aspect/RedirectionStorageAspect.php
aertmann: kinda agree with @bwaidelich here and see your reasoning, however an abstract class with that signal implementing the interface would be a better approach IMO

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
```diff
@@ -0,0 +1,240 @@
...
+                $this->redirectionRepository->update($obsoleteRedirection);
+            }
+        }
+    }
+
+    /**
+     * Increment the hit counter for the given redirection
+     *
+     * @param RedirectionDto $redirection
+     * @return mixed
```
aertmann: @api?

aertmann:
Hey @dfeyer very nice work!

a few additional questions/suggestions:
1. not sure about the naming, reads in an odd way.. also there's a mixture of Redirection and Redirects.. would suggest to go with a Redirect model etc.
2. Postgres migration missing
3. replace `array(` syntax with `[`
4. ensure `@api` is marked where it should be

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
```diff
@@ -0,0 +1,240 @@
...
+        $this->redirectionRepository->removeAll($host);
+    }
+
+    /**
+     * Adds a redirection to the repository and updates related redirection instances accordingly
+     *
+     * @param string $sourceUriPath the relative URI path that should trigger a redirect
+     * @param string $targetUriPath the relative URI path the redirect should point to
+     * @param integer $statusCode the status code of the redirect header
+     * @param array $hosts the list of host patterns
```
aertmann: wait how does this work? never seen that before.. maybe provide an example in the description

radmiraal:
Nice to see this change, would love to discuss how we can use this to deprecate https://github.com/SimplyAdmire/SimplyAdmire.Redirects ;-)

That package btw also adds help for converting the redirects to .htaccess or nginx redirects, which could be moved to an export package (like with the csv) too.

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectionStorage.php
dfeyer: @aertmann I'm not sure if I understand the question ?

I need to rename host patterns everywhere (because even in Neos we don't support host patterns since sometimes now). So if the question is about host pattern, read "hostname" ;)

dfeyer:
Rebased on master, will tackle the next steps during the weekend.

dfeyer:
Next steps:

* [ ] Documentation
* [ ] Fix naming Redirection / Redirect
* [ ] Remove the RedirectionStorageAspect
* [ ] Improve import notifications and fix error reported by @gerhard-boden 

@aertmann @kitsunet @gerhard-boden if you think other point need to tackled feel free to add them here ;)

dfeyer:
@aertmann I try to make the naming a bit more consistent, your feedback is welcome

I also remove the aspect by a trait

Now try to fix the tests ;)

dfeyer:
My collegue just ask me "What happen when I move a page ?" ... my first answer was "Not tested ... but works fine ... awesome what we work on a good API"

gerhard-boden:
Hehe, moved pages while testing and successfully checked for redirects.
Didn't know that feature was in there by accident ;)

Dominique Feyer <notifications@github.com> schrieb am Fr., 8. Apr. 2016,
18:51:

gerhard-boden:
Can't look at the code at the moment, but I guess it works because the redirect is created on publish (not just when applying a change to the Uri path segment in the Inspector)

dfeyer:
Tests are running again, but look like I break something in the ComponentChainTest, if somebody can found why, i will be happy ;)

    There was 1 failure:
    1) TYPO3\Flow\Tests\Unit\Http\Component\ComponentChainTest::handleResetsTheCancelParameterIfItWasTrue
    Expectation failed for method name is equal to <string:getParameter> when invoked at sequence index 0.
    Mocked method does not exist.

dfeyer:
@gerhard-boden 9da9267 should solve most of your issue with importing from a CSV, not it replace existing redirection and should update every correctly.

dfeyer:
I love working with CLI: https://dl.dropboxusercontent.com/s/itmzftdzoysrxhm/2016-04-08%20at%2023.15.png?dl=0

So for me this one is ready for hardcore review, I remove the WIP tag, last missing pieces:

- [ ] Broken test regarding ComponentChainTest (see bellow, I need help on this one)
- [ ] Documentation (the work has been started by @gerhard-boden here https://github.com/dfeyer/flow-development-collection/pull/1)

Have a good review session, this feature is really nice ;)

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -42,6 +41,63 @@ class RedirectionCommandController extends CommandController
...
+            }
+        };
+        if ($host !== null) {
+            $outputByHost($host);
+        } else {
+            $hosts = $this->redirectStorage->getDistinctHosts();
+            if ($hosts !== []) {
+                array_map($outputByHost, $hosts);
+            }
+            $outputByHost($host);
```
aertmann: `$outputByHost(null):`?

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,336 @@
...
```
aertmann: `$outputByHost(null):`?

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -210,15 +210,49 @@ public function removeAllCommand($host = null)
...
+                $redirect->getSourceUriPath(),
+                $redirect->getTargetUriPath(),
+                $redirect->getStatusCode(),
+                $redirect->getHost() ?: 'no host'
+            ]);
+        };
+        $this->outputLine();
+        $redirect = $this->redirectStorage->getOneBySourceUriPathAndHost($source, $host);
+        if ($redirect !== null && $force === false) {
+            $this->outputLine('A redirection with the same source URI exist, see bellow:');
```
aertmann: below

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,336 @@
...
+     */
+    public function addCommand($source, $target, $statusCode, $host = null, $force = false)
+    {
+        $this->outputLine();
+        $this->outputLine('<b>Create a redirection ...</b>');
+        $this->outputLine();
+        $redirect = $this->redirectStorage->getOneBySourceUriPathAndHost($source, $host);
+        $isSame = $this->isSame($source, $target, $host, $statusCode, $redirect);
+        if ($redirect !== null && $isSame === false && $force === false) {
+            $this->outputLine('A redirection with the same source URI exist, see bellow:');
```
aertmann: below

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,336 @@
...
+        $redirects = $this->redirectStorage->addRedirection($source, $target, $statusCode, [$host]);
+        $redirect = reset($redirects);
+        $this->outputRedirectLine('<info>++</info>', $redirect);
+        $this->outputLine();
+        $this->outputLegend();
+    }
+
+    /**
+     * @param string $prefix
+     * @param RedirectInterface $redirect
```
aertmann: @return void

Neos.RedirectHandler.DatabaseStorage/composer.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "neos/redirecthandler-databasestorage",
+    "type": "neos-package",
+    "description": "A plugin for neos/redirecthandler to store redirection in the database",
+    "license": "GPL-3.0+",
+    "require": {
+        "typo3/flow": "~3.2",
```
aertmann: this dependency would be inherited, so I'd suggest to leave it out

Neos.RedirectHandler/composer.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "neos/redirecthandler",
+    "type": "neos-package",
+    "description": "Basic API to handle http redirect with Flow Framework",
+    "license": "GPL-3.0+",
+    "require": {
+        "typo3/flow": "~3.2"
```
aertmann: not sure if this would have unwanted side effects like not being able to use Neos master with Flow master

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Repository/RedirectRepository.php
```diff
@@ -0,0 +1,164 @@
...
+use Doctrine\ORM\Internal\Hydration\IterableResult;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\QueryBuilder;
+use Neos\RedirectHandler\DatabaseStorage\Domain\Model\Redirect;
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Persistence\QueryInterface;
+use TYPO3\Flow\Persistence\Repository;
+
+/**
+ * Repository for redirection instances.
```
aertmann: redirect instances?

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+use Neos\RedirectHandler\Redirect as RedirectDto;
+use Neos\RedirectHandler\RedirectInterface;
+use Neos\RedirectHandler\Storage\RedirectStorageInterface;
+use Neos\RedirectHandler\Traits\RedirectSignalTrait;
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Mvc\Routing\RouterCachingService;
+use TYPO3\Flow\Persistence\PersistenceManagerInterface;
+
+/**
+ * Database Storage for the Redirections
```
aertmann: Redirects

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+    protected $runtimeCache = [];
+
+    /**
+     * @Flow\InjectConfiguration(path="statusCode", package="Neos.RedirectHandler")
+     * @var array
+     */
+    protected $defaultStatusCode;
+
+    /**
+     * Returns one redirection for the given $sourceUriPath or NULL if it doesn't exist
```
aertmann: redirect?

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+    protected $persistenceManager;
+
+    /**
+     * @Flow\Inject
+     * @var RouterCachingService
+     */
+    protected $routerCachingService;
+
+    /**
+     * Runtime cache to avoid creating multiple time the same redirection
```
aertmann: creating the same redirect multiple times?

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+     * @var array
+     */
+    protected $defaultStatusCode;
+
+    /**
+     * Returns one redirection for the given $sourceUriPath or NULL if it doesn't exist
+     *
+     * @param string $sourceUriPath
+     * @param string $host Full qualified hostname or host pattern
+     * @return RedirectDto|null if no redirection exists for the given $sourceUriPath
```
aertmann: redirects

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+    {
+        $redirect = $this->redirectRepository->findOneBySourceUriPathAndHost($sourceUriPath, $host);
+        if ($redirect === null) {
+            return null;
+        }
+        return RedirectDto::create($redirect);
+    }
+
+    /**
+     * Returns all registered redirection records
```
aertmann: redirect

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+     * @return array
+     * @api
+     */
+    public function getDistinctHosts()
+    {
+        return $this->redirectRepository->findDistinctHosts();
+    }
+
+    /**
+     * Removes a redirection for the given $sourceUriPath if it exists
```
aertmann: redirect

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+    {
+        $redirect = $this->redirectRepository->findOneBySourceUriPathAndHost($sourceUriPath, $host);
+        if ($redirect === null) {
+            return;
+        }
+        $this->redirectRepository->remove($redirect);
+    }
+
+    /**
+     * Removes all registered redirection records
```
aertmann: redirect

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+                $this->redirectRepository->remove($obsoleteRedirect);
+            } else {
+                $obsoleteRedirect->setTargetUriPath($newRedirect->getTargetUriPath());
+                $this->redirectRepository->update($obsoleteRedirect);
+            }
+        }
+    }
+
+    /**
+     * Increment the hit counter for the given redirection
```
aertmann: redirect

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+        $redirect = new Redirect($sourceUriPath, $targetUriPath, $statusCode, $host);
+        $this->updateDependingRedirects($redirect);
+        $this->redirectRepository->add($redirect);
+        $this->routerCachingService->flushCachesForUriPath($sourceUriPath);
+        $this->runtimeCache[$hash] = $redirect;
+        return RedirectDto::create($redirect);
+    }
+
+    /**
+     * Updates affected redirection instances in order to avoid redundant or circular redirects
```
aertmann: redirect

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+     * Adds a redirection to the repository and updates related redirection instances accordingly
+     *
+     * @param string $sourceUriPath the relative URI path that should trigger a redirect
+     * @param string $targetUriPath the relative URI path the redirect should point to
+     * @param integer $statusCode the status code of the redirect header
+     * @param string $host the host patterns for the current redirection
+     * @return Redirect the freshly generated redirection instance
+     * @api
+     */
+    protected function addRedirectionByHost($sourceUriPath, $targetUriPath, $statusCode, $host = null)
```
aertmann: redirect

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /**
+     * Adds a redirection to the repository and updates related redirection instances accordingly
+     *
+     * @param string $sourceUriPath the relative URI path that should trigger a redirect
+     * @param string $targetUriPath the relative URI path the redirect should point to
+     * @param integer $statusCode the status code of the redirect header
+     * @param string $host the host patterns for the current redirection
+     * @return Redirect the freshly generated redirection instance
```
aertmann: redirect

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+        return $redirects;
+    }
+
+    /**
+     * Adds a redirection to the repository and updates related redirection instances accordingly
+     *
+     * @param string $sourceUriPath the relative URI path that should trigger a redirect
+     * @param string $targetUriPath the relative URI path the redirect should point to
+     * @param integer $statusCode the status code of the redirect header
+     * @param string $host the host patterns for the current redirection
```
aertmann: redirect

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+            }, $hosts);
+        } else {
+            $redirects[] = $this->addRedirectionByHost($sourceUriPath, $targetUriPath, $statusCode);
+        }
+        $this->emitRedirectionCreated($redirects);
+        return $redirects;
+    }
+
+    /**
+     * Adds a redirection to the repository and updates related redirection instances accordingly
```
aertmann: redirect x 2

Neos.RedirectHandler.DatabaseStorage/Readme.rst
```diff
@@ -0,0 +1,18 @@
+---------------------------------------------
+Neos RedirectHandler Database Storage Package
+---------------------------------------------
+
+.. note:: This repository is a **read-only subsplit** of a package that is part of the
+          Neos project (learn more on `www.neos.io <https://www.neos.io/>`_).
+
+A plugin for neos/redirecthandler to store redirection in the database
```
aertmann: redirects

Neos.RedirectHandler.DatabaseStorage/Tests/Unit/RedirectStorageTest.php
```diff
@@ -0,0 +1,195 @@
...
+use Neos\RedirectHandler\DatabaseStorage\Domain\Repository\RedirectRepository;
+use Neos\RedirectHandler\DatabaseStorage\RedirectStorage;
+use Neos\RedirectHandler\Redirect as RedirectDto;
+use Neos\RedirectHandler\RedirectionService;
+use TYPO3\Flow\Log\SystemLoggerInterface;
+use TYPO3\Flow\Mvc\Routing\RouterCachingService;
+use TYPO3\Flow\Tests\UnitTestCase;
+
+/**
+ * Test case for the RedirectionService class
```
aertmann: RedirectStorage class

Neos.RedirectHandler.DatabaseStorage/Tests/Unit/RedirectTest.php
```diff
@@ -0,0 +1,201 @@
...
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\RedirectHandler\DatabaseStorage\Domain\Model\Redirect;
+use TYPO3\Flow\Tests\UnitTestCase;
+use TYPO3\Flow\Utility\Now;
+
+/**
+ * Test case for the RedirectionService class
```
aertmann: Redirect class

Neos.RedirectHandler.DatabaseStorage/composer.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "neos/redirecthandler-databasestorage",
+    "type": "neos-package",
+    "description": "A plugin for neos/redirecthandler to store redirection in the database",
```
aertmann: redirects

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,336 @@
...
+    protected $redirectStorage;
+
+    /**
+     * @Flow\Inject
+     * @var PersistenceManagerInterface
+     */
+    protected $persistenceManager;
+
+    /**
+     * @param string $host Filter redirection by the given hostname
```
aertmann: redirects

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,336 @@
...
+        }
+        if ($filename === null) {
+            $writer->output();
+        } else {
+            file_put_contents($filename, (string)$writer);
+        }
+    }
+
+    /**
+     * Load redirection from a CSV file
```
aertmann: redirects

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,336 @@
...
+        $this->outputLine();
+        if (!class_exists(Reader::class)) {
+            $this->outputWarningForLeagueCsvPackage();
+        }
+        if (!is_readable($filename)) {
+            $this->outputLine('<error>Sorry, but the file "%s" is not readable or does not exist...</error>', [$filename]);
+            $this->outputLine();
+            $this->sendAndExit(1);
+        }
+        $this->outputLine('<b>Import redirection from "%s"</b>', [$filename]);
```
aertmann: redirects

aertmann:
Hey @dfeyer, thanks for pushing this further. Certainly a lot of work.

Started commenting on remaining usages of "Redirection", but was a bit too many for it to make sense to continue.

Commented on some other nitpicks. Otherwise it looks good in general.

One blocker is the missing Postgres migration though.

Also are any of these points still valid? https://github.com/neos/flow-development-collection/pull/195#issuecomment-172252371

dfeyer:
> Allow only 3xx and 418 status codes, everything else makes no sense in my opinion

Will not be strictly enforced, let the user be serious ;) a redirection with 200 does not make sense ;)

The other issue from https://github.com/neos/flow-development-collection/pull/195#issuecomment-172252371 should be solve by persisting every redirect in the for each loop.

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
dfeyer: Should be replaced by ```$outputByHost();```, will push this later today

But a redirect can be attached to a host or not and outputByHost handle correctly null value.

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
dfeyer: Yes per example during a batch import, a CSV can contains duplicated redirect (with the same URI), this cache avoid key contraint errors

dfeyer:
> One blocker is the missing Postgres migration though.

I will try to finally setup a PostreSQL setup, it's a shame i don't do that earlier.

@aertmann Did you have any idea about the failing test for the ComponentChainTest ?

I will push a follow up this evening with more fix redirection/redirect in comments (look like I forgot to check the comments.

Thanks for the review.

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
radmiraal: Aren't you swallowing expected exceptions then? Only writing the last instance for a certain redirect would not be up to the base storage but more an importer responsibility right?

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
```diff
@@ -0,0 +1,247 @@
...
+
+    /**
+     * Increment the hit counter for the given redirection
+     *
+     * @param RedirectInterface $redirect
+     * @return void
+     * @api
+     */
+    public function incrementHitCount(RedirectInterface $redirect)
+    {
```
radmiraal: did you consider using DQL for the update query?

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
radmiraal: "So dependency is not really important here." I can't really agree here, no matter where: dependencies are always important... In a redirect package I created last year (https://github.com/SimplyAdmire/SimplyAdmire.Redirects) we also created an import and even a writer (to write to for example an nginx config file or output for a .htaccess). I do feel like import / export / writing is a bit much for the base package though.

The one time we really used the import it was for a onetime import during a migration, and I think that's the most common usecase. That would actually be an argument for me to move such commands to a separate package (also removes the dependencies and makes the base package more slim)

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,336 @@
...
+                $this->persistenceManager->persistAll();
+            }
+            try {
+                $redirects = $this->redirectStorage->addRedirection($sourceUriPath, $targetUriPath, $statusCode, $hosts);
+                foreach ($redirects as $redirect) {
+                    $this->outputRedirectLine('<info>++</info>', $redirect);
+                }
+                $this->persistenceManager->persistAll();
+            } catch (Exception $exception) {
+                \TYPO3\Flow\var_dump($exception);
```
radmiraal: var_dump ;)

Neos.RedirectHandler/Classes/RedirectionService.php
```diff
@@ -0,0 +1,106 @@
...
+use TYPO3\Flow\Mvc\Routing\RouterCachingService;
+
+/**
+ * Central authority for HTTP redirects.
+ * This service is used to redirect to any configured target URI *before* the Routing Framework kicks in and it
+ * should be used to create new redirection instances.
+ *
+ * @Flow\Scope("singleton")
+ */
+class RedirectionService
```
radmiraal: RedirectService

gerhard-boden:
Should I do some final testing on this or should I wait until the latest changes are commited?

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
dfeyer: Let's keep some improvements for an other PR ;) but look like a good idea

dfeyer:
@gerhard-boden Feel free to start a final review

@aertmann @radmiraal @kitsunet @robertlemke You are also welcome for the final review ;)

Neos.RedirectHandler.DatabaseStorage/Classes/RedirectStorage.php
dfeyer: Now we use DQL to increment the counter at the DB level.

gerhard-boden:
YES! I was hoping for this to happen, because I have some time this weekend for testing. Will report back with the results. Thanks for the updates!

kitsunet:
Looks pretty cool, :+1 by reading, will test as well...

gerhard-boden:
Thanks a lot again for the update! 
Great work, I could only find two little bugs, the rest is just a list of suggestions 

**Bugs**
- [ ] When I add a redirect for a specific host, it also listed as a general redirect (it will work correctly though, the `:list` command is just showing it wrong).
- [ ] The source URI check is broken when working with multiple hosts. 
*Example:* I add a general redirect for A -> B. Now i add a redirect for my host "test.alias" with "A -> C".
At first I received a message that the redirect already exists with the same source (which is correct to some extend, since redirects without hosts are valid for all alias domains)
But when I add "--force" the general redirect is removed and the host specific redirect is added.
Shouldn't it be possible to "overrule" general redirects for host specific redirects, without removing the general redirect?
The user can circumvent this bug by always supplying a `host` argument.

**Suggestions**
None of this is merge critical, except the fist thing (maybe), but I guess this is an easy fix.
So if you don't have time to care of them at the moment them so be it.

- [ ] 307 as the standard redirect status code might be harmful for SEO (Google could refuse to index the target, similar like the 302 issues with google)
   I know this can be customized via `Settings.yaml` but 307 is also the default for automatic redirects in Neos and some users might not be aware of that.
- [ ] `redirection:export` (and `import`) could use the `Data` folder as default root. Now it's dropped into the Vhost root. 
- [ ] When the import finished with errors, show errors at the very end of importlog or write something like `!! Import contains errors` at the very end. 
    If you import 10000 redirects at a time, you would have to search through the log for errors. This way you can tell instantly if something went wrong.
- [ ] When I have a redirect for a specific host and want to remove it without adding a host argument I get `"There is no redirect with the source URI path "test12"`. 
  Maybe we could add a hint to this message for the user to add a host argument.


I also noticed that the target for existing redirects is updated when adding a redirect using said target as source. This is awesome!
The match command is also a very useful addition (really like how you structured and styled the CLI output in general!)

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,342 @@
...
+     */
+    public function listCommand($host = null, $match = null)
+    {
+        $outputByHost = function ($host = null) use ($match) {
+            $redirects = $this->redirectStorage->getAll($host);
+            $this->outputLine();
+            if ($host !== null) {
+                $this->outputLine('<info>==</info> <b>Redirect for %s</b>', [$host]);
+            } else {
+                $this->outputLine('<info>==</info> <b>Redirect without host attached</b>', [$host]);
```
gerhard-boden: Maybe say "Redirects valid for all hosts" (underwise this could be misunterstood)

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,342 @@
...
+            }
+            if ($match !== null) {
+                $this->outputLine('   <info>++</info> <b>Show only if source or target URI match <u>%s</u></b>', [$match]);
```
gerhard-boden: `Only showing redirects where source or target URI matches <u>%s</u></b>`

Neos.RedirectHandler/Classes/Command/RedirectionCommandController.php
```diff
@@ -0,0 +1,342 @@
...
+    /**
+     * @return void
+     */
+    protected function outputLegend()
+    {
+        $this->outputLine('<b>Legend</b>');
+        $this->outputLine();
+        $this->outputLine('   <info>++</info> Redirect created');
+        $this->outputLine('   <info>--</info> Redirect removed');
+        $this->outputLine('   <comment>~~</comment> Redirect do not need update');
```
gerhard-boden: `Redirect not modified`

dfeyer:
@gerhard-boden Thanks for the review, it's really a pleasure to work with you ;)

I solve you two bugs, the last one solved by 70f07a1 is really nasty, and nice to we catch it now.

I also improve the remove commands, now we have `redirection:removeall` and `redirection:removebyhost`. It's important because with host === null we should only remove redirect with property host set as null and not all redirect. This is a bit more clear now.

We should also communicated clearly that we fallback to redirect with host property setted as null, if no more specific redirect is found (understand with host value that match the current host).

gerhard-boden:
Hey @dfeyer,
you already took care of the list? Awesome! The pleasure is all mine by the way. You're the one doing the awesome coding, and I'm the guy nagging and making lists :smile: 

I retested the two bugs and the new command. Everything is working perfectly now!

One tiny nitpick for the `removebyhost` command since it feels redundant with with the `removeall` command (it does the same when no `host` argument is supplied).
My suggestion would be to either add the `host` option to `removeall` or make `host` a required argument for `removebyhost`. While testing I expected the command to ask me for the host, but it just went ahead an deleted all my fallback redirects. I thought "whoops" and it felt a little like I just ran the infamous `rm ....` command :smiley:

radmiraal:
hey @dfeyer, seems like you're on a 'weekend write redirect code'-spree ;-) nice work

Without taking the code for a spin / in depth review of the commits you added:
* I still see 'redirection' used a lot where 'redirect' would be the correct word. Kind of a nitpick, but better fix it pre-merge than after-merge ;-)
* I still have doubts about the export / import stuff. I do see the usecase, but the fact it adds dependencies and is probably not used in most installations makes me feel like we should better substract it. Even when this package is not really a 'core package' or so. There are nitpickers like me that would prefer maintaining a custom package over including one with extra dependencies for features you don't need in a project :-)

kdambekalns:
As I see it, this adds two new packages (`Neos.RedirectHandler` and `Neos.RedirectHandler.DatabaseStorage`), right? So when this is merged we need to

- [ ] add two new repositories on GitHub
- [ ] adjust the Slicer job on Jenkins

And a question on DB migration: Why is the MySQL migration dated 2013? It should be renamed to a current date to avoid confusion.

kdambekalns:
Ah, and the license for Flow is MIT, not GPL.

kdambekalns:
The MySQL migration uses `VARCHAR(4000)`, something we have not used until now. It works in MySQL 5.0.3 and later, so I guess it's not a problem. Still, since the matching is done on the hash (I assume), there is nothing speaking against changing this to `TEXT` to stay closer to standard SQL (for what it's worth). Opinions?

kdambekalns:
@dfeyer see https://github.com/dfeyer/flow-development-collection/pull/2

dfeyer:
> I still have doubts about the export / import stuff. I do see the usecase, but the fact it adds dependencies and is probably not used in most installations makes me feel like we should better substract it. Even when this package is not really a 'core package' or so. There are nitpickers like me that would prefer maintaining a custom package over including one with extra dependencies for features you don't need in a project :-)

@radmiraal No hard dependency, you need league/csv only if you want to use the import/export stuff, if you don't care, you don't have a new dependency, just a suggestion

> I still see 'redirection' used a lot where 'redirect' would be the correct word. Kind of a nitpick, but better fix it pre-merge than after-merge ;-)I still see 'redirection' used a lot where 'redirect' would be the correct word. Kind of a nitpick, but better fix it pre-merge than after-merge ;-)

@radmiraal  I know ... not easy to rename this in the middle of this PR. Feel free to send a PR to this one. Not sure if I have time today or tomorrow to tackle this.

> One tiny nitpick for the removebyhost command since it feels redundant with with the removeall command (it does the same when no host argument is supplied).

@gerhard-boden  As host === null is valid value for us ;) I can not remove everything if the host is null, i need to remove only the redirect with null as host value. So removeall is our rm -Rf and removebyhost, remove only per host (or null if you target fallback redirect).

radmiraal:
@dfeyer whoops, missed the suggest for league/csv :) Awesome... Will fix the redirection PR to make up for misreading ;)

gerhard-boden:
@dfeyer Thanks for explaining, I get it now from the technical perspective! Will have to explain the difference between the two commands in the documentation. Maybe we can use a different value for the `removebyhost` command. My suggestion: Make `host` mandatory. You can either enter a host name or enter `all` to delete the redirects that are valid for all domains. I opened an PR for that. This is not important at all, but i think it would make the command safer if you just enter `redirection:removebyhost`. https://github.com/dfeyer/flow-development-collection/pull/4

dfeyer:
Will be offline tomorrow, and in the train Thursday morning (EAT, Rosenheim, Thursday, 14:30), feel free to push a PR on my repo if anything need to fixed to have the PR in a mergeable state ASAP ;)

gerhard-boden:
Improved the behaviour of the `redirect:export` command:
https://github.com/dfeyer/flow-development-collection/pull/5

LIke `redirect:list` it will also take the redirects for certain hosts into account, if host is `null`. Otherwise you would have to export all redirects for each host separately (which is still possible).

@radmiraal Thanks a lot for taking care of the renaming part! Would you mind having a look at https://github.com/neos/neos-development-collection/pull/323/files ? It's way smaller, but no broken because of the renaming.

Also did some testing again, no major issues! Really looks like we're nearing the home stretch!

radmiraal:
@gerhard-boden I already looked at https://github.com/neos/neos-development-collection/pull/323, see: https://github.com/dfeyer/neos-development-collection/pull/7
@dfeyer can you have a look at that redirection rename PR?

gerhard-boden:
@radmiraal great! Sorry, should have checked if a PR was out already. Just did this about a minute ago:
https://github.com/gerhard-boden/neos-development-collection/commit/40b29237a18fa0a528f1b6db9ddead453078f9a2
:smile:

gerhard-boden:
Found yet another bug in the ``export`` command.
@dfeyer please merge this: https://github.com/dfeyer/flow-development-collection/pull/6 

Could not find any other bugs. After merging the PR mentioned above i can give :+1:

alexfruehwirth:
@dfeyer Tested it today and found two little storage issues.  
- `lasthit`is not set yet, or did i miss something?
- `incrementHitCount` update statement won't work if host is null. made a PR dfeyer/neos-development-collection#8 

All other things work as expected. so :+1:

dfeyer:
@alexfruehwirth  check your PR, and fix the query based on my comment if it make sense ;)

@gerhard-boden Will spend some time tomorrow to test your issue. 

Hopefully, we are OK tomorrow evening with the PR ;)

gerhard-boden:
@dfeyer Awesome, if you run into any trouble how to reproduce the issue don't hesitate to ping me.

kdambekalns:
Could everyone who agrees with this please add a **reaction**? The comments are so many by now, that it's hard to understand if this is good to go or not.

dfeyer:
@kdambekalns Currently checking the issue reported by @gerhard-boden and @alexfruehwirth I ping you as soon as it's fixed except that I think we are good to go

kitsunet:
Yep, looked quite good for me, if the issue is fixed we shuold get it merged...

dfeyer:
> Found yet another bug in the export command.

@gerhard-boden I don't fix your PR, but remove an array_filter in the Storage, like this findDistinctHosts also return null as this is a valid value for our case

> incrementHitCount update statement won't work if host is null. made a PR dfeyer/flow-development-collection#8

@alexfruehwirth Thanks for the PR, merge and solved the issue

> lasthitis not set yet, or did i miss something?

@alexfruehwirth my last commit solve this issue. The update of the lastHit date, was lost during the DQL refactoring 

Dear fellow reviewer, I think that one is ready to be merged 👍

gerhard-boden:
oh no... i hate to be the devils advocate, but now export command is broken again. If a have redirects valid for all hosts and redirects valid for specific hosts, only the specific ones are exported. Let me try something if I can fix that real quick.

radmiraal:
regarding the export bug: that is in the redirect package not in Flow, and as such only be part of the development collection. I'd be fine with merging a fix ofc., but I also wouldn't object if it would be fixed later if it needs more time

gerhard-boden:
Turns out the export command is fine after all(was just an issue with my local setup as it seems, because the problem vanished after a couple of more checkouts of the PR and deleting cache)
BUT we have this now
image

Neos.RedirectHandler.DatabaseStorage/Classes/Domain/Repository/RedirectRepository.php
```diff
@@ -173,9 +173,9 @@ public function findDistinctHosts()
     {
         /** @var Query $query */
         $query = $this->entityManager->createQuery('SELECT DISTINCT r.host FROM Neos\RedirectHandler\DatabaseStorage\Domain\Model\Redirect r');
-        return array_filter(array_map(function ($record) {
+        return array_map(function ($record) {
```
gerhard-boden: this will result in duplicate NULL values for hosts

gerhard-boden:
fixed it, setting up PR right away

gerhard-boden:
Please merge this to fix the latest list issue. Everything else is working perfect now!
https://github.com/dfeyer/flow-development-collection/pull/9

gerhard-boden:
Thanks for the merge @dfeyer just gave :+1:

PR #196, opened 2016-01-04
BUGFIX: Prevent overwriting source files

hhoechtl:
After switching from `FileSystemSymlinkTarget` to `FileSystemTarget`
the symlinks are still present in the `targetPath` and therefore Flow
tries to copy the stream content in the symlink (original file, same
stream) which results in a 0 byte file.

We handle this issue by removing the symlinks if present.

FLOW-384 #close

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
+++ b/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
@@ -315,7 +315,7 @@ class FileSystemTarget implements TargetInterface
 
         $targetPathAndFilename = $this->path . $relativeTargetPathAndFilename;
         $streamMetaData = stream_get_meta_data($sourceStream);
-        $sourcePathAndFilename = $streamMetaData['uri'];
+        $sourcePathAndFilename = $streamMetaData['uri'] ?? null;
 
         if (@fstat($sourceStream) === false) {
             throw new TargetException(sprintf('Could not publish "%s" into resource publishing target "%s" because the source file is not accessible (file stat failed).', $sourceStream, $this->name), 1375258499);
--- a/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
+++ b/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
@@ -181,7 +181,7 @@ class FileSystemTarget implements TargetInterface
      * @param StorageInterface $storage
      * @return void
      */
-    public function checkAndRemovePackageSymlinks(StorageInterface $storage)
+    protected function checkAndRemovePackageSymlinks(StorageInterface $storage)
     {
         if (!$storage instanceof PackageStorage) {
             return;
--- a/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
+++ b/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
@@ -20,6 +20,8 @@ use Neos\Flow\ResourceManagement\Publishing\MessageCollector;
 use Neos\Flow\ResourceManagement\PersistentResource;
 use Neos\Flow\ResourceManagement\ResourceMetaDataInterface;
 use Neos\Flow\ResourceManagement\ResourceRepository;
+use Neos\Flow\ResourceManagement\Storage\PackageStorage;
+use Neos\Flow\ResourceManagement\Storage\StorageInterface;
 use Neos\Flow\ResourceManagement\Storage\StorageObject;
 use Neos\Utility\Files;
 use Neos\Utility\Unicode\Functions as UnicodeFunctions;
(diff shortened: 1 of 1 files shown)


dfeyer:
:+1: from my side, look like the old PR was targeted the 3.0, and this one to 3.1, maybe I miss parts of the discussion

hhoechtl:
Ah, should the first pull request target the lowest version? @dfeyer

kdambekalns:
Yes, the PR must be against the lowest affected branch still maintained.

kdambekalns:
Also, check the (new) way of writing the subject prefix and issue references (I adjusted this PR). And the `Releases` footer is no longer needed (since the change will be merged into all higher branches).

kdambekalns:
@robertlemke might want to have a look on this, because resource management… ;)

dfeyer:
@hhoechtl Did you have some answer to @robertlemke comments ?

What's the status of this PR ?

hhoechtl:
Oh, I almost forgot about this PR.

$streamMetaData['uri'] is always set according to http://php.net/manual/en/function.stream-get-meta-data.php

albe:
Needs a rebase to retarget 3.0

albe:
Any chance of getting this rebased and merge ready?

hhoechtl:
It might take a few weeks until I can have a look

albe:
Ok, no worries. Would be nice to maybe get this finished for 4.1 since it's already pretty far, so I might ping again later

albe:
ping @hhoechtl - if you don't find time to look at this again don't worry :) btw, by now this is even supposed to be retargeted at 3.3 instead

jonnitto:
What is the status here? Any progress?

hhoechtl:
Should have been merged in 2016. Would need to copy paste it to the current flow version. I'm currently very busy.

jonnitto:
Sorry for that it wasn't merged in 2016. It would be great if you could manage it to solves this in the next days, as we soon have a new version…

Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
```diff
@@ -172,6 +174,26 @@ public function getName()
...
     }
 
+    /**
+     * Checks if the PackageStorage has been previously initialized with symlinks
+     * and clears them. Otherwise the original sources would be overwritten.
+     *
+     * @param StorageInterface $storage
+     * @return void
+     */
+    public function checkAndRemovePackageSymlinks(StorageInterface $storage)
```
kitsunet:
```suggestion
    protected function checkAndRemovePackageSymlinks(StorageInterface $storage)
```

Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
```diff
@@ -290,11 +314,19 @@ protected function publishFile($sourceStream, $relativeTargetPathAndFilename)
         }
 
         $targetPathAndFilename = $this->path . $relativeTargetPathAndFilename;
+        $streamMetaData = stream_get_meta_data($sourceStream);
+        $sourcePathAndFilename = $streamMetaData['uri'];
```
kitsunet: IMHO this is error prone as we cannot assume a stream with URI in all cases. So we need to at least null coalesce here.

```suggestion
        $sourcePathAndFilename = $streamMetaData['uri'] ?? null;
```

Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
```diff
@@ -290,11 +314,19 @@ protected function publishFile($sourceStream, $relativeTargetPathAndFilename)
...
 
         if (@fstat($sourceStream) === false) {
```
kitsunet: As we do a fstat here we might as well check the srouce stream path after this...

albe:
So is it right this bugfix is supposed to not land in 4.3+ but only next release (targeting master)?

Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
hhoechtl: I don't get it. If the `$sourceStream` is not available it stops here because of the exception

hhoechtl:
> So is it right this bugfix is supposed to not land in 4.3+ but only next release (targeting master)?

It could land in 4.3. I'll change it

Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
kitsunet: Yes but until that point you don't need the two added lines above, only after. so I would move them after to avoid the extra work in case the fstat check fails. But I don't mind, it's a minor thing.

PR #197, opened 2016-01-05
BUGFIX: ResourceStreamWrapper now saves handles for resource:// - URIs

fcool:
Trying to use file_get_contents("resource://yourhashhere") led to an exception in the StreamWrapper.
This fix addresses the issue by correctly setting the StreamWrapper handle to the opened resource.

Additionally a test is provided to demonstrate the problem.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Streams/ResourceStreamWrapper.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Streams/ResourceStreamWrapper.php
@@ -302,6 +302,7 @@ class ResourceStreamWrapper implements StreamWrapperInterface
         }
 
         if (is_resource($resourceUriOrStream)) {
+            $this->handle = $resourceUriOrStream;
             return $resourceUriOrStream;
         }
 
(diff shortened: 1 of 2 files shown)


kitsunet:
Looks like a pretty obvious bug! Will have closer look and try it later. Thanks a lot! Any idea if that problem also exists in 3.0? Because then the PR would need to be changed to target 3.0

fcool:
Yes, if I remember correctly in 3.0 it did hit me the first time. But you released 3.1 before we got stable, so we moved on to 3.0. Is there anything, I can do to help with targeting? (Merge to the 2.3 branch?) Sorry, for being stupid. This is my first pull request ever. ;)

fcool:
Just checked it - yes, this bug is also in 3.0. So pushing it there and make an new pull request?

fcool:
The failing test is strange - seems to be a cache issue - because this patch EXCACTLY tries to remove the bug, the tests fails at.

kitsunet:
> Just checked it - yes, this bug is also in 3.0. So pushing it there and make an new pull request?
Yest that would be great :)

fcool:
Closing this and create a new pull request at 3.0 Branch.

PR #199, opened 2016-01-05
BUGFIX: ResourceStreamWrapper now saves handles for resource:// - URIs

fcool:
Trying to use file_get_contents("resource://yourhashhere") led to an exception in the StreamWrapper.
This fix addresses the issue by correctly setting the StreamWrapper handle to the opened resource.

Additionally a test is provided to demonstrate the problem.

FLOW-302 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Streams/ResourceStreamWrapper.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Streams/ResourceStreamWrapper.php
@@ -303,7 +303,7 @@ class ResourceStreamWrapper implements StreamWrapperInterface
 
         if (is_resource($resourceUriOrStream)) {
             $this->handle = $resourceUriOrStream;
-            return $resourceUriOrStream;
+            return true;
         }
 
         $handle = ($resourceUriOrStream !== false) ? fopen($resourceUriOrStream, $mode) : false;
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Streams/ResourceStreamWrapper.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Streams/ResourceStreamWrapper.php
@@ -302,6 +302,7 @@ class ResourceStreamWrapper implements StreamWrapperInterface
         }
 
         if (is_resource($resourceUriOrStream)) {
+            $this->handle = $resourceUriOrStream;
             return $resourceUriOrStream;
         }
 
--- a/TYPO3.Flow/Tests/Functional/Resource/ResourceTest.php
+++ b/TYPO3.Flow/Tests/Functional/Resource/ResourceTest.php
@@ -1,0 +1,54 @@
+<?php
+namespace TYPO3\Flow\Tests\Functional\Resource;
+
+/*
+ * This file is part of the TYPO3.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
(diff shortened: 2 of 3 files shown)


kitsunet:
change makes sense, but tests need ot be fixed and it seems it needs to be rebased because the style issues are not in changed files.

fcool:
I'm not quite sure how the process here now would be. I'm sure I can handle the rebasing... about the two failing tests I am not that sure...
But the tests seem strange to me anyway - as the stream_wrapper interface tells us to return a bool for open and not a resource.
What to do next?

kitsunet:
I think it makes sense to adjust or even remove the failing tests. I will have a look.

kdambekalns:
Can we please have an issue fir this in JIRA? I am surprised to see this change and wonder why it breaks/is noticed only now. AFAIK there have nit been recent changes ti the StreamWrapper code, have there?

fcool:
No its broken since 3.0 - it could be, that it has been changes to the php interpret though. But the current 5.5 and 5.6 are broken even with just the "handle" - fix. That has not been the case with php 7.0 as the tests did clearly show. :(

kdambekalns:
Hm, could this be caused by https://github.com/neos/flow-development-collection/pull/43 - it's the only change around there…

kdambekalns:
No, it actually is broken since https://github.com/neos/flow-development-collection/blob/21c8b9495bf6f6877e5804794a85547a5efda7e8/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Streams/ResourceStreamWrapper.php

kdambekalns:
And the tests that fail, assume a return value that should never have been returned. Added with 5994ccd91464ccb385ff8efff4f9b1012e7e1ad0

kdambekalns:
@fcool I created a PR to against your branch to fix these tests: https://github.com/fcool/flow-development-collection/pull/1

kdambekalns:
:+1: from my side now…

fcool:
@kdambekalns Wooohoo! Thanks. Now all is green. ;) And verified with my current development project.

The Multi-Storage / Multi-Target Resource Management came with Flow 3.0 "stable", did it?

kdambekalns:
Yes, it did.

kdambekalns:
But, https://jira.neos.io/browse/FLOW-302 says 2.3.3 is affected, would you mind checking this, @fcool ?

fcool:
No problem! Which one to test best? 2.3.3 or 2.3.10?

(For 2.3.10 I'm already to 99.9% sure this problem does not exist their because or current working project works with the semantics and is currently productive with 2.3.10)

kdambekalns:
2.3.x should be fine…

kitsunet:
Mmm in the ticket it says "This was introduced by the change "Multi-Storage / Multi-Target Resource Management" (37396c7)", this was definitely done for 3.0, so 2.3 shouldn't be affected. I think the ticket is mis-tagged.

fcool:
I just copied the functional Test from https://github.com/neos/flow-development-collection/pull/199 to 2.3.10, adjusted to the old methods
$resource = $this->resourceManager->createResourceFromContent('fixture', 'fixture.txt');
$this->assertEquals('fixture', file_get_contents('resource://'.$resource->getResourcePointer()->getHash()));
And let it run. There has been no problem in 2.3.x.

The Test passes without errors.

kdambekalns:
Ok, updated the affected version in the ticket.

PR #200, opened 2016-01-07
BUGFIX: AOP Proxies properly check method existance before upward delegation

albe:
The usage of method_exists() checks inside the proxy class code was wrong by checking
the existance on the parent class of $this, which might in some circumstances not be
the current proxy class (ie. when checking method was originally called from a doctrine proxy).

This change fixes the method_exists calls by using the get_parent_class() without
parameter. Also, the method_exists check is added to the remaining method checks, that
currently only rely on is_callable, which in turn is always true on hierarchies that implement
__call magic method.


dfeyer:
:+1: make sense by reading

kdambekalns:
:+1: by reading, absolutely makes sense

kdambekalns:
Verified in a Neos 2.1 setup after having this upmerged from 2.3, works fine.

PR #201, opened 2016-01-11
BUGFIX: Default constructor arguments only autowired if needed

kitsunet:
So far all constructor arguments were treated as autowired unless
configured otherwise. This can result in problems and one example
is seen with the constructor of ``ArrayObject`` starting from PHP 7.0.2.

Unfortunately PHP does not allow Reflection of build-in functions
arguments. This includes methods of build-in classes like ``ArrayObject``.
This results in our ReflectionService to report ``null`` as default for
those arguments which will then be autowired. Since PHP 7.0.2 all three
arguments of the ``ArrayObject`` constructor are reported correctly by
Reflection but due to our autowiring the third argument will be given
as ``null`` in the proxy class which results in a fatal error.

The solution is to mark configured arguments that use the default or null
to be not autowired and do not give them to the original constructor if
they are not followed by an autowired argument.

FLOW-431 #close Fixes the issue by changing the way we apply default values

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/DependencyInjection/ProxyClassBuilder.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/DependencyInjection/ProxyClassBuilder.php
@@ -243,35 +243,43 @@ class ProxyClassBuilder
     {
         $assignments = array();
 
-        $argumentConfigurations = $objectConfiguration->getArguments();
+        $configurationArguments = $objectConfiguration->getArguments();
         $constructorParameterInfo = $this->reflectionService->getMethodParameters($objectConfiguration->getClassName(), '__construct');
         $argumentNumberToOptionalInfo = array();
         foreach ($constructorParameterInfo as $parameterInfo) {
             $argumentNumberToOptionalInfo[($parameterInfo['position'] + 1)] = $parameterInfo['optional'];
         }
 
-        foreach ($argumentConfigurations as $argumentNumber => $argumentConfiguration) {
-            if ($argumentConfiguration === null) {
+        $highestArgumentPositionWithAutowiringEnabled = -1;
+
+        /**  @var ConfigurationArgument $configurationArgument */
+        foreach ($configurationArguments as $argumentNumber => $configurationArgument) {
+            if ($configurationArgument === null) {
                 continue;
             }
-            $argumentValue = $argumentConfiguration->getValue();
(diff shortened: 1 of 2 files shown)


kitsunet:
I wanted to push the possible solution quickly to start refining it and get it merge ready. I guess in the end it must be merged to the 2.3 branch.

radmiraal:
:+1: works for me

kitsunet:
Note to self, should probaly be documented in the code, this still leaves one breaking case open. Lets imagine a build-in constructor with the following signature:

```
__construct($foo, $bar = 'bar', $baz = 'baz')
```

ReflectionService would report defaults for `$bar` and `$baz` as `nul`l (because build-in). If we now configure (via Objects.yaml) arguments 0 and 2 (but not 1), we will have the same fatal error occuring because in this case we need to autowire the optional argument 1 as 0 and 2 are given and will do so with a `null` value. Note this problem only exists for build-in constructors so IMHO the scope is very limited. We might just need to point this special case out somewhere.

kdambekalns:
Yes, must go into 2.3, as that is affected.

TYPO3.Flow/Classes/TYPO3/Flow/Object/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -243,35 +243,43 @@ protected function buildConstructorInjectionCode(Configuration $objectConfigurat
     {
         $assignments = array();
 
-        $argumentConfigurations = $objectConfiguration->getArguments();
+        $configurationArguments = $objectConfiguration->getArguments();
```
kdambekalns: I think the old name is better. It's about the "argument  configuration", not about "configuration arguments".

kdambekalns:
Tested on PHP 5.6.17 and 7.0.2, works fine with both (yes, Travis was faster than me–but I applied it to the 3.1 branch…).

The change does not apply to 3.0 and 2.3, though.

TYPO3.Flow/Classes/TYPO3/Flow/Object/DependencyInjection/ProxyClassBuilder.php
kitsunet: Agree, BUT the object is called `ConfigurationArgument` I just wanted to harmonize that...

kitsunet:
See  #202

PR #202, opened 2016-01-12
BUGFIX: Default constructor arguments only autowired if needed

kitsunet:
So far all constructor arguments were treated as autowired unless
configured otherwise. This can result in problems and one example
is seen with the constructor of ArrayObject starting from PHP 7.0.2.

Unfortunately PHP does not allow Reflection of built-in functions
arguments. This includes methods of built-in classes like ArrayObject.
This results in our ReflectionService to report null as default for
those arguments which will then be autowired. Since PHP 7.0.2 all three
arguments of the ArrayObject constructor are reported correctly by
Reflection but due to our autowiring the third argument will be given
as null in the proxy class which results in a fatal error.

The solution is to mark configured arguments that use the default or null
to be not autowired and do not give them to the original constructor if
they are not followed by an autowired argument.

FLOW-431 #close Fixes the issue by changing the way we apply default values

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/Configuration/ConfigurationBuilder.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/Configuration/ConfigurationBuilder.php
@@ -334,37 +334,42 @@ class ConfigurationBuilder
     {
         foreach ($objectConfigurations as $objectConfiguration) {
             $className = $objectConfiguration->getClassName();
+            if (!$this->reflectionService->hasMethod($className, '__construct')) {
+                continue;
+            }
+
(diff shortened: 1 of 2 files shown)


pide2000:
This Bugfix solves the following error: 

```
ArrayObject::__construct() expects parameter 3 to be a class name derived from Iterator, '' given
Type: TypeError
File: Data/Temporary/Development/Cache/Code/Flow_Object_Classes/TYPO3_Flow_Mvc_Controller_Arguments.php
Line: 311
Type: TYPO3\Flow\Core\Booting\Exception\SubProcessException
Code: 1355480641
File: Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
Line: 551
```

PR #203, opened 2016-01-13
BUGFIX: Make sure Fluid parser is configured correctly

kdambekalns:
When ``AbstractTemplateView::renderSection`` is called on a section that
has partials inside, it can happen that for parsing the partial the
ParserConfiguration is not set. This depends on the cache status of the
section and partial template.

If this happens, no Interceptors are applied to the partial. As that
includes the Escaping interceptor it might be a "security issue" when
the system runs in development context.

FLOW-430 #close


kitsunet:
:+1:

kdambekalns:
:+1:

PR #204, opened 2016-01-13
BUGFIX: StringHelper::startsWith broken with multiple occurance

kitsunet:
In case the search string appeared multiple times in the haystack
startsWith would not correctly report the occurance at the beginning
as it used the same code as endsWith. This is now fixed by using
``mb_strpos`` instead ``mb_strrpos``.

FLOW-423 #close


kdambekalns:
:+1: by reading

aertmann:
:+1: by reading

PR #205, opened 2016-01-14
BUGFIX: Value objects can be property mapped when submitted by identifier only

albe:
The identifier is unset from the submitted properties for Value Objects, because
they should use constructor arguments to be reconstituted. However, in forms
value objects are currently submitted by identifier, which will make property
mapping fail with an error.

This change fixes that by only unsetting the identifier if there are other
properties submitted.


albe:
This is a retargeted follow-up to #90

:+1 from me, see old PR for discussion

dfeyer:
:+1:

albe:
ping @kaystrobach @kitsunet

kaystrobach:
Will Troy

Von meinem iPhone gesendet

kitsunet:
I guess better than not working ;) :+1:

PR #206, opened 2016-01-14
BUGFIX: Allow creation of value objects during property mapping by default

albe:
Currently it is impossible to edit value objects without explicitly
allowing creation. This is because value objects are always
created, never loaded from persistence.

This change alters the PersistentObjectConverter to allow creation
of value objects by default.

FLOW-373 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/PersistentObjectConverter.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/PersistentObjectConverter.php
@@ -12,6 +12,7 @@ namespace TYPO3\Flow\Property\TypeConverter;
  */
 
 use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Annotations\ValueObject;
 use TYPO3\Flow\Persistence\PersistenceManagerInterface;
 use TYPO3\Flow\Property\Exception\DuplicateObjectException;
 use TYPO3\Flow\Property\Exception\InvalidPropertyMappingConfigurationException;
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/PersistentObjectConverter.php
```diff
@@ -226,7 +227,7 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
     protected function handleArrayData(array $source, $targetType, array &$convertedChildProperties, PropertyMappingConfigurationInterface $configuration = null)
     {
         if (!isset($source['__identity'])) {
-            if ($configuration === null || $configuration->getConfigurationValue('TYPO3\Flow\Property\TypeConverter\PersistentObjectConverter', self::CONFIGURATION_CREATION_ALLOWED) !== true) {
+            if (($configuration === null && $this->reflectionService->isClassAnnotatedWith($targetType, 'TYPO3\Flow\Annotations\ValueObject') === false) || $configuration->getConfigurationValue('TYPO3\Flow\Property\TypeConverter\PersistentObjectConverter', self::CONFIGURATION_CREATION_ALLOWED) !== true) {
```
dfeyer: Can we use ValueObject::class here ?

dfeyer:
Nice one, :+1: by ready, just a small nitpick

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/PersistentObjectConverter.php
albe: I wondered too, but it is not used in the other places in 3.0 branch (https://github.com/neos/flow-development-collection/blob/3.0/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/PersistentObjectConverter.php), so I did it the same way. Probably needs a follow-up for 3.1/master to use ::class

albe:
Had to push a revised logic, that always allows creation for value objects. This means it is not possible to disallow creation even through explicit configuration. Anyone has some thoughts about that? (The logic for making disallowing possible is really ugly and complex)

albe:
And one more push, because reflectionService in tests would return null for method calls instead of the expected false.

kitsunet:
Mmm, I am not sure about the not being able to disable. In general probably fine, but doesn't that pose a possible security or at least the possibility to flood the DB?

albe:
That's true (flooding DB), but under that perspective the whole idea needs to be reconsidered, because even with being able to disallow it explicitly the problem still exists. Actually, the main problem is how ValueObjects are stored (own table) currently.

I would see it this way: ValueObjects should never be created lonely, or else they are in fact AggregateRoots - they always belong to an entity. So if you are in a request where creation or update of the parent entity is allowed, the creation of value objects inside should always be allowed by default.

That still leaves room for flooding the DB, but no more than it were possible through the parent entity itself already, which needs to be explicitly configured.

albe:
Pushed a revised logic that allows disallowing the creation of value objects explicitly.

kitsunet:
k, then :+1:

kdambekalns:
I have the feeling this should be mentioned in the documentation somewhere, maybe?

Also, is DB flooding really an issue? We have code that "uniqueifies" value objects during persistence, IIRC.

kitsunet:
DB flooding in the sense of creating just always changing value objects. the same is uniqueified

albe:
Good point regarding the documentation, I always forget that this is really important. Now the question is, if this should still be considered a BUGFIX, plus it should probably be marked breaking as it changes behaviour.

albe:
So, still any concerns about this?

PR #207, opened 2016-01-14
BUGFIX: Respect constructor arguments that are no properties during property mapping

albe:
The PersistentObjectConverter does not evaluate constructor arguments
when determining the type of its children.

This patch adds the check and now constructor arguments that are not
also a property are mapped again.

FLOW-371 #close


kitsunet:
I agree that this makes sense. I would love to get rid of reflectionService in runtime though... But that can be refactored...

albe:
That's a good point, didn't even think about that yet, just copied/rebased the original changeset. Could that be solved with a statically compiled method?

albe:
:+1: from my side, makes sense and tests pass. Is there some performance test setup that could give feedback if that has performance implications?

kdambekalns:
:+1: by reading. On performance, there is use of the `ReflectionService` already, the change just invokes that every time now, instead of only when the property is not found in the schema. Could that order be switched maybe?

kdambekalns:
I switched that around and it still works (all tests pass). Anyway, this can be refactored later on as well.

PR #210, opened 2016-01-15
BUGFIX: Advices from parents are evaluated

albe:
Given an inheritance hierarchy like this: A <: B
when a advice for A and B was defined, the advices
for B were lost because they were
a) overwritten by A and
b) not accessible by B

This change modifies the visibility of the advice
properties to be protected and alters the generated
advice code in such a way that Flow_Aop_Proxy_target
MethodsAndGroupedAdvices is no longer overwritten

FLOW-367 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
@@ -477,9 +477,9 @@ class ProxyClassBuilder
         $this->buildInvokeJoinPointMethodCode($targetClassName);
         $this->buildMethodsInterceptorCode($targetClassName, $interceptedMethods);
 
-        $proxyClass->addProperty('Flow_Aop_Proxy_targetMethodsAndGroupedAdvices', 'array()');
-        $proxyClass->addProperty('Flow_Aop_Proxy_groupedAdviceChains', 'array()');
-        $proxyClass->addProperty('Flow_Aop_Proxy_methodIsInAdviceMode', 'array()');
+        $proxyClass->addProperty('Flow_Aop_Proxy_targetMethodsAndGroupedAdvices', 'array()', 'protected');
+        $proxyClass->addProperty('Flow_Aop_Proxy_groupedAdviceChains', 'array()', 'protected');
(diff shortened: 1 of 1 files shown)


kitsunet:
Test failures need to be tackled...

albe:
Ok, so I'm not sure what the reasoning behind "advicesAreExecutedAgainIfAnOverriddenMethodCallsItsParentMethod" is, maybe someone can enlighten me?

This test currently guarantees that if both A and B have a method X which is adviced, and A::X calls parent::X, then the advice is executed twice (see expected output "...Hello World World").
From a pure method-call POV this might sound reasonable, but I cannot for the life of me find a real world scenario, where this would be expected behaviour.
If I advice X and A->X() gets called, I expect that this single method call is adviced once, and not N times just because there is a deep inheritance hierarchy that deeply calls the parents method implementation. From an API POV I cannot/dont want to know if that method implementation internally depends on an abstract or parent class, that's an implementation detail of that method.

Looking at the history, maybe @kdambekalns or @robertlemke can shed some light?

kdambekalns:
Sorry, I cannot enlighten you in this case. Your reasoning makes sense, but… why we have that test for that behavior, I don't remember. Would need to research, but lack the time right now. :/

kitsunet:
IMHO the current behavior does make sense. If you advice an Interface method *ALL* implementations should trigger the pointcut. If that method does a parent call, well then the advice should be called twice. IMHO that's just reasonable. If you want to target only one of the two then you target the specific class. Which is what you basically would remove. How would you target a method only in the parent class and not in the child class. I think after all the current behavior is right looking at it. Otherwise we should write down cases and try to figure this out together at some point. Also the test was written by @foerthner so maybe he can enlighten us.

kitsunet:
Could it be that this is pretty much solved via #425 now?

kitsunet:
Could it be that this is pretty much solved via #425 now?

albe:
Will check

kitsunet:
Ping @albe how is the status here?

jonnitto:
Ping @albe

albe:
I'm not entirely sure the issue is fully resolved with #425, but given Christians explanation and the fact that this issue exists since 2014 and does not seem to cause much pain, I'd close this for now.

PR #211, opened 2016-01-19
BUGFIX: Don't leak version as "X-Flow-Powered" Header

obi12341:
The Version shouldn't be leaked as a Header value.
This is kind of a security hole. Maybe we should make it optional, but I don't understand the advantage of leaking the version anyway.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Response.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Response.php
@@ -111,7 +111,7 @@ class Response extends AbstractMessage implements ResponseInterface
     public function __construct(Response $parentResponse = null)
     {
         $this->headers = new Headers();
-        $this->headers->set('X-Flow-Powered', 'Flow/' . FLOW_VERSION_BRANCH);
+        $this->headers->set('X-Flow-Powered', 'Flow');
         $this->headers->set('Content-Type', 'text/html; charset=' . $this->charset);
         $this->parentResponse = $parentResponse;
     }
--- a/TYPO3.Flow/Tests/Unit/Http/ResponseTest.php
+++ b/TYPO3.Flow/Tests/Unit/Http/ResponseTest.php
@@ -217,7 +217,7 @@ class ResponseTest extends \TYPO3\Flow\Tests\UnitTestCase
 
         $expectedHeaders = array(
             'HTTP/1.1 123 Custom Status',
-            'X-Flow-Powered: Flow/' . FLOW_VERSION_BRANCH,
+            'X-Flow-Powered: Flow',
             'Content-Type: text/html; charset=UTF-8',
             'MyHeader: MyValue',
             'OtherHeader: OtherValue',


kitsunet:
Thanks, I like :) :+1: from my side

albe:
:+1: too, version information should not be given easily, true

obi12341:
any status on this?

daniellienert:
:+1: from my side. Could be merged I think.

albe:
Could this also make sense for down to 2.3? Or could this be breaking for applications that somehow depend on the version string in the header?

Otherwise I would just merge.

obi12341:
I think this should be merged in all versions. Of course only when it is possible.

kdambekalns:
Well, this isn't the _version_ but only the branch being in the header. And "security by obscurity" isn't security. So, if everyone likes this, go ahead. As for the branch, it could go into 2.3 and up. If that is desired, please close this PR and create a new one against that branch.

bwaidelich:
I agree that it may should be easier to adjust/disable that header but I wouldn't consider it a security issue.
Remember that this only outputs the *branch* not the *version*. If you're running a branch that is no longer supported (as in security fixes) *that* would be the vulnerability - not that you expose that fact.

bwaidelich:
damn, those dambekalns are always faster

kdambekalns:
yeah, but six seconds was close…

kitsunet:
Totally, agree otherwise I would have moved this to be a security relevant issue, but it's still unfortunate that it cannot be changed AND IT departments/clients might require it removed despite any arguments.

robertlemke:
Hi folks. I don't consider this a bug fix because I added that behaviour on purpose.

This header has the purpose to allow statistic crawlers detect Flow based applications and help us understanding how much Flow and Neos are used in public websites. I deliberately did not include the full version number, but only the branch and I don't consider that a real security problem.

You can switch that header off already by removing the header via the API, but I'm totally fine with adding an option to remove it completely, like Apache's server signature / server token options.

And yes, damn, Karsten is fast.

robertlemke:
@kitsunet I guess that those kind of departments/clients will simply remove the header via Nginx or some proxy if they really need to.

kitsunet:
I am all for a setting/option

obi12341:
@bwaidelich and @kdambekalns : I agree that you shouldn't run a branch which is not supported by fixes anymore. But believe me or not, it will happen that people are running outdated versions. I think it is not a good publicity if some bot can easy check if you are running an outdated branch and deploy some hacks for this branch.
@robertlemke I get your point and it is awesome to check how well people adopt to newer versions. But I think it is enough to just publish you are running a flow app and not which branch and stuff. And btw it should be documented somewhere how to disable it and it should be easy to disable it (I still think this should be disable by default in Production Context) and not by overwriting some classes via php or stuff like that.

robertlemke:
Suggestion for a setting:

TYPO3.Flow.Http.ApplicationToken: `value`

`value` could be one of:

- `ApplicationName` (the application name only, determined via the TYPO3.Flow.core.applicationKey setting)
- `MajorVersion`(the application name + major version, e.g. "Neos/2"
- `MinorVersion`(the application name + minor version, e.g. "Neos/2.1"
- `PatchVersion`(e.g. "Neos/2.1.1")
- `Full` (e.g. "Neos/2.1.1 Flow/3.0.2")

I suggest to have "MinorVersion" as the default.

And yes, I'm totally in favour of making this transparent and document how to change the token.

obi12341:
:+1: for @robertlemke but i would not print any Version into the header by default. My suggestion default : "Neos"

robertlemke:
Sorry for being so insistent with this small change ... but, Nginx and Apache both have server tokens on by default. What would be the issue for us doing the same?

bwaidelich:
:+1: for a setting, +-0 re the default value

obi12341:
Yeah and best pratice is to disable it directly :P

fcool:
Well working for several big groups - and after being pentested many times - we opted out to tell anyone, anything. But we went the easy way: Let apache do unset the filter. ;) 
My personal opinion is to let flow tell the world proudly who served the page!

robertlemke:
@obi12341 yes, I know. I mean, of course there's not need to waste too much time with this, and I'd give in if needed. But I really don't believe in security by obscurity and if the default is secure enough for Apache and Nginx, I wouldn't feel bad doing it as well. Only showing the branch and not exact version is a compromise, I think.

obi12341:
yeah maybe we waste to much time on this. I would be fine with an option to easy disable the header.

albe:
@robertlemke re it's good enough for Apache/Nginx, so it should be for us: I think the problem is more PHP here, so it's not completely fair to compare Flow with a widely tested and hardened webserver security wise.

And re security by obscurity: I don't believe in it either, but I think this is another viewpoint here. It's not about if Flow is secure if it doesn't leak version information, but rather that any known existing security hole can be automatically exploited. I've seen this problem a lot with e.g. wordpress/joomla installations.

+1 for documented setting, but don't allow patch level information - there's no benefit to telling that
+1 for defaulting to something less verbose, though that's by no means a blocking decision

robertlemke:
@albe all right.

How about `MajorVersion` as the default then and leave out the `PatchVersion` and `Full` options ... deal?

albe:
sounds good

kitsunet:
Can we progress on this?

robertlemke:
Closing this PR in favour of #299

PR #212, opened 2016-01-20
BUGFIX: InstallerScripts can run ``rescanPackages``

kitsunet:
In order to update package states when composer changes packages
the PackageManager must work with minimal dependencies.
To make that possible all neccessary initialisations are moved to
the constructor and ``InstallerScripts`` define necessary
constants.


kitsunet:
Merging as builds will fail anyway until this is merged.

PR #214, opened 2016-01-21
TASK: The ``package:rescan`` command is callable without proxies

kitsunet:
To mitigate problems with packages not being loaded while creating
proxies the ``package:rescan`` command now becomes a compile time 
command, so it must be called via ``typo3.flow:package:rescan``.


kitsunet:
@skurfuerst A little follow up which turns out to be quite helpful.

PR #215, opened 2016-01-21
TASK: Deprecate EarlyLogger

kitsunet:
The ``EarlyLogger`` was explicitly meant for the ``PackageManager``
which is not using it anymore. Therefore the EarlyLogger is deprecated
and bound for removal in the next major Flow version.

To underline the fact that it was never meant for user land code the
``@api`` annotations were removed.


albe:
:+1: by reading

daniellienert:
:+1: by reading

bwaidelich:
:+1: nice one

PR #216, opened 2016-01-21
TASK: Split Flow Log to separate package

kitsunet:
Creates the new `neos/flow-log` package containing the original
log infrastructure with the original namespace (`Neos\Flow`). So
nothing should change for userland code.

The Readme.rst contains a summary of dependencies of this package
to make it easier to work on removing them and making this package
more independent of Flow.

The choosen way to split allows for transitioning to a new namespace
while keeping the old classes.

The `SystemLoggerInterface` and `SecurityLoggerInterface` are kept
in Flow as they have not much meaning in the Logger package. Additionally
the `EarlyLogger` was not moved as it depends on those interfaces.


kitsunet:
Note that Travis will fail here because the autoloading cannot succeed for the new split package unless installed via composer (which is not the case for travis). That also means just checking out this branch will not help. You have to install this branch as master alias via composer.

kitsunet:
To test this change your base composer.json so that it includes:

```
    "repositories": [
        {
            "type": "git",
            "url": "https://github.com/kitsunet/flow-development-collection.git"
        }
    ],
```
and 
`"neos/flow-development-collection": "dev-split-log-package as dev-master",`

in the require section.

kitsunet:
Tests run fine if run locally with aforementioned config, so :+1:
Note that before merging we need to figure out/prepare subsplits for this new package.

kitsunet:
@bwaidelich There we go :) Still some stuff to do until this is really dependency free but a good start. I didn't want to modify too much in this PR as testing it is quite tedious.

composer.json
```diff
@@ -36,10 +37,14 @@
...
     "autoload": {
         "psr-0": {
+            "TYPO3\\Flow\\Log\\": "Neos.Flow.Log/Classes",
             "TYPO3\\Eel": "TYPO3.Eel/Classes",
             "TYPO3\\Flow": "TYPO3.Flow/Classes",
             "TYPO3\\Fluid": "TYPO3.Fluid/Classes",
             "TYPO3\\Kickstart": "TYPO3.Kickstart/Classes"
+        },
+        "psr-4": {
+            "Neos\\Flow\\Log\\": "Neos.Flow.Log/Classes"
```
bwaidelich: PSR-0 *and* PSR-4 ?

kitsunet: Yes, nothing speaks against that and I wanted to prevent WTF moments when we actually want to transfer the classes to a new namespace. Basically we could already create dummy classes extending the old ones (or vice versa) and it would work.

bwaidelich:
Nice one!
Could you maybe elaborate (in the change or - better - a corresponding ticket) what the mid- and longterm plan is regarding this?
So, e.g. what are the steps to be able to hook in one of https://packagist.org/providers/psr/log-implementation without further adjustments?
What I could imagine (without having thought about it too much):
* Adjust `TYPO3.Flow.Log` to be a valid `PSR-3` implementation
* Deprecate the Flow `*LoggerInterfaces` and create a slim adapter so that they work with the default logger (`TYPO3.Flow.Log`) as well as any other PSR-3-Logger..

Could that work?

composer.json
bwaidelich: Sorry if I'm thick, but I just *had* a WTF moment ;)
So you want to create dummy classes at the PSR-4 location and keep the PSR-0 ones as well??

kitsunet:
@bwaidelich yep just what you wrote. That's roughly my plan :)

gerhard-boden:
I think the comments need to be updated like https://github.com/kitsunet/flow-development-collection/blob/split-log-package/Neos.Flow.Log/Classes/TYPO3/Flow/Log/Backend/AbstractBackend.php#L5

kitsunet:
Merged master so no conflicts anymore, but this needs some more thought and time, might not be ready for 3.3 because I need to decouple some stuff still.

kitsunet:
Completely new change from scratch...

kdambekalns:
@kitsunet I added this to the 3.2 board, since @gerhard-boden mentioned the change yesterday. Anyway, what's up with the WIP label? Is that still correct?

kitsunet:
Unfortunately yes, despite everything working I think I need to go over some comments and dependencies.

Neos.Flow.Log/Classes/DefaultLogger.php
```diff
@@ -0,0 +1,161 @@
...
+     *
+     * This method allows for conveniently injecting a backend through some Objects.yaml configuration.
+     *
+     * @param Backend\BackendInterface $backend A backend implementation
+     * @return void
+     * @api
+     */
+    public function setBackend(Backend\BackendInterface $backend)
+    {
+        $this->backends = new \SplObjectStorage();
```
albe: Should this maybe walk the existing backends and close them first?

kitsunet:
@albe all fixed :)

Neos.Flow.Log/Classes/FormatUtility.php
```diff
@@ -0,0 +1,67 @@
+<?php
+namespace Neos\Flow\Log;
+
+/**
+ *
```
kdambekalns: Some comment, maybe?

Also, the name `FormatUtility` sounds rather generic. Something completely different would be a `FormattedVariable` class, where you pass the `$variable` to the constructor and then call `getAsPlaintext()` on that. Phew. :)

Speaking of which, I think it should be "plain text", not "plaintext".

Neos.Flow/Classes/Error/AbstractExceptionHandler.php
```diff
@@ -356,4 +356,8 @@ protected function splitExceptionMessage($exceptionMessage)
             'body' => $body
         ];
     }
+
+    protected function collectAdditionalInformationForException()
```
kdambekalns: Add a docblock, please. What does this do?

Neos.Flow/Classes/Log/Logger.php
```diff
@@ -10,116 +10,36 @@
...
-use Neos\Flow\Core\Bootstrap;
-use Neos\Flow\Error\Debugger;
-use Neos\Flow\Exception;
-use Neos\Flow\Http\HttpRequestHandlerInterface;
-use Neos\Flow\Log\Exception\NoSuchBackendException;
-use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 
 /**
- * The default logger of the Flow framework
  *
```
kdambekalns: Why did the docblock content disappear?

Neos.Flow/Classes/Log/LoggerFactory.php
```diff
@@ -21,6 +26,44 @@
  */
 class LoggerFactory
 {
+    /**
+     * @var array
+     */
+    protected $logInstanceCache = [];
+
+    /**
+     * @var
```
kdambekalns: What type is this?

Neos.Flow/Classes/Log/LoggerFactory.php
```diff
@@ -33,8 +76,24 @@ class LoggerFactory
...
+        return $this->logInstanceCache[$identifier];
+    }
 
+    /**
+     * Create a new logger instance.
+     *
+     * @param $loggerObjectName
+     * @param $backendObjectNames
+     * @param array $backendOptions
+     * @return mixed
```
kdambekalns: There should be something more specific than `mixed` here, no?

Neos.Flow/Classes/Log/LoggerFactory.php
```diff
@@ -33,8 +76,24 @@ class LoggerFactory
...
+            $this->logInstanceCache[$identifier] = $this->instantiateLogger($loggerObjectName, $backendObjectNames, $backendOptions);
+        }
+
+        return $this->logInstanceCache[$identifier];
+    }
 
+    /**
+     * Create a new logger instance.
+     *
+     * @param $loggerObjectName
```
kdambekalns: The first two parameters are `string`, right?

Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
```diff
@@ -0,0 +1,184 @@
+<?php
+namespace Neos\Flow\Log\ThrowableStorage;
+
+use Neos\Flow\Log\FormatUtility;
+use Neos\Flow\Log\ThrowableStorageInterface;
+use Neos\Utility\Files;
+
+/**
+ *
```
kdambekalns: Ah, it stores files? Hm. No comment?

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -149,6 +149,15 @@ Neos\Flow\Log\SecurityLoggerInterface:
     4:
       setting: Neos.Flow.log.securityLogger.backendOptions
 
+Neos\Flow\Log\ThrowableLoggerInterface:
+  scope: singleton
+  className: Neos\Flow\Log\ThrowableStorage\FileStorage
```
kdambekalns: This feels strange… How can a storage be a logger? A logger should be _using_ a storage, not be one. No?

composer.json
```diff
@@ -25,9 +25,13 @@
         "ext-mbstring": "*"
     },
     "replace": {
+        "typo3/eel": "self.version",
+        "typo3/flow": "self.version",
+        "typo3/kickstart": "self.version",
```
kdambekalns: Why do these three lines appear in this change?

composer.json
```diff
@@ -98,7 +105,10 @@
     },
     "require-dev": {
         "mikey179/vfsstream": "~1.6",
-        "phpunit/phpunit": "~6.0"
+        "phpunit/phpunit": "~6.0",
+        "neos/flow": "*",
```
kdambekalns: A bit self-requiring, no?

albe:
Eagle-Eye Karsten :) I agree to all your "complaints", though the first one (FormattedVariable vs. FormatUtility) is debatable.

kitsunet:
Thanks for looking closely @kdambekalns maybe little late but well. Should have tackled all of it.

PR #217, opened 2016-01-21
BUGFIX: Evaluate property conditions without security checks

kdambekalns:
The PropertyConditionGenerator for entity privileges allows the use of
global objects from the configured global context. If those in turn
may be secured, the system runs into an endless loop.

To avoid this, the fetching of the value for the operand is done without
security checks after this change.


kdambekalns:
moved from https://review.typo3.org/41942

bwaidelich:
:+1:  by reading

kdambekalns:
Maybe @foerthner and/or @kitsunet  could have another look?

kitsunet:
:+1:

foerthner:
:+1:

PR #218, opened 2016-01-21
TASK: When reconnecting to DB, log preceding exception

kdambekalns:
When flushing fails with Doctrine, we try to reconnect and flush again,
to work around dropped connections.

If the disconnection was caused by a "real" error, the cause was lost.
This change logs the exception that caused the reconnection to ease
debugging.


kdambekalns:
moved from https://review.typo3.org/41532/

albe:
:+1:

daniellienert:
:+1: by reading

bwaidelich:
:+1:  by reading

PR #219, opened 2016-01-21
BUGFIX: Resolve type in UniqueEntityValidator

astehlik:
Run the given validator value through TypeHandling::getTypeForValue()
to make sure doctrine proxies are resolved to the actual domain
model type.

Resolves: NEOS-1764


daniellienert:
@astehlik Thanks for that patch - also fixes my issue!
:+1: by reading and testing.

The bug also exists in older versions. Would you please reopen the pull request for the oldest supported version where the bug exists - which is 2.3 so it could be upmerged.

astehlik:
Thanks for testing :)

I created a new pull request for the 2.3 branch:

https://github.com/neos/flow-development-collection/pull/220

PR #220, opened 2016-01-22
BUGFIX: Resolve type in UniqueEntityValidator

astehlik:
Run the given validator value through TypeHandling::getTypeForValue()
to make sure doctrine proxies are resolved to the actual domain
model type.

Resolves: FLOW-433


daniellienert:
@astehlik Thanks for that patch - also fixes my issue!
:+1: by reading and testing.

The bug also exists in older versions. Would you please reopen the pull request for the oldest supported version where the bug exists - which is 2.3 so it could be upmerged

astehlik:
This pull request ist for the 2.3 branch, I hope this is OK. Please let me know if I should change anything.

bwaidelich:
:+1: by reading - this should also solve FLOW-433

daniellienert:
@bwaidelich I moved the ticket from Neos to Flow, so NEOS-1764 actually was FLOW-433.

bwaidelich:
@daniellienert awesome, thanks

PR #221, opened 2016-01-24
BUGFIX: Fix missing refactoring

futjikato:
Missing refactoring after 30b326a9735c80cd88c0acebcde127c02d6a198b
Method getComposerPackageNameFromPackageKey was moved from PackageManager to ComposerUtility but method call was still pointing to PackageManager.


kitsunet:
Good catch! Thanks. :+1:

PR #222, opened 2016-01-24
BUGFIX: Fix argument must be array but string given

futjikato:
Fix missing refactoring from 30b326a9735c80cd88c0acebcde127c02d6a198b


futjikato:
What I just notice is that the method now returns a path without trailing slash.
The doc comment above shows an example with trailing slash and the kickstart module seem to expect a trailing slash.

See example output for kickstart:package with this fix.
```
Please specify the required argument "packageKey": Futjikato.PluginTsView
Created .../Futjikato.PluginTsView/Classes/Futjikato/PluginTsViewController/StandardController.php
```

futjikato:
Method is already declared as deprecated so I added the trailing slash to provide backward compatibility.

kitsunet:
Yep, good call, I like :+1:

PR #223, opened 2016-01-26
BUGFIX: Make withoutAuthorizationChecks example realistic

kitsunet:
The example for usage of ``withoutAuthorizationChecks`` in the docblock
is wrong in as it shows variables used inside the closure as closure
arguments but that is impossible. Instead they must be added to the
closure context via ``use``.


bwaidelich:
:+1: can be merged as no-brainer IMO. Thanks!

PR #224, opened 2016-01-27
TASK: Split unicode utilities off from Flow

kitsunet:
Moves the unicode utilities of Flow to a separate package
retaining the original class namespace for backwards compatibility.
Additionally a few dependencies were cut without changing functionality.

### DIFF
--- a/Neos.Utility.Unicode/Classes/TYPO3/Flow/Utility/Unicode/Functions.php
+++ b/Neos.Utility.Unicode/Classes/TYPO3/Flow/Utility/Unicode/Functions.php
@@ -2,7 +2,7 @@
 namespace TYPO3\Flow\Utility\Unicode;
 
 /*
- * This file is part of the TYPO3.Flow package.
+ * This file is part of the Neos.Utility.Unicode package.
  *
  * (c) Contributors of the Neos Project - www.neos.io
  *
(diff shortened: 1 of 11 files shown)


kitsunet:
As with all package splits travis will always fail due to the fact that the composer autoloader doesn't have the information about the new package and therefore cannot load any of the moved files.

dfeyer:
Nice idea, we can maybe try the same way for the HostPattern matching strategy, like this we can break the coupling between the RedirectHandler package and Neos (love to have more really small package, NPM style)

kitsunet:
Absolutely

kitsunet:
I guess I will change the package type for this to library as it doesn't need any flow specific handling, like boot or proxies.

Neos.Utility.Unicode/Classes/TYPO3/Flow/Utility/Unicode/Functions.php
```diff
@@ -11,29 +11,28 @@
...
 
 /**
  * A class with UTF-8 string functions, some inspired by what might be in some
  * future PHP version...
  *
- * @Flow\Scope("singleton")
  * @api
  */
-class Functions
+abstract class Functions
```
kdambekalns: Why did you make that abstract?

kitsunet: It's something I personally prefer for classes that contain only static methods. Just to prevent instanciation. Can be removed of course.

PR #225, opened 2016-01-27
TASK: Split Filemonitor off to a separate package

kitsunet:
Splits the file monitoring into a separate package retaining the
namespace for backwards compatibility. No further refactoring
was done.


kitsunet:
As with all package splits travis will always fail due to the fact that the composer autoloader doesn't have the information about the new package and therefore cannot load any of the moved files.

dfeyer:
:+1: by reading

kdambekalns:
@kitsunet I changed `Readme.rst` and `composer.json` a bit (the same applies to the other "split jobs":

- the future Flow website URL is already reachable
- we should use the new package type
- license can be a plain string
- the PHP dependency should be a bit more strict (who knows what PHP 8 will bring)

What are your plans on the "leftover" dependencies? Solve in a later step?

kitsunet:
Yep, we need to iterate over the dependencies as soon as more is split.
Cache is the first where I tried to actually make the tests fully independent, now that I have an example that can be done to the other splits as well, but I think I would like to merge this and iterate on it.

kitsunet:
Needs some more cleanup!

kitsunet:
I am considering to drop this change. After having another look it seems to ingrained into the core of Flow. Up for discussion of course...

PR #227, opened 2016-01-27
FEATURE: Register privilege targets during runtime

RafaelKa:
Adds a signal for privilege targets initialization into PolicyService 
to allow adding privilege targets during runtime.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Security/Policy/PolicyService.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Security/Policy/PolicyService.php
@@ -201,6 +201,7 @@ class PolicyService
                 $this->privilegeTargets[$privilegeTargetIdentifier] = $privilegeTarget;
             }
         }
+        $this->emitPrivilegeTargetsInitialized($this->privilegeTargets);
     }
 
     /**
@@ -309,6 +310,20 @@ class PolicyService
      * @Flow\Signal
      */
     protected function emitConfigurationLoaded(array &$policyConfiguration)
+    {
+    }
+    
+    /**
+     * Emits a signal when privilege targets have been initialized
+     *
+     * This signal can be used to register privilege targets during runtime. 
+     * In the slot make sure to receive the $privilegeTargets array by reference so you can alter it.
+     *
+     * @param array<PrivilegeTarget> $privilegeTargets All initialized privilege targets
+     * @return void
+     * @Flow\Signal
+     */
+    protected function emitPrivilegeTargetsInitialized(array &$privilegeTargets)
     {
     }
 


foerthner:
Hi Rafael,

thanks for the PR! The general idea absolutely makes sense. However, the problem with privilege targets is, that we have privilege types that have to be evaluated during compile time (everything that's based on AOP, like method privileges). Adding or changing them during runtime will not work. So we would have to differentiate between runtime and compile time privilege types... I guess introducing interfaces for this would make sense. In the end you must not be able to change compile time privileges during runtime...

Greets Andi

kitsunet:
I also really dislike the idea to use an array reference received via signal to manipulate the privilegeTargets.

kitsunet:
Can someone more knowledable about these topics then me (so @foerthner or @bwaidelich for example) decide about the fate of this and either close this or move this forward in some other way?

bwaidelich:
I'd be interested in the use case, maybe @RafaelKa can enlighten us?

bwaidelich:
..to add to this:
I think `PrivilegeTargets` and `Roles` are a _fixed_ concept that should never change during runtime (and IMO even shouldn't be changed without re-deployment).
PrivilegeTarget matchers can be made very flexible (e.g. using runtime constraints) and I can't think of a case where you would need additional ones at runtime (but I'm very keen to hear about your scenario!).
One thing that _is_ still missing very much is a concept of `Groups` on top of the `Roles` concept, as this would allow a more flexible rights management which maybe covers your use case, too. See
https://jira.neos.io/browse/FLOW-386 (@foerthner, any news on this one)

RafaelKa:
This is for long running processes. In my case was it for [WAMP](https://tools.ietf.org/html/draft-oberstet-hybi-tavendo-wamp-02), which allows clients to register RPCs and publish and define rules for own RPCs and Events, who can call and listen on them.

Currently I do not use Flows content security for WAMP in any way. Also all content security things are from [Thruway](https://github.com/voryx/Thruway), but they are not so flexible as by Flow.

bwaidelich:
> This is for long running processes [...] which allows clients to register RPCs

What about a custom `Privilege` that looks up its rules? That might also be less fragile and easier to monitor. Have a look at the `NodePrivileges` from the Neos package. They implement `MethodPrivilegeInterface` so they can intercept method calls and do other things _additionally_.

You can use this without content security

RafaelKa:
Also can client define this privileges and add it to the privilege collection on long running process, which can test then if someone is granded or not?

bwaidelich:
> Also can client define this privileges

The privilege itself not (that's what I questioned above). Allowing the client to change the core policy is really dangerous & fragile IMO.
But in practice, yes, that's something you could achieve: Some kind of service that abstracts which RPCs a certain role / account is allowed to execute. You might even be able to that w/o a custom Privilege using runtime constraints.
Feel free to ask in the public slack channels if you need support with that!

RafaelKa:
> Allowing the client to change the core policy is really dangerous

Nooooo, only own Pubs and RPCs!

albe:
So is this still a topic or can this be closed?

albe:
Closing for now, please feel free to reopen if you think this is still relevant.

PR #229, opened 2016-01-30
TASK: RsaWalletService tweaks

kdambekalns:
Improves naming and documentation in RsaWalletService and the related command controller.


kdambekalns:
@foerthner @hlubek

foerthner:
Thanks for fixing the namings here. Totally makes sense by reading!

hlubek:
:+1: by reading, maybe add a note into the docs on how to generate that fingerprint from a PEM-encoded (private) key?

RafaelKa:
:+1:, but what about doc. comments and parameters in RsaWalletServiceInterface?

kdambekalns:
@RafaelKa Yes, what about them?  Can you elaborate?

@hlubek With docs you mean CLI docs (docblock) or manual?

kitsunet:
:+1: definitely an improvement. Will merge in 24h if no changes or objections.

RafaelKa:
@kitsunet then #230 to at the same time ;)

PR #230, opened 2016-01-30
FEATURE: Command to generate key pair for RsaWalletService

kdambekalns:
A new CLI command to generate a key pair:

  ./ flow security:generatekeypair

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/SecurityCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/SecurityCommandController.php
@@ -97,6 +97,20 @@ class SecurityCommandController extends CommandController
     }
 
     /**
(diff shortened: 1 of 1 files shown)


kdambekalns:
@foerthner @hlubek

foerthner:
Thanks, Karsten! Looks good by reading!

RafaelKa:
What about ECDSA? 
I mean if EcdsaWalletService is introduced in Flow, how to use this command?

kdambekalns:
Well, then we could add a type option. But adding that now seems useless, and the import commands would also need to be adjusted in that case… So, no blocker IMHO. Agreed?

RafaelKa:
@kdambekalns yes, agreed

kitsunet:
:+1:

PR #231, opened 2016-02-01
TASK: Extract caching to separate package

kitsunet:
!!!TASK: Extract caching to separate package

Splits off the caching framework into a separate package. 
All classes and tests have been decoupled and depend only on 
the independent utilities. Tests have been adapted and run 
with PHPUnit only.

With the ``CacheManager``, an extended ``CacheFactory`` and
the (now) deprecated ``AbstractCacheBackend``, remaining in 
Flow there should be no breaking changes to any public API.

This is marked breaking as it deprecates the ``@api`` marked
``AbstractCacheBackend``. Instead cache backends should now
extend from the ``Neos\Cache\Backend\AbstractBackend`` and
adapt to the new constructor.


kitsunet:
As this is a package split off again the autoloading in travis doesn't work out, travis will always fail.

kitsunet:
By now tests are running. Next step would be some readme and basic documentation. After that this is good to go for now. After this is merged the Utilities should be split so that the cache can actually be used standalone, additionally adding a PSR compatible frontend would be good.

kitsunet:
Readme is added, IMHO ready for merge now. Everything else can be done successively.

kitsunet:
This is pretty much good to go, but has a performance impact that is fixed by #300

dfeyer:
Awesome, will do review next week. We need to communicate when this one is ready to used by other people in the PHP community, really cool.

kitsunet:
Yep, when the split is done I would add a PSR compatible Frontend as well, I think then it's really a nice caching package for the PHP community.

kitsunet:
Updated to be mergeable in master...

kitsunet:
And again updated.

kitsunet:
@kdambekalns pls :)

albe:
Rebased, was rather easy after all. Looks good and tests pass locally.

kitsunet:
OK, so I rebased this (again) and checked the Cache tests. This needs careful (local) testing still, so please check!

kaystrobach:
@kitsunet atleast Travis is complaining ...

```
PHP Fatal error:  Class 'Neos\Cache\Backend\AbstractBackend' not found in /home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Cache/Tests/Unit/Backend/AbstractBackendTest.php(54) : eval()'d code on line 2
Fatal error: Class 'Neos\Cache\Backend\AbstractBackend' not found in /home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Cache/Tests/Unit/Backend/AbstractBackendTest.php(54) : eval()'d code on line 2

```

Neos.Cache/Classes/TYPO3/Flow/Cache/Backend/MemcachedBackend.php
```diff
@@ -2,7 +2,7 @@
 namespace TYPO3\Flow\Cache\Backend;
 
 /*
- * This file is part of the TYPO3.Flow package.
+ * This file is part of the Neos.Cache package.
```
johannessteu: Is it? Isn't it still in TYPO3.Flow ?

Neos.Cache/Tests/Functional/Backend/RedisBackendTest.php
```diff
@@ -54,9 +58,9 @@ public function setUp()
             $this->markTestSkipped('redis server not reachable');
         }
         $this->backend = new RedisBackend(
-            new ApplicationContext('Testing'), array('hostname' => '127.0.0.1', 'database' => 0)
+            new EnvironmentConfiguration('Redis a wonderful color Testing', '/some/path', PHP_MAXPATHLEN), ['hostname' => '127.0.0.1', 'database' => 0]
         );
-        $this->cache = $this->createMock(\TYPO3\Flow\Cache\Frontend\FrontendInterface::class);
+        $this->cache = $this->getMock(FrontendInterface::class);
```
kdambekalns: No, it must be `createMock()`, `getMock()` is deprecated and will trigger warnings.

Neos.Cache/Classes/TYPO3/Flow/Cache/Backend/AbstractBackendBase.php
```diff
@@ -0,0 +1,168 @@
+<?php
+namespace TYPO3\Flow\Cache\Backend;
+
+/*
+ * This file is part of the TYPO3.Flow package.
```
kdambekalns: Neos.Cache, eh?

kdambekalns:
Hm, I *do* get class loading errors, and I guess this is the culprit:


            "TYPO3\\Flow\\Cache": [
                "Neos.Cache/Classes"
            ],

Because we have matching classes in both packages… or is there anything do be aware of for testing, @kitsunet? I (locally) merged your branch into master, ran `composer dump-autoload`, even updated another package using composer. The functional tests pass, but the unit tests (of TYPO3.Flow and Neos.Cache) fail with

    Fatal error: Class 'Neos\Cache\Backend\AbstractBackend' not found in /…/Packages/Framework/Neos.Cache/Tests/Unit/Backend/AbstractBackendTest.php(54) : eval()'d code on line 2

and

    Fatal error: Class 'TYPO3\Flow\Cache\Exception' not found in /…/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Log/Exception/CouldNotOpenResourceException.php on line 19

kdambekalns:
Stepped through class loading in the `AbstractBackendTest` and the class loader doesn't know anything about that class, indeed.

kdambekalns:
Tried running the tests like this:

    phpunit -c Build/BuildEssentials/PhpUnit/UnitTests.xml Packages/Framework/Neos.Cache/Tests/Unit
    cd Packages/Framework/Neos.Cache ; phpunit
    cd Packages/Framework/Neos.Cache ; ../../../bin/phpunit

all to no avail. But a new message appeared: `Cannot stub or mock class or interface "TYPO3\Flow\Cache\Frontend\AbstractFrontend" which does not exist`

kdambekalns:
Oh, and this:

        "repositories": [{"type": "git","url": "git@github.com:kitsunet/flow-development-collection.git"}],

combined with

    "neos/flow-development-collection": "dev-split-cache",

leads to `Your requirements could not be resolved to an installable set of packages.` with about 60 lines of output…

kdambekalns:
The upside: I had this applied all day long while working on a project, and noticed nothing wrong.

kitsunet:
To run the Cache package tests:
`phpunit -c Packages/Framework/Neos.Cache/phpunit.xml.dist`

All other tests as usual. I see there is one test with wrong mocks in TypoScript, I will fix that in a separate PR.

kitsunet:
Newest commit fixes the create/get Mock thingy

kitsunet:
The correct way to install this for testing:


        "repositories": [{"type": "git","url": "git@github.com:kitsunet/flow-development-collection.git"}],

        "neos/flow-development-collection": "dev-split-cache as dev-master",


Then run your installation, test for the cache package:

    phpunit -c Packages/Framework/Neos.Cache/phpunit.xml.dist

All other tests as usual.
Works on my machine

robertlemke:
Thanks for driving this so far! I read almost all of the code and tested it based on Neos master, which works fine so far.

I personally don't like {inheritdoc} as it doesn't help me really while I'm reading the code, only if I read a rendered API documentation (which I very rarely do). But apart from my personal preferences, I suggest that we stick to one style unless we decided to go for another. Let's keep it consistent and just write down and probably repeat comments for every class and method.

More importantly though, regarding the backwards compatibility: I know that there is userland code out there which instantiates cache backends and frontends, combines them and then uses the `CacheManager` to register them. That's all marked as public API, so as far as I see it (and I may miss something because it was a long day), the constructor of the backends is incompatible to earlier versions.

Is that so?

Neos.Cache/Classes/TYPO3/Flow/Cache/GenericCacheFactory.php
```diff
@@ -0,0 +1,93 @@
...
+        return $cache;
+    }
+
+    /**
+     * @param string $backendObjectName
+     * @param array $backendOptions
+     * @return BackendInterface
+     * @throws Exception\InvalidBackendException
+     */
+    protected function instanciateBackend($backendObjectName, $backendOptions)
```
robertlemke: that's a typo - must be "instantiate" (also everywhere else)

Neos.Cache/Classes/TYPO3/Flow/Cache/GenericCacheFactory.php
```diff
@@ -0,0 +1,93 @@
...
+        return $backend;
+    }
+
+    /**
+     * @param string $cacheIdentifier
+     * @param string $cacheObjectName
+     * @param BackendInterface $backend
+     * @throws Exception\InvalidCacheException
+     */
+    protected function instanciateCache($cacheIdentifier, $cacheObjectName, $backend)
```
robertlemke: that's a typo - must be "instantiate" (also everywhere else)

kitsunet:
Will fix the naming issues. The `\TYPO3\Flow\Cache\Backend\AbstractBackend` still exists and is untouched so that should work backwards compatible if you extended that and register it. I do think we should discourage not using the Manager and Factory.

robertlemke:
@kitsunet yes, I've seen that the old `AbstractBackend` is still there. What I meant though was that there is application code out there which uses the existing concrete file backends and relies on their constructors.

In my eyes, this is a breaking change, because the constructor was marked as `@api` is in use by application code. The question is, how relevant that really is for 99% of the users.

If it's breaking, we can't just merge it as a TASK into 3.3.

kitsunet:
IMHO only the AbstractBackend constructor is marked `@api` (eg. Memcached has it's own which is not tagged) but I am fine moving this to next major...

kitsunet:
@robertlemke Added backwards compat layer for extended backends. Maybe you could quickly check and remove the -1, then I would retag it for 3.3.

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/ApcBackend.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    use BackendCompatibilityTrait;
+
+    /**
+     * Constructs this backend
+     *
+     * @param ApplicationContext $context Flow's application context
+     * @param array $options Configuration options - depends on the actual backend
+     */
+    public function __construct(ApplicationContextt $context, array $options = [])
```
robertlemke: There's a typo in `ApplicationContextt`

robertlemke:
Looks good in general, and I think the compatibility layer is a good compromise.
We should make sure that all of these backends really still work after this bigger change by quickly testing each of them with Neos for example.

albe:
The Tests needed to be adjusted to import the `Neos\Cache\Backend\FileBackend` instead of the deprecated `TYPO3\Flow\Cache\Backend\FileBackend`, since some were already adjusted to the new interface and failed with 

> Argument 1 passed to TYPO3\Flow\Cache\Backend\FileBackend::__construct() must be an instance of TYPO3\Flow\Core\ApplicationContext, instance of Neos\Cache\EnvironmentConfiguration given

Also, the `initializeObject` method of the Frontend needs to stay backwards compatible for now, so I readded the `$this->backend->setCache($this);` or else things will break when a cache is not created through the factory (in tests).

albe:
Ok, finally the Flow and the Neos.Cache UnitTests now pass for me (Windows 7, PHP 5.6, PHPUnit 5.5; & Ubuntu 14.04, PHP 5.5, PHPUnit 4.8).

albe:
@robertlemke Could you please take a quick look again and possibly remove your down-vote?

kitsunet:
robert is in holidays. Will remove LTS label and hope we get this in ASAP after 3.3

albe:
ping @robertlemke @kitsunet Can we get this resolved soonish?

dfeyer:
Look like the test of the cache package are green, but something is broken in the unit test of Flow, will check tomorrow ... sync with master has been quiet complex, so maybe I broke some stuff. Now time to sleep

kdambekalns:
I too am guilty of not having reviewed this, but… @robertlemke should have a look again, since this is going to master a potential BC break is not impossible to merge.

kitsunet:
Well, I fixed roberts comment already back in august but the -1 was never removed. So never -1 unless you are available for checking again please...

kdambekalns:
I dared to merge master into this… had failing tests, but I had to hack composer autoloading a bit to get it working, so I may have false negatives. But this does need thorough checking… Feel free to throw the merge commit away if it's too broken. 🙈

kdambekalns:
So, installed properly via composer. It comes down to 5 errors (4 in `TYPO3\Flow\Cache\Tests\Unit\Backend\RedisBackendTest`, one in `TYPO3\Flow\Cache\Tests\Unit\Backend\RedisBackendTest`) and 7 failures (all in `TYPO3\Flow\Cache\Tests\Unit\Backend\RedisBackendTest`) for the unit tests.

kdambekalns:
So, the Redis tests must have always been (in this branch). The mocked `\Redis` isn't even used in the `RedisBackdend`, so…

robertlemke:
@kitsunet I removed my -1 but didn't have the chance to read all of the final code again.

The questions I have, which you probably can answer easily, are:

- is this a breaking change concerning any methods which previously have been marked as `@api`?
- which steps do I have to take as a user to adjust my project to this change? Is this documented?
- are tests fine?
- did at least one person test each backend with a real Neos instance?

I suggest that we have answers for these, and if you do, go for it!

kitsunet:
- [x] * breaking change concerning any methods which previously have been marked as @api?

Nope, one api class is deprecated though and should be replaced with the new one until the major version after the upcoming (there is no hurry, the fallback code is minimal).


- [x] * which steps do I have to take as a user to adjust my project to this change? Is this documented?

No changes are needed for existing projects unless you extended an existing backend (because the class name changed) but those itself where never part of the public api.
The cache usage and configuration stays exactly as it was.

- [x] * Are tests fine?

Yes, fixed the last failures in the redis tests (used wrong constructor parameter).

- [x] * Did at least one person test each backend with a real Neos instance?

I did checks with various configurations and didn't encounter any problems.

PR #232, opened 2016-02-01
BUGFIX: Initialize Bootstrap::requestHandlers with empty array

kitsunet:
The ``requestHandlers`` property of the Bootstrap is used as an
array but is never initialized. For sake of cleanliness it should be.


kitsunet:
no brainer, will merge in 24h if no one objects.

bwaidelich:
:+1:

kdambekalns:
:+1:

PR #233, opened 2016-02-01
TASK: InternalRequestEngine should clear persistence after request

albe:
To correctly simulate a real http request, the internal request engine should clear the state
of the persistence manager after a request.


kitsunet:
Makes sense :+1: from me. Should probably do a whole lot more, but a start. Would merge soonish if no one objects.

PR #234, opened 2016-02-02
TASK: Change typo3.org reference to neos.io

albe:
This change uses www.neos.io instead of typo3.org for testing the CurlEngine.


kitsunet:
Nice, no-brainer, merging.

PR #235, opened 2016-02-05
BUGFIX: setHeader fails if 'Content-Type' value is not string

mkeitsch:
If you add 'Content-Type' to the automatic request headers of a virtual browser, it fails every time. The given value will be casted to an array automatically. But 'Http\AbstractMessage::setHeader()' expects a string value for 'Content-Type'.
UnitTest extended especially for this header field.
See: https://jira.neos.io/browse/FLOW-305


mkeitsch:
Sorry, have seen, that there is a workflow with branches. Was my first pull request ever. I try it again...

PR #236, opened 2016-02-05
BUGFIX: 'Content-Type' (automatic) header fails in virtual browser

mkeitsch:
With adding the 'Content-Type' header to the automatic headers of a virtual browser, the request fails every time.
The given value is casted to an array by the 'Http\Headers::set()' method. When setting this header in a 'Header' (!) instance ('Http\AbstractMessage::setHeader()') of the request, a string is expected especially for 'Content-Type', but an array is given.

'BrowserTest' extended especially for this header field.
See Issue: https://jira.neos.io/browse/FLOW-305

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/AbstractMessage.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/AbstractMessage.php
@@ -114,8 +114,15 @@ abstract class AbstractMessage
     {
         switch ($name) {
             case 'Content-Type':
+                $typeOfValues = gettype($values);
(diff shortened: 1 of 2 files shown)


kitsunet:
Hey, thanks for the bug fix :)
I see that this would work, but in general a multi value "Content-Type" doesn't make much sense so I would handle the case even stricter and just get the first element if it's an array and if there are more elements throw an Exception. That also makes the conversion back to array unnecessary. 
Additionally our way of merging bug fixes is always to the lowest supported and affected branch, I guess that would be 2.3, so it would be awesome of you would create a new PR for that branch. Thanks :)

PR #237, opened 2016-02-07
TASK: Split Lock utility off

kitsunet:
This separates the Flow Lock implementation into a separate
package removing almost all dependencies on Flow.
The only dependencies left is on the File utility, so as soon
as that is separated this package can be used without Flow core.

This change is supposed to be non breaking and no fucntionality
has been changed. The Flow bootstrap takes care of initializing
the new ``LockManager``. The only actual change is that the
Exception ``LockNotAcquiredException`` has a different
namespace now.

### DIFF
--- a/Neos.Flow.Lock/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
+++ b/Neos.Flow.Lock/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
@@ -2,7 +2,7 @@
 namespace TYPO3\Flow\Utility\Lock;
 
 /*
- * This file is part of the TYPO3.Flow package.
+ * This file is part of the Neos.Flow.Lock package.
  *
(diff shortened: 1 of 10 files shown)


kitsunet:
Again a package split so tests will probably break... Test locally please.

dfeyer:
:+1: Work fine, maybe it's time to make this package PSR4, but can be done in an other PR ?

For me this one is merge ready

kitsunet:
I didn't switch to PSR-4 on purpose so we can introduce a new namespace ("Neos\...") on PSR-4 level later and migrate to that by having bc classes in the old namespace still around.

kitsunet:
@kdambekalns as soon as any of these split changes is to be merged we need to take care of git split and deployment stuff. Can you prepare that?

kdambekalns:
Yes, I'd prepare that (and document it along the way). Basically it should be enough to create a new target repo and adjust the configuration of the neos/slicer repo. About the release script(s) I'd need to check myself.

PR #238, opened 2016-02-08
BUGFIX: package:create allows flow package typep

kitsunet:
The condition to check for a valid Flow package type in
the ``package:create`` command was not negated correctly and
so didn't allow creation of any flow package type.

This is now fixed and creation of packages works again.


kitsunet:
Merging as no-brainer.

PR #239, opened 2016-02-08
BUGFIX: 'Content-Type' (automatic) header fails in virtual browser

mkeitsch:
With adding the `Content-Type` header to the automatic headers of a virtual browser, the request fails every time.

The given value is cast to an array by the `Http\Headers::set()` method. When setting this header in a `Header` (!) instance (`Http\AbstractMessage::setHeader()`) of the request, a string is expected especially for `Content-Type`, but an array is given.

`BrowserTest` extended especially for this header field.

FLOW-305 #close


kitsunet:
:+1:

kdambekalns:
This replaces https://github.com/neos/flow-development-collection/pull/236 and handles things a bit stricter. 'AbstractMessage' is first available in 3.0. In 2.3 there are no automatic headers in the virtual browser.

kdambekalns:
:+1:

PR #240, opened 2016-02-09
BUGFIX: PackageManager only uses MetaData type if it was set

kitsunet:
The deprecated package metadata can contain a type but we should
use it for generating a composer manifest only if the type was
actually set otherwise the given type should be used.


kdambekalns:
:+1:

PR #241, opened 2016-02-09
BUGFIX: Inactive packages are handled correctly

kitsunet:
After the package manager refactoring several inconsistencies with
inactive packages and moving them to a separate directory arose.
The code was adapted to make this more consistent and allow loading
of composer manifests from deactivated packages.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
@@ -336,7 +336,7 @@ class PackageManager implements PackageManagerInterface
         $filteredPackages = [];
         /** @var $package Package */
         foreach ($packages as $package) {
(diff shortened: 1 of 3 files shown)


kitsunet:
@kdambekalns please also test, but this should fix all the problems we encountered. I tested a full cycle of creating, deleting, deactivating and activating packages. Also try to manually set a package inactive in `PackageStates.php` and then run `flow:package:rescan`. The package should be moved to `Inactive` and the PackageStates packagePath updated accordingly.

kdambekalns:
This fixes the leftovers the `scanAvailablePackagesTraversesThePackagesDirectoryAndRegistersPackagesItFinds` test created in the "real world" (on the actual filesystem instead of in vfs). Good start. Now for a review. :)

kdambekalns:
:+1: by reading and testing

PR #242, opened 2016-02-10
FEATURE: Allow setting validator options for element validators

kitsunet:
The element validators configured for a collection validator couldn't
accept options so far, with this change it is possible to set the
``elementValidatorOptions`` option on the ``CollectionValidator`` to
give options to the element validator.

FLOW-435 #resolve


kdambekalns:
Would be nice if the test actually checked the options are passed on, otherwise :+1:

kitsunet:
Right, will add a test for that!

kitsunet:
test added, IMHO good to go :+1:

aertmann:
Hey. What about documentation? Is it already documented?

kitsunet:
Right, but I think the validator documentation is automatically generated?

PR #243, opened 2016-02-10
TASK: Fix comment on Flow.session.name setting

kdambekalns:
Leaving the name empty will not work as advertised.


kdambekalns:
@robertlemke @kitsunet

robertlemke:
who believes in advertising promises anyway?

:+1:

kitsunet:
Fine, but did the comment have any plan behind it maybe? Could it be useful to have different cookie names per domain? Probably not... :+1:

kdambekalns:
I didn't check, but _maybe_ it worked at some point… Anyway, it doesn't make sense. And you don't need different names per domain, since they are sandboxed from each other anyway (through the domain they relate to).

PR #247, opened 2016-02-10
BUGFIX: Fix session garbageCollector probability of 0

daniellienert:
I would expect a session garbageCollection.probability of zero would never
trigger garbage collection.

Previously it actually did with a probability 1 / 101 as a it results in the
equation rand(0, 100) <= 0.


albe:
:+1: by reading

PR #251, opened 2016-02-12
TASK: Exclude classes pulled in by doctrine/migrations 1.3

kdambekalns:
Pulling in doctrine/migrations 1.3 works fine, but it comes with some
new dependencies that need to be excluded from reflection.


albe:
:+1: by reading

kitsunet:
:+1:

kitsunet:
narf, too late... We need to revert this sorry, doctrine/migrations 1.3 requires PHP > 5.5 but Flow 2.3 is still PHP > 5.3.2

kdambekalns:
Sorry, I knew this, but didn't get around to fixing it. :)

PR #253, opened 2016-02-15
Revert "TASK: Exclude classes pulled in by doctrine/migrations 1.3"

kitsunet:
Reverts neos/flow-development-collection#251

Flow 2.3 still runs on PHP 5.3 but doctrine/migrations 1.3 requires at least 5.5, therefore
we need to revert this requirement to make Flow 2.3 work smoothly with PHP 5.3.


kitsunet:
@kdambekalns

PR #256, opened 2016-02-15
Corrected error message in package:create command

SoulCover:
`$packageType` is used in the error message if is is not a valid Flow package tpye.


kdambekalns:
This will never land in the maintained branches, I fear… Flow 2.3 is what we currently support, this went into 2.1.

kdambekalns:
See #257…

kitsunet:
Ups, I was confused here for a sec... Thanks!

kdambekalns:
No problem, I just wondered why an "ancient" ancestor commit suddenly appeared at the "top" of my git history… ;)

PR #257, opened 2016-02-15
BUGFIX: Correct error message in package:create command

kdambekalns:
$packageType is used in the error message if is is not a valid Flow package
type.


kdambekalns:
Paging @SoulCover and @kitsunet :)

kitsunet:
Great, thanks!
Note for @SoulCover: In master this has changed a lot and the check isn't make anymore.

kitsunet:
:+1:

skurfuerst:
:+1: by reading. merging with previous votes!

PR #260, opened 2016-02-15
BUGFIX: Avoid warning on empty controller class

daniellienert:


kdambekalns:
Hm, why would one want to have an empty controller class? Anyway, :+1: by reading.

daniellienert:
If someone does, I don't want to see a red warning ;)

albe:
Would prefer to avoid the additional indenting and instead do an early "continue" statement if methodNames is not an array. Otherwise +1

kdambekalns:
Agree to the early continue - would be consistent with the way it's done earlier in the method.

kitsunet:
Yes please, less nesting = less context in your head while reading :)
Could also do early return on the `if´ around the `foreach` which takes a nesting level away.

daniellienert:
Thanks a lot for the reviews! You are absolutely right.

albe:
:+1: by reading

PR #261, opened 2016-02-16
FEATURE: Implement Lock strategy based on Semaphore

dfeyer:
This change introduce a new LockStrategy based on IPC Semaphore. This
strategy use only memory so no lock file stay on disk. This strategy can be used to avoid the I/O problem of the default `FlockLockStrategy`. 

_Warning_: This strategy don't work for clustered Flow application, because the lock is only available locally for each server. A more complex strategy, based on Redis, is required for a clustered Flow application.

You can change the default Lock strategy in Flow settings:

```
TYPO3:
    Flow:
        utility:
            lockStrategyClassName: 'TYPO3\Flow\Utility\Lock\SemaphoreLockStrategy'
```


albe:
Nice, the Flock files issue is really a problem and that solves it elegantly! Just one nitpick at the moment: I'd prefer to use "use TYPO3\Flow\Utility\Lock\Factory as LockFactory;" to have the Factory method calls read more clearly. Otherwise total +1 so far, keep it up :)

bwaidelich:
Nice one and I agree to Alex's comment!

kitsunet:
What is still needed to get this out of WIP?

kitsunet:
Ok, I like it a lot. Works great. I would personally split this into separate PRs. One just the strategy which works well on it's own and can be merged right away IMHO. Then we need to look at the other stuff. I generally like the idea of the LockFactory but would think about that again. I think what it does could be integrated into the Lock class which hasn't much use currenly.

And this needs a rebase to the split package. Note that the Lock package shouldn't have a dependency on Flow so the Scope annotation must be removed and the exception was renamed...

albe:
One thing to consider/properly document: This strategy doesn't work for a multi-server setup with a shared storage, since the semaphores are not shared among servers. For that another strategy with memcache/redis would be needed, but that's a different topic.
It should just be clearly communicated what the limits/use cases of this strategy are.

kitsunet:
Wait, we use the locking for filecaches only anyway, so that's not a problem.

albe:
Quite the opposite, no? If a file cache is stored on a shared folder, exactly that case would break badly.

kitsunet:
file cache on shared folder is broken by design, IMHO also the current flock lock won't save you in that case.

kitsunet:
@dfeyer Can you rebase this on the split package? :)

dfeyer:
@kitsunet thanks for the reminder, I offline tomorrow but can rebase this one for the end of the week

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Factory.php
```diff
@@ -34,13 +34,18 @@ public static function acquire($subject, $exclusiveLock = true)
...
-    public static function acquireCallback($subject, $exclusiveLock, $callback)
+    public static function acquireCallback($subject, $exclusiveLock, \Closure $callback)
     {
         $lock = self::acquire($subject, $exclusiveLock);
-        $callback();
-        $lock->release();
+        try {
+            $callback();
+            $lock->release();
+        } catch (\Exception $exception) {
```
albe: `finally` instead maybe?

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Lock.php
```diff
@@ -78,6 +77,18 @@ public static function setLockManager(LockManager $lockManager = null)
 
     /**
      * Releases the lock
+     * @param boolean $exclusiveLock TRUE to, acquire an exclusive (write) lock, FALSE for a shared (read) lock. An exclusive lock is the default.
+     * @return void
+     */
+    public function acquire($exclusiveLock)
```
albe: That's a bit irritating IMO - the constructor already takes the argument for exclusiveLock, but it must also be provided in acquire again?

Also, since acquire is now a public method, what is the expected behaviour if someone does this:
`$lock = new Lock(...); $lock->acquire(..);`?

albe:
Just left two quick structural comments, will take a deeper look at the implementation later.

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Lock.php
dfeyer: > Also, since acquire is now a public method, what is the expected behaviour if someone does this: $lock = new Lock(...); $lock->acquire(..);?

I think the Lock must not be part of the public API because of this

albe: As @kdambekalns put it - if it is there it will be used ;)

kdambekalns: On top of all that, the docblock says "Releases the lock"… ;)

dfeyer: Arf ... we acquire the lock in the constructor and in the aquire method ... that's bad :(

dfeyer: Basically removing Lock::acquire method is the best solution. When you instanciate a Lock, it acquire the lock, you can release. But if you need a new Lock, you need to instanciate a new object

kitsunet:
Ok @dfeyer I would love to push this forward. Can you rebase?

albe:
Ok, looking good so far.
Just one last nit-pick: When trying to use the SemaphoreLockStrategy without semaphore extension being available, you get an error that `\TYPO3\Flow\Utility\Lock\sem_get()` could not be found. That's pretty misleading and I had to check PHP docs first to realize that this extension is just not available for `Windows` at all. So it would be nice to get a useful message then.

Next on this feature should be properly documented so that people can make use of it. Unfortunately it's not possible to make it the default LockStrategy. That can go into another PR though, since I can't see any fitting spot in the current Documentation structure for this. Probably needs a new Chapter "Advanced" or similar, where optimizations and generic cookbook style tips are collected.

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Factory.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+class Factory
+{
+    /**
+     * @param string $subject
+     * @param boolean $exclusiveLock TRUE to, acquire an exclusive (write) lock, FALSE for a shared (read) lock. An exclusive lock is the default.
+     * @return Lock
+     * @deprecated Use Factory::acquireCallback, deprecated in Flow 3.3
+     */
+    public static function acquire($subject, $exclusiveLock = true)
```
kitsunet: Why do we have this new method that is deprecated right away? Can't we just offer acquireCallback as new method?

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Factory.php
```diff
@@ -0,0 +1,49 @@
...
+        return new Lock($subject, $exclusiveLock);
+    }
+
+    /**
+     * @param string $subject
+     * @param boolean $exclusiveLock TRUE to, acquire an exclusive (write) lock, FALSE for a shared (read) lock. An exclusive lock is the default.
+     * @param \Closure $callback A callback executed before the relase of the lock
+     * @return void
+     */
+    public static function acquireCallback($subject, $exclusiveLock, \Closure $callback)
```
kitsunet: Shouldn't we drop the $exclusiveLock argument in favor of always exclusively locking? That would prevent race conditions with FlockLock and removing lock files.

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Lock.php
```diff
@@ -41,17 +36,21 @@ class Lock
...
      */
-    protected $exclusiveLock = true;
+    protected $exclusiveLock;
 
     /**
      * @param string $subject
-     * @param boolean $exclusiveLock TRUE to, acquire an exclusive (write) lock, FALSE for a shared (read) lock. An exclusive lock ist the default.
+     * @param boolean $exclusiveLock TRUE to, acquire an exclusive (write) lock, FALSE for a shared (read) lock. An exclusive lock is the default.
      */
     public function __construct($subject, $exclusiveLock = true)
```
kitsunet: We should probably add @deprecated here to notify people that they should use the LockFactory?

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/SemaphoreLockStrategy.php
```diff
@@ -0,0 +1,173 @@
...
+use TYPO3\Flow\Annotations as Flow;
+
+/**
+ * A Semaphore based lock strategy.
+ *
+ * This lock strategy is based on Sempahore.
+ *
+ * @see http://php.net/manual/fr/ref.sem.php
+ *
+ * @Flow\Scope("prototype")
```
kitsunet: No Flow annotations in this package, it's the default anyway.

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Factory.php
```diff
@@ -0,0 +1,49 @@
...
+ * This file is part of the TYPO3.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use TYPO3\Flow\Annotations as Flow;
```
kitsunet: Needs to be removed.

kitsunet:
Added a few final comments, if oyu agree but don't have time I can take care as well.

dfeyer:
I can take care of the comment tomorrow, Neos day here

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Lock.php
dfeyer: Basically we should remove the API annotation from this class, the factory is the API

kitsunet:
Any news here?

albe:
Just found this while digging why the sem_get would fail for me: https://forge.typo3.org/issues/66503

Could it be that we acquire too many semaphores at the same time? I checked and on our system `cat /proc/sys/kernel/sem` returns "250  32000  32  128" - so effectively limited to 128 semaphores at a time for the *whole system*

albe:
Ok, so after further digging, it exactly fails at the 128th acquired semaphore. Houston, we got a problem...

kitsunet:
Works like a charm on mac. I mean, it doesn't hurt to have this in right. Either your system supports it or not... @albe ?

kitsunet:
Would merge soon if no objections. It's optional anyway.

albe:
But not cool if you configure to use a strategy and then your setup blows with an unspecifc "could not acquire resource semaphore". We'd at least need to revamp the documentation a bit, and there's also some left-over code from the prior READ/WRITE implementation (why was that dropped anyway?).

I'll take another look at getting this to run on our linux machine, can't imagine why it should work flawlessly on mac but not on a widespread linux dist.

Also, having adjusted Flow to the new Factory locally I'm no longer so confident about that Factory class (it's now just a single static wrapper around Lock instanciation with try .. finally, so that could as well go directly into Lock class instead). Plus it's a bit cumbersome currently to work with the `acquireCallback` method, as you have to use the `use` statement to let the callback know on which subject to work and to return anything:

```
LockFactory::acquireCallback($subject, function() use ($subject, &$someData) {
   // Some locked calculation/IO on $subject
  $someData = $result;
});
```
Also, this method completely loses the distinction of exclusive and shared lock by ignoring the second constructor parameter of `Lock`.

I'd rather suggest s.th. like this:
```
$someData = Lock::exclusive($subject, function($subject) {
    // Some locked calculation/IO on $subject
    return $result;
});
```

by simply changing the static method a bit:
```
public static function exclusive($subject, \Closure $callback) {
   $lock = new Lock($subject, true);
   try {
      return $callback($subject);
   } finally {
      $lock->release();
   }
}
```

If no one objects, I'll push something later.

kitsunet:
Thing is, as discussed with multiple people, a shared lock is something you would have to solve on a lower level than PHP IMHO. I thought it through and I don't see a way to work with shared locks in a good way. That's why I suggested to just drop shared lock functionality.

albe:
Oh, ok... fair enough. But has anyone measured the implications of dropping the existing Flock shared locks on higher load setups? I mean we basically serialize **all file cache accesses**, silently and in a **minor** version?!

hlubek:
This has a very high risk assigned to me. We needed many versions to get locking at least to the point where random requests didn't blow up our caches. I'd not support this as the new default without extensive testing (which is not realistic IMHO). We can however include the `SemaphoreLockStrategy` as a new strategy.

albe:
I'd suggest finally switching the Cache FileBackends to use locking on the cache files directly as already suggested in our long discussions when trying to solve the race condition issues with external flock files.

Once the FileBackend is decoupled from this, we're basically free to deprecate shared locking and can use this generic class much more freely in other places where exclusive locking is required (compile time anyone?).

I'll get a PR up soon, as it is much less work than I ever thought and it should really finally solve the lock file issue we still suffer from.

albe:
How about this: https://github.com/albe/flow-development-collection/commit/1a54705be46a76f9bc2eef5d4d0a580002674103

I reverted *FileBackend to old `new Lock(...)` usage (to keep it simpler for merging #417 and the split-cache) and removed the Lock Factory class in favor of `Lock::acquire(...)`. Also fixed the leftover code and added (copied) tests for the new Strategy.

kitsunet:
So can we keep the file backend as is for now and introduce the rest anyway for LTS?

albe:
Can't push to Dominiques fork, so should I create a new PR with my commit above?

kitsunet:
Good question, but I like @dfeyer could you have a look? Otherwise I would say new PR and we merge that then.

dfeyer:
@albe You have push access now

dfeyer:
I've restarded the travis build as the problem seems on the travis side

kitsunet:
Will check as well. Looks good to me though.

kitsunet:
Looks to me as if that ends in a deadlock somewhere...

kitsunet:
Note: changed base to `3.3` branch to get this into the release still.

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Lock.php
```diff
@@ -13,17 +13,10 @@
 
 /**
  * A general lock class.
- *
- * @api
```
albe: Just noticed. Maybe this should stay now that the `acquire` method is in this class

Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/Lock.php
```diff
@@ -41,17 +34,22 @@ class Lock
     /**
      * @var boolean
      */
-    protected $exclusiveLock = true;
+    protected $exclusiveLock;
```
albe: That's not even used anywhere... guess we can safely get rid

dfeyer:
Let's don't merge this one. And thinks twice about how we handle lock (why, ...). I just pushed an old local branch #436 that replace our own code, by an external library and add strategy for Flock (like our, never remove lock file), Redis and Memcached, but only with exclusive lock.

dfeyer:
Now https://github.com/neos/flow-development-collection/pull/436 also support Semaphore based strategy, if the system is correctly configured (with enough semaphore, it should work fine)

dfeyer:
I close this one, we can reopen later if needed

PR #259, opened 2016-02-15
BUGFIX: Rendering docs produced errors and warnings

DanielSiepmann:
Removed all possible errors and warnings from documentation. This will
allow new errors to be visible as soon as they occur and to reduce the
overall errors.

Some errors lead to non working links or syntax highlighting.

Also fixed indentation at some places. As the indentation wasn't the
same everywhere.

PR #263, opened 2016-02-16
TASK: Fix rST errors and warnings for 3.1

DanielSiepmann:
I've tried to fix as much as possible.

But I can't resolve some errors, e.g. in ChangeLog some labels exist twice as there were two pull requests for the same topic leading to the same title.

Beside that, there are two code blocks I can't fix with my knowledge and without breaking the code.


DanielSiepmann:
First commit is cherry-picked from #259.
Rest is documented in each commit itself.

kdambekalns:
:+1: (the Travis CI failure can be ignored for this PR)

PR #266, opened 2016-02-17
TASK: Do not initialize ReflectionService on shutdown if not needed

hlubek:
The ReflectionService would be still initialized before having an early
return in the saveToCache() method. This change makes sure to not
initialize the ReflectionService which results in less cache entries
being loaded and therefore less memory consumption and I/O.

NEOS-571 #comment Optimize ReflectionService shutdown


kitsunet:
Nice! :+1:

kdambekalns:
:+1: by reading

albe:
:+1: by reading, makes sense

dfeyer:
:+1:

PR #267, opened 2016-02-17
TASK: Don't throw exception on broken resource when publishing

dfeyer:
This change replace the exception by a log message.


dfeyer:
Maybe this one can be target to a lower branch, even if it's not really a bugfix

And maybe we can keep track of those error and output them at the end of resource publishing ?

kdambekalns:
Nice one! Agree to both suggestions (branch and collect errors).

dfeyer:
I close this on a open a new one on branch 2.1

PR #268, opened 2016-02-18
TASK: Handle missing data as a soft error during resource publishing

dfeyer:
This change log missing data during resource publishing. By default all notifications are logged in the system log. This change also introduce ```MessageCollector``` as the default API for resource storages and targets to provide error messages during publishing.


dfeyer:
:+1: Tested on 3.0, 3.1 and master

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Publishing/Notification.php
```diff
@@ -0,0 +1,78 @@
...
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use TYPO3\Flow\Annotations as Flow;
+
+/**
+ * Notification
+ */
+class Notification
```
hlubek: Why not use the Flow Message (Error) class here?

dfeyer: Good idea

hlubek:
I really like to improve that part because it caused us a lot of trouble already (failed deployment / update, debugging through the database, dumping SHA1s, etc.).

Question is, if we need to implement two new classes that are concerned with logging and messages / results especially for the resource publishing. I think we can implement that in a more lightweight way by just returning an array of `TYPO3\Flow\Error\Message` or a `TYPO3\Flow\Error\Result` object from `publishCollection` (and `publishResource` respectively) without breaking the interface (yeah PHP for not forcing return types here). We could handle a `null` return value as no errors and then just print them in the command controller. Would feel easier for me.

dfeyer:
I see other place where notification can be useful (like timing monitoring in cloud deployment, ...), and having a central class is way more easy to manage them. 

I'm OK to replace the Notification class by resusing the Flow Message class family, so we just introduce a single new class, the NotificationCollection or MessageCollector if we rename it.

hlubek:
Yes, that would also work for me, but Notification is just too similar to the base `Message` class IMHO

dfeyer:
:+1:

TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
```diff
@@ -86,6 +94,14 @@ public function publishCommand($collection = null)
                 $this->outputLine('Publishing resources of collection "%s"', array($collection->getName()));
                 $collection->publish();
             }
+
+            if ($this->messageCollector->hasNotification()) {
+                $this->outputLine();
+                $this->outputLine('The system has generate some notifications, please check if you can solve some issue:');
```
kdambekalns: make that
> The system has generated some notifications, please check if you can solve those issues:

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Publishing/MessageCollector.php
```diff
@@ -0,0 +1,106 @@
...
+ * Notification Collector
+ *
+ * @Flow\Scope("singleton")
+ */
+class MessageCollector
+{
+    /**
+     * @var \SplObjectStorage
+     */
+    protected $backends;
```
kdambekalns: `$backends` is a weird name. This contains messages, so why not name it `$messages`?

dfeyer:
:+1:

TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
```diff
@@ -86,6 +94,14 @@ public function publishCommand($collection = null)
                 $this->outputLine('Publishing resources of collection "%s"', array($collection->getName()));
                 $collection->publish();
             }
+
+            if ($this->messageCollector->hasNotification()) {
+                $this->outputLine();
+                $this->outputLine('The system has generate some notifications. You can check and probably fix the integrity of the resource registry by using the resource:clean command.');
```
robertlemke: A little typo ("generate*d*"), but maybe in general rephrase like this?

"The resources were published, but a few inconsistencies were detected. You can check and probably fix the integrity of the resource registry by using the resource:clean command."

robertlemke:
:+1: by reading - I would have tested it, too, but somehow my local setup is pretty much broken while trying to update Neos after the Flow split while being on the train.

dfeyer:
@robertlemke Can you test this one, it's really nice if we can merge it for 2.2

aertmann:
:+1: by reading

robertlemke:
Tested it, works fine and is a great improvement!

Just two thoughts:

- since it's a `MessageCollector`, wouldn't the name of the method then have to be `hasMessages()` (also note the plural)?
- are we sure enough about this so we can make it part of the API? I think it's only really useful, if also other resource storages and targets can provide error messages through that mechanism

dfeyer:
> since it's a MessageCollector, wouldn't the name of the method then have to be hasMessages() (also note the plural)?

Agree, will push a new version later today

> are we sure enough about this so we can make it part of the API? I think it's only really useful, if also other resource storages and targets can provide error messages through that mechanism

I think yes, it can be nice to have the MessageCollector a single API to collect any kind of notification, but maybe we can discuss a little bit more how the handle Message severity, better reporting, flushing non read message to the system log, ... before making it parts of the public API. I'm fine to wait 3.3 release to make it part of the API

kdambekalns:
We wait for @dfeyer to push the changes around the `MessageCollector` and the give it another review an merge it, I'd suggest.

dfeyer:
Sorry for the hug delay, I lost control on my planning during the last two weeks.

@robertlemke All public method are now annotated as API, it just make sense.

Ready to merge and upmerge up to 3.1 and master

TYPO3.Flow/Classes/TYPO3/Flow/Resource/Publishing/MessageCollector.php
```diff
@@ -0,0 +1,112 @@
...
+     * @param callable $callback a callback function to process every notification
+     * @return void
+     * @api
+     */
+    public function flush(callable $callback = null)
+    {
+        foreach ($this->messages as $message) {
+            /** @var Message $message */
+            $this->messages->detach($message);
+            $this->systemLogger->log('ResourceMessageCollectorMessage: ' . $message->getMessage(), $message->getSeverity());
```
johannessteu: Shouldn't this be `RessourcePublishingMessageCollectorMessage`?

dfeyer: RessourcePublishingMessage should be OK ;)

johannessteu: Yeah, at least something with Publishing :-P

johannessteu:
👍

PR #269, opened 2016-02-23
TASK: Use composer autoloader as fallback to Flow autoloader

kitsunet:
The Flow autoloader relied on much of the autoload information
that composer provides anyway. With this change we initialize
the composer autoloader first thing in the bootstrap and later
our own on top of that. That means the responsibility is kept
in the composer autoloader and we can get rid of many early include
statements which the composer autoloader fullfills instead.


dfeyer:
:+1: make sense, good one

PR #270, opened 2016-02-23
TASK: Split utilities to a separate package

kitsunet:
With this change all Utility classes that can be used
independently of Flow are moved to a separate package.


kitsunet:
Note this needs #269 because the removed requires in there would fail with this change as the classes are moved.

kitsunet:
will probably split more, don't merge

PR #273, opened 2016-02-25
BUGFIX: Update tagFormats to match defined severities

langeland:
Uses the severities descriped in \TYPO3\Flow\Log\LoggerInterface

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Log/Backend/AnsiConsoleBackend.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Log/Backend/AnsiConsoleBackend.php
@@ -58,12 +58,14 @@ class AnsiConsoleBackend extends ConsoleBackend
     {
         parent::open();
         $this->tagFormats = array(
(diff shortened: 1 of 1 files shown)


kdambekalns:
I'd like to see `info` included to make it B/C and then would be fine with adding to 3.0 even…

kitsunet:
Info is included?

kdambekalns:
Sorry, I meant `success`

langeland:
Hmm.. As such I have no problem with adding `success`, but from what I can see there's no such thing as `success` in the rfc3164. See page 8 Table 2.

And if I look at: `\TYPO3\Flow\Log\Backend\ConsoleBackend::open` then `success` can never be called.

kdambekalns:
That's what you get from purely reading a change: I just saw something that was there, was now gone. If it can never end up there, that obviously doesn't matter at all.

:+1: after reading the related code as well. Still, that looks like a bugfix and should even go into 2.3…

kitsunet:
looks good :+1:

PR #274, opened 2016-02-25
TASK: Don't set a default host for persistence backend

johannessteu:
Flow won't work on a server without a running mysql service because a default host ist configured. This fix will set the default host to an empty string so flow does also run without a running  mysql service.

FLOW-379 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
@@ -97,6 +97,17 @@ class EntityManagerFactory
         $config->setProxyNamespace('TYPO3\Flow\Persistence\Doctrine\Proxies');
         $config->setAutoGenerateProxyClasses(false);
 
+        // Set default host to 127.0.0.1 if there is no other host configured and at least one other necessary option
+        if ($this->settings['backendOptions']['host'] == '') {
+            if (
+                $this->settings['backendOptions']['dbname'] != '' ||
+                $this->settings['backendOptions']['user'] != '' ||
+                $this->settings['backendOptions']['password'] != ''
+            ) {
+                $this->settings['backendOptions']['host'] = '127.0.0.1';
+            }
+        }
+
         // The following code tries to connect first, if that succeeds, all is well. If not, the platform is fetched directly from the
         // driver - without version checks to the database server (to which no connection can be made) - and is added to the config
         // which is then used to create a new connection. This connection will then return the platform directly, without trying to
--- a/TYPO3.Flow/Configuration/Settings.yaml
+++ b/TYPO3.Flow/Configuration/Settings.yaml
@@ -215,7 +215,7 @@ TYPO3:
       # Options for the default Doctrine persistence
       backendOptions:
         driver: 'pdo_mysql'
-        host: '127.0.0.1'
+        host: ''
         dbname: NULL
         user: NULL
         password: NULL


bwaidelich:
is that completely backwards compatible? So, does it still work if you use a mysql server and don't explicitly set the host?

johannessteu:
ah ok, you are talking about configurations like: 
```
TYPO3:
  Flow:
    persistence:
      backendOptions:
        dbname: 'flow'
        user: ''            
        password: ''         
```

Hmm, then check in EntityMangerFactory if one of the values is set and host is missing then default host to `127.0.0.1`?

johannessteu:
I'm not sure if the `EntityManagerFactory` is the right place @bwaidelich ?!

kitsunet:
No technically not, because the config is not doctrine specific. It *could* be used by other persistence backends in theory.

kitsunet:
So might be better to fix that in master only?

johannessteu:
@kitsunet in master but the code itself is ok? I'd create a new PR then

TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
```diff
@@ -97,6 +97,17 @@ public function create()
...
+        if ($this->settings['backendOptions']['host'] == '') {
+            if (
+                $this->settings['backendOptions']['dbname'] != '' ||
+                $this->settings['backendOptions']['user'] != '' ||
+                $this->settings['backendOptions']['password'] != ''
+            ) {
+                $this->settings['backendOptions']['host'] = '127.0.0.1';
+            }
+        }
+
```
bwaidelich: We should always use type safe comparison according to the [CGL](http://flowframework.readthedocs.org/en/latest/TheDefinitiveGuide/PartV/CodingGuideLines/PHP.html#php-in-general). And `password` could be left empty intentionally (not a good idea, but...).
Also, why nest the conditions?
Couldn't something like this work out:
```
if (empty($this->settings['backendOptions']['host']) && !empty($this->settings['backendOptions']['dbname'])) {
  $this->settings['backendOptions']['host'] = '127.0.0.1';
}
```
?

johannessteu: The idea was to check if at least one of them is set. But `dbname` is a must have in any kind of persistence backend? Just thinking if there is one wich could default to a database. If there's none that would be enough, sure!

TYPO3.Flow/Configuration/Settings.yaml
```diff
@@ -215,7 +215,7 @@ TYPO3:
       # Options for the default Doctrine persistence
       backendOptions:
         driver: 'pdo_mysql'
-        host: '127.0.0.1'
+        host: ''
```
aertmann: a NULL value would be more suiting than an empty string here, it clearly indicates an unset value unlike a empty string

TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
aertmann: at least the code needs to be more robust, this could easily lead to issues with missing array keys and unsetting to null instead of an empty string..

Example:
```
if (empty($this->settings['backendOptions']['host']) &&
  (
    !empty($this->settings['backendOptions']['dbname']) ||
    !empty($this->settings['backendOptions']['user']) ||
    !empty($this->settings['backendOptions']['password'])
  )
) {
  $this->settings['backendOptions']['host'] = '127.0.0.1';
}

johannessteu:
OK, to sum it up
- change to master
- `host: ~`
- just check for db name && host 

I'll adjust and create a new PR this evening

TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
kitsunet: I think this change in itself is a good idea but the above code feels too magic. If I don't set a host but everything else the host will be magically 127.0.0.1, I am not really happy with that yet. I would strongly suggest to deprecate this behavior right away and document that the host MUST be set beginning with the next major Flow version.

bwaidelich:
@johannessteu did you already get around recreating the PR? I can't find it

johannessteu:
Ah, sorry. Didn't made it yet. I'll try to do that tomorrow on my rest day :D. 
Currently in holiday right now. But maybe i find some time to do it. Fell free to take over if you ok with it. Otherwise i will do so. My sum up is ok so far? Or should i respect anything else in a new PR @bwaidelich ?

johannessteu:
New PR #308

PR #277, opened 2016-03-02
BUGFIX: Fix regression with frozen reflection service

hlubek:
The commit 409a45aaf9550b2df790c9730b7f002bd609624b introduced a change
for early return inside the reflection service that did not catch a case
where the context was not initialized from the environment.


kitsunet:
:+1: Makes sense, I guess another round of refactorings is needed. This kind of assignment isn't really useful and leads to this kind of bugs. We should either get rid of environment and make sure that context is always there or remove `$this->context`. But that's for another change.

PR #278, opened 2016-03-03
TASK: Split ObjectAccess utility off

kitsunet:
Separates the ObjectAccess utility to a new package and cuts
dependencies to Flow.

Last dependency is the TypeHandling utility which first needs to
be split as well.

### DIFF
--- a/Neos.PropertyAccess/Classes/TYPO3/Flow/Reflection/Exception/PropertyNotAccessibleException.php
+++ b/Neos.PropertyAccess/Classes/TYPO3/Flow/Reflection/Exception/PropertyNotAccessibleException.php
@@ -2,7 +2,7 @@
 namespace TYPO3\Flow\Reflection\Exception;
 
 /*
- * This file is part of the TYPO3.Flow package.
+ * This file is part of the Neos.PropertyAccess package.
  *
  * (c) Contributors of the Neos Project - www.neos.io
  *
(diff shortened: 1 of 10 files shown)


bwaidelich:
Hey! Nice one.
What about a more general package that contains `TypeHandling` and the `ObjectAccess` parts?

kitsunet:
Yep very good, I already thought about it, but didn't do it yet. IMHO would make sense and isn't too far away in terms of responsibilities.

bwaidelich:
Perfect! Do you have an idea for a good package name? Maybe just `neos/typehandling` or `neos/types` or `neos/objects`..?  mh

kitsunet:
good question. I find neos/objects too generic even if it would fit well. neos/typehandling yes, but that neglects the property get/set story somehow.
I think that was one of the points that let me not do it right away ;)
Lets think a bit more, maybe someone else has a good idea...

dfeyer:
Has this package contain mainly ObjectAccess, _Neos.ObjectAccess_ sounds OK for me

bwaidelich:
But it would also contain all the `TypeHandling` things, like `parseType()`, `isSimpleType()`, `getTypeForValue()`, ...
I think "ObjectAccess" doesn't really express a lot. Maybe `neos/object-utilities`?

kitsunet:
@bwaidelich see #279 :)

PR #279, opened 2016-03-06
TASK: Create ObjectUtilities package

kitsunet:
This package contains the ObjectAccess and TypeHandling utilities from Flow.
Both are related and are useful if working with objects and types.


kitsunet:
Alternative to https://github.com/neos/flow-development-collection/pull/278

bwaidelich:
Nice one, thanks for taking care!
A couple of questions:
* Could it be an issue that the new Exceptions don't extend `\TYPO3\Flow\Exception` any longer? They have some special features regarding custom error handling and existing try/catches might fail now
* What about adjusting the namespace too? Now this package contains `TYPO3\Flow\Reflection`, `TYPO3\Flow\Utility` and `Neos\ObjectUtilities` classes. But this would be the chance to unify that to `Neos\ObjectUtilities` and only keep the existing FQN for B/C !?
* ..and related to that: What about using PSR-4 for those new classes?

kitsunet:
@bwaidelich I want to separate the split from the unification and deprecation of existing packages, but yes, that would be the next step and of course those should then use PSR-4 :)

I don't think it's a huge issue in regards to the exceptions. Yes, someone might use the file utility and only catch \TYPO3\Flow\Exception and that would potentially break, but I consider that a very low risk chance.

BTW I closed this and continued with a https://github.com/neos/flow-development-collection/tree/splits branch to make this easier to review.

If you wanted to check that just require `"dev-splits as dev-master"` in your root composer.json and `composer update`.

For us here this works, tests run as well.

PR #280, opened 2016-03-06
Extend RenderViewHelper to accept arguments from a child ArgumentViewHelper

vjanoch:
[FEATURE] RenderViewHelper accepts arguments from a child ArgumentViewHelper


TYPO3.Flow/Classes/TYPO3/Flow/Error/Debugger.php
```diff
@@ -175,7 +175,7 @@ protected static function renderObjectDump($object, $level, $renderProperties =
         $scope = '';
         $additionalAttributes = '';
 
-        if ($object instanceof \Doctrine\Common\Collections\Collection) {
+        if ($object instanceof \Doctrine\Common\Collections\Collection || $object instanceof \ArrayObject) {
```
bwaidelich: This might make sense, but it's not related to this feature so I'd suggest to skip it or extract it to its own PR

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/ArgumentViewHelper.php
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use TYPO3\Fluid\Core\ViewHelper\AbstractViewHelper;
+
+/**
+ * "ARGUMENT" -> only has an effect inside of "RENDER". See Render-ViewHelper for documentation.
```
bwaidelich: A little usage example would make sense (similar to the ones in other ViewHelpers)

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/RenderViewHelper.php
```diff
@@ -101,9 +137,34 @@ public function render($section = null, $partial = null, $arguments = array(), $
     }
 
     /**
+     * Loads the "argument" child nodes.
+     *
+     * @param $arguments
```
bwaidelich: `@param array $arguments`

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/RenderViewHelper.php
```diff
@@ -101,9 +137,34 @@ public function render($section = null, $partial = null, $arguments = array(), $
...
+     *
+     * @return array
+     */
+    protected function loadArgumentChildrenIntoArguments($arguments)
+    {
+        foreach ($this->childNodes as $childNode) {
+            if ($childNode instanceof ViewHelperNode
+                && $childNode->getViewHelperClassName() === \TYPO3\Fluid\ViewHelpers\ArgumentViewHelper::class
```
bwaidelich: If the Namespace is imported this could be put into one line:

```
if ($childNode instanceof ViewHelperNode $childNode->getViewHelperClassName() === ArgumentViewHelper::class) {
```

bwaidelich:
This looks good already, thanks for this! I just added some nitpick-comments.

Apart from that it was [discussed][1] to split up this ViewHelper into `render.section` and `render.partial` and to keep `render` only for backwards compatibility.

But IMO that could be done in a separate PR, too

  [1]: https://neos-project.slack.com/archives/flow-general/p1455618538000136

TYPO3.Flow/Classes/TYPO3/Flow/Error/Debugger.php
vjanoch: Hi Bastian,

this part has been merged already. That was a suggestion I made a few days/weeks ago. I think I messed it up by pulling the latest changes from neos and pushed it to my repo. Next time I will take care to provide a clean pull request.

Sorry!

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/RenderViewHelper.php
vjanoch: I totally agree. Additionally it should be

```
protected function loadArgumentChildrenIntoArguments(array $arguments)
```

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/RenderViewHelper.php
vjanoch: :+1:

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/RenderViewHelper.php
bwaidelich: Yes, that's even better but the type hint is missing in `loadSettingsIntoArguments()` too, so I didn't complain ;)

TYPO3.Flow/Classes/TYPO3/Flow/Error/Debugger.php
bwaidelich: :+1: no worries

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/RenderViewHelper.php
vjanoch: Hmm, we don't even need to import the namespace, because it is the same in both classes :-)

vjanoch:
@bastian: Do you have an idea why the build does not pass? It seems that the described error is not caused from the changes made within this PR. Could you please have a look at it?

kitsunet:
@vjanoch There you go. Restarted the build, works! :+1:

albe:
Really looks nice and I can relate to the use case. Probably needs a little squash to tighten the commits. Anything else keeping this from getting merged?

kitsunet:
I would like to block this for now as the standalone fluid integration progresses nicely and I don't want to be behind in features please :)

mficzel:
This is a generic fluid-feature that makes much sense. 

The native fluid has landed in flow-master some time ago and all that remains of fluid in flow is the Neos.FluidAdapter package. I think this feature should go to the main fluid repo here https://github.com/TYPO3/Fluid.

albe:
So I guess this can be closed here, since it is now something that should go into Fluid. If anyone feels like this should be part of Flow still, feel free to reopen.

PR #287, opened 2016-03-07
TASK: extracted Neos.Utility.Arrays and Algorithms from TYPO3.Flow

christoph-daehne:


kitsunet:
This was not exactly what I had in mind. I wanted to put Arrays and PositionalArraySorter into one package. See https://github.com/neos/flow-development-collection/pull/288

PR #289, opened 2016-03-08
TASK: Memcached cache backend compatibility with memcached

aertmann:
Adds support for the "memcached" extension in addition to the existing
support for the "memcache" extension.

"memcached" is a newer version with several improvements and often both
extensions aren't available simultaneously.

FLOW-440 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/MemcachedBackend.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/MemcachedBackend.php
@@ -13,9 +13,9 @@ namespace TYPO3\Flow\Cache\Backend;
 
 
 /**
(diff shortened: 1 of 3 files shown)


aertmann:
not sure if this should be backported to stable versions, it's very low risk and a useful improvement

:+1: verified using the "memcached" extension

should be very safe to merge, since nothing should have changed for the "memcache" extension except if both are installed "memcached" is used since it's better

@kdambekalns @dfeyer maybe you have some interest in this area? thanks

kdambekalns:
Very nice. Read through it and it seems indeed fit for inclusion in 2.3 and up IMHO.

aertmann:
Pushed against 2.3 #296

kitsunet:
Looks good and agree, could be backported.

PR #297, opened 2016-03-09
BUGFIX: Accept multiple autoload paths for Flow packages

kitsunet:
Flow packages with a PSR-4 and a PSR-0 autoload path will only search for a ``Package.php``
file in the PSR-4 class path, which breaks if the ``Package.php`` is actually in the PSR-0 path.
With this change Flow will search in both paths with a preference on the PSR-4 path as PSR-0
is deprecated. Additionally only Flow package types will be searched for ``Package.php`` files.
If both paths contain a ``Package.php`` file Flow will throw an Exception because we cannot
know which one to load.


kdambekalns:
Looks good, seen it pass. :)

PR #300, opened 2016-03-11
TASK: Fix Configuration cache usage

kitsunet:
The Objects configuration was never checked against the loaded
configuration cache but still saved sometimes. If a specific configuration type
was saved depended on the fact that it was loaded at the time of writing the
configuration cache. We now make sure that Objects configurations are also used
if they are cached and that we load all configuration types before writing the
cache.


kitsunet:
@kdambekalns @skurfuerst @bwaidelich 

Stumbled on that while splitting packages. Basically using Objects configuration in runtime is impossible because it will be build on every request. Also means it could eventually take multiple requests to have a configuration cache with all types.

kitsunet:
:+1:

kdambekalns:
:+1:

dfeyer:
:+1:

PR #302, opened 2016-03-13
TASK: Monitor only classes of packages with registered objects

kitsunet:
With this change the amount of monitored class files is massively reduced
by only monitoring class files of packages that have active object
configurations.
This improves request time in development context noticeably.


albe:
:+1: by reading, nice one! Would it maybe even be possible to statically compile the getListOfPackagesWithConfiguredObjects method?

kitsunet:
I thought about that as well but then realized it is pointless because static compile happens only in Production but file monitoring doesn't happen in production, so makes no sense ;)

albe:
Haha, indeed! :)

kitsunet:
anyone? Really nice performance boost for development @kdambekalns maybe? :)

kdambekalns:
Nice, shaved some 80-100 ms from my request (down to ~540 ms, on PHP 5.6 with Xdebug enabled).

PR #303, opened 2016-03-14
BUGFIX: Log throwable errors in PHP 7

aertmann:
Throwable errors (``TypeError``, ``ParseError``, ..) introduced in PHP 7.0
are not logged in the system log. Which means these errors can only
be shown using the debug exception handler.

To fix this a new logger interface ``ThrowableLoggerInterface``
that extends the existing ``LoggerInface`` is introduced,
which adds supports throwables in addition to exceptions.

To support PHP7 in custom logger interfaces, the new interface needs
to be implemented. Otherwise throwables won't get logged.

FLOW-441 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Log/EarlyLogger.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Log/EarlyLogger.php
@@ -17,7 +17,7 @@ namespace TYPO3\Flow\Log;
  *
  * @api
  */
-class EarlyLogger implements SystemLoggerInterface
+class EarlyLogger implements ThrowableLoggerInterface
 {
     /**
      * @var array
@@ -28,6 +28,11 @@ class EarlyLogger implements SystemLoggerInterface
      * @var array
      */
     protected $exceptions = array();
+
+    /**
+     * @var array
+     */
+    protected $throwables = array();
 
     /**
      * Adds a backend to which the logger sends the logging data
@@ -63,6 +68,7 @@ class EarlyLogger implements SystemLoggerInterface
     {
         $this->logEntries = array();
         $this->exceptions = array();
+        $this->throwables = array();
     }
 
     /**
@@ -96,6 +102,19 @@ class EarlyLogger implements SystemLoggerInterface
     }
 
     /**
+     * Writes information about the given exception into the log.
+     *
+     * @param \Throwable $throwable The exception to log
+     * @param array $additionalData Additional data to log
+     * @return void
+     * @api
+     */
+    public function logThrowable(\Throwable $throwable, array $additionalData = array())
+    {
+        $this->throwables[] = func_get_args();
+    }
+
+    /**
      * Replays internal logs on provided logger. Use to transfer early logs to real logger when available.
      *
      * @see \TYPO3\Flow\Package\PackageManager
@@ -114,8 +133,13 @@ class EarlyLogger implements SystemLoggerInterface
             $logger->log('[Done replaying logs from instance of EarlyLogger.]');
         }
         if (count($this->exceptions) > 0) {
-            foreach ($this->logEntries as $logEntry) {
-                call_user_func_array(array($logger, 'logException'), $logEntry);
+            foreach ($this->exceptions as $exception) {
+                call_user_func_array(array($logger, 'logException'), $exception);
+            }
+        }
+        if (count($this->throwables) > 0 && $logger instanceof ThrowableLoggerInterface) {
+            foreach ($this->throwables as $throwable) {
+                call_user_func_array(array($logger, 'logThrowable'), $throwable);
             }
         }
         if ($resetLogs === true) {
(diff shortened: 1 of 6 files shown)


aertmann:
Bugs me having to push this one, but it's pretty bad.

@kdambekalns @robertlemke @kitsunet @bwaidelich: please respond on this one

robertlemke:
@aertmann thanks for taking care of this one.

I actually created almost exactly the same patch when I was dealing with PHP 7 compatibility last year, but then didn't push it because it changes the `LoggerInterface`.

As far as I can see, all we can do right now is to include as much code from your PR as possible (where it still makes sense) but leave out the interface change and put it on the list for Flow 4.0.

kitsunet:
Shoudl be cool if we just skip the Interface change and add a comment there about Throwable and that the interface will require it for next major.

aertmann:
@robertlemke @kitsunet: that would be great, but how do you fix this without touching the interface? to do that we need to implement a new method, is that what you're suggesting?

kitsunet:
You are right, that's not really possible :/

Mmm, what do... Just to have those logged does not really justify the breaking change right now, on the other hand there will probably not many implementations of the logger interface around...

aertmann:
IMO it actually does to a certain extent, not having them logged is likely going to cause a frustration for quite some people. As an example, I spend a quite some time on a wild goose chase debugging if there was anything wrong with the server setup since some 500's didn't give end up in any log. You don't expect Flow to have such a bug (silencing errors) and only really suspected it after remembering someone mentioning something similar in Slack. I'm sure that person wasted some time figuring out the same.

Now whether or not fixing it by breaking the interface or adding a new method that we add to the default logger is another thing.

kdambekalns:
Thinking aloud… What about adding a new interface and implementation (e.g. `ThrowableLoggerInterface` extending the `LoggerInterface`) and then on PHP7 it can be configured as the default implementation by the user, and the default would be the unchanged interface?

robertlemke:
@kdambekalns yes, something like that is feasible. But I really don't see the justification for releasing Flow 4.0 and leaving our release plan for only this change. I actually consider it a (useful) feature and not a bug fix, because it provides functionality which wasn't there before.

kitsunet:
I tend to agree about the major release, but we need a solution, this is not just about some nice functionality that you can live without. @kdambekalns I don't see how that would work with two interfaces but I'd love to be proven wrong. I expect you cannot implement two interfaces with conflicting method signatures... Or is that possible if you extend an interface? Strikes me as odd if so.

bwaidelich:
I'm a bit puzzled. The [related change](https://github.com/neos/flow-development-collection/commit/8f051a1f4d21f8ba995dff4ae89db66d535a68fa) was as breaking, requiring adjustments to all custom error handler implementations. Why is it a different situation here?

I'd say: Given the modified behavior of the ExceptionHandler it would be OK to make the exception in this case if it leads to less frustration. And it's probably less frustrating having to adjust the signature of your implementation than finding out that errors are not logged..

Yet another thought: Would it be possible (and feasible) to convert `Throwable`s into `Exception`s so that the regular exception handling could kick in?

aertmann:
Hey everyone, thanks for the input. Gave it another go, not breaking and fallback to converting into exceptions for non supporting loggers.

TYPO3.Flow/Classes/TYPO3/Flow/Log/EarlyLogger.php
```diff
@@ -114,8 +133,13 @@ public function replayLogsOn(SystemLoggerInterface $logger, $resetLogs = true)
             $logger->log('[Done replaying logs from instance of EarlyLogger.]');
         }
         if (count($this->exceptions) > 0) {
-            foreach ($this->logEntries as $logEntry) {
-                call_user_func_array(array($logger, 'logException'), $logEntry);
+            foreach ($this->exceptions as $exception) {
+                call_user_func_array(array($logger, 'logException'), $exception);
```
aertmann: fixed this one along the way..

robertlemke:
@aertmann thanks for the extra effort to make this backwards compatible. My only concern is, if this works for real with PHP <7, not only with the tests (because I'm not sure if PHP really doesn't mind the \Throwable type hints). I can't test it right now, but if someone of you really really tested this with PHP 5.6 manually, I'll gladly give my +1

robertlemke:
something also worth considering is a polyfill approach, like https://packagist.org/packages/symfony/polyfill-php70 – any opinions?

aertmann:
PHP doesn't mind type hint with non-existing class names, see https://3v4l.org/pDZaj

however travis fails since it's doing some AOP tests using the logger, so it's more Flow that doesn't handle missing classes in this case :)

aertmann:
okay fixed the tests, should be good to go

aertmann:
@kitsunet: would be great if you could test it on your setup since I don't have a running 5.5 setup

kitsunet:
Works fine on PHP5.5 :+1:

aertmann:
@robertlemke @kdambekalns: wanna change to a :+1:? :)

robertlemke:
@aertmann I gladly do (did), but I didn't get a chance yet to really test it. Trusting you guys ...

kdambekalns:
@aertmann now it's only your own down vote left. :)

aertmann:
fixed regression 09948e2b96999c9036ad9c2caa39e1e4b86eec42

PR #304, opened 2016-03-14
BUGFIX: If query is null the condition break

dfeyer:
This change trim the query string, without this change the condition to build the request URI is wrong when the query is null and the URI contains a question mark without request arguments.


kitsunet:
I would rather prefer to cast to string there, it might be a rare case but a space might be significant at some point

albe:
Good catch @kitsunet +1 that string cast

PR #307, opened 2016-03-15
BUGFIX: Remove duplicate registration of ClassLoader

kitsunet:
The Flow class loader is registered in doctrines annotation reader twice.
First in ``Booting\Scripts::registerClassLoaderInAnnotationRegistry()``
then again in the ReflectionService. As this is unnecessary and is not
directly related to the ReflectionService the second registration is
removed with this change.


kitsunet:
replaces #305

kitsunet:
Merging with 3 upvotes.

PR #308, opened 2016-03-22
TASK: Don't set a default host for persistence backend

johannessteu:
Flow won't work on a server without a running MySQL service because a default host is configured.
This fix will set the default host to an empty string so Flow also runs without a running MySQL service.

FLOW-379 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
@@ -117,6 +117,11 @@ class EntityManagerFactory
         $config->setProxyNamespace('TYPO3\Flow\Persistence\Doctrine\Proxies');
         $config->setAutoGenerateProxyClasses(false);
 
+        // Set default host to 127.0.0.1 if there is no host configured but a dbname
+        if (empty($this->settings['backendOptions']['host']) && !empty($this->settings['backendOptions']['dbname'])) {
+            $this->settings['backendOptions']['host'] = '127.0.0.1';
+        }
+
         // The following code tries to connect first, if that succeeds, all is well. If not, the platform is fetched directly from the
         // driver - without version checks to the database server (to which no connection can be made) - and is added to the config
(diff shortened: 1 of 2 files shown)


johannessteu:
@bwaidelich

kitsunet:
I have another show stopper I fear or maybe not? I do think that you can set a default host in php.ini? I mean what if you want to use that, basically you cannot if we include this code... This is probably an edge case but I see it happen...

bwaidelich:
I don't know why the travic CI build fails, but apart from that it looks good by reading. thx!

dfeyer:
> I have another show stopper I fear or maybe not? I do think that you can set a default host in php.ini? I mean what if you want to use that, basically you cannot if we include this code... This is probably an edge case but I see it happen...

Really egde case and sounds event more than edge ;) i don't remember an application configured based on the default host configured in the php.ini

kitsunet:
Yeah, I won't block it because of that, just wanted to mention it to see if someone reacts and puts up a hand...

johannessteu:
Does behat uses some other host configuration then `127.0.0.1` ?

kitsunet:
I wouldn't know where from. I mean it uses Testing/Behat context but I don't see that we have any config for it.

johannessteu:
Closed/Reopend to trigger the tests again

kitsunet:
ok, currently tests are broken due to https://jira.neos.io/browse/FLOW-447
Needs retest after but is probably fine...

kitsunet:
restarting travis

bwaidelich:
what's happening here?

kdambekalns:
Well, pretty obvious, no?

> Database schema creation has been SKIPPED, the driver and host backend options are not set in /Configuration/Settings.yaml.

gerhard-boden:
:+1: by reading

albe:
needs adjustment in DoctrineCommandController right?

aertmann:
needs to be set for the Testing context, https://github.com/neos/flow-development-collection/blob/master/TYPO3.Flow/Configuration/Testing/Settings.yaml#L46

kdambekalns:
No, needs to be adjusted in `Build/BuildEssentials/TravisCi/SetupDatabase.sh` :)

PR #309, opened 2016-03-24
TASK: Adjust .editorconfig to use 4 spaces for JS files

kdambekalns:


kdambekalns:
@grebaldi @Inkdpixels @aertmann

aertmann:
guess we can merge this, although there's only 6 js files all from an external library https://github.com/neos/flow-development-collection/search?l=javascript

PR #312, opened 2016-03-25
BUGFIX: orphanRemoval must be a true integer

lorenzulrich:
Doctrine expects a true integer instead of a string (true in double quots) for the orphanRemoval property.


kitsunet:
Thanks! But this should go into the lowest branch, so 2.3 actually. Thanks for taking care!

PR #315, opened 2016-03-27
TASK: Remove Environment dependency of ConfigurationManager

kitsunet:
The ConfigurationManager needs only the temporary path to store cached
configurations so we can inject just that and get rid of the dependency
to the ``Environment`` Utility.


albe:
:+1: by  reading. Nice simple dependency refactoring

PR #316, opened 2016-03-27
TASK: Initialize MediaTypeConverter only if request has body content

kitsunet:
Instead of trying to convert an empty string we return early in
the ``DispatchComponent`` if the request body was empty, which
avoids initializing type conversion needlessly.


albe:
:+1: by reading. Good move to improve 80% of use cases

PR #317, opened 2016-03-28
TASK: Get rid of PropertyMappingConfigurationBuilder

kitsunet:
The ``PropertyMappingConfigurationBuilder`` its superfluous and contains
only one method that can easily be integrated into the ``PropertyMapper`` as
you will need that anyway to do something with the configuration.

Therefore ``PropertyMappingConfigurationBuilder::build()`` is deprecated
and fully replaced by ``PropertyMapper::buildPropertyMappingConfiguration()``.
The ``PropertyMappingConfigurationBuilder`` will be removed for the next
major Flow version.


TYPO3.Flow/Classes/TYPO3/Flow/Property/PropertyMapper.php
```diff
@@ -104,7 +104,7 @@ public static function getTypeConverterImplementationClassNames($objectManager)
     public function convert($source, $targetType, \TYPO3\Flow\Property\PropertyMappingConfigurationInterface $configuration = null)
     {
         if ($configuration === null) {
-            $configuration = $this->configurationBuilder->build();
```
daniellienert: The configurationBuilder injection should also be removed.

kitsunet: Right!

kitsunet: Done

lorenzulrich:
It looks as if the documentation needs to be adjusted here:
http://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/PropertyMapping.html#configuring-the-conversion-process

PR #318, opened 2016-03-28
TASK: Cache TypeConverterMap to avoid instantiating all TypeConverters

kitsunet:
The source and target types as well as priorities for all registered
``TypeConverters`` can be cached in Production context as they are not
changing. That avoids instantiating all ``TypeConverters`` on each
request saving some time.

NEOS-1288 #resolve

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Property/PropertyMapper.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Property/PropertyMapper.php
@@ -12,6 +12,7 @@ namespace TYPO3\Flow\Property;
  */
 
 use TYPO3\Flow\Annotations as Flow;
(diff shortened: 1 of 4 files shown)


albe:
Would it maybe make sense to statically compile the method (maybe even refactor getTypeConverterImplementationClassNames to getTypeConverterImplementationMap or sth.)?

kitsunet:
See #118 
You cannot do that because the TypeConverters cannot be instanciated during compile time, tried that, didn't work. the only way to make that happen would be to change source/target types and priority to be statically accessible which is a rather big breaking change...

albe:
Ah ok, guess due to some dependencies in the more sophisticated TCs. Well, maybe that's a route to go for something like 4.0 or so (statically accessible TC source/target) then. But thanks for the clearup.

Looks good for me then from reading :+1:

PR #319, opened 2016-04-02
BUGFIX: Always set subpackage key when setting package key in forward

albe:
The subpackage key is directly derived from the package key in the forward
method. Previously, if the subpackage key was not provided, it would not be
set in the next request.
Therefore to forward to a non-subpackage from a subpackage it was required
to specify the package key with trailing backslashes, so the subpackage key
would resolve to an empty string.

With this change it is possible by just specifying the actual package key
to forward to.


albe:
Followup for #172

albe:
Looks good for me :+1:

kitsunet:
:+1:

PR #321, opened 2016-04-17
FEATURE: Add support for configuring trusted proxies

albe:
This change introduces a configuration setting `TYPO3.Flow.http.trustedProxies.proxies`
for so called "Trusted Proxies", which is a list of IP address ranges that are trusted
to provide header overrides for some import request meta informations like the host,
port, protocol and client IP.

        proxies:
          - '216.246.40.0/24'
          - '216.246.100.0/24'

Also the header names that are trusted for those informations can be set via the configuration
option `TYPO3.Flow.http.trustedProxies.headers` and default to this:

        headers:
          clientIp: 'X-Forwarded-For'
          host: 'X-Forwarded-Host'
          port: 'X-Forwarded-Port'
          proto: 'X-Forwarded-Proto'

To get the most trusted client IP adress, the Http Request provides a new method
`getTrustedClientIpAddress()` which resolves the clientIp header, matching against
trusted proxy addresses. So this might not always end up as the actual user IP address.

By default all IP addresses are trusted to stay mostly backwards compatible, but a more
safe setting would be to trust no proxies by default:

        proxies: []

FLOW-414 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
@@ -91,21 +91,7 @@ class Request extends AbstractMessage
      */
     public function __construct(array $get, array $post, array $files, array $server)
     {
-        if (self::$trustedProxiesSettings === null) {
-            self::$trustedProxiesSettings = [
-                'proxies' => '*',
-                'headers' => [
-                    self::HEADER_CLIENT_IP => 'X-Forwarded-For',
-                    self::HEADER_HOST => 'X-Forwarded-Host',
-                    self::HEADER_PORT => 'X-Forwarded-Port',
-                    self::HEADER_PROTOCOL => 'X-Forwarded-Proto'
-                ]
-            ];
-            if (Bootstrap::$staticObjectManager !== null) {
-                $configurationManager = Bootstrap::$staticObjectManager->get(ConfigurationManager::class);
-                self::$trustedProxiesSettings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'TYPO3.Flow.http.trustedProxies');
-            }
-        }
+        $this->initializeTrustedProxiesConfiguration();
 
         $this->headers = Headers::createFromServer($server);
         $this->server = $server;
@@ -230,6 +216,18 @@ class Request extends AbstractMessage
     {
         $actionRequest = new ActionRequest($this);
         return $actionRequest;
+    }
+
+    /**
+     * Initialize the static trustedProxiesSettings property from ConfigurationManager if not yet set.
+     * @return void
+     */
+    protected function initializeTrustedProxiesConfiguration()
+    {
+        if (self::$trustedProxiesSettings === null) {
+            $configurationManager = Bootstrap::$staticObjectManager->get(ConfigurationManager::class);
+            self::$trustedProxiesSettings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'TYPO3.Flow.http.trustedProxies');
+        }
     }
 
     /**
@@ -414,46 +412,6 @@ class Request extends AbstractMessage
             $this->content = file_get_contents($this->inputStreamUri);
         }
         return $this->content;
-    }
-
-    /**
-     * Set a trusted proxy header.
-     *
-     * This is primarily useful for testing. If you want to set trusted headers, use the configuration setting
-     * "TYPO3.Flow.http.trustedProxies.headers" instead.
-     *
-     * @param string $type One of the HEADER_* constants
-     * @param string $name The header name to trust or an empty string to not trust this header
-     * @return void
-     */
-    public static function setTrustedProxyHeader($type, $name)
-    {
-        if (!in_array($type, array(self::HEADER_CLIENT_IP, self::HEADER_HOST, self::HEADER_PORT, self::HEADER_PROTOCOL))) {
-            return;
-        }
-        self::$trustedProxiesSettings['headers'][$type] = strval($name);
-    }
-
-    /**
-     * Set the trusted proxies IP addresses.
-     * If the argument is an empty array, no proxies are trusted.
-     * If the argument is the string '*', all proxies are trusted.
-     *
-     * This is primarily useful for testing. If you want to set trusted headers, use the configuration setting
-     * "TYPO3.Flow.http.trustedProxies.proxies" instead.
-     *
-     * @param array|string $trustedProxies
-     * @return void
-     */
-    public static function setTrustedProxies($trustedProxies)
-    {
-        if (is_string($trustedProxies) && $trustedProxies !== '*' && $trustedProxies !== '') {
-            $trustedProxies = [$trustedProxies];
-        }
-        if (!is_array($trustedProxies) && $trustedProxies !== '*') {
-            $trustedProxies = [];
-        }
-        self::$trustedProxiesSettings['proxies'] = $trustedProxies;
     }
 
     /**
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
@@ -429,7 +429,7 @@ class Request extends AbstractMessage
      * If the argument is the string '*', all proxies are trusted.
      *
      * This is primarily useful for testing. If you want to set trusted headers, use the configuration setting
-     * "TYPO3.Flow.http.trustedProxies.headers" instead.
+     * "TYPO3.Flow.http.trustedProxies.proxies" instead.
      *
      * @param array|string $trustedProxies
      * @return void
@@ -507,7 +507,7 @@ class Request extends AbstractMessage
      * If no proxies are trusted or no client IP header is trusted, this is the remote address of the machine
      * directly connected to the server.
      *
-     * @return string The most trusted client's IP address
+     * @return string|bool The most trusted client's IP address or FALSE if no remote address can be found
      * @api
      */
     public function getTrustedClientIpAddress()
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
@@ -92,8 +92,19 @@ class Request extends AbstractMessage
     public function __construct(array $get, array $post, array $files, array $server)
     {
         if (self::$trustedProxiesSettings === null) {
-            $configurationManager = Bootstrap::$staticObjectManager->get(ConfigurationManager::class);
-            self::$trustedProxiesSettings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'TYPO3.Flow.http.trustedProxies');
+            self::$trustedProxiesSettings = [
+                'proxies' => '*',
+                'headers' => [
+                    self::HEADER_CLIENT_IP => 'X-Forwarded-For',
+                    self::HEADER_HOST => 'X-Forwarded-Host',
+                    self::HEADER_PORT => 'X-Forwarded-Port',
+                    self::HEADER_PROTOCOL => 'X-Forwarded-Proto'
+                ]
+            ];
+            if (Bootstrap::$staticObjectManager !== null) {
+                $configurationManager = Bootstrap::$staticObjectManager->get(ConfigurationManager::class);
+                self::$trustedProxiesSettings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'TYPO3.Flow.http.trustedProxies');
+            }
         }
 
         $this->headers = Headers::createFromServer($server);
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
@@ -17,7 +17,7 @@ use TYPO3\Flow\Core\Bootstrap;
 use TYPO3\Flow\Mvc\ActionRequest;
 use TYPO3\Flow\Utility\Arrays;
 use TYPO3\Flow\Utility\MediaTypes;
-use TYPO3\Flow\Security\RequestPattern\Ip as IpPattern;
+use TYPO3\Flow\Utility\Ip as IpUtility;
 
 /**
  * Represents an HTTP request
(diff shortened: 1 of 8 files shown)


dfeyer:
Thanks for taking care, looks good, if we can make this feature for the upcoming release it should really nice

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -373,6 +406,137 @@ public function getContent($asResource = false)
...
+        self::$trustedProxiesSettings['headers'][$type] = strval($name);
+    }
+
+    /**
+     * Set the trusted proxies IP addresses.
+     * If the argument is an empty array, no proxies are trusted.
+     * If the argument is the string '*', all proxies are trusted.
+     *
+     * This is primarily useful for testing. If you want to set trusted headers, use the configuration setting
+     * "TYPO3.Flow.http.trustedProxies.headers" instead.
```
albe: "trustedProxies.proxies"

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -373,6 +406,137 @@ public function getContent($asResource = false)
...
+
+    /**
+     * Get the most trusted client's IP address.
+     *
+     * This is the right-most address in the trusted client IP header, that is not a trusted proxy address.
+     * If all proxies are trusted, this is the left-most address in the header.
+     * If no proxies are trusted or no client IP header is trusted, this is the remote address of the machine
+     * directly connected to the server.
+     *
+     * @return string The most trusted client's IP address
```
albe: "|bool" .. " or boolean FALSE if no remote address can be found."

TYPO3.Flow/Configuration/Settings.yaml
```diff
@@ -184,6 +184,25 @@ TYPO3:
               position: 'end'
               component: 'TYPO3\Flow\Http\Component\StandardsComplianceComponent'
 
+      trustedProxies:
+
+        # Defines Proxy servers which are trusted for setting request headers
+        # By default all proxies are trusted ('*')
+        # If you have some reverse proxies or CDN running in front of your server, you should provide
+        # an array list of those servers' addresses or address ranges (in CIDR notation) here:
+        #
```
albe: If you do definitely not have any reverse proxy or CDN in front of your server, you should not trust any proxies and set this option to "[]", ie. an empty list.

albe:
I'd say the support for rfc7239 "Forwarded" headers should go into a seperate PR.

albe:
So I guess this is pretty much ready for final review now. Afaik @dfeyer, @robertlemke and @bwaidelich could be candidates for review.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -71,7 +91,24 @@ class Request extends AbstractMessage
...
+                'headers' => [
+                    self::HEADER_CLIENT_IP => 'X-Forwarded-For',
+                    self::HEADER_HOST => 'X-Forwarded-Host',
+                    self::HEADER_PORT => 'X-Forwarded-Port',
+                    self::HEADER_PROTOCOL => 'X-Forwarded-Proto'
+                ]
+            ];
+            if (Bootstrap::$staticObjectManager !== null) {
+                $configurationManager = Bootstrap::$staticObjectManager->get(ConfigurationManager::class);
+                self::$trustedProxiesSettings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'TYPO3.Flow.http.trustedProxies');
```
bwaidelich: So you set the `$trustedProxiesSettings` in 95ff and then override it completely from settings here!?
Wouldn't it make sense to do either one or the other or *merge* the configured ones over the defaults?

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -71,7 +91,24 @@ class Request extends AbstractMessage
...
+            }
+        }
+
         $this->headers = Headers::createFromServer($server);
+        $this->server = $server;
```
bwaidelich: any particular reason why you moved this up here from line 152? (just curious)

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -373,6 +417,136 @@ public function getContent($asResource = false)
...
+     * If the argument is the string '*', all proxies are trusted.
+     *
+     * This is primarily useful for testing. If you want to set trusted headers, use the configuration setting
+     * "TYPO3.Flow.http.trustedProxies.proxies" instead.
+     *
+     * @param array|string $trustedProxies
+     * @return void
+     */
+    public static function setTrustedProxies($trustedProxies)
+    {
```
bwaidelich: same here.
But if this is considered useful in general, I'd recommend to rename this to `addTrustedProxy()` so that we avoid the mixed argument type

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -373,6 +417,136 @@ public function getContent($asResource = false)
...
+        self::$trustedProxiesSettings['proxies'] = $trustedProxies;
+    }
+
+    /**
+     * Get the values of trusted proxy header.
+     *
+     * @param string $type One of the HEADER_* constants
+     * @return array|null An array of the values for this header type or NULL if this header type should not be trusted
+     */
+    protected function getTrustedProxyHeaderValues($type)
```
bwaidelich: In all the places this is used the code actually only works with the **first** item of the array.. Wouldn't it make sense to rename this to `getTrustedProxyHeaderValue()` and have that logic encapsulated here?

bwaidelich:
Nice one. Just left some comments regarding low-level coding style, I'll test asap

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
albe: Had problem with just relying on the configuration from configuration manager in tests (`cannot call get() on non-object`). Also thought about a merge, but that would possibly create unwanted merge results.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
albe: Yeah, the later code (`isFromTrustedProxy()`) will always use $this->server[...] for getting remote address. Otherwise I would have had to provide the $server values as a method argument there.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -373,6 +417,136 @@ public function getContent($asResource = false)
...
+     * Set a trusted proxy header.
+     *
+     * This is primarily useful for testing. If you want to set trusted headers, use the configuration setting
+     * "TYPO3.Flow.http.trustedProxies.headers" instead.
+     *
+     * @param string $type One of the HEADER_* constants
+     * @param string $name The header name to trust or an empty string to not trust this header
+     * @return void
+     */
+    public static function setTrustedProxyHeader($type, $name)
```
albe: Well, actually it's supposed to only overwrite the already configured headers (say you want to trust the "Client-Ip" header instead for this single request). In testing it's ofc used to directly set/add the headers, as they are not loaded from configuration there. Not sure if there's really a plausible use case outside of testing, but it could be.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
albe: In all but `getTrustedClientIpAddress()`, where the list of values is actually iterated. Had that idea too before :)

kdambekalns:
I agree to @bwaidelich - please get rid of code used only for testing. Any method actually merged will be used by people, intended or not, `@api` or not. So, switch to `$this->inject(…)` in tests or use `ObjectAccess`, instead of writing setters more or less only for test use…

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
bwaidelich: > Also thought about a merge, but that would possibly create unwanted merge results.

Mh.. I think it can also be confusing to suddenly replace the settings during runtime.. Again, if its about tests lets try to get the tests deal with this (be it by manually setting `Bootstrap::$staticObjectManager` where it would break otherwise  (or extract that part out to a method like `getTrustedProxiesSettings()` and mock that in tests)

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
bwaidelich: ah, sure, makes sense of course

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
bwaidelich: Ok, got you
> Not sure if there's really a plausible use case outside of testing, but it could be.
I call YAGNI ;)  lets inject this from the tests and implement it once actually needed.

But if we decide to keep it, we should probably throw an exception for a non-allowed $type instead of silently ignoring that case

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
bwaidelich: Fair enough. We made some bad experiences with mixed argument types though, as it makes the code more complex and less comprehensive.
What about changing this to:

```php
/**
 * @param array $trustedProxies
 * @return void
*/
public static function setTrustedProxies(array $trustedProxies)
 {
    self::$trustedProxiesSettings['proxies'] = $trustedProxies;
}
```
?

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
bwaidelich: Ah right

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
albe: Sounds good, will take care of that

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
albe: Currently it's actually a mixed type setting, since the '*' placeholder will denote a blanket whitelist. So either we then add a special method for that case (`trustAllProxies()` or sth.) or we change the configuration for that to ['*'] - but IMO that's odd, what if you have a list of IPs and then one entry '*'? That's ambiguous then, so I'm more leaning towards the special setter method

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
albe: Yeah, probably about right. Will remove that method and rather use a workaround in tests

albe:
Ok, actually completely got rid of those static setters now. They're just bad after all and might introduce hard to track behaviour if used wrongly, so you are right @bwaidelich @kdambekalns. Also moved the settings initialization into its own method, so this can be easily mocked away in other tests when needed and added the missing schema for the new settings (thanks @mficzel for reminding me indirectly ;).

dfeyer:
Work fine

@kdambekalns Can you remove your downvote, now that all tests related methods are gone ?

albe:
:tada: Thx everyone! :)

kdambekalns:
Oh, well, this broke TYPO3.Form tests… see https://github.com/neos/form/pull/16

skurfuerst:
Hey everybody, this broke the ./flow server:run command:
```
[Wed Jun 15 19:07:14 2016] PHP Fatal error:  Uncaught Error: Call to a member function get() on null in /Users/sebastian/src/neosmaster/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php:228
Stack trace:
#0 /Users/sebastian/src/neosmaster/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php(94): TYPO3\Flow\Http\Request->initializeTrustedProxiesConfiguration()
```

All the best,
Sebastian

skurfuerst:
OK, I think I understand why ./flow server:run breaks:

- PhpDevelopmentServerRouter.php calls Bootstrap->run()
- this triggers the active request handler.
  - this, in turn, directly creates the Request.
    - this, in turn, initializes the trusted proxies configuration.
  - Afterwards, this calls $this->boot(), leading to the Object Management and Configuration init. Which is too late for the trusted proxies config.

*However, I do not understand yet why this can ever work when not using ./flow server:run, but classical FCGI based hosting.*

Opinions?

kdambekalns:
Hm, same error as in the test in TYPO3.Form I mentioned above.

Too bad that neither @dfeyer nor @albe ran into troubles - I have to admit I trusted the tests. Then again, I'd not have checked `server:run` nor FCGI anyways…

It boils down to the `Request` no longer being "completely standalone" and relying on the `ObjectManager` being available.

kitsunet:
Wait what? Nooooooo, please. Http Request must be standalone. Inject this config from outside, optionally.

albe:
Ok, I think it is even worse. The issue is actually that the `Request` is created even before the essential bootstrap sequence is run (inside the requestHandler). So in fact, the ConfigurationManager isn't even available when `Request::__construct` is executed. Question is how to get something from Configuration that early?

I didn't like the $staticObjectManager thing from the start, but didn't see any other way.

Only thing I see atm is reordering the calls inside the requestHandler a bit (move `$this->boot()` before the `Request` creation). With that it would be possible to inject the settings from there.

kdambekalns:
Well, whenever a `Request` is created, this setting needs to be given. Automatic injection isn't possible, because of `Proxy(false)`–and that probably is there for a reason.

albe:
Yeah, `Proxy(false)` is definitely the only sane thing to do on a `Request`. Just had a quick talk with Christian and probably the best solution would be to handle this via an **http component** instead. It could either wrap the `Request` in sth. like a `TrustedRequest` or overwrite the `Request` infos (or possibly even just reject Requests that contain untrusted headers).

Obvious drawback is that anything before the component can not be sure about the Request being trusted. In return all this logic is completely separated from the Request and hence easily configurable. Also makes a PSR-7 switch easier.

So should we revert this for now?

bwaidelich:
I like the idea of a HTTP component. If it is configured to be at the **start** of the `preprocess` chain it's hopefully early enough. The chain is created in the `Http\RequestHandler` right after creating the initial Http\Request

kdambekalns:
But… will that work for case like `server:run`? I have to admit I am not sure if the component chain is active for those "edge cases".

bwaidelich:
> But… will that work for case like `server:run`?

yep ;)

PR #322, opened 2016-04-20
TASK: Remove unused Security JavaScript

kitsunet:


gerhard-boden:
If it's unused I suppose we can do without it. :+1: How did it get in there in the first place?

kdambekalns:
This is a leftover from a login view that was removed with commit ec4dab595e4a7275e85f885bba4ef583d775c8f4. Oh, the pain of looking at that old stuff ;)

PR #324, opened 2016-04-20
TASK: Translator uses LocaleChain

SoulCover:
Actually implemented the usage of the LocalChain way down in the ``XliffTranslationProvider``.

FLOW-356 #close Fixed by this

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/I18n/TranslationProvider/XliffTranslationProvider.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/I18n/TranslationProvider/XliffTranslationProvider.php
@@ -67,35 +67,65 @@ class XliffTranslationProvider implements \TYPO3\Flow\I18n\TranslationProvider\T
     }
 
     /**
+     * Returns translated $stringToTranslate from a file defined by $sourceName using the function $functionName of the XliffModel.
+     *
+     * Chooses particular form of label if available and defined in $pluralForm.
+     *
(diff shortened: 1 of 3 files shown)


SoulCover:
I would be grateful if somebody would adjust the unit tests.

**Breaking?!?**

TYPO3.Flow/Classes/TYPO3/Flow/I18n/TranslationProvider/XliffTranslationProvider.php
```diff
@@ -67,35 +67,65 @@ public function injectPluralsReader(\TYPO3\Flow\I18n\Cldr\Reader\PluralsReader $
...
-        $model = $this->getModel($packageKey, $sourceName, $locale);
+        if ($locale === null) {
+            $locale = $this->localizationService->getConfiguration()->getCurrentLocale();
+        }
+
+        $translation = false;
 
-        if ($pluralForm !== null) {
-            $pluralFormsForProvidedLocale = $this->pluralsReader->getPluralForms($locale);
+        foreach ($this->localizationService->getLocaleChain($locale) as $localeInChain) {
```
albe: Not sure if the iteration should be so deep down in the specific TranslationProvider. It's a behaviour I would rather expect to be independent of the specific implementation.

albe:
Yes, I'd say breaking, as it changes interface (public methods signature) of core classes that are probably used by third parties.

I can take a look at the tests, but the current error should go away with my comment.

PR #326, opened 2016-04-21
TASK: Update OneToMany annotation in Persistence.rst

koernchen02:
Remove quotes in example, since the option orphanRemoval expects a boolean value


albe:
:+1:

PR #327, opened 2016-04-21
!!! TASK: Make Translator use locale chain

ComiR:
This change makes `getTranslationById` and `getTranslationByOriginalLabel` use the configured
locale chain.

Followup to #327

FLOW-356 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/I18n/TranslationProvider/TranslationProviderInterface.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/I18n/TranslationProvider/TranslationProviderInterface.php
@@ -27,12 +27,12 @@ interface TranslationProviderInterface
      *
      * @param string $originalLabel Label used as a key in order to find translation
      * @param \TYPO3\Flow\I18n\Locale $locale Locale to use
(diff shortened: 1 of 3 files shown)


albe:
> Wouldn't it be best to implement the LocaleChain as deep down as possible? Especially to keep Plurals in the same locale.

I still think the behaviour of the LocaleChain should be independent of the specific TranslationProvider, as it is a generic feature of i18n. But the plurals/quantity definitely should go down.

Let's see what the others think

kitsunet:
I am not sure, perfect would be if there is some basic logic in the Translator for example but I think that there should be a possibility for the Provider to handle fallbacks for optimized handling. (Eg. a database provider might want to fetch all languages of the fallback chain in one query instead of doing multiple).

Maybe the logic could go to a trait or Abstract Provider to allow it to be easily reused by other Providers? That would maybe be a good compromise.

robertlemke:
Reading from your discussion state I guess that this patch won't reach consensus / maturity before the release, right?

jonnitto:
Ping

ComiR:
Closing in favor of #1451.
Please continue discussion there.

PR #328, opened 2016-04-21
TASK: Make XliffTranslationProvider use locale chain

ComiR:
This change makes `getTranslationById` and `getTranslationByOriginalLabel` use the configured
locale chain.

Replaces #324, succession by #327 needs to be discussed.

FLOW-356 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/I18n/Translator.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/I18n/Translator.php
@@ -125,16 +125,21 @@ class Translator
         if ($locale === null) {
             $locale = $this->localizationService->getConfiguration()->getCurrentLocale();
         }
-        $pluralForm = $this->getPluralForm($quantity, $locale);
+        $localeInChain = $locale;
+        $translatedMessage = $originalLabel;
+        foreach ($this->localizationService->getLocaleChain($locale) as $localeInChain) {
+            $pluralForm = $this->getPluralForm($quantity, $localeInChain);
 
-        $translatedMessage = $this->translationProvider->getTranslationByOriginalLabel($originalLabel, $locale, $pluralForm, $sourceName, $packageKey);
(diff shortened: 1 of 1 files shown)


albe:
Tests need a proper LocaleChain mock it seems, other than that looks good. I can take care of the tests if no one else does beforehand.

albe:
Created a PR on your fork for the tests. Would be cool if you could allow all Neos team members push access to your fork to improve working together.

albe:
So, looks good for me :+1:

kdambekalns:
One thing that bothers me out of the box: now `getTranslationBy…` loops over the locale chain and in turn calls `getModel` and thus `getXliffFilenameAndPath` which again loops over the locale chain.

Sounds like a lot of looping to me… the again, it could very well be a negligible amount of time spent there. Any numbers on this, @ComiR or @albe?

albe:
Good catch, but I think even if performance weren't a problem, there is a functional problem. The translator deals with choosing the locale to use if none is specified, while the providers always expect a valid locale - but the Xliff Provider might return a file that does not match that locale, because `getLocaleChain` will for example use configured fallbacks or even the default locale.
So IMO the `I18n\Service::getXliffFilenameAndPath` should not deal with the locale chain at all. Since it's only used in XliffTranslationProvider it should not break anything. Alternatively, for maximum b/c, it should have a $strict parameter like `I18n\Service::getLocalizedFilename` which should be set to true when calling from XliffTranslationProvider.

dfeyer:
This change is breaking so should target master ... or maybe only the next major ?

albe:
Is this really breaking? I mean it will possibly return translations from the chain where it wouldn't return anything before, but is this considered *breaking*? Just curious

jonnitto:
Any progress on this? Is this something for the new release?

ComiR:
Closing in favor of #1451.
Please continue discussion there.

PR #332, opened 2016-04-25
BUGFIX: package:rescan must happen before booting

kitsunet:
If the user issued a ``typo3.flow:package:rescan`` command this should
happen very early in the compiletime boot sequence as otherwise the
ObjectManager will try to load a non existing package if the rescan was
issued after a package was removed from the installation.

With this change package rescans are possible in all package state
situations in case you manipulate the packages without using composer
or our ``PackageManager``.

Additionally adds a bunch of use statements to the ``Bootstrap``.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
@@ -11,16 +11,33 @@ namespace TYPO3\Flow\Core\Booting;
  * source code.
  */
 
+use Doctrine\Common\Annotations\AnnotationRegistry;
 use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Cache\CacheFactory;
+use TYPO3\Flow\Cache\CacheManager;
+use TYPO3\Flow\Configuration\ConfigurationManager;
+use TYPO3\Flow\Configuration\Source\YamlSource;
 use TYPO3\Flow\Core\Bootstrap;
 use TYPO3\Flow\Core\ClassLoader;
+use TYPO3\Flow\Core\LockManager as CoreLockManager;
+use TYPO3\Flow\Error\Debugger;
+use TYPO3\Flow\Error\ErrorHandler;
+use TYPO3\Flow\Log\Logger;
+use TYPO3\Flow\Log\LoggerFactory;
+use TYPO3\Flow\Log\SystemLoggerInterface;
 use TYPO3\Flow\Monitor\FileMonitor;
+use TYPO3\Flow\Object\CompileTimeObjectManager;
+use TYPO3\Flow\Object\ObjectManager;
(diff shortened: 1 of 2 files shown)


kitsunet:
@kdambekalns that should definitely go into 2.2. To see how it works simply add a package to the installation make sure it works and finally remove it by straight deleting it. Should give you an exception if you run the mentioned package rescan command unless you have this change applied.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -161,10 +178,16 @@ public static function initializeSignalSlot(Bootstrap $bootstrap)
      */
     public static function initializePackageManagement(Bootstrap $bootstrap)
     {
-        $packageManager = new \TYPO3\Flow\Package\PackageManager();
-        $bootstrap->setEarlyInstance(\TYPO3\Flow\Package\PackageManagerInterface::class, $packageManager);
+        $packageManager = new PackageManager();
+        $bootstrap->setEarlyInstance(PackageManagerInterface::class, $packageManager);
+
+        // The package:rescan must happen as early as possible, compiletime alone is not enough.
```
kitsunet: The next three lines contain the actual bugfix, the rest is just use statements.

albe: Something similar is already in the forced cache flushing. Maybe makes sense to wrap the $_SERVER['argv'] somewhere inside the Bootstrap?

Also, is the rescanPackages() method now called twice? Once here in bootstrap and once in the commandController?

kitsunet: Right! That should probably be fixed, even though it doens't really hurt.

kitsunet: Actually it needs to happen twice if we want to deliver the freshly sorted list of packages as command output...

kdambekalns:
@kitsunet you mean an error like this?

`Fatal error: require_once(): Failed opening required '…/Neos/GoogleAnalytics/Package.php' (include_path='.:') in …/Package/PackageFactory.php on line 49`

kdambekalns:
So, I `composer require`d the Google Analytics package and then did `rm -rf Packages/Application/TYPO3.Neos.GoogleAnalytics`. Now all I get is the aforementioned error, even on `cache:flush --force` and even when the caches are empty…

kitsunet:
This is a problem that needs to be fixed separately, try a package without  `Package.php` for testing this.

kdambekalns:
Ah, it worked when using `flow:package:rescan` (as opposed to `package:rescan`).

kdambekalns:
@kitsunet why only master, btw?

kitsunet:
Great, uhm new package manager is master only :) thanks for taking care.

PR #333, opened 2016-04-25
TASK: Allow wildcard propertymapping configuration to set defaults for trustedProperties configuration

albe:
Before, when specifying a PropertyMappingConfiguration with a wildcard inside the initializeAction() of
a controller, this configuration would not be taken into account when a Fluid form was submitted, that
set configuration for specific properties.

This change makes the trustedProperties configuration happen *after* the user configuration, which in turn
will lead to trustedProperties always only extend the user configuration and wildcard configurations to
be used as the template configuration for the following specific trustedProperties configurations.


koernchen02:
:+1:

kitsunet:
Looks very nice

PR #334, opened 2016-04-25
FEATURE: Validate only up to Aggregate boundaries

albe:
This is one specific attempt at optimizing entity hierarchy validation, by stopping validation at unloaded Aggregate boundaries.

It is based on the idea that unloaded aggregates (doctrine proxy instances) do not carry changes, and since aggregates form a natural consistency boundary validation should not change and therefore can be stopped.

This will improve performance noticeably if you designed relatively small aggregates, but have lots of aggregate relations in your domain model.

FLOW-17 #comment This optimizes validation performance for well designed Aggregates

[snip] the following can be removed for merging: [/snip]

See also https://discuss.neos.io/t/rfc-optimized-validation-of-entities/340/5

Pro:
- simple implementation
- makes sense from a DDD perspective
- it's easy to define validation cascading through eager loading relations

Con:
- as is, only works for lazily loaded aggregate relations (which is default in doctrine)
- loading of proxies can sometimes happen unwantedly
- does not help on deep aggregates

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
@@ -10,25 +10,20 @@ namespace TYPO3\Flow\Validation\Validator;
  * information, please view the LICENSE file which was distributed with this
  * source code.
  */
-use Doctrine\ORM\Proxy\Proxy;
 
 /**
- * A validator which will only validate up to the Aggregate boundary, or if an Aggregate relation
- * is specifically annotated to be cascade validated.
+ * A validator which will not validate Aggregates that are lazy loaded and uninitialized.
+ * Validation over Aggregate Boundaries can hence be forced by making the relation to
+ * other Aggregate Roots eager loaded.
  *
  * @api
  */
 class AggregateBoundaryValidator extends GenericObjectValidator
 {
     /**
-     * @var array
-     */
-    protected $supportedOptions = array(
-    );
-
-    /**
      * Checks if the given value is valid according to the validator, and returns
-     * the Error Messages object which occurred.
+     * the Error Messages object which occurred. Will skip validation if value is
+     * an uninitialized lazy loading proxy.
      *
      * @param mixed $value The value that should be validated
      * @return \TYPO3\Flow\Error\Result
@@ -40,7 +35,7 @@ class AggregateBoundaryValidator extends GenericObjectValidator
         if ($this->acceptsEmptyValues === false || $this->isEmpty($value) === false) {
             if (!is_object($value)) {
                 $this->addError('Object expected, %1$s given.', 1241099149, array(gettype($value)));
-            } elseif ($value instanceof Proxy && !$value->__isInitialized()) {
+            } elseif ($value instanceof \Doctrine\ORM\Proxy\Proxy && !$value->__isInitialized()) {
                 return $this->result;
             } elseif ($this->isValidatedAlready($value) === false) {
                 $this->isValid($value);
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
@@ -1,0 +1,52 @@
+<?php
+namespace TYPO3\Flow\Validation\Validator;
+
+/*
+ * This file is part of the TYPO3.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+use Doctrine\ORM\Proxy\Proxy;
+
+/**
+ * A validator which will only validate up to the Aggregate boundary, or if an Aggregate relation
+ * is specifically annotated to be cascade validated.
+ *
+ * @api
+ */
+class AggregateBoundaryValidator extends GenericObjectValidator
+{
+    /**
+     * @var array
+     */
+    protected $supportedOptions = array(
+    );
+
+    /**
+     * Checks if the given value is valid according to the validator, and returns
+     * the Error Messages object which occurred.
+     *
+     * @param mixed $value The value that should be validated
+     * @return \TYPO3\Flow\Error\Result
+     * @api
+     */
+    public function validate($value)
+    {
+        $this->result = new \TYPO3\Flow\Error\Result();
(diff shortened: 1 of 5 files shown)


kdambekalns:
Noticed @albe brought up another discussion point in the RFC linked above…

albe:
This PR is relatively independent of the discussion in the RFC though. It can/should be finalised anyway as it just makes sense to validate like this - after all, persistence cascading is done similarly by default (`cascade all` unless it's an AggregateRoot).

Things I'd like feedback on specifically: Is it good enough to depend on eager loading to define the validation cascading over the aggregate boundary? Or should a specific annotation be introduced for that?

kitsunet:
I would say it's fine for now. We can introduce an additional annotation if the need comes up.

TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+use Doctrine\ORM\Proxy\Proxy;
+
+/**
+ * A validator which will only validate up to the Aggregate boundary, or if an Aggregate relation
+ * is specifically annotated to be cascade validated.
```
albe: This needs a correction, as there currently isn't a simple cascade validation annotation.

TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @api
+ */
+class AggregateBoundaryValidator extends GenericObjectValidator
+{
+    /**
+     * @var array
+     */
+    protected $supportedOptions = array(
```
albe: Left-over from an attempt to make behaviour configurable. Can be removed.

albe:
Ok, I'll clean-up the commit and add documentation then.

albe:
StyleCI issues are unrelated.

albe:
Rebased to fix StyleCI issues.

albe:
@kdambekalns @kitsunet would you please take a quick lock again that the added documentation cleanup is good to go? Didn't change anything else.

TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
```diff
@@ -0,0 +1,47 @@
...
+     * the Error Messages object which occurred. Will skip validation if value is
+     * an uninitialized lazy loading proxy.
+     *
+     * @param mixed $value The value that should be validated
+     * @return \TYPO3\Flow\Error\Result
+     * @api
+     */
+    public function validate($value)
+    {
+        $this->result = new \TYPO3\Flow\Error\Result();
```
bwaidelich: <nitpick>The namespace could be imported for better readability</nitpick>

TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
```diff
@@ -0,0 +1,47 @@
...
+        if ($this->acceptsEmptyValues === false || $this->isEmpty($value) === false) {
+            if (!is_object($value)) {
+                $this->addError('Object expected, %1$s given.', 1241099149, array(gettype($value)));
```
bwaidelich: the exception code 1241099149 is copied from the `GenericObjectValidator`. Please create a new one based on the current unix timestamp
<nitpick>`array()` => `[]` for new code</nitpick>

TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
```diff
@@ -0,0 +1,47 @@
...
+            } elseif ($value instanceof \Doctrine\ORM\Proxy\Proxy && !$value->__isInitialized()) {
```
bwaidelich: Maybe we could add a inline comment explaining why uninitialized doctrine proxies are skipped?
&lt;nitpick&gt;The namespace could be imported&lt;/nitpick&gt;

TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
bwaidelich: hang on, probably it makes sense to keep the same exception code - its the same exception reason at least..
Couldn't we simplify the whole method to sth like:
```php
public function validate($value)
{
    if ($value instanceof DoctrineProxy && !$value->__isInitialized()) {
        return new Result();
    }
    return parent::validate($value);
}
```

bwaidelich:
I added a few nitpicks (that can be ignored).
But I think the AggregateBoundaryValidator could be simplified a little

TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
albe: Good idea!

TYPO3.Flow/Classes/TYPO3/Flow/Validation/Validator/AggregateBoundaryValidator.php
albe: In this case I actually did not import the namespace to make it more clear that we are dealing with a doctrine proxy here.

bwaidelich: ```use Doctrine\ORM\Proxy\Proxy as DoctrineProxy```
but, as written, just a minor nitpick that can be safely ignored (I just don't get used to the ugly php namespace separator ;)

albe:
Refactored the code as to Bastians suggestion and added an inline documentation for the intention. Also, I removed the @api annotation from the validator, since it is rather an internal implementation detail than something that should be used by developers directly.

bwaidelich:
Thanks, looks good (by reading)

albe:
Now just one question remaining: Should this maybe be retargeted at a lower version, since it's rather a performance optimization than a real feature? Guess at least 3.2 would be ok, but maybe even lower? wdyt

bwaidelich:
> Should this maybe be retargeted at a lower version, since it's rather a performance optimization than a real feature?

There's a risk of broken (as in "changed") behavior, so I'd vote for keeping this in master.
Hopefully it won't be long till the next release

kitsunet:
ACK, I also vote for master.

PR #335, opened 2016-04-25
FEATURE: Validate only objects that actually need validation

albe:
This change optimizes object validation, especially for large
hierarchies. Objects are only validated during persistence when
they contain changes and only validated in controller arguments
when they were submitted with set properties.

This is a second (unfinished) attempt at optimizing object hierarchy validation, which
is much more careless about edge use-cases.
Also see #334 and https://discuss.neos.io/t/rfc-optimized-validation-of-entities/340/5

Pro:
 - pretty much optimal validation performance
 - based on simple (naive) convention "what didn't change cannot fail"

Con:
 - relatively complex to implement (needs to store change state information at some place accesible for validators)
 - as is, ignores use-cases where partial object submission might fail on validators of previously submitted objects, depending for example on validation groups (see also https://neos-project.slack.com/archives/guild-validation/p1441885466000005)
 - will cause some coupling between property mapping and validation

FLOW-17 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/ActionController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/ActionController.php
@@ -159,6 +159,8 @@ class ActionController extends AbstractController
 
         $this->mapRequestArgumentsToControllerArguments();
 
+        // TODO: We could do the validation initialization and invocation here and then make use of change information inside the propertyMapper
+
         if ($this->view === null) {
             $this->view = $this->resolveView();
         }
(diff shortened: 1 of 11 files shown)


albe:
@bwaidelich

kitsunet:
I read coupling... :/

albe:
Yeah, of the above cons that's actually the second-most bad side-effect after the breaking behaviour of edge-cases. The latter could be solved with my suggestion to separate **user input** and **business constraint** validation (see my last post in the RFC) and the first con is actually just a matter of time/effort.

Maybe, once that separation of validation is in place, we can just avoid the loading of lazy relations inside the user input validation that is currently happening. If we also take `extra lazy` collections into consideration, that could be a way to deal with the left-over use case that #334 won't handle:
An Aggregate that holds a huge list (hundreds to thousands) of entities that can not be separated and that have validations. In that case the only sane way to deal with this would be to use extra-lazy relations after all.

Thanks for triggering this thought process Christian :) Just shows how any feedback to PRs is invaluable!

albe:
Closing and will start something new for separating user input and business rule validation, then optimize from there.

PR #339, opened 2016-04-27
TASK: Provide comments used as command reference for ``redirect`` command

gerhard-boden:
Add PHPDoc comments used for command reference in the ``redirect`` command controller.


gerhard-boden:
I merged the text from https://github.com/neos/flow-development-collection/pull/339 into the PHPDoc comments. When you find a lot of things to change please just create a PR with the changes, because that way i can merge things fast when I'm at work

PR #340, opened 2016-04-29
BUGFIX: Add translation for error message (string length)

kdambekalns:
The translation is broken in 3.0 and up, even though it appears to be
correctly translated in Crowdin.

This is try to fix this.

### DIFF
--- a/TYPO3.Flow/Resources/Private/Translations/de/ValidationErrors.xlf
+++ b/TYPO3.Flow/Resources/Private/Translations/de/ValidationErrors.xlf
@@ -116,24 +116,20 @@
 				<source>The length of this text must be between {0,number} and {1,number} characters</source>
 			<target xml:lang="de">Die Länge des Textes muss zwischen {0,number} und {1,number} Zeichen liegen</target></trans-unit>
       <group id="1238108068" restype="x-gettext-plurals">
-        <trans-unit id="1238108068[0]">
-          <source/>
-          <target xml:lang="de"/>
-        </trans-unit>
-        <trans-unit id="1238108068[1]">
-          <source/>
-          <target xml:lang="de"/>
-        </trans-unit>
+        <trans-unit id="1238108068[0]" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


kdambekalns:
By analyzing using my project knowledge, I identifier @aertmann, @kitsunet and @albe to be potential reviewers

albe:
The new kdambekalns-mentionbot works pretty well ;)

aertmann:
it's a general problem, not just with German.. probably due to the identifier containing `[]`

thus it will probably just break next time it's overridden and still broken with all other languages

@kdambekalns is there anything we can do to fix it or do we need to open an issue with Crowdin?

kdambekalns:
Hm, that sounds like an issue with CrowdIn then. Also I found http://crowdin.betaeasy.com/thread/4164 which sound slightly similar. I contacted the CrowdIn support… let's see what happens.

kdambekalns:
So I got a reply already:

> I assume the issue is caused on our side, somehow it's related with plural forms.
> 
> The task is already created and we will do our best to find the solution this week. Due to holidays here in Ukraine, our developers can review this bug not earlier than Wednesday this week, sorry for a little delay.

kdambekalns:
Got this in the mail today:

> I want to inform you that we have already fixed the issue with the exported XLIFF files and pushed the fix to the production.
> It would be great if you can confirm that everything works fine!

I'll check the generated XLIFF files…

kdambekalns:
Their fix doesn't change anything. I'll let them know, let's hope for more feedback.

kdambekalns:
Ok, they got back to me and after updating the english (source) XLIFF file, the plural form shows up in the translation screen and is built correctly. So we can close this. 🎉 

plural-forms-crowdin

kdambekalns:
FYI: We have one more plural in Neos "Modules.xlf", that still doesn't translate correctly (is not showing up at all in the Crowdin UI). I got in touch with their support…

PR #341, opened 2016-05-01
TASK: Add additional detection of changed package files

kitsunet:
`Package.php` files are used for specific Flow bootstrap behavior
in packages and the information if such a file exists is cached in
`PackageStates.php`. In development it can happen that a new file
is created or an existing one removed and that should be detected.

This change adds detection for change (creation) and removal of
`Package.php` files.


kitsunet:
@kdambekalns we talked about this. Needs good testing but should probably go into 3.2 ASAP.

kdambekalns:
Looks good, will test ASAP

kitsunet:
I still have no clue why the tests fail :( Probably some mysterious behat / travis magic failing.

albe:
Looks good from reading. Failing behat tests look unrelated (table *typo3_flow_security_account* doesn't exist?)

kitsunet:
Problem is that they consistently fail with this change and AFAIK only with this change.

kitsunet:
Something is indeed not right yet, will investigate a bit further.

kitsunet:
Need help, no clue what the behat tests do there and how that is influenced by this change. IMHO the behat tests are the fault here not the change.

kitsunet:
Ok, so I am trying this again and I get the same error locally with AND WITHOUT the change... ???

albe:
Have you tried rebasing?

kitsunet:
I just rebased, will push.

Actually the test works for me IF I use only flow. (With rebased PR)

kitsunet:
Nope. didn't help :(

kitsunet:
locally the test is fine. I don't get it

albe:
Hm, ok, so the core of the issue comes from the "FIXME" documented `$this->objectManager->get(\TYPO3\Flow\Security\AccountRepository::class)->countAll();` in `SecurityOperationsTrait::authenticateRoles()`.

What would be helpful to check is if this happens with any PR onto 3.2 now, ie. if is 3.2 simply broken currently (for travis)?

kitsunet:
haha, good point

kdambekalns:
Maybe something for @robertlemke  ;)

albe:
If only failing travis keeps this from getting merged, it could be worth another look.

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageFactory.php
```diff
@@ -41,10 +43,15 @@ public function create($packagesBasePath, $packagePath, $packageKey, $composerNa
...
         $packageClassName = Package::class;
+        $packageClassExists = true;
         if (!empty($packageClassInformation)) {
             $packageClassName = $packageClassInformation['className'];
             $packageClassPath = Files::concatenatePaths(array($absolutePackagePath, $packageClassInformation['pathAndFilename']));
-            require_once($packageClassPath);
+            $packageClassExists = @include_once($packageClassPath);
+        }
+
+        if ($packageClassExists === false) {
```
kdambekalns: The `require_once()` used to include a package class _if it existed_. If not, the generic package class was used. Now an exception is thrown, which seems wrong… Let's see if that fixes the tests if changed.

kdambekalns:
Hm. Still think my last change makes sense, but didn't help with the tests…

albe:
What if we just try applying it on 4.0 and see if it works there? If it does, I'd say go with it in 4.x... if not, we at least know for sure it's an deeper lying issue

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageFactory.php
```diff
@@ -41,15 +43,31 @@ public function create($packagesBasePath, $packagePath, $packageKey, $composerNa
...
+            $packageClassExists = @include_once($packageClassPath);
+
+            if ($packageClassExists === false) {
+                throw new Exception\PackageClassMissingException(
+                    sprintf(
+                        'The package class "%s" (expected in file "%s" of package "%s") could not be loaded.',
+                        $packageClassName,
+                        $packageClassPath,
+                        $$packageKey
+                    ), 1461911652);
```
aertmann: this check is to make sure that it works when `$packageClassInformation` is provided directly to the function right? because this shouldn't be possible when it comes from `detectFlowPackageFilePath`

TYPO3.Flow/Classes/TYPO3/Flow/Package.php
```diff
@@ -125,6 +125,21 @@ public function boot(Core\Bootstrap $bootstrap)
...
         });
 
+        $dispatcher->connect(\TYPO3\Flow\Monitor\FileMonitor::class, 'filesHaveChanged', function ($monitorIdentifier, $changedFiles) use ($bootstrap) {
+            if ($monitorIdentifier !== 'Flow_ClassFiles') {
+                return;
+            }
+            $packageManager = $bootstrap->getEarlyInstance(PackageManagerInterface::class);
+            $packagePhpReversed = strrev('Package.php');
+            foreach ($changedFiles as $pathAndFilename => $status) {
+                if (strpos(strrev($pathAndFilename), $packagePhpReversed) === 0) {
```
aertmann: what's the benefit of the string reversing?

albe: It is basically a check for "endsWith" done a bit cryptic for performance.

i.e. `strpos(strrev($pathAndFilename), "php.egakcaP") === 0 <=> strpos($pathAndFilename, "Package.php") === strlen($pathAndFilename) - strlen("Package.php")`

aertmann: ah got it, thanks!

kdambekalns:
So, #1020 led to @kitsunet saying:

> Ok, so the check via file monitor creates an infinite loop apparently and thus breaks the compile step. I am not sure how fix this yet...

So trying this on 4.0 has *some* effect, at least.

albe:
> the check via file monitor creates an infinite loop apparently

Maybe just do a stupid recursion check and jump out of the File_Monitor handler...

PR #343, opened 2016-05-04
BUGFIX: Fix try/catch in ObjectAccessorNode::getPropertyPath for is/has access

whythecode:
Related to: #108

As a safer measure, we're now catching any `\Throwable`.

- if rendered without isSomething in the arguments (implicit falsy), we then get `Call to undefined method TYPO3\Fluid\Core\ViewHelper\TemplateVariableContainer::isSomething()`

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
@@ -93,9 +93,13 @@ class ObjectAccessorNode extends AbstractNode
             try {
                 $subject = ObjectAccess::getProperty($subject, $propertyName);
             } catch (PropertyNotAccessibleException $exception) {
-                if (is_object($subject) && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) > 0) {
+                if (is_object($subject)
+                    && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) > 0
+                    && is_callable([$subject, $propertyName])) {
                     try {
                         $subject = $subject->$propertyName();
+                    } catch (\Throwable $exception) {
+                        $subject = null;
                     } catch (\Exception $exception) {
                         $subject = null;
                     }


whythecode:
maybe someone can write a test for this? I have 0 familiarity with the Unit and Functional tests in Flow :)

I suppose it would be either in `ObjectAccessorNodeTest` or `ParserIntegrationTest`.

kdambekalns:
Well, `Throwable` is new in PHP 7, so I wonder what happens with older versions… I guess this needs to be taken care of.

albe:
Guess for backwards compatibility it needs to look something like this:

https://github.com/neos/flow-development-collection/pull/303/files#diff-2a982b6b93e930471d554015c31f5ebeL113

MarkusRodler:
@whythecode damn you're fast. i just wanted to open an bugreport on this topic because my installation is broken with 3.2.0

whythecode:
@kdambekalns ahhh LOL, I wasn't aware of that, I'll do as @albe suggested then

btw, @albe thanks for this feature, I love it and it actually will help readability in our project

@MarkusRodler hehe well, I had the luck of getting the Exception right on my homepage :)

whythecode:
@albe did you ever get \Exception thrown? then I'll keep it and just add \Error and \Throwable instead of replacing it

albe:
Well, the idea was that the non-existing method would throw an exception, for example if a class implements __call and signals non-existing methods with an exception. But after looking again at it, that whole method invocation is pretty naive, because in any other case (of PHP < 7) if the method doesn't exist, a fatal error happens that wouldn't be caught. So I guess there should at least be a `is_callable(array($subject, $propertyName))` check around it.

TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
```diff
@@ -96,7 +96,9 @@ public static function getPropertyPath($subject, $propertyPath, RenderingContext
                 if (is_object($subject) && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) > 0) {
                     try {
                         $subject = $subject->$propertyName();
-                    } catch (\Exception $exception) {
+                    } catch (\Throwable $exception) {
+                        $subject = null;
+                    } catch (\Error $exception) {
```
aertmann: `\Throwable` would catch `\Error` since it extends it

and this wouldn't work in PHP 5.5/5.6 since `\Exception` would be thrown there..

so it has to be

    } catch (\Throwable $exception) {
        $subject = null;
         $subject = null;
    }

aertmann:
yeah like @albe said, also it has to be pushed against the `3.2` branch instead

albe:
@whythecode can you push a new PR with the try/catch as suggested targeting 3.2? I'll then provide a PR to add a test and then we should probably get a bugfix release out asap

bwaidelich:
As @albe mentioned for PHP < 7 this has to be explicitly tested like so:
```php
if (is_object($subject)
    && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) === 1
    && is_callable([$subject, $propertyName])) {
    $subject = $subject->$propertyName();
} else {
    $subject = null;
}
```

All this within a try/catch... not so nice but IMO OK for a hotfix.

For the next major version I'd really like to move this whole special case into `ObjectAccess` to keep that in sync and because it makes sense outside of Fluid, too (see https://github.com/bwaidelich/Wwwision.GraphQL/blob/master/Classes/AccessibleObject.php#L82-L84 for example)

whythecode:
closed in favor of #348

PR #344, opened 2016-05-05
BUGFIX: PackageManager uses incorrect packageKey for split libraries

radmiraal:
The class loader derives the packageKey from the namespace declaration
for non-Flow packages. This is a problem for the split library package
which then share the same packageKey (being ``TYPO3.Flow.Utility``).

This change adds special handling for packages of type ``library`` and
a composer name starting with ``neos/``.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
@@ -1196,7 +1196,7 @@ class PackageManager implements PackageManagerInterface
     {
         $packageKey = '';
 
-        if ($packageType !== null && ComposerUtility::isFlowPackageType($packageType)) {
+        if (($packageType !== null && ComposerUtility::isFlowPackageType($packageType)) || ($packageType === 'library' && substr($composerName, 0, 5) === 'neos/')) {
             $lastSegmentOfPackagePath = substr(trim($packagePath, '/'), strrpos(trim($packagePath, '/'), '/') + 1);
             if (strpos($lastSegmentOfPackagePath, '.') !== false) {
                 $packageKey = $lastSegmentOfPackagePath;


radmiraal:
I quickly pushed this to start the discussion already but have to get out of the door now, would be nice to have some first feedback and will then check to what branch this should've been pulled (probably 3.2)

kitsunet:
:-1: The composer class loader is perfectly capable of loading library classes as for any other third party library as well. Our class loader shouldn't and doesn't have to take care of those.

radmiraal:
Depending on my dependency order I now have issues that the classes from the schema library can not be loaded. While debugging I found that the schema package is not in the list of activePackages in the packageManager which is (if composer loads something yes or no) an obvious bug. The reason this happens is because the neos/utility-schema package has the packageKey TYPO3.Flow.Utility in the PackageStates.php. And so does neos/utility-array and the other utility packages with the TYPO3\Flow\Utility namespace in the classloader configuration.

I'm now working on a unit test that proves the issue so this change is tested. It seems like there're people that still have an issue loading classes from the Schema package so I'm not sure if it's the only issue. But this should be fixed either way...

radmiraal:
The package btw also doesn't show up in flow package:list if the order in the PackageStates.php if the bug appears... And this just looks wrong in the PackageStates ;-)

    'neos/utility-schema' => 
    array (
      'state' => 'active',
      'packageKey' => 'TYPO3.Flow.Utility',
      'packagePath' => 'Framework/Neos.Utility.Schema/',
      'composerName' => 'neos/utility-schema',
      'autoloadConfiguration' => 
      array (
        'psr-0' => 
        array (
          'TYPO3\\Flow\\Utility' => 'Classes',
        ),
      ),
      'packageClassInformation' => 
      array (
      ),
    ),

bwaidelich:
for the record: I also experience a lot of hickups with the class loader after upgrading.
Neos.* Packages are installed twice at `Packages/Framework` and `Libraries/neos` and are not properly autoloaded in testing context.

radmiraal:
ah, btw.. ClassLoader => wrong title in the PR ;-) I meant PackageManager

kitsunet:
Ah I see, fair enough, PackageManager can/should take care of that indeed. IMHO we need to get rid of those package keys. They are just bad. Composer name all the way! But I know that is a long term goal. IMHO the better solution is to add the package key to extras in the composer.json. See
https://github.com/neos/composer-plugin/pull/7

Too bad I never saw this happen otherwise we could've done this from the start.

radmiraal:
reading through the package manager I wonder why we use the first configured autoloader namespace in the derivePackageKey() method at all btw, why not skip that and just generate the packagekey based on the composer package name? What is the usecase so to say...

kitsunet:
it was backwards compatibility because htat's how it worked before as well. See comment above, I think that's the way to go.

radmiraal:
@kitsunet thanks, will have a look... but that does make it kind of fragile if we have a new package... shouldn't we protect the package manager from overwriting packagekey configuration at all?

As I now end up with either about 70 packages installed, or if I add some WIP change to take require-dev and suggest into account while ordering I suddenly only have about 50 or 60 packages installed...

kitsunet:
Sure, I am fine with changing it in the PackageManager,  IMHO the very original logic was rather bend to our universe than adapting flow to composer. I think if we say for library packages we just use the composer name to create a package key that is fine because anything not a flow package type shouldn't expect anything else anyway.

kitsunet:
So my suggest:
if (is flow package type) -> logic as it is (possibly to be changed in next major)
else -> create package key from composer name

radmiraal:
I'll play around a bit with the package manager to see if I can fix the overwrite of packages with the same key, but till then: https://github.com/neos/flow-development-collection/pull/345

kdambekalns:
Those double installs never happened for me. Sure those are not just not removed for some reason?

Also I tagged a new version of the composer plugin today and upmerged to master, if that is of any relevance.

radmiraal:
@kdambekalns The problem is a combination between the fact we just overwrite already existing configuration on this line https://github.com/neos/flow-development-collection/blob/master/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php#L991
and the usage of the class namespace (for BC reasons) in https://github.com/neos/flow-development-collection/blob/master/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php#L1206

To me it happened because I was working on taking require-dev and suggest into account in the package loading order. This put the schema package before some of the other packages with the Utility namespace, so the configuration in the activePackages was overwritten.

I tried to quickly use the composer name internally in the PackageManager, but it looks like this needs some more work than just a quick fix...

kitsunet:
Is this change still needed?

kitsunet:
Closing this, no answer, seems unnecessary

PR #345, opened 2016-05-05
BUGFIX: Class loading errors because library packageKeys are incorrect

radmiraal:
The package manager derives the packageKey from the class loader configuration
if the package type is ``library``. This leads to problems with the split
libraries that share the namespace ``TYPO3\\Flow\\Utility`` because the
package state configuration for is overwritten.

Depending on the dependency order of the packages this will lead to symptoms
like class loading errors for the ``Neos.Utility.Schema`` package.

This change adds the specific packageKey to use to the composer.json to
load the libraries correctly.

### DIFF
--- a/Neos.Utility.Arrays/composer.json
+++ b/Neos.Utility.Arrays/composer.json
@@ -21,5 +21,10 @@
     "psr-4": {
       "Neos\\Utility\\Arrays\\Tests\\": "Tests"
     }
+  },
+  "extra": {
+    "neos": {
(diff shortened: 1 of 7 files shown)


dimaip:
Weird, but I'm hitting this issue even with 2.1... https://github.com/psmb/BratstvoMovieDistr/blob/master/composer.json

jonnitto:
Tested. This bugfix needs really relesed asap.

kitsunet:
:+1:

PR #346, opened 2016-05-05
FEATURE: Use require-dev and suggest for package ordering

radmiraal:
This change adds require-dev and suggest to the package order resolver.
This allows packages to be sure they're loaded after a package
they suggest so it is still possible to overwrite configuration.
Besides that it is now no longer required to require a package that is
meant for development only

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageOrderResolver.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageOrderResolver.php
@@ -93,7 +93,12 @@ class PackageOrderResolver
         $unsortedPackages[$packageKey] = -1;
         $packageComposerManifest = $this->manifestData[$packageKey];
         $packageRequirements = isset($packageComposerManifest['require']) ? array_keys($packageComposerManifest['require']) : [];
-        // HINT: at this point, we do not support require-dev dependencies yet (but we could).
+        if (isset($packageComposerManifest['require-dev'])) {
+            $packageRequirements = array_merge($packageRequirements, array_keys($packageComposerManifest['require-dev']));
+        }
+        if (isset($packageComposerManifest['suggest'])) {
+            $packageRequirements = array_merge($packageRequirements, array_keys($packageComposerManifest['suggest']));
+        }
         $unresolvedDependencies = 0;
 
         foreach ($packageRequirements as $requiredComposerName) {
--- a/TYPO3.Flow/Tests/Unit/Package/PackageOrderResolverTest.php
+++ b/TYPO3.Flow/Tests/Unit/Package/PackageOrderResolverTest.php
@@ -31,13 +31,10 @@ class PackageOrderResolverTest extends \TYPO3\Flow\Tests\UnitTestCase
                         'name' => 'doctrine/orm',
                         'require' => ['doctrine/dbal' => 'dev-master'],
                     ],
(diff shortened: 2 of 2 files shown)


albe:
Looks good by reading.

radmiraal:
I just realised something, maybe we should skip the 'suggest' and move something like 'load-after' to the extras.neos in the composer.json.

Because do you always want to load suggest packages before your own package? I've had multiple usecases for that over the last year, but is it 'always' true... ?

aertmann:
I wondered why the suggest stuff was even needed since a suggest package would normally depend on the package that suggests it?

radmiraal:
@aertmann because it completely depends if you need to extend / overwrite the configuration of the package but don't want to require it. There's now only one possibility to overwrite config: a hard dependency.

But as there's also a definite usecase for *not* wanting this I think it makes sense to move it to the extras section of the composer.json file.

kitsunet:
Ok, thinking this though:

First of all the loading order should really only matter for flow packages, everything else cannot matter.
"suggest" is a suggestion, IMHO saying nothing about dependencies. If you need a specific order you need to use require in either of the packages that are linked though a suggest. So suggest should be ignored.

radmiraal:
I agree about leaving out suggest, and will remove that.

But I choose to disagree to "if you need a specific order you should use require". This adds a hard dependency which we found really annoying if you want to create for example an application that can work with Neos, but requires only CR. In that case you wouldn't have a chance to have the loading order of configuration defined which is a problem in some cases. This can only be fixed by adding an extra 'in between package' which is furthermore completely useless.

So what about adding a 'load-before' or whatever named configuration option to composer extras.neos like we now also have the package-key?

aertmann:
Thanks for explaining when it's useful. I agree with @radmiraal here, it's definitely a valid scenario and should probably be supported somehow.

Although a simple solution would of course be to add the dependency to your application/site package, which would normally depend on Neos or Flow as well.

kitsunet:
right, that would be a way. Also if it can work with Neos but doesn't need it shouldn't the loading order be no problem?
I just want to avoid to introduce a new config option, but if we need it I would agree, but then suggest to make ordering of dependencies deprecated and put that up as sole way to order loading of packages.

radmiraal:
downside of the whole issue is that the loading of configuration has in essence not much to do with composer... Maybe even the 'require' is an abuse of the composer feature...

But it's the easiest implementation to put this in the composer.json because it is there already...

Regarding the loading order: in our case we had issues with for example overriding the basic Neos nodetypes like TYPO3.Neos:Document. We use Neos just as a 'possible way to edit content', but only have it as a requirement if there's really also a website involved in the project (we also use the same codebase just as an api).

dfeyer:
Any update on this one ? I think @radmiraal use case is valid

radmiraal:
I was just thinking about this again last week. Currently I think we should do the following:

- deprecate usage of sorting based on require (doesn't have to happen *now*, can still support it, but in the end I think support for it should be dropped regarding the configuration loading)
- add options to the composer.json extras section to support 'load before' or 'load after' that (and make sure that this doesn't break if configured package is not installed...)

kitsunet:
See #406 which we could merge instead of this as it adds the option to configure loading order independently.

radmiraal:
@kitsunet awesome, that makes this PR obsolete imho ;-)

PR #348, opened 2016-05-06
BUGFIX: Fix ObjectAccessorNode::getPropertyPath for is/has access in PHP7

whythecode:
Related to: #108, #343

As a safety measure for PHP7, we're now catching any `\Throwable` as well.
Also added is_callable check.

- if rendered without isSomething in the arguments (implicit falsy), we then get `Call to undefined method TYPO3\Fluid\Core\ViewHelper\TemplateVariableContainer::isSomething()`

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
@@ -93,9 +93,13 @@ class ObjectAccessorNode extends AbstractNode
             try {
                 $subject = ObjectAccess::getProperty($subject, $propertyName);
             } catch (PropertyNotAccessibleException $exception) {
-                if (is_object($subject) && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) > 0) {
+                if (is_object($subject)
+                    && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) > 0
+                    && is_callable([$subject, $propertyName])) {
                     try {
                         $subject = $subject->$propertyName();
+                    } catch (\Throwable $exception) {
+                        $subject = null;
                     } catch (\Exception $exception) {
                         $subject = null;
                     }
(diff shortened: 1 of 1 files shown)


whythecode:
I integrated all of the suggestions in the previous thread, but this is probably overkill now with is_callable plus try-catch. just let me know how you all want it ;)

bwaidelich:
Thanks for taking care. Right, the double try/catch is not necessary here. And we should add a test case for this

albe:
Well, the double try/catch is due to the backwards compatibility we strived for as the direct method invocation should only occur when the property could not be accessed by ObjectAccess. I'll take care of providing a test asap.

whythecode:
no problem, @bwaidelich :)

with overkill, I didn't mean the double catch though, since that's necessary for PHP 5+7... I meant using `is_callable` *and* a try-catch... theoretically it would never get to the try-catch if it's not callable, so the try-catch is redundant.
Unless of course, one wants to make sure the function itself doesn't fail... but I think a true Exception here, coming directly from the called function, would be the clearer and more expected behaviour rather than having the variable be null.

albe:
Yeah, that's really an edge case and good to bring it up. But I guess the main rationale is rather to not break unintentionally here, because previously the method wouldn't have been called at all and null returned. Plus, it's expected Fluid behaviour that unaccessible properties/accessors silently resolve to null. But maybe @bwaidelich has a different opinion on that.

bwaidelich:
> theoretically it would never get to the try-catch if it's not callable, so the try-catch is redundant.

That's what I meant, I think this should be enough for the whole block:

```php
try {
    $subject = ObjectAccess::getProperty($subject, $propertyName);
} catch (PropertyNotAccessibleException $exception) {
    if (is_object($subject)
        && preg_match('/^(is|has)[A-Z]/', $propertyName) === 1
        && is_callable([$subject, $propertyName])) {
            $subject = $subject->$propertyName();
    } else {
        $subject = null;
    }
}
```

bwaidelich:
> it's expected Fluid behaviour that unaccessible properties/accessors silently

Right. But if a getter throws an exception that should not be disguised IMO

albe:
ah, there you go :) fine for me

whythecode:
the try-catch is now gone :)

albe:
Added a PR for your branch with a test that fails without this change.

aertmann:
@whythecode: would be great if you could merge https://github.com/whythecode/flow-development-collection/pull/1 into your branch, then we can merge this one

daniellienert:
👍 
The bug seems to effect all installations, as we use "hasCaption" uninitialized by default in https://github.com/neos/neos-nodetypes/blob/master/Resources/Private/Templates/NodeTypes/Partials/Image.html.

The fix also fixes this issue.

kdambekalns:
I just pushed #350 which is this one and whythecode/flow-development-collection#1 combined.

whythecode:
ah, sorry ppl, my computer gave up on me over the weekend :)

PR #350, opened 2016-05-08
BUGFIX: Fix ObjectAccessorNode::getPropertyPath for is/has access in PHP7

kdambekalns:
Since non-callable method call should not be attempted in the first place, this change
adds a is_callable check. Therefore a try/catch block is no longer necessary, as any exception thrown inside the callable method should bubble up.

- if rendered without isSomething in the arguments (implicit falsy), we then get
  `Call to undefined method TYPO3\Fluid\Core\ViewHelper\TemplateVariableContainer::isSomething()`

Related to: #108, #343, #348, whythecode/flow-development-collection#1

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/Core/Parser/SyntaxTree/ObjectAccessorNode.php
@@ -93,9 +93,13 @@ class ObjectAccessorNode extends AbstractNode
             try {
                 $subject = ObjectAccess::getProperty($subject, $propertyName);
             } catch (PropertyNotAccessibleException $exception) {
-                if (is_object($subject) && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) > 0) {
+                if (is_object($subject)
+                    && preg_match('/^(is|has)([A-Z].*)/', $propertyName, $matches) > 0
+                    && is_callable([$subject, $propertyName])) {
                     try {
                         $subject = $subject->$propertyName();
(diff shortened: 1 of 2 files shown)


kdambekalns:
@albe Your test failed because `null !== ''` - did you run the test after you wrote it? Because I wonder if this is actually correct (returning `null`) or if `render()` should return an empty string here…

albe:
Yep, and it passed. Rendering a view should always return a string, even if it only consists of a single variable that resolves to null, no? IIRC I even had it check for null beforehand and that failed. I'll check again tomorrow, maybe I overlooked something or some weird cache misbehaviour made it pass after that null fail before?

albe:
After a quick look it seems that indeed a View render might indeed return null. ParsingState::render is documented to return a string, but the called RootNode::evaluate resp. AbstractNode::evaluateChildNodes can possibly return null (if the tree is empty for example or, as in this case, only consists of a single node evaluating to null).
Couldn't find a test at first look that covers this behaviour, so maybe this is unintended. Do you know anything maybe @bwaidelich?

bwaidelich:
Thanks everybody for fixing this. I'll prepare a PR for master that moves the logic to `ObjectAccess` - that'll simplify & unify things and should be able to keep the new tests as is.

> After a quick look it seems that a View render might indeed return null

Right.. `AbstractTemplateView::render()` internally calls `NodeInterface::evaluate()` with an unknown return type. The reason is that e.g. a ViewHelper node can return anything you want it to.
I'd suggest to cast the result to a `string` in the View implementation. Or do you see any problems with that, @skurfuerst ?

whythecode:
thx, all!

skurfuerst:
@bwaidelich agreed, that should work I think if it is casted inside the *View*

PR #351, opened 2016-05-10
TASK: Use version of composer.json as fallback for installed version of packages

dlubitz:
In case of missing the installed version of a package, we provide the given version out of the composer.json of the package.

FLOW-455 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
@@ -465,7 +465,7 @@ class Package implements PackageInterface
     public function getInstalledVersion()
     {
         $installedVersion = PackageManager::getPackageVersion($this->composerName);
-        return  $installedVersion === '' ? $installedVersion : $this->getPackageMetaData()->getVersion();
+        return  $installedVersion !== '' ? $installedVersion : $this->getPackageMetaData()->getVersion();
     }
 
     /**
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
@@ -456,7 +456,7 @@ class Package implements PackageInterface
     }
 
     /**
-     * Get the installed package version (from composer) and as fallback the version given by package meta data.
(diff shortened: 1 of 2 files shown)


dlubitz:
See also: https://github.com/neos/neos-development-collection/pull/511

bwaidelich:
looks good (by reading). You're the man!

TYPO3.Flow/Classes/TYPO3/Flow/Package/Package.php
```diff
@@ -456,15 +456,16 @@ public function getComposerManifest($key = null)
...
      *
      * @return string
      * @api
      * TODO: Should be added to the interface in the next major Flow version (4.0)
      */
     public function getInstalledVersion()
     {
-        return PackageManager::getPackageVersion($this->composerName);
+        $installedVersion = PackageManager::getPackageVersion($this->composerName);
+        return  $installedVersion === '' ? $installedVersion : $this->getPackageMetaData()->getVersion();
```
aertmann: I'd go with

   return PackageManager::getPackageVersion($this->composerName) ?: $this->getPackageMetaData()->getVersion();

dlubitz: Thanks for the hint. But I think the type safe comparisson has less of possible side effects as @bwaidelich already mentioned: https://github.com/neos/neos-development-collection/pull/511#issuecomment-218163485

aertmann: true, although that's a bit of a stretch.. valid version numbers wouldn't cause that, only something like "0" which isn't a really valid version number

dlubitz:
Used shorthand ternary as @aertmann suggested

bwaidelich:
Makes sense. The untyped comparison only "fails" if the composer version would be "0" which should never really happen

aertmann:
thanks for taking care of this @dlubitz

kitsunet:
👎  Please don't use the  `MetaData` as it is deprecated and will be removed in the next major Flow version, rather use the Manifest directly.

dlubitz:
@kitsunet Done.

PR #352, opened 2016-05-11
BUGFIX: Packages without classes should still monitor configuration and xliff

kitsunet:
Packages without configured class files were exempted from monitoring to reduce
the amount of monitored files. This is fine for class files but monitoring of
configuration and xliff files should happen for all packages regardless if they
have configured class files or not.


soee:
Looks like it fixes my problem when NodeTypes configuration was not updated without flushing caches.

+1

robertlemke:
Okay, gotta be careful with my mobile Github app, that merge wasn't on purpose. But I did read the code and think that it will fix the bug.

kdambekalns:
Yes, looks good.

PR #353, opened 2016-05-11
BUGFIX: TypeHandling assumes vendor name to be uppercase

fcool:
The TypeHandling PARSE_TYPE_PATTERN assumes vendors in namespace do always
begin with uppercase letters. Reality shows, that this is not the case.
So this changes the regular expression accordingly

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/TypeHandling.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/TypeHandling.php
@@ -22,7 +22,7 @@ class TypeHandling
     /**
      * A property type parse pattern.
      */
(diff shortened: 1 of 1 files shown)


fcool:
Ok... Example of an unprepared pull request. The Unit-Tests fails - and if I interpret them correctly, they should fail. Nevertheless: This commit fixes a bug. ;)
*sigh* Honestly: What do no now? Negativ look aheads regex?

robertlemke:
Thanks for the Pull Request! I wouldn't say it fixes a bug (that behaviour was intentional) but if it helps with real-world needs, I'm fine with the change (when tests succeed).

fcool:
Hmmm... well. I can see it this way either... But shouldn't TypeHandling be able to deal with Packages (and classes) that do not follow Flows conventions?

albe:
If Flow weren't opinionated, yes ;) It tries to force some (sane) conventions on the developer purposefully.

What about this pattern:
`([A-Z]|[a-zA-Z0-9]+\\\\)[a-zA-Z0-9\\\\]+`

Maybe also add a test case to TypeHandlingTest::types() for that like:
`array('foo\Bar\baz', array('type' => 'foo\Bar\baz', 'elementType' => null))`

kitsunet:
the problem is that TypeHandling utility also needs to be able to deal with third party packages. AND it's a standalone package now, so 👍 from me.

kitsunet:
I would at least fix it in the 3.2 branch, not sure if we need to go back to 3.0.

aertmann:
also tests are broken

kitsunet:
This has been fixed with #441

PR #354, opened 2016-05-12
BUGFIX: Fallback if custom exception rendering fails

aertmann:
If a custom exception rendering fails, the exception would not be
caught and thus difficult to debug. Instead as a fallback exceptions
thrown during rendering will be rendered statically.


bwaidelich:
looks good (by reading)

PR #355, opened 2016-05-14
BUGFIX: Missing use statement in HelpCommandController

dfeyer:


dfeyer:
Argggg wrong branch ...

PR #356, opened 2016-05-14
BUGFIX: Missing use statement in HelpCommandController

dfeyer:
This fix a regression introduced in 3.2, the use statement for the CommandManager is missing


kdambekalns:
The commit the changed that was from May 15th, 2015… and merged December 19th. Funny no one realised this earlier.

bwaidelich:
If this would break it would have been realized, I'm sure. But it won't break for me without the fix.
The reason is that `commandManager` is already used in the parent class (`TYPO3\Flow\Cli\CommandController`).. So instead of adding the use statement we should have removed the field IMO

kitsunet:
@bwaidelich right!

dfeyer:
Based on Feedback on Discuss look like it break for PHP7 but not for PHP 5.6 ... who sounds pretty weird. But agree that we can remove the property from the HelpCommandController:

https://github.com/neos/flow-development-collection/pull/358

PR #357, opened 2016-05-17
BUGFIX: Get package with case insensitive name

daniellienert:
While `isPackageAvailable` checks the package name
case insensitive in the packageKeys array, it then
tries to return the package object assuming a correct
case sensitive package name.

If the correct pacakage key is Vendor.CustomerProject
and you request Vendor.Customerproject the check 
succeeds but returns no object.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
@@ -200,7 +200,7 @@ class PackageManager implements \TYPO3\Flow\Package\PackageManagerInterface
         if (!$this->isPackageAvailable($packageKey)) {
             throw new \TYPO3\Flow\Package\Exception\UnknownPackageException('Package "' . $packageKey . '" is not available. Please check if the package exists and that the package key is correct (package keys are case sensitive).', 1166546734);
         }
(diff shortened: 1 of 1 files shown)


aertmann:
makes sense, a test would be nice though

kdambekalns:
Hm, while this is touched anyway… There is inconsistency in the way the case of a package key is handled currently: `isPackageAvailable()` "normalizes" the key, but other methods (`isPackageActive()`, `getPackage()`, `deactivatePackage()`, `activatePackage()`,
`freezePackage()`, `isPackageFrozen()`,`unfreezePackage()`, `refreezePackage()`, `unregisterPackageByPackageKey()`, `deletePackage()`, `sortPackagesByDependencies()`) **do not** "normalize"

So IMHO we should rather not "fix" this here. People should be using the correct key and use `getCaseSensitivePackageKey()` if needed. I'd even say `isPackageAvailable()` should not normalize the key…

kitsunet:
Yes agreed, that was my original intend, actually I would be happy to get rid of the case insensitive key handling, but as I saw that might be rather difficult I didn't finish the task. IMHO `isPackageAvailable()` should not normalize.

kitsunet:
I agree to @kdambekalns, just removing the check should be fine.

PR #359, opened 2016-05-18
BUGFIX: Ignore abstract command-controllers

mficzel:
If a command controller is declared abstract the cli ignores this class because it cannot be instantiated.
The commands from derived classes will of course be available.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cli/CommandManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cli/CommandManager.php
@@ -261,7 +261,7 @@ class CommandManager
 
         $commandControllerMethodArgumentMap = [];
         foreach ($reflectionService->getAllSubClassNamesForClass(CommandController::class) as $className) {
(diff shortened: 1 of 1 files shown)


aertmann:
why 3.2? if it's a bugfix it should go to 3.0 or 2.3

mficzel:
That part was hugely refactored between 3.1 and 3.2 and the code was centralized in the CommandManager class. There was no good way to create an upmergable change.

mficzel:
On the other hand this is quite a rare case. I only stumbled over this issue when i created a bunch of importers that had some commands in common but used different service-implementation.

mficzel:
This change here introduced the current structure. https://github.com/neos/flow-development-collection/commit/feded93a4f030bf0de5d296e7a7e63b33d47ad67

kdambekalns:
For the record: fine in 3.2 for me…

aertmann:
true, although the line is exactly the same just moved.. 3.2 should be fine though

PR #360, opened 2016-05-20
Add a note to rescan packages to Signals and Slots

koernchen02:
I'm not sure if this should make it in the docs, since https://github.com/neos/flow-development-collection/pull/341 is already pending, but if the auto-detection is not implemented soon, I would vote for it, since a cache flush does not get rid of the problem and beginners working with the docs could really have a hard time figuring this out!


kdambekalns:
Totally cool with that, but two wishes:

- wrap the line at about the same length as the existing lines
- add that this only affects 3.2.0 to 3.2.2

Thanks!

koernchen02:
I'm thankful for the insights, and now I totally get, that we don't want to have notes in the docs that are only relevant because of a regression, which itself was only relevant for a very short timeframe...

Best,
Christian!

kdambekalns:
Ok, so let's make sure #341 is merged for 3.2.3 :)

PR #361, opened 2016-05-25
BUGFIX: Clear opcode cache before compiling proxy classes

kdambekalns:
Fixes a bug with statcaches that would lead to Flow trying to delete
configuration files that no longer existed.


kdambekalns:
My analysis shows @kitsunet is involved, this has been pulled from https://github.com/neos/flow-development-collection/pull/75

kitsunet:
Yep, that definitely makes sense.

PR #362, opened 2016-05-27
BUGFIX: Avoid type-casting errors if the validation of complex types fails

mficzel:
If an error is created in some cases the value was added to the result-message without checking for the type beforehand, that could lead to php errors if the value could not be casted to a string.

This change centralizes the rendering of the value in a  'renderValue' - method.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/SchemaValidator.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/SchemaValidator.php
@@ -78,7 +78,7 @@ class SchemaValidator
                 }
             }
             if ($isValid === false) {
-                $result->addError($this->createError('None of the given schemas matched ' . $value));
+                $result->addError($this->createError('None of the given schemas matched ' . $this->renderValue($value)));
             }
         } elseif ($this->isSchema($schema)) {
             if (isset($schema['type'])) {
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Utility/SchemaValidator.php
```diff
@@ -571,6 +571,21 @@ protected function validateAnyType($value, array $schema)
     }
 
     /**
+     * Create a string information for the given value
+     *
+     * @param $value
```
aertmann: mixed

TYPO3.Flow/Classes/TYPO3/Flow/Utility/SchemaValidator.php
```diff
@@ -571,6 +571,21 @@ protected function validateAnyType($value, array $schema)
...
+     * @return string
+     */
+    protected function renderValue($value)
+    {
+        if (is_scalar($value)) {
+            return (string)$value;
+        } else {
+            return 'type=' . gettype($value);
+        }
```
aertmann: could be a one liner
`return is_scalar($value) ? (string)$value : 'type=' . gettype($value);`

mficzel:
@aertmann adjusted the lines marked

PR #363, opened 2016-05-27
!!!TASK: Deprecate unused methods in ``Argument`` class

kitsunet:
Some methods haven't been removed or deprecated earlier but are in fact
no longer used in the Flow core and therefore deprecated to be removed
in the next major Flow version.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/Argument.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/Argument.php
@@ -29,12 +29,14 @@ class Argument
     /**
      * Short name of this argument
      * @var string
+     * @deprecated Will be removed for next major Flow version.
      */
     protected $shortName = null;
 
     /**
      * Short help message for this argument
      * @var string
+     * @deprecated Will be removed for next major Flow version.
      */
     protected $shortHelpMessage = null;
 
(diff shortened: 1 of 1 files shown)


kitsunet:
@kdambekalns

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/Argument.php
```diff
@@ -148,6 +152,7 @@ public function getShortName()
      * Sets the data type of this argument that is also used for property mapping.
      * @param string $dataType
      * @return \TYPO3\Flow\Mvc\Controller\Argument $this
+     * @deprecated Will be removed for next major Flow version. Set the DataType via constructor.
```
kdambekalns: Well, in a current project the client uses exactly that, since the action has an abstract class as type hint. The concrete type to use is set in `initializeAction()` using this method.

kdambekalns:
Aside from my comment on `setDataType()` this is obvious, IMHO.

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/Controller/Argument.php
kitsunet: Can't you just create a new Argument object then?

kdambekalns: Yes, that's an option, of course.

PR #365, opened 2016-05-29
BUGFIX: After creating a package all packages should be sorted

kitsunet:
After creating a package only a rescan of existing packages was triggered
but not a resorting. That would lead to arbitrary package loading order
which only depended on the way the filesystem would find the packages.
This has been changed to trigger a full rescan including sorting of packages.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
@@ -424,18 +424,11 @@ class PackageManager implements PackageManagerInterface
 
         $manifest = ComposerUtility::writeComposerManifest($packagePath, $packageKey, $manifest);
 
-        $packagePath = str_replace($this->packagesBasePath, '', $packagePath);
-        $package = $this->packageFactory->create($this->packagesBasePath, $packagePath, $packageKey, $manifest['name'], (isset($manifest['autoload']) ? $manifest['autoload'] : []), null);
-
-        $refreshedPackageStatesConfiguration = $this->scanAvailablePackages($this->packageStatesConfiguration);
-        $this->savePackageStates($refreshedPackageStatesConfiguration);
+        $refreshedPackageStatesConfiguration = $this->rescanPackages(false);
         $this->packageStatesConfiguration = $refreshedPackageStatesConfiguration;
+        $this->registerPackageFromStateConfiguration($manifest['name'], $this->packageStatesConfiguration['packages'][$manifest['name']]);
 
-        $this->packages[$packageKey] = $package;
-        $this->activePackages[$packageKey] = $package;
-        $this->packageKeys[strtolower($packageKey)] = $packageKey;
-
-        return $package;
+        return $this->packages[$packageKey];
     }
 
     /**


TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
```diff
@@ -427,13 +427,9 @@ public function createPackage($packageKey, \TYPO3\Flow\Package\MetaData $package
         $packagePath = str_replace($this->packagesBasePath, '', $packagePath);
         $package = $this->packageFactory->create($this->packagesBasePath, $packagePath, $packageKey, $manifest['name'], (isset($manifest['autoload']) ? $manifest['autoload'] : []), null);
 
-        $refreshedPackageStatesConfiguration = $this->scanAvailablePackages($this->packageStatesConfiguration);
-        $this->savePackageStates($refreshedPackageStatesConfiguration);
+        $refreshedPackageStatesConfiguration = $this->rescanPackages(false);
```
kitsunet: Not only scan and save but also sort inbetween... ;)

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
```diff
@@ -427,13 +427,9 @@ public function createPackage($packageKey, \TYPO3\Flow\Package\MetaData $package
...
         $this->packageStatesConfiguration = $refreshedPackageStatesConfiguration;
-
-        $this->packages[$packageKey] = $package;
-        $this->activePackages[$packageKey] = $package;
-        $this->packageKeys[strtolower($packageKey)] = $packageKey;
+        $this->registerPackageFromStateConfiguration($manifest['name'], $this->packageStatesConfiguration['packages'][$manifest['name']]);
```
kitsunet: more encapsulated way of doing the three lines that were removed.

kitsunet:
I consider this a no-brainer as the problem is rather obvious...
Will merge and release later today if no one objects.

kitsunet:
The style CI failures are not even in files I touched... Ignoring.

kitsunet:
The test failures were correct... Fixed.

aertmann:
thanks for taking care @kitsunet, verified 👍

TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php
```diff
@@ -424,18 +424,11 @@ public function createPackage($packageKey, \TYPO3\Flow\Package\MetaData $package
 
         $manifest = ComposerUtility::writeComposerManifest($packagePath, $packageKey, $manifest);
 
-        $packagePath = str_replace($this->packagesBasePath, '', $packagePath);
-        $package = $this->packageFactory->create($this->packagesBasePath, $packagePath, $packageKey, $manifest['name'], (isset($manifest['autoload']) ? $manifest['autoload'] : []), null);
-
-        $refreshedPackageStatesConfiguration = $this->scanAvailablePackages($this->packageStatesConfiguration);
-        $this->savePackageStates($refreshedPackageStatesConfiguration);
+        $refreshedPackageStatesConfiguration = $this->rescanPackages(false);
```
ngocanh1233: dsadasdasa

PR #366, opened 2016-05-31
FEATURE: Allow use of PDO backend for caches

malteriechmann:
This pull request contains two features:

1. Make PdoBackend iterable so it can be used for caches
2. Use default filter expression to ignore other packages when generating migrations

I tested this with PostgreSQL as a backend for sessions.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/DoctrineCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/DoctrineCommandController.php
@@ -419,6 +419,16 @@ class DoctrineCommandController extends CommandController
      *
      * would only create a migration touching tables starting with "acme_com".
      *
+     * It is also possible to set a default filter expression within the settings.
+     *
+     * TYPO3:
+     *   Flow:
+     *     persistence:
+     *       doctrine:
+     *         migrations:
+     *           generate:
+     *             defaultFilterExpression: '/^acme_com/'
+     *
      * @param boolean $diffAgainstCurrent Whether to base the migration on the current schema structure
      * @param string $filterExpression Only include tables/sequences matching the filter expression regexp
      * @return void
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/DoctrineCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/DoctrineCommandController.php
@@ -436,6 +436,11 @@ class DoctrineCommandController extends CommandController
             $this->quit(1);
         }
 
+        // use default filter expression from settings
+        if (isset($this->settings['doctrine']['migrations']['generate']['defaultFilterExpression']) && $filterExpression === null) {
+            $filterExpression = $this->settings['doctrine']['migrations']['generate']['defaultFilterExpression'];
+        }
+
         list($status, $migrationClassPathAndFilename) = $this->doctrineService->generateMigration($diffAgainstCurrent, $filterExpression);
 
         $this->outputLine('<info>%s</info>', [$status]);
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/PdoBackend.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/PdoBackend.php
@@ -10,7 +10,6 @@ namespace TYPO3\Flow\Cache\Backend;
  * information, please view the LICENSE file which was distributed with this
  * source code.
  */
-
 use TYPO3\Flow\Annotations as Flow;
 
 /**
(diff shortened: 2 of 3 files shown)


kdambekalns:
It would be nice to have two different PRs for those two features… and about the StyleCI failure, no idea why that happens.

malteriechmann:
> It would be nice to have two different PRs for those two features…
Would that be nice or is this an instruction?

kdambekalns:
It's "Would be nice". Maybe for the next PR :)

But now for an "instruction"… please mention the setting in the documentation for the `migrate` command…

malteriechmann:
What can I do to speed this up?

bwaidelich:
dear trigger happy folks ;)
Please make sure that new configuration options are always documented and covered by a corresponding change to the schema files:

```
./flow configuration:validate
Validating all configuration

1 errors were found:
 - Settings.TYPO3.Flow.persistence.doctrine.migrations -> This property is not allowed here, check the spelling if you think it belongs here.
```

TYPO3.Flow/Classes/TYPO3/Flow/Command/DoctrineCommandController.php
```diff
@@ -436,6 +446,11 @@ public function migrationGenerateCommand($diffAgainstCurrent = true, $filterExpr
             $this->quit(1);
         }
 
+        // use default filter expression from settings
+        if (isset($this->settings['doctrine']['migrations']['generate']['defaultFilterExpression']) && $filterExpression === null) {
```
bwaidelich: the second check is obsolete because
```isset(null) === false```

**Comment by bwaidelich** [comment]
I'd push a follow-up adding the modified schema but I'd like to extend this (btw quite useful) feature because as is it is not very reliable because every package setting that flag will *override* the setting for others. That's why we started switching to `dictionaries` with `true` as value so that adding/overriding gets more explicit (see #130 for a related example)..
In this case I'd like to go even a step further and allow to hook into this somehow: The Doctrine JobQueue implementation I'm currently working on will by default create a table per queue and it would be more than useful to be able to skip those from automatic generated migraions..
@kitsunet @kdambekalns any opinions on that?

**Comment by kdambekalns** [comment]
Sorry about the config schema… we should really make that part of the Travis CI builds…

About this not being a dictionary: it didn't even occur to me this might come packaged with some code. I (wrongly) assumed everyone would see it like I did: "Oh, I can set that in `Configuration/Development/Settings.yaml` for my projects, awesome!"…

Making this more useful? Hell, yeah! ;)

**Comment by bwaidelich** [comment]
> Sorry about the config schema

No worries, happens to me all the time

> we should really make that part of the Travis CI builds…

Yes.. A simple functional test might be enough.. I'll give this a shot later today

> Making this more useful? Hell, yeah! ;)

I'll have a look (and might ping you for some doctrine related questions along the way)

**Comment by kitsunet** [comment]
Just please lets have #231 first ;) It's a PITA to get that rebased/merged.

**Comment by kdambekalns** [comment]
@bwaidelich Travis could just run `configuration:validate` and check the output.

**Comment by bwaidelich** [comment]
even better!

**Comment by bwaidelich** [comment]
@kitsunet re
> Just please lets have #231 first ;)
I agree that we should tackle this, but it's not really related/in conflict with doctrine migration filter, is it?
If there are no objections I'd prepare a PR that:
* fixes the schema
* adds corresponding documentation
* fixes indention of the command help that is a bit hard to understand right now:
image

I'd also try to make the configuration "multi-tenant-capable" by turning it into some kind of dictionary.. But looking at the code I was wondering about two things in particular:

1. wouldn't this mostly be used to *exclude* certain tables? So shouldn't at least the configuration be a *blacklist* rather than a *whitelist*?  (I know we can use negative look aheads in the regex, but..)
2. It seems the filter is ignored for other doctrine commands like `./flow doctrine:update`. Also: wouldn't the `DoctrineService` be a better place to enforce the defaults?

**Comment by bwaidelich** [comment]
Suggestion for the settings (in sync with `debugger.ignoredClasses`):

Instead of:
```yaml
...
      doctrine:
        migrations:
          generate:
            defaultFilterExpression: '/^some_expression$/'
```
->

```yaml
...
      doctrine:
        migrations:
          ignoredTables:
            'autogenerated_.*': TRUE
```

kdambekalns:
@bwaidelich 

>wouldn't this mostly be used to exclude certain tables?

No, it would be used to only include stuff from your own vendor name… ;)

> It seems the filter is ignored for other doctrine commands

Yes, it says `migrate` - and other command could (would?) lead to a DB state beyond repair if not done consistently on the full set of needed changes.

kdambekalns:
Agreed! Now let's stop hijacking this PR thread… ;)

PR #367, opened 2016-06-01
TASK: Generated database migrations are PSR-2 compatible

gerdemann:
This task makes generated database migrations compatible with
the PSR-2 coding style guide.

--

Just a question: Is it okay to use here the same commit message? (see https://github.com/neos/flow-development-collection/commit/d47f7114b7f27989100d3ee1e09050d855221490)


kitsunet:
Just fine :)

PR #375, opened 2016-06-02
TASK: Followup cleanup for #333 to make use of ::class

albe:
This change fixes the usage of class names as strings that were necessary in 2.3 branch for PHP backwards compatibility in #333 and replaces them with the recommended ::class calls.


albe:
Hm, Travis failure seems (and should be) unrelated...

PR #376, opened 2016-06-02
TASK: Provide proper error message when no request handler can be resolved

albe:
Previously Flow would just fail with an error message:
`Warning: ksort() expects parameter 1 to be array, null given`
when for some reason no suitable request handler can be found for the current request.
This change will instead throw a useful exception and hint the user at a setup problem.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
@@ -422,6 +422,9 @@ class Bootstrap
                 $suitableRequestHandlers[$priority] = $requestHandler;
             }
         }
+        if (empty($suitableRequestHandlers)) {
+            throw new FlowException('No suitable request handler could be found for the current request. This is most likely a setup-problem, so please check your package.json and/or try removing Configuration/PackageStates.php', 1464882543);
+        }
         ksort($suitableRequestHandlers);
         return array_pop($suitableRequestHandlers);
     }


TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
```diff
@@ -422,6 +422,9 @@ protected function resolveRequestHandler()
                 $suitableRequestHandlers[$priority] = $requestHandler;
             }
         }
+        if (empty($suitableRequestHandlers)) {
+            throw new FlowException('No suitable request handler could be found for the current request. This is most likely a setup-problem, so please check your package.json and/or try removing Configuration/PackageStates.php', 1464882543);
```
bwaidelich: Does `FlowException` actually resolve here? Did anyone test this?

albe: Hm, there's indeed no import of the namespace here - I was damn sure the class was imported, actually I always auto-complete classes in IntelliJ to make sure I don't forget. Only thing I could imagine is I first added this change on master, then rebased to 2.3 and didn't notice that 2.3 didn't import the FlowException yet. Also tests didn't break because there is no test that simulates an unresolved RequestHandler. Argh, having a bad commit streak this week :(

bwaidelich:
@albe This seems to break 2.3, because `FlowException` is not imported there:
image

TYPO3.Flow/Classes/TYPO3/Flow/Core/Bootstrap.php
bwaidelich: No worries, happens to all of us ;)

bwaidelich:
Do you take care of a follow-up?

bwaidelich:
@albe *ping*. Will you be able to take care of this one? otherwise I'll open a ticket so we won't forget about it

kitsunet:
@bwaidelich he is in holidays

PR #377, opened 2016-06-03
!!! TASK: More consistent translation behavior

bwaidelich:
This patch adjusts the behavior of id-based translations in order to make it deterministic and consistent.

It mainly adjusts `Translator::translateById()` to return `NULL` if the given `id` couldn't be translated. Previously it returned the `id` making it impossible to detect that case.

This has an effect to the two related Fluid ViewHelpers in certain conditions:

`{f:translate(id: 'some.id', value: 'default')}`:

previously returned the *translated* default value if the `id` wasn't translated, possibly leading to unexpected behavior.
Now it will just return the raw default value ("default" in the example above instead of "translatedDefault").

Besides the `TranslateViewHelper` has been tweaked to always return an empty string rather than `NULL` if neither id nor value could be resolved.

`<f:form.select translate="{by:'id', by:'value', prefix: 'prefix.'}" />`:

Previously this would render

`<option value="...">prefix.someValue</option>`

if the id could not be translated.
With `by:'label'` it would return the prefixed label instead.

Now we always render the actual key (= label) of the option in that case:

`<option value="...">Original Option</option>`

FLOW-456 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/I18n/EelHelper/TranslationParameterToken.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/I18n/EelHelper/TranslationParameterToken.php
@@ -178,10 +178,8 @@ class TranslationParameterToken implements ProtectedContextAwareInterface
         }
 
         $translation = $this->translator->translateById($id, $arguments, $quantity, $locale, $source, $package);
-        if ($translation === $id) {
-            if ($value) {
-                return $this->translator->translateByOriginalLabel($value, $arguments, $quantity, $locale, $source, $package);
-            }
+        if ($translation === null && $value !== null) {
+            return $this->translator->translateByOriginalLabel($value, $arguments, $quantity, $locale, $source, $package);
         }
 
         return $translation;
--- a/TYPO3.Flow/Classes/TYPO3/Flow/I18n/Translator.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/I18n/Translator.php
@@ -159,7 +159,7 @@ class Translator
      * @param \TYPO3\Flow\I18n\Locale $locale Locale to use (NULL for default one)
      * @param string $sourceName Name of file with translations, base path is $packageKey/Resources/Private/Locale/Translations/
      * @param string $packageKey Key of the package containing the source file
-     * @return string Translated message or $labelId on failure
+     * @return string Translated message or NULL on failure
      * @api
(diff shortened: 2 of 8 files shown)


albe:
Makes sense, will take a closer look asap, but couldn't find anything on first sight.

Only this

> Now we always render the actual key (= label) of the option in that case:

is a bit confusing: isn't the key (in the select options iterable) the value of the option tag? Or which "key" are you referring to?

bwaidelich:
> isn't the key (in the select options iterable) the value of the option tag?

This *is* quite confusing indeed. I changed that to just "label" now.
But the example should make it clear

kitsunet:
currently in most projects I use only `{f:translate(id: 'some.id')}`
That would now return empty string instead of `some.id`, right? IMHO that is rather breaking at least for me.

bwaidelich:
@kitsunet Thanks for the feedback.

> currently in most projects I use only {f:translate(id: 'some.id')}
> That would now return empty string instead of some.id, right?

No, that was maybe a little bit confusing. Only the lowlevel `Translator::translateById()` will now return `NULL` for an unresolved ID so that we can distinguish this case.
The Translate-ViewHelper behaves like before with following exceptions:

* It will return an empty string when it returned NULL before (basically if neither *id* nor *value* are specified)
* It won't fallback to the *translated* value, but to the one specified: `{defaultLabel -> f:translate(id: 'some.id')}` won't try to translate the *defaultLabel* if *some.id* can't be translated - IMO it's a conscious decision to go for *id*- vs *label*-based translation and we should not automagically switch between them as this could result in unexpected behavior.

You can see a complete comparison of the behavior in the [ticket description](https://jira.neos.io/browse/FLOW-456#f%3Atranslate) and the diff for the VHs is also pretty simple: https://github.com/bwaidelich/flow-development-collection/commit/a0ec1df2b45e0b340e19977eff91c91af2ce32fa#diff-ebae97d1889485be90da5e2d0aa41e84

kitsunet:
Sounds good then, thanks for the explanation :) 👍

bwaidelich:
*ping* would be nice to have this in 3.3lts. also it's required for https://github.com/neos/form/pull/12

kitsunet:
Has conflicts, but when those are solved we could merge IMHO

bwaidelich:
@kitsunet thanks for the pointer. I fixed the conflicts, lets see what travis has to say

bwaidelich:
looking good. anyone up for a last validation?

bwaidelich:
*ping*

PR #379, opened 2016-06-07
TASK: Clarify positioning of arguments and options in CommandController

lorenzulrich:
In the method signature, all (mandatory) arguments must be placed in front of the options. If not, all options placed in front of the last argument become mandatory, too in the command line prompt. This should be stated explicitely in the documentation.


kitsunet:
IMHO that's a question of best practice and doesn't need to be stated. The PHP docs already say it's best practice to do that for all argument declarations. I would still be ok merging this...

lorenzulrich:
I would prefer to merge it for the following reason: If not following best practise, the command line states ```Please specify the required argument "foo":``` which is not correct.

albe:
True that it is a rather generic thing to know about php specifics, but then OTOH the little additional sentence doesn't hurt in the documentation. So :+1: from me

PR #380, opened 2016-06-08
BUGFIX: Improving argument handling of FormatResolver::resolvePlaceholders

fcool:
If placeholders will be replaced by arguments containing chars "{}",
replacing led to an exception.

This fix solves this behaviour by not visiting replaced content again.
This helps a lot translating error messages from Validators which give
their content back to their error messages - and can contain literally
anything.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/I18n/FormatResolver.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/I18n/FormatResolver.php
@@ -99,9 +99,9 @@ class FormatResolver
         if ($locale === null) {
             $locale = $this->localizationService->getConfiguration()->getDefaultLocale();
         }
-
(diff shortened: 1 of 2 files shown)


albe:
Looks good, but shouldn't this maybe be targeted even lower? It's a bugfix after all

fcool:
Well I think it is a bugfix - but if someone would have relied on being able to introduce new placeholders through arguments... Of course one should not, but...

If you think so, I would love to lower the target.

albe:
Well, IMO makes not much sense to only target 3.2+ either way ;)

fcool:
So close this one and start another one? Or ist there a away to thange the base here?

fcool:
So close this one and start another one? Or ist there a away to thange the base here?

albe:
Unfortunately not, need to start a new PR

fcool:
Continued there: https://github.com/neos/flow-development-collection/pull/383

PR #383, opened 2016-06-09
BUGFIX: Improving argument handling of FormatResolver::resolvePlaceholders

fcool:
If placeholders will be replaced by arguments containing chars "{}",
replacing led to an exception.

This fix solves this behaviour by not visiting replaced content again.
This helps a lot translating error messages from Validators which give
their content back to their error messages - and can contain literally
anything.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/I18n/FormatResolver.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/I18n/FormatResolver.php
@@ -100,8 +100,9 @@ class FormatResolver
             $locale = $this->localizationService->getConfiguration()->getDefaultLocale();
         }
 
-        while (($startOfPlaceholder = strpos($textWithPlaceholders, '{')) !== false) {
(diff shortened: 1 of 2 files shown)


albe:
Thx for the retargeting. Looks good for me.

kitsunet:
I am not really happy with the assignments in the `while`

fcool:
Ok. Do you have a "constructive" suggestion? I must admit, that I am too stupid to come with an alternative myself.

albe:
I think christian is referring to the startOfPlaceholder assignment itself. It's generally best practice to avoid assignments within conditions. To be fair though, the assignment was already there before.

Best option without completely rewriting the logic would be to just duplicate the `$startOfPlaceholder`  `strpos` once outside the while.

skurfuerst:
Hey everybody, 

just talked again with @kitsunet and we're both fine with merging it as is :) And then we can refactor at a later point if needed -- great that it has tests.

All the best,
Sebastian

PR #384, opened 2016-06-12
BUGFIX: Query count with groupBy and join is off

albe:
FLOW-104 #close


kdambekalns:
Now the failing Travis test actually shows a small, albeit real issue with this change… ;)

albe:
This could actually be related to #416

albe:
Ah, no that's nonsense. #416 only relevant for Query model

kitsunet:
Is this still relevant?

albe:
If the test still fails, yes. Guess a rebase would be good, but I doubt the behaviour changed.

PR #385, opened 2016-06-13
FEATURE: Allow excluding paths from scanning for available locales

albe:
This change introduces the configuration setting "TYPO3.Flow.i18n.scan.excludePaths", which is a
dictionary list of regex patterns as keys. Paths inside all packages Resources directories that match
the patterns that are enabled, will be skipped, therefore greatly improving performance when
lots of folders and files exist without any localized filenames, like `node_modules` or `.git`.

By default the following folders will be skipped for scanning:
 - '/node_modules'
 - '/.git'
 - '/CLDR/Sources'

FLOW-376 #close


dlubitz:
Looks good by reading. One little nitpick.
To be honest, I don't get the unit test. Where is any assertation of the resulting/scanned folders?

TYPO3.Flow/Classes/TYPO3/Flow/I18n/Service.php
```diff
@@ -279,6 +281,9 @@ protected function generateAvailableLocalesCollectionByScanningFilesystem()
             $directories = array(Files::getNormalizedPath($packageResourcesPath));
             while ($directories !== array()) {
                 $currentDirectory = array_pop($directories);
+                if ($blacklistPattern !== '' && preg_match('#' . str_replace('#', '\#', $blacklistPattern) . '#', $currentDirectory) > 0) {
```
dlubitz: Why not simply `preg_match('#' . str_replace('#', '\#', $blacklistPattern) . '#', $currentDirectory) == true`?

bwaidelich: `... === 1`

albe:
@dlubitz The test checks that there are only 2 locales added to the available, instead of the 4 it would find if the "node_modules" folder were scanned too.

TYPO3.Flow/Classes/TYPO3/Flow/I18n/Service.php
dlubitz: They had no boolean return values leftover on the day they implemented the 'preg_match' method in PHP. *facepalm*

dlubitz:
@albe: I see. Thanks.

aertmann:
why not make use of the existing `.flowFileMonitorIgnore` concept? see http://gerrit.neos.io/#/c/37487/

see https://jira.neos.io/browse/NEOS-563 and https://jira.neos.io/browse/NEOS-111

albe:
Didn't know that existed, thanks for the hint! But tbh I'm not sure it's the best idea to add an additional file check as that will firstly degrade performance (I'm on windows, the filesystem is not my friend ;)). Also one of the reasons why it's suggested to avoid using apaches ".htaccess" functionality.

Also, this is not really about FileMonitor, but rather a very specific file scan for filenames matching a pattern, so I'm also not sure if mixing contexts is best option.

I see that having multiple ways of ignoring dictories for different scannings is suboptimal though.

What do the others think?

albe:
Still probably makes sense to add "/bower_components", "/.sass-cache" and "/.svn" to the defaults. Maybe even exclude all folders that start with a `.` by default?

kitsunet:
How about we white list folders? I mean the convention is `Translations` right? Ofc that would be a breaking change for next major, but I would like that much more.

albe:
Translations and localized assets. We could go for whitelisting the Resources/Public (no one would put a node_modules or sth there, right?) and Resources/Private/Translations by default, then allowing to add other folders via configuration.

kitsunet:
That could work. This `Resources/Public` makes me a bit nervous though....

albe:
At least needs documentation to avoid placing dependency managers/build tools/vcs/etc into `Resources/Public`, yes. It won't make anything worse though, but will work nicely with a sane project structure. 
Or are there really any valid use cases of having such folders in Public?
In that case we could go for "whitelist via configuration, blacklist via .*ignore files".

aertmann:
There will certainly be people checking out npm and bower modules in `Resources/Public`, why wouldn't there? that's would be perfectly "sane" in cases where external resources are needed by the client..

Didn't even know about this feature, but it seems like a poor default behavior that almost no one benefits from and everyone pays a unnecessary performance penalty for. Would even suggest to disable it by default instead.

Regarding whitelisting, I agree it should be `Resources/Private/Translations/` by default. Even if disabled the feature by default. Then those few who actually use the feature can configure it to their needs.

albe:
Hm, I always assumed one would use some bundling tool when using npm/bower anyway, since a lot of node_modules are not directly browser-compatible. In that case it only makes sense to keep bower/npm in `Resources/Private` and set the build target inside `Resources/Public` (that's how I do it currently). Maybe we should document some best-practice for setting up a Package with npm/bower for managing scripts.

Not sure if disabling the scan is doable, as else the availableLocales/localeCollection will be empty, so most of i18nService functionality will not work (getParentLocale/findBestMatchingLocale/etc.). So I'd say let's go for whitelist + .ignore?
In that case my main question would be: What name should the ignore file have? Really `.flowFileMonitorIgnore`?

dlubitz:
We also work with npm just in `Resource/Private`. I don't like the `.ignore` idea, so I have to put e.g the `node_modules` folder to the repostiory, just for adding the `.ignore` file. 

What's wrong with the initital idea of just blacklisting the pattern/folders we really don't need to browse for i18n content? Performance is not really an issue: AFAIK we browse this just on first call/compile of Flow.

albe:
@dlubitz:
> Performance is not really an issue: AFAIK we browse this just on first call/compile of Flow.

Right, only on compile, but then again this whole change is only about optimizing this step.

>  I don't like the .ignore idea, so I have to put e.g the node_modules folder to the repostiory, just for adding the .ignore file.

You'd only need the `.ignore` file if you have `node_modules` (or similar heavy folder) in the whitelisted path (e.g. `Resources\Public`).


So any further opinions on this? I'm not totally decided yet, but currently leaning towards the **whitelist** aproach.

But I would also be fine with the current blacklist aproach, though it has the drawback of having to define sane defaults, which might not catch all cases or would need to change easily in the future.

The whitelist aproach OTOH has the benefit of the configuration being more stable to changes in project structure (ie. uprising of new tooling), but would introduce a "recommended" structure of keeping tools/package managers outside the `Resources\Public` folder, which would need some documentation (I would deliver that). Also, an option to further exclude single folders from the whitelist would be required for special cases, e.g. the `.*ignore` file (`.flowI18nIgnore` maybe?), which in turn would need to commit such folders to VCS as @dlubitz stated. Not sure that's that much of a drawback considering the specialness of the use case though.

@aertmann 
> that would be perfectly "sane" in cases where external resources are needed by the client..

Is this really a common use of npm/bower to have all the modules in public web folder and browser loads each dependency individually? Do you have an example setup?
I'm not the very proficient with those scripting tools yet, but everything I read up until now ended in having something like webpack/browserify for bundling all those dependencies into singular build files that the browser would load then. Those can easily be configured to land in `Public` then.

Still, even if someone really decides for whatever reason to put bower/npm into `Public`, he would at most have a little longer compile step (still not worse than currently), or need to place the `.*ignore` file there.

aertmann:
@albe: I've done it personally and seen others do it as well, so yes. We cannot assume people use a build system and put stuff in `Private` and compile that to `Public`. For a simple project, including s.th. like Bootstrap using bower directly in `Public` makes sense.

The conventions for `Public` and `Private` are very vague so I don't think it's a good idea to make such assumptions.

albe:
Ok. So can we agree on implementing a whitelist/`.*ignore` combination?

albe:
Ok, so I now implemented this as a combined white- and blacklist, as it just makes sense.
That way only folders that are whitelisted (setting TRUE) and *not* blacklisted (no pattern with setting FALSE matches) will actually be scanned.

However, there are two breaking behaviours to be solved:

  - `Unit\I18n\ServiceTest::initializeCorrectlyGeneratesAvailableLocales` expects the `Private` folder to be scanned and now fails

  - The folder structure in `Public\Translations` will no longer generate available locales, as those are only retrieved from file names and not folders. Therefore the functional `TranslatorTest::simpleTranslationByIdWorks` fails now. The reason it didn't fail before was that the "de_CH" locale was made available through the CLDR resources, which are now no longer scanned.

IMO the latter should actually be fixed in the scanner to also allow locale folder names.

So how should we proceed? Make that change fully b/c by scanning all of Resources by default and only excluding some common patterns (node_modules, bower_components, dot-folders) or make the change breaking?

From a semver POV we'd need to make it b/c for 3.3, and could then restrict the scanned paths further for next major.

TYPO3.Flow/Configuration/Settings.yaml
```diff
@@ -238,6 +238,20 @@ TYPO3:
...
+        # Example:
+        # '/node_modules': TRUE
+        # '/src/.*/test': TRUE
+        #
+        # Note: Setting a pattern to FALSE will not whitelist it, but is an option for packages to remove patterns
+        #       from this blacklist again.
+        excludePaths:
+          '/CLDR/Sources': TRUE
+          '/node_modules': TRUE
+          '/\.git': TRUE
```
aertmann: might as well add `bower_components` here, as mentioned it's not unlikely to helpful to some and won't hurt anyone

aertmann:
sounds good to me making it b/c for 3.3 since we can make 4.0 changes rather soon anyway so it's likely to happen..

dfeyer:
By reading it look good, maybe we can have small documentation about this ?

TYPO3.Flow/Classes/TYPO3/Flow/I18n/Service.php
```diff
@@ -279,6 +281,9 @@ protected function generateAvailableLocalesCollectionByScanningFilesystem()
             $directories = array(Files::getNormalizedPath($packageResourcesPath));
             while ($directories !== array()) {
                 $currentDirectory = array_pop($directories);
+                if ($blacklistPattern !== '' && preg_match('#' . str_replace('#', '\#', $blacklistPattern) . '#', $currentDirectory) === 1) {
```
dlubitz: Maybe you can move the part building and escaping the pattern up to the other pattern building code (L271).
Would make this line easier to read, and it is also not needed per directory.

dlubitz:
Just one little nitpick. Rest looks good by reading. Would also vote for some documentation, maybe here:
https://github.com/neos/flow-development-collection/blob/master/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Internationalization.rst#file-locations-and-naming
Thanks @albe

TYPO3.Flow/Classes/TYPO3/Flow/I18n/Service.php
albe: Funny, that's exactly what I already did today :)

TYPO3.Flow/Configuration/Settings.yaml
albe: Already done :) will push my latest tomorrow

albe:
Already have a little documentation snippet in my branch. Will make it b/c and push tomorrow. Then add another dependant PR that just restricts the scan paths further for next major. Thanks for all your feedback!

kitsunet:
Looks quite good already. I do think your current master is outdated though? The test failures should already be fixed.

albe:
Ugh, that styleCI is surely pedantic about anonymous functions... Everything should be fine now.

Should definitely be squashed for merge.

albe:
Ok, there is still one small issue currently: The whitelist will not work when it is more than one level deep, because the scan iteratively traverses the folders, so something like '/Private/Translations/' will not match on the '/Private/' folder and hence skip it, not even ending up in '/Private/Translations/' in the first place.

albe:
Ok, this should be it now. I tried with still iterating folders that did not match the whitelist patterns and only skip adding the locale, but that would not lead to the desired performance gain expected with explicit whitelisting.

So I redid this to have now two settings: 
- `scan.includePaths` is a list of actual folder names inside `Resources` that should exclusively be scanned
- `scan.excludePatterns` as a list of regex patterns that further avoid descending into matching folders during scanning

This will now work nicely with e.g. only scanning `Private/Translations/` instead of the full `Private` Resources.

Please review!

kitsunet:
Noooooo, white AND blacklist is always a hassle. I mean how is the priority in that case, what does the whitelist maybe disallow from third-party packages, can I combine both. Do we really need an additional whitelist.

albe:
There's no "priority" involved actually - only included (whitelist) *paths* are actually scanned through, and those are then checked against the exclusion pattern. That's why I named them differently,

We could of course replace that with the `.*ignore` file thingy discussed earlier, but that would just complicate explaining how to include a path and skipping some sub-folders, plus it would only work for the Packages that already do use the `.*ignore` file. Since it's a performance optimization though, I'd say it makes more sense to have it configurable globally.

kitsunet:
Fair enough, fair enough.

albe:
Just noticed, this got merged with the old commit message, which both stated the wrong configuration setting and that "CLDR" was excluded by default, which is not true (yet). Anything we can do about this to at least fix the Release Changelog @kitsunet @kdambekalns ?

The correct commit message is in https://github.com/neos/flow-development-collection/pull/385/commits/ad97cdacaa8e157bf80948be098ee4a92604f5b4

kdambekalns:
Just edit the PR description, that is what the changelog is based on.

albe:
Thx! Didn't know the changelog was built from GH PRs. I guessed it used the commit message, but makes sense otherwise after thinking for a second :)

PR #386, opened 2016-06-14
BUGFIX: Input field name for multiple checkbox is generated correctly

albe:
Checkboxes that were bound to collection properties or had the `multiple` attribute set,
were generating invalid input field names like this:

    <input type="checkbox" name="post[tags][__identity][]" ...>

Since the internal identity is not required for checkboxes, it is stripped from the generated name.
Also, any such checkboxes now avoid generating an empty value hidden field, as this will at most
lead to empty values being additionally submitted to the collection property.

FLOW-419 #close

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/AbstractFormFieldViewHelper.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/AbstractFormFieldViewHelper.php
@@ -81,9 +81,14 @@ abstract class AbstractFormFieldViewHelper extends AbstractFormViewHelper
         } else {
             $name = $this->arguments['name'];
         }
-        if ($this->hasArgument('value') && is_object($this->arguments['value'])) {
-            if (null !== $this->persistenceManager->getIdentifierByObject($this->arguments['value'])
-                && (!$this->persistenceManager->isNewObject($this->arguments['value']))) {
+        if ($this->hasArgument('value')) {
+            /** @var object $value */
+            $value = $this->arguments['value'];
+            $multiple = $this->hasArgument('multiple') && $this->arguments['multiple'] === true;
+            if (!$multiple
+                && is_object($value)
+                && $this->persistenceManager->getIdentifierByObject($value) !== null
+                && (!$this->persistenceManager->isNewObject($value))) {
                 $name .= '[__identity]';
             }
         }
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/CheckboxViewHelper.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/CheckboxViewHelper.php
@@ -75,7 +75,6 @@ class CheckboxViewHelper extends AbstractFormFieldViewHelper
     {
         $this->tag->addAttribute('type', 'checkbox');
 
-        $nameAttribute = $this->getName();
         $valueAttribute = $this->getValueAttribute(true);
         $propertyValue = null;
         if ($this->hasMappingErrorOccurred()) {
@@ -92,11 +91,14 @@ class CheckboxViewHelper extends AbstractFormFieldViewHelper
             if ($checked === null) {
                 $checked = in_array($valueAttribute, $propertyValue);
             }
-            $nameAttribute .= '[]';
-        } elseif ($multiple === true) {
-            $nameAttribute .= '[]';
-        } elseif ($propertyValue !== null) {
+            $this->arguments['multiple'] = true;
+        } elseif (!$multiple && $propertyValue !== null) {
             $checked = (boolean)$propertyValue === (boolean)$valueAttribute;
+        }
+
+        $nameAttribute = $this->getName();
+        if ($this->arguments['multiple']) {
+            $nameAttribute .= '[]';
         }
 
         $this->registerFieldNameForFormTokenGeneration($nameAttribute);
@@ -111,19 +113,5 @@ class CheckboxViewHelper extends AbstractFormFieldViewHelper
 
         $this->renderHiddenFieldForEmptyValue();
         return $this->tag->render();
-    }
-
-    /**
-     * Get the name of this form element, without prefix.
-     *
-     * This is done to prevent the extra __identity being added for objects
-     * since it leading to property mapping errors and it works without it.
-     *
-     * @return string name
-     */
-    protected function getNameWithoutPrefix()
-    {
-        $name = parent::getNameWithoutPrefix();
-        return str_replace('[__identity]', '', $name);
     }
 }
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/CheckboxViewHelper.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/CheckboxViewHelper.php
@@ -97,7 +97,7 @@ class CheckboxViewHelper extends AbstractFormFieldViewHelper
         }
 
         $nameAttribute = $this->getName();
-        if ($this->arguments['multiple']) {
+        if (isset($this->arguments['multiple']) && $this->arguments['multiple'] === true) {
             $nameAttribute .= '[]';
         }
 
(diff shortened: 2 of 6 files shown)


aertmann:
well hello old friend, thanks for taking care @albe 👍

kitsunet:
Seems like a crude solution, but well... I just wonder why that didn't come up before.

aertmann:
@kitsunet: it did, I did it more than two years ago but we had already released Flow stable for Neos 1.0 stable so put it in media.. see the Jira ticket

kitsunet:
Ouch, good to get it solved then.

bwaidelich:
My bad, I lost sight of this issue. Thanks for taking care.
Let me please try to find a way without `str_replace` – that seems a bit flaky to me (obviously better than a broken implementation).
I'll report back here within the next 30 mins

albe:
Looking forward to what you might come up with :)

bwaidelich:
It took me a bit longer, there are quite a few cases to consider..
Same effect but maybe a little bit more stable could be to change the last bit of `AbstractFormFieldViewHelper::getNameWithoutPrefix()` from
```php
if ($this->hasArgument('value') && is_object($this->arguments['value'])) {
    if (null !== $this->persistenceManager->getIdentifierByObject($this->arguments['value'])
        && (!$this->persistenceManager->isNewObject($this->arguments['value']))) {
        $name .= '[__identity]';
    }
}
```
to something like
```php
if ($this->hasArgument('value')) {
    $value = $this->arguments['value'];
    $multiple = $this->hasArgument('multiple') && $this->arguments['multiple'] === true;
    if (!$multiple
        && is_object($value)
        && $this->persistenceManager->getIdentifierByObject($value) !== null
        && (!$this->persistenceManager->isNewObject($value))) {
        $name .= '[__identity]';
    }
}
}
```

wdyt?

albe:
This doesn't handle the case when the bound property is a Collection though, does it?

`<input property="someCollection" value="{someItem}" ...>`

`{someItem}` will be a entity with identifier and "multiple" is not set, but checkbox VH will still append `[]` because it is part of a collection.

bwaidelich:
> This doesn't handle the case when the bound property is a Collection though

You're right.. Another case of magic getting in our way..
But that could be fixed in the Checkbox-VH by setting the `multiple` argument for array-values:

```php
if (is_array($propertyValue)) {
    if ($checked === null) {
        $checked = in_array($valueAttribute, $propertyValue);
    }
    $this->arguments['multiple'] = true;
} elseif ($propertyValue !== null) {
    $checked = (boolean)$propertyValue === (boolean)$valueAttribute;
}

$nameAttribute = $this->getName();
if ($this->arguments['multiple']) {
    $nameAttribute .= '[]';
}
```

albe:
That could work, but I'm not yet sure about the side-effects of having that handled down in the `AbstractFormFieldViewHelper`. What about other viewhelpers with a "multiple" attribute? Does this apply to select for example? Probably yes, but what about custom VH with "multiple" attribute in a different context?

bwaidelich:
> I'm not yet sure about the side-effects

Good point, but it's unlikely IMO because the `[__identity]` suffix will now only be added to the name attribute, if the ViewHelper...
* ...extends `AbstractFormFieldViewHelper``
* ...has a `value`-argument with an object known to the PersistenceManager
* ...has no `multiple`-argument with a value of `TRUE`.

I just gave it a try and the tests work still without modifications..

The patch to your PR:

```patch
diff --git a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/AbstractFormFieldViewHelper.php b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/AbstractFormFieldViewHelper.php
index 761bc4d..860e14a 100644
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/AbstractFormFieldViewHelper.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/AbstractFormFieldViewHelper.php
@@ -81,9 +81,14 @@ abstract class AbstractFormFieldViewHelper extends AbstractFormViewHelper
         } else {
             $name = $this->arguments['name'];
         }
-        if ($this->hasArgument('value') && is_object($this->arguments['value'])) {
-            if (null !== $this->persistenceManager->getIdentifierByObject($this->arguments['value'])
-                && (!$this->persistenceManager->isNewObject($this->arguments['value']))) {
+        if ($this->hasArgument('value')) {
+            /** @var object $value */
+            $value = $this->arguments['value'];
+            $multiple = $this->hasArgument('multiple') && $this->arguments['multiple'] === true;
+            if (!$multiple
+                && is_object($value)
+                && $this->persistenceManager->getIdentifierByObject($value) !== null
+                && (!$this->persistenceManager->isNewObject($value))) {
                 $name .= '[__identity]';
             }
         }
diff --git a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/CheckboxViewHelper.php b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/CheckboxViewHelper.php
index c3caa92..de3bb11 100644
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/CheckboxViewHelper.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/CheckboxViewHelper.php
@@ -75,7 +75,6 @@ class CheckboxViewHelper extends AbstractFormFieldViewHelper
     {
         $this->tag->addAttribute('type', 'checkbox');
 
-        $nameAttribute = $this->getName();
         $valueAttribute = $this->getValueAttribute(true);
         $propertyValue = null;
         if ($this->hasMappingErrorOccurred()) {
@@ -92,13 +91,16 @@ class CheckboxViewHelper extends AbstractFormFieldViewHelper
             if ($checked === null) {
                 $checked = in_array($valueAttribute, $propertyValue);
             }
-            $nameAttribute .= '[]';
-        } elseif ($multiple === true) {
-            $nameAttribute .= '[]';
-        } elseif ($propertyValue !== null) {
+            $this->arguments['multiple'] = true;
+        } elseif (!$multiple && $propertyValue !== null) {
             $checked = (boolean)$propertyValue === (boolean)$valueAttribute;
         }
 
+        $nameAttribute = $this->getName();
+        if ($this->arguments['multiple']) {
+            $nameAttribute .= '[]';
+        }
+
         $this->registerFieldNameForFormTokenGeneration($nameAttribute);
         $this->tag->addAttribute('name', $nameAttribute);
         $this->tag->addAttribute('value', $valueAttribute);
@@ -112,18 +114,4 @@ class CheckboxViewHelper extends AbstractFormFieldViewHelper
         $this->renderHiddenFieldForEmptyValue();
         return $this->tag->render();
     }
-
-    /**
-     * Get the name of this form element, without prefix.
-     *
-     * This is done to prevent the extra __identity being added for objects
-     * since it leading to property mapping errors and it works without it.
-     *
-     * @return string name
-     */
-    protected function getNameWithoutPrefix()
-    {
-        $name = parent::getNameWithoutPrefix();
-        return str_replace('[__identity]', '', $name);
-    }
 }
```

bwaidelich:
@albe if you agree to my proposal you can apply the patch using `git apply <patch>` in your fork.

albe:
Well, I guess the possibility of breakiness is close to none, so I'm fine with going that route. Will update the PR this evening if I find time, otherwise feel free to push the diff yourself :)

bwaidelich:
> Will update the PR this evening if I find time

Thanks!

> otherwise feel free to push the diff yourself

In this case you'd need to give me write access to your fork

albe:
You should have already :) Added most neos members pretty early on

bwaidelich:
ah, right, done

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/AbstractFormFieldViewHelper.php
```diff
@@ -81,9 +81,14 @@ protected function getNameWithoutPrefix()
         } else {
             $name = $this->arguments['name'];
         }
-        if ($this->hasArgument('value') && is_object($this->arguments['value'])) {
-            if (null !== $this->persistenceManager->getIdentifierByObject($this->arguments['value'])
-                && (!$this->persistenceManager->isNewObject($this->arguments['value']))) {
+        if ($this->hasArgument('value')) {
+            /** @var object $value */
```
albe: Is this type-hint docblock really needed? There is the `is_object` check anyway in the condition.

albe:
Do the others have any objections to the new fix?

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/AbstractFormFieldViewHelper.php
bwaidelich: right, not needed

PR #387, opened 2016-06-14
TASK: Make expected exception class name fully qualified

kdambekalns:
This is rather cosmetic, it fixes IDE navigation and error detection.


afoeder:
should the class name be namespace-imported when on it?

kdambekalns:
Well, I was disappointed to find that in PhpStorm you can mass-convert arrays to short syntax, but apparently not mass-import or simplify fully qualified class names. So… I didn't touch namespace imports at all in my recent changes. Keeps it simpler. :)

Also, those mass-changes cause conflicts, so… maybe we just keep doing that along the way.

dfeyer:
👍 IX (IDE Experience)

PR #390, opened 2016-06-17
FEATURE: Allow asynchronous execution of commands

bwaidelich:
Adds a new convenience method `Scripts::executeCommandAsync()` that can be
used to execute commands without waiting for their result.

This is especially useful for time-consuming tasks whose result is not
(instantly) required.

Example::

     Scripts::executeCommandAsync('my.package:registration:sendconfirmationmail', $this->flowSettings, ['email' => $emailAddress]);

FLOW-458 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
@@ -637,6 +637,27 @@ class Scripts
     }
 
     /**
+     * Executes the given command as a sub-request to the Flow CLI system without waiting for the output.
+     * 
+     * Note: As the command execution is done in a separate thread potential exceptions or failures will *not* be reported
+     *
+     * @param string $commandIdentifier E.g. typo3.flow:cache:flush
+     * @param array $settings The TYPO3.Flow settings
+     * @param array $commandArguments Command arguments
+     * @return void
+     * @api
+     */
+    public static function executeCommandAsync($commandIdentifier, array $settings, array $commandArguments = array())
+    {
+        $command = self::buildSubprocessCommand($commandIdentifier, $settings, $commandArguments);
+        if (DIRECTORY_SEPARATOR === '/') {
+            exec($command . ' > /dev/null 2>/dev/null &');
+        } else {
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -637,6 +637,27 @@ public static function executeCommand($commandIdentifier, array $settings, $outp
...
+     * @return void
+     * @api
+     */
+    public static function executeCommandAsync($commandIdentifier, array $settings, array $commandArguments = array())
+    {
+        $command = self::buildSubprocessCommand($commandIdentifier, $settings, $commandArguments);
+        if (DIRECTORY_SEPARATOR === '/') {
+            exec($command . ' > /dev/null 2>/dev/null &');
+        } else {
+            pclose(popen('START /B CMD /S /C "' . $command . '" > NUL 2 > NUL &', 'r'));
```
bwaidelich: This looks weird but it seems to be the only reliable way to trigger async commands on Windows according to http://stackoverflow.com/questions/22154506/asynchronous-popen-exec-in-php-on-windows

I tested it on a Win7-Installation on a virtual machine

dlubitz: I guess there is a space to much, shouldn't it be:
`pclose(popen('START /B CMD /S /C "' . $command . '" > NUL 2> NUL &', 'r'));`

I've tested the command itself (Win 10):
`START /B CMD /S /C "ping 127.0.0.1 -n 10 > nul & echo %PATH% > test.txt" > NUL 2> NUL &`
And this works.

bwaidelich: Good catch, does it really make a difference though?

bwaidelich: > does it really make a difference though?

Tested it again and: yes, it does in case of an error:
`START /B CMD /S /C "unknowncommand > test.txt" > NUL 2 > NUL &`
vs
`START /B CMD /S /C "unknowncommand > test.txt" > NUL 2> NUL &`

pushing a follow-up, thx!

PR #391, opened 2016-06-20
BUGFIX: FlowException is no imported

albe:
This change fixes a missing class import introduced with PR #376 into the 2.3 branch.

### DIFF
--- a/TYPO3.Flow/Tests/Unit/Core/BootstrapTest.php
+++ b/TYPO3.Flow/Tests/Unit/Core/BootstrapTest.php
@@ -56,7 +56,7 @@ class BootstrapTest extends \TYPO3\Flow\Tests\UnitTestCase
      */
     public function resolveRequestHandlerThrowsUsefulExceptionIfNoRequestHandlerFound()
     {
-        $bootstrap = $this->getAccessibleMock(Bootstrap::class, array('dummy'), array(), '', false);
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/Tests/Unit/Core/BootstrapTest.php
```diff
@@ -49,4 +49,14 @@ public function isCompileTimeCommandControllerChecksIfTheGivenCommandIdentifierR
...
+
+    /**
+     * @test
+     * @expectedException \TYPO3\Flow\Exception
+     */
+    public function resolveRequestHandlerThrowsUsefulExceptionIfNoRequestHandlerFound()
+    {
+        $bootstrap = $this->getAccessibleMock(Bootstrap::class, array('dummy'), array(), '', false);
+        $bootstrap->_call('resolveRequestHandler');
+    }
```
albe: Yes, this should be a test for this exact exception. Couldn't properly fix it up earlier as I'm on vacation here and only have short time spans before my little one wants attention.

bwaidelich: Ah right, I overlooked that this is a `Test` class. Thanks for this and enjoy your holidays, I'll push a follow up (since I have push-rights to your forks *har har*)

bwaidelich:
I just wanted to push the same change, but you already did. Awesome. Now back to holiday!! ;)

PR #394, opened 2016-06-29
BUGFIX: Support PHP 7.0 type declarations on parameters and return types

hlubek:
Adds methods to the reflection service to access declared types in
a backwards compatible way (returning null for PHP < 7.0) and extends
the proxy builder to add type declarations if they were used on the
original method. This allows to use type declarations for parameters
(scalar values) or return type declarations even for advised methods.

FLOW-460 #fixes

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php
@@ -152,11 +152,15 @@ class ProxyMethod
 
         $visibility = ($this->visibility === null ? $this->getMethodVisibilityString() : $this->visibility);
 
+        $returnType = $this->reflectionService->getMethodDeclaredReturnType($this->fullOriginalClassName, $this->methodName);
+        $returnTypeDeclaration = ($returnType !== null ? ' : ' . $returnType : '');
+
+
         $code = '';
         if ($this->addedPreParentCallCode !== '' || $this->addedPostParentCallCode !== '' || $this->methodBody !== '') {
             $code = "\n" .
                 $methodDocumentation .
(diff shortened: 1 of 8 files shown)


kitsunet:
So how is the test fixture supposed to work below PHP7 in regards to proxy building?

hlubek:
@kitsunet In my tests the scalar type declarations on parameters didn't count as a syntax errors (they were taken as classes which do not exist and thus the method rejects to be called with a simple value). What I couldn't get working was a return type declaration so I left that out. As long as we don't instantiate / call the method the tests should work fine even for lower PHP versions.

hlubek:
Fixed the ReflectionService to not override array subtype information from scalar type declarations (this caused the failed functional test) and updated a unit test to include the new method parameter key.

hlubek:
The tests are actually okay.

The error `syntax error, unexpected '\' (T_NS_SEPARATOR), expecting identifier (T_STRING)` is caused by a bug in `zend-code` (3.0.3) and unrelated to this change.

dfeyer:
Maybe we can target 2.2 ?

hlubek:
@dfeyer I'm not sure we should backport this to a lower major version. It's quite a change to the proxy class builder and reflection service.

albe:
Let's keep this in master for now and get it merged for the LTS.

kitsunet:
I think the test fails are legit, can we get that fixed this week?

hlubek:
@kitsunet It seems to be a problem of the`zend-code` package last time I digged into the test failures, I'll check the tests locally again

hlubek:
I don't get the error locally, so I retriggered the TravisCI job, let's see if it succeeds

hlubek:
@kitsunet Tests did run fine this time, could you merge this?

PR #395, opened 2016-07-02
BUGFIX: Check if array is ``null`` at the beginning of JSON conversion

gerhard-boden:
Prevent an exception to be thrown if the array passed for conversion is ``null``.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/DataTypes/JsonArrayType.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/DataTypes/JsonArrayType.php
@@ -118,13 +118,12 @@ class JsonArrayType extends DoctrineJsonArrayType
      */
     public function convertToDatabaseValue($array, AbstractPlatform $platform)
     {
-        $this->initializeDependencies();
-
-        $this->encodeObjectReferences($array);
-
         if ($array === null) {
             return null;
         }
+
+        $this->initializeDependencies();
+        $this->encodeObjectReferences($array);
 
         return json_encode($array, JSON_PRETTY_PRINT | JSON_FORCE_OBJECT | JSON_UNESCAPED_UNICODE);
     }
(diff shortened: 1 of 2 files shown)


albe:
Makes sense, thx for the PR! A test to cover this would make it perfect.

gerhard-boden:
Hi @albe!

You're right, a test would be awesome. Tried around a bit and it feels like https://github.com/neos/flow-development-collection/blob/master/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/DataTypes/JsonArrayType.php#L119 isn't that much fun to test (at least for me... not really a PHPUnit Pro)

When I want to create an instance the complaints go on and on about things like private constructors. No luck with static method calls either (since `$this` is used). Could need a hint how to write a useful test, maybe it's really easy and I'm not seeing it.

Btw. I just saw that the file permissions changed with this PR. Is this something to obey or are permissions reset while building?

albe:
Mocking the `JsonArrayType` without mocking the methods `encodeObjectReferences` and `convertToDatabaseValue` should already be enough.
Then you can just call `$mockJsonArrayType->convertToDatabaseValue(null, null);` and assert that it returns `null`. The dependencies are not accessed unless you have an array with an persisted entity inside, which does not need to be tested (for this PR).

gerhard-boden:
@albe Thanks a lot for the hint! 

I got pretty close in the afternoon (some extra calls like ``disableOriginalConstructor`` where still necessary), but after reading your comment the rest was pretty straight forward and I completed the test in the evening. I took the liberty and added another test method where a simple array with a couple of values is converted. The test case is far from complete, but having two small tests beats having none.

gerhard-boden:
One machine fails with the test, because ``setMethodsExcept()`` is not a method. Works fine on my test setup and on all the others. Is it a local issue?

albe:
Very nice! The failing test is due to PHPUnit 4.8 which is used for PHP 5.5 - that probably needs a workaround until we drop 5.5 support finally.

Seems about right regarding the pgsql/`jsonb` TODO - or is there something missing @kdambekalns?

gerhard-boden:
just took another look at the test, turns out i can do without the method that is not in PHPUnit 4.8. 
(squashed the commit, so the history stays clean)

albe:
Thanks a lot! Nothing from my side any more

gerhard-boden:
You're very welcome, thanks for the lightning fast feedback!

kdambekalns:
Late reply, but yes: the todo can be removed it seems…

PR #396, opened 2016-07-04
BUGFIX: Fix converter determination for the ObjectConverter

beheist:
When mapping a plain object with the ObjectConverter, root namespace 
properties like \DateTime do not find a converter. This is because
the leading backslash is not removed by the ObjectConverter (i.e. we
looked for a "\DateTimeConverter" instead of the DateTimeConverter).

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/ObjectConverter.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/ObjectConverter.php
@@ -16,7 +16,10 @@ use TYPO3\Flow\Property\Exception\InvalidDataTypeException;
 use TYPO3\Flow\Property\Exception\InvalidPropertyMappingConfigurationException;
 use TYPO3\Flow\Property\Exception\InvalidTargetException;
 use TYPO3\Flow\Property\PropertyMappingConfigurationInterface;
+use TYPO3\Flow\Reflection\Exception\InvalidPropertyTypeException;
 use TYPO3\Flow\Reflection\ObjectAccess;
+use TYPO3\Flow\Utility\Exception\InvalidTypeException;
+use TYPO3\Flow\Utility\TypeHandling;
 
 /**
  * This converter transforms arrays to simple objects (POPO) by setting properties.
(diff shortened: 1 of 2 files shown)


albe:
Wondering why I never noticed this, even though we definitely use `@var \DateTime` even together with a custom `DateTimeConverter`. Actually, the TypeConverter map is built from reflection data, which is always normalized to FQCN and the TypeConverters themself should also state their targetType as FQCN. So there should be no mismatch, unless either Reflection was wrong or the TypeConverter::$targetType. Maybe I'm missing something?

Still, this would then need to be adjusted accordingly in PersistentObjectConverter, no? Or maybe it would even make more sense to fix it inside PropertyMapper::findTypeConverter directly and normalize to FQCN there.

beheist:
The comparison with PersistentObjectConverter was actually how I found this out. In my entities, I regularly use @var \DateTime without issues. However the ObjectConverter doesn't use the class schema (like the PersistentObjectConverter does), but reads the annotations directly, first trying __construct @param s, then setter @param s, then the @var annotation at the property.

albe:
Ah ok, I see... thanks for the hint. ClassSchema does a `TypeHandling::parseType` on the `@var` tag value, which in turn creates the FQCN.

That should still only have effect on the last `else`, since only there the `ObjectConverter` is different from the `PersistentObjectConverter`. Also, it would be more correct to also use `TypeHandling::parseType` there too, as that also works with array/collection element types. See `ClassSchema::addProperty()`, as you need to catch the InvalidTypeException then.

beheist:
Hi Alex,
I updated the change following your hints. Also removed the superfluous tests. Thanks for the feedback!

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/ObjectConverter.php
```diff
@@ -138,9 +141,18 @@ public function getTypeOfChildProperty($targetType, $propertyName, PropertyMappi
...
+                    // This ensures that FQCNs are returned without leading backslashes. Otherwise, something like @var \DateTime
+                    // would not find a property mapper. It is needed because the ObjectConverter doesn't use class schemata,
+                    // but reads the annotations directly.
+                    $declaredType = strtok(trim(current($varTagValues), " \n\t"), " \n\t");
+                    try {
+                        $parsedType = TypeHandling::parseType($declaredType);
+                    } catch (InvalidTypeException $exception) {
+                        throw new \InvalidArgumentException(sprintf($exception->getMessage(), 'class "' . $targetType . '" for property "' . $propertyName . '"'), 1467699674);
+                    }
+                    return $parsedType['type'];
```
albe: Last nitpick: This should return the full type including `elementType` if it exists. Else collections/arrays of objects/entities would not be converted correctly.

`return $parsedType['type'] . ($parsedType['elementType'] !== null ? '<' . $parsedType['elementType'] . '>' : '');`

albe:
Thanks, that looks good apart from the last tiny issue.

Probably would make sense to even move that logic into some kind of `getPropertyType` reflectionService method in a follow-up.

beheist:
Last issue fixed as well, and you're certainly right about refactoring that, since I duplicated quite a bit of code from the reflectionService now.

albe:
Thanks! Ok from my side now :)

dfeyer:
As this is a bugfix, maybe point to an other branch ?

kitsunet:
Probably also in older branches, so needs to go to lowest please :)

albe:
2.3 is security-fixes only by now, so targeting 3.0 would be correct, right?

beheist:
Ok, can target this to 3.0.

beheist:
Can be closed, I created https://github.com/neos/flow-development-collection/pull/397

PR #397, opened 2016-07-07
BUGFIX: Fix Property Mapper determination for the ObjectConverter

beheist:
When mapping a plain object with the ``ObjectConverter``, root namespace 
properties like ``\DateTime`` do not find a converter. This is because
the leading backslash is not removed by the ``ObjectConverter`` (i.e. we
looked for a ``\DateTimeConverter`` instead of the ``DateTimeConverter``


kitsunet:
I am pretty sure we have similar code somewhere else and should extract it to a utility method, but that can be done in a separate PR.

kitsunet:
Makes sense. Just rechecked. Will follow up with a refactor to avoid duplications of this `@var` parsing.

PR #398, opened 2016-07-07
BUGFIX: Allow SliceOperation to work on \Iterator

beheist:
The Eel slice operation only works with arrays so far. Since the FlowQuery
context can be anything that implements `\Iterator`, we now convert iterators
to arrays if they are passed in as the context.


johannessteu:
👍  by reading

TYPO3.Eel/Classes/TYPO3/Eel/FlowQuery/Operations/SliceOperation.php
```diff
@@ -39,7 +39,9 @@ class SliceOperation extends AbstractOperation
     public function evaluate(FlowQuery $flowQuery, array $arguments)
     {
         $context = $flowQuery->getContext();
-
+        if (is_object($context) && $context instanceof \Iterator) {
```
albe: the `is_object($context)` is not necessary here

beheist:
@albe fixed.

PR #399, opened 2016-07-09
BUGFIX: Use Unix paths in InstallerScripts

WouterJ:
The Files utility used by InstallerScripts will use these constants
to transform an absolute path to a relative one. As the compared path
will always be a Unix path, the path to replace needs to be completely
Unix as well to make replacing working. This prevents "mkdir(): invalid
arguments" errors on Windows.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Composer/InstallerScripts.php
@@ -33,15 +33,15 @@ class InstallerScripts
     public static function postUpdateAndInstall(Event $event)
     {
         if (!defined('FLOW_PATH_ROOT')) {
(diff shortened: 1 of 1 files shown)


kitsunet:
Makes sense! Thanks for the PR :)

kdambekalns:
Looks good, thanks! But: no lower branches affected?

WouterJ:
@kdambekalns no, this code was introduced by https://github.com/neos/flow-development-collection/pull/212 in 3.2.0

WouterJ:
Discovered that a second change was in code added back in 2.2. So I've removed those changes from this PR and created https://github.com/neos/flow-development-collection/pull/400 to fix them for 3.0.

kitsunet:
Thanks @WouterJ !

albe:
will try to take a look later to test it, but looks good from reading

PR #400, opened 2016-07-10
BUGFIX: Use Unix paths in InstallerScripts

WouterJ:
The Files utility used by InstallerScripts will use these constants
to transform an absolute path to a relative one. As the compared path
will always be a Unix path, the path to replace needs to be completely
Unix as well to make replacing working. This prevents "mkdir(): invalid
arguments" errors on Windows.

Similar to neos/flow-development-collection#399 , but then for the 3.0
version.


kitsunet:
Will merge those in 24h if no one objects

PR #401, opened 2016-07-11
FEATURE: New parameter to keep output ordered by loading order instead of name

dlubitz:
Very helpful parameter to debug loading order of packages.

`package:list --loading-order`


dlubitz:
Is the correspponding documentation generated automatically? Or do I have to add the new param?
http://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartV/CommandReference.html
http://neos.readthedocs.io/en/stable/References/CommandReference.html

kdambekalns:
The references are auto-generated (at least those that say so, including those two).

But… what about the `git` in the change?

dlubitz:
@kdambekalns Sorry, was an unmeant change. I've fixed it.

PR #402, opened 2016-07-12
TASK: Fix use of deprecated getMock()

kdambekalns:


kdambekalns:
Um, why is there no Travis check?

PR #403, opened 2016-07-12
BUGFIX: Tweaked "ignoredTables" behavior

bwaidelich:
This is a follow-up to a feature merged with #366 that allows
a configurable whitelist for tables respected by the `doctrine:migrationgenerate` command.

This change turns that setting into a *blacklist* and makes the list a
dictionary so that 3rd party packages can extend the list:

```yaml
TYPO3:
  Flow:
    persistence:
      doctrine:
        migrations:
          ignoredTables:
            'autogenerated_.*': TRUE
            'wp_.*: TRUE
```

This would skip all tables starting with `autogenerated_` or `wp_` from migrations when
using the `./flow doctrine:migrationgenerate` command.

This also adjusts the Settings schema and adds a section to the documentation.

Related: #366


TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
```diff
@@ -964,6 +995,7 @@ Doctrine tries to keep existing data as far as possible, avoiding lossy actions.
 
 	Be careful, the update command might destroy data, as it could drop tables and fields
 	irreversibly.
+	It also ignores any ``ignoredTables`` settings (see previous section).
```
kdambekalns: This is not true, is it?

bwaidelich:
@kdambekalns I tweaked the docs a little bit to make this part clearer. Apart from that: Do you agree to this change in general?

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
```diff
@@ -943,6 +943,37 @@ migration file only deals with one package and then move it to the *Migrations*
 in that package. This way different packages can be mixed and still a reasonable migration
 history can be built up.
 
+Ignoring tables
+---------------
+
+For tables that are not known to the schema because they are code-generated or come from a
+different system sharing the same database, the ``flow:doctrine:migrationgenerate`` command
+will generate corresponding ``DROP TABLE`` statements.
+In this case you can use the ``--filter-expression`` flag to generate migrations only for tables
```
kdambekalns: Is it on purpose to have only "half a paragraph" here? If you want to have separate paragraphs, an empty line must be present, and I think you want one, because you added a linebreak…

kdambekalns:
Nice, looks good (didn't test)…

albe:
Good change from reading.

PR #404, opened 2016-07-14
FEATURE: Extend AuthenticationManagerInterface with getter for providers

dlubitz:
Adds a new getter method to the AuthenticationManagerInterface, that has to return all provided authentication providers.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Security/Authentication/AuthenticationManagerInterface.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Security/Authentication/AuthenticationManagerInterface.php
@@ -28,6 +28,13 @@ interface AuthenticationManagerInterface
     public function getTokens();
 
     /**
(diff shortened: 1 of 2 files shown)


dlubitz:
screen

Unattractive is at the moment the TYPO3SetupProvider. Maybe it make sence to filter providers by provider type `PersistedUsernamePasswordProvider`?

gerhard-boden:
please add a prefix to commit message "FEATURE: " before the merge.  Aside from that, :+1:

kdambekalns:
Out of curiosity: what is the use case?

dlubitz:
@kdambekalns E.g to provide the possibility to create accounts with different auth providers in Neos backend. For example the FrontendLogin:Frontend of Flowpack.
see https://github.com/neos/neos-development-collection/pull/612

kdambekalns:
Ah, ok. Thanks, that was the missing link…

PR #405, opened 2016-07-15
!!! TASK: Fluid standalone

kitsunet:
Replaces ``TYPO3.Fluid`` with ``Neos.FluidAdaptor`` integrating standalone fluid.

This change brings the following:

 * Standalone Fluid integration (see https://github.com/typo3/fluid)
 * Flow no longer depends on Fluid, the default View is configurable
 * Partials can be called with a package prefix "Vendor.Package:Path/Partial"

Standalone Fluid in general is faster and many of the Flow specific ViewHelpers were
rewritten to take advantage of compiling as well to make it even faster.

The change is breaking because:

 * Standalone Fluid is a major rewrite, it might react differently for edge cases
 * Notably escaping now also escapes single quotes.
 * The ``ViewInterface`` got a new static ``createWithOptions(array $options)`` construct
method, which needs to be implemented by custom view classes to have a defined way to
instantiate views.
 * Flow no longer depends on Fluid, which means you might need to require it yourself in your
distribution or package(s)
 * ``TYPO3\Fluid\*`` classes have moved to ``Neos\FluidAdaptor\*`` and a lot of classes are gone and instead to be used from the standalone fluid package if needed.
 * Boilerplate code to create Fluid views is slightly different and might need to be adapted in projects.


kitsunet:
@bwaidelich To make it easier to look at things. This is not polished at all yet. ;)

kitsunet:
At least style ci green now. Travis will not be due to the new dependency at least, I guess also some tests are not in order yet.

mficzel:
I ran directly into this error in a customer project with custom ViewHelpers
> Required class "TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper"
could not be loaded properly for reflection. ...

mficzel:
The error above was a composer issue ... after manually installing typo3fluid/fluid the cli worked but in the frontend i run into the following error.

> Exception: Argument 1 passed to TYPO3\Fluid\Core\Rendering\RenderingContext_Original::__construct() must implement interface TYPO3Fluid\Fluid\View\ViewInterface, null given

kitsunet:
@mficzel are you sure you also have the neos dev collection change applied? That should fix it

robertlemke:
@kitsunet from you side, would this be ready to be merged? Or do you still need some general feedback?

If you personally are fine with this state, let's remove the WIP label and ask for reviews.

kitsunet:
Removed WIP label, IMHO this change is too big to say, I am done, done and do not need feedback. No one has really used this so far, so I am sure feedback is a good idea.

NamelessCoder:
I've read through the first couple of thousand lines of this and it looks *really* good, excellent work so far! I've got a couple of generic review comments but rather than putting them in the code I'll mention them here (I bet it's easier for you to search for pointers than scroll through this diff).

* At least one ViewHelper still uses render method arguments - for the best result (and of course if signatures and inheritance allows it) you should move those to initializeArguments() like you did for many others.
* Some test fixture templates have added a `{test}` namespace import - you could put this in the test class and use `addNamespace` on ViewHelperResolver to add that namespace programmatically.
* At least one place still mentions `CMS` in a code comment, you can probably search-and-replace this to Neos.
* Some classes like the TemplateView or TemplatePaths can probably be simplified a little; but as a first step this porting of the pattern expanding is a good start.
* You might want to `class_alias` RootNode, ViewHelperNode and perhaps others. I found that these were sometimes referenced by third-party ViewHelpers and aliases for those would prevent most issues (note though that ViewHelperNode has a different set of constructor arguments; but requiring instances of the old class in method signatures would work with the alias in place).
* For PHP7 signature compatibility I recommend using `class_alias` on a few key interfaces: RenderingContextInterface and ViewHelperInterface are the most important. CompilableInterface, PostParseInterface and ChildNodeAccessInterface can be aliased to TYPO3Fluid's ViewHelperInterface. This may mean having to drop some methods from the interface while preserving them on the implementation - I see you already do a check for the instance of the rendering context in one place before you call getControllerContext on it, and this is the solution TYPO3 CMS also uses for this problem.

But overall this looks really good to me. I see pretty much all of the special implementations that I had to do for TYPO3 CMS so it looks like a good candidate. And do let me know if there is more I can do to help here. With the exception of Tuesdays and Wednesdays I'm also available for Skype calls to get things solved quicker.

Steam on!

kdambekalns:
@NamelessCoder Thanks for checking this out!

You wrote

> For PHP7 signature compatibility […]

Do you have any pointers to what the exact problem is?

NamelessCoder:
> Do you have any pointers to what the exact problem is?

@kdambekalns sure thing. Briefly descried, PHP insists on signature compatibility with parent class, interface and traits alike. In a subclass you cannot require an extension of an interface, for example, if the parent, interface or trait demands a specific class.

So when you subclass an interface that subclass definitely works when seen alone - but requiring it in methods which are declared in interfaces or just overridden from parent (which means it also affects protected methods that might require such classes) exists with a fatal PHP error that signatures are not compatible.

This even happens when a class file is not declared with strict mode, so it's "in there deep". The only thing that avoids it, is a `class_alias` which makes overridden signatures backwards compatible even if they require the old interface name - but of course this prevents the interface from adding any methods; it must be a 100% pure alias.

Hope this explains it! I think there was a report of this fatal PHP error in the second PR...

Neos.FluidAdaptor/Classes/Service/Exception.php
```diff
@@ -11,7 +11,7 @@
  * source code.
  */
 
-use TYPO3\Fluid;
+use Neos\FluidAdaptor;
```
dfeyer: The exception extend with wrong namespace Fluid\Exception on line 21

kitsunet:
:-1: Do not merge yet, @dfeyer noticed a problem with media browser. Seems I forgot to implement some view option... Will do later and remove my -1 when fixed

kitsunet:
Ok, problem solved @dfeyer Media module works now, I think that's it!

NamelessCoder:
The patch for CycleViewHelper is merged and version 2.0.4 is tagged/released - in case you want to update and remove that bug as well before you merge ;)

kitsunet:
Ok, here we go, adapted/updated to 2.0.4 of Fluid and added some better description to the change itself.

dfeyer:
@kitsunet I try a last time this evening, and I think we can merge

NamelessCoder:
This calls for a celebratory glass of wine :)

If you find any post-merge issues do ping me!

kitsunet:
YAY, YAY, YAY, definitely will drink a glass of scotch later on it! now fixing small test failures that were overseen...

PR #406, opened 2016-07-18
TASK: Add option to sort packages by extra configuration

kitsunet:
Adds a way to configure packages to be loaded before via `composer.json`
besides using require in a backwards compatible way.

The configuration option is documented and additionally package key
configuration is clarified.


kitsunet:
Note I only implemented "after" so far but constructed the config so that "before" could be added later. But before would need a lot more logic to work, so I skipped it for now. I guess after helps quite a lot already.

kitsunet:
Has not gotten any reviews, will not go into LTS.

PR #408, opened 2016-07-25
TASK: Update used Symfony components to 2.8

cedricziel:
Since symfony follows semver (http://symfony.com/doc/current/contributing/code/bc.html) it's almost always possible to upgrade components within the given major without BC breaks.

This updates the used symfony components to 2.8, which also matches the most current [LTS](http://symfony.com/doc/current/contributing/community/releases.html#long-term-support-versions) version supported.

I would also bring a `^3.0` requirement to the table, but that one would mean a little more work and may be done in a second step.


composer.json
```diff
@@ -89,7 +89,16 @@
     "extra": {
         "installer-name": "Framework",
         "neos": {
-            "warning": "AUTOGENERATED FILE, ONLY MODIFY THE .composer.json IN THIS DIRECTORY AND RUN THE MERGE SCRIPT."
+            "warning": "AUTOGENERATED FILE, ONLY MODIFY THE .composer.json IN THIS DIRECTORY AND RUN THE MERGE SCRIPT.",
+            "package-key": [
```
cedricziel: I'm not so sure about this passage. Please tell me if I am supposed to cut it out.

kitsunet: It wouldn't matter much and I guess as such I prefer to leave it in as to avoid manual edititing of this file.

kitsunet:
Tested locally, on Neos and tests. 👍

cedricziel:
Thx for merging!

PR #409, opened 2016-07-25
FEATURE: Add "Trusted Proxies" as a HTTP component

albe:
This change introduces a HTTP component for so called "Trusted Proxies" that is registered
at the start of the HTTP chain.

It is configurated via the setting `TYPO3.Flow.http.trustedProxies.proxies`, which is a list of
IP address ranges that are trusted to provide header overrides for some import request meta
informations like the host, port, protocol and client IP.

    proxies:
      - '216.246.40.0/24'
      - '216.246.100.0/24'

Also the header names that are trusted for those informations can be set via the configuration
option `TYPO3.Flow.http.trustedProxies.headers` and default to this:

    headers:
      clientIp: 'Client-Ip, X-Forwarded-For, X-Forwarded, X-Cluster-Client-Ip, Forwarded-For, Forwarded'
      host: 'X-Forwarded-Host'
      port: 'X-Forwarded-Port'
      proto: 'X-Forwarded-Proto'

The `clientIp` setting is chosen for backwards compatibility of the `getClientIpAddress` method.
This method will now resolve to the first IP address that is not a trusted proxy inside the first of the configured headers that exists in order from left to right.
This setting should be changed to only a single trusted header instead.

By default all IP addresses are trusted to stay mostly backwards compatible, but a more
safe setting would be to trust no proxies by default:

    proxies: []

FLOW-414 #close


albe:
Follow-Up for #321 implemented as HTTP component. Still WIP so checks fail.

All the logic around the trusted Proxies is encapsulated into the component and removed from the Request. This also added the need to make the Request properly clonable and opened it for starting to implement some PSR-7 methods to properly access the request information from the component.

Things I'd still like feedback on:
 - `Request::isFromTrustedProxy()` - yes or no? It could also be moved away, but I see use cases for third parties to check if the current request is considered trusted

 - Configuration inside `TYPO3.Flow.http.trustedProxies` or inside the components options? The latter would need a new schema specifically for this component, which would fix it to the "preprocess" chain and the given name, or we completely drop schema validation for the component

 - `clientIp` is currently configured for `getClientIpAddress` to work b/c - should this configuration be changed in next major or should it just be documented explicitly that user should change it?

cc @bwaidelich

mficzel:
`Request::isFromTrustedProxy()` - yes or no? :: Determining wether a request is from a trusted proxy or not could imho be by done with a specific (maybe customizable) http-header that is written by the http-component. So no additional method in the request is needed.

Configuration: Adding the configuration to `TYPO3.Flow.http.trustedProxies` seems correct to me but i suggest to set `proxies: []`by default.

albe:
> a specific (maybe customizable) http-header that is written by the http-component

Good idea! Currently have it in the PSR-7 request attributes, but headers are probably better

> suggest to set proxies: []by default.

That's the sane default ofc, but would be breaking (headers would not change request params any more by default as they do now), so this needs to wait until next major.

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Ip.php
```diff
@@ -0,0 +1,76 @@
...
+     *          127.0.0.254/31 and 127.0.0.255/31 will both match the IP addresses 127.0.0.254 and 127.0.0.255
+     *          1:2::3:4 will match the IPv6 address written as 1:2:0:0:0:0:3:4 or 1:2::3:4
+     *          ::7F00:1 will match the address written as 127.0.0.1, ::127.0.0.1 or ::7F00:1
+     *          ::1 (IPv6 loopback) will *not* match the address 127.0.0.1
+     *
+     * @param string $ip The IP to match
+     * @param string $range The CIDR range pattern to match against
+     * @return boolean TRUE if the pattern matched, FALSE otherwise
+     */
+    public static function cidrMatch($ip, $range)
```
kitsunet: I am pretty sure there is a tried and tested package for this. I want to get rid of `TYPO3.Flow/Classes/TYPO3/Flow/Utility` and IMHO this is a prime example of something that I am sure was implemented already.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -373,6 +382,156 @@ public function getContent($asResource = false)
...
+     * This method MUST be implemented in such a way as to retain the
+     * immutability of the message, and MUST return an instance that has the
+     * updated attribute.
+     *
+     * @see getAttributes()
+     * @param string $name The attribute name.
+     * @param mixed $value The value of the attribute.
+     * @return self
+     */
+    public function withAttribute($name, $value)
```
kitsunet: Problem with those `with` methods is that we currently expect to retain the request object and references to it. Not sure if it works in all places if we start using this now. But if it works I am happy to have it already.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -373,6 +382,156 @@ public function getContent($asResource = false)
...
+    {
+        return $this->withAttribute('clientIpAddress', $clientIpAddress);
+    }
+
+    /**
+     * Check if this request is coming from a trusted proxy.
+     *
+     * @return boolean
+     */
+    public function isFromTrustedProxy()
```
kitsunet: As this is not PSR-7 compatible I would be much happier to have it removed.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -373,6 +382,156 @@ public function getContent($asResource = false)
...
+    {
+        return $this->getAttribute('fromTrustedProxy', false);
+    }
+
+    /**
+     * @return Request
+     */
+    public function fromTrustedProxy()
```
kitsunet: As this is not PSR-7 compatible I would be much happier to have it removed.

kitsunet:
Left some comments.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
albe: OTOH, that's exactly the intention of those immutable methods - you get a new copy that does not influence the instance that is still used somewhere else. If you expect to use your new copy, you need to explicitly tell the other places about it (i.e. in the HTTP component replace the current Request).

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Ip.php
albe: Yeah, that's just a simple refactor of the existing code from the Ip Pattern class into an Utility class, and it's thoroughly tested ( I need to adjust the tests still).

I'm open to replacing it with an external dependency in another PR, though I'm a little cautious with adding external dependencies (that bring their own take of problems) only for singular needs (i.e. this single CIDR matcher).

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
albe: Yes, guess it makes sense to just set the attribute or an HTTP header for this as @mficzel suggested.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
kitsunet: I know and I like it. I just say that the current expectation for a Flow HttpRequest (which is not PSR-7 yet) is different.

albe: Ok, so would you rather completely avoid those `with*` methods or would it be enough to have them more clearly documented to return new instances of the request?

albe:
FYI, I kept the fromTrustedProxy and clientIpAddress in request attributes, because headers come from the outside and could hence be tempered.

Please check again, looks good from my side now.

TYPO3.Flow/Configuration/Settings.yaml
```diff
@@ -168,7 +168,10 @@ TYPO3:
 
         'preprocess':
           position: 'before process'
-          chain: []
+          chain:
+            'trustedProxies':
+              position: 'start'
+              component: 'TYPO3\Flow\Http\Component\TrustedProxiesComponent'
```
bwaidelich: Wouldn't it make sense to only activate this component if needed?

albe: It's currently needed for b/c because it provides some functionality that was previously in the Request (overriding host/port from headers, getting client IP address).

bwaidelich: Ah, sweet!

TYPO3.Flow/Classes/TYPO3/Flow/Http/Request.php
```diff
@@ -84,23 +102,15 @@ public function __construct(array $get, array $post, array $files, array $server
         }
         $this->setMethod($method);
 
-        if ($this->headers->has('X-Forwarded-Proto')) {
```
kitsunet: Mmm, is that replaced somewhere? I think it's important for vagrant for example.

kitsunet:
Looks merge ready to me now.

albe:
Ok, verified locally with tests and on our project that actually is behind a proxy (CacheFly).

kitsunet:
ok @albe if you can do a last rebase, I would push the button tomorrow on this.

albe:
Will do tomorrow.

albe:
Rebased on latest master. Only conflict was the request pattern options change in the Ip pattern.

PR #410, opened 2016-07-27
TASK: Create composer manifests with default PSR-4 autoloading

kitsunet:


robertlemke:
The classes string needs to end with \\ for psr-4

kitsunet:
Good catch, fixed that.

albe:
Wonder where the failing tests come from...

kitsunet:
I guess we do some strange things in the tests that don't adhere to PSR-4, will have a look....

kitsunet:
Ok, test failures should be fixed now, it came from some namespace voodoo we did assuming it would be PSR-0 for all times.

albe:
Ah, hail the all-mighty Voodoo-priests. Gonna take a look again, but looks reasonable

kdambekalns:
So, does the kickstarter create the files in the correct places as well?

kitsunet:
@kdambekalns yes it does :)

PR #411, opened 2016-07-27
BUGFIX: Make I18nService scan Translation folders for available locales

albe:
Before, the I18nService only scanned file names for locale identifiers, but the main source
for available locales are the translations, which reside in locale folders. Therefore it was not
possible to add an available locale through an additional translation file.

This went unnoticed, because the available locales are still filled with all locales available
in the CLDR.

This change will now also scan folder names inside `/Private/Translations/` for valid
locale identifiers.


albe:
Related to #385

kitsunet:
That somehow makes sense. 👍

albe:
btw. I targeted master because this fix still slightly changes behaviour, though it should not be "breaking". Still, this will be preferable with #385 in place, so it should maybe be merged before 3.3 LTS.

albe:
Trivial rebase.

TYPO3.Flow/Classes/TYPO3/Flow/I18n/Utility.php
```diff
@@ -111,6 +111,26 @@ public static function extractLocaleTagFromFilename($filename)
...
+    public static function extractLocaleTagFromDirectory($directory)
+    {
+        $directoryParts = explode('/', rtrim($directory, '/'));
+        $lastDirectoryPart = array_pop($directoryParts);
+
+        if ($lastDirectoryPart !== null && preg_match(Locale::PATTERN_MATCH_LOCALEIDENTIFIER, $lastDirectoryPart) === 1) {
+            return $lastDirectoryPart;
+        } else {
+            return false;
+        }
```
aertmann: maybe get rid of unnecessary else statement? *hint* https://github.com/neos/flow-development-collection/pull/443

albe:
Ah, finally a little bit of time for neos/flow stuff... feels good :)

PR #412, opened 2016-07-27
!!!BUGFIX: I18nService ignores .xml, .php and .rss files for scanning locales

albe:
The i18nService was supposed to ignore .xml, .php and .rss files when scanning for available
locales in filenames. However, the implementation suffered from a off-by-one error, which
made the check useless.

This is breaking, because it will lead to much less locales being collected for the available
locales, since the CLDR resources will no longer be considered. Also without the fix to
actually scan the Translation folders for available locales, it will even end up with a nearly
empty list of available locales.

Depends on #411

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/I18n/Utility.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/I18n/Utility.php
@@ -99,7 +99,7 @@ class Utility
 
         $filenameParts = explode('.', $filename);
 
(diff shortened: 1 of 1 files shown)


albe:
Ok, so the current behaviour is actually expected and not supposed to avoid adding all of CLDR to available locales.

See Unit tests `UtilityTest::localeIdentifiersAreCorrectlyExtractedFromFilename`.

PR #413, opened 2016-08-03
FEATURE: Allow replacement of configuration with environment-variable

mficzel:
Currently configuration can be replaced with constants via the ``%CONST_NAME%``syntax.
This change adds a special prefix ENV:: to the configuration post-processing that replaces markers of the form ``%ENV::ENV_VAR_NAME%`` with the content of the environment variable if one is defined

FLOW-212  #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
@@ -656,13 +656,25 @@ EOD;
                 preg_match_all('/(?:%)((?:\\\?[\d\w_\\\]+\:\:)?[A-Z_0-9]+)(?:%)/', $configuration, $matches);
                 if (count($matches[1]) > 0) {
                     foreach ($matches[1] as $match) {
+                        $replace = false;
+                        $replacement = null;
                         if (defined($match)) {
+                            $replace = true;
+                            $replacement = constant($match);
+                        } elseif (strpos($match, 'ENV::') === 0) {
+                            $environmentValue = getenv(substr($match, 5));
+                            if ($environmentValue !== false) {
+                                $replace = true;
+                                $replacement = $environmentValue;
+                            }
+                        }
+                        if ($replace) {
                             if ($configurations[$key] === '%' . $match . '%') {
                                 // the constant expression spans the complete directive, assign directly to keep type
-                                $configurations[$key] = constant($match);
+                                $configurations[$key] = $replacement;
                             } else {
                                 // the constant is only a substring of the directive, replace that part accordingly
-                                $configurations[$key] = str_replace('%' . $match . '%', constant($match), $configurations[$key]);
+                                $configurations[$key] = str_replace('%' . $match . '%', $replacement, $configurations[$key]);
                             }
                         }
                     }
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Configuration.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Configuration.rst
@@ -122,14 +122,14 @@ which come with the Flow distribution for getting more examples.
   Always use *two spaces* for indentation in YAML files. The parser will not
   accept indentation using tabs.
 
-Constants
----------
+Constants and Environment
+-------------------------
 
 Sometimes it is necessary to use values in your configuration files which are defined as
-PHP constants. These values can be included by special markers which are replaced by the
-actual value during parse time. The format is ``%<CONSTANT_NAME>%`` where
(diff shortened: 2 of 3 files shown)


mficzel:
This pr goes well together with pr https://github.com/neos/neos-development-collection/pull/639 to do this ``SITE_NODE_NAME=neos-demo ./flow server:run`` with such a setting ``defaultSiteNodeName: '%ENV::SITE_NODE_NAME%'``

bwaidelich:
Thanks for this, useful addition!
A couple of comments though:
1. There should be a corresponding ticket for this in our issue tracker IMO
2. I'm a little bit reluctant about the special syntax and that the `ConfigurationManager` suddenly has to interpret environment variables (IMO that shouldn't be its concern). Maybe there's a way to turn special env variables (e.g. those prefixed with `NEOS_*` or whitelisted via Settings.yaml) into PHP constants early during the bootstrap so that we can just use the existing mechanism!?

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
```diff
@@ -656,13 +656,25 @@ protected function postProcessConfiguration(array &$configurations)
...
+                        }
+                        if ($replace) {
                             if ($configurations[$key] === '%' . $match . '%') {
                                 // the constant expression spans the complete directive, assign directly to keep type
-                                $configurations[$key] = constant($match);
+                                $configurations[$key] = $replacement;
                             } else {
                                 // the constant is only a substring of the directive, replace that part accordingly
-                                $configurations[$key] = str_replace('%' . $match . '%', constant($match), $configurations[$key]);
+                                $configurations[$key] = str_replace('%' . $match . '%', $replacement, $configurations[$key]);
```
bwaidelich: I think the readability of the code could be greatly improved by using `preg_replace_callback`

mficzel:
@bwaidelich 
1. Jira ticket: I can create one if we agree that something like this makes sense. This pr was merely a byproduct of a short experiment this afternoon.
2. Regarding the envar question: I personally find it much safer/cleaner if the  configuration reads some environment than if all env-vars of a given format are mapped to constants.

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
mficzel: I just wanted to add the ENV:: reading with as little changes as possible to the existing codebase to have the diff really reflect my intention.

mficzel:
I found a jira issue https://jira.neos.io/browse/FLOW-212

bwaidelich:
Thanks for linking the jira ticket!
Just for the record: I don't want to block this by any means, I was rather just "thinking out loud".. You're right, that we probably shouldn't just map env-vars to constants.. (btw I just found out that this is exactly what [symfony does](http://symfony.com/doc/current/configuration/external_parameters.html#environment-variables)). An explicit whitelist would be another option or maybe there are others.
Having expressed my concerns this might be just as well a pragmatic and working solution as is - and definitely 👍  for the feature itself!

mficzel:
@bwaidelich thanks for your feedback. I just thought about thy symfony approach and i still do'nt like it. Adding an option to override internal constants is dangerous because not all constants are meant to be changed fro outside. On the other hand the configuration is the place wehre the flexibility is intentionally exposed. So interpreting the Environment here seems the right place to me. 

Offcourse the concrete implementation i suggested can be discussed and i will gladly adjust the pr.

radmiraal:
Thanks for this PR :-) This weekend I stumbled upon this configuration setting, I was playing around with linked docker containers. To explain why I'd vote for registering env vars as constants:

I found out that it didn't work because I checked Flow's Settings.yaml, noticed the %PHP_BINDIR% and %FLOW_PATH_DATA% and just took that for the syntax to load environment vars and never ever thought about it being constants :-) So I feel like it's the most convenient approach to just use at least the same syntax.

A con would be that I wouldn't like constants like HTTP_HOST to be registered, so would it be possible to whitelist them?

And (sidetrack!) what about loading .env files? They seem to become pretty common

Just thinking out loud btw :-)

bwaidelich:
> And (sidetrack!) what about loading .env files?

This one is more about _interpreting_ env vars, see #56 for _adding_ env vars without actually setting them

kitsunet:
Can we get at least this one merged for the LTS, which means ASAP (next two days) ?

mficzel:
I am on vacation right now and have almost no internet here. If you consider this approach valid feel free to remove the needs feedback label and merge. I added the label only because i knew that this could be controversial.

kitsunet:
sorry for hijacking.... -> see #431

PR #414, opened 2016-08-04
BUGFIX: Require latest version of neos/composer-plugin

bwaidelich:
Require version 2.x of the ``composer-plugin`` package. Previous versions
use the ``excludeClasses`` setting which has been deprecated with Flow 3.0.

### DIFF
--- a/TYPO3.Flow/composer.json
+++ b/TYPO3.Flow/composer.json
@@ -25,7 +25,7 @@
         "symfony/dom-crawler": "~2.5.0",
         "symfony/console": "~2.0",
 
-        "neos/composer-plugin": "2.0.0"
+        "neos/composer-plugin": "^2.0.0"
     },
     "suggest": {
         "ext-curl": "To use the \\TYPO3\\Flow\\Http\\Client\\CurlEngine",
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/composer.json
```diff
@@ -25,7 +25,7 @@
         "symfony/dom-crawler": "~2.5.0",
         "symfony/console": "~2.0",
 
-        "neos/composer-plugin": "^1.0.1"
+        "neos/composer-plugin": "2.0.0"
```
aertmann: soo why no caret `^` here?

kitsunet: Good point, no fixed version pls. Caret is better :)

bwaidelich:
@aertmann good catch, fixed

bwaidelich:
@aertmann could you remove your 👎  so we can merge this?

kdambekalns:
https://travis-ci.org/neos/flow-development-collection/jobs/150878082 failed with this… why?

albe:
Guess you already fixed it, BuildEssentials was still requiring neos/composer-plugin "^1.0.1"

kdambekalns:
Yep, fixed by now.

PR #415, opened 2016-08-05
!!!TASK: Implicit joins in a query makes the result set distinct by default

### LINKED ISSUE #552: TYPO3\Flow\Persistence\Doctrine\Query builds query in a way that can lead to entities being fetched multiple times
Jira issue originally created by user cerlestes:

TYPO3\Flow\Persistence\Doctrine\Query, Line 107:
`$this->queryBuilder = $entityManager->createQueryBuilder()->select('e')->from($this->entityClassName, 'e');`

Because the QueryBuilder is only told to select 'e', it can happen that a query (especially with logical-or constraints and joins) fetches the same entity multiple times. This is a major bug when also a LIMIT is set for that query.
With de-duplication applied, this leads to a result set that is smaller than the desired LIMIT, as the duplicate rows are skipped.
Without de-duplication applied, this would result in a fatal double appearance of the entity in the result set.

Luckily, somewhere in Doctrine or Flow deduplication happens on the result sets (though I gave up after 30 minutes of trying to find where exactly this happens).
This leaves us with the query result that holds less than the desired amount of entities and thus breaks stuff like pagination.

The fix is very easy though: on the mentioned line 107 turn
`select('e')`
into
`select('DISTINCT e')`
as one should do, when querying for intrinsically unique datasets.

This might also benefit the overall performance of the TYPO3 Flow framework, as I'm sure there are some queries that run into the same problem, but don't experience it due to the de-duplication efforts and a missing LIMIT statement.

 Atachements:

Jira Image 

 Jira-URL: https://jira.neos.io/browse/FLOW-21

albe:
Before it was possible that a query with implicit joins created a result set with duplicate entries.
This change sets the `distinct` flag on the query object to true when an implicit join is created.

This is breaking, as it might change the result sets of your queries, if you used implicit joins (dot notation subproperty queries).

Resolves #552

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Query.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Query.php
@@ -660,6 +660,7 @@ class Query implements QueryInterface
             }
             $previousJoinAlias = $joinAlias;
         }
+        $this->setDistinct(true);
 
         return $previousJoinAlias . '.' . $propertyPathParts[$i];
     }
--- a/Neos.Flow/Tests/Functional/Persistence/Doctrine/QueryTest.php
+++ b/Neos.Flow/Tests/Functional/Persistence/Doctrine/QueryTest.php
@@ -241,22 +241,55 @@ class QueryTest extends FunctionalTestCase
     /**
      * @test
      */
-    public function subpropertyQueriesReuseJoinAlias()
+    public function implicitJoinQueryOnlyReturnsDistinctEntitiesByDefault()
     {
-        $testEntityRepository = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\TestEntityRepository();
+        $testEntityRepository = new Fixtures\TestEntityRepository();
         $testEntityRepository->removeAll();
 
-        $testEntity = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\TestEntity;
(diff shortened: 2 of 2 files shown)


kdambekalns:
Hm, maybe there is no better way, but `DISTINCT` has it's issues… Some writeups about what I mean:
- http://www.bimonkey.com/2011/01/why-to-avoid-distinct-and-group-by-to-get-unique-records/
- http://weblogs.sqlteam.com/markc/archive/2008/11/11/60752.aspx

albe:
Yeah, I'm aware. Guess we need to decide if we value ease of use over worst case performance or not. I could live with dropping this and just making clear that implicit joins might return unexpected results.

kitsunet:
I really have no opinion at this point \o/

I guess the easy use case should just be predictable, speed is something you can do by optimising to your use case. So that speaks for distinct I guess 👍

albe:
K, I'll rebase this thing then and adjust where necessary

kitsunet:
Implicit joins in sort expressions lead to nasty SQL errors because then the orderBy field does not appear in the distinct select but (my)sql requires it to be. Therefore considering:
https://github.com/neos/flow-development-collection/pull/1293

albe:
I'll investigate how a good solution could look. Worst case we just keep things as is and people need to take care of distinct result sets themself when creating implicit joins.

PR #416, opened 2016-08-05
!!! BUGFIX: Reuse join aliases on same property paths

albe:
Previously a query object like

`$query->logicalAnd($query->equals('related.property', 'foo'), $query->equals('related.otherProperty', 'bar'));`

would generate an SQL statement with two joins to the `related` entity, one for each condition, translating to

"get all objects that have any related entity with a `property` 'foo' and any related entity with `otherProperty` 'bar'".

With this change, it will only generate a single join and reuse the join for multiple conditionals, therefore translating the above query object to the more common

"get all objects that have a related entity with both a `property` 'foo' and `otherProperty` 'bar'"

This also improves performance of such queries by avoiding unnecessary joins.

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Query.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Query.php
@@ -651,9 +651,13 @@ class Query implements QueryInterface
         $propertyPathParts = explode('.', $propertyPath);
         $conditionPartsCount = count($propertyPathParts);
         for ($i = 0; $i < $conditionPartsCount - 1; $i++) {
-            $joinAlias = $propertyPathParts[$i] . $this->joinAliasCounter++;
-            $this->queryBuilder->leftJoin($previousJoinAlias . '.' . $propertyPathParts[$i], $joinAlias);
-            $this->joins[$joinAlias] = $previousJoinAlias . '.' . $propertyPathParts[$i];
+            $joinProperty = $previousJoinAlias . '.' . $propertyPathParts[$i];
+            $joinAlias = array_search($joinProperty, (array)$this->joins);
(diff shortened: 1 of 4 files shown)


albe:
Needs tests, documentation and also some API method to switch back to the other behaviour, if it is actually intended.

hlubek:
@albe I remember that issue from long ago. Is this a regression? And: let's introduce a small functional test that specifies that behavior. Shouldn't be difficult to formulate one with the existing test entities in Flow.

albe:
I don't think it's a regression, but can check again when the behaviour was introduced.

albe:
The current implementation of the join alias is around at least since 2011, so definitely no regression ;)

albe:
So, regarding making the old behaviour still available: That's probably only possible with introducing some fancy property selector markup - i.e. "$someSubentities.foo" - which I'm not so fond of.

bwaidelich:
Is this really a bugfix as in “fixes wrong behavior” or merely a performance optimization?  In any case there should be an issue for it IMO

albe:
Sorry I can't help out more to get this finished up, so if this doesn't make it, I have no bad feelings at all.
It's more a "fixes unexpected default behaviour" thing - the (very slim) performance thing comes as a side-effect.

bwaidelich:
@albe thanks for this and sorry for the late reviews..

albe:
Trivial rebase. I'm still not sure how a solution to support the old semantic could look - maybe it's good enough to just document this change and stay with hinting at DQL for the uncommon usage?

kdambekalns:
The test(s) need to be adjusted. I'd be fine with adjusting documentation and noting this change in the release notes / changelog.

PR #417, opened 2016-08-11
FEATURE: Make Cache FileBackends independent of external locks

albe:
This avoids using external locks, which are prone to platform issues
(race conditions and tombstones for lock files or missing semaphore extension)
and instead directly uses the file locking mechanism of PHP to lock the cache files.

This should noticeably improve performance for the FileBackend caches and avoid
having thousands of Lock files which clobber the file system.


albe:
See #261 #10 #138 #183

kitsunet:
👎 please not before the cache split.

albe:
Sure, I'll take care of rebasing on the split once that's merged.

albe:
I just benchmarked this to see the effects. Not that big, but surely measurable.
Biggest plus is "no more Lock files laying around" though.

As a bonus, the lock will now work if the cache files are stored on a shared network folder while the lock files aren't (very unlikely setup).

<pre>
Benchmark:
time "./flow flow:cache:flush --force && ./flow flow:core:compile && ./flow doctrine:compileproxies && ./flow flow:cache:warmup"


master:

TESTING:
real    0m29.191s
user    0m27.930s
sys     0m1.111s

real    0m29.052s
user    0m27.797s
sys     0m1.146s

real    0m29.863s
user    0m28.063s
sys     0m1.173s


PRODUCTION:
real    0m5.388s
user    0m4.734s
sys     0m0.609s

real    0m5.461s
user    0m4.831s
sys     0m0.591s

real    0m5.443s
user    0m4.823s
sys     0m0.569s


DEVELOPMENT:
real    0m11.141s
user    0m10.343s
sys     0m0.736s

real    0m11.143s
user    0m10.349s
sys     0m0.722s

real    0m11.582s
user    0m10.740s
sys     0m0.784s



flock_filecaches:

TESTING:
real    0m28.921s
user    0m27.826s
sys     0m0.990s

real    0m28.815s
user    0m27.647s
sys     0m1.048s

real    0m28.968s
user    0m27.829s
sys     0m1.018s


PRODUCTION:
real    0m5.138s
user    0m4.573s
sys     0m0.540s

real    0m5.220s
user    0m4.705s
sys     0m0.461s

real    0m5.062s
user    0m4.539s
sys     0m0.495s


DEVELOPMENT:
real    0m11.195s
user    0m10.600s
sys     0m0.534s

real    0m10.838s
user    0m10.171s
sys     0m0.612s

real    0m10.940s
user    0m10.298s
sys     0m0.583s
</pre>

kitsunet:
Actually I am not sure if that doesn't introduce race conditions that we fixed before by using locks for reading as well. Did you try two concurrent development context calls on empty cache for example?

albe:
Ah yes, that should probably be changed again to a single lock. Yeah, the Lock utility is currently useless that way, but I think it has its places elsewhere, especially for compile time locks

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
```diff
@@ -257,14 +252,7 @@ public function remove($entryIdentifier)
...
 
-        try {
-            $lock = new Lock($pathAndFilename);
-            unlink($pathAndFilename);
-            $lock->release();
-        } catch (\Exception $exception) {
-            return false;
-        }
-        return true;
+        return @unlink($pathAndFilename);
```
albe: This here is the only change of locking behaviour.

The reasoning to not lock the file for unlink is, that on Windows the unlink would then fail (can't unlink an opened file). In the worst-case, the file will be unlinked on *NIX while it is being locked in a different process, in which case the outcome is a non-existing cache entry/failed write. That should not be a problem as it will resolve itself with the next request that builds the cache.

The other way around would be worse on Win, as it would lead to the cache file never being removed.

albe:
Rebased onto 3.3, so we can deliver this into the LTS. If we decide against that, it should be trivial to retarget onto master again.

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
kitsunet: The reason for locking was to prevent concurrent reading on this file, which has happened in the past.

kitsunet:
So, I added one last question. TBH cache split doesn't happen, I give up, so if you rebase, I'll merge this.

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
albe: Hm, yeah I see... basically means that on Windows the delete will just fail if it is concurrently read somewhere else and on *NIX I'm not sure if the read would just create currupt data, fail or if the delete would fail. Do you have any idea what the behaviour is?

What if we keep the lock on non-windows only? That way we retain the concurrent-read safety on like 99% of used production environments and at least have an as-good-as-possible behaviour no windows.

albe:
Rebased once more and re-added locking on unlink for non-windows. This reatins concurrent-access safety for the widely used production environments.

On windows however, there is a slim chance that the remove continually fails if the cache entry is under high pressure. There is not much we can do about that other than adding external locking again.

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/SimpleFileBackend.php
```diff
@@ -260,14 +255,23 @@ public function remove($entryIdentifier)
...
-            $lock->release();
-        } catch (\Exception $exception) {
+        if (strncasecmp(PHP_OS, 'WIN', 3) == 0) {
+            // On Windows, unlinking a locked/opened file will not work, so we just attempt the delete straight away.
+            // In the worst case, the unlink will just fail due to concurrent access and the caller needs to deal with that.
+            return @unlink($pathAndFilename);
+        }
+
+        $file = @fopen($pathAndFilename, 'r');
+        if ($file === false) {
```
robertlemke: I have concerns about the shut-up-operator – wouldn't `try`/`catch` work here?

It's a bit of history repeating – we had this several times in and out in similar places, see also https://karsten.dambekalns.de/blog/a-story-about-include-once.html

So, don't rush this until we're really sure.

albe: Good call.

So what is the intended/expected API here on file opening failure? I'd say for the remove operation just returning false and ignoring anything else is fine (as before). But for the reading/writing part it would make sense to somehow bubble exceptions up... actually there was no `try / catch` before, so it would maybe even make sense to throw exceptions on lock failure, too.

robertlemke: In an earlier version (without `flock`) I implemented it so that errors where caught and the operation was retried 3 times or so, with a random sleep in between. And when the lock didn't happen then, an error / exception was thrown.

kitsunet:
Ok, rebased to master, adapted to Neos.Cache, replaced shutup with try cache, added retry for failsafety.

Also benchmarked with AB, speed is slightly better with the change and both show same behavior when deleting the cache in the middle of the AB run. Both recover and continue after rebuilding.

kitsunet:
@kdambekalns maybe you can have a check here?

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -454,19 +471,28 @@ protected function configureCacheDirectory()
...
+                }
+                if (flock($file, LOCK_SH) !== false) {
+                    if ($offset !== null) {
+                        fseek($file, $offset);
+                    }
+                    $data = fread($file, $maxlen !== null ? $maxlen : filesize($cacheEntryPathAndFilename) - (int)$offset);
+                    flock($file, LOCK_UN);
+                }
+                fclose($file);
+            } catch (\Exception $e) {}
```
kitsunet: In favor of speed I wouldn't even, but we can ofc.

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -467,4 +460,74 @@ protected function configureCacheDirectory()
...
+    protected function writeCacheFile($cacheEntryPathAndFilename, $data)
+    {
+        for ($i = 0; $i < 3; $i++) {
+            // This can be replaced by a simple file_put_contents($cacheEntryPathAndFilename, $data, LOCK_EX) once vfs
+            // is fixed for file_put_contents with LOCK_EX, see https://github.com/mikey179/vfsStream/wiki/Known-Issues
+            $result = false;
+            try {
+                $file = fopen($cacheEntryPathAndFilename, 'wb');
+                if ($file === false) {
+                    continue;
```
skurfuerst: in this case, don't we need to wait a bit (usleep) before continuing? Otherwise the three writes will be really quick after each other.

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -467,4 +460,74 @@ protected function configureCacheDirectory()
...
+     */
+    protected function writeCacheFile($cacheEntryPathAndFilename, $data)
+    {
+        for ($i = 0; $i < 3; $i++) {
+            // This can be replaced by a simple file_put_contents($cacheEntryPathAndFilename, $data, LOCK_EX) once vfs
+            // is fixed for file_put_contents with LOCK_EX, see https://github.com/mikey179/vfsStream/wiki/Known-Issues
+            $result = false;
+            try {
+                $file = fopen($cacheEntryPathAndFilename, 'wb');
+                if ($file === false) {
```
skurfuerst: can we add a comment which case this is? I am currently not sure in which (race-condition) case this might happen.

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -467,4 +460,74 @@ protected function configureCacheDirectory()
...
+     * @return boolean|string The contents of the cache file or FALSE on error
+     */
+    protected function readCacheFile($cacheEntryPathAndFilename, $offset = null, $maxlen = null)
+    {
+        for ($i = 0; $i < 3; $i++) {
+            $data = false;
+            try {
+                $file = fopen($cacheEntryPathAndFilename, 'rb');
+                if ($file === false) {
+                    continue;
```
skurfuerst: see below. Can we add a comment when this case can happen? IMHO it can happen when the file was removed in the meantime

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -19,7 +19,6 @@
 use Neos\Cache\Frontend\PhpFrontend;
 use Neos\Cache\Frontend\FrontendInterface;
 use Neos\Utility\Exception\FilesException;
-use Neos\Utility\Lock\Lock;
```
skurfuerst: Do we still use Neos\Utility\Lock\Lock somewhere or can we remove the thing then?

Neos.Cache/Classes/Backend/FileBackend.php
```diff
@@ -268,16 +248,14 @@ public function findIdentifiersByTag($searchedTag)
...
-            $expiryTime = (integer)substr($metaData, 0, self::EXPIRYTIME_LENGTH);
+            $fileSize = filesize($cacheEntryPathAndFilename);
+            $index = (integer)$this->readCacheFile($cacheEntryPathAndFilename, $fileSize - self::DATASIZE_DIGITS, self::DATASIZE_DIGITS);
+            $metaData = $this->readCacheFile($cacheEntryPathAndFilename, $index, $fileSize - $index - self::DATASIZE_DIGITS);
+            $expiryTime = (integer)substr($metaData, -self::EXPIRYTIME_LENGTH, self::EXPIRYTIME_LENGTH);
             if ($expiryTime !== 0 && $expiryTime < $now) {
                 continue;
             }
-            if (in_array($searchedTag, explode(' ', substr($metaData, self::EXPIRYTIME_LENGTH, -self::DATASIZE_DIGITS)))) {
+            if (in_array($searchedTag, explode(' ', substr($metaData, 0, -self::EXPIRYTIME_LENGTH)))) {
```
skurfuerst: I don't understand what is actually changing here :-)

albe: I basically just moved the expiry time to the end of the metadata block. That allows for quicker expiry checks, because it's no longer necessary to read the data size first to find the start of the arbitrary sized metadata. You can just read from the end of the file -EXPIRYTIME_LENGTH bytes and check.

albe:
@skurfuerst 

> inlined into here; and instead of separate lock files we directly use the cache-files themselves now.

Yes, exactly. The only functional difference is, that on windows we can't unlink the file while it is locked (see https://github.com/neos/flow-development-collection/pull/417#r75280544 and following) - so we unlink without locking. It's a small drawback, but IMO worth it - especially considering how windows is not necessarily the OS of choice for production.

robertlemke:
As far as I can see, there are only some minor nitpicks left (@skurfuerst asking for additional comments).

I reviewed the code again and tested the patch as thorough as I could, mostly in Production context. this definitely improves the stability a lot. Even though caches might become corrupt when you destroy them in midst of 100 requests hammering on the site at the same time, they will eventually repair themselves and after a few seconds you have a stable system again. That's great and didn't work with the old locking solution.

Since this improves stability a lot already, I'll merge this now. @skurfuerst if you still would like to have the mentioned comments, please talk that through with @albe and put them into a new PR.

albe:
:tada: Thx! I'll gladly answer on any left-over questions and improve in a follow-up if necessary

peterbucher:
If you run into the error "could not lock file" when executing a flow-Command within a vagrant box which runs on top of an Ubuntu Linux system, this can maybe help you.

The error went away after installing the vagrant plugin "vagrant-bindfs" and also installing nfs-kernelserver, as well as nfs-common.

Like:
```
vagrant plugin install vagrant-bindfs

sudo apt-get install nfs-kernel-server
sudo apt-get install nfs-common
```

PR #418, opened 2016-08-12
BUGFIX: Sub classes of adviced classes should initialize advices

kitsunet:
Any sub class of a class that has an advice also needs to initialize
the advices array in the constructor or otherwise that will lead to
exceptions later on. The only reason it works currently is the
``\TYPO3\Flow\Object\Proxy\DoctrineProxyFixingTrait``
which (as a side effect) masks this bug. If we want to get rid of this
trait with newer Doctrine versions we need to fix the bug on our side.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
@@ -222,11 +222,14 @@ class ProxyClassBuilder
         }
         $targetClassNameCandidates->sort();
 
+        $treatedSubClasses = new ClassNameIndex();
+
         foreach ($targetClassNameCandidates->getClassNames() as $targetClassName) {
             $isUnproxied = $this->objectConfigurationCache->has('unproxiedClass-' . str_replace('\\', '_', $targetClassName));
             $hasCacheEntry = $this->compiler->hasCacheEntryForClass($targetClassName) || $isUnproxied;
             if ($rebuildEverything === true || $hasCacheEntry === false) {
                 $proxyBuildResult = $this->buildProxyClass($targetClassName, $this->aspectContainers);
+                $treatedSubClasses = $this->proxySubClassesOfClassToEnsureAdvices($targetClassName, $targetClassNameCandidates, $treatedSubClasses);
                 if ($proxyBuildResult !== false) {
                     if ($isUnproxied) {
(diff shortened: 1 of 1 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -474,6 +477,45 @@ public function buildProxyClass($targetClassName, array &$aspectContainers)
...
+            }
+            if ($this->reflectionService->isClassReflected($subClassName) === false || $targetClassNameCandidates->hasClassName($subClassName)) {
+                continue;
+            }
+            $proxyClass = $this->compiler->getProxyClass($subClassName);
+            if ($proxyClass === false) {
+                continue;
+            }
+            $callBuildMethodsAndAdvicesArrayCode = "        if (method_exists(get_parent_class(), 'Flow_Aop_Proxy_buildMethodsAndAdvicesArray') && is_callable('parent::Flow_Aop_Proxy_buildMethodsAndAdvicesArray')) parent::Flow_Aop_Proxy_buildMethodsAndAdvicesArray();\n";
+            $proxyClass->getConstructor()->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
```
albe: Don't we need the same PreParentCallCode for __clone and __wakeup, too?

albe: Also, isn't that fixed already by just making sure `parent::__construct()` is called?

kitsunet: Not necessarily, no. As if the class that is proxied has a constructor it might or might not call the parent constructor, but we cannot rely on that.

kitsunet: Ok, added wakeup as well, but looking at the current code we don't do it in clone so far, so I wouldn't add it here. Might be something to have a look at though.

albe: Well, that's what was missing for the ORM 2.5 change, so that probably only needs change together with the ORM upgrade

kitsunet:
So, can we get this merged to finally also merge the ORM change?

albe:
No objections here

kitsunet:
Ok, unfortunately I found another case that needs attention when the doctrine fix is removed... Will push another change ASAP...

PR #419, opened 2016-08-12
BUGFIX: DateTimeConverter accepts timestamps when configuration without format is provided

albe:
Previously, the DateTimeConverter would fail converting timestamps when a configuration for the
property existed, but did not configure the DateTimeConverter format.
This is commonly the case in REST Apis, where type conversion needs to be configured for allowing
creation and modification, but does not necessarily explicitly configure the DateTime format.


albe:
Hm, should probably be targeted at 3.0 instead...

albe:
Closing in favor of #420 which targets 3.0

PR #420, opened 2016-08-12
BUGFIX: DateTimeConverter accepts timestamps when configuration without format is provided

albe:
Previously, the DateTimeConverter would fail converting timestamps when a configuration for the
property existed, but did not configure the DateTimeConverter format.
This is commonly the case in REST Apis, where type conversion needs to be configured for allowing
creation and modification, but does not necessarily explicitly configure the DateTime format.


TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/DateTimeConverter.php
```diff
@@ -143,7 +143,8 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
         if ($dateAsString === '') {
             return null;
         }
-        if (ctype_digit($dateAsString) && $configuration === null && (!is_array($source) || !isset($source['dateFormat']))) {
+        $configurationForm = $configuration === null ? null : $configuration->getConfigurationValue('TYPO3\Flow\Property\TypeConverter\DateTimeConverter', self::CONFIGURATION_DATE_FORMAT);
```
bwaidelich: What does the name `configurationForm` stand for?
And a nitpick: although we didn't replace all occurrences yet we should use the `class` keyword in new code:
```
$configuration->getConfigurationValue(DateTimeConverter::class);
```

bwaidelich: OK, now I understood the code I guess (it was already hard to read before). Can we maybe make it a bit more explicit, e.g. sth like:
```php
$isDateFormatSpecified = $configuration->getConfigurationValue(DateTimeConverter::class, self::CONFIGURATION_DATE_FORMAT) !== null || isset($source['dateFormat']);
if (!$isDateFormatSpecified && ctype_digit($dateAsString)) {
 // ...
```
?

albe: Thx for the quick feedback, that was a typo actually and should have read `configurationForm*at*`.
I was under the (false?) impression that 3.0 should still work with PHP < 5.5, so ::class was avoided, but I'm probably mixing things up based on the fact that 3.0 still contains lots of FQCN strings instead of ::class.

Otherwise I'm with you on making that part more explicit, I'll push a follow-up.

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/DateTimeConverter.php
```diff
@@ -138,13 +139,14 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
...
                 $dateFormat = $source['dateFormat'];
+                $isFormatSpecified = true;
             }
         }
         if ($dateAsString === '') {
             return null;
         }
-        $configurationForm = $configuration === null ? null : $configuration->getConfigurationValue('TYPO3\Flow\Property\TypeConverter\DateTimeConverter', self::CONFIGURATION_DATE_FORMAT);
-        if (ctype_digit($dateAsString) && $configurationForm === null && (!is_array($source) || !isset($source['dateFormat']))) {
+        $isFormatConfigured = ($configuration !== null && $configuration->getConfigurationValue(self::class, self::CONFIGURATION_DATE_FORMAT) !== null);
```
bwaidelich: Careful: `self:class` won't work if the class is proxied!. `DateTimeConverter::class` should do the trick

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/DateTimeConverter.php
```diff
@@ -191,7 +193,7 @@ protected function getDefaultDateFormat(\TYPO3\Flow\Property\PropertyMappingConf
         if ($configuration === null) {
             return self::DEFAULT_DATE_FORMAT;
         }
-        $dateFormat = $configuration->getConfigurationValue('TYPO3\Flow\Property\TypeConverter\DateTimeConverter', self::CONFIGURATION_DATE_FORMAT);
+        $dateFormat = $configuration->getConfigurationValue(self::class, self::CONFIGURATION_DATE_FORMAT);
```
bwaidelich: same here

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/DateTimeConverter.php
albe: Ah, damn... of course, the Proxy class... good catch, thx!

albe:
Regarding the failing test, I actually think the test is wrong:

If input is `array('date' => '1308174051', 'dateFormat' => '')` then the test expects it to not convert. However in the converter, there is a check `if (isset($source['dateFormat']) && strlen($source['dateFormat']) > 0) {`, i.e. only if the dateFormat is set and not empty, it is considered a specified format. That also makes sense from a web form POV, where fields that are unselected might still get submitted with an empty value.

So I'd suggest changing that test actually, but the question then is if that is to be considered breaking or fixing wrong behaviour.

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/DateTimeConverter.php
```diff
@@ -138,12 +139,14 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
...
                 $dateFormat = $source['dateFormat'];
+                $isFormatSpecified = true;
             }
         }
         if ($dateAsString === '') {
             return null;
         }
-        if (ctype_digit($dateAsString) && $configuration === null && (!is_array($source) || !isset($source['dateFormat']))) {
+        $isFormatConfigured = ($configuration !== null && $configuration->getConfigurationValue(DateTimeConverter::class, self::CONFIGURATION_DATE_FORMAT) !== null);
+        if (!$isFormatConfigured && !$isFormatSpecified && ctype_digit($dateAsString)) {
```
kitsunet: Is that enough to decide it's format 'U'?  I am not sure...
It could as well be YYMMDD, which is an allowed ISO notation (19800901 -> 1st September 1980), so IMHO you NEED to define the format.

kitsunet:
See inline comment... I am unsure about this one. No vote from me so far.

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/DateTimeConverter.php
aertmann: well that's similar to how it is already, it's just that if there is configuration, but it hasn't defined dateFormat then it works similar to when there is no configuration.. which makes more sense than how it currently is IMO

kitsunet: LOL, yep, I thought this was totally new, but in fact it isn't. Then I think it's a good thing to do!

kitsunet:
Updated my vote with <3

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/DateTimeConverter.php
albe: Well, there is one slim change on top of the "configuration but no format": Previously it also wouldn't trigger if `$source['dateFormat']` was set but empty, which it now will - hence the test failing. But as stated I think this is rather odd behaviour.

I could be convinced to move that additional (breaking?) change to next major only.

albe:
Rebased and changed the test in question.

albe:
Rebased and retargeted onto 3.1

albe:
Meh, rebase + force push from the wrong laptop without prior pull is a bad bad thing.

albe:
Would be cool to get this merged, so if anyone could do the required verify and push the button I'd be glad :) And thanks for the down-merge Christopher!

PR #421, opened 2016-08-12
FEATURE: Allow automatic injection of singleton constructor arguments

bwaidelich:
This allows to omit singleton constructor arguments when using the
``PropertyMapper`` to convert objects.

Background:
Constructor arguments can already be omitted if there is a default
value specified in the class to convert to:

```
class SomeClass {
  public function __construct($optionaArgument = null) {
    // ..
  }
}
```

With this change this will also work for (singleton) classes:

```
class SomeClass {
  public function __construct(SomeService $someService) {
    // ..
  }
}
$someInstance = $this->propertyMapper->convert([], SomeClass::class);
```

FLOW-463 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/ObjectConverter.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/ObjectConverter.php
@@ -12,6 +12,7 @@ namespace TYPO3\Flow\Property\TypeConverter;
  */
 
 use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Object\Configuration\Configuration;
 use TYPO3\Flow\Property\Exception\InvalidDataTypeException;
 use TYPO3\Flow\Property\Exception\InvalidPropertyMappingConfigurationException;
 use TYPO3\Flow\Property\Exception\InvalidTargetException;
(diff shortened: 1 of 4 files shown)


albe:
Makes sense and looks good from reading. The small CGL issue needs a fix and the `convertFromShouldThrowExceptionIfRequiredConstructorParameterWasNotFound` test needs to be adjusted it seems. Otherwise :heart:

TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/ObjectConverter.php
```diff
@@ -237,6 +238,8 @@ protected function buildObject(array &$possibleConstructorArgumentValues, $objec
                     unset($possibleConstructorArgumentValues[$constructorArgumentName]);
                 } elseif ($constructorArgumentReflection['optional'] === true) {
                     $constructorArguments[] = $constructorArgumentReflection['defaultValue'];
+                } elseif ($this->objectManager->getScope($constructorArgumentReflection['type']) === Configuration::SCOPE_SINGLETON) {
```
kitsunet: Note: This is breaking in case the type is not a registered object in the object manager (So any third party package object basically). I would suggest to revert and try/catch the exception....

kitsunet:
@bwaidelich see inline comment....

bwaidelich:
@kitsunet ah, good catch.. I'll try to cover this case with a test and push a follow-up

bwaidelich:
@kitsunet ok, #422 should cater for this case, although it's probably more of a theoretical issue because that code is only hit when building objects with (non-optional) constructor arguments - so this fails anyways when referring to a class unknown to the OM. But the safeguard will lead to a more meaningful exception, so good catch!

PR #422, opened 2016-08-15
BUGFIX: Safeguard for unmananged constructor injections

bwaidelich:
This adds an additional check to ``ObjectConverter::buildObject()``
in order to prevent misleading errors when trying to convert objects
with constructor injections that are unknown to the ``ObjectManager``.

FLOW-463 #related


kitsunet:
YAY :)

albe:
Looks reasonable.

PR #424, opened 2016-08-16
TASK: Remove doctrine/orm dependency

kitsunet:
As the doctrine classes are only used for type checks in
``instanceof`` and ``is_subclass_of`` and both work with
non existing classes we can drop the dependency on
``doctrine/orm`` in this package.

### DIFF
--- a/Neos.Utility.ObjectHandling/composer.json
+++ b/Neos.Utility.ObjectHandling/composer.json
@@ -8,7 +8,6 @@
     "php": ">=5.5.0"
   },
   "require-dev": {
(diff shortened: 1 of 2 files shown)


kdambekalns:
What about updating the collection `composer.json`?

kitsunet:
Shouldn't change, as the dependency still exists in Flow

kdambekalns:
Of course it does, one is `require`, one is `require-dev`…

kitsunet:
Right! And fixed :)

PR #425, opened 2016-08-16
BUGFIX: MethodsAndAdvices fix for non proxied target classes

kitsunet:
The fix done in commit 8ca79cda53f90415458eb825c5bac52105d08d40
did not catch all cases as target classes that do not have an
advice applied (usually the parent class then has the advice),
was not treated with the fix for methods and advices.

This change adds treatment for those classes as well.


kitsunet:
@albe @kdambekalns 

Part two, with this the functional tests of the ORM 2.5 change are finally green 👍

albe:
Awesomeness! Thx Christian!

PR #427, opened 2016-08-17
TASK: Fix CGL issue

albe:


albe:
not sure why tests fail... only syntax change

albe:
Closing in favor of #429

PR #428, opened 2016-08-17
TASK: Avoid PHPUnit warnings on deprecated getMock method

albe:


kdambekalns:
Only needed for master?

albe:
Yep, think they got introduced with #385

PR #429, opened 2016-08-17
TASK: Adjust unit tests to use "real mocks"

kdambekalns:
Mocking a "virtual" class is no longer allowed, so we use fixture classes.


kdambekalns:
@albe

kdambekalns:
@kitsunet

PR #430, opened 2016-08-17
TASK: Adjust some unit tests to avoid warnings

kdambekalns:


kdambekalns:
Fun fact: everything passes on Travis, but I had warnings locally. So this fixes those ;)

kdambekalns:
@kitsunet

PR #431, opened 2016-08-19
FEATURE: Allow replacement of configuration with environment-variable

kitsunet:
Currently configuration can be replaced with constants via the ``%CONST_NAME%``syntax.
This change adds a special prefix env: to the configuration post-processing that replaces markers of the form ``%env:ENV_VAR_NAME%`` with the content of the environment variable if one is defined

FLOW-212  #close


kitsunet:
Hijacking of #431 as @mficzel is in holidays and I don't have push rights.

preg_replace_callback doesn't work because of the special handling of full matches and retaining the original constant variable type (eg. an integer constant) which breaks with preg_replace_callback as that always sets a string.
Still tried to clean that a bit and make it nicer to read.

Also rebased to 3.3

@bwaidelich @radmiraal @kdambekalns @albe maybe someone could have a look?

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Configuration.rst
```diff
@@ -122,14 +122,14 @@ which come with the Flow distribution for getting more examples.
...
+Constants and Environment
+-------------------------
 
 Sometimes it is necessary to use values in your configuration files which are defined as
-PHP constants. These values can be included by special markers which are replaced by the
-actual value during parse time. The format is ``%<CONSTANT_NAME>%`` where
-``<CONSTANT_NAME>`` is the name of a constant. Note that the constant name must be all
-uppercase.
+PHP constants or are environment variables. These values can be included by special markers
+which are replaced by the actual value during parse time. The format is ``%<CONSTANT_NAME>%``
```
albe: maybe add " or ``%ENV:<VARIABLE_NAME>%`` here, otherwise the "%ENV:" prefix is only mentioned in the "%ENV:HOME%" example

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
```diff
@@ -651,24 +651,51 @@ protected function postProcessConfiguration(array &$configurations)
...
+     * @param string $value
+     * @return mixed
+     */
+    protected function replaceVariablesInValue($value)
+    {
+        $matches = [];
+        $replacements = [];
+        $replacementCount = 0;
+
+        preg_match_all('/(?:%)((?:\\\?[\d\w_\\\]+\:\:)?[A-Z_0-9]+)(?:%)/', $value, $matches);
```
albe: I guess the non-capturing groups around the % can be removed to make the regex a little bit more readble

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Configuration.rst
```diff
@@ -148,6 +148,9 @@ Some examples:
   but is not recommended due to CGL (stringed class names should not
   have a leading backslash).
 
+``%ENV::HOME%``
+  Will be replaced by the environment variable's value.
```
albe: by the value of the environment variable "HOME"

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
```diff
@@ -651,24 +651,51 @@ protected function postProcessConfiguration(array &$configurations)
...
+        array_walk($matches[1], function ($match, $key) use ($matches, &$replacements, &$replacementCount) {
+            $replacementCount++;
+
+            if (defined($match)) {
+                $replacements[$matches[0][$key]] = constant($match);
```
albe: Had a little bit hard time to find what ``$matches[0][$key]`` actually is. Maybe we could assign that to an expressive variable beforehand, sth. like "$matchWithPercentages = ..."

albe:
Left a few comments but looks good otherwise

dfeyer:
Agree with the comment from @albe and left an open question about performance

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
```diff
@@ -651,24 +651,51 @@ protected function postProcessConfiguration(array &$configurations)
         foreach ($configurations as $key => $configuration) {
             if (is_array($configuration)) {
                 $this->postProcessConfiguration($configurations[$key]);
```
kitsunet: Probably would be, would also enable replacing keys (could be done in iterating as well, but not so nice). But IMHO that's a topic for another PR. This change won't impact performance much compared to what we have now, so....

kitsunet:
Ok, new commit has a totally new regex, making the code much clearer, the prefix is now `env:` instead of `ENV::`

dfeyer:
Work as expected, I just edit the first comment to match the syntax change

albe:
Very nice, I like :)

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
kitsunet: Oh, just realize whenever we want to do that, this has a couple of implications:
We would need to take care of (correct) string quoting and we need to figure out a way to understand if a match is the complete assignment and take care of the variable type thingy. So all in all that will not be that easy.

kitsunet:
Thanks for the quick round of reviews, feedbacks and stuff

mficzel:
Thanks for taking over ... I'm glad that this finally went into 3.3

PR #432, opened 2016-08-19
BUGFIX: ArrayFromObjectConverter correctly converts object properties with property name

albe:
Before the ``ArrayFromObjectConverter`` lost the property name of object properties when converting child properties.


albe:
Uhhh... "PHP Fatal error:  Uncaught TYPO3\Flow\Package\Exception\InvalidPackageKeyException: "paragonie/random_compat" is not a valid package key. " ??

albe:
Ah, everything OK now

kdambekalns:
Looks good, but… what about test?

albe:
Fair enough. There you go ;)

PR #433, opened 2016-08-19
FEATURE: Implement 'flow:cache:collectgarbagesessions' command

cerlestes:
This command is intended to trigger a manual garbage collection (via cron et al) with big Flow-based applications that have a lot of sessions.

With hundreds of thousands of sessions, even the fastest cache backend will take some time to clean them up; doing this probabilistically at the end of regular HTTP requests causes some unlucky requests to take very long, and, in some rare situations, can lead to data loss. Hence it's a better choice to do garbage collection from the CLI whenever needed.


albe:
Looks good from reading, thx!

cerlestes:
Superceded by #434

PR #434, opened 2016-08-19
FEATURE: Implement 'flow:session:collectgarbage' command

cerlestes:
This command is intended to trigger a manual garbage collection (via cron et al) with big Flow-based applications that have a lot of sessions.

With hundreds of thousands of sessions, even the fastest cache backend will take some time to clean them up; doing this probabilistically at the end of regular HTTP requests causes some unlucky requests to take very long, and, in some rare situations, can lead to data loss. Hence it's a better choice to do garbage collection from the CLI whenever needed.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/SessionCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/SessionCommandController.php
@@ -1,0 +1,46 @@
+<?php
+namespace TYPO3\Flow\Command;
+
+/*
+ * This file is part of the TYPO3.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Cli\CommandController;
+use TYPO3\Flow\Session\SessionManagerInterface;
+
+/**
+ * Command controller for managing sessions
+ *
+ * @Flow\Scope("singleton")
+ */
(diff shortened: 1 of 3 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Session/Session.php
```diff
@@ -602,10 +602,11 @@ public function destroy($reason = null)
      * Iterates over all existing sessions and removes their data if the inactivity
      * timeout was reached.
      *
+     * @param boolean $maximumSessionsToRemove How many sessions to remove per run
```
kdambekalns: Not a boolean…

TYPO3.Flow/Classes/TYPO3/Flow/Command/SessionCommandController.php
```diff
@@ -0,0 +1,46 @@
...
+     * instead that calls this command whenever.
+     *
+     * @return void
+     */
+    public function collectGarbageCommand()
+    {
+        $count = $this->sessionManager->getCurrentSession()->collectGarbage();
+        $this->outputLine('Removed %d expired sessions.', [$count]);
+    }
+}
```
kdambekalns: Add a newline here and StyleCI will be happy…

kdambekalns:
To me the test failures say: _this changes expected behaviour_. Previously you could just call `collectGarbage()` and it would respect the maximum defined. Now it defaults to "remove all sessions", and the per-run maximum is only used when `collectGarbage()` is triggered in `shutdownObject()`.

Suggestion: Default to 0 and respect the defined per-run maximum like before, and use -1 for "collect everything".

kitsunet:
@cerlestes would you update this? Or do you want someone to take over?

PR #436, opened 2016-08-19
TASK: Replace our LockStrategy by malkusch/lock

dfeyer:
This change implement Flock, Redis, Memcached and Semaphore based strategy. The external
library support only exclusive lock.

### DIFF
--- a/Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
+++ b/Neos.Utility.Lock/Classes/TYPO3/Flow/Utility/Lock/FlockLockStrategy.php
@@ -27,13 +27,6 @@ class FlockLockStrategy implements LockStrategyInterface
     protected $temporaryDirectory;
 
     /**
-     * File pointer if using flock method
-     *
-     * @var resource
-     */
-    protected $filePointer;
-
-    /**
      * FlockLockStrategy constructor.
      *
      * @param array $options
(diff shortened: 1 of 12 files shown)


dfeyer:
Tests need to be adapted, but this PR can be used to evaluate the performance impact of the exclusive lock.

8863c35e4a690862588a25c183384a98bdfcb1eb and 6383b814310d4cf5f91debd7691f70b90f4710c9 remove some lock to improve performance

kdambekalns:
Hm, this week I told @kitsunet that I'd love to keep out of anything related to locking by now… but still, just one arbitrary "example question":

Why do you consider it safe to skip the lock when checking the expiry? Is file_get_contents an atomic operation? If not, this will break if at the time of checking the expiry the cache file is written half-way.

dfeyer:
> Why do you consider it safe to skip the lock when checking the expiry? Is file_get_contents an atomic operation? If not, this will break if at the time of checking the expiry the cache file is written half-way.

Basically the only case where the isCacheFileExpired can be called with the acquire lock set to true is in the internalGet and in this PR the internalGet has acquired the lock on $pathAndFilename basically before in the current code (not the PR), the lock on $pathAndFilename is acquired 3 times in a few lines of code. 

> Hm, this week I told @kitsunet that I'd love to keep out of anything related to locking by now… but still, just one arbitrary "example question":

I prefer to keep locking related stuff out of my radar too, but I found this little branch on my local disk, so I push it ... and start working on it ... too bad I know, but hard to resist ;)

albe:
I'd really like you to take a look at #417 before spending too much time adjusting FileBackend to this.

dfeyer:
I close this one

PR #439, opened 2016-08-20
BUGFIX: Add missing Doctrine 2.5 mappings and fix Embeddables

albe:
This adds mappings for the @ORM\Index annotation attributes "flags" and "options" as well as the @ORM\UniqueConstraints annotation attribute "options".

Also, this adds a test for and fixes the Doctrine Embeddables.


albe:
The imports change included can easily be removed if anyone doesn't like it, but I found it easier to read and my IDE stopped complaining so much about unneccesarily full class names

albe:
Meh, cloning a proxied entity will somehow not load an embeddable and hence it turns out to be null... so unless the embedded properties are annotated nullable, database will blow up.

Since seemingly only embeddables inside a cloned doctrine proxy are affected, I consider this non-blocking. A skipped test to reproduce and possibly fix that later on is provided.

PR #437, opened 2016-08-20
BUGFIX: TypeHandling assumes vendor name to be uppercase

kitsunet:
The TypeHandling PARSE_TYPE_PATTERN assumes vendors in namespace do always
begin with uppercase letters. Reality shows, that this is not the case.
So this changes the regular expression accordingly.

PR #441, opened 2016-08-23
BUGFIX: Type Handling should not break when classnames contain underscores, or when classnames are lowerCamelCase

skurfuerst:
While underscores in class names are not used in Flow code itself, it might
happen that TypeHandling is used outside Flow - where having underscores
in class names is perfectly valid.

This change ensures that underscores in class names are recognized properly.


kitsunet:
Nice! I think together with #437  we should have tackled everyhing :)

skurfuerst:
@kitsunet can I help you getting #437 in (for 3.2)? Then I'd like to do another 3.2 release, as I need this change in a project :-)

All the best,
Sebastian

PR #442, opened 2016-08-23
BUGFIX: Doctrine controller should allow empty host for database

kitsunet:
The doctrine controller checked for correct configuration of a database
before calling the doctrine methods. As the drivers all have different
options the ``host`` is a not a conclusive check. The only thing that
can be checked is if a driver was set.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/DoctrineCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/DoctrineCommandController.php
@@ -121,18 +121,16 @@ class DoctrineCommandController extends CommandController
      */
     public function createCommand($output = null)
     {
-        // "driver" is used only for Doctrine, thus we (mis-)use it here
-        // additionally, when no path is set, skip this step, assuming no DB is needed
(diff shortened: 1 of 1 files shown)


kitsunet:
@robertlemke @kdambekalns This should fix the configuration problem regarding sockets and/or no host.

IMHO that makes much more sense for now. Please quickly check. Behavior is as expected.

kdambekalns:
What happens when no host is configured, though? Does some default come from us, Doctrine or even MySQL itself?

kitsunet:
@kdambekalns yes the EntityManagerFactory currently defaults that to 127.0.0.1 as was before via settings

kdambekalns:
I see. Well, I like. :)

TYPO3.Flow/Classes/TYPO3/Flow/Command/DoctrineCommandController.php
```diff
@@ -499,4 +485,14 @@ protected function handleException(\Exception $exception)
         $this->systemLogger->logException($exception);
         $this->quit(1);
     }
+
+    protected function isDatabaseConfigured()
```
robertlemke: lacks a DocComment

kitsunet:
Tests pass, will merge for now :)

PR #443, opened 2016-08-23
TASK: Improve code readability

albe:
This change avoids needless `else` statements and tries to minimize the amount of
 context required to keep in mind by reducing deep indentations.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Advice/AdviceChain.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Advice/AdviceChain.php
@@ -54,11 +54,9 @@ class AdviceChain
     {
         $this->adviceIndex++;
         if ($this->adviceIndex < count($this->advices)) {
-            $result = $this->advices[$this->adviceIndex]->invoke($joinPoint);
-        } else {
-            $result = $joinPoint->getProxy()->Flow_Aop_Proxy_invokeJoinpoint($joinPoint);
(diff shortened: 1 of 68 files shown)


kitsunet:
OMG, generally <3
Just this will break any single PR afterwards, so we need to find a good time to merge this.

albe:
Yep, also thought about that after pushing. Also will make upmerges more complicate... maybe retarget lower?

albe:
PS: kudos to @ocramius for making me take a look and walk through all that with his "Birthday wish" ;)

kitsunet:
Nah, for targeting lower I will it'S too big a change... But I mean, we have to chnage namespaces, move files etc. anyway. So upmerges will be a mess afterwards regardless.

albe:
True

kdambekalns:
So, this has a ton of merge conflicts now… anyone up for rebasing or merging in master? Do we consider it worth the effort?

Ocramius:
Easier to just re-run the CS rules against latest master.

albe:
If we agree to merge afterwards I'm willing to go the round again. As stated, it will cause merge conflicts in pretty much every other PR once merged, so we need to agree on the merge time first.

And I'd say a clear "yes!" for considering it worth the effort :)

albe:
So as discussed in https://discuss.neos.io/t/bigger-code-cleanup-effort/1983/3 I will restart the effort in several small PRs. Closing this now.

PR #446, opened 2016-08-30
!!! TASK: Remove deprecated setting injection

albe:
This removes the deprecated injection of settings via the @Flow\Inject annotation.


albe:
According to docblocks this should in fact have been removed with 3.3 already.

kitsunet:
Not the only deprecated thing to remove still I am sure of

albe:
Yeah, found a couple, was just the first I decided to quickly create the PR for :)

kitsunet:
Nice! 
https://github.com/neos/Neos.Contribute/pull/11
https://github.com/neos/Neos.Demo/pull/12

are needed to run the standard neos (dev) installation

PR #451, opened 2016-08-30
!!!TASK: Remove deprecated ``EarlyLogger`` from Flow

kitsunet:


kitsunet:
Well yes, @skurfuerst  there is a good reason, which is #466 
Feel free to apply this and try with empty caches, you will get an error about Logger not being singleton and therefore not being able to inject it.

PR #452, opened 2016-08-30
TASK: Allow configuration of doctrine second level cache

albe:
This change introduces the configuration settings `TYPO3.Flow.persistence.doctrine.secondLevelCache`
which allows configuring the second level cache that is available since Doctrine ORM 2.5
Currently, this allows for enabling the second level cache, setting default lifetime and region specific
lifetimes.
The used cache can be configured as `Flow_Persistence_Doctrine_SecondLevel`.


albe:
Hun... that failing test passes for me locally... also can't see the relation to the change?

albe:
Rebased in hope of passing tests.

PR #457, opened 2016-08-30
!!!TASK: Remove relations to party in ``Account`` and ``Security\Context``

kitsunet:
Since 3.0 something like a `Party` is not attached to the account directly anymore. Fetch your user/party/organization etc. instance on your own using Domain Services or Repositories.

One example is `TYPO3\Party\Domain\Service\PartyService`.


TYPO3.Flow/Classes/TYPO3/Flow/Security/Account.php
```diff
@@ -203,46 +203,6 @@ public function setCredentialsSource($credentialsSource)
     }
 
     /**
-     * Returns the party object this account corresponds to
```
daniellienert: `use TYPO3\Flow\Security\Exception as SecurityException;` can be removed

albe:
Tests for the get/setParty need to be removed too

kdambekalns:
Hm. Bummer. `BackendControllerSecurityTest` uses `setParty()` and the `RegistrationController` in the demo site does, too.

PR #459, opened 2016-08-30
!!! TASK: Remove deprecated methods from ``TemplateView``

kitsunet:


robertlemke:
Checked usages in the Neos distribution.

PR #462, opened 2016-08-30
FEATURE: Allow more flexible Doctrine Entity Manager configuration

albe:
Currently, we need to add new Configuration Settings for every feature of Doctrine that
is configurable through the Entity Manager.

This change adds a signal that triggers before the Doctrine Entity Manager is configured to
allow changing configuration settings of Doctrine programmatically.


kdambekalns:
TBH, having to add things to the settings make sure they are at least documented there…

Here is a suggestion: With adding that signal, move the setup code we have in place to a slot, document this and thus make sure everyone can easily learn how this is done and used. And it makes sure the feature works and will not break due to not being sued actively… ;)

albe:
Sounds good :)

kdambekalns:
So, who does it? ;)

albe:
I'll try to find time this weekend.

albe:
Sorry for losing track of this and thanks for picking it up Karsten!

> the PR title was misleading IMHO, so I changed it

Thanks :)

> the signal name is wrong. The EM isn't even created at that point… what about emitBeforeDoctrineEntityManagerCreation instead?

:+1:

> what code of ours could be moved to a slot?

- `$this->applySecondLevelCacheSettingsToConfiguration(...)`
- setting `sqlLogger` in config
- application of `eventSubscribers`, `eventListeners` from `buildEventManager()`
- application of `mappingTypes`, `filters`, `dql functions`

plus maybe:

- setting caches (metadata/query/result), could be done together with the second level cache
- setting FlowAnnotationDriver as driver implementation
- set proxy settings (directory/namespace/autogeneration)

that would tremendously reduce the Factory complexity and move those things to distinct slots for the specific concerns (caches, annotationdriver, proxy-generation, logging, doctrine hooks). Maybe makes sense to do this fine-grained separation in another PR as it is a bigger effort though.

albe:
Moved the above parts to a Slot and renamed the Signal. Also added another Signal/Slot for after creation of EntityManager.

Still unsure about the naming of the Slots and the class. I'm open to any suggestion.

albe:
Added some documentation, did some refactoring for readability and made tests pass. Only thing left to do, though that would be the extra mile, is to add more tests to `EntityManagerConfigurationTest` for the other settings we already have.

albe:
So can we get this reviewed and merged as is @kdambekalns ? I'd like to add the additional tests in a separate PR then, since that's basically a different issue

PR #466, opened 2016-08-31
BUGFIX: Configured scope for interfaces is not overwritten

kitsunet:
Currently the behaviour of a `Scope` annotation on an interface depends
on the amount of implementations. If there are multiple implementations
for the interface, then the configured interface scope is used all the time.
When there is only a single implementation, then that implementation will
override the configured interface scope.

This breaks for example in case of the `SystemLoggerInterface`. This
interface is marked singleton but the implementation is not because it is
reused for other loggers. The expected result is that the configured scope
of the interface is kept.

This change adjusts the behaviour and makes sure the `Scope` defined in
an interface is always respected.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
@@ -142,7 +142,7 @@ class ConfigurationBuilder
 
         // If an implementation class could be determined for an interface object configuration, set the scope for the
         // interface object configuration to the scope found in the implementation class configuration, but
-        // only if the interface doesn't have a specifically configured scope (eg. is prototype so far)
+        // only if the interface doesn't have a specifically configured scope (i.e. is prototype so far)
         foreach (array_keys($interfaceNames) as $interfaceName) {
             $implementationClassName = $objectConfigurations[$interfaceName]->getClassName();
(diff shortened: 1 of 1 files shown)


kitsunet:
This fixes #451

albe:
A test that shows the errorenous behaviour would really be nice.

kdambekalns:
So, I can only override the scope defined in an interface if it is "prototype", correct? Why not let the interface always win? And it would be great to have that added to the documentation on object configuration.

kitsunet:
Currently there is no way to know if the scope was set explicitly. Currently if you have an implementation and an interface were the implementation is singleton usually the interface is not configured (which will result in prototype in there) and in that case the implementation must win. Otherwise this would be a breaking change. It would be nice to add a flag to object configurations if the scope was set or just defaulted to prototype, but that is a bigger change that I didn't want to do for this bugfix (and IMHO is strictly not necessary to fix the bug).

kitsunet:
It would be nice actually if the interface could never dictate scope, but that will need serious changes in our Logger so that would also be a breaking change right now...

kitsunet:
This problem is loosely related to the discussion in #671 and should be fixed. Shameless plug, I know...

kitsunet:
This is merely a result of

kitsunet:
To clarify:

Currently the behavior depends on on the amount of implementations. If there are multiple implementations for the interface, then the configured interface scope is used all the time.
When there is only a single implementation, then that implementation will override the configured interface scope.

kdambekalns:
I adjusted the PR description to something I personally find easier to understand. @kitsunet please read it and verify it's not wrong now… :) Thanks!

kitsunet:
Wonderful! Thanks a lot

PR #469, opened 2016-08-31
TASK: Refactor absolute namespaces in  Flow / Command classes

daniellienert:
Additionally refactor legacy array syntax.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
@@ -20,8 +20,10 @@ use TYPO3\Flow\Package\PackageManagerInterface;
 use TYPO3\Flow\Persistence\PersistenceManagerInterface;
 use TYPO3\Flow\Resource\CollectionInterface;
 use TYPO3\Flow\Resource\Publishing\MessageCollector;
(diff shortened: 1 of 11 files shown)


albe:
The one `Resource` import could be also named `PersistentResource`, but since it's only used in the `@var` doc comment I see this not so much as issue. Looks good :)

TYPO3.Flow/Classes/TYPO3/Flow/Command/ResourceCommandController.php
```diff
@@ -189,7 +191,7 @@ public function cleanCommand()
                             }
                         }
                     }
-                    $this->outputLine('Removed %s resource object(s) from the database.', array(count($brokenResources)));
+                    $this->outputLine('Removed %s resource object(s) from the database.', [count($brokenResources)]);
                 break;
```
kitsunet: Why remove all this?

daniellienert: Sorry Christian - what do you mean?

kitsunet: Uhm, I think it was a display error in the diff view, now iwth the comments it looks normal so nevermind. It looked like the rest of the cases was deleted as well....

PR #470, opened 2016-08-31
TASK: Refactor absolute namespaces in Flow / Configuration classes

daniellienert:
Also refactor legacy array syntax


albe:
That one was easy :)

PR #471, opened 2016-08-31
BUGFIX: Do not throw exception when concurrently unlinking files

hlubek:
Fixes a race condition where concurrent emptying of a directory causes
errors because files were already unlinked by another process.

FLOW-283 #close
FLOW-345 #close

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Files.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Files.php
@@ -364,9 +364,10 @@ class Files
     /**
      * A version of unlink() that works on Windows regardless on the symlink type (file/directory).
      *
-     * If this method could not unlink the specified file, it will clear the stat cache for its filename and check if
-     * the file still exist. If it does not exist, this method assumes that the file has been deleted by another process
-     * and will return TRUE. If the file still exists though, this method will return FALSE.
+     * If this method could not unlink the specified file or it doesn't exist anymore (e.g. because of a concurrent
+     * deletion), it will clear the stat cache for its filename and check if the file still exist. If it does not exist,
+     * this method assumes that the file has been deleted by another process and will return TRUE. If the file still
+     * exists though, this method will return FALSE.
      *
      * @param string $pathAndFilename Path and name of the file or directory
      * @return boolean TRUE if file/directory was removed successfully
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Files.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Utility/Files.php
@@ -377,7 +377,11 @@ class Files
         try {
             // if not on Windows, call PHPs own unlink() function
             if (DIRECTORY_SEPARATOR === '/' || is_file($pathAndFilename)) {
-                return @\unlink($pathAndFilename);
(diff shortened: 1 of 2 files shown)


hlubek:
Mmmh, I could bet that I checked unit tests before submitting this, anyway I'l, fix the test.

albe:
Makes sense, though it should probably be marked breaking since it changes an @api method behaviour.

hlubek:
@albe Well, IMHO it's not really. If you didn't rely on `unlink` telling you if a file "existed" (before it's gone) that's true.

So to make this bulletbroof we _could_ introduce a parameter (like `ignoreMissingFile`) that enables the behaviour only for recursive deletion (which is mainly the underlying problem from concurrent cache flushing).

hlubek:
Or: we just fix the case for the `emptyDirectoryRecursively` function which duplicates some of the code but doesn't change the `unlink` behaviour at all.

hlubek:
My understanding of the fix by @robertlemke to this problem was that unlinking a file and then discovering it's not existing should never return `false`. But a case to handle this besides catching the exception was missing. Maybe he can tell more.

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Files.php
```diff
@@ -377,7 +378,11 @@ public static function unlink($pathAndFilename)
         try {
             // if not on Windows, call PHPs own unlink() function
             if (DIRECTORY_SEPARATOR === '/' || is_file($pathAndFilename)) {
-                return @\unlink($pathAndFilename);
+                if (!@\unlink($pathAndFilename)) {
```
gerhard-boden: Would `if (file_exists(` have a huge performance impact instead of `@`?

kdambekalns: It wouldn't be the same thing. Even after `file_exists` you still have the `unlink` fail, so… Even though `@` as evil as it gets, I think this is the only sane way.

gerhard-boden: We even do check for the file existence in the line above (did not see that at first). I understand now that checking if the file exists and trying to fail silently are two different concerns. Thanks for explaining!

TYPO3.Flow/Classes/TYPO3/Flow/Utility/Files.php
```diff
@@ -377,7 +378,11 @@ public static function unlink($pathAndFilename)
         try {
             // if not on Windows, call PHPs own unlink() function
             if (DIRECTORY_SEPARATOR === '/' || is_file($pathAndFilename)) {
```
gerhard-boden: Another question out of curiosity:
Why do we use an `OR` operator and not `AND` here? I mean we want to proceed if where not on Windows AND when  the path leads to file, correct?

kdambekalns: The reason is symlinks, yes. See https://github.com/hlubek/flow-development-collection/commit/04aaa0248744ae3d19f2094e2583fdfde53de268 an/or the `unlink` method in this very file.

kdambekalns: Second thought: yes, the comment is misleading/incomplete.

PR #472, opened 2016-08-31
TASK: Improve StringHelper documentation

hlubek:
Adds missing PHPDoc to StringHelper functions and add more examples.

### DIFF
--- a/TYPO3.Eel/Classes/TYPO3/Eel/Helper/StringHelper.php
+++ b/TYPO3.Eel/Classes/TYPO3/Eel/Helper/StringHelper.php
@@ -189,8 +189,13 @@ class StringHelper implements ProtectedContextAwareInterface
     /**
      * Match a string with a regular expression (PREG style)
      *
-     * @param string $string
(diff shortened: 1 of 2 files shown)


hlubek:
StyleCI failing is not caused by this commit, I'd rather not want to fix this here because of possible merge errors.

TYPO3.Eel/Classes/TYPO3/Eel/Helper/StringHelper.php
```diff
@@ -555,4 +574,5 @@ public function allowsCallOfMethod($methodName)
     {
         return true;
     }
+
```
albe: This addition of a line break is the reason for the Style-CI error, can easily be fixed ;)

hlubek: Ah okay, didn't see that, I'll push a fix

hlubek:
Should be fixed.

PR #473, opened 2016-08-31
FEATURE: Add String.pregSplit Eel helper

hlubek:
Adds a new helper method to the string helper for splitting strings with
a PREG pattern.

Example::

    String.pregSplit("foo bar   baz", "/\s+/") == ['foo', 'bar', 'baz']


kitsunet:
I mean it's already in use, but according to our "similar to javascript" idea it should be String.split allowing regex I guess?

hlubek:
@kitsunet That was the idea with `String.match` which is deprecated now ;) IMHO the PREG functions differ pretty much, but I'm open to adding a `String.split` function that emulates the JS behaviour.

kitsunet:
Let's do it like this, `String.split` exists already, unfortunately without regex, maybe we can add that later...

PR #474, opened 2016-09-01
TASK: Add `useMainRequest` to ActionViewHelper

gerdemann:
With this change it is possible to use the main Request instead
of the current Request.

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Link/ActionViewHelper.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Link/ActionViewHelper.php
@@ -72,13 +72,14 @@ class ActionViewHelper extends AbstractTagBasedViewHelper
      * @param array $additionalParams additional query parameters that won't be prefixed like $arguments (overrule $arguments)
      * @param boolean $addQueryString If set, the current query parameters will be kept in the URI
      * @param array $argumentsToBeExcludedFromQueryString arguments to be removed from the URI. Only active if $addQueryString = TRUE
-     * @param boolean $useParentRequest If set, the parent Request will be used instead of the current one
+     * @param boolean $useParentRequest If set, the parent Request will be used instead of the current one. Note: using this argument can be a sign of undesired tight coupling, use with care
      * @param boolean $absolute By default this ViewHelper renders links with absolute URIs. If this is FALSE, a relative URI is created instead
+     * @param boolean $useMainRequest If set, the main Request will be used instead of the current one. Note: using this argument can be a sign of undesired tight coupling, use with care
      * @return string The rendered link
      * @throws ViewHelper\Exception
      * @api
      */
-    public function render($action, $arguments = array(), $controller = null, $package = null, $subpackage = null, $section = '', $format = '', array $additionalParams = array(), $addQueryString = false, array $argumentsToBeExcludedFromQueryString = array(), $useParentRequest = false, $absolute = true)
+    public function render($action, $arguments = array(), $controller = null, $package = null, $subpackage = null, $section = '', $format = '', array $additionalParams = array(), $addQueryString = false, array $argumentsToBeExcludedFromQueryString = array(), $useParentRequest = false, $absolute = true, $useMainRequest = false)
     {
         $uriBuilder = $this->controllerContext->getUriBuilder();
         if ($useParentRequest) {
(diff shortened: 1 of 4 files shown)


TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Link/ActionViewHelper.php
```diff
@@ -88,6 +89,13 @@ public function render($action, $arguments = array(), $controller = null, $packa
             }
             $uriBuilder = clone $uriBuilder;
             $uriBuilder->setRequest($request->getParentRequest());
+        } elseif ($useMainRequest === true) {
+            $request = $this->controllerContext->getRequest();
+            if ($request->isMainRequest()) {
```
albe: Should we really throw an exception in this case and not just code around to allow that? I think in the useParentRequest case that makes sense, because "parentRequest" could unexpectedly be the main request in that case, but when the user explicitly wants the main request and is already in it - it's safe to just gracefully use the current request.

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Uri/ActionViewHelper.php
```diff
@@ -68,7 +69,15 @@ public function render($action, array $arguments = array(), $controller = null,
             }
             $uriBuilder = clone $uriBuilder;
             $uriBuilder->setRequest($request->getParentRequest());
+        } elseif ($useMainRequest === true) {
+            $request = $this->controllerContext->getRequest();
+            if ($request->isMainRequest()) {
```
albe: see above.

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Link/ActionViewHelper.php
gerdemann: I adjusted it

gerdemann:
@albe: It should be labeled with Fluid and not with Flow. 😉

albe:
Good catch :) thx!

albe:
Oh, and this should probably be retargeted at 3.1 at most, since 2.3 and 3.0 are supposed to only receive security updates by now (https://www.neos.io/features/release-roadmap.html - Flow ~ Neos + 1).

bwaidelich:
Could you please add a scenario on _why_ you need to access the main request?
When we added `useParentRequest` we had a quick discussion about that and thought that this should never be required because you should usually only ever know about the exact parent container of your context.. I doubt that we covered all cases back then. It would be interesting to hear some scenarios for that anyways. And IMO we should add it to the commit message and to the VHs inline docs.
Apart from that I agree that it should be targeted to master

gerdemann:
My current scenario is, that i have the Neos-request (first request), in this request i've a Plugin-NodeType (second request) with a TYPO3.Form in the action (third request) . So i'm in a Sub-Sub-Request and the ParentRequest is the Plugin-NodeType-Request, but for a link to a "main" route i need the MainRequest.

bwaidelich:
OK. I think we were probably thinking too restrictive back then fearing high coupling. But it's probably quite a useful addition.. Thanks for clarifying!

gerdemann:
I changed the target to master. What should I add to the CommitMessage and the inline docs?

bwaidelich:
> I changed the target to master
Thanks!
> What should I add to the CommitMessage and the inline docs?
Don't feel obliged but if you're at it you could maybe add a simple phrase to both, the `useParentRequest` and `useMainRequest` param doc comments saying something like
```
Note: using this argument can be a sign of undesired tight coupling, use with care
```

But nothing to block this obviously, unless @kitsunet sees troubles with this and the standalone Fluid efforts

kitsunet:
Nope, all good from my side!

gerdemann:
I added the comment to both params.

bwaidelich:
@gerdemann Great, thank you! @kitsunet could you remove your 👎

kitsunet:
Changed to <3

mficzel:
@gerdemann can you rebase this on current master or open the pr for pushes from maintainers? This change conflicts with the Standalone Fluid that was merged recently and has to be adjusted.

mficzel:
I hijacked this pr to #720 because it needed a rebase and i have no push rights.

PR #477, opened 2016-09-01
BUGFIX: adds missing migration for flow base tables

funkvn:
After a clean installation flow doctrine:migrationgenerate generates
migrations for this missing table/fields.


daniellienert:
Hey @funkvn, thanks a lot for your PR!!
The "solution" for your experienced problem is to run `flow doctrine:migrate` before generating a new migration.

PR #483, opened 2016-09-04
TASK: Refactor absolute namespaces in Flow / Object

daniellienert:

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/Configuration/ConfigurationBuilder.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/Configuration/ConfigurationBuilder.php
@@ -16,7 +16,12 @@ use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Annotations\Inject;
 use TYPO3\Flow\Annotations\InjectConfiguration;
 use TYPO3\Flow\Configuration\ConfigurationManager;
+use TYPO3\Flow\Log\SystemLoggerInterface;
+use TYPO3\Flow\Object\Exception as ObjectException;
 use TYPO3\Flow\Object\Exception\InvalidObjectConfigurationException;
+use TYPO3\Flow\Object\Exception\UnknownClassException;
+use TYPO3\Flow\Object\Exception\UnresolvedDependenciesException;
+use TYPO3\Flow\Reflection\ReflectionService;
 
 /**
  * Object Configuration Builder which can build object configuration objects
@@ -29,29 +34,29 @@ use TYPO3\Flow\Object\Exception\InvalidObjectConfigurationException;
 class ConfigurationBuilder
 {
     /**
-     * @var \TYPO3\Flow\Reflection\ReflectionService
+     * @var ReflectionService
      */
     protected $reflectionService;
 
     /**
-     * @var \TYPO3\Flow\Log\SystemLoggerInterface
+     * @var SystemLoggerInterface
      */
     protected $systemLogger;
 
     /**
-     * @param \TYPO3\Flow\Reflection\ReflectionService $reflectionService
+     * @param ReflectionService $reflectionService
      * @return void
      */
-    public function injectReflectionService(\TYPO3\Flow\Reflection\ReflectionService $reflectionService)
+    public function injectReflectionService(ReflectionService $reflectionService)
     {
         $this->reflectionService = $reflectionService;
     }
 
     /**
-     * @param \TYPO3\Flow\Log\SystemLoggerInterface $systemLogger
+     * @param SystemLoggerInterface $systemLogger
      * @return void
      */
-    public function injectSystemLogger(\TYPO3\Flow\Log\SystemLoggerInterface $systemLogger)
+    public function injectSystemLogger(SystemLoggerInterface $systemLogger)
     {
         $this->systemLogger = $systemLogger;
     }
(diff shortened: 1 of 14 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Object/ObjectManager.php
```diff
@@ -50,22 +50,22 @@ class ObjectManager implements ObjectManagerInterface
     /**
      * @var array
      */
-    protected $objects = array();
+    protected $objects = [];
 
     /**
      * @var array<\TYPO3\Flow\Object\DependencyInjection\DependencyProxy>
```
albe: Missing import here

TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyClass.php
```diff
@@ -43,32 +45,32 @@ class ProxyClass
...
      * @var array
      */
-    protected $constants = array();
+    protected $constants = [];
 
     /**
      * @var array
      */
-    protected $interfaces = array('\TYPO3\Flow\Object\Proxy\ProxyInterface');
+    protected $interfaces = ['\TYPO3\Flow\Object\Proxy\ProxyInterface'];
```
albe: Use ::class instead here too maybe?

TYPO3.Flow/Classes/TYPO3/Flow/Object/ObjectSerializer.php
```diff
@@ -348,7 +352,7 @@ protected function reconstituteArray($dataArray)
      *
      * @param string $type The collection type (class name) to create
      * @param array $dataArray The data array to reconstitute from
-     * @return \Doctrine\Common\Collections\Collection The reconstituted Collection
+     * @return Collection The reconstituted Collection
```
albe: I would actually prefer to make it explicit that we are referring the 3rd party doctrine collection here - either through "use Doctrine\Common\Collections as Doctrine;" or keep the FQCN.

TYPO3.Flow/Classes/TYPO3/Flow/Object/ObjectSerializer.php
```diff
@@ -165,15 +169,15 @@ public function serializeObjectAsPropertyArray($object, $isTopLevelItem = true)
...
+                            $this->reflectionService->isClassAnnotatedWith($propertyClassName, Flow\Entity::class)
+                            || $this->reflectionService->isClassAnnotatedWith($propertyClassName, Flow\ValueObject::class)
+                            || $this->reflectionService->isClassAnnotatedWith($propertyClassName, ORM\Entity::class)
                         )) {
                 $propertyArray[$propertyName][self::TYPE] = 'persistenceObject';
                 $propertyArray[$propertyName][self::VALUE] = get_class($propertyValue) . ':' . $this->persistenceManager->getIdentifierByObject($propertyValue);
             } elseif (is_object($propertyValue)) {
                 $propertyObjectName = $this->objectManager->getObjectNameByClassName($propertyClassName);
-                if ($this->objectManager->getScope($propertyObjectName) === \TYPO3\Flow\Object\Configuration\Configuration::SCOPE_SINGLETON) {
+                if ($this->objectManager->getScope($propertyObjectName) === Configuration\Configuration::SCOPE_SINGLETON) {
```
albe: That could very well be just `Configuration` (and is somewhere above) ;)

TYPO3.Flow/Classes/TYPO3/Flow/Object/ObjectSerializer.php
```diff
@@ -146,13 +150,13 @@ public function serializeObjectAsPropertyArray($object, $isTopLevelItem = true)
...
                 $propertyArray[$propertyName][self::TYPE] = 'SplObjectStorage';
-                $propertyArray[$propertyName][self::VALUE] = array();
+                $propertyArray[$propertyName][self::VALUE] = [];
 
                 foreach ($propertyValue as $storedObject) {
                     $propertyArray[$propertyName][self::VALUE][] = spl_object_hash($storedObject);
                     $this->serializeObjectAsPropertyArray($storedObject, false);
                 }
-            } elseif (is_object($propertyValue) && $propertyValue instanceof \Doctrine\Common\Collections\Collection) {
+            } elseif (is_object($propertyValue) && $propertyValue instanceof Collection) {
```
albe: As stated below, prefer explicit external dependencies.

TYPO3.Flow/Classes/TYPO3/Flow/Object/Configuration/ConfigurationBuilder.php
```diff
@@ -412,7 +417,7 @@ protected function autowireArguments(array &$objectConfigurations)
      *
      * @param array &$objectConfigurations
      * @return void
-     * @throws \TYPO3\Flow\Object\Exception if an injected property is private
+     * @throws Exception if an injected property is private
```
albe: `Object\Exception`

kitsunet:
Couldn't push my fix for the two comments @daniellienert please give access...

TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyClass.php
daniellienert: I tried that but a ::class doesn't has a leading "\" and a `protected $interfaces = ['\' . ProxyInterface::class];` is not valid in PHP 5.6

TYPO3.Flow/Classes/TYPO3/Flow/Object/ObjectSerializer.php
daniellienert: I wanted to avoid the use statement here. But if you both think it would better that way ..

daniellienert:
@kitsunet I granted you access to the repo ...

TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyClass.php
kdambekalns: Exactly like that it cannot be valid, as the backslash escapes the single quote… But in general that works, see https://3v4l.org/H6eAe

kitsunet:
That should fix the failing test and take care of the last comment.

kitsunet:
Note: We should probably refactor the interface and trait behavior to accept also classnames without leading backslash and add it later in the generation of the proxy class.

PR #491, opened 2016-09-06
TASK: Refactor absolute namespaces in Flow / SignalSlot and Utility

daniellienert:


robertlemke:
Thanks! Just wondering if we should target to older branches as well or if we don't, what it will mean for conflicts in future upmerges ... Any opinions on that? @kdambekalns

kdambekalns:
Well… doing those cleanups in older branches will make merging easier. Then again with rerere this is manageable. 

That being said, I'd welcome this in older branches. :)

kitsunet:
That's why this is targeted at 3.0 ;)
It's what we discussed on the Sprint.

PR #493, opened 2016-09-07
BUGFIX: Port is set correctly when Trusted Proxy only sends protocol override

albe:
Before the port of the request URI was not set correctly if the Trusted Proxy would 
send an `X-Forwarded-Proto: HTTPS` header without also specifying a forwarded port.
This resulted in generated URLs like "https://acme.com:80".

The change fixes that by doing a case insensitive check for the protocol to be
'https' and set port 443 accordingly.

### DIFF
--- a/TYPO3.Flow/Tests/Unit/Http/Component/TrustedProxiesComponentTest.php
+++ b/TYPO3.Flow/Tests/Unit/Http/Component/TrustedProxiesComponentTest.php
@@ -440,6 +440,12 @@ class TrustedProxiesComponentTest extends UnitTestCase
                 'requestUri' => 'https://acme.com',
                 'expectedUri' => 'https://acme.com:80',
             ),
+            array(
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/Tests/Unit/Http/Component/TrustedProxiesComponentTest.php
```diff
@@ -440,6 +440,12 @@ public function forwardHeaderTestsDataProvider()
                 'requestUri' => 'https://acme.com',
                 'expectedUri' => 'https://acme.com:80',
             ),
+            array(
```
daniellienert: Small nitpick: Use short array syntax.

kdambekalns: Not if the other declarations use the long form. And since it's better not to mix functional and cleanup, let's leave it like that.

PR #499, opened 2016-09-09
BUGFIX: Http RequestHandler returns correct instance of Request and Response

albe:
Before the Http RequestHandler would always return the initial instance of the `Request` and `Response`
before the ComponentChain.
Since the ComponentChain is supposed to possibly replace the Request and Response instances, this
lead to cases where retrieving the current Request from the RequestHandler would not match the actual
Request that is in place inside the MVC Controller for example.

This change fixes that by always delegating the request and response getters to the ComponentContext.

Also, this change fixes a regression introduced with 3.3 when generating URLs behind a https offload proxy. The UriBuilder was using the untouched Request instance ending up with generating URLs like `https://acme.com:80`.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php
@@ -46,6 +46,11 @@ class RequestHandler implements HttpRequestHandlerInterface
     protected $baseComponentChain;
 
     /**
+     * @var Component\ComponentContext
+     */
+    protected $componentContext;
+
+    /**
      * The "http" settings
      *
      * @var array
@@ -111,8 +116,8 @@ class RequestHandler implements HttpRequestHandlerInterface
             $this->request->setBaseUri(new Uri($this->settings['http']['baseUri']));
         }
 
-        $componentContext = new ComponentContext($this->request, $this->response);
-        $this->baseComponentChain->handle($componentContext);
+        $this->componentContext = new ComponentContext($this->request, $this->response);
+        $this->baseComponentChain->handle($this->componentContext);
         $this->response = $this->baseComponentChain->getResponse();
 
         $this->response->send();
@@ -129,7 +134,7 @@ class RequestHandler implements HttpRequestHandlerInterface
      */
     public function getHttpRequest()
     {
-        return $this->request;
+        return $this->componentContext->getHttpRequest();
     }
 
     /**
@@ -140,7 +145,7 @@ class RequestHandler implements HttpRequestHandlerInterface
      */
     public function getHttpResponse()
     {
-        return $this->response;
+        return $this->componentContext->getHttpResponse();
     }
 
     /**
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php
@@ -108,6 +108,7 @@ class RequestHandler implements HttpRequestHandlerInterface
         // Create the request very early so the Resource Management has a chance to grab it:
         $this->request = Request::createFromEnvironment();
         $this->response = new Response();
+        $this->componentContext = new ComponentContext($this->request, $this->response);
 
         $this->boot();
         $this->resolveDependencies();
@@ -116,7 +117,6 @@ class RequestHandler implements HttpRequestHandlerInterface
             $this->request->setBaseUri(new Uri($this->settings['http']['baseUri']));
         }
 
-        $this->componentContext = new ComponentContext($this->request, $this->response);
         $this->baseComponentChain->handle($this->componentContext);
         $this->response = $this->baseComponentChain->getResponse();
 
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/ComponentContext.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/ComponentContext.php
@@ -21,7 +21,10 @@ use TYPO3\Flow\Http\Response;
  * An instance of this class will be passed to each component of the chain allowing them to read/write parameters to/from it.
  * Besides handling of the chain is interrupted as soon as the "cancelled" flag is set.
  *
+ * The instance will be created before the bootstrap, so AOP/DI proxying is not possible.
+ *
  * @api
+ * @Flow\Proxy(false)
  */
 class ComponentContext
 {


mficzel:
This is not all that has to be changed here. I get the following php error.

    Fatal error: Uncaught Error: Call to a member function getHttpRequest() on null in
    __project_folder__Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php:138 Stack trace: 
    #0 __project_folder__Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Log/Logger.php(240): TYPO3\Flow\Http\RequestHandler->getHttpRequest() 
    #1 __project_folder__Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Log/Logger.php(194): TYPO3\Flow\Log\Logger->renderRequestInfo() 
    #2 __project_folder__Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Log/Logger.php(164): TYPO3\Flow\Log\Logger->renderErrorInfo(Object(Error)) 
    #3 __project_folder__Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Log/Logger.php(133): TYPO3\Flow\Log\Logger->logError(Object(Error), Array) 
    #4 __project_folder__Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Error/AbstractExceptionHandler.php(94): TYPO3\Flow\Log\Logger->logThrowable(Obje in __project_folder__Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php on line 138

I am looking into the details of that problem.

TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php
```diff
@@ -111,8 +116,8 @@ public function handleRequest()
             $this->request->setBaseUri(new Uri($this->settings['http']['baseUri']));
         }
 
-        $componentContext = new ComponentContext($this->request, $this->response);
-        $this->baseComponentChain->handle($componentContext);
+        $this->componentContext = new ComponentContext($this->request, $this->response);
```
mficzel: This happens to late here and causes errors because the getRequest is called somehow in the $this->boot sequence. I suggest to move this to the top of the method.

mficzel:
With the initialisation of `$this->componentContext...` moved before $this->boot this works in my setup as expected. 

PS: Neos does not need a second change since the domain repository used the getRequest of the activeRequestHandler.

TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php
kitsunet: But that again is also not easily possible, it would move the ComponentContext and ALL components (also user land) to a pre boot time, which has tremendous implications like no Configuration, no Persistence, no Caches no nothing basically...

albe: Ah, well, that makes sense somehow - thanks for finding that.

kitsunet: Yep, ok, so we could only initialize the component context pre boot, that will work. We should declare it as Flow\Proxy false though and comment that it is created pre boot.

albe: Good point, I guess it even makes a lot of sense generally to make this unproxyable.

kitsunet:
maybe add deprecated to the internal properties request and response, so we don't forget to remove them at some point?

mficzel:
That works for me. Tested in Flow 3.3 with Neos 2.3

albe:
@kitsunet already have a PR for that open: #500

kitsunet:
Perfect!

gerhard-boden:
Tested on 3.3. and this also fixes the issue that `https://domain.com:80` URIs were generated because some checks where moved from `Http/Request.php`` to ``TrustedProxiesComponent``.

albe:
Guess there's enough reviews by now. Gonna merge tomorrow if no one objects.

albe:
FYI: I tried to create a functional test that shows the regression, but I could not get the functional test to fail for some unknown reasons (the Http\RequestHandler isn't even used in that case and the FunctionalTestRequestHandler has the same illbehaviour that is fixed in this PR). Seems the functional test bootstrap has some side-effects that mitigate the bug. I'll try to provide a PR that deals with this as soon as I find the cause.

bwaidelich:
It seems this is still broken with Neos 2.3. At least I get the same fatal error as described in #issuecomment-246266084 in the setup of a fresh 2.3 installation?!

TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php
```diff
@@ -129,7 +134,7 @@ public function handleRequest()
      */
     public function getHttpRequest()
     {
-        return $this->request;
+        return $this->componentContext->getHttpRequest();
```
bwaidelich: This only works if `handleRequest()` has been invoked before.. Why not check for null first and prevent the Fatal Error that occurs otherwise?

bwaidelich: ..even though this won't be the solution to the issue I commented

bwaidelich:
It seems the issue is related to #514

bwaidelich:
> It seems the issue is related to #514 

No, it's not (sorry for the spam, I'm a bit desperate).
It's because we use a custom `RequestHandler` in the Setup package that inherits from Flows `RequestHandler` and overrides it's `handleRequest()` method..

TYPO3.Flow/Classes/TYPO3/Flow/Http/RequestHandler.php
albe: Good point, I'll provide a follow-up, though normally you should never call `getHttpRequest()` *before* `handleRequest()` ;)

bwaidelich: famous last words ;)
But as the null-check doesn't really solve the core issue, it might be worth to think about a completely different approach to the whole HTTP handling. So, I'd say: don't worry about the follow-up but lets collect some ideas for a more stable HTTP foundation (backwards compatible ofc)

PR #500, opened 2016-09-09
!!! TASK: Remove internal properties request and response from RequestHandler

albe:
Since the Request and Response instances are supposed to change inside the ComponentChain,
it is error-prone to keep a reference to the initial instances inside the RequestHandler.
This change removes the class properties $request and $response and instead uses local variables.

This is marked breaking only for the reason that some RequestHandler implementations could
exist that still somehow depend on this internal detail. It is not really breaking as those properties
were never part of the public api though.


dfeyer:
Make sense, thanks for the PR

albe:
Rebased.

PR #502, opened 2016-09-10
BUGFIX: Temporary files are correctly moved across volumes

hhoechtl:
PHP throws a operation not permitted warning when using rename across
volumes, which happens e.g. if you have FLOW_PATH_TEMPORARY_BASE pointing
to a different (more performant) volume.

From the php documentation:
> More explicitly, rename() may still return (bool) true, despite the warnings that result from the underlying calls to chown() or chmod(). This behavior can be misleading absent a deeper understanding of the underlying mechanics. To rename across filesystems, PHP "fakes it" by calling copy(), unlink(), chown(), and chmod() (not necessarily in that order). See PHP bug #50676 for more information.

> On UNIX-like operating systems, filesystems may be mounted with an explicit uid and/or gid (for example, with mount options "uid=someuser,gid=somegroup"). Attempting to call rename() with such a destination filesystem will cause an "Operation not permitted" warning, even though the file is indeed renamed and rename() returns (bool) true.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Storage/WritableFileSystemStorage.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Storage/WritableFileSystemStorage.php
@@ -162,10 +162,10 @@ class WritableFileSystemStorage extends FileSystemStorage implements WritableSto
         if (!file_exists(dirname($finalTargetPathAndFilename))) {
             Files::createDirectoryRecursively(dirname($finalTargetPathAndFilename));
         }
-        if (rename($temporaryFile, $finalTargetPathAndFilename) === false) {
-            unlink($temporaryFile);
+        if (copy($temporaryFile, $finalTargetPathAndFilename) === false) {
             throw new StorageException(sprintf('The temporary file of the file import could not be moved to the final target "%s".', $finalTargetPathAndFilename), 1381156103);
         }
+        unlink($temporaryFile);
 
         $this->fixFilePermissions($finalTargetPathAndFilename);
     }


TYPO3.Flow/Classes/TYPO3/Flow/Resource/Storage/WritableFileSystemStorage.php
```diff
@@ -162,10 +162,10 @@ protected function moveTemporaryFileToFinalDestination($temporaryFile, $finalTar
         if (!file_exists(dirname($finalTargetPathAndFilename))) {
             Files::createDirectoryRecursively(dirname($finalTargetPathAndFilename));
         }
-        if (rename($temporaryFile, $finalTargetPathAndFilename) === false) {
-            unlink($temporaryFile);
+        if (copy($temporaryFile, $finalTargetPathAndFilename) === false) {
             throw new StorageException(sprintf('The temporary file of the file import could not be moved to the final target "%s".', $finalTargetPathAndFilename), 1381156103);
         }
+        unlink($temporaryFile);
```
kitsunet: The problem with copy and unlink is that it's not atomic. But should be fine in this case.

albe: Yep, but the whole moveTemporaryFileToFinalDestination is not atomic anyway.

hhoechtl:
@kitsunet Is there a vote missing or can you merge it?

hhoechtl:
Hmm, I was pretty sure I hat 2.0 as target so it can be upmerged. @kitsunet did you change that or didn't I pay attention?

albe:
2.0? Even 3.0 should only get security related bugfixes by now, so I guess that's fine as is.

kitsunet:
Don't confuse Flow and Neos @hhoechtl this is Flow 3.0 so Neos 2.0

hhoechtl:
Uff, somehow I feel like monday morning. 😅  Sorry

PR #504, opened 2016-09-12
BUGFIX: Wrong indentation in policy examples

rpflamm:
The indentation of the examples are wrong. They cause an exception if they are copied without the knowledge of the right indentation.


robertlemke:
Hey @rpflamm thanks a lot for spotting these wrong examples!

PR #505, opened 2016-09-14
FEATURE: Dynamic XLIFF label overrides

nezaniel:
Provides Flow core functionality for label overrides between packages based on loading order. Also prepares XLIFF 2.0 support and additional overrides from arbitrary XLIFF sources.

see https://discuss.neos.io/t/project-proposal-implementing-label-overrides-and-fallbacks-in-flow/1118
related: FLOW-61


TYPO3.Flow/Classes/TYPO3/Flow/I18n/Translator.php
```diff
@@ -166,6 +166,12 @@ public function translateByOriginalLabel($originalLabel, array $arguments = [],
      */
     public function translateById($labelId, array $arguments = [], $quantity = null, Locale $locale = null, $sourceName = 'Main', $packageKey = 'TYPO3.Flow')
     {
+        $combinedIdentifiers = explode(':', $labelId);
```
kitsunet: Do we want this special logic in here?

kitsunet:
I am not very fond of the "V12" namespace, but looks good otherwise on a first glance.

TYPO3.Flow/Classes/TYPO3/Flow/I18n/Translator.php
nezaniel: It was supposed to support stuff like <f:translate id="Package:Source:TranslationUnitId" />
Thinking about XLIFF 2.0, where files have identifiers themselves and can be resolved without a package name or source whatsoever, this seems no longer necessary. But that leads to a different topic

nezaniel:
V12 (in contrast to V20, which is not yet finished and thus stashed) is supposed to separate XLIFF 1.2 logic from 2.0 logic. In the current, simple version that includes only the parser, which may lead to the impression that an own namespace is overkill at this point. At the moment I'd be okay with omitting it, it might come in handy later on though. But since the parsers themselves are everything but public API, we could add it later on easily I think.

PR #507, opened 2016-09-14
TASK: Refactor absolute namespaces in Flow / Unit Tests Log - Package

daniellienert:


daniellienert:
Hey, This is the last refactored unit test block. As the unit tests are rather uncritical I would like to merge this PR in the next days if no one objects.

PR #512, opened 2016-09-16
BUGFIX: Prevent nesting level too deep error in checkbox view helper

aertmann:
To prevent a recursive comparison which can lead to this error, strict comparison is used which only compares the reference.


aertmann:
See http://stackoverflow.com/a/19097159/2311403 for details

aertmann:
Similar to https://github.com/neos/neos-development-collection/pull/762

PR #513, opened 2016-09-17
BUGFIX: ObjectAccess supports property bag implementation classes

albe:
Before any class that would implement any of the methods `get()`, `has()`, `set()` or `is()` would make
ObjectAccess fail when attempting to get/set empty propertyNames. Those methods are commonly
used when implementing the propertyBag pattern.

This change fixes that by making sure that the accessor methods are actually longer than the plain prefixes.

### DIFF
--- a/TYPO3.Flow/Tests/Unit/Reflection/Fixture/DummyClassWithGettersAndSetters.php
+++ b/TYPO3.Flow/Tests/Unit/Reflection/Fixture/DummyClassWithGettersAndSetters.php
@@ -30,6 +30,8 @@ class DummyClassWithGettersAndSetters
     public $publicProperty;
     public $publicProperty2 = 42;
 
+    protected $propertyBag = [];
+
     public function setProperty($property)
     {
         $this->property = $property;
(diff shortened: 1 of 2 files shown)


dfeyer:
;)

albe:
Would merge the next days if no one objects.

bwaidelich:
Nice one!
Just an idea that just popped up: Maybe it's more flexible (and even faster?) to use a regex in the form `/^is[A-Z]/`? That would also skip things like `issueNumber()` etc

TYPO3.Flow/Tests/Unit/Reflection/Fixture/DummyClassWithGettersAndSetters.php
```diff
@@ -92,4 +94,19 @@ protected function getPrivateProperty()
...
+
+    public function get($property)
+    {
+        return isset($this->propertyBag[$property]) ? $this->propertyBag[$property] : null;
+    }
+
+    public function set($property, $value)
+    {
+        $this->propertyBag[$property] = $value;
+    }
```
bwaidelich: Does this really break without the change in `ObjectAccess`?

PR #642, opened 2016-09-25
TASK: Add missing model annotation in documentation

DanielSiepmann:
- Without this annotation it will not be possible to generate doctrine
  migration.

### DIFF
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/ResourceManagement.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/ResourceManagement.rst
@@ -242,6 +242,8 @@ Provided that the ``Image`` class has a ``$title`` and a ``$originalResource`` p
 that they are accessible through ``setTitle()`` and ``setOriginalResource()`` respectively the
 above code will work just as expected::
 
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
Ok, thanks.

For me it's still a bug. As bugs, in my opinion, can also exist in documentation and the documented code will not fully work. But I'm fine with using Task here.

robertlemke:
@DanielSiepmann thanks for the fix! Can you rebase it on top of Flow 3.1 and then force-push it again? I tried to use the change-branch-feature in Github, but the would result in too many merge commits ...

DanielSiepmann:
I'll try tonight @robertlemke

DanielSiepmann:
Will it be adjusted for further releases, or do I have to create another pull request for later versions?

kitsunet:
No, we will upmerge this change into the newer branches up to master.

albe:
Hm, we somehow totally forgot to merge this at any point after the retarget... sorry! Guess by now this should actually be 3.2 or 3.3 but that's easy

albe:
Would merge (into 3.2) this evening if no one objects.

kitsunet:
Please go for it :)

PR #643, opened 2016-09-26
TASK: Update Installation Doc

chewbakartik:
Changing the web group name for the installation instructions on linux from the mac web group name to a more common linux one.


chewbakartik:
Same PR but for the 3.1 branch.  Please review @albe, @robertlemke, @dlubitz.

albe:
Would merge by end of this week if no one objects.

PR #644, opened 2016-09-27
FEATURE: Dependency Injection and AOP for final classes

### LINKED ISSUE #496: Better support for final classes and private methods
## Problem Description
In order to make use of dependency injection and AOP in general user-land classes must not be `final`. Besides, injected properties can not be `private` (but have to be `protected` or `public`).
Having everything extensible might be convenient but it can also have undesired side-effects so we should at least allow users to define a more strict API, using the `final` and `private` keywords.

This should be possible for example:

```php
final class SignUpUser implements CommandInterface
{
    private $username;
    private $password;
    private $hashedPassword;

    /**
     * @Flow\Inject
     * @var HashService
     */
    private $hashService;

    public function __construct(string $username, string $password)
    {
        $this->username = $username;
        $this->password = $password;
    }

   public function initializeObject()
   {
        $this->hashedPassword = $this->hashService->hashPassword($this->password);
        unset($this->password);
   }
}
```

This currently doesn't work because `final` classes are completely skipped from Flows object configuration. Besides `$hashService` wouldn't be allowed to be `private`.

Note: In a related matter I'd love to be able to allow `initializeObject()` to be `protected`.

## Suggestion

1. Don't skip `final` classes in `ConfigurationBuilder::buildObjectConfigurations()`
2. Don't skip `final` classes in `ProxyClassBuilder::build()` and `ProxyClassBuilder::getProxyableClasses()`
3. Adjust regex in `Proxy\Compiler::cacheOriginalClassFileAndProxyCode()` to remove the `final` keyword from `<class>_Original` declaration (and maybe add it to the new proxied `<class>` declaration)

Apart from adjusted tests there's some more work required to make this work. i.e. Flow complains about missing annotations:  `[Semantical Error] The class "\TYPO3\Flow\Annotations\Entity" is not annotated with @Annotation.`

bwaidelich:
This adds support for proxied ``final`` classes.

Previously those were always skipped from proxy building disallowing Dependency Injection.
Besides final classes could not be targeted by AOP advices.

With this change, final classes are now also proxied by default.
To *disable* AOP/DI for those the already existing ``Proxy`` annotation can be used::

    /**
     * @Flow\Proxy(false)
     */
    final class SomeClass
    {
        // ...

Background:

Marking classes ``final`` is an important tool for framework code as it allows to define extension points
more explicitly, but until now we had to avoid the ``final`` keyword in order to support DI and AOP.

Resolves #496


bwaidelich:
Tested with Flow & Neos base distributions

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -202,7 +202,7 @@ public function build()
     {
         $allAvailableClassNamesByPackage = $this->objectManager->getRegisteredClassNames();
         $possibleTargetClassNames = $this->getProxyableClasses($allAvailableClassNamesByPackage);
-        $actualAspectClassNames = $this->reflectionService->getClassNamesByAnnotation(\TYPO3\Flow\Annotations\Aspect::class);
+        $actualAspectClassNames = $this->reflectionService->getClassNamesByAnnotation(Flow\Aspect::class);
```
bwaidelich: I didn't import new namespaces but I couldn't resist simplifying this one ;)

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -282,12 +282,20 @@ protected function getProxyableClasses(array $classNamesByPackage)
...
-                        $proxyableClasses[] = $className;
+                if (in_array(substr($className, 0, 15), $this->blacklistedSubPackages)) {
+                    continue;
+                }
+                if ($this->reflectionService->isClassAnnotatedWith($className, Flow\Aspect::class)) {
+                    continue;
+                }
+                if ($this->reflectionService->isClassFinal($className)) {
+                    /** @var Flow\Proxy $proxyAnnotation */
+                    $proxyAnnotation = $this->reflectionService->getClassAnnotation($className, Flow\Proxy::class);
```
bwaidelich: I just restructured this a bit in order to avoid even more nesting, but this is the new behavior:
Previously we skipped if `isClassFinal()` was true, now only if it's true and the class is not whitelisted via `proxy` annotation.

TYPO3.Flow/Classes/TYPO3/Flow/Object/Configuration/ConfigurationBuilder.php
```diff
@@ -80,10 +80,6 @@ public function buildObjectConfigurations(array $availableClassAndInterfaceNames
                     continue;
                 }
 
-                if ($this->reflectionService->isClassFinal($classOrInterfaceName)) {
-                    continue;
-                }
-
```
bwaidelich: Without this code, we create object configuration for all final classes, but I don't think that this would cause any harm

TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/Compiler.php
```diff
@@ -227,12 +230,12 @@ protected function cacheOriginalClassFileAndProxyCode($className, $pathAndFilena
...
-        $classCode = preg_replace_callback('/^([a-z\s]*)(interface|class)\s+([a-zA-Z0-9_]+)/m', function ($matches) use ($pathAndFilename, $classNameSuffix) {
+        $classCode = preg_replace_callback('/^([a-z\s]*?)(final\s+)?(interface|class)\s+([a-zA-Z0-9_]+)/m', function ($matches) use ($pathAndFilename, $classNameSuffix, $proxyClassCode) {
             $classNameAccordingToFileName = basename($pathAndFilename, '.php');
-            if ($matches[3] !== $classNameAccordingToFileName) {
-                throw new Exception('The name of the class "' . $matches[3] . '" is not the same as the filename which is "' . basename($pathAndFilename) . '". Path: ' . $pathAndFilename, 1398356897);
+            if ($matches[4] !== $classNameAccordingToFileName) {
+                throw new Exception('The name of the class "' . $matches[4] . '" is not the same as the filename which is "' . basename($pathAndFilename) . '". Path: ' . $pathAndFilename, 1398356897);
             }
-            return $matches[1] . $matches[2] . ' ' . $matches[3] . $classNameSuffix;
+            return $matches[1] . $matches[3] . ' ' . $matches[4] . $classNameSuffix;
```
bwaidelich: This removes any `final` modifier from the `*_Original` class (it will be moved to the proxy below)

TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyClass.php
```diff
@@ -206,7 +206,12 @@ public function render()
         $namespace = $this->namespace;
         $proxyClassName = $this->originalClassName;
         $originalClassName = $this->originalClassName . \TYPO3\Flow\Object\Proxy\Compiler::ORIGINAL_CLASSNAME_SUFFIX;
-        $abstractKeyword = $this->reflectionService->isClassAbstract($this->fullOriginalClassName) ? 'abstract ' : '';
+        $classModifier = '';
+        if ($this->reflectionService->isClassAbstract($this->fullOriginalClassName)) {
+            $classModifier = 'abstract ';
+        } elseif ($this->reflectionService->isClassFinal($this->fullOriginalClassName)) {
+            $classModifier = 'final ';
+        }
```
bwaidelich: This moves and `final` modifier from the `*_Original` class to the proxy, just like we already did with the `abstract` modifier.

robertlemke:
@bwaidelich what would speak against automatically white-listing `final` classes, even without the proxy annotation - any drawbacks?

bwaidelich:
@robertlemke 
> what would speak against automatically white-listing final classes

That would also be my favorite, but I thought the discussion on #496 suggested otherwise?

> any drawbacks?

If I just disable the proxy-annotation checks I get an error during proxy building:
```
[Semantical Error] The class "\TYPO3\Flow\Annotations\Entity" is not annotated with @Annotation.
```
(because the `@Annotation`-annotation is not copied/moved to the proxy)
Nothing that couldn't be fixed I guess

robertlemke:
@bwaidelich I over-read that part with "Proxy(true)" in the #496 discussion. 

Still, I'd prefer the support of final proxies without the annotation, it just just work. And I don't see a real drawback in practice for it. If you don't want your final class to be proxied, you can still state "Proxy(false)", but why should you?

bwaidelich:
> Still, I'd prefer the support of final proxies without the annotation

I agree. Do you have an idea on how to solve the error above? Excluding classes marked `@Annotation` explicitly?

albe:
Well, the main point for making it "opt-in" is, that if you already choose to make a class `final`, you are very explicit about not wanting that class to be extended or modified outside it's scope. So the idea is that this implicitly also includes AOP.

It's probably a decision of *"is automatic DI support more important than securing your* **final** *class for unwanted extension by default?"*.

But the fact that it needs a workaround for `@Annotation`s is probably a small sign in itself.

bwaidelich:
>  to make a class final, you are very explicit about not wanting that class to be extended [...] includes AOP.

I wouldn't say that it includes AOP/DI. For those to opt-in/out the `@Flow\Proxy` annotation is used. The `final` keyword lets you define a stricter PHP API but IMO that is a different thing

dfeyer:
> I wouldn't say that it includes AOP/DI

Could be nice to have DI working on final class, but I think AOP must be disabled to respect the intention of the the developper, it's final WTF ;)

bwaidelich:
> Could be nice to have DI working on final class, but I think AOP must be disabled

That was my first impression, too. But playing with it I think it would be a big shame if we lost the ability to security-protect methods of final classes, log access to those, ....
Again, I don't think that it's the developers intention to disable AOP when he/she marks a class `final`. To disable AOP use `@Flow\Proxy(false)` (and maybe we should document that better).

If we disagree on that I can happily do a little poll

albe:
Yeah, let's do a quick poll on what the team thinks overall. I wouldn't mind too much having it opt-out, but my feeling is it should be opt-in.

> lost the ability to security-protect methods

That's indeed a good point, though I'm unsure about how much final classes are security relevant. But yeah, shifting my opinion a bit towards opt-out.

kitsunet:
I mean, I like the opt-in, but to fix the error above (with opt-out) we could just proxy false all our annotation classes. They definitely shouldn't need DI or AOP.

albe:
That would just be a symptomatic fix though. It would still break for any 3rd party annotation that isn't `Proxy(false)` with a very unhelpful exception.

kitsunet:
But we do not proxy anything in third party packages unless you enable it. Which then would make it impossible to enable it for a third party packagek so yes, probably the opt-in via proxy true should stay.

bwaidelich:
> we could just proxy false all our annotation classes

I started to prepare a follow-up that turns the whitelist into a blacklist (proxy all classes by default) but excludes classes marked `@Annotation` and that seems to work pretty nicely

bwaidelich:
After thinking about this further and working with it I'm even more convinced that we should just support final classes by default, everything else is just confusing (and documentation/logging doesn't really solve it IMO).
So I just went ahead and prepared a new version that turns the whitelist into a blacklist approach. If you guys don't agree we can always revert to 6ef8abf4c61c948ddc587ef88c95eedc06109283

BTW: I would even suggest to do the same for final *methods* in order to solve #649

kitsunet:
I like, works for me! Would merge soonish if no objections.

albe:
Alright, let's go the blacklist path then. The "documentation doesn't help confusion" convinced me ;)

kdambekalns:
Why don't you people use the review feature and do a "proper formal approval"?

TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/Compiler.php
```diff
@@ -141,6 +141,9 @@ public function getProxyClass($fullClassName)
         if (in_array(substr($fullClassName, 0, 14), $this->blacklistedSubPackages)) {
             return false;
         }
+        if ($classReflection->isFinal() && preg_match('/^\s?\*\s?\@Annotation\s/m', $classReflection->getDocComment()) === 1) {
```
kdambekalns: What is the purpose of this? I understand it skips classes that have `@Annotation` in their docblock, but why?

bwaidelich: See https://github.com/neos/flow-development-collection/pull/644#issuecomment-249910726
Without this, Flow complaints about proxied Annotation classes.
Maybe this could be solved otherwise but a `@Flow\Proxy(false)` annotation on each Annotation class is not a good option IMO

kitsunet: At least it feels like this rule is very specific and therefore dirty. I definitely prefer the proxy false solution then, also this might have impact on other peoples packages then...

bwaidelich: >  I definitely prefer the proxy false solution then, also this might have impact on other peoples packages then

But that was exactly my reasoning: If we suddenly require a `Proxy(false)` annotation that change might be breaking for 3rd parties (i.e. https://github.com/Flowpack/jobqueue-common/blob/master/Classes/Annotations/Defer.php).
A different approach might be to copy *all* of the classes doc block to the proxy class!?

bwaidelich: > Now is the chance for a breaking change

Mh, only because we _can_ break the API we should still try to avoid it by all means

> implementing such specific checks to avoid that is a hack IMHO

I agree that it is a bit "crude", especially because we can't use the `AnnotationParser` for that part. But I wouldn't consider it a hack. Especially if you look at the surrounding code of the Proxy building: There are lots of weird hard-coded specifics, like:

> Applying the docblock to the proxy could also have unintended results

We already do that, but we do it "selectively" for some reason: https://github.com/neos/flow-development-collection/blob/master/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyClass.php#L242-L256
We just skip `@Annotation` because it will *not* be returned by `ReflectionService::getClassAnnotations()`.
IMO it would be less of a hack to just copy the whole doc block

bwaidelich:
I added a short inline comment explaining why Annotation classes are skipped in `Proxy\Compiler:: getProxyClass()`. I'd like to merge this with the current votes within 24h if noone objects

PR #645, opened 2016-09-29
BUGFIX: Doctrine DataTypes should not be proxied

### LINKED ISSUE #516: Proxys for final methods should not be generated
Flow 3.3.3 generates a corrupt proxy object for `JsonArrayType`:

As `JsonArrayType` inherits from `Doctrine\DBAL\Types\JsonArrayType` which has a final constructor, this proxy class is invalid and causes the error: 

```
PHP Fatal error:  Cannot override final method Doctrine\DBAL\Types\Type::__construct() in /opt/tmp/Development/Cache/Code/Flow_Object_Classes/TYPO3_Flow_Persistence_Doctrine_DataTypes_JsonArrayType.php on line 235
Fatal error: Cannot override final method
Doctrine\DBAL\Types\Type::__construct() in
/opt/tmp/Development/Cache/Code/Flow_Object_Classes/TYPO3_Flow_Persistence_Doctrine_DataTypes_JsonArrayType.php
on line 235
```

```

/**
 * Extends the default doctrine JsonArrayType to work with entities.
 *
 * TOOD: If doctrine supports a Postgres 9.4 platform we could default to jsonb.
 */
class JsonArrayType extends JsonArrayType_Original implements \TYPO3\Flow\Object\Proxy\ProxyInterface {

    use \TYPO3\Flow\Object\Proxy\ObjectSerializationTrait;


    /**
     * Autogenerated Proxy Method
     */
    public function __construct()
    {
        if (method_exists(get_parent_class(), 'Flow_Aop_Proxy_buildMethodsAndAdvicesArray') && is_callable('parent::Flow_Aop_Proxy_buildMethodsAndAdvicesArray')) parent::Flow_Aop_Proxy_buildMethodsAndAdvicesArray();
        parent::__construct();
    }
```

albe:
Since 3.3 the doctrine types often ended up being proxied, due to a change in proxy generation
on aspected parent classes. This would throw an exception because the DBAL\Types base classes
constructor is private and hence cannot be called from the proxied constructor.

This works around that issue by making sure the custom DBAL Types are not proxied.

Fixes #516

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/DataTypes/JsonArrayType.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/DataTypes/JsonArrayType.php
@@ -2,7 +2,6 @@
 namespace TYPO3\Flow\Persistence\Doctrine\DataTypes;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\DBAL\Types;
 use Doctrine\DBAL\Types\JsonArrayType as DoctrineJsonArrayType;
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Core\Bootstrap;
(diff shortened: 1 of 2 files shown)


kitsunet:
Just a symptom fix but  lets have it for now.

PR #648, opened 2016-10-02
BUGFIX: Check for existing symlink in createDirectoryRecursively()

### LINKED ISSUE #647: TYPO3\Flow\Utility\Files::createDirectoryRecursively does not support symlinks
TYPO3\Flow\Utility\Files::createDirectoryRecursively is called during the NEOS bootstrap.
Composer dies with an exception if you symlink the Configuration directory.

Therefore createDirectoryRecursively should also support symlinks instead of only directories.

loeffert:
The `createDirectoryRecursively()` method now checks for an existing symlink before
trying to create a directory. Before it only checked for an existing directory, sometimes
leading to false negatives.

Fixes #647

### DIFF
--- a/Neos.Utility.Files/Classes/TYPO3/Flow/Utility/Files.php
+++ b/Neos.Utility.Files/Classes/TYPO3/Flow/Utility/Files.php
@@ -246,7 +246,7 @@ abstract class Files
         if (is_file($path)) {
             throw new Exception('Could not create directory "' . $path . '", because a file with that name exists!', 1349340620);
         }
-        if (!is_dir($path) && $path !== '') {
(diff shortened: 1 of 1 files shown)


kdambekalns:
Thanks for the PR! But let me point out what bothered me about it: the commit message was as bad as it can get. Well, almost, it wasn't _fix bug_, after all. :)

But really, we have a guideline for commit messages, and there is a more or less universally accepted view of a "good commit message". So, do everyone (including yourself) a favour and invest those 20 seconds to write a better commit message. It always helps when understanding how the code evolved…

I hope this doesn't come across as not being thankful!

albe:
Thx! But Karsten has a point there, investing a few more seconds in writing a proper commit message sure won't hurt the next time ;)

PR #649, opened 2016-10-03
BUGFIX: Do not proxy final methods in proxy classes

kitsunet:
To avoid problems with uninitialized AOP advices we need to proxy all
child classes of adviced classes as well. Due to that unplanned
classes can now be the target of a proxy, therefore Flow should
make sure to not proxy methods declared final on those classes.
As that might still result in non functioning AOP the skipped
class and method is logged for easier debugging.

Fixes: #516

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
@@ -540,8 +540,21 @@ class ProxyClassBuilder
         }
 
         $callBuildMethodsAndAdvicesArrayCode = "        if (method_exists(get_parent_class(), 'Flow_Aop_Proxy_buildMethodsAndAdvicesArray') && is_callable('parent::Flow_Aop_Proxy_buildMethodsAndAdvicesArray')) parent::Flow_Aop_Proxy_buildMethodsAndAdvicesArray();\n";
-        $proxyClass->getConstructor()->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
-        $proxyClass->getMethod('__wakeup')->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
+
+        $proxyMethods = ['__construct' => null, '__wakeup' => null];
+        if ($this->reflectionService->isMethodFinal($className, '__construct') === false) {
+            $proxyClass->getConstructor()->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
+            $proxyMethods['__construct'] = true;
+        }
+
+        if ($this->reflectionService->isMethodFinal($className, '__wakeup') === false) {
+            $proxyClass->getMethod('__wakeup')->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
+            $proxyMethods['__wakeup'] = true;
+        }
+
+        foreach (array_keys(array_filter($proxyMethods, 'is_null')) as $methodName) {
+            $this->systemLogger->log(sprintf('Could not proxy method "%s" in class "%s", as it is declared final, AOP might not be correctly applied.', $methodName, $className), LOG_DEBUG);
+        }
 
         return $treatedSubClasses;
     }


kitsunet:
This probably needs a test, but didn't have time now, wanted to find the cause and a fix first for now...

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -540,8 +540,21 @@ protected function addBuildMethodsAndAdvicesCodeToClass($className, ClassNameInd
...
+            $proxyClass->getConstructor()->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
+            $proxyMethods['__construct'] = true;
+        }
+
+        if ($this->reflectionService->isMethodFinal($className, '__wakeup') === false) {
+            $proxyClass->getMethod('__wakeup')->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
+            $proxyMethods['__wakeup'] = true;
+        }
+
+        foreach (array_keys(array_filter($proxyMethods, 'is_null')) as $methodName) {
```
albe: Can be simplified to `array_keys(array_filter($proxyMethods))` if `$proxyMethods` is initialized as `['__construct' => false, '__wakeup' => false]`.

kitsunet: Can't because I need the "other" value, not the true value. doing array_filter without anything would give me the methods that were proxied. I could name the variable negative, initialize to true and set to false if a proxy was created, but that felt like double negative. This is the most concise way to do this I could come up with.

albe: Ah yes, I see. Sorry for not fully following the logic before nitpicking!

bwaidelich:
Instead of silently* skipping final methods I'd prefer a solution that moves the `final` keyword to the proxied method similar to #644 if possible (I can investigate if you agree).

*logging doesn't count ;)

bwaidelich:
> I'd prefer a solution that moves the final keyword to the proxied method

Obviously that would be a solution for the next major release. The approach of this PR is super straight forward for 3.x.
But I wonder: I just tested this and still get a compilation error
```
Fatal error: Cannot override final method Some\Class_Original::__construct()...
```

A log entry
```
Could not proxy method "__construct" in class "Some\Class", as it is declared final, AOP might not be correctly applied.
```
is written (but in the middle of tons of "Build AOP Proxy/DI proxy" entries). Is that the expected behavior?

bwaidelich:
> I just tested this and still get a compilation error

Ah, the reason was a `@Flow\Inject` annotation in the class that leads the `constructor` to be proxied anyways.. I'll look into it after lunch.

ps. sorry for the monologue ;)

bwaidelich:
I gave it a shot and this seems to work quite nicely.
I don't know how to create a PR that depends on another one, but here are the required changes on top of #496:
https://github.com/bwaidelich/flow-development-collection/compare/496-better-support-for-final-classes...bwaidelich:649-better-support-for-final-methods

Again, this is just a suggestion for the upcoming major release. As a bugfix your solution seems to work nicely (except for classes that use DI). I would just suggest to simplify the code slightly for better readability (see inline comment)

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
kitsunet: Creates duplication of virtually the same log message and two else statements though.

kitsunet:
I would like to get this merge as the bug is really nasty...

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
albe: Hm, hard to decide which I find better. The above is indeed a bit easier to follow - OTOH the hard to follow stuff is only the (debug) logging in problematic cases.

Slightly in favor of not duplicating the error log atm, so I would go for keeping as is.

albe:
A test would really be nice, but I see how that is really really hard to properly test at all. So I'd be fine with merging for now to get rid of the bug.

TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php
bwaidelich: Wouldn't your implementation log two messages exactly the same way??
Anyways it can ofc be changed to sth like
```php
$skippedMethods = [];
if ($this->reflectionService->isMethodFinal($className, '__construct')) {
    $skippedMethods[] = '__construct';
} else {
}

if ($this->reflectionService->isMethodFinal($className, '__wakeup')) {
    $skippedMethods[] = '__wakeup';
} else {
}

if ($skippedMethods !== []) {
    $this->systemLogger->log(sprintf('Could not proxy the method(s) "%s" of class "%s", as it is declared final. AOP might not be correctly applied.', explode(', ', $skippedMethods), $className), LOG_DEBUG);
}
```
too.. Anyways, not a biggy of course.

bwaidelich:
I just tried to add a test but it seems the fix does not work!?
`ProxyClassBuilder::buildProxyClass()` has a similar code that is not fixed: https://github.com/neos/flow-development-collection/blob/3.3/TYPO3.Flow/Classes/TYPO3/Flow/Aop/Builder/ProxyClassBuilder.php#L468-L470

@kitsunet do I miss something?

albe:
Hmmm... maybe the check should go into the ProxyMethod class?

bwaidelich:
> maybe the check should go into the ProxyMethod class?

@albe good call, I'll give it a shot

bwaidelich:
Another PR that I just "quickly" wanted to add a test for.. ;)
Turns out, the fix only works if an AOP aspect targets a *non-existing* method of some class that has some parent with a final constructor. If the method exists somewhere in the hierarchy this still breaks..
Anyways, Alexanders suggestion to add this to the `ProxyMethod` greatly simplifies the fix:

```patch
From a5ca4837dbcb435f1482919d49ff122dbdb1785b Mon Sep 17 00:00:00 2001
From: Bastian <b.waidelich@wwwision.de>
Date: Mon, 10 Oct 2016 19:08:32 +0200
Subject: [PATCH] test

---
 TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyConstructor.php | 3 +++
 TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php      | 3 +++
 2 files changed, 6 insertions(+)

diff --git a/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyConstructor.php b/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyConstructor.php
index 41509e9..4289ac6 100644
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyConstructor.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyConstructor.php
@@ -41,6 +41,9 @@ class ProxyConstructor extends \TYPO3\Flow\Object\Proxy\ProxyMethod
      */
     public function render()
     {
+        if ($this->reflectionService->isMethodFinal($this->fullOriginalClassName, $this->methodName)) {
+            return '';
+        }
         $methodDocumentation = $this->buildMethodDocumentation($this->fullOriginalClassName, $this->methodName);
         $callParentMethodCode = $this->buildCallParentMethodCode($this->fullOriginalClassName, $this->methodName);
 
diff --git a/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php b/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php
index 583356c..41188ec 100644
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php
@@ -144,6 +144,9 @@ class ProxyMethod
      */
     public function render()
     {
+        if ($this->reflectionService->isMethodFinal($this->fullOriginalClassName, $this->methodName)) {
+            return '';
+        }
         $methodDocumentation = $this->buildMethodDocumentation($this->fullOriginalClassName, $this->methodName);
         $methodParametersCode = ($this->methodParametersCode !== '' ? $this->methodParametersCode : $this->buildMethodParametersCode($this->fullOriginalClassName, $this->methodName));
         $callParentMethodCode = $this->buildCallParentMethodCode($this->fullOriginalClassName, $this->methodName);
-- 
```

But I gave up trying to insert the `SystemLogger` in there (and it felt wrong anyways). Maybe you guys have an idea?

Anyways, we could ofc just use the original PR but we should add a test:

```patch
IFrom 6e3ab7b9e1418b9b1a9118db35d37d0d7922d9bc Mon Sep 17 00:00:00 2001
From: Bastian <b.waidelich@wwwision.de>
Date: Mon, 10 Oct 2016 19:12:34 +0200
Subject: [PATCH] test

---
 TYPO3.Flow/Tests/Functional/Aop/AopProxyTest.php   |  9 ++++++
 .../Fixtures/BaseFunctionalityTestingAspect.php    | 10 +++++++
 .../Fixtures/TargetClassWithFinalConstructor.php   | 32 ++++++++++++++++++++++
 .../TargetSubClassOfClassWithFinalConstructor.php  | 27 ++++++++++++++++++
 4 files changed, 78 insertions(+)
 create mode 100644 TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetClassWithFinalConstructor.php
 create mode 100644 TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetSubClassOfClassWithFinalConstructor.php

diff --git a/TYPO3.Flow/Tests/Functional/Aop/AopProxyTest.php b/TYPO3.Flow/Tests/Functional/Aop/AopProxyTest.php
index 1137457..8d6be9c 100644
--- a/TYPO3.Flow/Tests/Functional/Aop/AopProxyTest.php
+++ b/TYPO3.Flow/Tests/Functional/Aop/AopProxyTest.php
@@ -58,6 +58,15 @@ class AopProxyTest extends \TYPO3\Flow\Tests\FunctionalTestCase
     /**
      * @test
      */
+    public function subClassesWithFinalConstructorCannotBeAdvised()
+    {
+        $targetClass = new Fixtures\TargetSubClassOfClassWithFinalConstructor();
+        $this->assertSame('some value', $targetClass->someMethodOfSubClass());
+    }
+
+    /**
+     * @test
+     */
     public function canCallAdvicedParentMethodNotDeclaredInChild()
     {
         $targetClass = new Fixtures\ChildClassOfTargetClass01();
diff --git a/TYPO3.Flow/Tests/Functional/Aop/Fixtures/BaseFunctionalityTestingAspect.php b/TYPO3.Flow/Tests/Functional/Aop/Fixtures/BaseFunctionalityTestingAspect.php
index cdc8e87..6f2c78d 100644
--- a/TYPO3.Flow/Tests/Functional/Aop/Fixtures/BaseFunctionalityTestingAspect.php
+++ b/TYPO3.Flow/Tests/Functional/Aop/Fixtures/BaseFunctionalityTestingAspect.php
@@ -212,4 +212,14 @@ class BaseFunctionalityTestingAspect
     {
         return 'This is so NaN';
     }
+
+    /**
+     * @Flow\Around("method(TYPO3\Flow\Tests\Functional\Aop\Fixtures\TargetClassWithFinalConstructor->someMethod())")
+     * @param \TYPO3\Flow\Aop\JoinPointInterface $joinPoint
+     * @return string
+     */
+    public function classWithFinalConstructorAdvice(\TYPO3\Flow\Aop\JoinPointInterface $joinPoint)
+    {
+        return 'adviced';
+    }
 }
diff --git a/TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetClassWithFinalConstructor.php b/TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetClassWithFinalConstructor.php
new file mode 100644
index 0000000..9110640
--- /dev/null
+++ b/TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetClassWithFinalConstructor.php
@@ -0,0 +1,32 @@
+<?php
+namespace TYPO3\Flow\Tests\Functional\Aop\Fixtures;
+
+/*
+ *
+ *
+ * source code.
+ */
+
+
+/**
+ * A target class for testing proxied classes with final constructors
+ */
+class TargetClassWithFinalConstructor
+{
+
+    final public function __construct()
+    {
+    }
+
+    /**
+     * @return string
+     */
+    public function someMethod()
+    {
+        return 'some value';
+    }
+}
diff --git a/TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetSubClassOfClassWithFinalConstructor.php b/TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetSubClassOfClassWithFinalConstructor.php
new file mode 100644
index 0000000..77331c1
--- /dev/null
+++ b/TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetSubClassOfClassWithFinalConstructor.php
@@ -0,0 +1,27 @@
+<?php
+namespace TYPO3\Flow\Tests\Functional\Aop\Fixtures;
+
+/*
+ *
+ *
+ * source code.
+ */
+
+
+/**
+ * A target class for testing proxied subclasses of a class with final constructor
+ */
+class TargetSubClassOfClassWithFinalConstructor extends TargetClassWithFinalConstructor
+{
+    /**
+     * @return string
+     */
+    public function someMethodOfSubClass()
+    {
+        return 'some value';
+    }
+}
-- 
```

I hope this will be [obsolete with 4.0 anyways](https://github.com/bwaidelich/flow-development-collection/commit/10d17b921e2979d31ed6269735b15d51dce4007)

albe:
> But I gave up trying to insert the SystemLogger in there (and it felt wrong anyways). Maybe you guys have an idea?

Yeah, logging shouldn't happen in the ProxyMethod obviously. How about moving the check into the ProxyMethod constructor and throwing an exception there, then catching that and logging it in the builder? From the ProxyClass/Method POV trying to proxy a final method is an error condition, and just ignoring it (and logging the fact) is a caller-decision we make in the higher level.

robertlemke:
@bwaidelich @albe what's the state here, should I give it a try and review it or is there something else to be done first?

kitsunet:
@bwaidelich do you know the state? Can this be closed? I lost track 😆

bwaidelich:
Since #661 this can be considered resolved I guess

PR #651, opened 2016-10-04
TASK: Update standard view

grebaldi:
Hi there,

the fallback MVC view contained three invalid links and a couple of `TYPO3` labels and css class names, that are replaced with this PR. Not sure about the wording though, so please have a look if it's okay :)

(BTW: Just for completeness, I applied the same to the LockHoldingStackPage.html)


bwaidelich:
nice one, thanks!

albe:
Nice, thx! :)

PR #653, opened 2016-10-04
FEATURE: Support allowable tags in ``stripTags`` Eel String helper

gerhard-boden:
Now the ``stripTags`` string eel helper will accept a second optional argument in form of a list of allowed tags which will not be stripped from the string.


gerhard-boden:
Thanks for the reviews and the fast merge guys!

Issue #654, opened 2016-10-05
Object Management fails for classnames with underscores


bwaidelich:
When using class names that contain underscores e.g. when using the PSR-0 schema (see [example](https://github.com/google/google-api-php-client/)) Flow dies with a fatal error:

```
Required class "Some_Class" could not be loaded properly for reflection.
```

The reason for this broken behavior is that we use the underscore as namespace separator within cache identifiers/filenames (that won't allow the backslash).

See https://discuss.neos.io/t/error-trying-to-inject-google-client-via-objects-yaml-or-flow-inject/1468 for more details.

PR #655, opened 2016-10-05
BUGFIX: Don't use underscore as namespace separator

bwaidelich:
Work in progress

Fixes: #654


kitsunet:
Nasty! Thanks for starting out with this.

albe:
This sounds and looks pretty senseful - what's left to finish this up?

bwaidelich:
> what's left to finish this up?

Only tests IIRC, I'll look into it

bwaidelich:
FYI: It was _not_ only missing tests, but I observe some weird caching behavior with this patch applied – I'll look into it

bwaidelich:
OK, I replaced some more occurrences, adjusted tests accordingly and added a new one.
But this line currently breaks everything: https://github.com/neos/flow-development-collection/blob/3.3/TYPO3.Flow/Classes/TYPO3/Flow/Core/ClassLoader.php#L167
I don't understand why we split the class name by underscores to determine the path, this breaks loading of classes like `\Google_Client`. @kitsunet maybe you know how to fix that case, I failed to comprehend that namespacePart juggling in the `ClassLoader` ;)

kitsunet:
I would happily remove it, but:

http://www.php-fig.org/psr/psr-0/#mandatory

`Each _ character in the CLASS NAME is converted to a DIRECTORY_SEPARATOR...`

So according to PSR-0 this must be \Google\Client.

Maybe we can work around this. Have to have a look.

bwaidelich:
@kitsunet Ah, I remember now, thanks. And indeed, it is: https://github.com/google/google-api-php-client/blob/master/src/Google/Client.php so I should move the test fixture accordingly and it should work.. let's see

bwaidelich:
Didn't work as easy as I hoped: I renamed the fixture class to `SubNamespace_Underscores` and moved it to `SubNamespace/Underscores.php` but now I get:
```
Required class "TYPO3\Flow\Tests\Functional\Object\Fixtures\SubNamespace\Underscores" could not be loaded properly for reflection.
```
@kitsunet Would be great if you could also have a quick look sometimes next week, you're far more into that ;)

bwaidelich:
@kitsunet see discussion above. Maybe we can finally look into this again together? Would be nice to get this of my list ;)

bwaidelich:
Sorry for the long delay.. I had no personal itch with this one and it was too hot to merge it to an existing branch.. But IMO it should be fixed for 5.0+ so I created a new PR for master with: #1279

PR #656, opened 2016-10-07
TASK: Adjust documentation recommendations to ComponentChain

albe:
The documentation still recommended to create Requests through the `createFromEnvironment`
factory method, but that completely bypasses the HTTP components and hence leads to unexpected
results.
This change adjusts documentation and clearly states, that the factory method should not be used
to access the **current** request and response. Instead it is suggested to use the `RequestHandler`
for that purpose only.

### DIFF
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
@@ -252,8 +252,11 @@ Such a request is always bound to an ``Http\Request``::
 
 	...
 
-	$httpRequest = $this->bootstrap->getActiveRequestHandler()->getHttpRequest();
-	$actionRequest = new ActionRequest($httpRequest);
+  $requestHandler = $this->bootstrap->getActiveRequestHandler();
+  if ($requestHandler instanceof HttpRequestHandlerInterface) {
+    $actionRequest = new ActionRequest($requestHandler->getHttpRequest());
+    // ...
+  }
 
 Arguments
 ~~~~~~~~~
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
@@ -212,6 +212,16 @@ You can, in theory, create a new ``Request`` instance by simply using the ``new`
 arguments to the constructor. However, there are two static factory methods which make life much easier. We recommend
 using these instead of the low-level constructor method.
 
+.. warning::
+
+	You should only create a ``Request`` manually if you want to send out requests or if you know exactly what you are
+	doing. Creating a new ``Request`` will completely bypass all ``HTTP Components`` and might therefore lead to
+	unexpected results, like the trusted proxy headers ``X-Forwarded-*`` not being applied and the ``Request`` providing
+	wrong protocol, host or client IP address.
+	If you need access to the **current** HTTP ``Request`` or ``Response``, instead inject a ``HttpRequestHandlerInterface`` and
+	get the ``Request`` and ``Response`` from there. Alternatively, you can also get the active ``RequestHandler`` from the
+	``Bootstrap``.
+
 create()
 ~~~~~~~~
 
@@ -226,6 +236,8 @@ The second method, ``createFromEnvironment()``, take the environment provided by
 functions into account. It creates a ``Request`` instance which reflects the current HTTP request received from the web
 server. This method is best used if you need a ``Request`` object with all properties set according to the current
 server environment and incoming HTTP request.
+Note though, that you should not expect this ``Request`` to match the current ``Request``, since the latter will still
+have been affected by some ``HTTP Components``. If you need the **current** Request, get it from the ``RequestHandler`` instead.
 
 Creating an ActionRequest
 ~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -233,7 +245,15 @@ Creating an ActionRequest
 In order to dispatch a request to a controller, you need an ``ActionRequest``.
 Such a request is always bound to an ``Http\Request``::
 
-	$httpRequest = Request::createFromEnvironment();
+	/**
+	 * @var HttpRequestHandlerInterface
+	 * @Flow\Inject
+	 */
+	protected $requestHandler;
+
+	...
+
+	$httpRequest = $this->requestHandler->getHttpRequest();
 	$actionRequest = new ActionRequest($httpRequest);
 
 Arguments
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
@@ -215,12 +215,11 @@ using these instead of the low-level constructor method.
 .. warning::
 
 	You should only create a ``Request`` manually if you want to send out requests or if you know exactly what you are
-	doing. Creating a new ``Request`` will completely bypass all ``HTTP Components`` and might therefore lead to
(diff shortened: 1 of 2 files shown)


albe:
Should be updated and explicitly announced IMO, because 3.3 basically blew up the old recommended way of creating your own request instances when needed.

If it wouldn't be completely mad, I would even tag this as breaking to raise attention.

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
```diff
@@ -226,14 +236,24 @@ The second method, ``createFromEnvironment()``, take the environment provided by
...
 
 In order to dispatch a request to a controller, you need an ``ActionRequest``.
 Such a request is always bound to an ``Http\Request``::
 
-	$httpRequest = Request::createFromEnvironment();
+	/**
+	 * @var HttpRequestHandlerInterface
+	 * @Flow\Inject
+	 */
+	protected $requestHandler;
```
gerhard-boden: This leaves me with
`#1355480641: The object "TYPO3\Flow\Http\HttpRequestHandlerInterface" which was specified as a property in the object configuration of object "<path>\ConfirmationFinisher" (automatically registered class) does not exist` 

However
```
    /**
     * Bootstrap for retrieving the current HTTP request
     *
     * @Flow\Inject
     * @var Bootstrap
     */
    protected $bootstrap;

    $requestHandler = $this->bootstrap->getActiveRequestHandler()->getHttpRequest();
```
did work

albe: argh... HttpRequestHandlerInterface exists and is used all around, but isn't configured in `Objects.yaml`

albe:
Ok, so this actually also allows to inject `HttpRequestHandlerInterface` now.

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
gerhard-boden: Thanks for the hint! 
But couldn't we check for that with `if ($requestHandler instanceof HttpRequestHandlerInterface) {` I've seen other places where we do that.

About `HttpRequestHandlerInterface`:
Now I get ```Uncaught Exception in Flow
Argument 1 passed to TYPO3\Flow\Http\RequestHandler::__construct() must be an instance of TYPO3\Flow\Core\Bootstrap, none given, called in /var/www/kaprunerhof.at/kaprunerhof.getan.at/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Object/ObjectManager.php on line 530```

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
```diff
@@ -212,6 +212,16 @@ You can, in theory, create a new ``Request`` instance by simply using the ``new`
 arguments to the constructor. However, there are two static factory methods which make life much easier. We recommend
 using these instead of the low-level constructor method.
 
+.. warning::
+
+	You should only create a ``Request`` manually if you want to send out requests or if you know exactly what you are
+	doing. Creating a new ``Request`` will completely bypass all ``HTTP Components`` and might therefore lead to
```
bwaidelich: That sounds a bit dramatic, just *creating*  a `Request` won't have any side effects ;) Maybe we should put it a bit more neutral?

bwaidelich:
Thanks a lot for this. Apart from the little nitpick I think this could even be applied to lower branches..

For 4.0 I'd like to rework the HTTP implementation. I'll collect some ideas here #658

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
albe: argh2 :( I guess I trust the whole DI part too much... `RequestHandler` is `@Flow\Proxy(false)`, so no DI for that... guess the solution then really is to use the `Bootstrap`since the `RequestHandler` can currently not be injected

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
albe: Drama baby, drama! ;) Adjusted it a bit, if you have another suggestion for the wording, feel free to push something or post it here.

albe:
I targeted 3.3 only for the reason that the breaking nature of this only came alive in 3.3. So I'm not sure if it's not too confusing to change documentation of something in old versions where the problem doesn't occur out-of-the-box anyway. But feel free to retarget as you like, shouldn't be a problem.

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
```diff
@@ -226,14 +235,24 @@ The second method, ``createFromEnvironment()``, take the environment provided by
...
-	$httpRequest = Request::createFromEnvironment();
+	/**
+	 * @var Bootstrap
+	 * @Flow\Inject
+	 */
+	protected $bootstrap;
+
+	...
+
+	$httpRequest = $this->bootstrap->getActiveRequestHandler()->getHttpRequest();
```
bwaidelich: This will break if the active Request Handler is not an `HttpRequestHandlerInterface` that's why we do
```php
$requestHandler = $bootstrap->getActiveRequestHandler();
if ($requestHandler instanceof HttpRequestHandlerInterface) {
    $request = $requestHandler->getHttpRequest();
    // ...
}
```
mostly *würg ;)

albe:
Adjusted for the HttpRequestHandlerInterface... not nice, but that's the current state of affairs.

So if anyone feels like it, please merge.

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
```diff
@@ -226,15 +235,28 @@ The second method, ``createFromEnvironment()``, take the environment provided by
...
 Creating an ActionRequest
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In order to dispatch a request to a controller, you need an ``ActionRequest``.
 Such a request is always bound to an ``Http\Request``::
 
-	$httpRequest = Request::createFromEnvironment();
-	$actionRequest = new ActionRequest($httpRequest);
+	/**
+	 * @var Bootstrap
```
gerhard-boden: Maybe let's say `@var TYPO3\Flow\Core\Bootstrap` so this code works out of the box without an additional use statement?

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
```diff
@@ -226,15 +235,28 @@ The second method, ``createFromEnvironment()``, take the environment provided by
...
+	 * @Flow\Inject
+	 */
+	protected $bootstrap;
+
+	...
+
+  $requestHandler = $this->bootstrap->getActiveRequestHandler();
+  if ($requestHandler instanceof HttpRequestHandlerInterface) {
```
gerhard-boden: same thing here. 
`\TYPO3\Flow\Http\RequestHandler`

TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
bwaidelich: I'd suggest to in include the `use` statements instead as we encourage to import namespace for readabilitys sake in our CGL

bwaidelich:
Ah, wait, this is targetted for 3.3. I'll adjust – Damn namespaces ;)

PR #661, opened 2016-10-12
FEATURE: AOP for final methods

bwaidelich:
This adds support proxied `final` methods.
Previously those were always skipped from proxy building disallowing to advice them via AOP aspects.

Background:
Marking methods `final` is an important tool for framework code as it allows to define extension points
more explicitly, but until now we had to avoid the `final` keyword in order to support AOP.
With this change any `final` modifier is removed from the original class and moved to the proxy class so
that the behavior of the proxied class stays the same.

Related: #516


TYPO3.Flow/Tests/Functional/Aop/Fixtures/BaseFunctionalityTestingAspect.php
```diff
@@ -223,4 +223,16 @@ public function methodOfFinalClassAdvice(\TYPO3\Flow\Aop\JoinPointInterface $joi
...
+
+    /**
+     * @Flow\Around("method(TYPO3\Flow\Tests\Functional\Aop\Fixtures\TargetClass01->someFinalMethod())")
+     * @param \TYPO3\Flow\Aop\JoinPointInterface $joinPoint
+     * @return string
+     */
+    public function methodWithFinalModifierAdvice(\TYPO3\Flow\Aop\JoinPointInterface $joinPoint)
+    {
+        $originalValue = $joinPoint->getAdviceChain()->proceed($joinPoint);
+        return $originalValue . ' But, as said, nothing is final!';
```
albe: Well, there is the "Final countdown" ;)

bwaidelich:
As you can see, I finally try to take care of this one now

bwaidelich:
@mficzel good catch! 14f561e50fdbd665a499c3e3927f20fc03d9729f should take care of this

kitsunet:
Still unsure if that is _really_ a good idea

bwaidelich:
@kitsunet did you deliberately wait 6 month for the very moment this is merged to complain? ;)

What are your concerns?

kitsunet:
No, we talked about this already in the past 😆 I just saw this in my inbox on merge. I still think final is final is final. No AOP for you, but I can totally live with it, otherwise I would have blocked it back then. I just feel it opens a can of worms.

bwaidelich:
OK, got you. Thanks for clarifying.
I really don't think it's a contradiction: Marking a method final is (sometimes) a good idea as it can force a certain kind of architecture (extension points, ...). Still it can make perfect sense to cover these methods with policies or log calls to it for example.
With the same reason I opted to do this for final _classes_ (#644).

mea culpa if this turns out to open a can of worms – in that case I'll take care of reverting it with a corresponding migration

Issue #662, opened 2016-10-13
CompileStatic annotation is silently skipped for protected methods


bwaidelich:
When using the `@Flow\CompileStatic` annotation on a method that is not `public` and/or `static` this method won't be compiled. Instead it will just be skipped silently(!) and always executed in production, too.

This is for example the case for [ThumbnailGeneratorStrategy ::getThumbnailGeneratorClassNames ()](https://github.com/neos/neos-development-collection/blob/master/TYPO3.Media/Classes/TYPO3/Media/Domain/Strategy/ThumbnailGeneratorStrategy.php#L61).

bwaidelich:
Idea: throw exception during compile time for 3.x and fix support for `static protected` in 4.0

bwaidelich:
Fixed in master with #664 and we decided to leave it as is for older branches (see #663)

PR #663, opened 2016-10-13
BUGFIX: Reveal inaccessible CompileStatic methods

bwaidelich:
Previously when a protected or non-static method was annotated `@Flow\CompileStatic`
it was silently skipped from compilation.
With this change those cases will lead to an exception during compile time in
`Production` context.

Fixes: #662

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/DependencyInjection/ProxyClassBuilder.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/DependencyInjection/ProxyClassBuilder.php
@@ -647,6 +647,7 @@ class ProxyClassBuilder
      * @param string $className
      * @param \TYPO3\Flow\Object\Proxy\ProxyClass $proxyClass
      * @return void
+     * @throws \TYPO3\Flow\Object\Exception
      */
     protected function compileStaticMethods($className, $proxyClass)
     {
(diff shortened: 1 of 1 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Object/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -657,15 +658,24 @@ protected function compileStaticMethods($className, $proxyClass)
...
 
-        $methodNames = get_class_methods($className);
-        foreach ($methodNames as $methodName) {
-            if ($this->reflectionService->isMethodStatic($className, $methodName) && $this->reflectionService->isMethodAnnotatedWith($className, $methodName, \TYPO3\Flow\Annotations\CompileStatic::class)) {
-                $compiledMethod = $proxyClass->getMethod($methodName);
-
-                $value = call_user_func(array($className, $methodName), $this->objectManager);
-                $compiledResult = var_export($value, true);
-                $compiledMethod->setMethodBody('return ' . $compiledResult . ';');
+        $reflectionClass = new \ReflectionClass($className);
```
bwaidelich: `get_class_methods()` only returns `public` methods!

bwaidelich:
FYI: This is fixed in master with #664 and the Media package has been fixed for 2.1+ with neos/neos-development-collection#1182 – I wonder if we should just close this one and live with the limitation for Flow < 4.0 instead of risking a breaking change

kitsunet:
Yes please! Only master, it's not a serious issue...

PR #665, opened 2016-10-13
!!! TASK: Put properties of higher weighted requestFilters to the start of the merged viewConfiguration

mficzel:
Before this the higher weighted requestFilters were merged into the lower-weighted ones which placed the array-properties of the higher weighted filters last in the merged configuration. This made it impossible to add a new path templatePath that would be considered before.

This patch inverts the order of property merging and starts with the highest weighted configurations.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ViewConfigurationManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ViewConfigurationManager.php
@@ -83,12 +83,13 @@ class ViewConfigurationManager
             }
 
             usort($matchingConfigurations, function ($configuration1, $configuration2) {
-                return $configuration1['weight'] > $configuration2['weight'];
+                return $configuration1['weight'] < $configuration2['weight'];
             });
 
             $viewConfiguration = [];
             foreach ($matchingConfigurations as $key => $matchingConfiguration) {
-                $viewConfiguration = Arrays::arrayMergeRecursiveOverrule($viewConfiguration, $matchingConfiguration['configuration']);
+                $configuration = Arrays::arrayMergeRecursiveOverrule($matchingConfiguration['configuration'], $viewConfiguration);
+                $viewConfiguration = Arrays::arrayMergeRecursiveOverrule($viewConfiguration, $configuration);
             }
             $this->cache->set($cacheIdentifier, $viewConfiguration);
         }
--- a/TYPO3.Flow/Tests/Unit/Mvc/ViewConfigurationManagerTest.php
+++ b/TYPO3.Flow/Tests/Unit/Mvc/ViewConfigurationManagerTest.php
@@ -1,0 +1,222 @@
+<?php
+namespace TYPO3\Flow\Tests\Unit\Mvc;
+
+/*
+ * This file is part of the TYPO3.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use TYPO3\Flow\Configuration\ConfigurationManager;
+use TYPO3\Flow\Mvc\ViewConfigurationManager;
+use TYPO3\Flow\Mvc\ActionRequest;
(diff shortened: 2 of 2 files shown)


bwaidelich:
Thanks for this!
As discussed via [Slack](https://neos-project.slack.com/archives/flow-core-dev/p1476352744000434):
This would be a breaking change in behavior that is probably really hard to spot (btw: the commit message should mention why this is a breaking change).
We could maybe solve that with a core migration, but I'd still like to propose my suggested solution once again (just for the record):

## 1. Specify all array items for an explicit order

Currently if you just *add* an item to an array, the order is not really deterministic (it is technically, dependent on the package dependency order, but quite implicit and hard to tell). Also it can have side-effects if you use numeric keys.
So I think it would be perfectly OK to communicate the best practice that *all* items should be specified if you want to enforce a certain order.
For example:
I think it's fine (and even easier to comprehend) to write:
```yaml
-
  requestFilter: '<some-very-specific-filter>'
  options:
    templateRootPaths:
      'Your.Package': 'resource://Your.Package/Private/Templates/Media'
      'TYPO3.Neos': 'resource://TYPO3.Neos/Private/Templates/Media'
      'TYPO3.Media': 'resource://TYPO3.Media/Private/Templates'
```
if you want to add some custom templates for the media package but still want the fallbacks Neos > Media (in that order) to be retained.

## 2. Bug

Unfortunately exactly that does not work currently because `Arrays::arrayMergeRecursiveOverrule()` does not care about the key order.
This might be fixed in `arrayMergeRecursiveOverrule()` but that would be pretty dodgy (it could be done with yet another flag but..).
Another option could be to remove duplicates from the configuration before applying `arrayMergeRecursiveOverrule()`.


This would also be a new behavior but only in the very unlikely case that someone specified the packages in a *different* order than expected

mficzel:
@aertmann changing the behavior of arrayMergeRecursiveOverrule would be extremely breaky. It is used overall the core and i bet for we will find new bugs in unexpected places.

mficzel:
I see no way to do this unbreaky. I know some packages which had broken an thus not active  configurations that would start working by fixing this. 

What could be a solution would be to use the array sorter for this. But therefore the syntax of the pathes section would have to be changed.

bwaidelich:
> I see no way to do this unbreaky

Agreed, that's why I wrote "This would also be a new behavior but only in the very unlikely case that someone specified the packages in a different order than expected"

> What could be a solution would be to use the array sorter for this

The thing is: We just can't know what the correct behavior should be if the more specific configuration just sets *one* array item to an existing array.
In the case of TemplatePaths you probably want to add it to the *beginning* but there might be other options where you just want to append this.

One solution that would be fine with me: Don't merge array options at all.

bwaidelich:
BTW: We're on the same track that the current behavior is misleading and should be fixed. It's great that you started to work on this and I'm confident that the tests could be kept almost as-is anyways

mficzel:
Doing the pr now was the best way to save the learnings of todays debugging session. So i am not against changing this entirely ... just wanted to get this started. 

BTW ... what speaks against using array sorting:

```
templateRoots:
  'TYPO3.Neos': 
    path: 'resource://TYPO3.Neos/Private/Templates/Media'
    position: 'before TYPO3.Media'
  'TYPO3.Media':
    path: 'resource://TYPO3.Media/Private/Templates'
    position: '10' 
```

bwaidelich:
> what speaks against using array sorting

ah, that's a terrific idea!
That would mean we could keep the `ViewConfigurationManager` as  is and use the `PositionalArraySorter` in `TemplateView::getTemplateRootPaths()` (and the other multi-value getters).
We just have to make sure that the old syntax is still supported

mficzel:
@aertmann ... i tried the positional array sorting approach in https://github.com/neos/flow-development-collection/pull/666/files

kitsunet:
As written in slack I think we should have a proper discussion about the whole thing and the use cases and then do something. We can do it on discourse?

mficzel:
I started a discussion on discuss regarding this topic ... https://discuss.neos.io/t/templaterootpathes-overriding-in-views-yaml/1494

mficzel:
Close in favor of https://github.com/neos/flow-development-collection/pull/690

PR #666, opened 2016-10-14
FEATURE: Add positional array sorting to ``*RootPaths``  in Views.yaml

mficzel:
Add positional array sorting to `templateRootPaths`, `partialRootPaths` and `layoutRootPaths` in Views.yaml

The configurations of all patching requestFilters in Views.yaml are merged with the higher weighted filters applied last.  Since this process does not handle the order of the pathes the order in the resulting configuration can be a surprise. Especially keys added by higher weighted filters end up after the previously existing ones and thus are evaluated at last. This makes it hard to override a single templates by an external package.

This patch solves this by applying the positional array-sorter to those values while still supporting the old syntax. That way the order of the pathes can be controlled explicitly by the integrator.

Example:

```
-
  requestFilter: '...'
  options:
    templateRootPaths:
      'Vendor.Package':
        'position': 'before Vendor.OtherPackage'
        'value': 'resource://Vendor.Package/Private/Templates'
```

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/View/TemplateView.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/View/TemplateView.php
@@ -13,6 +13,7 @@ namespace TYPO3\Fluid\View;
 
 use TYPO3\Flow\Mvc\Controller\ControllerContext;
 use TYPO3\Flow\Utility\Files;
+use TYPO3\Flow\Utility\PositionalArraySorter;
 
 /**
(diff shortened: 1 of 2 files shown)


mficzel:
Please note: This is an alternate approach to https://github.com/neos/flow-development-collection/pull/665

mficzel:
I started a discussion on discuss regarding this topic ... https://discuss.neos.io/t/templaterootpathes-overriding-in-views-yaml/1494

mficzel:
https://github.com/neos/flow-development-collection/pull/690

PR #670, opened 2016-10-16
Duplicate error #1172215790 in Bootstrap.php fixed

RafaelKa:


kitsunet:
This branch is no longer maintained, the minimal version should be Flow 3.1 (which is released in parallel to Neos 2.1) according to our roadmap.

RafaelKa:
I'll close this PR without creating new one, because there is no duplicate since 3.0.

Issue #671, opened 2016-10-17
Object configuration presets


bwaidelich:
### Intro

I'd like to propose/discuss a feature for Flow 4.0+: The possibility to reuse the same object/singleton with different setups within one application.
### Scenarios
- Currently we need two different interfaces to achieve this feature for `SecurityLogger` vs `SystemLogger`
- For the CQRS/ES package it's important to be able to have logically separate `EventStore` instances per Bounded Context/Package possibly with different storage backends
- In general: Singletons with different setups
### "presets" Approach

My first approach was to extend the object configuration by some optional `presets` part that will _merge and overrule_ the basic object configuration:

``` yaml
TYPO3\Flow\Log\LoggerInterface:
  scope: 'singleton'
  factoryObjectName: 'TYPO3\Flow\Log\LoggerFactory'
  presets:
    'system':
      arguments:
        1:
          value: 'SystemLogger'
        # ...
    'security':
      arguments:
        1:
          value: 'Flow_Security'
        # ...
```

The corresponding injection would look sth like:

``` php
/**
 * @Flow\Inject(preset="system")
 * @var LoggerInterface
 **/
protected $systemLogger;
```
#### Advantages
- Central configuration around the interface in question
- Little duplication because presets inherit the main object configuration
#### Drawbacks/Challenges

This works fine with `prototypes` but for `singletons` we'd need to tweak the core of object management: `ObjectManagerInterface::get()` expects the object name and optionally some constructor arguments.
We could use a special format to refer to the `preset` here, like `<ObjectName>:<presetName>`. The generated proxy Code in `Flow_Proxy_injectProperties()` could then look something like

``` php
$this->Flow_Proxy_LazyPropertyInjection('TYPO3\Flow\Log\LoggerInterface:system', '', 'systemLogger', '6d57d95a1c3cd7528e3e6ea15012dac8', function() { return \TYPO3\Flow\Core\Bootstrap::$staticObjectManager->get('TYPO3\Flow\Log\LoggerInterface:system'); });
```

instead of the current

``` php
$this->Flow_Proxy_LazyPropertyInjection('TYPO3\Flow\Log\SystemLoggerInterface', '', 'systemLogger', '6d57d95a1c3cd7528e3e6ea15012dac8', function() { return \TYPO3\Flow\Core\Bootstrap::$staticObjectManager->get('TYPO3\Flow\Log\SystemLoggerInterface'); });
```

but it feels a bit "hacky"
### "objectName" Approach

[...]

PR #674, opened 2016-10-17
FEATURE: Object Configuration Presets

bwaidelich:
Work in progress

Resolves: #671

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
@@ -13,6 +13,7 @@ namespace TYPO3\Flow\Object;
 
 use TYPO3\Flow\Composer\ComposerUtility as ComposerUtility;
 use TYPO3\Flow\Object\Configuration\Configuration;
+use TYPO3\Flow\Object\Configuration\ConfigurationInterface;
 use TYPO3\Flow\Object\Configuration\ConfigurationProperty as Property;
 use Doctrine\ORM\Mapping as ORM;
 use TYPO3\Flow\Annotations as Flow;
@@ -339,36 +340,50 @@ class CompileTimeObjectManager extends ObjectManager
     protected function buildObjectsArray()
     {
         $objects = array();
+        /** @var Configuration $objectConfiguration */
         foreach ($this->objectConfigurations as $objectConfiguration) {
-            $objectName = $objectConfiguration->getObjectName();
-            $objects[$objectName] = array(
-                'l' => strtolower($objectName),
-                's' => $objectConfiguration->getScope(),
-                'p' => $objectConfiguration->getPackageKey()
-            );
-            if ($objectConfiguration->getClassName() !== $objectName) {
-                $objects[$objectName]['c'] = $objectConfiguration->getClassName();
-            }
-            if ($objectConfiguration->getFactoryObjectName() !== '') {
-                $objects[$objectName]['f'] = array(
-                    $objectConfiguration->getFactoryObjectName(),
-                    $objectConfiguration->getFactoryMethodName()
-                );
-
-                $objects[$objectName]['fa'] = array();
-                $factoryMethodArguments = $objectConfiguration->getArguments();
-                if (count($factoryMethodArguments) > 0) {
-                    foreach ($factoryMethodArguments as $index => $argument) {
-                        $objects[$objectName]['fa'][$index] = array(
-                            't' => $argument->getType(),
-                            'v' => $argument->getValue()
-                        );
-                    }
-                }
+            $objects[$objectConfiguration->getObjectName()] = $this->buildObjectArray($objectConfiguration);
+            foreach ($objectConfiguration->getPresets() as $presetConfiguration) {
+                $objects[$presetConfiguration->getObjectName()] = $this->buildObjectArray($presetConfiguration);
             }
         }
         $this->configurationCache->set('objects', $objects);
         return $objects;
+    }
+
+    /**
+     * @param ConfigurationInterface $objectConfiguration
+     * @return array
+     */
+    protected function buildObjectArray(ConfigurationInterface $objectConfiguration)
+    {
+        $objectName = $objectConfiguration->getObjectName();
+        $object = array(
+            'l' => strtolower($objectName),
+            's' => $objectConfiguration->getScope(),
+            'p' => $objectConfiguration->getPackageKey()
+        );
+        if ($objectConfiguration->getClassName() !== $objectName) {
+            $object['c'] = $objectConfiguration->getClassName();
+        }
+        if ($objectConfiguration->getFactoryObjectName() !== '') {
+            $object['f'] = array(
+                $objectConfiguration->getFactoryObjectName(),
+                $objectConfiguration->getFactoryMethodName()
+            );
+
+            $object['fa'] = array();
+            $factoryMethodArguments = $objectConfiguration->getArguments();
+            if (count($factoryMethodArguments) > 0) {
+                foreach ($factoryMethodArguments as $index => $argument) {
+                    $object['fa'][$index] = array(
+                        't' => $argument->getType(),
+                        'v' => $argument->getValue()
+                    );
+                }
+            }
+        }
+        return $object;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


bwaidelich:
This WIP supports the following new `Objects.yaml` syntax:

```yaml
TYPO3\Flow\Log\LoggerInterface:
  scope: singleton
  factoryObjectName: 'TYPO3\Flow\Log\LoggerFactory'
  presets:
    'system':
      arguments:
        1:
          value: 'SystemLogger'
        2:
          setting: TYPO3.Flow.log.systemLogger.logger
        3:
          setting: TYPO3.Flow.log.systemLogger.backend
        4:
          setting: TYPO3.Flow.log.systemLogger.backendOptions
    'security':
      arguments:
        1:
          value: 'Flow_Security'
        2:
          value: 'TYPO3\Flow\Log\Logger'
        3:
          setting: TYPO3.Flow.log.securityLogger.backend
        4:
          setting: TYPO3.Flow.log.securityLogger.backendOptions
```

Usage:
```php
/**
 * @Flow\Inject(preset="system")
 * @var LoggerInterface
 */
protected $systemLogger;

/**
 * @Flow\Inject(preset="security")
 * @var LoggerInterface
 */
protected $securityLogger;
```


It does **not** yet support nested object properties like:
```yaml
TYPO3\Flow\Security\Cryptography\HashService:
  properties:
    'cache':
      object:
        factoryObjectName: 'TYPO3\Flow\Cache\CacheManager'
        factoryMethodName: 'getCache'
        arguments:
          1:
            value: 'Flow_Security_Cryptography_HashService'
  presets:
    'somePreset':
      properties:
        'cache':
          object:
            arguments:
              1:
                value: 'Some_Other_Cache_Id'
```
and probably that won't really be possible

TYPO3.Flow/Classes/TYPO3/Flow/Object/CompileTimeObjectManager.php
```diff
@@ -339,36 +340,50 @@ function ($className) use ($filterExpression, $includeOrExclude) {
     protected function buildObjectsArray()
     {
         $objects = array();
+        /** @var Configuration $objectConfiguration */
         foreach ($this->objectConfigurations as $objectConfiguration) {
```
bwaidelich: The diff looks more complex than it is: I just extracted the loop as is to a new `buildObjectArray()` method so it can be called for the main object and it's presets

TYPO3.Flow/Classes/TYPO3/Flow/Object/Configuration/ConfigurationBuilder.php
```diff
@@ -191,36 +191,10 @@ protected function parseConfigurationArray($objectName, array $rawConfigurationO
...
-                                $property = $this->parsePropertyOfTypeObject($propertyName, $propertyValue['object'], $objectConfiguration);
-                            } elseif (array_key_exists('setting', $propertyValue)) {
-                                $property = new ConfigurationProperty($propertyName, array('type' => ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'path' => $propertyValue['setting']), ConfigurationProperty::PROPERTY_TYPES_CONFIGURATION);
-                            } else {
-                                throw new InvalidObjectConfigurationException('Invalid configuration syntax. Expecting "value", "object" or "setting" as value for property "' . $propertyName . '", instead found "' . (is_array($propertyValue) ? implode(', ', array_keys($propertyValue)) : $propertyValue) . '" (source: ' . $objectConfiguration->getConfigurationSourceHint() . ')', 1230563249);
-                            }
-                            $objectConfiguration->setProperty($property);
-                        }
-                    }
+                    $objectConfiguration->setProperties($this->parseProperties($optionValue, $objectConfiguration));
```
bwaidelich: extracted logic to method for less duplication

TYPO3.Flow/Classes/TYPO3/Flow/Object/Configuration/ConfigurationBuilder.php
```diff
@@ -191,36 +191,10 @@ protected function parseConfigurationArray($objectName, array $rawConfigurationO
...
-                                $argument = $this->parseArgumentOfTypeObject($argumentName, $argumentValue['object'], $configurationSourceHint);
-                            } elseif (array_key_exists('setting', $argumentValue)) {
-                                $argument = new ConfigurationArgument($argumentName, $argumentValue['setting'], ConfigurationArgument::ARGUMENT_TYPES_SETTING);
-                            } else {
-                                throw new InvalidObjectConfigurationException('Invalid configuration syntax. Expecting "value", "object" or "setting" as value for argument "' . $argumentName . '", instead found "' . (is_array($argumentValue) ? implode(', ', array_keys($argumentValue)) : $argumentValue) . '" (source: ' . $objectConfiguration->getConfigurationSourceHint() . ')', 1230563250);
-                            }
-                            $objectConfiguration->setArgument($argument);
-                        }
-                    }
+                    $objectConfiguration->setArguments($this->parseArguments($optionValue, $objectConfiguration));
```
bwaidelich: extracted logic to method for less duplication

kitsunet:
Cool! I mean, it adds some complexity to the already complex parts of the core, but for me the bigger issue here is the magic string in ObjectManager->get. I would avoid going down that path at all costs.

bwaidelich:
> but for me the bigger issue here is the magic string in ObjectManager->get

Yes, that's what I meant with 'it feels a bit "hacky"' in the issue description.. But IMO it is not *too* bad: `ObjectManagerInterface::get()` expects an **objectName** and we could argue that this is allowed to have the form `<ClassName>:<presetName>` if a preset with that name is configured for the given class.. And it would only be used by the framework anyways.

At least I prefer that to object names that are not related to a class name at all. And I guess I even prefer it to `ObjectManagerInterface::get($className, $presetName)` which is ambiguous with `ObjectManagerInterface::get($className, $argument)`.

kitsunet:
yes I guess what I am saying is that this needs serious thought and rework of our ObjectManager thinking IMHO. Adding some magic string way of getting a (service) implementation with a configured preset should not be the solution. To me it feels too hacky for something so deeply rooted into the framework. We will never get rid of this again if we add it.

bwaidelich:
Update: So the current WIP does not work with Singletons (and that's more or less the only usecase for it).
The reason is the way our proxy classes are built currently.

For the following `Objects.yaml`:
```yaml
'Wwwision\Test\EventStore':
  arguments:
    1:
      object:
        name: 'Wwwision\Test\FooEventStoreBackend'

'Wwwision\Test\EventStore:foo':
  className: 'Wwwision\Test\EventStore'
  arguments:
    1:
      object:
        name: 'Wwwision\Test\BarEventStoreBackend'
```

This constructor is built for the proxy class:

```php
public function __construct()
{
    $arguments = func_get_args();
    if (get_class($this) === 'Wwwision\Test\EventStore') \TYPO3\Flow\Core\Bootstrap::$staticObjectManager->setInstance('Wwwision\Test\EventStore', $this);
    if (get_class($this) === 'Wwwision\Test\EventStore') \TYPO3\Flow\Core\Bootstrap::$staticObjectManager->setInstance('Wwwision\Test\EventStore:foo', $this);

    if (!array_key_exists(0, $arguments)) $arguments[0] = new \Wwwision\Test\FooEventStoreBackend();
    if (!array_key_exists(0, $arguments)) throw new \TYPO3\Flow\Object\Exception\UnresolvedDependenciesException('Missing required constructor argument $backend in class ' . __CLASS__ . '. Please check your calling code and Dependency Injection configuration.', 1296143787);
    call_user_func_array('parent::__construct', $arguments);
}
```

So it sets the singleton instance in the `ObjectManager` for both object names to the same instance.

I tried to rework the proxy-builder so that it generates something like this instead for singletons:

```php
public function __construct($objectName)
{
    $this->Flow_objectName = $objectName;
    if (get_class($this) === 'Wwwision\Test\EventStore') \TYPO3\Flow\Core\Bootstrap::$staticObjectManager->setInstance($this->Flow_objectName, $this);

    if ($this->Flow_objectName === 'Wwwision\Test\EventStore') {
        $arguments = [
            new \Wwwision\Test\FooEventStoreBackend();
        ];
    } elsef ($this->Flow_objectName === 'Wwwision\Test\EventStore:foo') {
        $arguments = [
            new \Wwwision\Test\BarEventStoreBackend();
        ];
    }
    call_user_func_array('parent::__construct', $arguments);
}
```

But I got lost and also there needs to be yet another special case if the parent class is proxied too. Doh.


So FYI: I'll rest my case for the moment and use factories/managers.. But if anyone feels like taking this from here I'm happy to share my WIP & help

bwaidelich:
Closing in favor of #676 as discussed in #671

PR #675, opened 2016-10-18
TASK: Don’t list memcache as possible backend for Session handling

johannessteu:
Since memcache backend does not implement the IterableBackendInterface it can not be used for session handling


kitsunet:
LOL, I had hoped for an implemenation if iteratable, but this does the trick ;)

johannessteu:
Guess it's not possible to implement that >.>

kitsunet:
No clue ;)

PR #676, opened 2016-10-19
FEATURE: Virtual Object Configuration

bwaidelich:
Work in progress

Resolves: #671

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Annotations/Inject.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Annotations/Inject.php
@@ -31,6 +31,13 @@ final class Inject
     public $lazy = true;
 
     /**
+     * Optional object name
+     *
+     * @var string
+     */
+    public $name;
+
+    /**
      * Path of a setting (without the package key) which should be injected into the property.
      * Example: session.name
      *
@@ -58,6 +65,9 @@ final class Inject
     {
         if (isset($values['lazy'])) {
             $this->lazy = (boolean)$values['lazy'];
+        }
+        if (isset($values['name'])) {
+            $this->name = $values['name'];
         }
         if (isset($values['setting'])) {
             $this->setting = (string)$values['setting'];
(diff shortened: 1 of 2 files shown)


bwaidelich:
### Usage:

#### Objects.yaml:
```yaml
Flow:SystemLogger:
  className: 'TYPO3\Flow\Log\Logger'
  scope: singleton
  factoryObjectName: 'TYPO3\Flow\Log\LoggerFactory'
  arguments:
    1:
      value: 'SystemLogger'
    2:
      setting: TYPO3.Flow.log.systemLogger.logger
    3:
      setting: TYPO3.Flow.log.systemLogger.backend
    4:
      setting: TYPO3.Flow.log.systemLogger.backendOptions

Flow:SecurityLogger:
  className: 'TYPO3\Flow\Log\Logger'
  scope: singleton
  factoryObjectName: 'TYPO3\Flow\Log\LoggerFactory'
  arguments:
    1:
      value: 'Flow_Security'
    2:
      value: 'TYPO3\Flow\Log\Logger'
    3:
      setting: TYPO3.Flow.log.securityLogger.backend
    4:
      setting: TYPO3.Flow.log.securityLogger.backendOptions
```

#### PHP:
```php
/**
 * @Flow\Inject(name="Flow:SystemLogger")
 * @var LoggerInterface
 */
protected $systemLogger;

/**
 * @Flow\Inject(name="Flow:SecurityLogger")
 * @var LoggerInterface
 */
protected $securityLogger;
```

bwaidelich:
FYI: With the current implementation a "virtual" object is determined by a colon in the object name

albe:
Hm, can we somehow make sure that the `@var` annotation (which gives the class contract that is expected) is actually fulfilled by the injected virtual configuration? As in "Flow:SystemLogger instanceof LoggerInterface" check at compile time.

Otherwise the `@var` annotation will lose its meaning and unexpected failure could result at runtime.

bwaidelich:
> can we somehow make sure that the @var annotation [...] is actually fulfilled [...]

Good point.
In fact, currently the `className` *has* to be specified for virtual objects.

In theory it should also be possible to only specify a factory (i.e. that's possible with interfaces for example).
In that case we'd lose the connection between `@var` annotation and object type – like today

kitsunet:
if this check should happen, then please at compile time. there is nothing that could change the result at runtime, so to avoid speed loss in this sensible area, only compile time checks :)

bwaidelich:
> if this check should happen, then please at compile time

absolutely

bwaidelich:
Closing in favor of #674 as discussed in #671

bwaidelich:
Dang, closed the wrong one

jonnitto:
Ping @bwaidelich

bwaidelich:
This has been closed because it is stale by all means.. But I still have this on my radar and will re-open if I have enough personal itch to finish this one

PR #679, opened 2016-10-20
add exec() to the list of required system functions

tmotyl:
It's used e.g. in TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php


kitsunet:
Thanks for the PR, as we have a strategy of upmerging fixes this should go to the lowest supported branch, which would be 3.1 currently. Good pointer about this, because the section needs an overhaul anyway. `shell_exec` is for example not used from what I see.

dfeyer:
@tmotyl are your OK to retarget to a lower branch as @kitsunet suggest ? If you need help, just ask

tmotyl:
sorry, I will not have time to work on this in next days

kitsunet:
fixed via #856

Issue #681, opened 2016-10-21
HashService hashing strategy fallback is misleading


bwaidelich:
Due to some problems caused by older Flow installations that migrated from 1.0, a `fallback` mechanism for the password hashing strategies was implemented (see 4fe751f496f6c516790fc9c34a92ef1c7b0fedd7)

As a result the default strategy for `HashService::hashPassword()` is a different one than for `HashService::validatePassword()` unless specified explicitly because for the latter the configured *fallback* strategy would be used rather than the *default*.

PR #682, opened 2016-10-21
!!! TASK: Remove "fallback" password hashing strategy

bwaidelich:
This removes the fallback for password hashing strategies.

This is a breaking change for installations that had accounts created with a Flow version lower
than 1.1 (and whose passwords were never updated since then).
In that case make sure to add the prefix to the corresponding accounts in the accounts table.
For the default configuration the corresponding SQL query would be:

    UPDATE typo3_flow_security_account SET credentialssource = CONCAT('bcrypt=>', credentialssource)

Background:

Due to some problems caused by older Flow installations that migrated from 1.0, a fallback
mechanism for the password hashing strategies was implemented for password hashes that don't
contain the strategy prefix (i.e. "bcrypt=>").

As a result the default strategy for `HashService::hashPassword()` is a different one than for
`HashService::validatePassword()` unless specified explicitly because for the latter the configured
fallback strategy would be used rather than the default.

Fixes: #681


bwaidelich:
> Actually the inner if isset for default could be dropped because further down an exceptoin is thrown anywways...

Not really, it's a different exception and I even added a test that fails if I remove this condition

PR #686, opened 2016-10-28
BUGFIX: ObjectAccess should care for properties set to NULL

ComiR:
By using `array_key_exists()` instead of `isset()`, properties set to `null` will be detected as existing.

### DIFF
--- a/Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
+++ b/Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
@@ -67,7 +67,7 @@ abstract class ObjectAccess
         if (!is_object($subject) && !is_array($subject)) {
             throw new \InvalidArgumentException('$subject must be an object or array, ' . gettype($subject) . ' given.', 1237301367);
         }
-        if (!is_string($propertyName) && !is_integer($propertyName)) {
+        if (!is_string($propertyName) && !is_int($propertyName)) {
             throw new \InvalidArgumentException('Given property name/index is not of type string or integer.', 1231178303);
         }
 
@@ -98,12 +98,10 @@ abstract class ObjectAccess
             return null;
         }
         if (is_array($subject)) {
-            if (array_key_exists($propertyName, $subject)) {
-                $propertyExists = true;
-                return $subject[$propertyName];
-            }
-            return null;
-        } elseif (!is_object($subject)) {
+            $propertyExists = array_key_exists($propertyName, $subject);
+            return $propertyExists ? $subject[$propertyName] : null;
+        }
+        if (!is_object($subject)) {
             return null;
         }
 
@@ -115,20 +113,19 @@ abstract class ObjectAccess
                 $propertyReflection = new \ReflectionProperty($className, $propertyName);
                 $propertyReflection->setAccessible(true);
                 return $propertyReflection->getValue($subject);
-            } elseif (property_exists($subject, $propertyName)) {
-                return $subject->$propertyName;
-            } else {
-                throw new PropertyNotAccessibleException('The property "' . $propertyName . '" on the subject does not exist.', 1302855001);
             }
+            if (property_exists($subject, $propertyName)) {
+                return $subject->$propertyName;
+            }
+            throw new PropertyNotAccessibleException('The property "' . $propertyName . '" on the subject does not exist.', 1302855001);
         }
 
         if ($subject instanceof \stdClass) {
             if (array_key_exists($propertyName, get_object_vars($subject))) {
                 return $subject->$propertyName;
-            } else {
-                $propertyExists = false;
-                return null;
             }
+            $propertyExists = false;
+            return null;
         }
 
         $cacheIdentifier = $className . '|' . $propertyName;
@@ -137,14 +134,13 @@ abstract class ObjectAccess
         if (isset(self::$propertyGetterCache[$cacheIdentifier]['accessorMethod'])) {
             $method = self::$propertyGetterCache[$cacheIdentifier]['accessorMethod'];
             return $subject->$method();
-        } elseif (isset(self::$propertyGetterCache[$cacheIdentifier]['publicProperty'])) {
+        }
+        if (isset(self::$propertyGetterCache[$cacheIdentifier]['publicProperty'])) {
             return $subject->$propertyName;
         }
 
-        if (($subject instanceof \ArrayAccess) && !($subject instanceof \SplObjectStorage)) {
-            if (isset($subject[$propertyName])) {
-                return $subject[$propertyName];
-            }
+        if (($subject instanceof \ArrayAccess) && !($subject instanceof \SplObjectStorage) && $subject->offsetExists($propertyName)) {
+            return $subject->offsetGet($propertyName);
         }
 
         $propertyExists = false;
@@ -234,7 +230,7 @@ abstract class ObjectAccess
         if (!is_object($subject)) {
             throw new \InvalidArgumentException('subject must be an object or array, ' . gettype($subject) . ' given.', 1237301368);
         }
-        if (!is_string($propertyName) && !is_integer($propertyName)) {
+        if (!is_string($propertyName) && !is_int($propertyName)) {
             throw new \InvalidArgumentException('Given property name/index is not of type string or integer.', 1231178878);
         }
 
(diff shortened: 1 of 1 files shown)


Neos.Utility.ObjectHandling/Classes/TYPO3/Flow/Reflection/ObjectAccess.php
```diff
@@ -97,13 +97,10 @@ protected static function getPropertyInternal($subject, $propertyName, $forceDir
...
-        }
-        if (is_array($subject)) {
-            if (array_key_exists($propertyName, $subject)) {
-                $propertyExists = true;
-                return $subject[$propertyName];
-            }
-            return null;
+        } elseif (is_array($subject)) {
+            $propertySet = isset($subject[$propertyName]);
+            $propertyExists = $propertySet || array_key_exists($propertyName, $subject);
```
kdambekalns: Just `array_key_exists` should suffice, no need for the `isset()`, no? Also, `$propertyExists` is not used… ;)

aertmann:
Hey thanks for the fix. Is this only present in 3.2? otherwise it needs to be targeted at the lowest maintained branch.

kitsunet:
Mmm, knowing that ObjectAccess is used in so many places this behavior change is quite invasive. Is there a specific bug that needs this fix?

Neos.Utility.ObjectHandling/Classes/TYPO3/Flow/Reflection/ObjectAccess.php
ComiR: It would suffice, but `isset()` is much faster than `array_key_exists()`.

`$propertyExists` is passed by reference to the method ;)

ComiR:
@aertmann In 3.1 ObjectAccess wasn't in it's own subpackage ;)

@kitsunet invasive: true - wrong: also true.
I stumbled accross this when `node:repair` didn't delete undefined properties set to `null` (see https://github.com/neos/neos-development-collection/issues/1211).

kitsunet:
@ComiR true, but sometimes a known wrong behavior must be kept for b/c
I am not yet really confident bringing this into the LTS. The wrong behavior in Nodes can be fixed in there from all I see...

Neos.Utility.ObjectHandling/Classes/TYPO3/Flow/Reflection/ObjectAccess.php
kdambekalns: Fair enough. But check the next line… :)

kitsunet:
propertyExists is passed by reference, so even when not used in the method, it must be set correctly.

Neos.Utility.ObjectHandling/Classes/TYPO3/Flow/Reflection/ObjectAccess.php
ComiR: Now I see what you mean. There are two options.
Option 1 (the one I used):
```php
$propertySet = isset($subject[$propertyName]);
$propertyExists = $propertySet || array_key_exists($propertyName, $subject);
return $propertySet ? $subject[$propertyName] : null;
```
Option 2 (the one you'd prefer):
```php
$propertyExists = isset($subject[$propertyName]) || array_key_exists($propertyName, $subject);
return $propertyExists ? $subject[$propertyName] : null;
```
It really does have the same effect since `$propertyExists && !$propertySet => $subject[$propertyName] === null`.

ComiR: You're probably right. This is way too much micro-optimization. With PHP 5.6 `array_key_exists()` is about three times slower than `isset()` but with PHP 7 it's way closer ;)

kitsunet:
For me 👎 and master please, but I would really like to hear more opinions.

robertlemke:
If it changes the behaviour and could therefore potentially break existing applications then sure, it must be targeted to master.

But I'm also not sure about how such a real scenario would look like. If in doubt, I'd target it to master, but I'd also be fine with pushing it to 3.2.

bwaidelich:
Can you please come up with a commit message (or better: file an issue) that explains exactly what problem is solved by this?
Before we can decide upon this we should get a better understanding of the behavior and side effects.
Besides this should be covered by tests IMO

kdambekalns:
My vote: master only, fix known visible result of this in neos/neos-development-collection#1211.

albe:
So am I correct this is not considered strictly breaking (no !!! next-major-only) and hence could be merged to master after a rebase?

kdambekalns:
Which would be 4.1 at the time of writing. I'd say, yes, possible.

ComiR:
rebased on `master` and applied some minor changes:

> Some `if` where combined, unnecessary `else` clauses ommitted and `is_integer()` (is alias) replaced with `is_int()`.

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
```diff
@@ -287,14 +283,18 @@ public static function getGettablePropertyNames($object)
...
                 if (is_callable([$object, $methodName])) {
-                    if (substr($methodName, 0, 2) === 'is' && strlen($methodName) > 2) {
-                        $declaredPropertyNames[] = lcfirst(substr($methodName, 2));
-                    }
-                    if (substr($methodName, 0, 3) === 'get' && strlen($methodName) > 3) {
-                        $declaredPropertyNames[] = lcfirst(substr($methodName, 3));
-                    }
-                    if (substr($methodName, 0, 3) === 'has' && strlen($methodName) > 3) {
-                        $declaredPropertyNames[] = lcfirst(substr($methodName, 3));
+                    $methodNameLength = strlen($methodName);
```
albe: Not so sure about this restructuring. IMO this is now much harder to quickly grasp what this block is about

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
```diff
@@ -376,23 +374,15 @@ public static function isPropertyGettable($object, $propertyName)
...
         }
         $uppercasePropertyName = ucfirst($propertyName);
-        if (is_callable([$object, 'get' . $uppercasePropertyName])) {
-            return true;
-        }
-        if (is_callable([$object, 'is' . $uppercasePropertyName])) {
-            return true;
-        }
-        if (is_callable([$object, 'has' . $uppercasePropertyName])) {
+        if (is_callable([$object, 'get' . $uppercasePropertyName]) || is_callable([$object, 'is' . $uppercasePropertyName]) || is_callable([$object, 'has' . $uppercasePropertyName])) {
```
albe: Also not sure about this. I think I'd prefer the previous version.

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
kitsunet: nested if/elseifs, please no. This whole thing could get some nice refactoring touch, but I wouldn't want this. So agree to @albe

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
```diff
@@ -376,23 +374,15 @@ public static function isPropertyGettable($object, $propertyName)
         if (!is_object($object)) {
             throw new \InvalidArgumentException('$object must be an object, ' . gettype($object) . ' given.', 1259828921);
         }
-        if ($object instanceof \ArrayAccess && isset($object[$propertyName]) === true) {
-            return true;
-        } elseif ($object instanceof \stdClass && array_search($propertyName, array_keys(get_object_vars($object))) !== false) {
+        if (($object instanceof \ArrayAccess && $object->offsetExists($propertyName)) || ($object instanceof \stdClass && array_key_exists($propertyName, get_object_vars($object)))) {
```
kitsunet: Shouldn't the \stdClass case be already done in line 358 and so be superfluous here?

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
kitsunet: I am neutral on that. Both work for me. Could also be refactored into a nice condition function at some point.

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
ComiR: L358 is in another method ;)

ComiR:
rebased and reduced if-nesting.

albe:
Looks good from my side now apart from the two nitpicks on the `|| is_callable(...` chains, but not a blocker by any means.

PR #689, opened 2016-11-03
BUGFIX: Adapt the core migration manager for the new package manager

dfeyer:
Without this change PHP can print a warning message in the CLI output: PHP Warning:  Attempt to assign property of non-object in Manager.php on line 265.

This change fix access to currentPackageData['composerManifest']['extra'] to use the array
syntax and not the stdClass, used by the previous package manager.

### DIFF
--- a/TYPO3.Flow/Scripts/Migrations/Manager.php
+++ b/TYPO3.Flow/Scripts/Migrations/Manager.php
@@ -261,10 +261,10 @@ class Manager
         if ($appliedMigrationIdentifiers === array()) {
             return null;
         }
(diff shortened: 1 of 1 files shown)


kitsunet:
👍

bwaidelich:
Which bug does this fix?

dfeyer:
Not really a but, just an ugly warning
https://neos-project.slack.com/archives/flow-core-dev/p1478191870000666

bwaidelich:
> Not really a bug, just an ugly warning

Ah, I missed that one.. Could you add a quick note to the description above (i.e. to the merge commit message)? Then +1 and fine to go IMO

PR #690, opened 2016-11-04
!!!TASK: ViewConfiguration use only the settings of highest weighted request filter

mficzel:
Before this the higher weighted requestFilters were merged into the lower-weighted ones which placed the array-properties of the higher weighted filters last in the merged configuration. This made it impossible to add a new path templatePath that would be considered before.

This patch removes the merging of view-configurations entirely since this lead to confusion in the integration because the merging was unexpected.

This is breaking if you have multiple configurations with filters that apply to the same request and expect some option from one of the configurations to still be present despite another configuration having a higher weight.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ViewConfigurationManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ViewConfigurationManager.php
@@ -83,13 +83,15 @@ class ViewConfigurationManager
             }
 
             usort($matchingConfigurations, function ($configuration1, $configuration2) {
-                return $configuration1['weight'] > $configuration2['weight'];
+                return $configuration1['weight'] < $configuration2['weight'];
             });
-
-            $viewConfiguration = [];
-            foreach ($matchingConfigurations as $key => $matchingConfiguration) {
-                $viewConfiguration = Arrays::arrayMergeRecursiveOverrule($viewConfiguration, $matchingConfiguration['configuration']);
+            
+            if (count($matchingConfigurations) > 0) {
+                $viewConfiguration = array_shift($matchingConfigurations)['configuration'];
+            } else {
+                $viewConfiguration = [];
             }
+
             $this->cache->set($cacheIdentifier, $viewConfiguration);
         }
 
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ViewConfigurationManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ViewConfigurationManager.php
@@ -66,32 +66,25 @@ class ViewConfigurationManager
 
             $requestMatcher = new RequestMatcher($request);
             $context = new Context($requestMatcher);
-            $matchingConfigurations = [];
+
+            $viewConfiguration = [];
+            $highestWeight = -1;
             foreach ($configurations as $order => $configuration) {
                 $requestMatcher->resetWeight();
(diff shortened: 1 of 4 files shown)


mficzel:
This pr will replace https://github.com/neos/flow-development-collection/pull/665 and https://github.com/neos/flow-development-collection/pull/666 as discussed here https://discuss.neos.io/t/templaterootpathes-overriding-in-views-yaml/1494/1

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ViewConfigurationManager.php
```diff
@@ -83,13 +83,15 @@ public function getViewConfiguration(ActionRequest $request)
...
-            $viewConfiguration = [];
-            foreach ($matchingConfigurations as $key => $matchingConfiguration) {
-                $viewConfiguration = Arrays::arrayMergeRecursiveOverrule($viewConfiguration, $matchingConfiguration['configuration']);
+            
+            if (count($matchingConfigurations) > 0) {
+                $viewConfiguration = array_shift($matchingConfigurations)['configuration'];
+            } else {
+                $viewConfiguration = [];
             }
+
```
bwaidelich: Nitpick: Without the merging the whole block could be simplified IMO along those lines:
```php
$highestWeight = -1;
foreach ($configurations as $order => $configuration) {
    $requestMatcher->resetWeight();
    if (!isset($configuration['requestFilter']) && $order > $highestWeight) {
        $weight = $order;
    } else {
        $result = $this->eelEvaluator->evaluate($configuration['requestFilter'], $context);
        if ($result === false) {
            continue;
        }
        $weight = $requestMatcher->getWeight() + $order;
    }
    if ($weight > $highestWeight) {
        $viewConfiguration = $configuration;
        $highestWeight = $weight;
    }
}
```

mficzel:
@aertmann thanks for the nitpic ... that simplifies the code. The documentation does not mention the merging as far as i see so i think the new behavior is in line with the current documentation.

I still have some trouble with a functional test that fails ... will look into that the next days.

bwaidelich:
> The documentation does not mention the merging as far as i see

Right, that was a mistake. But we should at least add a *note* now that the behavior changed, for example:
```rst
.. note::

	Previously the configuration of all matching ``Views.yaml`` filters was merged.
        From version 4.0 only the matching filter with the highest weight is respected
        in order to reduce ambiguity.
```

at the end of this section: http://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/ModelViewController.html#configuring-views-through-views-yaml

mficzel:
@bwaidelich thanks, i added your note (btw. the pr is open to pushes from maintainers). I will have to find some time to understand the failing functional test case in the next days.

bwaidelich:
Thanks!

mficzel:
I am still not sure what the failing test actually tries to test.

bwaidelich:
> I am still not sure what the failing test actually tries to test.

@mficzel it tests this line https://github.com/neos/flow-development-collection/blob/master/TYPO3.Flow/Configuration/Testing/Views.yaml#L14-L16 IIRC

mficzel:
Ahh i finally found it. The following test-configurations were more specific due to the ``parentRequest`` matcher. I now added ``mainRequest`` to be even more specific. 

BTW: That is the only place i found where the core actually used the merging.

bwaidelich:
Wohoo, looks great. Thanks for taking care.
I can't test atm, but IMO this is good to go

PR #691, opened 2016-11-05
TASK: Remove unnecessary ``IncludeConfigurations.php`` file

kitsunet:


kitsunet:
Failing test is PHP5.5 because I used a feature in PHPUnit that is not available there. I can try to fix this ASAP or we first upgrade our requirements for the next version as this is master anyway and will be PHP7 only.

PR #692, opened 2016-11-05
FEATURE: Convert objects with multiple identity properties

kitsunet:


TYPO3.Flow/Classes/TYPO3/Flow/Property/TypeConverter/PersistentObjectConverter.php
```diff
@@ -273,7 +273,18 @@ protected function handleArrayData(array $source, $targetType, array &$converted
...
+        $identityPropertyCount = count($identityProperties);
+        if ($identityPropertyCount === 1 && (is_string($identity) || is_integer($identity))) {
+            ObjectAccess::setProperty($object, reset($identityProperties), $identity, true);
+        }
+
+        if (is_array($identity) && count($identity) === $identityPropertyCount) {
+            foreach ($identityProperties as $propertyName) {
+                ObjectAccess::setProperty($object, $propertyName, $identity[$propertyName], true);
+            }
+        }
```
aertmann: would it make sense to have some more checking here? this would fail fairly silently if the provided identity stuff didn't match

also should have some tests

albe:
Good move!

robertlemke:
Lacks description / commit message body

PR #693, opened 2016-11-07
TASK: Require PHP 7.0 as minimal version

kitsunet:


robertlemke:
I guess we can't test this without merging it, right?

kitsunet:
well, I mean, the travis result is already with PHP 7.0 only so technically it should be fine. But finally we will only see after merging..

PR #694, opened 2016-11-07
FEATURE: ``Array.flip()`` Eel helper

jonnitto:
With this helper it is possible to flip the keys and values from an array.

`Array.flip(array)`

### DIFF
--- a/TYPO3.Eel/Classes/TYPO3/Eel/Helper/ArrayHelper.php
+++ b/TYPO3.Eel/Classes/TYPO3/Eel/Helper/ArrayHelper.php
@@ -340,6 +340,7 @@ class ArrayHelper implements ProtectedContextAwareInterface
      * If a value has several occurrences, the latest key will be used as its value, and all others will be lost.
      *
      * @param array $array
+     * @return array The array with flipped keys and values
      */
     public function flip(array $array)
     {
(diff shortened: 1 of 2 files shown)


TYPO3.Eel/Classes/TYPO3/Eel/Helper/ArrayHelper.php
```diff
@@ -334,6 +334,20 @@ public function splice(array $array, $offset, $length = 1, $replacements = null)
     }
 
     /**
+     *  Exchanges all keys with their associated values in an array
+     *
+     * Note that the values of array need to be valid keys, i.e. they need to be either integer or string.
+     * If a value has several occurrences, the latest key will be used as its value, and all others will be lost.
+     *
+     * @param array $array
```
jonnitto: I've added the annotation , thanks for the review

TYPO3.Eel/Classes/TYPO3/Eel/Helper/ArrayHelper.php
```diff
@@ -334,6 +334,21 @@ public function splice(array $array, $offset, $length = 1, $replacements = null)
...
+     * @return array The array with flipped keys and values
+     */
+    public function flip(array $array)
+    {
+        $newArray = array_flip($array);
+        return $newArray;
```
aertmann: not really important, but would personally just `return array_flip($array);`

PR #701, opened 2016-11-10
TASK: Add support to cache PHP in APC, Memcached, PDO and Redis

dfeyer:
This change adds support for caching PHP code in more cache backends.

With this change you can cache Eel expressions and Fluid templates to any supported backend. Flow proxy class can not use this new feature. If you use in-memory caches for your dynamic PHP code, you need to understand that opcache is disabled with those cache backends!

To use this feature you need to enable `allow_url_include = On` in your `php.ini`.


dfeyer:
Don't merge this one before https://github.com/neos/flow-development-collection/pull/231

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/RequireOnceFromValueTrait.php
```diff
@@ -0,0 +1,46 @@
...
+     * @api
+     */
+    public function requireOnce($entryIdentifier)
+    {
+        $value = trim($this->get($entryIdentifier));
+        if ($value === '') {
+            return false;
+        }
+        $hash = md5($value);
+        if (isset($this->_requiredEntryIdentifiers[$hash])) {
```
dfeyer: Maybe here we need to cache the result of the include and send it again. This runtime cache is required because include_once can include multiple time the same code, as we are not reading a file.

Maybe we should move to a simple include also

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/RequireOnceFromValueTrait.php
```diff
@@ -0,0 +1,46 @@
...
+            return false;
+        }
+        $this->_requiredEntryIdentifiers[$hash] = true;
+        return include_once('data:text/plain;base64,' . base64_encode($value));
```
dfeyer: Here two option, base64_encode or urlencode ... not benchmarked, but maybe urlencode is the fastest.

kdambekalns:
/cc @robertlemke 

@dfeyer did you do this primarily for performance reasons? or to avoid having to use a writable filesystem?

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/RequireOnceFromValueTrait.php
kdambekalns: If you ensure the "once" with this, I guess it should be `include` afterwards. Otherwise there is some overhead from PHP (trying to) check the file for being included already… no?

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/RequireOnceFromValueTrait.php
```diff
@@ -0,0 +1,46 @@
...
+     * @return mixed Potential return value from the include operation
+     * @api
+     */
+    public function requireOnce($entryIdentifier)
+    {
+        $value = trim($this->get($entryIdentifier));
+        if ($value === '') {
+            return false;
+        }
+        $hash = md5($value);
```
kdambekalns: Why can't the `$entryIdentifier` be used directly? It has to be unique anyway…

kitsunet: 👍

TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/RequireOnceFromValueTrait.php
kitsunet: I would go with base64.

dfeyer:
> did you do this primarily for performance reasons? or to avoid having to use a writable filesystem?

@kdambekalns For educational purpose ;) seriously to limit the amount of I/O (Filesystem) as much as possible first to try to improve the performance in Docker container for developement. For the performance in production, it need some benchmark. Less IO but no opcache ...

robertlemke:
Hey. My concern would be that people will now try to use these backends for caching PHP code like proxy classes etc, which doesn't work. How are these guys protected? Is there a meaningful error message if they try to do that?

But overall, I'm not yet really convinced by the use case. Opcache will usually be much faster (depending on your RAM of course).

IMO we should have some actually measured performance benefits from this in order to consider it for core.

radmiraal:
@robertlemke can we turn your statement around maybe? Have some pretty decent documented alternatives and explanation why only 1 php capable backend should be allowed? Like for example: can you be sure there's ALWAYS opcache available (and then please arguments from outside systems you have root access to).

I can imagine the change as I've have been looking for cache optimization for fluid caches and stuff to limit I/O on docker, while doing so I noticed there's only 1 php capable backend available. Just from that (and the pretty theoretical documentation) I can see more incoming pr's if this one wouldn't make it.

dfeyer:
> How are these guys protected?

@robertlemke They are protected, the system will just not work ;) we can document or even better add an inline comment in the Caches.yaml that this one require a file based cache, or we can even throw a proper exception during exception by testing the cache implementation for the proxy class. So no blocker here, just give us your preference.

> But overall, I'm not yet really convinced by the use case. Opcache will usually be much faster (depending on your RAM of course).

It's an option people are free to use it or not. In some use case you can be I/O bound and not CPU bound, think about small cloud instance with poor disk access (or only network FS, like EBS for AWS), hosting with readonly filesystem, ...

A benchmark will cost tons of time and solve none of your points. Benchmark suck at showcase real situation, especially here where tons of factors can impact the resolution (IO, CPU, opcache available or not, number of cache segment). 

Like @radmiraal say, currently with only one PhpCapable backend, there is no way to try to improve performance in some situation. With this change it's possible to test for your specific use case if switching to an other cache make sense.

kdambekalns:
I had my doubts about the use case, but for "dynamic PHP code" this could indeed be an option, @robertlemke. 

If I may request an enhancement, @dfeyer – amend the caching backend documentation to point out the specifics of using those for PHP code. Then we should be good to go. Agreed?

kitsunet:
👍 I really think we should have this in

mficzel:
@kdambekalns i added the following warning:

This backend is php-capable. Nevertheless it cannot be used to store the proxy-classes from the ``FLOW_Object_Classes`` Cache. It can be used for other code-caches like ``Fluid_TemplateCache``, ``Eel_Expression_Code`` or ``Flow_Aop_RuntimeExpressions``. This can be usefull in certain situations to avoid file operations on production environments. If you want to use this backend for code-caching make sure that ``allow_url_include`` is enabled in php.ini

PR #705, opened 2016-11-15
FEATURE: Add more optional arguments to currency viewhelper

gerhard-boden:
Adds ``prependCurrency``, ``separateCurrency`` and ``decimals`` as optional arguments to the ``CurrencyViewHelper``.

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Format/CurrencyViewHelper.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Format/CurrencyViewHelper.php
@@ -114,14 +114,21 @@ class CurrencyViewHelper extends AbstractLocaleAwareViewHelper
             } catch (I18nException $exception) {
                 throw new ViewHelperException($exception->getMessage(), 1382350428, $exception);
             }
-        } else {
-            $output = number_format((float)$stringToFormat, $decimals, $decimalSeparator, $thousandsSeparator);
-            if ($currencySign !== '' && $prependCurrency === false) {
-                $output .= ($separateCurrency === true ?  ' ' : '') . $currencySign;
-            } elseif ($currencySign !== '' && $prependCurrency === true) {
-                $output = $currencySign . ($separateCurrency === true ?  ' ' : '') . $output;
-            }
+
+            return $output;
         }
+
+        $output = number_format((float)$stringToFormat, $decimals, $decimalSeparator, $thousandsSeparator);
+        if (empty($currencySign)) {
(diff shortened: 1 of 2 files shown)


gerhard-boden:
For a feature this should actually target master, but I was thinking if we could make an exception in this case. 

The reasons:

- In the current state this ViewHelper is incomplete and the newly added arguments are dearly missed in the LTS version when working a lot with prices in different layouts.

- The new arguments are completely optional and fall back to the current behavior.

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Format/CurrencyViewHelper.php
```diff
@@ -75,13 +91,16 @@ class CurrencyViewHelper extends AbstractLocaleAwareViewHelper
      * @param string $currencySign (optional) The currency sign, eg $ or €.
      * @param string $decimalSeparator (optional) The separator for the decimal point.
      * @param string $thousandsSeparator (optional) The thousands separator.
+     * @param boolean $prependCurrency (optional) Indicates if currency symbol should be placed before or after the numeric value.
+     * @param boolean $separateCurrency (optional) Indicates if a space character should be placed between the number and the currency sign.
```
albe: `$currencySeparator` would probably be more in line with the other params for separation characters

gerhard-boden:
@albe Thanks a lot for your review!
I added a note (which will also be shown in the documentation), with a little more information about the ``locale`` behaviour and some more information that came up today in slack.

albe:
Thanks for the documentation update, that helps :)

gerhard-boden:
Tests are green now after the travis fixes and rebasing.

TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Format/CurrencyViewHelper.php
mficzel: Since decimal- and thousandsSeperator are also given explicitly why not also define currencySeperator as string with default ' '.

gerhard-boden: I thought about that, but I couldn't really see many use cases for that. Plus currently the ViewHelper is in synch with the Fluid ViewHelper:
https://fluidtypo3.org/viewhelpers/fluid/master/Format/CurrencyViewHelper.html

(not that it should be a priority, but it comes up when you google for `fluid currency viewhelper`):

PR #716, opened 2016-11-21
!!!TASK: Adjust Package API

kitsunet:
Adjusts the Package API and classes to the pending breaking changes to interfaces and removes
deprecated classes.


mficzel:
Here is a pr for the neos side of this change https://github.com/neos/neos-development-collection/pull/1234

mficzel:
@kitsunet the unit tests are working now and also the demo site ... the remaining problem is that flowpack/behat also needs an autload-dev section ... i think that causes the failing of the behat tests.

mficzel:
The adjustments for loading the test classes are extracted to this pr https://github.com/neos/flow-development-collection/pull/728

PR #717, opened 2016-11-23
!!!FEATURE: Runtime evaluation of env and constants in Configuration

kitsunet:
The configuration is now cached with php expressions that read
environment variables and constants at runtime to allow writing
the configuration cache on a different environment.

In the process the setting to configure the temporary path which was
``TYPO3.Flow.core.temporaryDirectoryBase`` got finally removed as it
never fully worked as intended. You can instead set the environment
variable ``FLOW_PATH_TEMPORARY_BASE`` if you need a different
temporary path than the default one.


kitsunet:
Will iterate a bit over this to streamline it more. It works though and unit tests could be adapted to new code structure.

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
```diff
@@ -125,11 +125,11 @@ class ConfigurationManager
      * @var array
      */
     protected $configurationTypes = [
+        self::CONFIGURATION_TYPE_SETTINGS => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_SETTINGS, 'allowSplitSource' => false],
```
robertlemke: Does changing the order of types have an impact of the compiled configuration in general? Or did you change the order for other reasons?

TYPO3.Flow/Classes/TYPO3/Flow/Configuration/ConfigurationManager.php
```diff
@@ -389,16 +387,18 @@ public function shutdown()
     }
 
     /**
-     * Warms up the complete configuration cache, i.e. fetching every configured configuration type
-     * in order to be able to store it into the cache, if configured to do so.
-     *
-     * @see \TYPO3\Flow\Configuration\ConfigurationManager::shutdown
-     * @return void
+     * @param array $packages
```
robertlemke: A little doc comment would have been nice, but well …

kitsunet:
Closing in favor of #741

PR #718, opened 2016-11-23
TASK: Require Fluid Adaptor in Flow

kitsunet:
Despite the fact that Flow technically doesn't depend
on Fluid anymore we require it as the majority of projects
including Neos uses it anyway. Otherwise developers would
need to require it themselves in Flow projects.


kitsunet:
@robertlemke

PR #720, opened 2016-11-23
TASK: Add `useMainRequest` to the ActionViewHelpers

mficzel:
With this change it is possible to use the main Request instead of the current Request.
Note: using this argument can be a sign of undesired tight coupling, use with care


mficzel:
I tested this and it looks good and works as expected.

PR #722, opened 2016-11-23
BUGFIX: Use correct namespaces in ViewHelperBaseTestCase

nezaniel:
It seems like that during the switch to the Fluid Adaptor, some namespaces in test were not properly adapted


kdambekalns:
Unfortunately broke a lot of tests that extend this…

PR #724, opened 2016-11-23
BUGFIX: Fix regression introduced by test namespace change

kdambekalns:
In neos/flow-development-collection#722 the namespace was changed but
the classes extending it were not adjusted. Cough, cough.

This fixes that.


kdambekalns:
And I are say @johannessteu and me are involved as well…

PR #727, opened 2016-11-24
TASK: Remove deprecated ``TagBuilder`` ViewHelper

johannessteu:


gerhard-boden:
Fixes the issue, but wasn't this marked as `@depricated` for the upcoming release?

kdambekalns:
I think it can be removed, since an adjustment is needed anyway (the old namespace is gone…). But feedback from @kitsunet would be appreciated.

johannessteu:
reviewed together with @skurfuerst, fine as is.

PR #728, opened 2016-11-24
!!! TASK: Remove special handling of test-classes from classLoader

mficzel:
Instead of the previous logic the autoload-dev feature of composer is used now.

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -84,6 +84,7 @@ class Scripts
         $bootstrap->setEarlyInstance(ClassLoader::class, $classLoader);
         if ($bootstrap->getContext()->isTesting()) {
             self::requireAutoloaderForPhpUnit();
+            $classLoader->setConsiderTestsNamespace(true);
             require_once(FLOW_PATH_FLOW . 'Tests/BaseTestCase.php');
             require_once(FLOW_PATH_FLOW . 'Tests/FunctionalTestCase.php');
         }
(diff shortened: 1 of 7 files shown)


mficzel:
We need a good strategy how to get this merged together with https://github.com/neos/neos-development-collection/pull/1250

robertlemke:
@mficzel I didn't start reviewing this yet, but have a question: you marked this change as breaking (!!!) - what do users have to adjust after this has been merged / what makes it a breaking change? I think you should explain this in the PR description.

mficzel:
@robertlemke i had serious problems getting the tests to run with this change and eventually gave up. Maybe someone with intimate knowledge of composer can get this to work.

This pr is extracted from an bigger and more important one that was merged on the november sprint.

kitsunet:
I will have a look at it after we got this release out, it's too late now anyways I think.

robertlemke:
@kitsunet agreed. "Serious problems to run tests" doesn't sound like a last-minute change for a release ;-)

mficzel:
@robert @kitsunet the problems with the autoload-dev feature of composer are those:

1. To be able to extend a testCase from another package as we do in Neos and Flow those BaseTestCases have to be in Classes not in Tests because the autoload-dev of depending packages is not included
2. If this is done accepted this would add a production dependency to phpunit to the BasePackage wich seems not acceptable to me.

So the question is how can we write baseTestClasses that are included from other Packages. As soon as we have a solution for that this should work.

kitsunet:
I think we can scrap this as it won't really work for the reasons you pointed out. Maybe we can think about some nice solution at some point, but for now this isn't it, right?

mficzel:
@kitsunet true i will close this now

PR #729, opened 2016-11-24
TASK: Use string for "virtual" namespace name

kdambekalns:
This reverts the use of the `::class` constant for the ORM proxy
namespace configuration, since that is not "real" and is not
even a "class" but just a "namespace prefix".


kdambekalns:
@daniellienert

PR #734, opened 2016-11-25
TASK: Add GitHub templates and contribution guidelines

kdambekalns:


kdambekalns:
Because if we have it for Neos, we should have it for Flow…

PR #738, opened 2016-11-25
TASK: Add Doctrine migrations to rename role identifiers

kdambekalns:
See neos/neos-development-collection#1240


kdambekalns:
@skurfuerst @kitsunet @robertlemke

PR #739, opened 2016-11-25
BUGFIX: Adjust RoleTest to renamed vendor

kdambekalns:
See neos/neos-development-collection#1240


kdambekalns:
@kitsunet @skurfuerst @robertlemke

PR #741, opened 2016-11-25
FEATURE: Runtime evaluation of env and constants in Configuration

kitsunet:
The configuration is now cached with php expressions that read
environment variables and constants at runtime to allow writing
the configuration cache on a different environment.


kitsunet:
Lightweight implementation of #717 without breaking.

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -679,6 +689,51 @@ protected function replaceVariablesInValue($value)
...
+     * @param string $phpString
+     * @return mixed
+     */
+    protected function replaceVariablesInPhpString($phpString)
+    {
+        $phpString = preg_replace_callback('/
+            (?<startString>\s\')?      # optionally starting a string
+            (?P<fullMatch>%            # an expression is indicated by %
+            (?P<expression>
+            (?:(?:\\\?[\d\w_\\\]+\:\:)    # either a class name followed by ::
```
skurfuerst: Why do we match for class names here? don't understand that yet :)

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -679,6 +689,51 @@ protected function replaceVariablesInValue($value)
...
+     * This is applied before caching to alllow runtime evaluation of constants and environment variables.
+     *
+     * @param string $phpString
+     * @return mixed
+     */
+    protected function replaceVariablesInPhpString($phpString)
+    {
+        $phpString = preg_replace_callback('/
+            (?<startString>\s\')?      # optionally starting a string
+            (?P<fullMatch>%            # an expression is indicated by %
```
skurfuerst: can you indent the regex properly, so one sees where a bracket starts / ends?

kitsunet:
We most certainly could, but for that see #717, the problem is, that I don't have a temp path to save the settings to before the context was injected as we still support setting temporary path by settings which is a hen/egg problem. I will try to eval or save to some other temp path...

Neos.Flow/Classes/Configuration/ConfigurationManager.php
kitsunet: You can use class constants.

robertlemke:
Same like in #717:

if there is a syntax error in the DevelopmentConfigurations.php caused by invalid configuration, you can neither call ./flow cache:flush nor ./flow cache:flush --force

you can provoke such a syntax error by setting a YAML value to %FOO% - because that constant is empty, no value is rendered in PHP, but a ,.

kitsunet:
@skurfuerst @robertlemke I took care of both your comments. The post processing logic is now entirely done in the resulting cache string, resulting in nicely reduced amount of code.

Also the problems with missing constants have been tackled, the conditions in there had some errors still. Now a missing constant is not resulting in a wrong PHP string and the existence of constants is checked at runtime as well.

kitsunet:
\o/ awesome! <3

PR #749, opened 2016-12-02
TASK: Add PHP 7.1 to .travis.yml

kdambekalns:
Now that PHP 7.1 is released, we should make sure we are compatible.


kitsunet:
As the failing test makes it not less broken but just more obvious, merging...

kdambekalns:
👍

PR #750, opened 2016-12-02
FEATURE: NodeType Configuration enhancements

sbruggmann:
Allow to change the YAML Configuration interpretation according to the RFC
https://discuss.neos.io/t/rfc-enable-inheritance-for-nodetype-properties/1607

This PR need the PR https://github.com/neos/neos-development-collection/pull/1292 to work.

Please look at https://github.com/webexcess/WebExcess.InheritProperties for a now possible implementation.

### DIFF
--- a/Neos.Flow/Classes/Configuration/ConfigurationManager.php
+++ b/Neos.Flow/Classes/Configuration/ConfigurationManager.php
@@ -23,7 +23,6 @@ use Neos\Utility\PositionalArraySorter;
  * A general purpose configuration manager
  *
  * @Flow\Scope("singleton")
- * @Flow\Proxy(FALSE)
  * @api
  */
 class ConfigurationManager
@@ -149,7 +148,7 @@ class ConfigurationManager
     protected $orderedListOfContextNames = [];
 
     /**
-     * @var Source\YamlSource
+     * @var Source\YamlSourceInterface
      */
     protected $configurationSource;
 
@@ -189,6 +188,11 @@ class ConfigurationManager
     protected $temporaryDirectoryPath;
 
     /**
+     * @var array
+     */
+    protected $configurationPostProcessors;
+
+    /**
      * Constructs the configuration manager
      *
      * @param ApplicationContext $context The application context to fetch configuration for
@@ -208,12 +212,23 @@ class ConfigurationManager
     /**
      * Injects the configuration source
      *
-     * @param Source\YamlSource $configurationSource
+     * @param Source\YamlSourceInterface $configurationSource
      * @return void
      */
-    public function injectConfigurationSource(Source\YamlSource $configurationSource)
+    public function injectConfigurationSource(Source\YamlSourceInterface $configurationSource)
     {
         $this->configurationSource = $configurationSource;
+    }
+
+    /**
+     * Injects the configuration post processors
+     *
+     * @param array $configurationPostProcessors
+     * @return void
+     */
+    public function injectConfigurationPostProcessors(array $configurationPostProcessors)
+    {
+        $this->configurationPostProcessors = $configurationPostProcessors;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


skurfuerst:
Hey,

Thanks for your pull request and contribution :heart:

just a very quick comment -- you re-enabled the proxying of the Configuration Manager; this won't work because we need the configuration manager extremely early in the bootstrapping process in order to initialize the Object framework (Objects.yaml) and build proxy classes.

We'd need a way to configure this using Package.php - via PHP source code / Signals / Slots. These should work so early in the process.

I'll try to do a more in-depth review later :-)

All the best,
Sebastian

kitsunet:
I agree and feel this implementation is way too low level for the use case. This might still work but as @skurfuerst said must be configured somewhere else. Also the YamlSourceInterface seems unnecessary. We once decided to remove options for alternate configuration sources and thereby removed the ConfigurationSourceInterface. If we wanted to have an interface then that, A lone YamlSourceInterface seems bloated for no reason.

sbruggmann:
> just a very quick comment -- you re-enabled the proxying of the Configuration Manager; this won't work because we need the configuration manager extremely early in the bootstrapping process in order to initialize the Object framework (Objects.yaml) and build proxy classes.

True, i struggled with that early processing and it's ugly, but it worked ;-)
If there is a cleaner way that works too, let me know and i will implement it.
Maybe there is a chance in the near future t implement it right in the core..

> We once decided to remove options for alternate configuration sources and thereby removed the ConfigurationSourceInterface.

This looks like a general question.
Stay fully compatible to the YAML Spec and don't allow additions (secure),
stay compatible in the core and allow differing implementations (risky),
or integrate a File- & Package-wide "YAML Anchors" impenentation (pretty nice).

dfeyer:
Did you test this PR with Symfony YAML component and the PHP YAML Extension ? Really important to keep Flow compatible with the PHP extension as it speed up the configuration parsing a lots.

sbruggmann:
@dfeyer the default behavior has not changed.
But the PR allows to replace the YAML Implementation (not recommended) and (more interesting) to manipulate the merge of the individual configuration array results.

dfeyer:
@sbruggmann Thanks for the details, so basically you need something to post process the configuration before it's cached, right ?

dfeyer:
> We once decided to remove options for alternate configuration sources and thereby removed the ConfigurationSourceInterface.

skurfuerst:
Hey everybody,

> It use existing concept (based on the same pattern as the superTypes, easy to extend, and not yet an other pattern to put in your brain ;). And property can be based on an other property in the same node type or a preset available in the current node type. Really flexible, and the syntax is easy to override (like disable a preset for a extended node type, ....)

I personally would still like to have this implemented in an additional package; because I think we already have node type mixins etc; allowing to cater for (quite many) use cases. I feel adding another level of indirection would just make it harder for people to understand... At least that's what I fear.

As said, I think it'd be great if the core allowed this to be done; but I would not add it to the core.

@sbruggmann just an idea -- do you know that a feature called "Node Type Postprocessors" exist? Basically they allow to hook into the Node type generation; and run PHP code which add/modify/... properties.

IMHO if you add your stuff as a generic post processor to Neos.Neos:Node (or so) this might actually already work. See this code example for how the postprocessor is invoked: https://github.com/neos/neos-development-collection/blob/master/Neos.ContentRepository/Classes/Domain/Model/NodeType.php#L226
https://github.com/neos/neos-development-collection/blob/master/Neos.Neos/Configuration/NodeTypes.yaml#L254

Maybe this works and we don't need a core patch at all? 🙃 

All the best,
Sebastian

skurfuerst:
Hey @sbruggmann,

did you have the chance to look into my idea from above already? :-)

All the best,
Sebastian

sbruggmann:
Hi @skurfuerst ,

Sorry the last days were heavy.
The "Node Type Postprocessor" test produced for example wrong i18n keys and was not able to handle a deep inheritance like "Package.XY:BaseType, Package.XY:ZType, Site.XY:ZTypeFinal".

albe:
So is there still something left to do or can we close this issue?

sbruggmann:
@albe after thinking about some "weeks".. it would be really nice, but I can totally understand what @skurfuerst said:
> I feel adding another level of indirection would just make it harder for people to understand...

So from my side, we can close this PR unresolved.

PR #755, opened 2016-12-05
!!! TASK: Remove deprecated support of temporary path setting

kitsunet:
The setting ``TYPO3.Flow.utility.environment.temporaryDirectoryBase``
was deprecated and with this change finally removed.

The temporary path defaults to FLOW_PATH_ROOT . 'Data/Temporary', but
you can always override the temporary path via the environment variable
``FLOW_PATH_TEMPORARY_BASE`` instead.

Note that in either case a sub path will be created based on the
current application context.

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -61,7 +61,7 @@ class Scripts
      */
     public static function initializeClassLoader(Bootstrap $bootstrap)
     {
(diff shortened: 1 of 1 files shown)


kitsunet:
Note that the current state has strong performance drawbacks basically rendering our configuration cache useless as @johannessteu proofed in profiling.
The environment variable was introduced over a year ago and this is a major release so our chance to change this.

kdambekalns:
Is there really no documentation related to this? Defaults in Settings.yaml? There should be more that needs to be removed, no?

johannessteu:
At least i can't find anything related @kdambekalns

PR #756, opened 2016-12-06
BUGFIX: Invalid namespaces and class references

robertlemke:
This change fixe a couple of stale reference to classes which
have been moved around recently.

### DIFF
--- a/Neos.Flow/Classes/Cache/CacheFactory.php
+++ b/Neos.Flow/Classes/Cache/CacheFactory.php
@@ -15,6 +15,7 @@ use Neos\Cache\Backend\BackendInterface;
 use Neos\Cache\Backend\SimpleFileBackend;
 use Neos\Cache\CacheFactoryInterface;
 use Neos\Cache\EnvironmentConfiguration;
+use Neos\Cache\Frontend\FrontendInterface;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Cache\Backend\AbstractBackend as FlowAbstractBackend;
 use Neos\Flow\Cache\Backend\FlowSpecificBackendInterface;
@@ -101,8 +102,8 @@ class CacheFactory extends \Neos\Cache\CacheFactory implements CacheFactoryInter
      * @param string $cacheObjectName
      * @param string $backendObjectName
      * @param array $backendOptions
-     * @param boolean $persistent
-     * @return Frontend\FrontendInterface
+     * @param bool $persistent
+     * @return FrontendInterface
      */
     public function create($cacheIdentifier, $cacheObjectName, $backendObjectName, array $backendOptions = [], $persistent = false)
     {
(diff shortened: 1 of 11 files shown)


Neos.Flow/Classes/Cache/CacheFactory.php
```diff
@@ -101,8 +102,8 @@ public function __construct(ApplicationContext $context, Environment $environmen
      * @param string $cacheObjectName
      * @param string $backendObjectName
      * @param array $backendOptions
-     * @param boolean $persistent
-     * @return Frontend\FrontendInterface
```
kdambekalns: That isn't invalid, it just relative to the current namespace. Just for the record. :)

Neos.Flow/Classes/Cache/CacheFactory.php
```diff
@@ -131,8 +132,8 @@ protected function instantiateCache($cacheIdentifier, $cacheObjectName, $backend
      * @param string $backendObjectName
      * @param array $backendOptions
      * @param boolean $persistent
-     * @return FlowAbstractBackend|Backend\BackendInterface
-     * @throws Exception\InvalidBackendException
+     * @return FlowAbstractBackend|BackendInterface;
```
kdambekalns: No semicolon, please.

Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
```diff
@@ -775,7 +776,7 @@ protected function evaluatePropertyAnnotations(ORM\ClassMetadataInfo $metadata)
                         'allocationSize' => $seqGeneratorAnnotation->allocationSize,
                         'initialValue' => $seqGeneratorAnnotation->initialValue
                     ]);
-                } elseif ($this->reader->getPropertyAnnotation($property, ORM\TableGenerator::class) !== null) {
```
kdambekalns: This is "correct", carried over from https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L332-L333 - it could be dropped, but then again, our driver is deliberately kept as close to the original one to ease updates…

Neos.Flow/Classes/Persistence/Doctrine/Service.php
```diff
@@ -201,7 +201,7 @@ function ($message) use ($that) {
         }
 
         $configuration = new Configuration($connection, $outputWriter);
-        $configuration->setMigrationsNamespace(\Neos\Flow\Persistence\Doctrine\Migrations::class);
```
kdambekalns: IMHO this should be really a string to begin with. It a namespace, not a `::class`. And even `Migrations::class` does not resolve.

robertlemke:
Thanks for the review!

Neos.Flow/Classes/Cache/CacheFactory.php
```diff
@@ -131,8 +132,8 @@ protected function instantiateCache($cacheIdentifier, $cacheObjectName, $backend
      * @param string $backendObjectName
      * @param array $backendOptions
      * @param boolean $persistent
-     * @return FlowAbstractBackend|Backend\BackendInterface
-     * @throws Exception\InvalidBackendException
+     * @return FlowAbstractBackend|BackendInterface
+     * @throws InvalidBackendException
```
bwaidelich: Is the `InvalidBackendException` imported? I fail to see the corresponding `use` statement

kdambekalns: Yes, in line 22 :)

bwaidelich: Doh, must be the age ;)

PR #758, opened 2016-12-06
FEATURE: Scalar to object converter

robertlemke:
This introduces a simple type converter which can convert
a scalar value (string, integer, float or boolean) into an
object by passing that value to the class constructor.

This converter helps developers using Value Objects (not
managed by the persistence framework) or other Data
Transfer Objects in places where type conversion is
supported. One common case is to use Value Object class
names as a type hint for arguments in a command line
controller method.


robertlemke:
@kitsunet thanks!

kitsunet:
Narf, but the priority breaks tests...

PR #761, opened 2016-12-08
TASK: Add PSR-7 compatibility to HTTP stack

kitsunet:
This adds all missing methods and implementations to make our
HTTP implementation PSR-7 compatible.

Further adjustments to Flow are not included yet.


Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -295,6 +296,184 @@ public function removeCookie($name)
...
+    }
+
+    /**
+     * Gets the body of the message.
+     *
+     * @return StreamInterface Returns the body as a stream.
+     */
+    public function getBody()
+    {
+        $streamResource = fopen('php://temp', 'w+');
```
albe: Wouldn't  `fopen('php://memory','r+');` be better? Don't write that body content to disk unneccessarily

albe: Also needs a rewind after the write, no?

Neos.Flow/Classes/Http/BaseRequest.php
```diff
@@ -0,0 +1,377 @@
...
+     * changed request target.
+     *
+     * @link http://tools.ietf.org/html/rfc7230#section-2.7 (for the various
+     *     request-target forms allowed in request messages)
+     * @param mixed $requestTarget
+     * @return self
+     */
+    public function withRequestTarget($requestTarget)
+    {
+    }
```
albe: TODO reminder comment.

composer.json
```diff
@@ -6,11 +6,12 @@
     ],
     "type": "neos-package-collection",
     "require": {
-        "php": ">=7.0.0",
+        "php": ">=5.5.0",
```
albe: ?

kitsunet: Narf, a problem with our merge script...

Neos.Flow/Classes/Http/BaseRequest.php
kitsunet: implemented.

Neos.Flow/Classes/Http/BaseRequest.php
```diff
@@ -0,0 +1,389 @@
+<?php
+namespace Neos\Flow\Http;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+use Neos\Utility\MediaTypes;
+
+/**
+ *
```
robertlemke: I think this needs some explanation … what's a "Base Request"?

Neos.Flow/Classes/Http/BaseRequest.php
```diff
@@ -0,0 +1,389 @@
...
+     */
+    protected $requestTarget = null;
+
+    /**
+     * BaseRequest constructor.
+     *
+     * @param UriInterface $uri
+     * @param string $method
+     */
+    public function __construct(UriInterface $uri, $method = 'GET')
```
robertlemke: You expect `UriInterface` in the constructor, but the doc comment for `$this->uri` states that it is a concrete `Uri`. 

Why does the constructor argument `$method` need to be optional? 

The `$method` argument does not have a `string` type hint. And should be checked if it actually is a string with length > 0.

Neos.Flow/Classes/Http/BaseRequest.php
```diff
@@ -0,0 +1,389 @@
...
+        if ($this->baseUri === null) {
+            $this->detectBaseUri();
+        }
+
+        return $this->baseUri;
+    }
+
+    /**
+     * @param Uri $baseUri
+     */
```
robertlemke: is it on purpose that `setBaseUri` is not public API? And it lacks a type hint.

Neos.Flow/Classes/Http/BaseRequest.php
```diff
@@ -0,0 +1,389 @@
...
+    public function setBaseUri($baseUri)
+    {
+        $this->baseUri = $baseUri;
+    }
+
+    /**
+     * Tries to detect the base URI of request.
+     *
+     * @return void
```
robertlemke: I think the `if` clause should be outside this method, because running `detectBaseUri` and then not detecting the base uri is somewhat unexpected. Even better would be if this method would return the uri instead of setting it.

Neos.Flow/Classes/Http/BaseRequest.php
```diff
@@ -0,0 +1,389 @@
...
+    /**
+     * BaseRequest constructor.
+     *
+     * @param UriInterface $uri
+     * @param string $method
+     */
+    public function __construct(UriInterface $uri, $method = 'GET')
+    {
+        $this->uri = $uri;
+        $this->method = $method;
```
robertlemke: In this case `$method` can turn out to be a lowercase string. You should rather call `$this->setMethod()` to make sure it is the valid.

Neos.Flow/Classes/Http/BaseRequest.php
```diff
@@ -0,0 +1,389 @@
...
+     * resource, that resource will be returned, no matter if $asResource is set.
+     *
+     *
+     * @param boolean $asResource If set, the content is returned as a resource pointing to PHP's input stream
+     * @return string|resource
+     * @api
+     * @throws Exception
+     */
+    public function getContent($asResource = false)
+    {
```
robertlemke: does it have to be this way? I think that returning two different types depending on a given argument is a bad habit. I'd rather have `getContentAsResource` and `getContent` / `getContentAsString`.

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -0,0 +1,314 @@
...
+    /**
+     * Attach a new stream/resource to the instance.
+     *
+     * @param string|resource $stream
+     * @param string $mode
+     */
+    public function replace($stream, $mode = 'r')
+    {
+        $this->close();
+        $this->resource = $stream;
```
robertlemke: that's ugly. `$this->resource = $stream` and then afterwards checking if it's really a resource?

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -0,0 +1,314 @@
...
+     * @param string $mode Mode with which to open stream
+     * @throws \InvalidArgumentException
+     */
+    public function __construct($stream, $mode = 'r')
+    {
+        $this->replace($stream, $mode);
+    }
+
+    /**
+     * {@inheritdoc}
```
robertlemke: why do we have inheritdoc here, but not on the other methods? I'd pefer to have actual doc comments here.

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -0,0 +1,314 @@
...
+
+        if (!$this->hasValidResource()) {
+            throw new \InvalidArgumentException('Invalid stream provided; must be a string or stream resource', 1453891861);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSize()
```
robertlemke: what does it do? what does it return?
why can it actually return null and not throw an exception if the resource isn't valid?

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -0,0 +1,314 @@
...
+    {
+        if (!$this->hasValidResource()) {
+            return null;
```
robertlemke: this should throw an exception if the resource is not valid! "null" would mean "unknown", not "an error occurred".

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -0,0 +1,314 @@
...
+
+        $stats = fstat($this->resource);
+
+        return $stats['size'];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function tell()
```
robertlemke: this should be documented. Without knowing the standard, to me it's quite unexpected that `tell` returns something.

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -0,0 +1,314 @@
...
+     * @param string $stream String stream target or stream resource.
+     * @param string $mode Resource mode for stream target.
+     * @return void
+     */
+    protected function openStream($stream, $mode = 'r')
+    {
+        $resource = fopen($stream, $mode);
+        $this->resource = $resource;
+    }
+
```
robertlemke: doc comment missing

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -0,0 +1,314 @@
...
+    protected function ensureResourceReadable()
+    {
+        if ($this->isReadable() === false) {
+            throw new \RuntimeException('Stream is not readable.', 1453892039);
+        }
+    }
+
+    /**
+     *
```
robertlemke: comment

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -0,0 +1,314 @@
...
+        if (!$this->isReadable()) {
+            return '';
+        }
+
+        try {
+            $this->rewind();
+
+            return $this->getContents();
+        } catch (\Exception $e) {
+            return '';
```
robertlemke: why catch _any_ exception here? Shouldn't this better break if something is wrong?

Neos.Flow/Classes/Http/FlowUploadedFile.php
```diff
@@ -0,0 +1,50 @@
...
+     */
+    protected $originallySubmittedResource;
+
+    /**
+     * @var string
+     */
+    protected $collectionName;
+
+    /**
+     * @return array|string
```
robertlemke: please document what it does and how that array or string could look like

Neos.Flow/Classes/Http/BaseRequest.php
kitsunet: Many implementations, including our current HTTP Request default to GET, so I felt it sensible to apply the same here.

Neos.Flow/Classes/Http/BaseRequest.php
kitsunet: I also odn't like it but it's `@api` and is not touched by my change (merely moved around) so it is existing API. We can deprecate it and get rid of this in the next major, but that is something for a follow up, because it's unrelated to the PSR-7 additions.

Neos.Flow/Classes/Http/ContentStream.php
kitsunet: Can't throw exceptions in __toString or PHP will fatal.

Neos.Flow/Classes/Http/ContentStream.php
kitsunet: I somewhat agree but the Interface states throws for some methods, not for this one though. We could still throw one here but other implementations generally do not.

Neos.Flow/Classes/Http/ContentStream.php
kitsunet: See next comment.

hlubek:
I'd love to get PSR7 support (especially streams for requests and responses) into Flow. What's the current state here and what's missing to get this into the next release (hopefully)?

Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -295,6 +297,185 @@ public function removeCookie($name)
...
+    }
+
+    /**
+     * Gets the body of the message.
+     *
+     * @return StreamInterface Returns the body as a stream.
+     */
+    public function getBody()
+    {
+        $streamResource = fopen('php://memory', 'r+');
```
hlubek: Can we think about a way to use streams internally instead of treating the response "content" as a string?

kitsunet: Sure, but please let's make iterations. As long as there is nothing broken I would first like to get this in...

hlubek: 👍 Sure, let's get that in first

Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -295,6 +297,185 @@ public function removeCookie($name)
...
+     */
+    public function getHeaderLine($name)
+    {
+        $headerLine = $this->headers->get($name);
+        if ($headerLine === null) {
+            $headerLine = '';
+        }
+
+        if (is_array($headerLine)) {
+            $headerLine = explode(', ', $headerLine);
```
kdambekalns: Wait, `explode()`!? `implode()` is the better choice, no?

Neos.Flow/Classes/Http/ContentStream.php
kdambekalns: Generally swallowing exceptions without leaving a trace is a smell… what about logging the exception?

kitsunet: Would mean to inject a logger into this class. I really would like to avoid that. But we can ofc change that if there is a general agreement for this.

Neos.Flow/Classes/Http/AbstractMessage.php
kdambekalns: @kitsunet This still looks very wrong.

Neos.Flow/Classes/Http/ContentStream.php
kdambekalns: I'd vote for injecting a logger. It is injected lazily, so unless used not (much) waste…

kitsunet:
I disagree on the logger becauae dependencies... But let's do it...

kitsunet:
Done and done.

kdambekalns:
Now we have passing tests (again) and all should be settled. @robertlemke might want to have another look as to not block this… :)

robertlemke:
Actually, I didnt have another look, just wanted to remove my block.

gerhard-boden:
Ah yes, and one day someone will have to add some return type declarations, but let's merge it as is first.

Neos.Flow/Classes/Http/Request.php
```diff
@@ -889,7 +889,7 @@ public function withUploadedFiles(array $uploadedFiles)
      */
     public function getParsedBody()
     {
-        return isset($this->parsedBody) ? $this->parsedBody : $this->arguments;
+        return $this->parsedBody ?? $this->arguments;
```
kitsunet: That's how old this PR already is :D

Neos.Flow/Classes/Http/Uri.php
```diff
@@ -354,11 +356,276 @@ public function getFragment()
...
+
+        $this->fragment = $fragment;
+    }
+
+    /**
+     * @return string
+     */
+    protected function getHostAndOptionalPort()
+    {
+        $hostAndPort = (string)$this->host;
```
kitsunet: Probably not, but I have gotten inspiration from a bunch of other implementations that did it. Right now I must admit I no longer know exactly why, but I would for now rather want to leave it in.

kitsunet:
Yep probably we should add the @api annotations but IMHO also that can be done afterwards. This is the major work to get the code interface compatible. The next iteration would be to sharpen the implementation (by adding missing @api etc.) maybe also deprecating duplicate existing methods and finally ofc switching our codebase to (only) use the PSR-7 interface methods and then see if something is amiss.

kitsunet:
It's just too big to do it in one pass IMHO.

gerhard-boden:
Roger that, sounds like this is good to go! 

Will merge within 24 hrs if nobody objects

gerhard-boden:
@kitsunet So I locally rebased this PR on master, but image upload is still broken? 

Now I get 
```
[Thu Aug 10 15:03:10.241146 2017] [:error] [pid 3924] [client 192.168.230.1:58468] PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Neos\\Flow\\Http\\Request::createUploadedFilesFromUntangledUploads() must be of the type array, string given, called in /var/www/neos-master/Packages/Framework/Neos.Flow/Classes/Http/Request.php on line 627 and defined in /var/www/neos-master/Packages/Framework/Neos.Flow/Classes/Http/Request.php:622\nStack trace:\n#0 /var/www/neos-master/Packages/Framework/Neos.Flow/Classes/Http/Request.php(627): Neos\\Flow\\Http\\Request->createUploadedFilesFromUntangledUploads('neos-transition...', 'neos-transition...')\n#1 /var/www/neos-master/Packages/Framework/Neos.Flow/Classes/Http/Request.php(627): Neos\\Flow\\Http\\Request->createUploadedFilesFromUntangledUploads(Array, Array)\n#2 /var/www/neos-master/Packages/Framework/Neos.Flow/Classes/Http/Request.php(627): Neos\\Flow\\Http\\Request->createUploadedFilesFromUntangledUploads(Array, Array)\n#3 /var/www/neos-master/Packages/Framework/Neos.Flow/Classes/Http/Request.php(627): Neos\\Flow\\Http\\Request->createUploadedFilesFromUntangledUploads(Array, A in /var/www/neos-master/Packages/Framework/Neos.Flow/Classes/Http/Request.php on line 622, referer: https://neos.dev/neos/management/media```

Inside `error.log`

kitsunet:
Thanks <3

gerhard-boden:
It's been almost 25 hrs. 

PSR-7 here we come!

kitsunet:
Fantastic, my hero <3

gerhard-boden:
C'mon, you implemented all this  ❤️ . I just read it and clicked a shiny button 😃

albe:
Thanks to both of you!

PR #764, opened 2016-12-08
TASK: Remove deprecated ``Http\Request::createActionRequest``

kitsunet:
The static method was deprecated since Flow 2.3 and is due for
removal.
Instead create an instance of ``ActionRequest`` via the ``new``
keyword.


kitsunet:
Needs https://github.com/neos/neos-development-collection/pull/1303

PR #767, opened 2016-12-09
!!! TASK: Remove deprecated ``PropertyMappingConfigurationBuilder``

kitsunet:
The ``PropertyMappingConfigurationBuilder`` class was deprecated and
is bound to be removed.

It can be fully replaced by calling
``PropertyMapper::buildPropertyMappingConfiguration`` from now on.

Usage in tests is replaced in the same manner.


kitsunet:
Needs neos/neos-development-collection#1309

PR #775, opened 2016-12-14
BUGFIX: typeconverter:list CLI command is broken

dfeyer:


dfeyer:
I can push an other PR to make the output more nice, or if it's fine to include this in a bugfix, i can update this one

robertlemke:
@dfeyer some styling issues left …

And since this is a bug fix (plus some little feature …), can you add a little PR description which we can use as the commit body when we merge this?

Issue #748, opened 2016-12-01
TYPO3Fluid: Broken support for mixed type render parameters


bwaidelich:
### Description

Previously the render method could have arguments with mixed type annotations (divided by |)

### Steps to Reproduce

1. Create a custom ViewHelper:

```php
class MyViewHelper extends AbstractViewHelper
{
  /**
   * @param array|\Iterator $someArgument
   */
  public function render($someArgument)
  {
  }
}
```

2. Use it with a valid `someArgument` of type `\Iterator` (`array` works fine)

#### Expected behavior

no output

#### Actual behavior

Exception:
```
#1256475113: The argument "someArgument" was registered with type "array|\Iterator", but is of type "SomeIterator" in view helper…
```

### Affected Versions

Flow 4.0 (using TYPO3Fluid).

PR #785, opened 2016-12-19
BUGFIX: Fix support for multi-type render arguments

kitsunet:
ViewHelper render method arguments with multiple type
defintions in the docblock are now correctly registered
as mixed type arguments.

Fixes: #748


kitsunet:
@bwaidelich there we go

bwaidelich:
Thanks for this!
FYI: There are some StyleCI issues apparently: https://styleci.io/analyses/qv0lVx

robertlemke:
@bwaidelich I'll fix the style issues in a new PR

PR #788, opened 2016-12-20
BUGFIX: Fix core migration to Flow 4.0

skurfuerst:
This change, after thorough discussion, fixes the following issues:

- When somebody specified "flow core:migrate" without arguments, or mis-spells the package-key argument ("flow core:migrate package=foo"), he now gets an error message instead of all packages which are non-core-packages being migrated.

- We remove one migration which does not apply cleanly; which was part of Flow 3.x. This migration is also not needed anymore; as people are usually upgrading from Flow 3.2/3.3 where this migration was already applied.

This now also fixes the case that if you upgrade a package which was created for Flow 3.3 to 4.0, all old migrations are run -- and a fatal error existed at the Fluid migration step.


Neos.Flow/Scripts/Migrations/Manager.php
```diff
@@ -125,41 +120,22 @@ public function getStatus($packageKey = null, $versionNumber = null)
      * @param boolean $force if TRUE migrations will be applied even if the corresponding package is not a git working copy or contains local changes
      * @return void
      */
-    public function migrate($packageKey = null, $versionNumber = null, $force = false)
+    public function migrate($packageKey, $versionNumber = null, $force = false)
     {
+        if (!$packageKey) {
```
skurfuerst: I'd say no, it cannot happen; but I wanted to add this additional safeguard. fine for you?

I'll adjust the param annotation.

bwaidelich: > fine for you?

OFC, I was just a bit puzzled ;)

kdambekalns:
Just one thing: what if people actually have code that would need the removed migration? It will lead to escaped output, better safe than sorry, but they get no hint whatsoever. Maybe instead emit a notice during the migration? Feel free to ignore me, though…

bwaidelich:
> Maybe instead emit a notice during the migration?

I agree and can take care of this

Neos.Flow/Classes/Command/CoreCommandController.php
```diff
@@ -235,8 +235,8 @@ public function setFilePermissionsCommand($commandlineUser, $webserverUser, $web
     /**
      * Migrate source files as needed
      *
-     * This will apply pending code migrations defined in packages to all
-     * packages that do not yet have those migration applied.
```
bwaidelich: the package key is no longer optional, adjusted doc comment accordingly

Neos.Flow/Classes/Command/CoreCommandController.php
```diff
@@ -246,16 +246,16 @@ public function setFilePermissionsCommand($commandlineUser, $webserverUser, $web
...
      * @param boolean $status Show the migration status, do not run migrations
      * @param string $packagesPath If set, use the given path as base when looking for packages
-     * @param string $packageKey If set, migrate only the given package
      * @param string $version If set, execute only the migration with the given version (e.g. "20150119114100")
      * @param boolean $verbose If set, notes and skipped migrations will be rendered
      * @param boolean $force By default packages that are not under version control or contain local changes are skipped. With this flag set changes are applied anyways (changes are not committed if there are local changes though)
      * @return void
      * @see neos.flow:doctrine:migrate
      */
-    public function migrateCommand($status = false, $packagesPath = null, $packageKey = null, $version = null, $verbose = false, $force = false)
```
bwaidelich: just moving making the package key param non-optional and renaming it to `package` for consistency with other CLI commands

Neos.Flow/Scripts/Migrations/Manager.php
```diff
@@ -39,11 +39,6 @@ class Manager
     /**
      * @var array
      */
-    protected $ignoredPackageCategories = array('Framework', 'Libraries');
```
bwaidelich: No longer needed as we now always execute migrations for a specific package

Neos.Flow/Scripts/migrate.php
```diff
@@ -46,11 +46,18 @@
 
 $migrationsManager = new Manager();
 
-$packageKey = getFlagValue('package-key');
+if (!isset($GLOBALS['argv'][0]) || substr($GLOBALS['argv'][0], 0, 2) === '--') {
+    outputLine('EXCEPTION: No package key specified.', [], 0, STYLE_ERROR);
+    outputLine('  That package key has to be specified as first argument like "./flow flow:core:migrate Foo.Bar:MyPackage"', [], 0, STYLE_ERROR);
+
```
bwaidelich: The package key argument needs to be specified before any flags because otherwise we'd need to implement the whole flag vs argument logic which is not trivial..

Neos.Flow/Scripts/migrate.php
```diff
@@ -62,17 +69,15 @@
 $verbose = flagIsSet('verbose');
 
 if (flagIsSet('status')) {
-    outputLine('Fetching migration status...');
+    outputHeadline('Migration status for package "%s"', 1, [$packageKey]);
     try {
         $status = $migrationsManager->getStatus($packageKey, $versionNumber);
     } catch (\Exception $exception) {
-        outputLine('EXCEPTION: %s', array($exception->getMessage()), 0, STYLE_ERROR);
+        outputLine('EXCEPTION: %s', [$exception->getMessage()], 0, STYLE_ERROR);
```
bwaidelich: I couldn't resist replacing `array()` with `[]` on the way..

Issue #789, opened 2016-12-20
Add Flow subroutes via Settings


bwaidelich:
Since version 3.1 Flow supports [Routes to be configured via Settings](http://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Routing.html#subroutes-from-settings).
We should make that the new default and get rid of the global `Routes.yaml` for greater flexibility.

PR #790, opened 2016-12-20
TASK: Load Flow Routes from settings

bwaidelich:
This adjusts the default Configuration to include the Flow default routes
via Settings.

Background:
Since version 3.1 Flow supports sub routes to be included via configuration
(see http://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Routing.html#subroutes-from-settings).
This is the recommended way to register routes as it provides greater flexibility.
Note: Any existing `Routes.yaml` file in the global `Configuration` folder take
precedence over routes loaded via Settings.

Related: #789


aertmann:
yay thanks for taking care! big fan 😄

Issue #794, opened 2016-12-21
Do not throw exception for unknown core commands


gerhard-boden:
### Description

No exception should not be thrown when running an uknown core command.

### Steps to Reproduce

1. Try to run the following command: `./flow flow:core:setfilempermissions` (there is a typo in there)

#### Expected behavior

The general `command not found message` 
`No command could be found that matches the command identififier ....`

#### Actual behavior

The following exception is thrown:
```
Could not inject configured property "packageManager" into "Neos\Flow\Command\HelpCommandController" because no injection method exists, but for compile time use this is required.
Configuration source: automatically registered class.

  Type: Neos\Flow\ObjectManagement\Exception\UnresolvedDependenciesException
  Code: 1297110953
  File: Packages/Framework/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectMana
        ger.php
  Line: 453

Open Data/Logs/Exceptions/20161221004935cfa43e.txt for a full stack trace.
```

### Affected Versions

Flow: at least since 3.0

PR #795, opened 2016-12-21
BUGFIX: Call of non existing core command issues proper error

kitsunet:
And exception was thrown when a non existing core command was being
invoked, eg. due to a spelling mistake.
The exception stems from the fact that all core commands are considered
compiletime commands but handling of non existing command calls is
done by the ``HelpCommandController`` which itself was not compiletime
compatible. This change adapts it to use inject methods which resolves
the problem.

Note that in upmerges the injections must be adapted to changing dependencies.

Fixes: #794

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/HelpCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/HelpCommandController.php
@@ -50,6 +50,7 @@ class HelpCommandController extends CommandController
 
     /**
      * @param Bootstrap $bootstrap
+     * @return void
      */
     public function injectBootstrap(Bootstrap $bootstrap)
     {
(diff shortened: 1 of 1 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Command/HelpCommandController.php
```diff
@@ -29,30 +29,58 @@
...
     protected $applicationPackageKey;
 
     /**
-     * @Flow\Inject
      * @var CommandManager
      */
     protected $commandManager;
 
     /**
+     * @param Bootstrap $bootstrap
```
bwaidelich: <nitpick>Not sure about our current policy regarding `@return void` annotations, but it's missing here ;)</nitpick>

kdambekalns: Done.

gerhard-boden:
ahh please don't merge just found the time to test it :D

PR #803, opened 2016-12-28
FEATURE: webserver group membership check

wexi79:
File permission script checks that the command line user is a member of the webserver group

### DIFF
--- a/Neos.Flow/Scripts/setfilepermissions.sh
+++ b/Neos.Flow/Scripts/setfilepermissions.sh
@@ -32,6 +32,17 @@ COMMANDLINE_USER="$1"
 WEBSERVER_USER="$2"
 WEBSERVER_GROUP="$3"
 
+# Check that command line user is a member of the webserver group
+
+if id -nG "$COMMANDLINE_USER" | grep -qw "$WEBSERVER_GROUP"; then
+  echo "User is a member of the webserver group $WEBSERVER_GROUP, continuing"
+else
+  echo "User is not a member of the webserver group $WEBSERVER_GROUP, exiting"
+  echo
+  exit 1;
+fi
+
+
 echo "Checking permissions from here upwards."
 
 unset PARENT_PATH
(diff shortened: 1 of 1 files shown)


dfeyer:
I agree that this can be a best practice and a good setup. But dependending on the hosting environnement ... not sure if it's possible to enforce this as a hard rule.

wexi79:
Well, if the command line user is not a member of the webserver group, things will not work. Clearing the cache from the command line for example. So basically when running the script, you are supposed to give a group as the third option that the cli user is a member of.

dfeyer:
@wexi79 I'm all in favor a being more strict/clear about our required setup ;)

But let's wait some more review before merging this one

wexi79:
@dfeyer yeah, I havent' been abl to think about a situation when the group membership cannot be enforced but maybe someone can :)

kdambekalns:
Hm. On my current setup  my user is not in the webserver group and things still work. Don't ask why, but… maybe make this a warning instead of a failure?

wexi79:
@kdambekalns I'm still going to ask why :D 

Mainly I'm interested what kind of setup you have where your cli user IS NOT a member of the webserver group and things still work; how can the cli user remove cache/lock files from the Data directory if it is not the owner of those files or a member of the group? If the cli user and webserver user are the same, the user would still be a member of the webserver group also?

kdambekalns:
Probably because of ACLs…

wexi79:
Ah, you're right, then the check is not required. So maybe instead of displaying the warning, just move the check after the ACL support check? As if the script uses the file system method for setting permission and the cli user is not a member, you will get a broken setup (which is annoying).

albe:
Needs a rebase, but what else is keeping this from merging, since it already got two approvals?

albe:
Rebased. Would merge this evening if no one objects any more.

albe:
So congrats to your first contribution @wexi79 and sorry it took so long :)

PR #805, opened 2017-01-03
TASK: Remove Parameter as it is the default value of the optional

NHoeller:
The parameter is the default value of the optional parameter in the EnvironmentConfiguration class.
it is not necessary to insert it additionally.

### DIFF
--- a/Neos.Flow/Classes/Cache/CacheFactory.php
+++ b/Neos.Flow/Classes/Cache/CacheFactory.php
@@ -90,8 +90,7 @@ class CacheFactory extends \Neos\Cache\CacheFactory implements CacheFactoryInter
 
         $environmentConfiguration = new EnvironmentConfiguration(
             FLOW_PATH_ROOT . '~' . (string)$environment->getContext(),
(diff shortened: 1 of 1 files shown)


NHoeller:
i found a minor bug in the EnvironmentConfiguration class because the property $maximumPathLength got the Value PHP_MAXPATHLEN but MUST be declared in the constructor. So i did a PR to that class to change the constructor to
`public function __construct($applicationIdentifier, $fileCacheBasePath, $maximumPathLength = PHP_MAXPATHLEN)`
In Order to this the value in the constructor call in the CacheFactory set as $maximumPathLength has not to be omitted if it is identical to the value of PHP_MAXPATHLEN.

see [github.com/neos/cache](url:https://github.com/neos/cache/blob/master/Classes/EnvironmentConfiguration.php)

aertmann:
should to be applied against 4.0, not master.. although not that important since it's a cosmetic fix really

kdambekalns:
Has been solved in the meantime, but thanks for the PR!

PR #806, opened 2017-01-03
BUGFIX: Use missing method EnvironmentConfiguration::getApplicationContext

dfeyer:


aertmann:
this would mean that the caches are shared between the contexts right? that would be really bad.. why not get the context from Bootstrap instead?

kitsunet:
@aertmann this is the splitted cache package, it shoudl know nothing about flow. The application identifier needs to contain everything ot uniquely identify the application, so in our case also the context, which in turn is configured in Flows CacheManager or Factory.

aertmann:
@kitsunet: okay makes sense, but does it? that's the important part

aertmann:
nvm it does, https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Cache/Backend/BackendCompatibilityTrait.php#L48

PR #807, opened 2017-01-03
TASK: Fix wrong phpdoc in RedisBackend::setPort

dfeyer:


dfeyer:
Not really related to this one, but we should update our CGL for the integer -> int ... in phpdoc and adapt the code everywhere.

PR #808, opened 2017-01-03
BUGFIX: Missing $context variable in PdoBackend

dfeyer:
This change also cache the context value, to avoid too much call of the
md5 functions.


kitsunet:
Looks good!

PR #813, opened 2017-01-06
TASK: Use RUNLEVEL_* constants in Bootstrap

NHoeller:
Replace the plaintext run level by the defined class constants so changes on the constants will not take no effect there.

### DIFF
--- a/Neos.Flow/Classes/Core/Bootstrap.php
+++ b/Neos.Flow/Classes/Core/Bootstrap.php
@@ -295,7 +295,7 @@ class Bootstrap
      */
     public function buildCompiletimeSequence()
     {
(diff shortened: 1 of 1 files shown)


kdambekalns:
Thanks for the change! But… we cannot merge this.

As @dfeyer said, this is a mismatch in case (the tests still pass, since we actually use that check only for locking the site in case of parallel requests, something the tests don't cover).

But even if we adjusted that, we must not change this: the string is also used as identifier for the `Sequence` instance built, and that identifier is passed to any slots listening to the the `beforeInvokeStep` and `afterInvokeStep` signals. So a change would be breaking something that is public API.

PR #816, opened 2017-01-08
TASK: Remove try catch block in Files::getFileContents

dfeyer:
This change remove the try catch block, file_get_contents never throw an
exception and definitively not a Flow exception. Maybe a left over after
some refactoring ...


kdambekalns:
But if it emits a warning, that will be converted into an exception by our settings, so… Did you check what happens with a non-readable file, e.g.?

aertmann:
what @kdambekalns said

kitsunet:
Maybe a test for that case would help :)

PR #818, opened 2017-01-08
TASK: Cast arguments in TextIterator constructor

dfeyer:


Neos.Utility.Unicode/Classes/TextIterator.php
```diff
@@ -195,11 +195,11 @@ public function last()
     }
 
     /**
-     * Returns the next elment following the character of the original string
+     * Returns the next element following the character of the original string
      * given by its offset
      *
      * @param integer $offset The offset of the character
-     * @return string The element following this character
+     * @return integer The element following this character
```
kitsunet: `string` was right here? Or either the method name or implementation are wrong. I would expect that this method returns the next character in the subject after the one in offset.

Neos.Utility.Unicode/Classes/TextIterator.php
```diff
@@ -218,7 +218,7 @@ public function following($offset)
      * Returns the element preceding the character of the original string given by its offset
      *
      * @param integer $offset The offset of the character
-     * @return string The element preceding this character
+     * @return integer The element preceding this character
```
kitsunet: Same as above

bwaidelich:
What does this do? ;)

kdambekalns:
This is in master by now via #1018 in more or less the same way. Thanks anyway, @dfeyer!

PR #820, opened 2017-01-12
FEATURE: Add Tideways instrumentation for bootstrap process

kdambekalns:
This adds Tideways instrumentation to the Flow bootstrapping.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
@@ -12,6 +12,7 @@ namespace TYPO3\Flow\Core\Booting;
  */
 
 use Doctrine\Common\Annotations\AnnotationRegistry;
+use Tideways\Profiler;
 use TYPO3\Flow\Annotations as Flow;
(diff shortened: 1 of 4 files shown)


kdambekalns:
My major concern: can we somehow move this to an external package? The parts of the code that are adjusted are not really reachable via `Package::boot()` or AOP, so… any other ideas?

kdambekalns:
Ah, see also https://github.com/tideways/profiler/pull/6 and https://github.com/tideways/php-profiler-extension/pull/32

kitsunet:
Oh and IF everyone else should decide that we could merge this, then clearly not in 3.3 but in master.

kdambekalns:
As I wrote:

> My major concern: can we somehow move this to an external package?

Also, _WIP_ and _Help needed_ for the very same reason. So…

And no, external `Package` class is too late, at least if using the `boot` method. I could just put watch calls into the global scope inside such a `Package.php`… Might be worth a try, **but** violates autoloading rules and CGL (side-effect free class files.)

bwaidelich:
I'd also prefer a more general way to hook into bootstrapping for these kind of things
Couldn't we introduce some signals for that or is that too late as well`

jonnitto:
Any progress on this or should we close this one?

PR #821, opened 2017-01-12
BUGFIX: Files::getFileContents broken in PHP 7.1

dfeyer:
This change set the offset to 0 (default value in `file_get_contents`.)

PHP 7.1 changed the behaviour of `file_get_contents` regarding
negative offset (support for negative offsets has been added.)

Without this change the sites `Root.fusion` is not loaded, so that
`Files::getFileContents` return an empty string. Tested on PHP 7.1.0
and 7.0.12 (worked without this fix).

### DIFF
--- a/Neos.Utility.Files/Classes/Files.php
+++ b/Neos.Utility.Files/Classes/Files.php
@@ -306,6 +306,7 @@ abstract class Files
      * @param integer $offset (optional) Offset where reading of the file starts.
      * @param integer $maximumLength (optional) Maximum length to read. Default is -1 (no limit)
      * @return mixed The file content as a string or FALSE if the file could not be opened.
+     * @throws FilesException
      * @api
      */
     public static function getFileContents($pathAndFilename, $flags = 0, $context = null, $offset = -1, $maximumLength = -1)
@@ -314,12 +315,13 @@ abstract class Files
             $flags = FILE_USE_INCLUDE_PATH;
         }
         try {
+            $offset = $offset === -1 ? null : $offset;
             if ($maximumLength > -1) {
                 $content = file_get_contents($pathAndFilename, $flags, $context, $offset, $maximumLength);
             } else {
                 $content = file_get_contents($pathAndFilename, $flags, $context, $offset);
             }
-        } catch (ErrorException $ignoredException) {
+        } catch (FilesException $ignoredException) {
             $content = false;
         }
         return $content;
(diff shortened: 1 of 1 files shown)


dfeyer:
Please merge this one before https://github.com/neos/flow-development-collection/pull/816

kitsunet:
Looks good so far, shouldn't we also change the default to `null` instead of `-1` then?

Neos.Utility.Files/Classes/Files.php
```diff
@@ -306,6 +306,7 @@ public static function copyDirectoryRecursively($sourceDirectory, $targetDirecto
      * @param integer $offset (optional) Offset where reading of the file starts.
      * @param integer $maximumLength (optional) Maximum length to read. Default is -1 (no limit)
      * @return mixed The file content as a string or FALSE if the file could not be opened.
+     * @throws FilesException
```
kdambekalns: Erm, it doesn't throw anything, right? The exception is caught in line 324…

Neos.Utility.Files/Classes/Files.php
```diff
@@ -306,6 +306,7 @@ public static function copyDirectoryRecursively($sourceDirectory, $targetDirecto
...
      * @api
      */
     public static function getFileContents($pathAndFilename, $flags = 0, $context = null, $offset = -1, $maximumLength = -1)
```
kdambekalns: `$offset` should be changed to `null`, I think (too).

Neos.Utility.Files/Classes/Files.php
```diff
@@ -306,6 +306,7 @@ public static function copyDirectoryRecursively($sourceDirectory, $targetDirecto
      * @param integer $offset (optional) Offset where reading of the file starts.
```
kdambekalns: Add hint that on PHP 7.1 negative offset is supported?

kdambekalns:
I adjusted this a bit, so now the default is `null` and the check for `-1` is gone. Otherwise _that_ would never have worked, in the (unlikely?) case of being needed.

kitsunet:
ok, apparently this is the fix for the `>` bug in setup/login that people are talking about in slack  (https://neos-project.slack.com/archives/neos-general/p1485696935005590) so apparently we need to backport to lower branches as well :/

kdambekalns:
@kitsunet Why backport? Does everything else work with 7.1? Then I'd say yes. But we promise 7.1 compatibility starting with 4.0 only, so otherwise…

kitsunet:
right, to be checked, probably not, but we should add respective php requirements then. currently you can install flow 3.x / neos 2.x with php 7.1 it just doesn't work.

kdambekalns:
Ok, if that is the only thing breaking it on 7.1 for older releases, we should backport. If not fix the constraint on PHP -> see #847

bwaidelich:
This also fixes PHP 7.1 support for Flow 3.3 / Neos 2.3 - I'll take care of the backport

PR #825, opened 2017-01-18
BUGFIX: Prevent invalid accessor detection in Fluid

kitsunet:
The standalone Fluid library tries to detect possible accessors
to use for extracting variable paths to speed up rendering
afterwards. Unfortunately this may result in wrong accessors being
used currently. The respective upstream bug is described here:

https://github.com/TYPO3/Fluid/issues/255

This change overwrites the method that extracts the accessors and
prevents any precached accessors. Instead Fluid will always use
the variable provider to determine access to variable paths.

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/Parser/TemplateParser.php
+++ b/Neos.FluidAdaptor/Classes/Core/Parser/TemplateParser.php
@@ -1,6 +1,11 @@
 <?php
 namespace Neos\FluidAdaptor\Core\Parser;
 
+use TYPO3Fluid\Fluid\Core\Parser\InterceptorInterface;
+use TYPO3Fluid\Fluid\Core\Parser\ParsingState;
+use TYPO3Fluid\Fluid\Core\Parser\Patterns;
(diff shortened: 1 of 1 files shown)


NamelessCoder:
When we get https://github.com/TYPO3/Fluid/pull/256 in we will make a bugfix release - so you don't have to live with this big class override but can instead bump the min Fluid dependency version (or tell people to update, we follow semantic versioning).

kdambekalns:
So TYPO3/Fluid#255 is fixed with TYPO3/Fluid#256 but the latest release is from before that. So still needed.

kitsunet:
@kdambekalns I would be in favor of merging now, so that this isn't a blocker for our release and we can always remove it in the next bugfix release, if at that time the bugfix release of fluid is out.

NamelessCoder:
A new release of Fluid is coming in a few minutes ;)

NamelessCoder:
Fluid 2.1.3 and 2.2.0 have been released, both of which include the necessary fix. 2.2.0 also contains new `f:variable` to assign variables in the template.

kitsunet:
merci! Fantastic :)

PR #826, opened 2017-01-18
BUGFIX: Require specific version of utility packages

kdambekalns:


kdambekalns:
So, this would need to be covered in our `set-dependencies.sh` for each release, right? Or can we require `self.version`?

PR #827, opened 2017-01-19
BUGFIX: Use fallback if exception handler configured wrong

kdambekalns:
In case the configured exception handler class dies not exist,
the `ProductionExceptionHandler` is used instead. This allows
(force-)flushing the caches to work in more cases.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
@@ -212,7 +212,11 @@ class Scripts
 
         $errorHandler = new \TYPO3\Flow\Error\ErrorHandler();
         $errorHandler->setExceptionalErrors($settings['error']['errorHandler']['exceptionalErrors']);
-        $exceptionHandler = new $settings['error']['exceptionHandler']['className'];
+        if (class_exists($settings['error']['exceptionHandler']['className'])) {
+            $exceptionHandler = new $settings['error']['exceptionHandler']['className'];
+        } else {
+            $exceptionHandler = new \TYPO3\Flow\Error\ProductionExceptionHandler();
+        }
         $exceptionHandler->injectSystemLogger($bootstrap->getEarlyInstance('TYPO3\Flow\Log\SystemLoggerInterface'));
         $exceptionHandler->setOptions($settings['error']['exceptionHandler']);
     }


kdambekalns:
FYI: The "failed build" was cancelled by me…

gerhard-boden:
(by "testing" i mean not just renaming the class, but rather how to produce the error that lead to this fix)

kitsunet:
Nice idea! I like

kdambekalns:
@gerhard-boden You could `composer install flownative/prettyexceptions`, run a request and then `composer remove flownative/prettyexceptions` - that should fail, if run another request.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -212,7 +212,11 @@ public static function initializeErrorHandling(Bootstrap $bootstrap)
 
         $errorHandler = new \TYPO3\Flow\Error\ErrorHandler();
         $errorHandler->setExceptionalErrors($settings['error']['errorHandler']['exceptionalErrors']);
-        $exceptionHandler = new $settings['error']['exceptionHandler']['className'];
+        if (class_exists($settings['error']['exceptionHandler']['className'])) {
+            $exceptionHandler = new $settings['error']['exceptionHandler']['className'];
```
aertmann: should it add an error to the log maybe? just to help debug when this issue occurs..

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -212,7 +212,11 @@ public static function initializeErrorHandling(Bootstrap $bootstrap)
...
+        } else {
+            $exceptionHandler = new \TYPO3\Flow\Error\ProductionExceptionHandler();
```
aertmann: should it add an error to the log maybe? just to help debug when this issue occurs..

aertmann:
good idea, added a suggestion though

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
kdambekalns: Well, is the logging set up at that point? Would need to check, but I guess it's not.

aertmann: no idea, but if not then ignore the suggestion I suppose

kitsunet: No logging  available here. We could inject the logger later and also check if a logger instance is set before logging.

kdambekalns: Well… I'd skip this. I mean, if you really expect some specific handler you will probably notice it's missing the first time it's (not) hit.

kitsunet: Agreed, can also be done anytime later.

PR #830, opened 2017-01-21
TASK: Some namespace fixes

albe:
Just two small fixes sponsored by phpstan/phpstan.


Neos.FluidAdaptor/Classes/Core/Parser/TemplateProcessor/EscapingFlagProcessor.php
```diff
@@ -50,7 +50,7 @@ public function preProcessSource($templateSource)
             return $templateSource;
         }
         if (count($matches) > 1) {
-            throw new Exception('There is more than one escaping modifier defined. There can only be one {escapingEnabled=...} per template.', 1407331080);
+            throw new \Exception('There is more than one escaping modifier defined. There can only be one {escapingEnabled=...} per template.', 1407331080);
```
aertmann: maybe it should be `\Neos\FluidAdaptor\Core\Exception` instead as an import?

albe:
Good points.

PR #831, opened 2017-01-22
!!! TASK: Remove deprecated support for relative uri paths

beheist:
Removed the long-deprecated compat flag for relative uri paths and the according code in the UriBuilder and UriBuilder test.

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
+++ b/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
@@ -63,12 +63,6 @@ class UriBuilder
     protected $createAbsoluteUri = false;
 
     /**
-     * @Flow\InjectConfiguration("compatibility.uriBuilder.createRelativePaths")
-     * @var boolean
-     */
-    protected $createRelativePaths = false;
-
-    /**
      * @var boolean
      */
     protected $addQueryString = false;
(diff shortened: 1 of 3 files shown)


albe:
Maybe something for 4.0 still? Or too hot now?

beheist:
Could be - shouldn't break anything, should it? Will gladly repoint to 4.0 if you'd like to.

aertmann:
get rid of it! 😀

albe:
The UriBuilderTest is understandable, but why does the FluidAdaptor test fail when rendering a simple variable into a section?!

Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
```diff
@@ -383,8 +354,6 @@ public function build(array $arguments = [])
         $httpRequest = $this->request->getHttpRequest();
         if ($this->createAbsoluteUri === true) {
             $uri = $httpRequest->getBaseUri() . $uri;
-        } elseif (!$this->createRelativePaths) {
```
albe: Guess this just needs to be changed to
```
} else {
    $uri = $httpRequest->getScriptRequestPath() . $uri;
```
instead of removing the whole elseif - at least if I'm correct in assuming the intended behaviour is to keep the previous default behaviour.

kitsunet:
The fluid test is the fluid problem that is fixed in 4.0 and not upmerged yet. Btw. this change should target 4.0 branch anyway. Can definitely NOT go into master.

albe:
Sorry for hijacking, but that was just too easy to fix quickly :)

PR #836, opened 2017-01-25
!!! TASK: Remove deprecated ``ResourcePublisher`` and pointer

kitsunet:
The old resource management pre Flow 3.0 used the ``ResourcePublisher``
as main service to get public URLs to resources and the ``ResourcePointer``
to keep track of unique resources. Both became unnecessary and were
deprecated with Flow 3.0 and are therefore removed with this major release.


kitsunet:
related https://github.com/neos/neos-development-collection/pull/1398

PR #838, opened 2017-01-25
!!! TASK: Rename [TYPO3][Flow][Security][Authentication]

### LINKED ISSUE #837: Rename [TYPO3][Flow][Security][Authentication] to Neos
The path used for the POST argument for Username Password tokens also needs to be adjusted to the vendor namespace change.

See #823

robertlemke:
This change adjusts the path used for the POST argument
used for authentication with username and password to the
new vendor namespace.

Any application - and especially its Fluid templates and
JavaScript - relying on the old path needs to be updated.

This change provides a core migration which carries out
these changes.

Resolves #837


aertmann:
a few more occurrences fixed in https://github.com/neos/flow-development-collection/pull/841

Issue #823, opened 2017-01-17
Remove remaining occurrences of "TYPO3"


bwaidelich:
There are currently still over 1000 occurrences of "typo3" in the flow-development-collection alone.
Some of them don't really matter, but for example the default authentication still expects `__authentication[TYPO3][Flow][Security][Authentication][Token][UsernamePassword]` request arguments (see [UsernamePassword Token](https://github.com/neos/flow-development-collection/blob/4.0/Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php#L34-L50) and [AbstractAuthenticationController](https://github.com/neos/flow-development-collection/blob/4.0/Neos.Flow/Classes/Security/Authentication/Controller/AbstractAuthenticationController.php#L48-L49)

PR #839, opened 2017-01-25
TASK: Remove TYPO3 occurrences in docs and tests

robertlemke:
This change removes some further TYPO3 occurrences and replaces
them by Neos.

Relates to #823

### DIFF
--- a/Neos.Flow/Tests/Unit/Security/Authorization/FilterFirewallTest.php
+++ b/Neos.Flow/Tests/Unit/Security/Authorization/FilterFirewallTest.php
@@ -92,9 +92,9 @@ class FilterFirewallTest extends UnitTestCase
                 'interceptor' => 'AccessGrant'
             ],
             [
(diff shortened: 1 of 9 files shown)


Neos.Flow/Tests/Unit/Security/Authorization/FilterFirewallTest.php
```diff
@@ -37,7 +37,7 @@ public function configuredFiltersAreCreatedCorrectlyUsingLegacySettingsFormat()
 
             if ($args[0] === 'URI') {
                 return 'mockPatternURI';
-            } elseif ($args[0] === 'TYPO3\\TestRequestPattern') {
+            } elseif ($args[0] === 'Neos\\TestRequestPattern') {
```
kdambekalns: The name of that test contains `UsingLegacySettingsFormat` - is that needed still? Also, it fails, so maybe there is a settings that needs to be adjusted (`Testing` context!?)

PR #843, opened 2017-01-27
TASK: Replace TYPO3 with Neos

robertlemke:
This replaces further occurrences of "TYPO3" with "Neos".

Addresses #823


robertlemke:
@kitsunet well spotted, thanks!

Issue #804, opened 2016-12-30
Core migrations need to take care of boot step rename


kitsunet:
Although it was never public API many packages reference boot step names in their Package.php

We need to make sure that the name from typo3.flow:* to neos.flow:* is covered with a migration, so that a string reference `typo3.flow:systemfilemonitor` is changed to `neos.flow:systemfilemonitor`.

PR #844, opened 2017-01-28
TASK: Add migration for renamed boot steps

kitsunet:
Replaces boot step names starting with `typo3.flow` with the renamed ones
starting with `neos.flow` in all PHP files.


robertlemke:
Resolves #804

PR #850, opened 2017-02-04
BUGFIX: Allow factories to build prototypes

daniellienert:
This fix allowes to build prototypes using factories by configuring
`scope: prototype` in the Objects.yaml
Before the object manager ignores the configuration and generates singletons everytime.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/ObjectManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/ObjectManager.php
@@ -187,6 +187,10 @@ class ObjectManager implements ObjectManagerInterface
         }
 
         if (isset($this->objects[$objectName]['f'])) {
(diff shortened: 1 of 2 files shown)


daniellienert:
I cannot imagine, that such a bug never broke things and was never discovered...

@dfeyer: This is the one breaking the refactored elasticsearch adapter.

albe:
absolute +1, but can we have a test please? This is too important to not be covered

kitsunet:
that was included in a change from me back in gerrit still. at that point I had no immediate (blocking) use case for it and no one wanted to discuss the problem, so I just dropped the change when we moved to GH...

daniellienert:
@albe: Added a unit test

kdambekalns:
I just fixed the same thing while still on 3.3.7 - made my day. But @daniellienert, you could have pointed me here!!one!!eleven!! ;)

daniellienert:
Next time just ask 😂 
Always the latest version you shall use ☝🏻

kdambekalns:
I asked a lot today in guild-search… :)

Issue #743, opened 2016-11-29
Package freezing is broken


dfeyer:
Currently `ReflectionService::freezePackageReflection()` is broken because is call `Package::getNamespace()` ... and this method does not exist any more.

We should adapt the code to use `Package::getNamespaces()`

dfeyer:
Regarding the package manager removal from the backend I think about it a bit more. What about to make our package manager (PackageManagerInterface) read only ? Maybe we can keep the activate/desactivate, that can be useful some times, but the rest is "broken" duplication of composer. What do you think ?

kitsunet:
https://github.com/neos/flow-development-collection/pull/851

PR #851, opened 2017-02-05
BUGFIX: Cleanup of reflection freezing

kitsunet:
The reflection freezing process still used a removed method from packages,
this has been refactored as the information needed is already available and
just needed to be cached.

Additionally a bit of clean up was done while looking at it.

Fixes: #743

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -106,11 +106,6 @@ class ReflectionService
     protected $annotationReader;
 
     /**
-     * @var ClassLoader
-     */
-    protected $classLoader;
-
-    /**
      * @var array
      */
     protected $availableClassNames = [];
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -2063,17 +2049,14 @@ public function freezePackageReflection($packageKey)
      * Filter an array of entries were keys are class names by being in the given package namespace.
      *
      * @param array $array
-     * @param string $packageNamespace
+     * @param string $packageKey
      * @return array
      */
-    protected function filterArrayByClassesInNamespace(array $array, $packageNamespace)
+    protected function filterArrayByClassesInNamespace(array $array, $packageKey)
```
albe: Guess this method could be slightly renamed now (`filterArrayByClassesInPackage` or sth), but not sure about the consequences... so let it be.

kdambekalns: Renamed the method, seemed too confusing, indeed.

kitsunet:
👍

Issue #848, opened 2017-02-01
StandaloneView in Neos.FluidAdaptor does not work correctly with request.


mkuiphuis:
### Description

The StandaloneView is used in Famelo.PDF to create PDF documents. Any older version of Flow which does not use the Neos.FluidAdaptor, the PDF's are created fine. After the update to Flow 4.0 the following error is displayed:

Call to a member function getRequest() on null. This happens here:

Packages/Framework/Neos.FluidAdaptor/Classes/View/StandaloneView.php (line 159)

#### Expected behavior

A generated PDF to be downloaded

#### Actual behavior

500 Internal Server Error with the error message as described above.
<img width="838" alt="fluidadaptor" src="https://cloud.githubusercontent.com/assets/12618307/22500377/2a55cebe-e8b8-11e6-85a0-a621d176ac28.png">

### Affected Versions

Neos: N/A

Flow: 4.0

kitsunet:
Ok, this is embarassing. I just fonud the error^^ PR coming soonish

PR #852, opened 2017-02-07
BUGFIX: Use request correctly in Fluid Views

kitsunet:
The refactoring of Fluid to the Adaptor left some places with invalid handling
of request and format inside of Views. The code couldn't work in it's current
state and is now adapted.

Fixes: #848


kitsunet:
Will see to add a test to cover those spots...

Issue #746, opened 2016-12-01
TYPO3Fluid: Document changed behavior of `AbstractConditionViewHelper`


bwaidelich:
### Description

In TYPO3Fluid the behavior of the  `AbstractConditionViewHelper` has been changed and it's hardcoded to evaluate a boolean argument `condition` by default.
In order to implement a custom condition VH one has to override the static `evaluateCondition` function.

### Steps to Reproduce

1. Create a simple custom ViewHelper, e.g.:

```php
class MyViewHelper extends AbstractConditionViewHelper
{
  public function render()
  {
    return $this->renderThenChild();
  }
}
```

2. Put it in a template

```html
{namespace x:My\Package\ViewHelpers}
<x:my>
  <f:then>THEN</f:then>
  <f:else>ELSE</f:else>
</x:my>
```

#### Expected behavior

```
THEN
```

#### Actual behavior

```
ELSE
```

(only renders "THEN" correctly on the first hit, before the template is compiled)

### Affected Versions

Flow 4.0  (using TYPO3Fluid).

bwaidelich:
Was this closed automatically?? This is not fixed by #784

robertlemke:
@bwaidelich Sorry, my bad. So this still needs to be fixed and there's no pull request for it yet, right? Is it a blocker for the 4.0 release?

bwaidelich:
@robertlemke This is just about *documenting* the changed behavior. I already added it to the release todos, but IMO we should keep the ticket open just in case

kdambekalns:
So, that new behaviour is this, right?

> In order to implement a custom condition VH one has to override the static evaluateCondition function.

kitsunet:
I would *try* to document some changes and best practices aobut fluid anyway, but so far I couldn't bring myself to it.
All in all render is not really best practice anymore and render with arguments even less.

bwaidelich:
> All in all render is not really best practice anymore and render with arguments even less.
I'm quite sad about that. I built many custom condition VHs. They improved usability/readability a lot and they only make sense with custom arguments really.
A way to make this less what-the-fucky for developers might be:
Make `evaluateCondition` *abstract* in the [AbstractConditionViewHelper](https://github.com/TYPO3/Fluid/blob/master/src/Core/ViewHelper/AbstractConditionViewHelper.php) and implement it in the [IfViewHelper](https://github.com/TYPO3/Fluid/blob/master/src/ViewHelpers/IfViewHelper.php).

[...]

PR #855, opened 2017-02-07
BUGFIX: Documented correct implementation of condition ViewHelpers

kitsunet:
With the introduction of standalone Fluid the correct way to implement
custom condition ViewHelpers has changed but this was not reflected in
the doc block of the ``AbstractConditionViewHelper``. This change finally
adds the missing explanation.

Fixes: #746

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractConditionViewHelper.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractConditionViewHelper.php
@@ -18,21 +18,22 @@ use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
 
 /**
  * This view helper is an abstract ViewHelper which implements an if/else condition.
- * @see Neos\FluidAdaptor\Core\Parser\SyntaxTree\ViewHelperNode::convertArgumentValue() to find see how boolean arguments are evaluated
+ *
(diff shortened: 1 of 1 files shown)


albe:
Good point @gerhard-boden

https://www.neos.io/download-and-extend/upgrade-instructions-2-3-3-0.html only mentions it by linking to the issue, which is rather technical around the problem space, rather than the solution.

gerhard-boden:
@albe Yea I think we should make the issue a little more prominent, it's on the very end of the notes. And people just smoke testing their sites after the upgrade might not notice the inversed behaviour (worst yet, the first load migh even be right)

kitsunet:
Ok, so add to the upgrade instructions on the website?

gerhard-boden:
@kitsunet That would be great! Should I try to formulate something or can you add the warning? Basically one or two sentences explaing the issue with the link to the issue should suffice.

kitsunet:
There is even a link to this in the upgrade instructions. I think that's fine them.

gerhard-boden:
@kitsunet I know, but like mentioned above I think it should be a little more prominent, since it can be a rather sneaky bug after upgrading

gerhard-boden:
It's a link to all open issues, it's marked as closed now and therefore gone. We should mention it some others way

gerhard-boden:
I changed the link to: 
https://github.com/neos/flow-development-collection/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3A%22P%3A%20Fluid(Adaptor)%22%20

At least this way all Fluid issues are visible.
I'm also writing a special section for mistakes that could easily happen while upgrading.

PR #856, opened 2017-02-07
BUGFIX: Use ``exec`` everywhere instead of ``system``

kitsunet:
Instead of using both, the Flow core uses the ``exec`` function throughout
now and the requirements documentation is updated accordingly.

Fixes: #634

### DIFF
--- a/Neos.Flow/Scripts/Migrations/Git.php
+++ b/Neos.Flow/Scripts/Migrations/Git.php
@@ -82,7 +82,7 @@ class Git
     public static function move($source, $target)
     {
         $result = 255;
-        system('git mv ' . escapeshellarg($source) . ' ' . escapeshellarg($target), $result);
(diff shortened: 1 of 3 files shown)


Neos.Flow/Scripts/Migrations/Git.php
```diff
@@ -94,9 +94,9 @@ public static function remove($fileOrDirectory)
     {
         $result = 255;
         if (is_dir($fileOrDirectory)) {
-            system('git rm -qr ' . escapeshellarg($fileOrDirectory), $result);
+            exec('git rm -qr ' . escapeshellarg($fileOrDirectory), $output, $result);
```
gerhard-boden: we tell git to be quiet and yet we handle the output?

kitsunet: Well, we want the status code, but as that is the third argument I also need to give the second (See reference about exec). I guess that's why system was used in the first place. But unifying the usage is worth even if we assign a useless variable.

Issue #478, opened 2016-09-01
On a clean installation a doctrine:migrationgenerate creates migrations for base tables


daniellienert:
## Problem Description

On a clean installation. When you run `doctrine:migrationgenerate` before a first run
of `doctrine:migrate` migrations for the base tables are generated.
## Suggested Behavior

The command `doctrine:migrationgenerate` checks if open migrations exist and tells the user to run `doctrine:migrate` first.

kitsunet:
That sounds reasonable. I would maybe add some "force" argument to override this, just in case. Maintenance and such...

dfeyer:
I agree with @kitsunet 

Good idea, but add ```--force``` to support side case

bwaidelich:
Additional idea: Without the `--force` flag Flow should never create an empty Migration but state sth like ```The database schema is up to date, no migration necessary```

PR #857, opened 2017-02-07
TASK: Prevent generating migrations if new migrations available

kitsunet:
It is generally undesired to create migrations if there are still migrations
that could be migrated. This is now prevented in the command controller
unless the ``force`` argument is set to circumvent this security measure.

Fixes: #478


malteriechmann:
👍  by reading

bwaidelich:
This is neat, but unfortunately it introduced a little regression: #976
I'll push a follow-up

bwaidelich:
-> #977

Issue #503, opened 2016-09-12
Missing documentation for package types


robertlemke:
The Composer package types which are currently supported by the `neos/composer-plugin` are not documented and [the existing example](https://github.com/neos/flow-development-collection/edit/master/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/PackageManagement.rst) is misleading. This may lead to very nerve-wrecking debug sessions (see below).

We need to update the respective documentation and explain which package types are supported. Additionally we need to clearly communicate, that **packages relying on the newer package type `neos-package` must state that they only work with Flow 3.2.x or later** since Flow 3.1.x still depends on `neos/composer-plugin` 1.0.

One bad example from my last weekend:

I installed [neos/eventstore](https://github.com/neos/Neos.EventStore) via Composer into an old application using Flow 3.1.5.

I then used the Event Sourced Repository in my code which led to:

```
The object "Neos\EventStore\EventStoreInterface" which was specified as a
property in the object configuration of object
"Flownative\Beach\Domain\Organization\Repository\OrganizationRepository"
(automatically registered class) does not exist.
```

I dumped the Composer autoloader. On the next `./flow` call I saw:

```
PHP Fatal error:  Uncaught TYPO3\Flow\Package\Exception\InvalidPackageStateException: Package "neos.eventstore" is already registered as "Neos.Eventstore". in /application/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php:723
```

So I tried a `./flow flow:package:rescan`:

```
PHP Fatal error:  Uncaught TYPO3\Flow\Package\Exception\InvalidPackageStateException: Package "neos.eventstore" is already registered as "Neos.Eventstore". in /application/Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Package/PackageManager.php:723
```

Same error message even for a `./flow flow:cache:flush --force`. I also tried removing `Configuration/PackageStates.php` until I realized that this might be due to an old Composer plugin. But since Flow 3.1 relies on 1.0 of the plugin, `composer update neos/composer-plugin` doesn't provide any new version.

So, everybody using these new package types: require Flow 3.2 or higher!

kitsunet:
Note though, it has nothing to do with the composer plugin, but merely with the PackageManager rewrite as the old version had hardcoded references to `typo3-flow-*` types all over the place. 
The composer plugin supported that from day one:
[...]

PR #862, opened 2017-02-07
BUGFIX: Add documentation about package types

kitsunet:
Adds a note about the package type prefixes support.

Fixes: #503


kitsunet:
I guess at some point we also want to review the setting mentioned. IF you install such a package via composer it would end up in the right location anway, setting or not. Basically the setting is just to define the location for a newly created package.

kitsunet:
merged as only documentation involved.

gerhard-boden:
@kitsunet Thanks for the infos, will use the custom folders for my internal composer packages starting by tomorrow. It just never crossed my mind that there could be a custom package folder outside "Application" or "Plugin" etc.

PR #864, opened 2017-02-08
FEATURE: Add wordCount Eel Helper

jonnitto:
Return number of words in a string. Good for calculation of reading duration of an article

### DIFF
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -571,9 +571,9 @@ class StringHelper implements ProtectedContextAwareInterface
     }
 
     /**
-     * Return numbers of words used in a string. Remove marks & digits and
+     * Return the count of words for a given string. Remove marks & digits and
      * flatten all kind of whitespaces (tabs, new lines and multiple spaces)
-     * This helper is good for calculate the reading time of an article.
+     * For example this helper can be utilized to calculate the reading time of an article.
      *
      * @param string $string The input string
      * @return integer Number of words
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -571,23 +571,19 @@ class StringHelper implements ProtectedContextAwareInterface
     }
 
     /**
-     * Return numbers of words used in a string
+     * Return numbers of words used in a string. Remove marks & digits and
+     * flatten all kind of whitespaces (tabs, new lines and multiple spaces)
+     * This helper is good for calculate the reading time of an article.
      *
      * @param string $string The input string
-     * @return integer Length of the string
+     * @return integer Number of words
      */
-    public function wordCount($unicode_string)
+    public function wordCount($unicodeString)
     {
-        // First remove all the punctuation marks & digits
-        $unicode_string = preg_replace('/[[:punct:][:digit:]]/', '', $unicode_string);
-        // Now replace all the whitespaces (tabs, new lines, multiple spaces) by single space
-        $unicode_string = preg_replace('/[[:space:]]/', ' ', $unicode_string);
-        // The words are now separated by single spaces and can be splitted to an array
-        // I have included \n\r\t here as well, but only space will also suffice
-        $words_array = preg_split('/[\n\r\t ]+/', $unicode_string, 0, PREG_SPLIT_NO_EMPTY);
+        $unicodeString = preg_replace('/[[:punct:][:digit:]]/', '', $unicodeString);
+        $unicodeString = preg_replace('/[[:space:]]/', ' ', $unicodeString);
 
-        // Now we can get the word count by counting array elments
-        return count($words_array);
+        return count(preg_split('/[\n\r\t ]+/', $unicodeString, 0, PREG_SPLIT_NO_EMPTY));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -571,6 +571,26 @@ public function length($string)
...
+     *
+     * @param string $string The input string
+     * @return integer Length of the string
+     */
+    public function wordCount($unicode_string)
+    {
+        // First remove all the punctuation marks & digits
+        $unicode_string = preg_replace('/[[:punct:][:digit:]]/', '', $unicode_string);
+        // Now replace all the whitespaces (tabs, new lines, multiple spaces) by single space
+        $unicode_string = preg_replace('/[[:space:]]/', ' ', $unicode_string);
```
gerhard-boden: can't we combine this replacement with the one above?
Good thing you documented everything nicely, but mostly we try to avoid too many inline comments. 

Maybe you can summarize those comments within the method description?

Neos.Eel/Tests/Unit/Helper/StringHelperTest.php
```diff
@@ -564,4 +564,26 @@ public function lengthWorks($input, $expected)
...
         $this->assertSame($expected, $result);
     }
+
+    public function wordCountExamples()
+    {
+        return [
+            'null' => [null, 0],
+            'empty' => ['', 0],
+            'non-empty' => ['Hello fri3nd, you\'re
+       looking          good tod@y!', 6],
```
gerhard-boden: fix indentation

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -571,6 +571,26 @@ public function length($string)
...
+        // The words are now separated by single spaces and can be splitted to an array
+        // I have included \n\r\t here as well, but only space will also suffice
+        $words_array = preg_split('/[\n\r\t ]+/', $unicode_string, 0, PREG_SPLIT_NO_EMPTY);
+
+        // Now we can get the word count by counting array elments
+        return count($words_array);
```
gerhard-boden: you can combine this line with one above and save one variable assignment.

kitsunet:
Didn't look into it in detail, but `Neos\Utility\Unicode\TextIterator::parseSubjectByWord` might eventually reduce the amount of code needed for this.

jonnitto:
@kitsunet I will look at it, thanks!

Neos.Eel/Classes/Helper/StringHelper.php
jonnitto: Line comment are now cleaned up. About the functions: In line 582 I remove marks & digits and in line 584 I replace all kind of whitespaces with an space.

Neos.Eel/Classes/Helper/StringHelper.php
jonnitto: Done.

Neos.Eel/Tests/Unit/Helper/StringHelperTest.php
jonnitto: Also done, thank you

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -571,6 +571,22 @@ public function length($string)
     }
 
     /**
+     * Return numbers of words used in a string. Remove marks & digits and
```
gerhard-boden: Better "Return the count of words"?

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -571,6 +571,22 @@ public function length($string)
     }
 
     /**
+     * Return the count of words for a given string. Remove marks & digits and
+     * flatten all kind of whitespaces (tabs, new lines and multiple spaces)
+     * For example this helper can be utilized to calculate the reading time of an article.
+     *
+     * @param string $string The input string
```
daniellienert: Var name should be `unicodeString` here.

jonnitto: You're right, thank you

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -571,6 +571,22 @@ public function length($string)
...
+     *
+     * @param string $unicodeString The input string
+     * @return integer Number of words
+     */
+    public function wordCount($unicodeString)
+    {
+        $unicodeString = preg_replace('/[[:punct:][:digit:]]/', '', $unicodeString);
+        $unicodeString = preg_replace('/[[:space:]]/', ' ', $unicodeString);
+
+        return count(preg_split('/[\n\r\t ]+/', $unicodeString, 0, PREG_SPLIT_NO_EMPTY));
```
kdambekalns: Why split on `\n\r\t `? You replace `[:space:]` earlier, and the PHP manual says:

> The space characters are HT (9), LF (10), VT (11), FF (12), CR (13), and space (32).

kdambekalns:
Oh, and if the behaviour of `Neos\Utility\Unicode\TextIterator::parseSubjectByWord` is considered a but, please make sure to open an issue. It is documented to behave like it does, but if that makes no sense we could a) remove or b) change that method…

kdambekalns:
The more I think about it… Let's leave `TextIterator` alone (it does what it's native counterpart does) and have this merged, since counting words is a rather common task (and definitely more common that counting sentences e.g.)

Issue #866, opened 2017-02-10
ClassLoader not respecting FLOW_PATH_TEMPORARY_BASE


ddoehring:
### Description

Proxy class cannot be found if you have an existing code cache and changed FLOW_PATH_TEMPORARY_BASE.

### Steps to Reproduce

1. You have an existing warm code cache
2. You modify the ENV variable FLOW_PATH_TEMPORARY_BASE
3. You add some new classes (e.g. ViewHelper)
4. Classes don't get loaded via proxy => dependency injection broken

#### Expected behavior

New classes compile new proxies and are therefore loaded via AvailableProxies from the new FLOW_PATH_TEMPORARY_BASE.

#### Actual behavior

The Classloader loads the AvailableProxies from the FLOW_PATH_DATA, which didn't change.

### Affected Versions

Flow:  probably all

kitsunet:
Is this Development or Production context or both?

hhoechtl:
We had the issue in the Development context

PR #867, opened 2017-02-10
BUGFIX: ClassLoader should load AvailableProxies from TEMPORARY_BASE

ddoehring:
If you have an old version of the AvailableProxyClasses in FLOW_PATH_DATA
and you change FLOW_PATH_TEMPORARY_BASE, always the old proxy classes
get loaded.

Resolves: #866


kitsunet:
Sounds sensible 👍 by reading. Will test later.

PR #872, opened 2017-02-21
BUGFIX: Make FileSystemTarget work with PackageStorage 

maltemuth:
When using the `FileSystemTarget` for static resources, Flow throws an error since `array_merge` is being called on a Generator. This is fixed by replacing the `array_merge` call with `\AppendIterator::append` calls.


maltemuth:
After sleeping on this, I think the Unit Test needs some work - but my I'm not experienced in PHPUnit (nor writing Unit Tests properly). I'd like to ask for help on that.

albe:
UnitTest is good for covering this. I would go for testing if the return type of the method is always a `Generator` or implements `Iterator` for that matter.
Basically, try to always test the interface of a method instead of it's implementation, i.e. verify that any possible input leads to a valid output according to type annotations.

maltemuth:
@albe, thanks for your input. I've cleaned up the test somewhat.

I'm using quite a lot of lines to set up the situation I found to be faulty - is that "okay" for a Unit Test?

After reading the test again, I don't really know what can be left out; so if you're okay with the unit test, the "help wanted" can be removed.

TYPO3.Flow/Users/maltemuth/src/vendor/neos/flow-development-3.2/Packages/Application/TYPO3.Flow/composer.json
```diff
@@ -0,0 +1,18 @@
...
+        "psr-0": {
+            "TYPO3\\Flow": "Classes/"
+        }
+    },
+    "extra": {
+        "neos": {
+            "package-key": "TYPO3.Flow"
+        }
+    }
+}
```
bwaidelich: What's this?

albe: Seems that accidentially sneaked in with 41656005ee9cf652165bd36f06b5aaceea5e89a6

albe:
> I'm using quite a lot of lines to set up the situation I found to be faulty - is that "okay" for a Unit Test?

Unfortunately, that's often unavoidable, especially with test subjects that do IO or use other dependencies heavily.

I'd be fine with leaving the test as is.

albe:
So if anyone else approves we can merge this I guess @bwaidelich @aertmann

PR #875, opened 2017-02-22
BUGFIX: Consider accounts with ``expirationDate`` in the future as active

### LINKED ISSUE #873: Account::isActive returns false when expirationDate is set
### Description

The Security\Account::isActive should return true if expirationDate is set in the past, but doesn't.

### Steps to Reproduce

Set the expiration date one minute into the future.
```
$account->setExpirationDate( (new \DateTime())->add(new \DateInterval("PT1M") );
```

#### Expected behavior

```
$account->isActive() // true, since it's not expired yet
```

#### Actual behavior

```
$account->isActive() // false - either $this->now is null when Injection didn't happen, or a Proxy (since it's lazy-injected); in either case the > check return false
```

### Affected Versions

Flow: 4.0.1, 3.2

maltemuth:
**What I did**

I fixed ``Account::isActive()`` behavior when an expirationDate is set.

**How I did it**

The "Now" injection is now non-lazy, thus the comparison of DateTime objects works.

**How to verify it**

Run the new functional test ``TYPO3.Flow/Tests/Functional/Security/AccountTest.php``

**Checklist**

Resolves #873

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Security/Account.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Security/Account.php
@@ -100,7 +100,7 @@ class Account
     protected $objectManager;
 
     /**
-     * @Flow\Inject
+     * @Flow\Inject(lazy=false)
      * @var Now
      */
     protected $now;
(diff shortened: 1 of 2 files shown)


aertmann:
tests don't pass

maltemuth:
@aertmann it does; Now is the timestamp of object instantiation and not equal to \DateTime("now") - this is why tests failed if the suite takes longer than one minute to run.

However, lazy-loading had the side effect of breakage.

PR #883, opened 2017-03-08
FEATURE: Skip unknown properties for ObjectConverter

robertlemke:
This change enables the ObjectConverter (for simple PHP objects)
to acknowledge the `skipUnknownProperties` flag of the property
mapping configuration and thus ignore properties from the source
which don't exist in the target.

If a source array contains properties "foo" and "bar" and the
target class constructor only contains "foo", the array can now
be mapped if `skipUnknownProperties` is set in the respective
property mapping configuration.

### DIFF
--- a/Neos.Flow/Classes/Property/PropertyMapper.php
+++ b/Neos.Flow/Classes/Property/PropertyMapper.php
@@ -187,6 +187,9 @@ class PropertyMapper
             }
 
             $targetPropertyType = $typeConverter->getTypeOfChildProperty($targetType, $targetPropertyName, $configuration);
+            if ($targetPropertyType === null) {
+                continue;
+            }
 
             $subConfiguration = $configuration->getConfigurationFor($targetPropertyName);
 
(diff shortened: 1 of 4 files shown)


robertlemke:
This feature helps us with mapping of events from the event stream: https://github.com/neos/Neos.EventSourcing/issues/94

Neos.Flow/Classes/Property/TypeConverter/ObjectConverter.php
```diff
@@ -160,6 +160,11 @@ public function getTypeOfChildProperty($targetType, $propertyName, PropertyMappi
                 }
             }
         }
+
+        if ($configuration->shouldSkipUnknownProperties() || $configuration->shouldSkip($propertyName)) {
```
aertmann: only the first part check is covered by a test

Neos.Flow/Classes/Property/PropertyMapper.php
```diff
@@ -187,6 +187,9 @@ protected function doMapping($source, $targetType, PropertyMappingConfigurationI
             }
 
             $targetPropertyType = $typeConverter->getTypeOfChildProperty($targetType, $targetPropertyName, $configuration);
+            if ($targetPropertyType === null) {
```
aertmann: a test for this case would be good

robertlemke:
Thanks a lot @aertmann for pushing me to write more tests. Here they are … 👻

Issue #884, opened 2017-03-08
PdoBackend for caches does not create the needed tables with mysql


jhinzmann:
### Description

The PDO-Backend creates the caching-tables only if sqlite is used.
When mysql is used, no tables are generated and the caching does not work.
If sqlite is used, the tables are generated from `TYPO3.Flow/Resources/Private/Cache/SQL/DDL.sql` by calling `PdoBackend->createCacheTables()` in `PdoBackend->connect()`

After creating the tables manually from the mentioned DDL.sql the caching does work.
The documentation does not mention this, therefore I guess it's a bug.

Maybe this should go in another issue, but when creating the tables manually, it's a good idea to not go with `text` (as in the DDL.sql) for `content` since the content of one cache entry can easily be longer than 64kb.
This is no problem in sqlite because there are no such limits on the columns.

### Steps to Reproduce

1. add a Caches.yaml with the following config:
```TYPO3_TypoScript_Content:
  backend: TYPO3\Flow\Cache\Backend\PdoBackend
  backendOptions:
    dataSourceName: 'mysql:host=XXX;dbname=XXX'
    username: 'XXX'
    password: 'XXX'`
```
2. Open the Frontend or Backend or try to flush the cache

#### Expected behavior

The caching tables should be created if needed and should be used for storing the caches

#### Actual behavior

The caching tables are not created.
The Frontend and Backend throw:
`SQLSTATE[42S02]: Base table or view not found: 1146 Table 'XXX.cache' doesn't exist`

Trying to flush the caches throws:
```
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'XXX.tags' doesn't exist

  Type: PDOException
  Code: 42S02
  File: Packages/Framework/TYPO3.Flow/Classes/TYPO3/Flow/Cache/Backend/PdoBackend.php
  Line: 217
```
### Affected Versions

This bug occurred with
Neos: 2.3
Flow: 3.3
but by checking the PdoBackend implementation I guess this is still the case for Flow 4.0 and was the case since the PDO-Backend was introduced (https://github.com/neos/flow-development-collection/commit/92ad0c115da5dca736d9a1cf6cfd138d70f2403d#diff-d0f3ffadf243879866b8b49abddce315R345)

aertmann:
Thanks for opening, been surprised by this too. Can't really be considered a bug though, but definitely a missing feature and an undocumented manual step.

jhinzmann:
[...]

PR #885, opened 2017-03-09
TASK: Add documentation for using PdoBackend with MySQL

jhinzmann:
When using the PdoBackend with MySQL for caching,
the needed caching tables have to be created manually.
This adds a hint on creating the tables to the documentation
on the PdoBackend.

Solving the documentation part of neos/flow-development-collection#884

This depends on #886

### DIFF
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Caching.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Caching.rst
@@ -356,6 +356,15 @@ to clean up hard disk space or memory.
 	not with a capable database like Oracle. We appreciate any feedback for real life use
 	cases of this cache.
 
+.. warning::
+
+	When using MySQL with the PDO backend, you have to create the needed caching tables manually.
+	The table definition for MySQL is the same as the definition for SQLite, which can be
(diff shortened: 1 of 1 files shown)


kdambekalns:
@jhinzmann Thanks! Tweaked it a bit, good to go IMHO.

If not yet existing: would you open an issue so we can adjust the length for the content in the DDL? Or maybe you can open a PR yourself? ;)

jhinzmann:
@kdambekalns just to clarify:

the new pullrequest should also go into 3.2, change the column from ``TEXT`` to  ``MEDIUMTEXT`` and remove the warning from the documentation?
I guess 16MB for one cache entry should be enough?
Or should the limitation still be mentioned?

albe:
Thanks for this! I guess together with #886 the last part can be removed again, but a notice to the limit of the content column doesn't hurt IMO, even if it feels relatively safe with 16MB.

jhinzmann:
@albe So #886 should be merged first and then I should updated this pull request?
Or should both pull requests be merged and I open a new one to adjust the warning text?

albe:
I think it's ok to just change this one as if #886 were already merged. You could add a note to the PR message that this depends on #886 to be sure that whoever wants to merge should make sure that the other one is also merged.

PR #886, opened 2017-03-09
TASK: Increase max-length for cache entries in PdoBackend

jhinzmann:
When not using SQLite for the `PdoBackend` and manually creating
the cache tables by using the `DDL.sql`, the content column of the
cache table with type `TEXT` could easily lead to cropped
caching entries because of the length restriction.
Using `MEDIUMTEXT` practically prevents this issue.

See also: neos/flow-development-collection#884 and
neos/flow-development-collection#885


aertmann:
can't seem to find any information about max lengths for `TEXT` fields in SQLite, neither any definition of a `MEDIUMTEXT` column type.

http://stackoverflow.com/questions/33910827/is-sqlite-database-support-mediumtext-like-mysql-in-android

kdambekalns:
@aertmann (and anyone else interested)

- size limits for SQLite: https://www.sqlite.org/limits.html
- `MEDIUMTEXT` works on SQLIte, see https://www.sqlite.org/datatype3.html#determination_of_column_affinity (which is a rather weird concept, IMHO, with all the assumptions involved…)

PR #892, opened 2017-03-15
BUGFIX: Detect base uri in cli mode

daniellienert:
When trying to render links to assets when
the request is triggered using a command controller
the base uri cannot be detected.

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
+++ b/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php
@@ -12,6 +12,7 @@ namespace Neos\Flow\ResourceManagement\Target;
  */
 
 use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Cli\CommandRequestHandler;
 use Neos\Flow\Core\Bootstrap;
 use Neos\Flow\Http\HttpRequestHandlerInterface;
 use Neos\Flow\Log\SystemLoggerInterface;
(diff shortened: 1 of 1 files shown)


daniellienert:
Hey, 
currently trying to render Fusion objects using a command controller. When It comes to image tag rendering, the filesystem target is not able to detect a base Uri for a request with a CommandRequestHandler. 

I assume this might not be the best way to solve it. So feedback and ideas are much appreciated.

kdambekalns:
As mentioned in Slack: What about using a different context for CLI and set the `baseUri` as needed like `FLOW_CONTEXT=Production/Cli ./flow dlienert:renderstuff`?

daniellienert:
Also from slack: @kdambekalns: I thought about that. But it would complicate things. In my current situation, the generation of an image URL fails silently with just no URL is generated. I would like to have a better default - which would be to have a relative URL when no baseUri is given

aertmann:
Hey, not sure this is a good approach. As mentioned on Slack, Flow lacks handling of relative resource URIs and this is mostly a workaround as I see it.

I'd suggest to solve it using configuration and maybe a context if the base URI cannot be shared for all contexts.

However the best solution would of course be to add support for relative resource URIs to Flow's resource handling.

bwaidelich:
I don't understand how the change is supposed to fix the bug. IMO the only reliable way is to configure the `http.baseUri` otherwise lots of things, like generated links in Fluid Templates (i.e. email templates) won't work either.
So does this break even if the HTTP baseUri is configured?
And could you provide a little more context?

kdambekalns:
Closing this for now… @daniellienert – feel free to reopen if you disagree and want to continue the discussion.

PR #893, opened 2017-03-17
!!! BUGFIX: fix Entity Security depending on more than just roles

skurfuerst:
## Problem Description

If you had used entity security and wanted to secure entities not just
based on the user's role, but on some property of the user (like the company
he belongs to), entity security did not work properly together with doctrine
caching.

## Visible Error

If you had Entity Constraints which did not only rely
on different user roles, but needed the domain model to build up SQL 
constraints (through entity security), the query
cache in production cached the SQL queries "wrongly", so you were able to
see data from other accounts.

## Bugfix

Inside the Doctrine CacheAdapter, we already namespace the full Doctrine
cache with the SecurityContext's ContextHash. However, only roles were
the source of the ContextHash.

We now make it possible to add custom sources to the calculation of the
ContextHash, by making aop.globalObjects able to contribute to the context
hash calculation if they implement CacheAwareInterface.

## Custom SQL Filters / Root Cause Analysis

It is not effective to call "setParameter" in the addFilterConstraint method
of an SqlFilter; as this will interfere with Doctrine's query caching:
Doctrine tries to avoid building queries as much as possible. It also
takes the applied SqlFilters *and their parameters* into account when
building cache identifiers, and if it has a matching cache entry, it will
not call "addFilterConstraint". Thus, "setParameter" must be called as early
as possible; way before the "addFilterConstraint" method might be called.

We did not manage to find the fully right spot to do it inside the Framework,
that's why we chose the more minimally-invasive solution here.

## Breakingness

This change is only breaking if:

- You used Entity Security AND custom context objects registered in
  `TYPO3.Flow.aop.globalObjects`:

  In this case, your global objects should very likely implement
  CacheAwareInterface.

- You used custom SQL Filters:

  Basically you are not allowed to call `setParameter` inside `addFilterConstraint`;
  but setParameter must be called *before* the SQL query is actually executed.
  Currently, there's no standard Doctrine way to provide this; so you manually
  can receive the filter instance from `$entityManager->getFilters()->getEnabledFilters()`
  and call `setParameter()` then.

  Alternatively, you can influence the Doctrine caching by registering a global context
  object in `TYPO3.Flow.aop.globalObjects` and use it to provide additional identifiers
  for the caching; effectively seggregating the Doctrine cache some more.

## Thanks

Thanks to Kay Strobach for sponsoring working on this issue!

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
@@ -19,7 +19,6 @@ use Doctrine\DBAL\Logging\SQLLogger;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManager;
-use Doctrine\ORM\Query\Filter\SQLFilter;
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Cache\CacheManager;
 use TYPO3\Flow\Configuration\Exception\InvalidConfigurationException;
@@ -59,11 +58,6 @@ class EntityManagerFactory
      * @var array
      */
     protected $settings = [];
-
-    /**
-     * @var SQLFilter[]
-     */
-    protected $enabledFilterInstances = [];
 
     /**
      * Injects the Flow settings, the persistence part is kept
@@ -152,7 +146,7 @@ class EntityManagerFactory
         if (isset($this->settings['doctrine']['filters']) && is_array($this->settings['doctrine']['filters'])) {
             foreach ($this->settings['doctrine']['filters'] as $filterName => $filterClass) {
                 $config->addFilter($filterName, $filterClass);
-                $this->enabledFilterInstances[$filterName] = $entityManager->getFilters()->enable($filterName);
+                $entityManager->getFilters()->enable($filterName);
             }
         }
 
@@ -161,17 +155,6 @@ class EntityManagerFactory
         }
 
         return $entityManager;
-    }
-
-    /**
-     * Returns the active and instanciated SQL Filters. Needed to lateron call setParameter() on custom
-     * SQL filter instances before doing queries.
-     *
-     * @return SQLFilter[] the SQL filters, indexed by filter name as given in Settings.yaml
-     */
-    public function getEnabledFilterInstances()
-    {
-        return $this->enabledFilterInstances;
     }
 
     /**
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
@@ -19,6 +19,7 @@ use Doctrine\DBAL\Logging\SQLLogger;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\Query\Filter\SQLFilter;
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Cache\CacheManager;
 use TYPO3\Flow\Configuration\Exception\InvalidConfigurationException;
@@ -58,6 +59,11 @@ class EntityManagerFactory
      * @var array
      */
     protected $settings = [];
+
+    /**
+     * @var SQLFilter[]
+     */
+    protected $enabledFilterInstances = [];
 
     /**
      * Injects the Flow settings, the persistence part is kept
@@ -146,7 +152,7 @@ class EntityManagerFactory
         if (isset($this->settings['doctrine']['filters']) && is_array($this->settings['doctrine']['filters'])) {
             foreach ($this->settings['doctrine']['filters'] as $filterName => $filterClass) {
                 $config->addFilter($filterName, $filterClass);
-                $entityManager->getFilters()->enable($filterName);
+                $this->enabledFilterInstances[$filterName] = $entityManager->getFilters()->enable($filterName);
             }
         }
 
@@ -155,6 +161,17 @@ class EntityManagerFactory
         }
 
         return $entityManager;
+    }
+
+    /**
+     * Returns the active and instanciated SQL Filters. Needed to lateron call setParameter() on custom
+     * SQL filter instances before doing queries.
+     *
+     * @return SQLFilter[] the SQL filters, indexed by filter name as given in Settings.yaml
+     */
+    public function getEnabledFilterInstances()
+    {
+        return $this->enabledFilterInstances;
     }
 
     /**
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
@@ -12,8 +12,10 @@ namespace TYPO3\Flow\Security;
  */
 
 use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Cache\CacheAwareInterface;
 use TYPO3\Flow\Log\SecurityLoggerInterface;
 use TYPO3\Flow\Mvc\RequestInterface;
+use TYPO3\Flow\Object\ObjectManagerInterface;
 use TYPO3\Flow\Security\Authentication\TokenInterface;
 use TYPO3\Flow\Security\Policy\Role;
 use TYPO3\Flow\Mvc\ActionRequest;
@@ -191,6 +193,20 @@ class Context
      * @var string
      */
     protected $contextHash = null;
+
+    /**
+     * @Flow\Inject
+     * @var ObjectManagerInterface
+     */
+    protected $objectManager;
+
+    /**
+     * Array of registered global objects that can be accessed as operands
+     *
+     * @Flow\InjectConfiguration("aop.globalObjects")
+     * @var array
+     */
+    protected $globalObjects = [];
 
     /**
      * Inject the authentication manager
@@ -837,7 +853,18 @@ class Context
             $this->initialize();
         }
         if ($this->contextHash === null) {
-            $this->contextHash = md5(implode('|', array_keys($this->getRoles())));
+            $contextHashSoFar = implode('|', array_keys($this->getRoles()));
+
+            $this->withoutAuthorizationChecks(function () use (&$contextHashSoFar) {
+                foreach ($this->globalObjects as $globalObjectsRegisteredClassName) {
+                    if (is_subclass_of($globalObjectsRegisteredClassName, CacheAwareInterface::class)) {
+                        $globalObject = $this->objectManager->get($globalObjectsRegisteredClassName);
+                        $contextHashSoFar .= '<' . $globalObject->getCacheEntryIdentifier();
+                    }
+                }
+            });
+
+            $this->contextHash = md5($contextHashSoFar);
         }
         return $this->contextHash;
     }
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
@@ -496,6 +496,20 @@ Filter System.
   ``@Flow\Proxy(false)`` to your filter class to prevent Flow from building a proxy,
   which causes this error.
 
+.. warning:: Custom SqlFilter implementations - watch out for data privacy issues!
+
+  If using custom SqlFilters, you have to be aware that the SQL filter is cached by doctrine, thus your SqlFilter might
+  not be called as often as you might expect. This may lead to displaying data which is not normally visible to the user!
+
+  Basically you are not allowed to call `setParameter` inside `addFilterConstraint`; but setParameter must be called *before*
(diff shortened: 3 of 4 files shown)


albe:
Uff, quite some fiddling in this, so yes, definitely needs thorough discussion and possibly some explanation to the chosen solution. But thanks you both for deep diving into this mess!

TYPO3.Flow/Classes/TYPO3/Flow/Security/GlobalObjectStateInterface.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Created by IntelliJ IDEA.
+ * User: sebastian
+ * Date: 17.03.17
+ * Time: 09:58
+ */
+
+namespace TYPO3\Flow\Security;
+
+interface GlobalObjectStateInterface
```
aertmann: this doesn't seem to be used by any classes, is that on purpose?

TYPO3.Flow/Classes/TYPO3/Flow/Security/GlobalObjectStateInterface.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * Created by IntelliJ IDEA.
+ * User: sebastian
+ * Date: 17.03.17
+ * Time: 09:58
+ */
```
aertmann: this should be removed

TYPO3.Flow/Classes/TYPO3/Flow/Security/GlobalObjectStateInterface.php
kaystrobach: see https://github.com/neos/flow-development-collection/pull/893/commits/87c341f54624844abc7fee9ee73e01231db8cdb5#diff-51acb520957e5e74e17bfd768b12d9d2R875 it is important, you want to have your own global object to have the ability to add a context hash ... very important, i think @skurfuerst can easily demonstrate that with the example project given 😄

TYPO3.Flow/Classes/TYPO3/Flow/Security/GlobalObjectStateInterface.php
kaystrobach: agree

TYPO3.Flow/Classes/TYPO3/Flow/Security/GlobalObjectStateInterface.php
aertmann: figured as much, however this should be documented somehow in https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Security.rst

kaystrobach: agree, I'm only able to fork an add another PR, maybe you can grant me the right - or add the todos directly to the PR description as tasklist?

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/DispatchComponent.php
```diff
@@ -79,6 +87,14 @@ public function handle(ComponentContext $componentContext)
...
         $this->securityContext->setRequest($actionRequest);
 
+        // TODO: un-nice position; and not nice to inject the entityManagerFactory...
+        $enabledFilterInstances = $this->entityManagerFactory->getEnabledFilterInstances();
+        foreach ($enabledFilterInstances as $filter) {
+            if ($filter instanceof EarlyParameterInitializationSqlFilterInterface) {
+                $filter->initializeParameters();
+            }
+        }
+
```
skurfuerst: maybe Signal: "SecurityContextWasInitialized"

TYPO3.Flow/Classes/TYPO3/Flow/Security/GlobalObjectStateInterface.php
skurfuerst: maybe use CacheAwareInterface?

TYPO3.Flow/Classes/TYPO3/Flow/Security/Authorization/Privilege/Entity/Doctrine/SqlFilter.php
```diff
@@ -37,6 +37,26 @@ class SqlFilter extends DoctrineSqlFilter
...
+        }
+        if (!$this->securityContext->isInitialized()) {
+            if (!$this->securityContext->canBeInitialized()) {
+                return '';
+            }
+            $this->securityContext->initialize();
+        }
+
+        // This is needed to include the current context of roles into query cache identifier
+        $this->setParameter('__contextHash', $this->securityContext->getContextHash(), 'string');
```
skurfuerst: remove setParameter in "addFilterConstraint" below

skurfuerst: done

skurfuerst:
Hey @albe @aertmann @kaystrobach @bwaidelich @foerthner @kdambekalns ,

I finally polished the Pull Request, and I think it is quite ready to go in now. See the updated pull request description please :)

NOTE: As this is such a critical change, I backported it to the earliest possible version where the problem occurs, and not just the currently-working ones.

skurfuerst:
@kitsunet actually I am not sure on this one - I just programmed it as defensively as possible so far.

kitsunet:
Fair enough, I just thought that this _might_ be a security issue later on if we don't fail there. But not a blocker right now.

TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
```diff
@@ -837,7 +864,18 @@ public function getContextHash()
...
         }
         if ($this->contextHash === null) {
-            $this->contextHash = md5(implode('|', array_keys($this->getRoles())));
+            $contextHashSoFar = implode('|', array_keys($this->getRoles()));
+
+            $this->withoutAuthorizationChecks(function () use (&$contextHashSoFar) {
+                foreach ($this->globalObjects as $globalObjectsRegisteredClassName) {
+                    $globalObject = $this->objectManager->get($globalObjectsRegisteredClassName);
+                    if ($globalObject instanceof CacheAwareInterface) {
+                        $contextHashSoFar .= $globalObject->getCacheEntryIdentifier();
```
bwaidelich: I'd suggest to separate the individual context hash segments with some character in order to reduce the risk of clashes

TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
```diff
@@ -837,7 +864,18 @@ public function getContextHash()
             $this->initialize();
         }
         if ($this->contextHash === null) {
-            $this->contextHash = md5(implode('|', array_keys($this->getRoles())));
+            $contextHashSoFar = implode('|', array_keys($this->getRoles()));
+
+            $this->withoutAuthorizationChecks(function () use (&$contextHashSoFar) {
+                foreach ($this->globalObjects as $globalObjectsRegisteredClassName) {
+                    $globalObject = $this->objectManager->get($globalObjectsRegisteredClassName);
```
bwaidelich: do we have to initialize the object in order to check it's interface?

skurfuerst: good point, we can use class_implements

TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
skurfuerst: ok!

albe: Good point

TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
kitsunet: 👍

TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
kitsunet: 👍

TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
albe: You mean check if it's a lazy loading proxy? AFAIS `objectManager->get()` always returns a real class instance, lazy loading only in effect for injected properties. Or am I wrong?

bwaidelich: No, to check whether the class implements `CacheAwareInterface`

TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
bwaidelich: Cool. Let's just use a different one than we already use for the role identifiers to make super sure ;)

TYPO3.Flow/Classes/TYPO3/Flow/Security/Context.php
albe: Yep, I thought you pointed at `instanceof` not returning correct result because the object might be a proxy, but it's just about avoiding to load the object at all :)

skurfuerst:
fixed your comments!

skurfuerst:
I think I can merge this now, can't I?

skurfuerst:
@bwaidelich @albe

bwaidelich:
Did you force-push, there's only one commit!?

bwaidelich:
Looks good to me.
(please keep the single commits until merged or one has to re-review the complete changeset for every follow-up ;))

albe:
Looks good now, but tests need adjustment it seems

skurfuerst:
FYI: Actually the tests hint at some missing cases; though for me it is still hard to track them all down. working on it!

skurfuerst:
Once more discussed with @bwaidelich and we're coming up with a different/other solution which is less breaking. I'll push something soon ;)

skurfuerst:
@bwaidelich @albe @kitsunet @kdambekalns @foerthner 

## Finally.

I rewrote the patch once again after talking to Bastian, seeing that the tests failed and I did not cover all cases. The new patch should be a lot less breaking and should have no unintended side-effects (which I was not entirely sure in the version before). Now I feel a lot better actually merging this :-)

See the updated message above for the latest description; Sorry for the force-push; but I changed the mechanism completely so I thought actually reviewing it from scratch  is easier to follow than with diff-commits.

Let's get this in :heart:

Sebastian

albe:
So, isn't the `setParameter('__contextHash', ...)` call supposed to be moved anymore? I thought that was ill-placed?

Since the issue is rather concerning I'd be inclined to have this merged, though I'd highly like to have some tests for the functionality.

kitsunet:
Ok, so this is rather scaled down to only add the possibility to move setParameter calls out AND add global objects to the context hash. Still makes sense to me.

skurfuerst:
@albe:

"So, isn't the setParameter('__contextHash', ...) call supposed to be moved anymore? I thought that was ill-placed? "

-> this call can technically be completely removed; but it will break our existing functional tests... However functionality wise it is not needed anymore...

@kitsunet: Exactly!

skurfuerst:
@albe tests would be nice, though I am quite unsure on how to write them and the (at least) functional security test setup is so brittle...

I spent now so many hours on that; I'd finally like to get it in (if it makes sense for you); upmerge, release, and do an extra release announcement to make people aware of it :)

kitsunet:
Agreed, test would be super nice and we should get one (maybe lets open an issue just for that) but this fix is important, so instead of taking longer let's get it in.

TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
```diff
@@ -146,7 +152,7 @@ public function create()
         if (isset($this->settings['doctrine']['filters']) && is_array($this->settings['doctrine']['filters'])) {
             foreach ($this->settings['doctrine']['filters'] as $filterName => $filterClass) {
                 $config->addFilter($filterName, $filterClass);
-                $entityManager->getFilters()->enable($filterName);
+                $this->enabledFilterInstances[$filterName] = $entityManager->getFilters()->enable($filterName);
```
albe: After working again a bit on #462 I noticed one thing: there might be cases where some custom code adds filters outside the configured filters in `Settings.yaml`, at least once that change gets merged. Hence, it is/becomes unreliable to keep track of the enabled filters ourself.

Luckily, Doctrine already provides an API for this:
`$entityManager->getFilters()->getEnabledFilters();`

kitsunet: Ok, but the configured filters still need to be enabled.

kaystrobach:
It's really awesome to see your effort fixing that issue. Thank you guys!

albe:
Tried to push a change to remove the enabledFilterInstances again and document to use doctrine's method instead, but have no rights to do so @skurfuerst :/

TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
skurfuerst: thanks, fixed :)

albe:
:+1: :)

PR #894, opened 2017-03-18
FEATURE: Allow for dynamic label overrides

nezaniel:
This replaces the previous physical file model with the file model defined by the XLIFF standard. 

Files IDs are no longer defined by their location in the file system but by their _product-name_ and _orginal_ XLIFF attributes. The location serves as a fallback to prevent this from being a breaking change.

The XLIFF file provider therefore parses all registered packages and in addition the Data/Translations folder in search of file entries in .xlf files.
The translation file monitor is now also registered for said Data/Translations folder to enable automatic cache clearing.

In preparation for support of XLIFF version 2, the XLIFF parser has been created in a v12 namespace and the class for holding the file data is modeled as an adapter that should be able to handle both formats.

Overriding labels is now be as easy as putting a valid XLIFF file with defined _product-name_ (Package) and _orginal_ (Source) in any translation folder.

I targeted Flow 4.1 since this is a breaking change for Neos, which before used the non-API XLIFF parser directly which is no longer available.

Update: Re-Added old XLIFF parser for backwards compatibility and flagged it as deprecated. We could now also target 4.0 if we wanted to


Neos.Flow/Classes/I18n/Xliff/XliffParser.php
```diff
@@ -28,67 +26,26 @@
  * @throws Exception\InvalidXliffDataException
  * @see http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html [1]
  * @see http://docs.oasis-open.org/xliff/v1.2/xliff-profile-po/xliff-profile-po-1.2-cd02.html#s.detailed_mapping.tu [2]
+ * @deprecated use \Neos\Flow\I18n\Xliff\V12\XliffParser instead. Will be removed in Flow 5.0
```
gerhard-boden: Can we already remove this in the next major, when we deprecate something in a minor release? Or do we have to wait for 6.0 to finally remove it?

Neos.Flow/Classes/I18n/Xliff/XliffParser.php
```diff
@@ -28,67 +26,26 @@
...
+class XliffParser extends V12\XliffParser
 {
     /**
      * Returns array representation of XLIFF data, starting from a root node.
      *
      * @param \SimpleXMLElement $root A root node
      * @return array An array representing parsed XLIFF
      * @throws Exception\InvalidXliffDataException
-     * @todo Support "approved" attribute
+     * @deprecated use \Neos\Flow\I18n\Xliff\V12\XliffParser::doParsingFromRoot instead
```
gerhard-boden: Can we already remove this in the next major, when we deprecate something in a minor release? Or do we have to wait for 6.0 to finally remove it?

Neos.Flow/Classes/I18n/Xliff/XliffParser.php
albe: AFAIK it's ok to remove in 5.0, there is still a LTS before that, so there should be no problem for users to avoid this deprecation

albe:
Looks like the test job hang up, I restarted it. Let's see if that works out.

albe:
Ok, nope. That wasn't the issue - travis just somehow fails to show the full output of the build runs. But there was something about the `ConfigurationValidationTest` failing

Neos.Flow/Classes/I18n/Xliff/XliffParser.php
kdambekalns: The class has no `@api` flag, so in theory we can adjust is as we like. So, yes, this is fine.

Neos.Flow/Classes/I18n/Xliff/XliffParser.php
kdambekalns: Same here, the class has no `@api` flag, the method has nop `@api`flag and is protected, so in theory we can adjust is as we like. So, yes, this is fine.

gerhard-boden:
@albe good catch! Totally overlooked the old namespace

kdambekalns:
@albe What do you mean with

>  travis just somehow fails to show the full output of the build runs.

For me that's all there… at e.g. https://travis-ci.org/neos/flow-development-collection/jobs/263063997

gerhard-boden:
@kdambekalns on the first run I also had the issue with the missing output. Stopped shortly after cloning the repo

kdambekalns:
> Stopped shortly after cloning the repo

Ah, well, could be that the job was aborted because a new push was seen.

gerhard-boden:
YES! 💚 again. Thanks guys!

PR #895, opened 2017-03-20
Bugfix accessing empty array

DanielSiepmann:
As ``$this->typeConverters`` is not filled during this method, as this method will provide the necessary information, we need to access ``$typeConverterMap`` which contains the class name already.

**What I did**
Fix accessing empty array.

**How I did it**
Accessing method internal existing array containing the necessary information.

**How to verify it**
Register two TypeConverters for same conersion with same priority.

**Checklist**

I could not find any information about Flow at the given url.


kitsunet:
Good catch, thanks! But I assume this bug exists in earlier versions as well? So the fix should target the lowest still supported version in which the bug exists.

DanielSiepmann:
Of course, if you can provide the information, the provided link only contains information about neos, but not flow: https://www.neos.io/features/release-roadmap.html

DanielSiepmann:
I've updated the base branch for PR and the commit accordingly to be merged into Flow 3.2.

Issue #896, opened 2017-03-20
AjaxWidgetConfiguration is not used in controller


DanielSiepmann:
### Description

If you modify Widget configuration for ajax widgets inside of ``getAjaxWidgetConfiguration`` it's not used inside of the controller. Instead the non ajax widget configuration is used.

### Steps to Reproduce

1. Build a WidgetViewHelper
2. Configure ViewHelper to use Ajax
3. Extend ``getAjaxWidgetConfiguration`` and modify the returned configuration

#### Expected behavior

The returned configuration is accessable inside of the controller in ``$this->widgetConfiguration``.

#### Actual behavior

The configuration returned by ``getNonAjaxWidgetConfiguration`` is provided instead.

### Research / bug report

The bug is introduced through https://github.com/neos/flow-development-collection/blob/4.0.2/Neos.FluidAdaptor/Classes/Core/Widget/AbstractWidgetViewHelper.php#L131 where the ``NonAjaxWidgetConfiguration`` is always set, even for ajax widgets. Afterwards https://github.com/neos/flow-development-collection/blob/4.0.2/Neos.FluidAdaptor/Classes/Core/Widget/WidgetContext.php#L127 is called by https://github.com/neos/flow-development-collection/blob/4.0.2/Neos.FluidAdaptor/Classes/Core/Widget/AbstractWidgetController.php#L53 which will first check whether ``nonAjaxWidgetConfiguration`` exists, which is always the case, and return that instead of the ``ajaxWidgetConfiguration``.

The easiest possible solution would be to invert the condition / check in https://github.com/neos/flow-development-collection/blob/4.0.2/Neos.FluidAdaptor/Classes/Core/Widget/WidgetContext.php#L127 to return the ajax config if it's not null.

### Affected Versions

Flow: 4.0.2, master

fcool:
The behaviour is correct. Everything behaves as it should be.

What really is confusing, that, if a widget-ViewHelper no ChildNodes (say: empty tag), it does not get the viewHelperNodes and the crucial ViewHelperVariableContainer set in its RoodNode.

I'm not sure if this is an issue or a real cool Fluid feature...

PR #897, opened 2017-03-20
BUGFIX: Make ajaxWidgetConfiguration accessible in controller

DanielSiepmann:
**What I did**

* Allow ajax widget configuration to be available in controller.

**How I did it**

* Inverting the existing condition.

**How to verify it**

* See bug report neos/flow-development-collection/issues/896

**Checklist**

Fixes: #896


DanielSiepmann:
This is just a proposal. Feel free to ignore it and provide another solution.

albe:
Fix looks reasonable, I'm just not sure what possible side-effects of this change could be. i.e. are there use cases that would depend on the old (buggy) behaviour?

bwaidelich:
Thanks for your contribution!
Could you please add a quick note what bug this fixes, i.e. _why_ you did this?

DanielSiepmann:
Sure @bwaidelich inside your Widgets, you can modify the widget configuration via `setAjaxWidgetConfiguration` and `setNonAjaxWidgetConfiguration `.

I would expect that using `setNonAjaxWidgetConfiguration` for an ajax widget is the way to go. But since this does not have any effect, the method is useless. That's what this fix is for. An alternative would be to remove the method and calls.

Hope that's enough? Also I tried to be more explicit already in https://github.com/neos/flow-development-collection/issues/896 .

bwaidelich:
Thanks for your comment, I missed the link to the bug report (added a "Fixes" footer now).
Unfortunately this seems to break the `WidgetContextTest` unit test (see https://travis-ci.org/neos/flow-development-collection/jobs/212957918#L737)

DanielSiepmann:
Ok, by investigating the tests I've found out that:
`ajaxWidgetConfiguration` is only available inside of an ajax request, not inside of the controller during the same request. That was what I've tried to fix. But then I just didn't get the idea behind the code and everything should be fine without adjustments.

bwaidelich:
Thanks a lot for investing the time. I wasn't sure either, the code could be more explicit there..

PR #898, opened 2017-03-20
BUGFIX: Fix 4.0 composer manifests

kdambekalns:
A number of inconsistencies and errors are in the composer manifests of the 4.0 branch:

- some libraries pinned to strictly 4.0.0
- outdated/wrong names and namespaces in merged collection manifest
- inconsistent dependencies on PHP
- inconsistent dependencies on PhpUnit

The first is an actual bug, since even with 4.0.2 of Flow, those packages are pinned to
the 4.0.0 version, even though it should be 4.0.2.

While the last two might in fact be correct, the merged manifest is prone to containing
wrong constraints, so this change sync the related constraints with the Flow package
itself.


aertmann:
looks good overall, but would be reluctant about requiring phpunit 5.4+ as we've had multiple issues with minor versions of that library in the past.. would prefer locking to a known working version like 5.4.0 instead

only reason 4.8 wasn't locked was that 4.x branch was deprecated

kdambekalns:
I conquer - in the dev distribution we use `~5.4` so this should be fine. I'll go one step further: PhpUnit usually does a very good job with semantic versioning - plus I'll prepare another update so we use PhpUnit 6, as that is the release we should be using… There is no reason not to, since we also do PHP 7 only…

aertmann:
That could easily be due to someone not considering the point I'm making last time it was updated. I can just say that I remember more than once PHPUnit breaking our tests on minor version upgrades, so I cannot agree that they do a good job with semantic versioning. Anyway if you prefer a loose constraint and risk it happening again, feel free to do so.

kdambekalns:
I'm in the "stay current" camp in this case. But I'll gladly pin it, if people think it's important. I'll prepare the 6.0 PR and then we can still change/leave it however we see fit. But I'd like to see this merged so 4.0 is actually having the correct packages and I can fix master (so the base dist can be installed again) after merging this in.

_Yes, I know that there is little use in running the base dist off of master, but…_ :)

aertmann:
Sure, no need to block this one because of that of course. Just mentioning it. Change looks good.

kdambekalns:
Thanks, I'll be working on this now and check for anything that needs to be done following this merge…

PR #902, opened 2017-03-21
TASK: Remove PhpUnit 4.8 support code

kdambekalns:
We no longer support PhpUnit 4.8 so some code could be removed.


kitsunet:
Mmm, but isn't this strictly a breaking change?

aertmann:
shouldn't be as you need 5.x if using php 7 anyway which Flow 4.x requires

kitsunet:
Right, good point!

PR #903, opened 2017-03-21
TASK: Use PhpUnit 6

kdambekalns:


kdambekalns:
Most probably TravisCI will fail, since the dev dist requires PhpUnit 5.4 (see https://github.com/neos/flow-development-distribution/pull/20). I'll know after lunch…

kdambekalns:
So @kitsunet merged the dist change and restarted the TravisCI build for this one (I think). All green!

kdambekalns:
Now on to Neos… :)

kitsunet:
Yes, Yes, that's what I did 👍 Thanks for emrging

PR #906, opened 2017-03-23
TASK: Fix "undefined class" warnings

kdambekalns:
Mostly fixes comments, but some use statements as well.


gerhard-boden:
Can we take care of ```Fatal error: Class 'PHPUnit_Framework_TestCase' not found in /home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Flow/Tests/BaseTestCase.php on line 22````?

I sleep much better when approving clean up tasks when the tests are green 😄 (although I'm aware that the error is unrelated to this PR in particular)

kdambekalns:
Unrelated, but disturbing. See #907 for a fix (I hope).

PR #910, opened 2017-03-24
!!! TASK: Add rescanPackages() to PackageManagerInterface

kdambekalns:
This was originally scheduled for 4.0 but forgotten. It is added now,
excuse the inconvenience. But you should really not be implementing
your own PackageManager.


kitsunet:
@albe that's what I wrote in Slack:

> I argued for removing the PackageManagerInterface while I was refactoring 
> but that was blocked by someone
> IMHO no one should replace our package manger, full stop
> so yes, let's add it to the interface IMHO
> but let's really think about deprecating the interface

PR #915, opened 2017-03-28
TASK: Remove requireAutoloaderForPhpUnit() from Booting\Scripts

kdambekalns:
This is legacy code in such a bad way...


kdambekalns:
Thanks for fixing, @kitsunet :)

PR #916, opened 2017-03-28
TASK: Declare transitive dependencies

kitsunet:
Additionally move an Object configuration to the right place.

### DIFF
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -22,7 +22,6 @@
         "neos/utility-mediatypes": "*",
         "neos/utility-objecthandling": "*",
         "neos/utility-opcodecache": "*",
-        "neos/utility-pdo": "*",
         "neos/utility-schema": "*",
         "neos/utility-unicode": "*",
         "neos/fluid-adaptor": "*",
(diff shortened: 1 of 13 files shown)


Neos.Flow/composer.json
```diff
@@ -32,6 +31,7 @@
         "doctrine/orm": "~2.5.0",
         "doctrine/migrations": "~1.3.0",
         "doctrine/dbal": "~2.5.0",
+        "doctrine/common": ">=2.4,<2.8-dev",
```
kdambekalns: Should be `2.8.x-dev` AFAIK - it works, but there is a difference, see https://semver.mwl.be/#?package=doctrine%2Fcommon&version=%3E%3D2.4,%3C2.7-dev&minimum-stability=stable vs >=2.4,<2.7.x-dev (using 2.7 there to make the difference visible). What is the intended result?

kitsunet: In fact this is copied from the `doctrine/dbal` and `doctrine/orm` requirements as I thought it makes sense to match it to them...

kdambekalns: In that case probably intended. Fine. :)

PR #919, opened 2017-03-29
TASK: Adjust DocTools settings

kdambekalns:
Fixes neos/neos-development-collection#1490


aertmann:
Something seems off, `TYPO3FluidViewHelperReference.rst` only contains one view helper.

PR #920, opened 2017-03-30
FEATURE: add Date.formatCldr helper

litovchenko1:
**What I did**

This change adds CLDR formatting capabillity to Date Eel helper,
to match functionality of format.date Fluid ViewHelper.

**How to verify it**

Just use `${Date.formatCldr(someDateObject, 'dd MMMM yyyy HH:mm:ss', 'de')}` to test. Also test different combinations with missing arguments.

**Checklist**

### DIFF
--- a/Neos.Eel/Classes/Helper/DateHelper.php
+++ b/Neos.Eel/Classes/Helper/DateHelper.php
@@ -75,19 +75,19 @@ class DateHelper implements ProtectedContextAwareInterface
      *
      * @param \DateTime $date
      * @param string $cldrFormat Format string in CLDR format (see http://cldr.unicode.org/translation/date-time)
-     * @param string $locale String locale - example (de|en|ru_RU). Must be one of Neos\Flow\I18n\Cldr\Reader\DatesReader::FORMAT_LENGTH_*'s constants.
+     * @param null|string $locale String locale - example (de|en|ru_RU)
      *
      * @return string
      */
-    public function formatCldr($date = null, $cldrFormat = null, $locale = null)
+    public function formatCldr($date, $cldrFormat, $locale = null)
     {
-        if ($date === null) {
-            $date = new \DateTime('now');
+        if (!$date instanceof \DateTimeInterface) {
+            throw new \InvalidArgumentException('"' . $date . '" could not be parsed by \DateTime constructor.');
         }
-        if ($cldrFormat === null) {
-            $cldrFormat = 'dd MMMM yyyy HH:mm:ss';
+        if (empty($cldrFormat)) {
+            throw new \InvalidArgumentException('CLDR date formatting parameter not passed.');
         }
-        if ($locale === null) {
+        if (empty($locale)) {
             $useLocale = $this->localizationService->getConfiguration()->getCurrentLocale();
         } else {
             $useLocale = new Locale($locale);
--- a/Neos.Eel/Classes/Helper/DateHelper.php
+++ b/Neos.Eel/Classes/Helper/DateHelper.php
@@ -13,14 +13,28 @@ namespace Neos\Eel\Helper;
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Eel\ProtectedContextAwareInterface;
+use Neos\Flow\I18n\Exception as I18nException;
+use Neos\Flow\I18n\Formatter\DatetimeFormatter;
+use Neos\Flow\I18n\Locale;
+use Neos\Flow\I18n\Service as I18nService;
 
 /**
  * Date helpers for Eel contexts
- *
- * @Flow\Proxy(false)
  */
 class DateHelper implements ProtectedContextAwareInterface
 {
+    /**
+     * @Flow\Inject
+     * @var DatetimeFormatter
+     */
+    protected $datetimeFormatter;
+
+    /**
+     * @Flow\Inject
+     * @var I18nService
+     */
+    protected $localizationService;
+
     /**
      * Parse a date from string with a format to a DateTime object
      *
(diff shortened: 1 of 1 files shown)


Neos.Eel/Classes/Helper/DateHelper.php
```diff
@@ -57,6 +71,31 @@ public function format($date, $format)
     }
 
     /**
+     * Format a date to a string with a given cldr format
+     *
+     * @param \DateTime $date
+     * @param string $cldrFormat Format string in CLDR format (see http://cldr.unicode.org/translation/date-time)
+     * @param string $locale String locale - example (de|en|ru_RU). Must be one of Neos\Flow\I18n\Cldr\Reader\DatesReader::FORMAT_LENGTH_*'s constants.
```
albe: I think the second part of that line is a bit off, it's not a format length parameter :)

Neos.Eel/Classes/Helper/DateHelper.php
```diff
@@ -57,6 +71,31 @@ public function format($date, $format)
...
+     */
+    public function formatCldr($date = null, $cldrFormat = null, $locale = null)
+    {
+        if ($date === null) {
+            $date = new \DateTime('now');
+        }
+        if ($cldrFormat === null) {
+            $cldrFormat = 'dd MMMM yyyy HH:mm:ss';
+        }
+        if ($locale === null) {
```
aertmann: probably a good idea to not be strict about non set values to allow `${Date.formatCldr(date, '', 'fr')}` as well as `${Date.formatCldr(date, null, 'fr')}`

albe: Good point, bu then please be explicitly unstrict ;) i.e. `$cldrFormat === null || $cldrFormat === ''`
Relying on the PHP type juggling behaviour is a fallacy

kitsunet:
I would actually simply enforce giving a format and not fallback to any default.

litovchenko1:
VThank you very much for your support
My first function for CMS NEOS.

Examples:
dateObject_1 = ${Date.formatCldr(q(node).property("testMyDate"), 'dd MMMM yyyy HH:mm:ss', 'ru_RU')}
dateObject_2 = ${Date.formatCldr(q(node).property("testMyDate"), 'dd MMMM yyyy HH:mm:ss')}

Neos.Eel/Classes/Helper/DateHelper.php
```diff
@@ -57,6 +71,31 @@ public function format($date, $format)
     }
 
     /**
+     * Format a date to a string with a given cldr format
+     *
+     * @param \DateTime $date
+     * @param string $cldrFormat Format string in CLDR format (see http://cldr.unicode.org/translation/date-time)
+     * @param null|string $locale String locale - example (de|en|ru_RU)
+     *
```
aertmann: no empty line between these

aertmann:
regarding documentation, I think that is auto-generated but not sure if it happens regularly for the master branch though

Neos.Eel/Classes/Helper/DateHelper.php
dimaip: We tried adding the tests, but there are like 5 methods to mock for 3 lines of actual code, don't know if it's worth to load Ivan with such task :)

dimaip:
@kitsunet On the other hand `Date.format` does just that, would make sense to handle input the way it does?

Neos.Eel/Classes/Helper/DateHelper.php
```diff
@@ -13,15 +13,29 @@
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Eel\ProtectedContextAwareInterface;
+use Neos\Flow\I18n\Exception as I18nException;
```
dimaip: This exception is currently not used. Remove it if you don't need it.

kdambekalns:
- Eel helper ref is rendered automatically
- why not mock lines for testing 3 lines?
- would be awesome to have this in 4.1, as it seems to be really almost done :)

Neos.Eel/Classes/Helper/DateHelper.php
```diff
@@ -57,6 +71,31 @@ public function format($date, $format)
...
+     * @param \DateTime $date
+     * @param string $cldrFormat Format string in CLDR format (see http://cldr.unicode.org/translation/date-time)
+     * @param null|string $locale String locale - example (de|en|ru_RU)
+     *
+     * @return string
+     */
+    public function formatCldr($date, $cldrFormat, $locale = null)
+    {
+        if (!$date instanceof \DateTimeInterface) {
+            throw new \InvalidArgumentException('"' . $date . '" could not be parsed by \DateTime constructor.');
```
bwaidelich: Can't we just use a `\DateTimeInterface` type hint instead?

bwaidelich: Also this would break if `$date` is not castable to string

dimaip:
@litovchenko1 how is your progress? Do you need any help to move forward?

dimaip:
@litovchenko1 ping

Neos.Eel/Classes/Helper/DateHelper.php
albe: good call

Neos.Eel/Classes/Helper/DateHelper.php
albe: Well, I prefer test coverage tbh. Basically that just shows that this method uses too much delegation, esp. the `$this->localizationService->getConfiguration()->getCurrentLocale();` for example violates the "Law of Demeter". So this should possibly be refactored, either by adding a method to localizationService or by adding a (protected) method in this helper, that fetches the current locale.

gerhard-boden:
@litovchenko1 ping :)

Would like to see this in 4.2, but not too much time left I'm afraid. Any chance you could find the time to get this finished within the upcoming weeks?

gerhard-boden:
Closed in favour of https://github.com/neos/flow-development-collection/pull/1032

PR #921, opened 2017-04-01
BUGFIX: Attempt reconnect only if EntityManager is open

### LINKED ISSUE #870: Actual Doctrine exceptions swallowed leading to difficult debug situations
### Description

When Doctrine throws an exception it is caught in a reconnect attempt solution hiding the actual issue leading to difficult to understand situations.

### Steps to Reproduce

1. Alter a database table resulting in a mapping error.
2. Connect to Doctrine somehow

#### Expected behavior

Actual Doctrine mapping error exception being shown.

#### Actual behavior

Other exception being thrown by Doctrine due to the entity manager being closed (side-effect).

### Affected Versions

3.0 to 4.0

### Information

This is the change that introduced the issue https://github.com/neos/flow-development-collection/commit/93b458812493f23c89d6ef8815acbe4d57b53256

Since https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php#L168 catches all exceptions, not just database connection failed exceptions (if they exist?) and tries to connect again it happens. Instead of catching all, it should catch a specific exception or at the very least check if the exception has anything to do with database connectivity issues. If not it should throw it.

The actual exception does actually gets logged, but if you're debugging and looking at the exception print it's not obvious that there might be other information available in the log, thus leading users on wild goose hunts.

jasinai:
If a `\Doctrine\DBAL\DBALException` is thrown in `persistAll()`, a reconnect will be attempted,
but only if `EntityManager->isOpen()` returns true. Otherwise the exception is re-thrown.

Before the connection to the server was sometimes never re-established, since only the specific
Flow\Error\Exception was caught.

Also this will prevent hiding exceptions that result in a closed EntityManager.

Fixes #870

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
@@ -163,6 +163,9 @@ class PersistenceManager extends AbstractPersistenceManager
         try {
             $this->entityManager->flush();
         } catch (\Doctrine\DBAL\DBALException $exception) {
+            if (!$this->entityManager->isOpen()) {
+                throw $exception;
+            }
             $this->systemLogger->logException($exception);
             /** @var Connection $connection */
             $connection = $this->entityManager->getConnection();
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
@@ -22,7 +22,6 @@ use TYPO3\Flow\Persistence\Exception\ObjectValidationFailedException;
 use TYPO3\Flow\Persistence\Exception as PersistenceException;
 use TYPO3\Flow\Persistence\Exception\UnknownObjectException;
 use TYPO3\Flow\Reflection\ClassSchema;
-use TYPO3\Flow\Error\Exception;
 use TYPO3\Flow\Reflection\ObjectAccess;
 use TYPO3\Flow\Reflection\ReflectionService;
 use TYPO3\Flow\Utility\TypeHandling;
@@ -163,7 +162,7 @@ class PersistenceManager extends AbstractPersistenceManager
 
         try {
             $this->entityManager->flush();
-        } catch (Exception $exception) {
+        } catch (\Exception $exception) {
             $this->systemLogger->logException($exception);
             /** @var Connection $connection */
             $connection = $this->entityManager->getConnection();
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
@@ -162,7 +162,7 @@ class PersistenceManager extends AbstractPersistenceManager
 
         try {
             $this->entityManager->flush();
-        } catch (\Exception $exception) {
+        } catch (\Doctrine\DBAL\DBALException $exception) {
             $this->systemLogger->logException($exception);
             /** @var Connection $connection */
             $connection = $this->entityManager->getConnection();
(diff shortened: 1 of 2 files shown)


albe:
Thanks a lot for this and congrats on your first PR :) This is actually directly related to/a fix for #870. Hence I guess it would make sense to even reduce that to only catch `\Doctrine\DBAL\DBALException`, since any application level exceptions are supposed to bubble up. @bwaidelich @aertmann

albe:
Ok, so the `persistAllReconnectsConnectionOnFailure` test definitely needs adjustment...

```
$this->mockEntityManager->expects($this->exactly(2))->method('flush')
->will($this->throwException(new **FlowError\Exception**('Dummy connection error')));
```

Out of interest/questioning the above test, I dug a bit deeper into this whole process - so sorry for the following wall of text:

AFAIS, what happens during flush from Flows side, is just the entity validation which possibly throws a `Neos\Flow\Error\Persistence\Exception` - however, that is also something, that should *not* be reattempted, because it will just fail again.

Doctrine seems to possibly throw the following exceptions (not limited to):
- `\Doctrine\DBAL\DBALException` - if the connection is not yet established before commit and there is an error during database connection. AFAIS this should never really be the case, because the connection is already established during creation of the doctrine EntityManager and it will not know it was lost since then. However, this exception is also raised when doctrine catches Driver exceptions during execution of a transaction.
- `\Doctrine\ORM\OptimisticLockException` - obviously on optimistic locking errors, which would lead to the same exception on immediate reattempt anyway
- `\Doctrine\DBAL\ConnectionException` - if no transaction active or transaction is rollback only during `commit` (note this one extends DBALException)
- any of the platform specific `Doctrine\DBAL\Driver\DriverException` if something goes wrong during execution of the commit

Also note that any exception raised during the actual commit (all of above doctrine specific cases except the first) will close the entity manager, which will lead to a failing reattempt to the flush in any case, hiding the original error (as described in #870). **I'd suggest checking for a closed entity manager inside the catch, and if it is closed, just rethrow the original exception.**

So the only plausible use case for this reconnection attempt is, when the connection to the database was lost between creation of the EntityManager and the flush, such that the attempt to flush will throw an exception *outside* of the doctrine commit try/catch. This ist mostly the `DBALException`, but might also be some platform dependant exception actually (e.g. `PDOConnection::beginTransaction` throwing `\PDOException` - http://php.net/manual/de/pdo.begintransaction.php).

So in conclusion I'm a bit stumbled as to the actual effectiveness of this whole reconnection attempt, and reading up on the original commit, it sounds like the current solution was rather a hacky attempt at fixing a problem that should rather be handled like this:
http://stackoverflow.com/a/15374063 / http://stackoverflow.com/a/32648896

albe:
See #926 for a follow-up/alternative solution to the issue at hand @bwaidelich @aertmann @kdambekalns

albe:
Ok, so let's go with this and adding the check for the closed entity manager inside the catch for now to get this solved. We can further discuss #926 and if we want to change the behaviour there.

kdambekalns:
Well, *originally* that was supposed to solve issues with lost connections. Everything else was not intended to be caught/solved/…

kdambekalns:
@albe I don't really get what you want to do after reading

> Ok, so let's go with this and adding the check for the closed entity manager inside the catch for now to get this solved. We can further discuss #926 and if we want to change the behaviour there.

Do you want to merge this (i.e. #921, after fixing the test) and the discuss further in #926?

albe:
Exactly @kdambekalns :) Sorry, the previous comment was typed in a bit of a rush

kdambekalns:
@albe No problem, just wanted to make sure. And, I agree to that approach.

albe:
Will fix the test when I'm home at my laptop. Not fun from Smartphone

albe:
Argh... that turned out to be tricky to get the tests right, mostly because PhpUnit is a bit unintuitive about overriding method expectations in tests (`isOpen` returning `false` on second invocation in the new test) and I didn't get that you need to completely override the mock object for that... one more reason to really go with my idea to refactor tests :/

Anyway, this should be good to merge now

Issue #899, opened 2017-03-21
Fizzle cannot filter by nested object property


kitsunet:
### Description

By looking at descriptions and grammer a fizzle filter should support filtering by nested object properties but that doesn't work in the way the parser works currently.

### Steps to Reproduce

Given I have an array of (Media) Asset objects $assets and I want to filter them via FlowQuery/Fizzle by fileExtension (path: `resource.fileExtension`), I would try the following expression:

```
$flowQuery = new FlowQuery($assets);
$flowQuery->filter('[resource.fileExtension = pdf]');
```

This will throw an error because Fizzle cannot parse the property path containing `.`, also alternatives like `resource/fileExtension` won't work.

#### Expected behavior

The FlowQuery context contains an array of assets only with "pdf" file extension.

#### Actual behavior

A Fizzle parser error is thrown.

### Affected Versions

all

kitsunet:
@hlubek Can you confirm this is a bug and not intended missing behavior? I think the fix is rather easy (to allow `.` in the filter parser)...

aertmann:
Would be surprised if it was intentional, rather not thought of. In any case makes sense to support.

joachimmathes:
Got some introduction by @kitsunet and give it a try now. :)

joachimmathes:
Made a pull request: https://github.com/neos/flow-development-collection/pull/924

PR #924, opened 2017-04-02
TASK: #899 Make Fizzle filter nested object properties

joachimmathes:
**What I did**
Implements #899. I extended the FlowQuery grammar to filter nested object properties.

**How I did it**
 It was not possible to just extend the `Identifier` rule in `AbstractParser.peg.inc` since that led to a lot of regressions in the Eel parser. Thus, a new rule `PropertyPath` was necessary.

**How to verify it**
Unit-Tests in Neos.Eel run successfully.

### DIFF
--- a/Neos.Eel/Resources/Private/Grammar/AbstractParser.peg.inc
+++ b/Neos.Eel/Resources/Private/Grammar/AbstractParser.peg.inc
@@ -12,7 +12,7 @@ namespace Neos\Eel;
  * source code.
  */
 
-require_once __DIR__ . '/../../../Resources/Private/PHP/php-peg/Parser.php';
+require_once __DIR__ . '/../Resources/Private/PHP/php-peg/Parser.php';
 
 /*!* !insert_autogen_warning */
 
@@ -42,6 +42,7 @@ StringLiteral: SingleQuotedStringLiteral | DoubleQuotedStringLiteral
 BooleanLiteral: 'true' | 'TRUE' | 'false' | 'FALSE'
 
 Identifier: / [a-zA-Z_] [a-zA-Z0-9_\-]* /
+PropertyPath: Identifier ( '.' Identifier )*
 */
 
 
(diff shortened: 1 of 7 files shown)


Neos.Eel/Resources/Private/Grammar/AbstractParser.peg.inc
```diff
@@ -12,7 +12,7 @@ namespace Neos\Eel;
  * source code.
  */
 
-require_once __DIR__ . '/../../../Resources/Private/PHP/php-peg/Parser.php';
+require_once __DIR__ . '/../Resources/Private/PHP/php-peg/Parser.php';
```
albe: Is this really correct? Not rather `'/../PHP/php-peg/Parser.php'`?

joachimmathes: Thank you! :)

And thanks for pointing this out. When I generated `AbstractParser.php` from `AbstractParser.peg.inc` for the first time, i.e. with `/../../../Resources ...`, the Unit-Tests didn't run at all due to the wrong path. So I changed it to `/../Resources/ ...`, since the resulting `AbstractParser.php` is placed in `Classes/`. That's why I'm pretty sure this is correct. And not only the Unit-Tests have worked as expected by then, but also the respective line in `AbstractParser.php` hasn't changed in the diff.

albe: Hm, must be a misconception on my side and I guess I have a clue on what it is (this just being the code template for the actual AbstractParser, which resides somewhere else and not being actually executed itself)... well, no objections from my side then :)

joachimmathes:
The final parsers are created from the `inc` templates by `Neos.Eel/Scripts/generateEelParser.sh`.

joachimmathes:
Lesson learned: Always check the NEOS development package when changing fundamental code in the Flow development package. :)

PR #925, opened 2017-04-03
!!! FEATURE: Allow bypassing Flow class loader for performance

kitsunet:
Currently the composer class loader is only used as a fallback to our own,
but especially if the optimized loader is used the composer one is much
faster.

On systems in which all packages/classes are registered correctly via
autoload statements in ``composer.json`` files using our own class loader
only for proxy classes can bring an substantial boost in performance for
every request.

In order to enable this feature you need to set an environment variable
``FLOW_ONLY_COMPOSER_LOADER=1``. Please test carefully if that breaks due
to your autoload configuration not being fully composer ready.

Additionally it is recommended to use the optimized composer loader by
calling ``composer dumpautoload -o``.

While not breaking in itself this change deprecates using our class loader
for anything else than proxy classes. In practice this means you should
always enable composer auto loader only by using above mentioned environment
variable. At least make sure that your projects work with this
enabled.

We will drop the variable and make this the default behavior in the next
major version of Flow (version 5.0) which means only classes that are
correctly added to composer (loaded) packages with autoload configuration
are being loaded correctly.


kdambekalns:
Is that flag supposed to stay? Could we invert the default? Use our loader as a fallback to it's faster but "unbreakable"?

albe:
Yes yes yes, please :)

kitsunet:
Ok so the following variants exist:

1) We split our loader between proxy and rest, so we can sandwich the composer loader:
- proxy loader
- composer loader
- flow class loader

That would work fine and I could do it, it's a bit bigger change though. We could then get rid of the flag of course. Has the drawback that our loader would still be initialized with a lot of information that is in theory superfluous (but we can't know that), so it eats up a bit of the performance gain.

2) My plan was to deprecate any non composer loaded classes and remove the flag and the respective parts of our loader on next major.

kitsunet:
Just for clarification, simply reversing the order is impossible as our proxy loading MUST come first.

albe:
> Has the drawback that our loader would still be initialized with a lot of information that is in theory superfluous (but we can't know that), so it eats up a bit of the performance gain.

Can we just keep the env var flag as an further optimization then and keep moving towards your goal in 2., using the flag as the "check that everything works when the flow class loader is disabled"?

kitsunet:
Ok, updated commit (should also fix tests) and added hint about deprecation to commit message.

albe:
Sorry, didn't come around to actually test it in our project and will also be on vacation the next 2 weeks now

kitsunet:
Now again up to master, please review.

albe:
Looks good on a first try with my local Neos.EventSourcing project. Could not yet try our work projects, since the servers can currently not easily be updated to PHP7 and hence Flow 4.

We really need to document the fact that non-composer-installed packages will break very well. Will try to get some more tests done.

robertlemke:
> but especially if the optimized loader is used the composer one is much
faster.

which speed gain did you measure, for example based on the Neos demo site?

dfeyer:
Can we merge this one, it just make non sense to have two class loader. And if we can slim down our code base, and maybe improve performance, it's just a win win. 

Fine also to deprecate our loader, and remove it in the next major. Now the community is bit more used to composer usages, and it's fairly easy to adapt project than can be broken by this the change in the next major. We can even write a blog post on the subjet. The problem concern mainly packages that are include in the distribution. It's easy to transform those distribution with local composer sources.

kitsunet:
WITH patch:

```
Lifting the server siege...
Transactions:		        6858 hits
Availability:		      100.00 %
Elapsed time:		       59.10 secs
Data transferred:	        4.09 MB
Response time:		        0.21 secs
Transaction rate:	      116.04 trans/sec
Throughput:		        0.07 MB/sec
Concurrency:		       24.93
Successful transactions:        6858
Failed transactions:	           0
Longest transaction:	        0.52
Shortest transaction:	        0.07
```


WITHOUT patch:
```
Lifting the server siege...
Transactions:		        6467 hits
Availability:		      100.00 %
Elapsed time:		       59.14 secs
Data transferred:	        3.85 MB
Response time:		        0.23 secs
Transaction rate:	      109.35 trans/sec
Throughput:		        0.07 MB/sec
Concurrency:		       24.94
Successful transactions:        6467
Failed transactions:	           0
Longest transaction:	        0.74
Shortest transaction:	        0.04
```

pretty arbitrary though. Can probably go the other way as well. Would be interesting to benchmark on a non SSD system for a change. I _think_ that might show more difference.

albe:
I think that's already good enough of an improvement to justify this, let alone the simplification of our code base in the mid term.

I'm unsure about if this should land in LTS, because the "breakiness" of this is now only on non-composer installed packages. Should we be strict about b/c with this?

kitsunet:
@albe it is fully backwards compatible right now, especially as long as you DON'T set the env. If you set it, ofc it might break, but you actively have to go in and set that env.

dfeyer:
I say merge it ;)

dfeyer:
Can we remove the !!! here ? It's not breaking if not enable, so no red flag required ?

kitsunet:
It deprecates the old way and we always marked these kinds of changes "breaking" to warn users that big change is happening in a future version.

kitsunet:
Probably, yes.

kitsunet:
Yes, well it is as easy as configuring functional test classes in autoload. But now the interesting question, do we actually want to do that. I would rather suggest the following: 

We add a testing autoloader that goes through composer.json of installed packages and uses the require-dev to build a map just like our autoloader does now. That way tests would work fine with autoload-dev configured in local packages and we can move forward with this.

albe:
Sounds like a plan.

kitsunet:
Here we go. This is not exactly the final state yet. But better than what we have now with the option of extra performance by usign the flag.

kitsunet:
Not sure what those styleci fails are. Nto exactly code I even touched.

kitsunet:
There we go, merged master into it which makes styleci happy again.

Would merge this weekend if no one else blocks it.

kdambekalns:
💯

PR #926, opened 2017-04-03
BUGFIX: PersistenceManager correctly checks and reattempts lost connections

### LINKED ISSUE #870: Actual Doctrine exceptions swallowed leading to difficult debug situations

albe:
This change replaces the previously put in place try/catch attempt at handling lost connections
with a proper explicit call to the doctrine connection's `ping()` method.
As a side-effect, this will no longer trigger the `AllObjectsPersisted` signal if flushing fails, however
this is not really breaking, since in those circumstances, the application is supposed to not do any
further work anyway.

Fixes #870

This is a follow-up to #921 - see discussion there.


bwaidelich:
Hey,
thanks for this but I have a couple of questions/concerns:
1. is it really feasible to perform a `ping` for _every_ `persistAll()` call and is that better than retrying on exceptions?
2. An actual `ping` is only supported by `PingableConnection`s (currently only `MysqliConnection` it seems) on other backends it performs a dummy SQL query
3. Is there a particular reason why you deviate form your (btw very helpful) suggestions mentioned in https://github.com/neos/flow-development-collection/pull/921#issuecomment-291180645 ?

kdambekalns:
I like this, it looks way cleaner than the current code. But… Did you check the history of this block of code to see the changes and check if there is any wisdom hidden? My gut feeling says, there's more to the current mess, than meets the eye. :)

albe:
My conclusion was fully based on https://github.com/neos/flow-development-collection/commit/93b458812493f23c89d6ef8815acbe4d57b53256 but you're right, looking at full history would be better. Unfortunately after taking a look now, nothing useful came up. That part of the code was effectively not touched since introduction. So in the end it's a decision of code clearity vs. performance I guess.

bwaidelich:
It _does_ make more sense, at least "at first sight". The reason why we added the `try/catch` is to prevent nasty to debug errors due to timeouts in long-running processes, if I remember correctly.

The change that originally introduced this is: https://github.com/jasinai/flow-development-collection/commit/93b458812493f23c89d6ef8815acbe4d57b53256 - so maybe we should check back with Martin whether the new version still works with him - or check with the Doctrine folks what they would recommend

albe:
Well, Marco Pivetta already brought that solution up in http://stackoverflow.com/questions/15362070/zf2-doctrine2-server-has-gone-away-how-to-jog-an-old-connection/15374063#15374063 (notice his comment from march 2014 where he hints at the new `ping()` feature). So I think it's safe to say this is the official way to go about checking lost connections.
But I have no issues in asking them again if there's a better solution now

albe:
Fixed tests and also slightly refactored in that process. `->willReturn('foo')` is just so much easier to read than `->will($this->returnValue('foo'))`.

bwaidelich:
Looks good to me, but tbh I'm still not 100% sure whether it's a problem to do that ping for every request, but surely better than what we have now

kdambekalns:
Ok, merged this now. Let's keep an eye open for any new issues being reported so we can revert/amend/fix this if needed… Thanks!

bwaidelich:
Is it possible that this one broke the Neos Setup?
With a fresh installation I now get a `ConnectionException` with ```An exception occured in driver: SQLSTATE[HY000] [1045] Access denied for user ''@'localhost' (using password: NO)```

kdambekalns:
Hm. Haven't used that one for a while… :/

kdambekalns:
Related to https://github.com/neos/neos-development-collection/issues/1658?

bwaidelich:
-> See https://github.com/neos/neos-development-collection/issues/1663

PR #928, opened 2017-04-04
TASK: throw InvalidConfigurationException if there are no secondLevelCache settings for doctrine

kabarakh:
#When using more than one doctrine backend, you have to make sure to set TYPO3.Flow.persistence.doctrine.secondLevelCache, or else you will get "An error occurred in the Database Abstraction Layer." as an exception which doesn't tell you much. By throwing an approriate exception we make clearer what the problem is.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/EntityManagerFactory.php
@@ -76,6 +76,9 @@ class EntityManagerFactory
         if (!is_array($this->settings['backendOptions'])) {
             throw new InvalidConfigurationException(sprintf('The TYPO3.Flow.persistence.backendOptions settings need to be an array, %s given.', gettype($this->settings['backendOptions'])), 1426149224);
         }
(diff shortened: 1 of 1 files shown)


albe:
Hey, thanks for this and congrats on your first PR :) This makes a lot of sense for developer experience, so thanks a lot for bringing it up! I even think we could just silently skip the
`$this->applySecondLevelCacheSettingsToConfiguration(...)` call if the setting is not an array. That way the developer would not unnecessarily be confronted with an exception at all. What do you think?

kitsunet:
Just to let you know I made this an example case for deciding how we want to "treat" developers 😆 
https://discuss.neos.io/t/discussion-to-be-meticulous-or-lenient/2115

Don't let that distract you from finishing this though. I really want to decide this long term. If this is merged either way we could change it even after a decision was taken...

daniellienert:
Thanks @kabarakh for improving!
And @kitsunet, thanks for bringing that topic up :) As for now I would say it makes sense to handle that misconfiguration like the other and display a more meaningful error message.

albe:
Not blocking this of course, so since we have two (and a half) approvals, I'll merge this for now. It can easily be improved further upon whatever we decide to and it's in any case much better than completely uninformal failure.

So congrats on your first contribution @kabarakh :)

PR #931, opened 2017-04-06
BUGFIX: Injected configuration should also be ignored in class schema

kitsunet:
Injected properties were always ignored in the persistence class schema as
it makes no sense to persist injected properties. Same is obviously true for
injected configuration, but this was never done so this change ignores
properties with a ``InjectConfiguration`` annotation.

This may be breaking if you use ``InjectConfiguration`` on a property that you
actually want to persist (which is not a good idea anyway). You would need to
add another property with the inject annotation and pass the value over to the
persisted property in ``initializeObject`` for example.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Reflection/ReflectionService.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Reflection/ReflectionService.php
@@ -1607,6 +1607,10 @@ class ReflectionService
             return false;
         }
 
+        if ($this->isPropertyAnnotatedWith($className, $propertyName, Flow\InjectConfiguration::class)) {
+            return false;
+        }
+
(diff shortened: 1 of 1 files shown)


Benjamin-K:
Can this pr also be merged to Neos 2.3? Since Neos 2.3.12 images do not work correct any more and this PR will fix it. More on that here: https://github.com/neos/neos-development-collection/pull/1513

gerhard-boden:
@kitsunet Could you set the base to 2.2?

kitsunet:
This is Flow! Flow 3.3 is the matching for Neos 2.3

kitsunet:
And yes I would also try to put a test in place but as our current builds are broken I would like to have this merged ASAP and release a working tagged version.

kitsunet:
@gerhard-boden why 2.2? You mean Flow 3.2 and if so then why?

PR #932, opened 2017-04-07
TASK: Adjust schema of routes configuration

mficzel:
Previously the route parts were evaluated correct regardless of the configured keys because the schema used the configuration `additionalAttributes = FALSE` instead of `additionalProperties = FALSE`. This is fixed and the schema is updated 

- validate the keys in route parts strictly
- allow configuration of `objectType` and  `uriPattern ` instead of `handler`
- allow `options` and `toLowerCase` for all routeParts


mficzel:
I changed the target branch to 3.3 since 3.2 is in security fix only mode now.

mficzel:
This one can be tested via `./flow configuration:validate --type Routes`. After this change the configuration should still be valid but detect if an unexpected key is added in a routePart configuration.

PR #933, opened 2017-04-08
BUGFIX: Correctly document usage of ``...cookie.domain`` setting

kitsunet:
The cookie domain was allowed to be set with a leading dot to signal
that it should be valid for all subdomains. This was later changed in
the current RFC 6265. Our ``Session`` implementation checks for domains
without leading dot, while the comment suggested to use it. This is now
matched.

Fixes: neos/neos-development-collection#813


kitsunet:
https://github.com/neos/neos-development-collection/issues/813

PR #935, opened 2017-04-12
TASK: Use generateRandomString() instead of uniqid()

kdambekalns:
This reduces the risk of collision on temporary filenames and other
identifiers and data.


hmrdk:
🙏

kdambekalns:
Yes, just strange it didn't happen locally (using PHP 5.6 as well for the tests). Whatever…

PR #939, opened 2017-04-13
TASK: Update Readme.rst and Upgrading.rst for 4.1

kdambekalns:


Neos.Flow/Resources/Private/Installer/Distribution/Essentials/Upgrading.rst
```diff
@@ -125,5 +77,5 @@ in *Development Context*, padded with some manual checking and adjustments needs
 That should result in a working package.
 
 If it does not and you have no idea what to do next, please get in touch
-with us. The `support page <https://www.neos.io/support/>`_ provides more
```
aertmann: doesn't this needs to be fixed in a lower branch?

kdambekalns: Yes, it does. Also in the Readme badges, we might want to point to the correct branch on RTFD instead of latest. But I decided to stick to this simple thing for now…

PR #941, opened 2017-04-13
TASK: Add release notes for 4.1

### LINKED ISSUE #917: Write release notes for Flow 4.1

gerhard-boden:
Fixes #917


kitsunet:
Meh, I changed base, but that is not that good of an idea... Needs to be moved to 4.1 branch. I see if I can do that this WE, would directly merge then as I also approve

gerhard-boden:
@kitsunet correctly moved this to 4.1, will do the same for Neos now

PR #944, opened 2017-04-14
TASK: Improve code readability in Cache package

albe:


albe:
This is a first of multiple future follow-ups to #443

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -493,7 +505,7 @@ protected function readCacheFile($cacheEntryPathAndFilename, $offset = null, $ma
             if ($data !== false) {
                 return $data;
             }
-            usleep(rand(10, 500));
+            usleep(random_int(10, 500));
```
kitsunet: IMHO this is not a good idea because random_int is slower than rand and no cryptographically secure value is needed here.

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -528,7 +540,7 @@ protected function writeCacheFile($cacheEntryPathAndFilename, $data)
                 clearstatcache(true, $cacheEntryPathAndFilename);
                 return $result;
             }
-            usleep(rand(10, 500));
+            usleep(random_int(10, 500));
```
kitsunet: IMHO this is not a good idea because random_int is slower than rand and no cryptographically secure value is needed here.

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -226,31 +254,16 @@ public function remove($entryIdentifier)
...
+                $result = $this->tryRemoveWithLock($cacheEntryPathAndFilename);
 
                 if ($result === true) {
                     clearstatcache(true, $cacheEntryPathAndFilename);
                     return $result;
                 }
             } catch (\Exception $e) {
             }
-            usleep(rand(10, 500));
+            usleep(random_int(10, 500));
```
kitsunet: IMHO this is not a good idea because random_int is slower than rand and no cryptographically secure value is needed here.

albe: good point

PR #945, opened 2017-04-14

albe:


albe:
I'd like to start an effort to add PHP 7.0 scalar typehints to the codebase step by step.

daniellienert:
@albe: If these files are typed completely, should we also enable strict typing for them?

albe:
Should be ok to do so in theory, right? I don't really have experience with the strict type flag yet

PR #946, opened 2017-04-14

albe:


Neos.Utility.Arrays/Classes/Arrays.php
```diff
@@ -25,11 +25,11 @@
      * @param string $string The string to explode
      * @return array Exploded values, all converted to integers
      */
-    public static function integerExplode($delimiter, $string)
+    public static function integerExplode(string $delimiter, string $string): array
     {
         $chunksArr = explode($delimiter, $string);
-        while (list($key, $value) = each($chunksArr)) {
-            $chunks[$key] = intval($value);
+        foreach ($chunksArr as $key => $value) {
```
albe: A foreach is a lot faster than the previous implementation. Was there a specific reason for choosing that?

Neos.Utility.Arrays/Classes/Arrays.php
```diff
@@ -25,11 +25,11 @@
...
+            $chunks[$key] = (int)$value;
```
albe: Again optimization, (int) cast is faster than `intval()` call

Neos.Utility.Arrays/Classes/Arrays.php
```diff
@@ -150,13 +150,13 @@ public static function arrayMergeRecursiveOverruleWithCallback(array $firstArray
...
      */
-    public static function containsMultipleTypes(array $array)
+    public static function containsMultipleTypes(array $array): bool
     {
         if (count($array) > 0) {
             reset($array);
             $previousType = gettype(current($array));
             next($array);
-            while (list(, $value) = each($array)) {
+            foreach ($array as $value) {
```
albe: See above

Issue #934, opened 2017-04-08
Custom configuration types break the compilation process


KaiLehmann:
### Description

Registering a custom configuration type produces a PHP undefined index error and breaks the compilation process.

This can be fixed in \Neos\Flow\Configuration\ ConfigurationManager:: loadConfiguration() in line 459 by replacing
`if ($this->configurations[$configurationType] !== []) {`
with
`if (isset($this->configurations[$configurationType])) {`

I tried to do a pull request, but my GIT knowledge is too bad and it somehow didn't work ...

### Steps to Reproduce

1. Register a custom configuration type like so:
```
class Package extends \Neos\Flow\Package\Package {
	public function boot(\Neos\Flow\Core\Bootstrap $bootstrap) {
		$dispatcher = $bootstrap->getSignalSlotDispatcher();
		$dispatcher->connect(\Neos\Flow\Configuration\ConfigurationManager::class, 'configurationManagerReady',
			function (\Neos\Flow\Configuration\ConfigurationManager $configurationManager) {
				$configurationManager->registerConfigurationType(
					'MyCustomConfiguration',
					\Neos\Flow\Configuration\ConfigurationManager::CONFIGURATION_PROCESSING_TYPE_SETTINGS
				);
			}
		);
	}
}
```
2. Flush all caches
3. CLI: ./flow or load any controller

#### Expected behavior

The custom configuration type is registered.

#### Actual behavior

```
Notice: Undefined index: MyCustomConfiguration in /xyz/Packages/Framework/Neos.Flow/Classes/Configuration/ConfigurationManager.php line 459

Type: Neos\Flow\Error\Exception
Code: 1
File: Packages/Framework/Neos.Flow/Classes/Error/ErrorHandler.php
Line: 81

Open Data/Logs/Exceptions/20170408101204dfb1f2.txt for a full stack trace.

Exception Code	1355480641
Exception Type	Neos\Flow\Core\Booting\Exception\SubProcessException
Log Reference	2017040810120348dee0
Thrown in File	Packages/Framework/Neos.Flow/Classes/Core/Booting/Scripts.php
Line	632
```

### Affected Versions

Flow: 4.*

### More information

This is a little difficult to debug since it happens during compilation. The following happens during compilation after all caches have been flushed:

1. CacheManager::flushConfigurationCachesByChangedFiles() calls $this->configurationManager->refreshConfiguration();
2. ConfigurationManager::refreshConfiguration() calls $this->saveConfigurationCache();
[...]

PR #948, opened 2017-04-19
BUGFIX: ConfigurationManager should never have empty configuration

kitsunet:
Custom configuration types of processing type settings could break
the compile process as the array wouldn't have the custom configuration
name as key set because it was never initialized in this processing
type.

This change initializes any configuration type with an empty array in
the beginning and provides a test to proof the behavior.

Fixes: #934

### DIFF
--- a/Neos.Flow/Classes/Configuration/ConfigurationManager.php
+++ b/Neos.Flow/Classes/Configuration/ConfigurationManager.php
@@ -423,6 +423,7 @@ class ConfigurationManager
      */
     protected function loadConfiguration($configurationType, array $packages)
     {
+        $this->configurations[$configurationType] = [];
         $this->cacheNeedsUpdate = true;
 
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -423,6 +423,7 @@ public function warmup()
      */
     protected function loadConfiguration($configurationType, array $packages)
     {
+        $this->configurations[$configurationType] = [];
```
bwaidelich: Is it impossible that `$this->configurations[$configurationType]` is already set at that point?

kitsunet: Pretty much and if it was, it should be reloaded anyway.

bwaidelich: Makes sense I guess, thx

PR #950, opened 2017-04-23
FEATURE: Undeclared tag ViewHelper arguments are rendered

kitsunet:
It is now possible to declared unregistered arguments for tag
based ViewHelpers in templates.
Those will be rendered to the tag as simple key/value.
If the value cannot be cast to string an exception will be thrown.

Example:

    <f:link.action action="index" my-attribute="test">

Results:

    <a href="path/to/index/action" my-attribute="test">

Previously an exception would be thrown for any undeclared argument given.

Resolves: #639

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractTagBasedViewHelper.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractTagBasedViewHelper.php
@@ -11,7 +11,6 @@ namespace Neos\FluidAdaptor\Core\ViewHelper;
  * source code.
  */
 
-use Neos\Flow\Annotations as Flow;
 use TYPO3Fluid\Fluid\Core\ViewHelper\TagBuilder;
 
 /**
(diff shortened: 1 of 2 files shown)


kitsunet:
It's a possible solution for the feature. Let's discuss pros and cons. Anything that cannot be converted to string is caught in the validation step, so I think that's fine. Also I don't think it's a _severe_ change, but I would like to hear opinions. IMHO it's fine like this, you can easily check the resulting markup and see if there was a typo. Also the example doesn't hold as you would get an exception from the VH because the required argumetn is missing...
We could add a flag to enable this behavior, but IMHO then you need to check for every single VH if it's enabled or not, which defeats the purpose of easy handling of additional arguments for tags.

bwaidelich:
> Anything that cannot be converted to string is caught in the validation step

True, I missed that part!

> I don't think it's a severe change

That wasn't meant judgemental - It's a change that affects the core behavior of Fluid. That's what I meant.

> We could add a flag to enable this behavior

Please no, lets agree on an expected behavior.
My personal opinion is: Let's extend the core helpers, add all allowed attributes from the HTML(5) spec  (`data` is already there and accepts any array) and look into the extensible approach by Marc for custom ViewHelpers.
But I'm happy to be convinced (and I'd still be interested in "TYPO3 Fluid"s stand on that - as it affects the way Fluid templates are interpreted)

kitsunet:
ok, added an RFC issue to typo3 fluid, let's wait and see.

kitsunet:
Btw: https://github.com/TYPO3/Fluid/issues/288#issuecomment-302385456

Claus also likes this very much. Maybe we simply wait for fluid to support it directly.

albe:
+1 for waiting for direct fluid support.

kitsunet:
Waiting for fluid support...

PR #953, opened 2017-04-24
TASK: Throw error if package exists two times

kitsunet:
Without this change a package (defined by composer name eg. "neos/party")
that exists two times in the installation would be silently ignored and added
with the second path found.	This could still result in strange effects
happening as autoload and Flow internal paths (configuration, resources) could
diverge.

With this change an exception is thrown when the package states are resolved
and a package is found a second time.

Fixes: #611


aertmann:
@kitsunet: did you see the tests fail?

kitsunet:
Didn't come around to check back, apparently we produce the errror in the tests 😆

kitsunet:
This should do the trick.

PR #954, opened 2017-04-24
BUGFIX: Integer argumentName should not result in an exception

kitsunet:
This prevents an exception in case an argument name is evaluated to
another simple type (integer) but can be converted to string easily.

Fixes: #612

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ActionRequest.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ActionRequest.php
@@ -469,11 +469,12 @@ class ActionRequest implements RequestInterface
      */
     public function setArgument($argumentName, $value)
     {
-        if (!is_string($argumentName) || strlen($argumentName) === 0) {
+        $argumentName = (string)$argumentName;
+        if (empty($argumentName)) {
             throw new Exception\InvalidArgumentNameException('Invalid argument name (must be a non-empty string).', 1210858767);
         }
 
-        if (substr($argumentName, 0, 2) === '__') {
+        if (strpos($argumentName, '__') === 0) {
             $this->internalArguments[$argumentName] = $value;
             return;
         }
@@ -482,7 +483,7 @@ class ActionRequest implements RequestInterface
             throw new Exception\InvalidArgumentTypeException('You are not allowed to store objects in the request arguments. Please convert the object of type "' . get_class($value) . '" given for argument "' . $argumentName . '" to a simple type first.', 1302783022);
         }
 
-        if (substr($argumentName, 0, 2) === '--') {
+        if (strpos($argumentName, '--') === 0) {
             $this->pluginArguments[substr($argumentName, 2)] = $value;
             return;
         }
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ActionRequest.php
```diff
@@ -469,11 +469,12 @@ public function getControllerActionName()
...
     public function setArgument($argumentName, $value)
     {
-        if (!is_string($argumentName) || strlen($argumentName) === 0) {
+        $argumentName = (string)$argumentName;
+        if (empty($argumentName)) {
             throw new Exception\InvalidArgumentNameException('Invalid argument name (must be a non-empty string).', 1210858767);
         }
 
-        if (substr($argumentName, 0, 2) === '__') {
+        if (strpos($argumentName, '__') === 0) {
```
bwaidelich: Is this related? What's better with this approach, performance?

bwaidelich:
OK, I checked this again and I would suggest the following:
Just remove the `!is_string($argumentName)` Check in the method and add a unit test like
```php
/**
 * @test
 */
public function setArgumentSupportsNumericArgumentNames()
{
    $this->actionRequest->setArgument(123, '');
    $this->assertSame('', $this->actionRequest->getArgument(123));
}
```

Regarding the `substr()` to `strpos()` change I'd like to understand the reasoning as it's quite a critical place.
We did change this from the direct array-accessor with https://github.com/kitsunet/flow-development-collection/commit/4f5f91a7d7ceff5ae58459cdc8bd7e19eebc644c  (And I just realized that the corresponding unit test got lost on the way(!))..

TYPO3.Flow/Classes/TYPO3/Flow/Mvc/ActionRequest.php
albe: Yep, according to http://maettig.com/code/php/php-performance-benchmarks.php#startswith `strpos` is faster, though `strncmp($argumentName, '__', 2)` would be even faster then.

I wonder how `strpos` performs in a non-matching case though, because obviously it would need to go through the whole input string then.

kitsunet: Was just a speed think that I apply whenever I pass by such a piece. We don't need the substr all we need to know is if the first two chars match. But I can change that back if you feel too uncomfortable with ti.

kitsunet:
ok @bwaidelich @kdambekalns I challenge the comment(s), IMHO all argument names should be strings. That it comes in as int is probably a result somewhere in routing or so. I don't see a good reason for keeping it an int. Therefore I would keep that code as is with the change (to cast to string) but I can obviously add a test if we agree on this.

kdambekalns:
I'd love a test, obviously. And I agree, all argument names are strings in PHP, just calling `phpinfo.php?123` results in `$_GET['123']` in the output so it is seen as a string.

So, maybe find out why it becomes an integer in Flow and fix it there? But I think I'd be fine with the current solution as well… 🤷‍♂️

kitsunet:
@kdambekalns added a small test, I cannot figure how you can get an integer there though, unless routing produces it somehow \o/
Anyway this should solve it.

kitsunet:
@kdambekalns maybe unblock/merge with the test in place?

bwaidelich:
Thanks for the updates. FYI: You can see how to reproduce this in the bug report (#612). It's not the routing directly, but the additional arguments

PR #966, opened 2017-05-02
BUGFIX: FileSystemTargetTest shouldn't have side effects

kitsunet:
The test would create a Flow package in the current installation as the
package manager in the test was not prepared with a virtual filesystem.


kitsunet:
I know this is EOL, but the test is borken an that sounds like not a good idea to leave it... So we need to upmerge this once after merge... Note this is not showing as test failure without the fix, but if you run it locally you will have a broken install.

Issue #900, opened 2017-03-21
Syntax Error for nullable return type with PHP 7.1 and Flow 4.0.2


p34ch:
I am writing a Flow application using PHP 7.1.0. In one of my domain models (postalAddress) I have a property named "postbox" for entering the postbox number of a users postal address.

The corresponding setter method looks like this:

    /**
     * @return string
     */
     public function getPostBox(): string
     {
         return $this->postBox;
     }

With the return type `string` this method MUST return a valid string (empty strings are allowed), even if the user has no postbox. With PHP 7.1 nullable return types may be defined by preceding the return type with a question mark: http://php.net/manual/en/migration71.new-features.php

    /**
     * @return string
     */
     public function getPostBox(): ?string
     {
         return $this->postBox;
     }

When I create an instance of this entity type in a commandcontroller command, I get this syntax error:
> syntax error, unexpected '?'

For testing purposes I created a simple test.php file in my localhost directory. In this file I wrote a single function for testing the return type definition. Both variants (with and without "?") for string return types worked perfectly fine in browser and on command line as well.

**==> Seems to be, this is a flow problem, isn't it?**

I'll attach you guys the complete error log below:

> Exception #1355480641 in line 321 of /Applications/MAMP/htdocs/volleymanager.local/Packages/Framework/Neos.Flow/Classes/Core/Booting/Scripts.php: syntax error, unexpected '?'
>
>   Type: ParseError
>   File: Packages/Application/SportManager.Core/Classes/Domain/Model/PostalAddress.p
>         hp
>   Line: 90
>
> 13 Neos\Flow\Core\Booting\Scripts::executeCommand("neos.flow:core:compile", array|16|)
> 12 Neos\Flow\Core\Booting\Scripts::initializeProxyClasses(Neos\Flow\Core\Bootstrap)
> 11 call_user_func(array|2|, Neos\Flow\Core\Bootstrap)
> 10 Neos\Flow\Core\Booting\Step::__invoke(Neos\Flow\Core\Bootstrap)
> 9 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
> 8 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
[...]

PR #967, opened 2017-05-02
BUGFIX: Return type declaration should be FQCN

### LINKED ISSUE #963: Incorrect FQCN for return types
### Description

When a class has the annotation 
```
/**
 * @Flow\Scope("session")
 */
```
and a method has an expected return type set like 
```
public function getNotificationPreferences(): \GesagtGetan\MyProjectAt\Dto\NotificationPreferences
````
the return type within the resulting proxy class is faulty 
```
public function getNotificationPreferences(): GesagtGetan\MyProjectAt\Dto\NotificationPreferences
```

if no FQCN is used, the import statement in the proxy class is missing.

#### Expected behavior

The same FQCN should be used in the proxy, otherwise an exception is thrown, because the return type is not compatible. If just a class name is given, the import statement should not be missing in the proxy class.

##### Used versions:
Neos: 3.0.3
Flow: 4.0.6

### PR DESCRIPTION
**Pull request by kitsunet**
Fixes #963

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php
@@ -14,6 +14,7 @@ namespace TYPO3\Flow\Object\Proxy;
 use Doctrine\ORM\Mapping as ORM;
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Reflection\ReflectionService;
+use TYPO3\Flow\Utility\TypeHandling;
 
 /**
  * Representation of a method within a proxy class
@@ -154,6 +155,10 @@ class ProxyMethod
         $visibility = ($this->visibility === null ? $this->getMethodVisibilityString() : $this->visibility);
 
         $returnType = $this->reflectionService->getMethodDeclaredReturnType($this->fullOriginalClassName, $this->methodName);
+        // TODO: Revise "TypeHandling::isSimpleType" and check if we want to add null.
+        if ($returnType !== null && !TypeHandling::isSimpleType($returnType) && !in_array($returnType, ['self', 'null', 'callable'])) {
+            $returnType = '\\' . $returnType;
+        }
         $returnTypeDeclaration = ($returnType !== null ? ' : ' . $returnType : '');
 
 

### CODE REVIEW

**Comment by kitsunet** [comment]
Ok, this is probably the right solution to the problem. But obviously we might want a test for this, which is hard to create. Also this calls for the question on how to (correctly) handle absolute class names. And finally this is obviously wrong for scalar return type declarations so we need to distinguish those and I am not exactly sure on how we can do that in the best way....

**Comment by kdambekalns** [comment]
Well, scalar types could be tested against a list of allowed scalar type names, easy. PHP itself won't change too fast to make this non-viable.

As far as absolute class names are concerned, `getMethodDeclaredReturnType()` should return the "correct" string already. So it could

- return a string already containing a leading backslash as needed, so the value can be used as is. That would violate the PHP core principle of classnames in string always being absolute and not having a leading backslash, though.
- always return a FQCN without the leading backslash, so it can always be prepended with one as needed. That would mean it has to expand imported classnames as needed.
- return imported classnames as is and indicate they don't need a backslash prefixed. This would complement either solution from above.

**Comment by kitsunet** [comment]
Currently it returns (I tested it with various examples) always the `FQCN without the leading backslash` regardless of how you coded it. I think that is generally fine although we could must easier distinguish if it would return the `FQCN WITH leading backslash` but that would be a breaking change of course and probably not in line with interfaces and parent class (although there it's guaranteed that it is a class name). So I guess I do the check in the proxy method and apply leading backslashes as needed.

**Comment by kdambekalns** [comment]
Ok, if it returns the FQCN "as common in PHP" we can indeed add a leading backslash. The missing piece would be to not do it, if the type is one of the scalar ones (checking for no backslash is not enough, because of classes like `Traversable` or `DOMNode`…)

**Comment by kdambekalns** [comment]
So, something like this could be added on 4.0 and up (since that's PHP7):

    const SCALAR_TYPES = ['array', 'callable', 'bool', 'float', 'int', 'string', 'null', 'self'];
    …
    $returnTypeDeclaration = (in_array($returnType, self::SCALAR_TYPES) ? $returnType : ' : \\' . $returnType);

**Comment by kitsunet** [comment]
that list exists in the TypeHandling Utility (however that is called, too lazy to check).

**Comment by kdambekalns** [comment]
Yeah, but it doesn't contain `self` (and `null`?) I think. Given that, the name is even wrong… :)

**Comment by kitsunet** [comment]
so amended commit should be pretty fine, definitely better than what we have.

**Comment by gerhard-boden** [comment]
For 4.0 we should think about adding these here:
http://php.net/manual/en/migration71.new-features.php

File: TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php [hunk 1]
```diff
@@ -154,6 +155,10 @@ public function render()
         $visibility = ($this->visibility === null ? $this->getMethodVisibilityString() : $this->visibility);
 
         $returnType = $this->reflectionService->getMethodDeclaredReturnType($this->fullOriginalClassName, $this->methodName);
+        // TODO: Revise "TypeHandling::isSimpleType" and check if we want to add null.
```
bwaidelich: That TODO should be moved to the `TypeHandling:: isSimpleType()` method ;)
But IMO we open Pandora's box if we change that.
I'll address that and try to add a test for this with a fix for #900

PR #969, opened 2017-05-04
TASK: Add a distinct exception for existing, but invalid composer manifests

beheist:
Since I just ran into that one and it took me a while to figure out why Flow was throwing a "Composer Manifest does not contain a "name" field", I added a distinct exception for the case when a composer manifest does exist, but cannot be parsed from JSON.

### DIFF
--- a/Neos.Flow/Classes/Composer/ComposerUtility.php
+++ b/Neos.Flow/Classes/Composer/ComposerUtility.php
@@ -101,6 +101,10 @@ class ComposerUtility
         $json = file_get_contents($manifestPathAndFilename);
         $composerManifest = json_decode($json, true);
 
+        if($composerManifest === null){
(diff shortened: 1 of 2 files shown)


kitsunet:
Do we need an extra exception class though? I find that a bit too much.

beheist:
I added it because the code around also uses "semantic" error classes, but I'm equally fine with removing it and throwing a more generic exception, as long as the error message stays the same ;)

albe:
I have no real issue with the exception class, especially since theres also the `MissingPackageManifestException`. We could also just remove that now and replace with the little bit more generic `InvalidPackageManifestException` or even get both down to just `PackageManifestException`.

albe:
@kitsunet ok with merging as is or should we refactor the Exception classes?

kitsunet:
Now I am agian slightly inhappy to have pressed the button 😆 I would like to push a follow up to remove the `@api` from the new exception os we *could* refactor it. I still think that custom Exception classes for every single exception is too much noise and I doubt this one will be caught that often in the future.

albe:
Go for it :) :+1:

PR #971, opened 2017-05-11
TASK: Ignore invalid cookie names instead of throwing an exception

Torsten85:
This prevents the exception that is thrown when a cookie has an invalid name.

See the discussion here: https://discuss.neos.io/t/ignore-invalid-cookies-instead-of-throwing-an-exception/2234

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Headers.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Headers.php
@@ -12,6 +12,7 @@ namespace TYPO3\Flow\Http;
  */
 
 use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Flow\Log\SystemLoggerInterface;
 
 /**
  * Container for HTTP header fields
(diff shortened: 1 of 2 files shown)


kitsunet:
Perfect, thanks a lot for your persistence.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Headers.php
```diff
@@ -450,8 +465,16 @@ protected function setCookiesFromRawHeader($rawFieldValue)
...
-            if (trim($name) !== '' && preg_match(Cookie::PATTERN_TOKEN, $name) === 1) {
-                $this->setCookie(new Cookie(trim($name), urldecode(trim($value, "\t ;\""))));
+            $trimmedName = trim($name);
+
+            if ($trimmedName === '') {
+                continue;
+            }
+
+            if (preg_match(Cookie::PATTERN_TOKEN, $trimmedName) !== 1) {
+                $this->systemLogger->log(sprintf('Cookie name %s is not RFC 2616 compliant, therefore it will be skipped', $trimmedName), LOG_INFO);
```
aertmann: question is if it should throw an exception in development context, but ignore in production or just always ignore..

kitsunet:
That would mean you couldn't really test it in dev

aertmann:
true

PR #973, opened 2017-05-16
BUGFIX: Check for link in unpublishFile()

### LINKED ISSUE #972: FileSystemSymlinkTarget fails to unpublish correctly
### Description

When unpublishing a resource, the file in `Data/Persistent/Resources/…` is removed correctly, but the symlink pointing to it below `Web/_Resources/Persistent/…` is left in place.

### Steps to Reproduce

1. Publish a file
2. Unpublish file

Note: this can easily be done by using the replace function on an asset in the Neos media management.

#### Expected behavior

The symlink is removed together with the "real" resource file.

#### Actual behavior

The symlink stays in place.

This leads to the rewrite rules "seeing" the requested file, even though it can non longer be served. One effect of this is redirects from the redirect handler are never even looked for.

### Affected Versions

Flow: 3.3.x and later

kdambekalns:
`unpublishFile()` in the `FileSystemTarget` does not remove symlinks
correctly, thus the method is adjusted in the `FileSystemSymlinkTarget`
to check using `is_link()` instead of `file_exists()`.

Fixes #972

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Target/FileSystemSymlinkTarget.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Target/FileSystemSymlinkTarget.php
@@ -11,8 +11,7 @@ namespace TYPO3\Flow\Resource\Target;
  * source code.
  */
 
-use TYPO3\Flow\Annotations as Flow;
-use TYPO3\Flow\Resource\Collection;
+use TYPO3\Flow\Error\Error;
 use TYPO3\Flow\Resource\CollectionInterface;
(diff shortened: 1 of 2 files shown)


kdambekalns:
I did not add a test here, since vfsStream has no support for symlinks (last time I checked, at least).

kdambekalns:
@kitsunet Which messages do you mean?

kdambekalns:
Well, I cannot think of a way that involves less code, TBH. Except handling both cases in the parent, but that feels dirty.

aertmann:
yeah it's a tough one, you could call the parent after first method, but that would mean it would do the first check twice of course

aertmann:
not blocking it btw.

Issue #974, opened 2017-05-17
Doctrine migrations are not "dry-runnable"


bwaidelich:
In some of the Flow doctrine migrations we execute SQL queries directly (for example to fetch all roles from the database to migrate them).
This is incompatible with the `--dry-run` and `--output` flags of the `doctrine:migrate` CLI command because even in "dry-run" mode, doctrine needs to execute migrations to determine the SQL they would generate.

Suggestion:
Don't execute *mutation* queries in dry-run and only run *select* queries if the respective table(s) actually exist.

Affected version: Almost all
Fix version: lowest supported (= Flow 3.3 LTS)

bwaidelich:
Fixed with #975

PR #975, opened 2017-05-17
BUGFIX: Doctrine migration SQL query safeguards

bwaidelich:
This adds `TABLE EXISTS` checks to all places where doctrine
migrations directly executes SQL queries in order to prevent
`./flow doctrine:migrate` to fail with `--dry-run` or `--output`
flag set.

Background:
Doctrine needs to execute the migrations in order to determine
the SQL they generate - even during "dry run".
In some migrations we access the database directly in order to
migrate data.
To avoid that these queries fail if the corresponding tables haven't
been created yet, this patch adds safeguards respectively.
NOTE: There are still a few places where we perform *mutations* in
the `postUp()`/`postDown()` hooks. Those will still be executed in
"dry run" mode if the table in question exists in the database!

Fixes: #974

### DIFF
--- a/TYPO3.Flow/Migrations/Mysql/Version20130319131400.php
+++ b/TYPO3.Flow/Migrations/Mysql/Version20130319131400.php
@@ -63,6 +63,9 @@ class Version20130319131400 extends AbstractMigration
      */
     protected function migrateAccountRolesUp()
     {
+        if (!$this->sm->tablesExist(['typo3_flow_security_account'])) {
+            return;
+        }
         $rolesSql = array();
         $accountRolesSql = array();
         $rolesToMigrate = array();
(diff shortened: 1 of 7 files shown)


kdambekalns:
It might be worth adding some output for those cases, since using the SQL emitted when using `--output` will not contain all needed statements. In fact, those migrations can only be run "for real", unfortunately. WDYT?

Other than that: thanks for taking care!

bwaidelich:
> It might be worth adding some output for those cases
Which cases do you mean exactly, the ones where a table is missing?
I don't think we have to deal with that case: They all migrate some data from deprecated tables. If those tables don't exist, there's nothing to migrate.
Or did I get you wrong?

aertmann:
well depends how it works, if the dry run only executes actual queries in the postUp / postDown methods, then it could be an earlier migration that would have created/renamed that table that it can't find and thus not showing that query in the dry run output

bwaidelich:
Right, we just can't guarantee that all cases are covered. But if an older migration that renames/creates a table hasn't been run yet, we don't have to care about migration of the respective data anyways..
I'd say there's no harm in merging this one, it improves the situation a littl eat least

bwaidelich:
@kdambekalns @aertmann Do you think that we can merge this one? (I'm not sure whether I completely got your remarks)

kdambekalns:
Well… the thing is, using `--dry-run` or `--output` will give different results that actually executing the migrations, in case those tables are coming into existence during the migration.

That being said, we have that problem anyway, so… this is better than before.

bwaidelich:
@kdambekalns OK, gotcha. But, still, I don't think that's an issue because the "data migration" queries will be empty anyways if a table came into existence during the migration. Cheers

PR #981, opened 2017-05-24
FEATURE: Make it possible to ignore annotations for a whole namespace

### LINKED ISSUE #980: Ignore Annotation namespaces
Right now it is not possible to disable a complete namespace in doctrine's annotation reader. It is possible to ignore single tags bot not a whole namespace.

johannessteu:
This change introduces a new configuration option `Neos.Flow.reflection.ignoredNamespaces`. This option can be used to configure doctrine’s AnnotationReader to ignore a whole namespace.

This configuration would ignore `@Foo\Bar` as well as `@Foo\Baz`:
```
Neos:
  Flow:
    reflection:
      ignoredNamepsaces:
        'Foo': true
```

Resolves #980

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -352,6 +352,12 @@ class ReflectionService
             }
         }
 
+        foreach ($this->settings['reflection']['ignoredNamespaces'] as $namespace => $ignoreFlag) {
+            if ($ignoreFlag === true) {
+                AnnotationReader::addGlobalIgnoredNamespace($namespace);
+            }
+        }
+
         $this->initialized = true;
     }
 
--- a/Neos.Flow/Configuration/Settings.yaml
+++ b/Neos.Flow/Configuration/Settings.yaml
@@ -369,7 +369,7 @@ Neos:
 
     reflection:
 
-      # A list of tags to be ignored during reflection
+      # A list of annotation tags to be ignored during reflection
       ignoredTags:
         'api': TRUE
         'package': TRUE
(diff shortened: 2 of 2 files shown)


Neos.Flow/Configuration/Settings.yaml
```diff
@@ -405,6 +405,9 @@ Neos:
         'WithoutSecurityChecks': TRUE
         'covers': TRUE
 
+      # A list of annotation namespaces to be ignored during reflection
+      ignoredNamespaces: []
```
albe: The new setting needs to be added to the schema

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -352,6 +352,12 @@ protected function initialize()
             }
         }
 
+        foreach ($this->settings['reflection']['ignoredNamespaces'] as $namespace => $ignoreFlag) {
```
albe: I think we also need to adjust the `isTagIgnored()` method, which is used to build the flow internal reflection data.

johannessteu: Yeah, we should also cover that with a test as well! I'll take care! Wasn't aware of that `isTagIgnored()` Method. What about a new `isNamespaceIgnored()` @albe ?

albe: There are already tests for the `isTagIgnored` method, so they can probably easily be extended/adjusted with namespace cases.

I'd assume Flow itself only cares/knows about ignoring single tags and `isTagIgnored` also checking against a namespace blacklist being an implementation detail, so I'd be in favour of keeping that in this single method. Otherwise we'd just move this complexity one level up in abstraction, which I see no real reason for.

albe:
*ping* Any news on finishing this up @johannessteu?

albe:
Is this still relevant? Otherwise I'd close this

albe:
Closing for now. Feel free to reopen if you feel the itch to work on this again.

PR #982, opened 2017-05-24
BUGFIX: Fix support for nullable return type declarations

bwaidelich:
Fixes: #900


bwaidelich:
This breaks functional tests with PHP < 7.1 because the compilation step fails (`syntax error, unexpected '?'`)

TYPO3.Flow/Classes/TYPO3/Flow/Object/Proxy/ProxyMethod.php
```diff
@@ -155,10 +155,6 @@ public function render()
         $visibility = ($this->visibility === null ? $this->getMethodVisibilityString() : $this->visibility);
 
         $returnType = $this->reflectionService->getMethodDeclaredReturnType($this->fullOriginalClassName, $this->methodName);
-        // TODO: Revise "TypeHandling::isSimpleType" and check if we want to add null.
-        if ($returnType !== null && !TypeHandling::isSimpleType($returnType) && !in_array($returnType, ['self', 'null', 'callable'])) {
-            $returnType = '\\' . $returnType;
-        }
```
bwaidelich: I moved the logic to the ReflectionService to be in sync with related code

TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetClassWithPhp71Features.php
```diff
@@ -0,0 +1,33 @@
...
+ * This file is part of the TYPO3.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+class TargetClassWithPhp71Features
```
bwaidelich: This is the culprit that breaks PHP < 7.1 support

TYPO3.Flow/Tests/Functional/Aop/Fixtures/TargetClassWithPhp7Features.php
```diff
@@ -21,10 +21,28 @@ class TargetClassWithPhp7Features
...
-        return "{$aString} and {$aNumber}";
+        return "{$aString} and {$aNumber} and {$anObject}";
     }
+
+    public function methodWithStaticScalarReturnTypeDeclaration(): string
+    {
+        return 'it works';
+    }
+
+    public function methodWithStaticObjectReturnTypeDeclaration(): TargetClassWithPhp7Features
```
bwaidelich: This implicitly tests #967 as well

bwaidelich:
I wonder if we should remove the tests in order to get this in.. At least I have no other idea on how we could solve the PHP < 7.1 issue..

kitsunet:
I guess I would be ok with that :)

albe:
Yes, fine with that. And let's increase PHP version requirement with Flow5

bwaidelich:
@aertmann hehe. I tried everything I could think of for several hours (including evil eval) but nothing worked out, so I just went for the easy way out for now

TYPO3.Flow/Classes/TYPO3/Flow/Reflection/MethodReflection.php
```diff
@@ -107,6 +107,18 @@ public function getDeclaredReturnType()
     }
 
     /**
+     * @return bool Whether a return type is declared (@see getDeclaredReturnType()) and has a nullable modified ("?")
```
kdambekalns: probably "a nullable modifier" was intended?

bwaidelich: Right, I'll push a follow-up and merge it. Thanks!

bwaidelich:
Merging with previous votes (just rebased and fixed a typo afterwards)

PR #984, opened 2017-05-30
TASK: Add missing documentation for trait introduction

kitsunet:
Trait introduction is available since some time, but it was never
documented. This change adds a short explanation with a code
example to fix that.

### DIFF
--- a/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/AspectOrientedProgramming.rst
+++ b/TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/AspectOrientedProgramming.rst
@@ -984,7 +984,7 @@ be a fully qualified "class" (trait) name without leading backslash.
 
 The following example introduces a trait ``SomeTrait`` to the class ``MyClass``.
 
-*Example: Interface introduction*::
+*Example: Trait introduction*::
 
 	namespace Example\MyPackage;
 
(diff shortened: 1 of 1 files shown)


TYPO3.Flow/Documentation/TheDefinitiveGuide/PartIII/AspectOrientedProgramming.rst
```diff
@@ -971,6 +971,34 @@ The following example introduces a new interface ``NewInterface`` to the class
...
+
+Like the interface introductions, also trait introductions are declared
+by annotation. It even uses the same annotation with a different argument:
+
+``@Flow\Introduce("PointcutExpression", traitName="NewTraitName")``
+
+Again only pointcuts filtering for classes make sense. The traitName must
+be a fully qualified "class" (trait) name without leading backslash.
+
+The following example introduces a trait ``SomeTrait`` to the class ``MyClass``.
```
aertmann: I tried to create a trait and add the annotation to the trait class itself, which didn't work. Could imagine others would try to the same to avoid having two classes to introduce a trait. Maybe we should mention that?

kitsunet: Mmm, but that is the same for interface introduction. The example says you need to create an (aspect) class.

kdambekalns: Right. I'd rather create a feature request for making introduction possible right in an interface or trait.

PR #985, opened 2017-05-31
BUGFIX: Keep format for URIs built in subrequests

hlubek:
Adds handling for the `@format` argument when merging arguments
from the request hierarchy to keep the format when building
URIs from a sub request.

Fixes neos/neos-development-collection#1596


hlubek:
@aertmann I'm not sure either, but I added the tests to check if it's somehow unwantedly propagating the format up to the main request, but IMHO it should work as before, but the bug I mentioned is fixed. This could mean if you relied on that bug it will break.

hlubek:
@albe Exactly that was my thinking, too. One case that could break is that a URL generated by a subrequest _should_ "loose" the format from the outer request. But why?

hlubek:
@robertlemke I'll merge this for now, if you see any issue we could revert this (until the next release ofc)

PR #991, opened 2017-06-11
!!!TASK: Add PHP 7.0 scalar type hints to method arguments and return types

albe:
This is breaking because it changes the cache interfaces and all implementations need to add scalar typehints.
The scope should be very limited though, since the amount of custom cache implementations is assumingly low. The most likely break will be with custom `*Privilege` implementations that also override the `getCacheEntryIdentifier()` method.


bwaidelich:
This seems to break unit tests (https://travis-ci.org/neos/flow-development-collection/jobs/243879407) or is that unrelated?

albe:
That's why it's marked as breaking !!! ;) Nah j/k, that's correct. Will check what happens this weekend.

Neos.Cache/Classes/Backend/BackendInterface.php
```diff
@@ -56,7 +56,7 @@ public function getPrefixedIdentifier($entryIdentifier);
      * @throws \Neos\Cache\Exception\InvalidDataException if $data is not a string
```
kdambekalns: With `$data` being declared as `string` that exception can never be thrown anymore, no? Might be the case for other places as well…

albe: Good catch

albe:
Rebased, removed the obsolete InvalidDataException hints and this should fix the test :crossed_fingers:

kitsunet:
We should actually provide a core migration for updating implementations of `CacheAwareInterface`

albe:
How would that migration look? A regex replacement for the method name `getCacheEntryIdentifier`? Isn't that a bit too generic?

kitsunet:
@albe I think we have code to detect implementations via lexer, so we could prefilter classes implementing the interface and then regex for the method, that should be quite safe. Can try to do that now.

kitsunet:
Ok, this is not what I had in mind but that would require change to the migration system. This should be good enough to catch 99% of all cases. Will merge in 24h if travis is happy and no objections.

PR #993, opened 2017-06-13
BUGFIX: Add handling of HTTP HEAD-request with curl

### LINKED ISSUE #992: Http-Request with Type HEAD runs into Timeout
### Description

When executing a HTTP-Request with type HEAD to an elasticsearch instance (Version 5.4+) curl waits for the content and runs into the timeout, even if the request ist completed.

We are using Flowpack/Elasticsearch for the connection. This package worked well until recently, when it was no more possible to create or delete an index. 

After debugging the issue, I came to the conclusion, that the problem lies in the usage of curl and HEAD request type. This problem was not visible since the last version of elasticsearch, where a change to the rest-api to fix the formaly malformed behavior occured. 
Here is the Issue describing the changes:
https://github.com/elastic/elasticsearch/pull/23112

In short, elasticsearch returns a content-length greater than 0, which I think causes curl to wait for the content which never arrives.

I will issue a Pull-Request to fix this.

#### Expected behavior
HTTP-Request with type HEAD executes and does not run into a timeout.

#### Actual behavior
The request runs into a timeout.

### Affected Versions
Flow: 4.0.5

thomasblass:
Extended CurlEngine for request-type HEAD to set the option to not expect a body.

resolves #992

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
@@ -107,6 +107,13 @@ class CurlEngine implements RequestEngineInterface
                     ]);
                 }
             break;
+            case 'HEAD':
+                if (!is_resource($content)) {
+                    $body = $content !== '' ? $content : http_build_query($request->getArguments());
+                    curl_setopt($curlHandle, CURLOPT_POSTFIELDS, $body);
+                }
+                curl_setopt($curlHandle, CURLOPT_NOBODY, true);
+            break;
             default:
                 if (!is_resource($content)) {
                     $body = $content !== '' ? $content : http_build_query($request->getArguments());


kitsunet:
Cool, that looks like a good idea :)
As this is a bugfix and I would agree to merge this as such, it could go into the 3.3 branch.

thomasblass:
Does it mean I have to create a new pull-request against the 3.3 branch?

Neos.Flow/Classes/Http/Client/CurlEngine.php
```diff
@@ -107,6 +107,14 @@ public function sendRequest(Http\Request $request)
                     ]);
                 }
             break;
+            case 'HEAD':
+                if (!is_resource($content)) {
+                    $body = $content !== '' ? $content : http_build_query($request->getArguments());
+                    curl_setopt($curlHandle, CURLOPT_POSTFIELDS, $body);
+                }
+                curl_setopt($curlHandle, CURLOPT_CUSTOMREQUEST, $request->getMethod());
```
kdambekalns: Not needed, since `CURLOPT_NOBODY` sets the method to `HEAD` automatically, see https://secure.php.net/manual/en/function.curl-setopt.php:
> CURLOPT_NOBODY: TRUE to exclude the body from the output. Request method is then set to HEAD.

kdambekalns:
Just now saw the question of @thomasblass - yes, you need to cherry-pick the change (or manually re-do it) on 3.3 and create a new PR (or force-push to your `head-request-timeout` branch.)

thomasblass:
Thanks for the hint @kdambekalns. I changed the code and rebased the pull-request to 3.3 @kitsunet

kdambekalns:
Closed/reopened to start Travis CI…

albe:
:heart: CURL with it's infinite options and how they play together.

According to manpage (http://www.manpagez.com/man/3/CURLOPT_NOBODY/CURLOPT_NOBODY-7.38.0.php), CURLOPT_NOBODY sets the request method to `HEAD` (for HTTP(s)) and prevents to **download/receive** the body. So I'm not sure this actually interferes with CURLOPT_INFILE, which specifies the file for the **upload**, does it?

Also, according to HTTP spec (https://tools.ietf.org/html/rfc7231#section-4.3.2)
> A payload within a HEAD request message has no defined semantics;
   sending a payload body on a HEAD request might cause some existing
   implementations to reject the request.

So should we maybe issue/log a warning if content is non-empty?

But another thing: According to HTTP spec, a `HEAD` request is the same as a `GET`, except that it doesn't return a response body. So shouldn't the HEAD request also ignore the request arguments for building the CURLOPT_POSTFIELDS?

kitsunet:
Ok, so that is indeed a bit confusing :D Got it now, good to go I think. Merging.

Issue #994, opened 2017-06-14
Rescanning packages deletes persistent caches


bwaidelich:
### Description

When the package state has been changed, i.e. by executing the `./flow neos.flow:package:rescan` command, caches are flushed correctly.
Unfortunately this also removes any _persistent_ caches though, including the Flow `encryptionKey`.
This also happens if no packages have actually changed.

### Steps to Reproduce

1. Make sure an `encryptionKey` has been generated, by default at `Data/Persistent/Cache/Data/Flow_Security_Cryptography_HashService/encryptionKey`
2. Execute `./flow neos.flow:package:rescan`

#### Expected behavior

The encryptionKey should still be there with the original value

#### Actual behavior

The encryptionKey was deleted.. (and any other persistent cache)

### Affected Versions

Flow: 3.3 (probably lower)

PR #995, opened 2017-06-14
BUGFIX: Don't flush persistent caches when package state changes

bwaidelich:
Previously the `bootstrapShuttingDown` signal was connected with the
`CacheManager::flushCaches()` method directly passing arguments of the
signal on. The first parameter, being the Boostrap `runLevel` flag was
casted to `true`, so that all caches were flushed.

This is fixed by explicitly calling `CacheManager::flushCaches()` in
a closure.

Fixes: #994

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Package.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Package.php
@@ -125,8 +125,10 @@ class Package extends BasePackage
             }
         });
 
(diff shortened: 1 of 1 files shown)


bwaidelich:
Writing a test for this would mean quite some work, but I tested it manually with the `encryptionKey` and a custom persistent cache

bwaidelich:
Thanks for the styling fixes

PR #997, opened 2017-06-16

albe:

### DIFF
--- a/Neos.Utility.Files/Classes/Files.php
+++ b/Neos.Utility.Files/Classes/Files.php
@@ -27,13 +27,12 @@ abstract class Files
      * @return string
      * @api
      */
-    public static function getUnixStylePath($path)
+    public static function getUnixStylePath(string $path): string
     {
         if (strpos($path, ':') === false) {
             return str_replace(['//', '\\'], '/', $path);
-        } else {
-            return preg_replace('/^([a-z]{2,}):\//', '$1://', str_replace(['//', '\\'], '/', $path));
         }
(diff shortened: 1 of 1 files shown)


Neos.Utility.Files/Classes/Files.php
```diff
@@ -432,16 +431,27 @@ public static function unlink($pathAndFilename)
...
+        if ($decimals === null) {
+            $decimals = 0;
+        }
+        if ($decimalSeparator === null) {
+            $decimalSeparator = '.';
+        }
+        if ($thousandsSeparator === null) {
+            $thousandsSeparator = ',';
+        }
+
```
bwaidelich: can you elaborate?
Why doesn't `public static function bytesToSizeString($bytes, int $decimals = 0, string $decimalSeparator = '.', string $thousandsSeparator = ','): string` work?

albe: Sure. Because we also want/need to accept nulls to not be breaking (see FilesTest::bytesToSizeStringDataProvider()), which we can not typehint with PHP < 7.2. So I guessed it would be better to typehint for now and place this workaround than not typehint those params (or change tests).

bwaidelich: OK, thanks. I wasn't aware of the fact that adding a type hint breaks compatibility with `null` :-/
But instead of that work around, personally I'd use:
```php
public static function bytesToSizeString($bytes, $decimals = 0, $decimalSeparator = '.', $thousandsSeparator = ','): string
```
We can't use the type hint for `$bytes` and no `void` return types yet anyways, so no problem in omitting some

PR #998, opened 2017-06-16
TASK: Skip symlink tests on windows when symlinking fails

albe:
On windows it is not unlikely that symlinking does not work due to missing access rights.
In that case a lot of the Files tests fail without any helpful notice and eventually lead
to the developer ignoring the failures, possibly overlooking real failures.

This change makes such symlink failures on Windows skip the test instead.

### DIFF
--- a/Neos.Utility.Files/Tests/Unit/FilesTest.php
+++ b/Neos.Utility.Files/Tests/Unit/FilesTest.php
@@ -52,7 +52,7 @@ class FilesTest extends \PHPUnit\Framework\TestCase
             return symlink($target, $link);
         } catch (\Exception $e) {
             if (DIRECTORY_SEPARATOR !== '/') {
-                $this->markTestSkipped('Your windows installation does not allow symlinks. Make sure these tests are run with appropriate rights.');
+                $this->markTestSkipped('Your Windows Installation does not allow the PHP process to create symlinks. Try running tests from an admin elevated command line.');
                 return false;
             }
             throw $e;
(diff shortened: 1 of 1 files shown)


Neos.Utility.Files/Tests/Unit/FilesTest.php
```diff
@@ -41,6 +41,25 @@ public function tearDown()
...
+     * @return boolean
+     * @throws \Exception
+     */
+    protected function trySymlink($target, $link)
+    {
+        try {
+            return symlink($target, $link);
+        } catch (\Exception $e) {
+            if (DIRECTORY_SEPARATOR !== '/') {
+                $this->markTestSkipped('Your windows installation does not allow symlinks. Make sure these tests are run with appropriate rights.');
```
bwaidelich: What about "Your Windows Installation does not allow the PHP process to create symlinks. Make sure that tests are executed with appropriate file permissions"?

albe:
> Did you actually test this on Windows with insuffizient permissions?

Yes, the description actually exactly describes my use-case. Those tests always failed for me (I never open an admin console to run the tests) and I started to just ignore those FilesTests, because I always assumed they were due to the missing rights for symlinks. Then I found there was a single actual failing test in between.

Neos.Utility.Files/Tests/Unit/FilesTest.php
albe: Maybe we could even be more concrete and state "Try running tests from an admin elevated command line".

albe:
Changed the message according to comment from Bastian.

PR #999, opened 2017-06-17
!!!TASK: Add PHP 7.0 scalar type hints to method arguments and return types

albe:
This is only breaking for classes implementing `LockStrategyInterface`, whose signature now requires scalar typehints.

### DIFF
--- a/Neos.Utility.Lock/Classes/FlockLockStrategy.php
+++ b/Neos.Utility.Lock/Classes/FlockLockStrategy.php
@@ -67,7 +67,7 @@ class FlockLockStrategy implements LockStrategyInterface
      * @return void
      * @throws LockNotAcquiredException
      */
-    public function acquire($subject, $exclusiveLock)
+    public function acquire(string $subject, bool $exclusiveLock)
     {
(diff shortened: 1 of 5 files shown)


Neos.Utility.Lock/Classes/FlockLockStrategy.php
```diff
@@ -132,7 +132,7 @@ protected function tryToAcquireLock($exclusiveLock)
      * @param boolean $exclusiveLock
      * @throws LockNotAcquiredException
      */
-    protected function applyFlock($exclusiveLock)
+    protected function applyFlock(bool $exclusiveLock)
```
kdambekalns: While we are at it, and since `void` as return type is only supported in PHP 7.1, why not add the missing `@return` annotation here? ;)

albe: Good call

albe:
I was thinking of doing a b/c version of this and the other PR if I still get it done before the release. That way we hopefully have "as much as possible typehinting" in the release without breaking stuff and can then add breaking typehint in interfaces and possibly even strict_types for next major

PR #1000, opened 2017-06-17

albe:


Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
```diff
@@ -173,7 +173,7 @@ public function mediaTypesWithAndWithoutParameters()
      * @test
      * @dataProvider mediaTypesWithAndWithoutParameters
      */
-    public function trimMediaTypeReturnsJustTheTypeAndSubTypeWithoutParameters($mediaType, $expectedResult)
+    public function trimMediaTypeReturnsJustTheTypeAndSubTypeWithoutParameters(string $mediaType, $expectedResult)
```
gerhard-boden: no typehint for `$expectedResult` here?

albe: Ah yes, the reason was that the dataProvider uses `null`, so that would have to be `string $expectedResult = null` for it to work then. Could be done though

gerhard-boden:
PR #1000 congrats @albe  😃

albe:
Whoop whoop! :)

PR #1005, opened 2017-06-26
BUGFIX: FileSystemTargetTest shouldn't create package stub

kdambekalns:
The FileSystemTargetTest was fixed earlier to no longer create a stub
of a package during test runs in the real filesystem see PR #966.)

The test still created a stub package, because the FileSystemTarget
was told to publish to a vfs location - but that option is only used
if initializeObject() is called, which in unit tests does not happen
automagically.


kdambekalns:
Thanks!

PR #1007, opened 2017-06-29
FEATURE: Add Eel String helper pregMatchAll

jonnitto:
This pull request add the Eel String helper `String.pregMatchAll`

### DIFF
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -198,7 +198,7 @@ class StringHelper implements ProtectedContextAwareInterface
         return $matches;
     }
 
-      /**
+    /**
      * Perform a global regular expression match (PREG style)
      *
      * Example::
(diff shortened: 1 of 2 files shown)


Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -198,6 +198,31 @@ public function pregMatch($string, $pattern)
         return $matches;
     }
 
+      /**
+     * Perform a global regular expression match (PREG style)
```
gerhard-boden: Fix docblock indentation

gerhard-boden:
Looks good except for one small indentation error

kitsunet:
This might be just me, but I feel if you need to use pregMatchAll you might want to have a specialised helper for the use case or do some more in PHP. Just a hunch though, maybe I need to see some actual fusion code with this to see how that would look like.

jonnitto:
In this case, I iterate over an SVG file string with symbols, and the goal was to get all the id in an array.

PR #1009, opened 2017-07-02
FEATURE: Allow Nullable action params to be annotated

### LINKED ISSUE #927: Property mapping for nullable parameters
Look at this example method:

```
/**
 * Helper action for generating a back button.
 *
 * @param Person|null $person
 */
public function fallbackAction(Person $person = null) { ... }
```
I implemented a fallback method which handles certain "exceptional" cases in my application logic. I may redirect to this action under various circumstances.

This doesn't work although since no property mapper for type "Person|null" can be found. Of course I may simple remove the null type hint for the parameter and everything works fine.

But I think it would be better to eliminate `|null` or `null|` out of the `@param` annotation for resolving property mappers.

albe:
This change allows action arguments to be annotated "|null" or "null|" when they are optional with default value `null`. Before the type conversion would fail because no matching type converter would be found.

Resolves #927

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/ActionController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/ActionController.php
@@ -29,6 +29,7 @@ use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 use Neos\Flow\Property\Exception\TargetNotFoundException;
 use Neos\Flow\Property\TypeConverter\Error\TargetNotFoundError;
 use Neos\Flow\Reflection\ReflectionService;
+use Neos\Utility\TypeHandling;
 
 /**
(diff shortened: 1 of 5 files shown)


kitsunet:
Shouldn't we actually read that from reflections and not docblock anymore? Given that ew only support 7.1 by now, Doing `function foo(Baz $bar = null)` should be giving us all the information we need?

albe:
Basically yes, but:

- there are still some kind of typehints that only work in docblocks (collection/generic types or plain `object`)
- not all code that runs on 7.1+ uses (nullable) typehints but instead uses the `|null` docblock hint and we should stay b/c with that

We might decide to go all out for Reflection-only in one of the next majors though, in which case all docblocks become pretty much eye-candy only.

PR #1010, opened 2017-07-02
BUGFIX: Make CLI subprocesses work on Windows

albe:
On Windows, the `PHP_BINDIR` constant will resolve to "C:\php" no matter what the actual php directory is, because this constant is decided at compile time. However, the `PHP_BINARY` constant will resolve to the current executable that runs the script at runtime, which at least for CLI will be the PHP executable.

### DIFF
--- a/Neos.Flow/Classes/Command/CoreCommandController.php
+++ b/Neos.Flow/Classes/Command/CoreCommandController.php
@@ -364,7 +364,7 @@ class CoreCommandController extends CommandController
      */
     protected function launchSubProcess()
     {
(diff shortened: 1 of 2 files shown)


kaystrobach:
please port that back to 3.3 as well

hosters like hetzner also compile somewhere on disk and then move that binary around.

so PHP_BINDIR returns: `/usr/local/php-56-cli/bin` 
while it should return the value of PHP_BINARY  `/usr/bin/php56.wrp.cli`

kitsunet:
I am not even sure why it didn't go into 3.3 in the first place...

albe:
That's somehow my lazy ass's fault for not paying attention on what branch to target. I'll take more care next time :)

Unless someone else comes first, I'll create at a backported PR for 3.3 this weekend

gerhard-boden:
I take the blame for merging into the wrong branch, so it's the fault of at least two people 😄

kaystrobach:
works like a charm now

kaystrobach:
thank you!

PR #1013, opened 2017-07-13
FEATURE: Add `range` method to Eel Array-helper

mficzel:
The range method is a wrapper around the php range function that creates a sequence of integers or characters as array. This is especially helpful to create paginations in pure fusion.

### DIFF
--- a/Neos.Eel/Tests/Unit/Helper/ArrayHelperTest.php
+++ b/Neos.Eel/Tests/Unit/Helper/ArrayHelperTest.php
@@ -444,4 +444,33 @@ class ArrayHelperTest extends \Neos\Flow\Tests\UnitTestCase
 
         $this->assertEquals($expected, $result);
     }
+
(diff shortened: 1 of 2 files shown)


Neos.Eel/Tests/Unit/Helper/ArrayHelperTest.php
```diff
@@ -444,4 +444,33 @@ public function flipWorks($array, $expected)
...
     }
+
+    public function rangeExamples()
+    {
+        return [
+            'array from one to three' => [
+                [1, 3],
+                [1, 2, 3]
+            ],
+            'array from one to severn in steps of two' => [
```
mstruebing: *seven

mficzel: The typo is fixed now ... thanks for finding

PR #1018, opened 2017-07-16

albe:


Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -567,7 +567,7 @@ public function md5($string)
      */
     public function length($string)
     {
-        return UnicodeFunctions::strlen($string);
+        return $string === null ? 0 : UnicodeFunctions::strlen($string);
```
albe: Since UnicodeFunctions no longer allow calling the methods with null subject (which btw. never was possible on HHVM - see https://3v4l.org/DUFhF#v560), but there are Eel StringHelper tests that expect length to allow null, this check is put in place. Anyone knows if Eel StringHelper has more such assumptions that the subject *may be* `null`?

Neos.Utility.Unicode/Classes/TextIterator.php
```diff
@@ -449,16 +449,19 @@ private function parseSubjectBySentence()
...
                 $whiteSpace .= ' ';
             }
-            if ($whiteSpace != '') {
+            if ($whiteSpace !== '') {
                 $this->iteratorCache->append(new TextIteratorElement($whiteSpace, $i, $count, true));
             }
             $i += $count;
 
-            if ($currentPart != '' && $j < count($delimitersMatches[0])) {
+            if ($j >= count($delimitersMatches[0])) {
```
albe: Avoid the double `count()` call and better to read IMO

Neos.Eel/Classes/Helper/StringHelper.php
kitsunet: No clue, but seems like a sensible solution to me for now.

PR #1019, opened 2017-07-16

albe:


albe:
Yep, was to be expected somehow with ObjectAccess though :) I will take care ASAP you fix things

Neos.Eel/Classes/FlowQuery/Operations/Object/PropertyOperation.php
```diff
@@ -49,17 +49,17 @@ class PropertyOperation extends AbstractOperation
      */
     public function evaluate(FlowQuery $flowQuery, array $arguments)
     {
-        if (!isset($arguments[0]) || empty($arguments[0])) {
+        if (!isset($arguments[0]) || empty($arguments[0]) || !is_string($arguments[0])) {
```
albe: Not 100% sure about this. Can this argument be for example an object that can be `toString()`'d?

kdambekalns: It shouldn't be, IMHO.

kdambekalns:
Enough of the waiting… :)

PR #1020, opened 2017-07-19
TASK: Add additional detection of changed package files

kitsunet:
`Package.php` files are used for specific Flow bootstrap behavior
in packages and the information if such a file exists is cached in
`PackageStates.php`. In development it can happen that a new file
is created or an existing one removed and that should be detected.

This change adds detection for change (creation) and removal of
`Package.php` files.

### DIFF
--- a/Neos.Flow/Classes/Package.php
+++ b/Neos.Flow/Classes/Package.php
@@ -12,6 +12,7 @@ namespace Neos\Flow;
  */
 
 use Neos\Flow\Package\Package as BasePackage;
+use Neos\Flow\Package\PackageManagerInterface;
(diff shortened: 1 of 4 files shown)


kitsunet:
Let's see how it goes with travis, looked good locally...
This is a 4.0 replacement for #341

kitsunet:
Nope, not really, but I think I have an idea...

kitsunet:
Ok, so the check via file monitor creates an infinite loop apparently and thus breaks the compile step. I am not sure how fix this yet...

kdambekalns:
Ok, but this is unrelated to the 3.2 vs 4.0 branch, right? Closing this again, we should continue in #341 IMHO.

PR #1021, opened 2017-07-21
BUGFIX: localeFormatType dateTime must use correct case

lorenzulrich:


albe:
Thank you for this. Such small errors in documentation can really be frustrating, so such fixes are highly appreciated!

Since the ViewHelper Reference documentation is generated from the ViewHelpers directly, this needs to be fixed inside the `DateViewHelper` render method docblock though.

PR #1022, opened 2017-07-23

albe:


albe:
This is WIP because I'm yet unsure how much of the Eel API expects to be able to work with dynamic types and where this breaks exactly. Most of the methods should in theory work correctly as long as `strict_types` isn't enabled, since PHP will automatically cast e.g. numeric strings to int/float that way. It will break for cases where an input of e.g. `sign('foo')` is supposed to evaluate to an actual value (`NaN`) instead of throw.

So if anyone with deep Eel knowledge can shed some light on this and how to proceed best, I'm happy about any tips/hints.

kitsunet:
Looking at the test failures and thinking about it, I guess we need to be very lenient on the helpers...

kdambekalns:
I guess we have that by now…

PR #1025, opened 2017-07-23

albe:


albe:
This is b/c version of #991 which should be safe to merge for 4.2

aertmann:
Hey just came across an issue related to this. There are several @api methods that have been changed, breaking b/c. Mentioning this since that shouldn't have happened until next major. However it's been so long that reverting it would probably be breaking for some projects as well, but let's try to avoid that in the future.

albe:
Hey, thanks for the hint, Aske! Anything specific that broke? Because this change intentionally left out the interfaces, so that only the concrete implementations added return typehints that should not have caused any noticeable effect, unless someone extends the implementation, but maybe I missed something.

aertmann:
@albe My specific case was extending a string cache implementation. Good to know the interfaces weren't adjusted. But yeah @api marked methods still works calling just fine with the added return type hint, but extending it breaks.

PR #1026, opened 2017-07-23

albe:


albe:
This is a b/c version of #999 that should be safe to merge for 4.2

Neos.Utility.Lock/Classes/FlockLockStrategy.php
```diff
@@ -146,7 +146,7 @@ protected function applyFlock($exclusiveLock)
      *
      * @return boolean TRUE on success, FALSE otherwise
      */
-    public function release()
+    public function release(): bool
```
albe: Apparently it's safe to typehint return types in implementations even though they're not present in the interface.

PR #1028, opened 2017-07-26
BUGFIX: Allow array and object literals to be padded with whitespace in Eel

### LINKED ISSUE #986: Eel parser fails if an object-literal starts with a whitespace
### Description

I just found that eel-expression object-literals that start with a whitespace after the leading curly-brace result in a exception from the parser. So wile this `{'foo' : 'bar' , 'bar' : 'baz'}` can be parsed this `{   'foo' : 'bar' , 'bar' : 'baz'}` cannot.

```Exception #1344513194 in line 86 of .../Data/Temporary/Development/Cache/Code/Flow_Object_Classes/Neos_Eel_CompilingEvaluator.php: Expression "{ 'foo' : 'bar' , 'bar' : 'baz'}" could not be parsed.```

This is especially problematic if you use multiline-eel in fusion and try to format the code of an object-literal.

### Steps to Reproduce

1. Use fusion to process an eel expression like  this one 
```
test = Neos.Fusion:Debug {
    value = ${{'foo' : 'bar' , 'bar' : 'baz'}}
}
```
2. Show the result
3. Add whitespace to the beginning of the object literal
```
test = Neos.Fusion:Debug {
    value = ${{   'foo' : 'bar' , 'bar' : 'baz'}}
}
```
4. Show the result again ... you will see an error instead

Another way to reproduce is to add whitespace to the test data in 
`\Neos\Eel\Tests\Unit\AbstractEvaluatorTest::objectLiteralExpressions`

#### Expected behavior

I expect both expressions to return an array [ 'foo' => 'bar' , 'bar' => 'baz' ] like JSON would

#### Actual behavior

The eel-expression object-literals that starts with a whitespace causes a parser error.

### Affected Versions

I think basically all Versions of Eel are affected.

### PR DESCRIPTION
**Pull request by grebaldi**
This change allows expressions like:

    { foo: "bar", bar: "baz" }

or

    {
      foo: "bar",
      bar: "baz"
    }

to be parsed correctly in EEL

The change extends the EEL PEG grammar (in `Eel.peg.inc`) to be tolerant towards
whitespace (as well as line wraps) in object and array literals.

This fixes #986

### DIFF
--- a/TYPO3.Eel/Classes/TYPO3/Eel/EelParser.php
+++ b/TYPO3.Eel/Classes/TYPO3/Eel/EelParser.php
@@ -803,67 +803,83 @@ function match_ProdCalculation ($stack = array()) {
 }
 
 
-/* ArrayLiteral: '[' < Expression? (< ',' > Expression)* > ']' */
+/* ArrayLiteral: '[' _ < Expression? (< _ ',' _ > Expression)* > _ ']' */
 protected $match_ArrayLiteral_typestack = array('ArrayLiteral');
 function match_ArrayLiteral ($stack = array()) {
 	$matchrule = "ArrayLiteral"; $result = $this->construct($matchrule, $matchrule, null);
-	$_163 = NULL;
+	$_167 = NULL;
 	do {
 		if (substr($this->string,$this->pos,1) == '[') {
 			$this->pos += 1;
 			$result["text"] .= '[';
 		}
-		else { $_163 = FALSE; break; }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by kdambekalns** [comment]
Updated the comment to fit our changelog (we should get rid of that template again…) Here is the review hint that @grebaldi left:

**How to verify it**
Try to use an expression in EEL like

```
{
  foo: "bar",
  bar: "baz"
}
```
which should work without error now.

PR #1029, opened 2017-07-26
FEATURE: Introduce ``ObjectAccess::instantiate``

kitsunet:
This static method allows instantiating a given class with
and array of arguments in an efficient way.

Most of this method was previously hidden in the ``ObjectManager``
of Flow but as the same code is replicated in other packages it
makes sense to open it as generic method for re-use.


kitsunet:
See eg. ImagineFactory in neos/imagine....

kdambekalns:
Is

> as the same code is replicated in other packages

really a widespread issue? I fear having this available will confuse the hell out of people, and using this should be really rare. No?

kitsunet:
As said, the imagine package uses it, and in my current work on image manipulation I also use it. I haven't checked if we duplicated it anywhere else...

kitsunet:
I don't think it's a big problem having this available, and it reduces code duplication so YAY?

ObjectConverter doesn't use the exact code but could also use it (with a slight performance benefit).

I just noticed the duplication and was unhappy with it. I can live without, so \o/

kdambekalns:
Job failure caused by merge, that removed the remote ref, it seems. Harmless.

PR #1031, opened 2017-07-28
TASK: Remove a "risky test" warning

kdambekalns:
A data provider named testSomething() is seen as a test, leading to
a warning about a test not doing any assertions.

Don't do that.


gerhard-boden:
Waiting for tests to finishing *fingers crossed* ;)

PR #1032, opened 2017-07-31
FEATURE: Add Date.formatCldr helper

zaveryukha:
This change adds CLDR formatting capability to the Date Eel helper, to match
the functionality of the `format.date` Fluid ViewHelper.

**P.S.** This PR has taken over from https://github.com/neos/flow-development-collection/pull/920
All the history can be found the. The original code was written by @litovchenko1, I
(@zaveryukha) have added the unit tests.

### DIFF
--- a/Neos.Eel/Classes/Helper/DateHelper.php
+++ b/Neos.Eel/Classes/Helper/DateHelper.php
@@ -13,7 +13,6 @@ namespace Neos\Eel\Helper;
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Eel\ProtectedContextAwareInterface;
-use Neos\Flow\I18n\Exception as I18nException;
 use Neos\Flow\I18n\Formatter\DatetimeFormatter;
 use Neos\Flow\I18n\Locale;
 use Neos\Flow\I18n\Service as I18nService;
@@ -73,15 +72,20 @@ class DateHelper implements ProtectedContextAwareInterface
     /**
      * Format a date to a string with a given cldr format
      *
-     * @param \DateTime $date
+     * @param integer|string|\DateTime $date
      * @param string $cldrFormat Format string in CLDR format (see http://cldr.unicode.org/translation/date-time)
      * @param null|string $locale String locale - example (de|en|ru_RU)
-     *
      * @return string
      */
     public function formatCldr($date, $cldrFormat, $locale = null)
     {
-        if (!$date instanceof \DateTimeInterface) {
+        if ($date === 'now') {
+            $date = new \DateTime($date);
+        } elseif (is_int($date)) {
+            $timestamp = $date;
+            $date = new \DateTime();
+            $date->setTimestamp($timestamp);
+        } elseif (!$date instanceof \DateTimeInterface) {
             throw new \InvalidArgumentException('"' . $date . '" could not be parsed by \DateTime constructor.');
         }
         if (empty($cldrFormat)) {
--- a/Neos.Eel/Classes/Helper/DateHelper.php
+++ b/Neos.Eel/Classes/Helper/DateHelper.php
@@ -80,18 +80,18 @@ class DateHelper implements ProtectedContextAwareInterface
     public function formatCldr($date, $cldrFormat, $locale = null)
     {
         if ($date === 'now') {
-            $date = new \DateTime($date);
+            $date = new \DateTime();
         } elseif (is_int($date)) {
             $timestamp = $date;
             $date = new \DateTime();
             $date->setTimestamp($timestamp);
         } elseif (!$date instanceof \DateTimeInterface) {
-            throw new \InvalidArgumentException('"' . $date . '" could not be parsed by \DateTime constructor.');
+            throw new \InvalidArgumentException('The given date "' . $date . '" was neither an integer, "now" or a \DateTimeInterface instance.');
         }
         if (empty($cldrFormat)) {
             throw new \InvalidArgumentException('CLDR date formatting parameter not passed.');
         }
-        if (empty($locale)) {
+        if ($locale === null) {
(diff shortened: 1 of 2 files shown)


gerhard-boden:
@zaveryukha Thanks a lot for taking over! Would you mind fixing the nitpicks from #920? 
This would go into the upcoming 4.2 release.

zaveryukha:
@gerhard-boden I have read all the comments and looks like I have fixed all of them.

Neos.Eel/Classes/Helper/DateHelper.php
```diff
@@ -50,13 +63,49 @@ public function format($date, $format)
...
+     * Format a date to a string with a given cldr format
+     *
+     * @param integer|string|\DateTime $date
+     * @param string $cldrFormat Format string in CLDR format (see http://cldr.unicode.org/translation/date-time)
+     * @param null|string $locale String locale - example (de|en|ru_RU)
+     * @return string
+     */
+    public function formatCldr($date, $cldrFormat, $locale = null)
+    {
+        if ($date instanceof \DateTimeInterface) {
```
dimaip: This condition is not needed

zaveryukha: @dimaip done

Neos.Eel/Tests/Unit/Helper/DateHelperTest.php
```diff
@@ -52,7 +53,7 @@ public function formatExamples()
         return [
             'DateTime object' => [$dateTime, 'Y-m-d H:i:s', '2013-07-03 12:34:56'],
             'timestamp as integer' => [1372856513, 'Y-m-d', '2013-07-03'],
-            'timestamp as string' => ['1372856513', 'Y-m-d', '2013-07-03'],
+            'timestamp as string' => ['@1372856513', 'Y-m-d', '2013-07-03'],
```
albe: The change in behaviour here would make this change breaking. Can we make this b/c by checking a `string` argument if it `is_numeric`?

zaveryukha: @albe we discuss this with @kitsunet 
The problem what the same string '20160101' means completely different things if it date or timestamp. 
Timestamp as a string '20160101' is equivalent to: 22/08/1970 08:08:41 (format 'd/m/Y H:m:s')
Date as a string '20160101' is equivalent to: 01/01/2016 00:01:00 (format 'd/m/Y H:m:s')
IMO we must clearly separate these things.

kitsunet: I am somewhat fine with this breakiness TBH. But Up to other reviewers.

albe: Well, yeah, I'm basically fine with the change of behaviour per se, just not sure if we should add it in a minor release.

kdambekalns:
Adding some of the things from the description I tweaked to fit our needs:

---

**How to verify it**

Just use `${Date.formatCldr(someDateObject, 'dd MMMM yyyy HH:mm:ss', 'de')}` to test. Also test different combinations with missing arguments.

**Checklist**

Neos.Eel/Tests/Unit/Helper/DateHelperTest.php
kdambekalns: The `$time` argument to `DateTime::__construct()` **is a string** and to use a unix timestamp one has to use `@1234567`. `date()` on the other hand only accepts integers. 

So what we have here is a case of _our own syntactic sugar broke a defined behaviour for the sake of making things "easier"_…

IMHO whenever I have a timestamp it is a proper integer, so this change would not affect me. But I like the `is_numeric` suhggestion…

kdambekalns: Scrap that. Indeed a string like `20160101` is parsed as a date and would still be numeric, leading to very strange behaviour. Talk about magic…

So my vote: drop string handling. If people want to parse a string, why not let them use `DateTime` themselves? Using a proper object to pass dates around feels better anyway.

gerhard-boden:
@zaveryukha I agree with Karten's request to drop the support for string to date altogether, since the behaviour is not really predictable.

The CLDR in Eel part would still be a nice feature for 4.2, would you mind removing the code portions?

albe:
Haha, my comment on keeping it b/c somehow backfired :D I guess it makes sense to completely avoid the fuzzy behaviour of the DateTime constructor with strings.

I'd still prefer to introduce the breaking changes for next major only, but if we do now, let's mark this PR as breaking (!!!) and exactly document in the PR description what changed.

dimaip:
Since it's extremely rare that we would need to render dates from strings and not from DateTime objects coming from DB/CR, I don't really mind how this helper will work, as long as the two helpers are consistent with each other.
@zaveryukha what do you think? Will you be able to finish this helper in the mean time or should someone take over?

zaveryukha:
@dimaip I'll update it today. No problem.

kdambekalns:
Well, for me "drop string handling" was more about the "new string handling", we can keep the cast in place so it behaves like before. Then drop it completely for the next major.

zaveryukha:
I have removed string handling from the format helper.

gerhard-boden:
@zaveryukha Thank you!
@kdambekalns Are you ready to remove your block on this one?

gerhard-boden:
Thanks to everyone for the blazing fast reactions and special thanks to @zaveryukha and @litovchenko1 for their code contributions.  This will make it into 4.2

dimaip:
Will just stop by here to congratulate the two new contributors!
Now Fusion got a little stronger :)

kitsunet:
Indee, great work!

PR #1034, opened 2017-08-01
BUGFIX: Avoid type error with Message->getCode()

kdambekalns:
The `getCode()` method declares int as return type, but it can happen
that the internal code is `null`, leading to a type error.

This adjusts the code so the internal code is always an integer.

### DIFF
--- a/Neos.Error.Messages/Classes/Message.php
+++ b/Neos.Error.Messages/Classes/Message.php
@@ -39,7 +39,7 @@ class Message
      * The error code.
      * @var integer
      */
-    protected $code = null;
+    protected $code = 0;
 
     /**
      * The message arguments. Will be replaced in the message body.
@@ -65,7 +65,7 @@ class Message
     public function __construct(string $message, int $code = null, array $arguments = [], string $title = '')
     {
         $this->message = $message;
-        $this->code = $code;
+        $this->code = (int)$code;
         $this->arguments = $arguments;
         $this->title = $title;
     }


bwaidelich:
Mh, I'm struggling with this one.. The message's code is nullable right now. To change this just to respect the newly introduced return type hint seems weird to me.. Why not remove the type hint in this case and add a `: ?int` once the minimum PHP requirement is raised to 7.1?

kdambekalns:
FYI: 3v4l testcase at https://3v4l.org/8ZXre

@bwaidelich IMHO in this case the intention clearly is for code to be integer, the null was simply a poorly executed way of making the constructor parameter optional…

bwaidelich:
OK, but are you are sure that there is no such thing as:
```php
foreach ($result->getErrors() as $error) {
 echo $error->getMessage() . ($error->getCode() !== null ? '(' . $error->getCode() . ') : '')
}
```
?

bwaidelich:
-> I just had a quick check and it seems there is no such null check in our framework code at least. But the Neos `AbstractServiceController` renders the code as is into its json response, so there could be other places..
Pretty unlikely that something relies on this though

albe:
I'm also fine with removing the typehint until master

kdambekalns:
Whatever we do, let's decide. Right now it breaks on "all is well" flashmessages, which is, well… weird. So far I count two 👍 and two "I don't really know". The latter don't really improve the decision making. ;)

bwaidelich:
My comment wasn't so much of a "I don't really know" more of a "heads up: this might break existing code and are you sure you want to even risk breaking behavior just to satisfy a newly introduced return type hint" ;)
But it's not critical, so I didn't want to block that. We have our Release Managers to take a decision in these cases

kdambekalns:
So, let's leave it to @gerhard-boden–adjust or merge, your call… :)

kitsunet:
merge, merge, merge, merge! 😆

gerhard-boden:
Alright guys, thanks a lot for the lively discussion on this topic and especially to @bwaidelich for raising awareness when it comes to changing the default value/type of a property. In this particular case I decided to merge anyways, since I consider it a fix for a code that was not correctly implemented in the first place and The risk that this might break existing code is there, but after some discussion yesterday with Karsten and Christian I consider it negligible.

Issue #930, opened 2017-04-05
Wrong type definition in RadioViewHelper for "value" argument


p34ch:
In the`RadioViewHelper` of the FluidAdaptor package the `value` attribute is defined with type `string`. I think the correct type should be `mixed` since it's `mixed` in the`CheckboxViewHelper` and objects/entities are handled correctly.

The `value` argument is overridden here: https://github.com/neos/flow-development-collection/blob/930e627d09367e944410397b9853b15c448e35ef/Neos.FluidAdaptor/Classes/ViewHelpers/Form/RadioViewHelper.php#L64

kitsunet:
Fixed for next version.

Issue #842, opened 2017-01-26
SelectViewHelper is missing the required attribute


knusperpixel:
### Description

When using the SelectViewHelper it is not possible to add the required attribute.

### Steps to Reproduce

1. Add the view helper to your template `<f:form.select property="someProperty" options="{someOptions}" required="true" />`

#### Expected behavior

The html output should contain the `required` attribute

#### Actual behavior

Exception: `The argument "required" was not registered`

### Affected Versions

Neos: 2.2.5

Flow: 3.2.5

kdambekalns:
None of the VHs support that, except through `additionalAttributes` (unless I cannot mentally parse the code anymore). That being said, it would be great to add that for all input VHs, except `hidden`, `range`, `color` types and the `<button>` elements (see https://html.spec.whatwg.org/multipage/forms.html#concept-input-apply)

knusperpixel:
Well, at least the TextfieldViewHelper does support the `required` attribute:
https://github.com/neos/flow-development-collection/blob/2.0/TYPO3.Fluid/Classes/TYPO3/Fluid/ViewHelpers/Form/TextfieldViewHelper.php#L78

kdambekalns:
True. It was my mental parsing that failed, I just checked the registered arguments, not the parameters to `render()`…

kdambekalns:
Now, the interesting question is this: how does this behave for Flow 4.0 with the use of `typo3/typo3fluid`? @kitsunet how much abstraction do we need to adjust to get "all features" of the "upstream library"?

kitsunet:
As form is tightly coupled to MVC it's fully our responsibility. There are zero form related VHs in typo3fluid. So we just implement it in our adaptor and are done.

albe:
One back question regarding implementation: in TextfieldViewHelper it's currently a *boolean* `render()` method argument. Should we keep it like that for all VHs or change to use `registerTagAttribute`? And if the latter, should we use `boolean` or `string` type - the latter is also used for the `disabled`, so that would be mis matching.

kitsunet:
`registerTagAttribute` is the now preferred way even if that is not communicated yet. Logic for render method arguments is ours.  In general no render method in favor of static render is preferred. 

As for string/boolean, no clue, depends on how it's used I guess. The only valid value for disabled is "disabled" right? so IMHO a bool should be enough?

albe:
[...]

PR #1036, opened 2017-08-03
!!! TASK: Fix ViewHelpers and remove render attributes

cwenzelg:
Refactor most view helpers to don't use arguments inside the render function to fit
more to the Fluid creation and get rid of the current reflection workaround in the
future, see #842.

Also changes the `disabled` and `required` attribute to boolean and fixes the
FluidAdaptor accordingly.

Also fixes issue #637 so security view helpers without the security context set
will return `false` and no error.

Also fixes issue #930 for the wrong type definition.


kitsunet:
Note this was discussed with me during the sprint. I like the change, removing render method arguments should be done as fluid standalone doesn't support them anymore and we add some stuff on top to make it possible, but generally it's a bad idea and I would want to deprecate the concept soonish.

kdambekalns:
Hear, hear… but the unit tests fail.

albe:
Sounds reasonable. Apart from the failing tests:
1) are the render methods still to be considered `@api`?
2) are the arguments still documented correctly from the `registerArgument()` calls alone?

kitsunet:
1. No, `@api` should probably be removed pretty much immediately.
2. I don't know, but as we had those before as well I do hope so. Otherwise I guess we can get inspiration from fluid itself, has somehow a documentation is generated there too.

gerhard-boden:
@cwenzelg Thanks a lot for you contribution. If we manage to fix the failing tests, this might make it into the upcoming release.  So this seems there is a TASK / BUGFIX mixture going on here. Are all bugfixes directly linked to the refactoring (it's hard to tell, because it's a lot of changes)?. Maybe split up multiple 
bugfixes into separate PRs next time, so it's easier to review. But apart that: Good job!

kitsunet:
I would suggest to merge it right after the release into master IMHO. I said it's ok to mix the bugfixes in as it is so difficult to separate the stuff out...

cwenzelg:
@kitsunet I fixes as much tests as i could. I'm lost for the last open tests. There are some that expect a Exception that currently will not come anymore and some other type issues i think. So if i still can help just write me on slack, otherwise you might need to fix the rest. ;)

albe:
Would be cool to finish this up at some point, as there is already so much work in this. Anyone up to look into the remaining failing tests?

kitsunet:
will tackle tests

Neos.FluidAdaptor/Tests/Unit/ViewHelpers/Format/StripTagsViewHelperTest.php
```diff
@@ -129,15 +129,15 @@ public function renderConvertsObjectsToStrings()
 
     /**
      * @test
+     * @expectedException \InvalidArgumentException
```
albe: See above.

Neos.FluidAdaptor/Tests/Unit/ViewHelpers/Format/HtmlentitiesViewHelperTest.php
```diff
@@ -162,12 +162,12 @@ public function renderConvertsObjectsToStrings()
 
     /**
      * @test
+     * @expectedException \InvalidArgumentException
```
albe: See above.

Neos.FluidAdaptor/Tests/Unit/ViewHelpers/Format/HtmlentitiesDecodeViewHelperTest.php
```diff
@@ -138,12 +138,12 @@ public function renderDecodesObjectsToStrings()
 
     /**
      * @test
+     * @expectedException \InvalidArgumentException
```
albe: Now that's a (breaking?) change in behaviour, isn't it? At any rate the test method name no longer matches the tested behaviour now.

kitsunet: Basically because now the validation is moved to standalone Fluid.

albe: Mhh... ok, so this will now error out where it previously just silently passed the argument back. Will this have any effect on real world use? i.e. would this have thrown an error later anyway?

mficzel:
Thinking about this i think this is breaky for all cases where the changed viewHelpers were extended.

Neos.FluidAdaptor/Classes/ViewHelpers/DebugViewHelper.php
```diff
@@ -55,22 +55,31 @@ class DebugViewHelper extends AbstractViewHelper
...
 
+    /**
+     * Arguments initialization
+     *
+     * @return void
+     */
+    public function initializeArguments()
+    {
+        $this->registerArgument('title', 'string', 'The title', false, null);
+        $this->registerArgument('ypeOnly', 'boolean', 'Whether only the type should be returned instead of the whole chain.', false, false);
```
kitsunet: That's wrong and needs to be fixed still. 'typeOnly'

mficzel:
This is to breaky for a minor release because of situations like this: `\Neos\Neos\ViewHelpers\Backend\TranslateViewHelper` extends  `\Neos\FluidAdaptor\ViewHelpers\TranslateViewHelper` and internally uses ` parent::render($id, $value, $arguments, $source, $package, $quantity, 'en');`  … this fails because `parent::render()` has no attributes on render any more.

kitsunet:
I would fix the conflicts on this ASAP and then merge it to master so there is ample time for people trying master to see what they need to adapt.

kitsunet:
Merged master, fixed conflicts. Will merge when Travis is happpy.

Neos.FluidAdaptor/Classes/ViewHelpers/DebugViewHelper.php
bwaidelich: That PR was merged in spite of this typo

bwaidelich: -> #1183

PR #1037, opened 2017-08-04
BUGFIX: Catch ConnectionExceptions in `PersistenceManager::persistAll()`

bwaidelich:
This adds a `try/catch` block around the `$connection->ping()` call that
throws an exception if no database credentials are configured (which is
the case when you do a fresh install of Flow or Neos).

Fixes: #1663

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
@@ -13,6 +13,7 @@ namespace TYPO3\Flow\Persistence\Doctrine;
 
 use Doctrine\Common\Persistence\ObjectManager;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception\ConnectionException;
 use Doctrine\ORM\Tools\SchemaTool;
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Log\SystemLoggerInterface;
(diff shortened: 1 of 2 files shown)


kdambekalns:
Well… do we consider it exceptional that no DB connection is configured? If not, we should check for a configured connection instead… _Sigh_

Ideally, if  no DB connection exists, disable the whole persistence subsystem 😎

albe:
What Karsten says. Also if no database is configured, the following `entityManager->flush()` would fail again anyway, wouldn't it?

TYPO3.Flow/Classes/TYPO3/Flow/Persistence/Doctrine/PersistenceManager.php
```diff
@@ -162,10 +163,14 @@ public function persistAll($onlyWhitelistedObjects = false)
 
         /** @var Connection $connection */
         $connection = $this->entityManager->getConnection();
-        if ($connection->ping() === false) {
-            $this->systemLogger->log('Reconnecting the Doctrine EntityManager to the persistence backend.', LOG_INFO);
-            $connection->close();
-            $connection->connect();
+        try {
```
kitsunet: for now that seems fine to me, but yeah, I would check for a configured DB early on and maybe set a magic config value as for context?

albe:
+1 for database-less a first-class citizen

bwaidelich:
Absolutely +1 for a more sophisticated solution, but I couldn't come up with one quickly enough. Let's do this!

PR #1040, opened 2017-08-07
Merge to 40

kitsunet:


kitsunet:
Just for testing this ;) Will close this and directly merge when jenkins is happy and then continue upmerging

PR #1041, opened 2017-08-07
TASK: Add PHP 7.0 scalar type hints to method arguments and return types [AOP/Cache]

albe:


albe:
Looking good now, yay. Please test very carefully

PR #1042, opened 2017-08-07
TASK: Add PHP 7.0 scalar type hints to method arguments and return types [Cli/Command]

albe:


kitsunet:
Merge in 24h if no objections.

albe:
Thx for the fix and cleanup

PR #1043, opened 2017-08-07
TASK: Add PHP 7.0 scalar type hints to method arguments and return types [Composer/Configuration/Core/Error]

albe:


albe:
Don't understand why the Bootstrap fails with an autowiring exception. Can't be from the typehint itself, can it?

Neos.Flow/Classes/Error/AbstractExceptionHandler.php
```diff
@@ -117,7 +117,7 @@ public function handleException($exception)
     /**
      * Echoes an exception for the web.
      *
-     * @param object $exception \Exception or \Throwable
+     * @param \Throwable $exception
      * @return void
      */
     abstract protected function echoExceptionWeb($exception);
```
bwaidelich: I guess you didn't add the type hint here for B/C!?

Neos.Flow/Classes/Error/AbstractExceptionHandler.php
```diff
@@ -240,15 +240,16 @@ protected function resolveRenderingGroup($exception)
...
 
     /**
      * Formats and echoes the exception and its previous exceptions (if any) for the command line
      *
-     * @param object $exception \Exception or \Throwable
+     * @param \Throwable $exception
      * @return void
      */
-    protected function echoExceptionCli($exception)
+    protected function echoExceptionCli(\Throwable $exception)
```
bwaidelich: ..but you _did_ add it here?!

Neos.Flow/Classes/Error/DebugExceptionHandler.php
```diff
@@ -54,7 +54,7 @@ class DebugExceptionHandler extends AbstractExceptionHandler
     /**
      * Formats and echoes the exception as XHTML.
      *
-     * @param object $exception \Exception or \Throwable
+     * @param \Throwable $exception
      * @return void
      */
     protected function echoExceptionWeb($exception)
```
bwaidelich: no type hint for B/C?

Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
```diff
@@ -213,22 +213,24 @@ protected function registerClassFiles(array $packages)
...
-                            $availableClassNames[$packageKey][] = $fullClassName;
-                        }
-                    }
-                }
-                if (isset($availableClassNames[$packageKey]) && is_array($availableClassNames[$packageKey])) {
-                    $availableClassNames[$packageKey] = array_unique($availableClassNames[$packageKey]);
-                }
+            }
+            if (isset($availableClassNames[$packageKey]) && is_array($availableClassNames[$packageKey])) {
+                $availableClassNames[$packageKey] = array_unique($availableClassNames[$packageKey]);
```
bwaidelich: This seems unrelated?!

Neos.Flow/Classes/Session/Session.php
```diff
@@ -287,7 +287,7 @@ public function start()
         if ($this->request === null) {
             $requestHandler = $this->bootstrap->getActiveRequestHandler();
             if (!$requestHandler instanceof HttpRequestHandlerInterface) {
-                throw new Exception\InvalidRequestHandlerException('Could not start a session because the currently active request handler (%s) is not an HTTP Request Handler.', 1364367520);
+                throw new Exception\InvalidRequestHandlerException(sprintf('Could not start a session because the currently active request handler (%s) is not an HTTP Request Handler.', gettype($requestHandler)), 1364367520);
```
bwaidelich: Is that related?

Neos.Flow/Tests/Unit/Session/SessionTest.php
```diff
@@ -300,8 +301,9 @@ public function startPutsACookieIntoTheHttpResponse()
      */
     public function startThrowsAnExceptionIfIncompatibleRequestHandlerIsUsed()
     {
+        $mockRequestHandler = $this->createMock(RequestHandlerInterface::class);
         $mockBootstrap = $this->createMock(Bootstrap::class);
-        $mockBootstrap->expects($this->any())->method('getActiveRequestHandler')->will($this->returnValue(new \stdClass()));
+        $mockBootstrap->expects($this->any())->method('getActiveRequestHandler')->willReturn($mockRequestHandler);
```
bwaidelich: Is this related?

Neos.Flow/Classes/Error/AbstractExceptionHandler.php
albe: I think, there was some reason for not adding it here. Could be b/c, because all classes implementing the abstract method would need to change.

Neos.Flow/Classes/Error/DebugExceptionHandler.php
albe: I just kept the exact abstract method signature without much thinking here I guess

Neos.Flow/Classes/Error/AbstractExceptionHandler.php
albe: I think that was just because it is implemented here, not api and hence my thinking process was probably: "No one overrides it anyway". But you're right that's inconsequent. It probably needs adjustment by people that implement an ExceptionHandler anyway, so we can just go full out with typehints and leave a nice message about this, or maybe even provide a migration?

Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
albe: I don't think I did this tbh

Neos.Flow/Classes/Session/Session.php
albe: Nope, I just fixed the missing sprintf when I stumbled over this "(%s)".

Neos.Flow/Tests/Unit/Session/SessionTest.php
albe: It needed to return an instanceof RequestHandlerInterface due to the typehints and then I also just switched to the more readable `->willReturn` in favor over `->will($this->returnValue(`

albe:
Woah, thanks a lot Christian! Impressive effort you put into that to fix this up, highly appreciated! Let's merge this if no other objections coming

PR #1046, opened 2017-08-09
TASK: Error Message accepts null as message title

albe:
Also, this allow the code of the message to be null in `getCode()`, since
that is the more sane behaviour from an API point of view.

Fixes neos/neos-development-collection#1666

### DIFF
--- a/Neos.Error.Messages/Classes/Message.php
+++ b/Neos.Error.Messages/Classes/Message.php
@@ -57,15 +57,15 @@ class Message
      * Constructs this error
      *
      * @param string $message An english error message which is used if no other error message can be resolved
-     * @param integer $code A unique error code
+     * @param integer|null $code A unique error code
      * @param array $arguments Array of arguments to be replaced in message
-     * @param string $title optional title for the message
+     * @param string|null $title optional title for the message
      * @api
      */
-    public function __construct(string $message, int $code = null, array $arguments = [], string $title = '')
+    public function __construct(string $message, int $code = null, array $arguments = [], string $title = null)
     {
         $this->message = $message;
-        $this->code = (int)$code;
+        $this->code = $code;
         $this->arguments = $arguments;
         $this->title = $title;
     }
@@ -84,10 +84,10 @@ class Message
     /**
      * Returns the error code
      *
-     * @return integer The error code
+     * @return integer|null The error code
      * @api
      */
-    public function getCode(): int
+    public function getCode()
     {
         return $this->code;
     }
@@ -102,10 +102,10 @@ class Message
     }
 
     /**
-     * @return string
+     * @return string|null
      * @api
      */
-    public function getTitle(): string
+    public function getTitle()
     {
         return $this->title;
     }
--- a/Neos.Error.Messages/Classes/Message.php
+++ b/Neos.Error.Messages/Classes/Message.php
@@ -62,7 +62,7 @@ class Message
      * @param string|null $title optional title for the message
      * @api
      */
-    public function __construct(string $message, int $code = null, array $arguments = [], string $title = null)
+    public function __construct(string $message, int $code = null, array $arguments = [], $title = '')
     {
         $this->message = $message;
         $this->code = $code;
@@ -79,6 +79,15 @@ class Message
     public function getMessage(): string
     {
         return $this->message;
+    }
+
+    /**
+     * @return bool
+     * @api
+     */
+    public function hasCode(): bool
+    {
+        return $this->code !== null;
     }
 
     /**
@@ -99,6 +108,15 @@ class Message
     public function getArguments(): array
     {
         return $this->arguments;
+    }
+
+    /**
+     * @return bool
+     * @api
+     */
+    public function hasTitle(): bool
+    {
+        return $this->title !== null && $this->title !== '';
     }
 
     /**


albe:
The reasoning for changing code and title to `nullable` values is, that those are really supposed to be nullable and the previous default values of `0` and `''` were just magic constants actually.

Also, probably a lot of code intentionally or unintentionally creates Error\Messages with null values, for example due to using a translator that might return null (as is the case in Neos AssetController).

kitsunet:
I like but where do the test failures come from?

albe:
That is a very good question... investigating

albe:
Arghh.... `$singleFlashMessage->getTitle() !== ''` - it's basically what I wrote in slack yesterday for the `code` ... guess we need to go with casting the null arguments in constructor for now...

bwaidelich:
> Ok, let's wait for PHP 7.1 and nullable return types. How could that not be implemented from the start, btw?

Oh yes, that half-baked implementation has caused major pain..

As discussed via Slack, I'd suggest to go for the fully b/c solution here:

```php
__construct(string $message, int $code = null, array $arguments = [], string $title = null)
// ...
public function getCode()
public function getTitle(): string
```

And possibly add `hasCode()` and `hasTitle()` methods to inline this logic.
And add the `getCode(): ?int` return type with PHP 7.1+

Neos.Error.Messages/Classes/Message.php
```diff
@@ -57,15 +57,15 @@ class Message
...
-     * @param string $title optional title for the message
+     * @param string|null $title optional title for the message
      * @api
      */
-    public function __construct(string $message, int $code = null, array $arguments = [], string $title = '')
+    public function __construct(string $message, int $code = null, array $arguments = [], $title = '')
     {
         $this->message = $message;
-        $this->code = (int)$code;
+        $this->code = $code;
```
albe: This was actually a non b/c change, since `code` really defaulted to `null` before.

Neos.Error.Messages/Classes/Message.php
```diff
@@ -57,15 +57,15 @@ class Message
...
      * @param string $message An english error message which is used if no other error message can be resolved
-     * @param integer $code A unique error code
+     * @param integer|null $code A unique error code
      * @param array $arguments Array of arguments to be replaced in message
-     * @param string $title optional title for the message
+     * @param string|null $title optional title for the message
      * @api
      */
-    public function __construct(string $message, int $code = null, array $arguments = [], string $title = '')
+    public function __construct(string $message, int $code = null, array $arguments = [], $title = '')
```
albe: Adding typehint for string not possible without adding subtle changes in behaviour. Defaulting to `null` would change the value when nothing is passed and casting to string would change the value when `null` is explicitly passed (for whatever reason).

Neos.Error.Messages/Classes/Message.php
```diff
@@ -57,15 +57,15 @@ class Message
      * Constructs this error
      *
      * @param string $message An english error message which is used if no other error message can be resolved
-     * @param integer $code A unique error code
+     * @param integer|null $code A unique error code
      * @param array $arguments Array of arguments to be replaced in message
-     * @param string $title optional title for the message
+     * @param string|null $title optional title for the message
```
bwaidelich: &lt;nitpick&gt;`@param string $title`&lt;/nitpick&gt;

Neos.Error.Messages/Classes/Message.php
```diff
@@ -102,10 +111,19 @@ public function getArguments(): array
...
+     * @return bool
+     * @api
+     */
+    public function hasTitle(): bool
+    {
+        return $this->title !== null && $this->title !== '';
+    }
+
+    /**
+     * @return string|null
```
bwaidelich: &lt;nitpick&gt;`@return string`&lt;/nitpick&gt;

Neos.Error.Messages/Classes/Message.php
```diff
@@ -102,10 +111,19 @@ public function getArguments(): array
...
      * @api
      */
-    public function getTitle(): string
+    public function getTitle()
```
bwaidelich: This typehint can stay

Neos.Error.Messages/Classes/Message.php
albe: Just want to make it very explicit that `null` is allowed (for now due to b/c) here

Neos.Error.Messages/Classes/Message.php
albe: See above :) We return what we get

Neos.Error.Messages/Classes/Message.php
albe: Nope, since it will fail if a null comes in (see neos/neos-development-collection#1666)

bwaidelich: IMO we have to fix the _calling_ side in this case. Explicitly passing `null` is not in sync with the signature and has been a bug all along (even if one could argue that PHP behaves weird in that case)..
But, anyways, it doesn't really make a difference. Let's just get this fixed and work on greater things :D

bwaidelich:
@albe Thanks for this!
With the new `hasCode()` and `hasTitle()` methods in place we should adjust all the code that still does this check "manually", but we can do that in a separate patch ofc

PR #1048, opened 2017-08-10
TASK: Make FlashMessageViewHelper use `hasTitle()` to check for existence

albe:
Depends on #1046


albe:
This will fail tests until rebased on a merged #1046

PR #1050, opened 2017-08-10
BUGFIX: Avoid a syntax error from hiding the intended error message

albe:
The error code for the Package Manager error on duplicate packages was
a millisecond timestamp. Since this is larger than MAX_INT PHP parses
this as float, which in turn is not accepted as parameter for an
exception code.


bwaidelich:
Wooha, nice catch

PR #1052, opened 2017-08-12
TASK: Improve readability of Role identifier pattern

albe:
This also removes the unused check for simple role name notation.

### DIFF
--- a/Neos.Flow/Classes/Security/Policy/Role.php
+++ b/Neos.Flow/Classes/Security/Policy/Role.php
@@ -19,6 +19,8 @@ use Neos\Flow\Security\Authorization\Privilege\PrivilegeInterface;
  */
 class Role
 {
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Security/Policy/Role.php
```diff
@@ -68,9 +70,7 @@ public function __construct($identifier, array $parentRoles = [])
         if (!is_string($identifier)) {
             throw new \InvalidArgumentException('The role identifier must be a string, "' . gettype($identifier) . '" given. Please check the code or policy configuration creating or defining this role.', 1296509556);
         }
-        if (preg_match('/^[\w]+((\.[\w]+)*\:[\w]+)?$/', $identifier) !== 1) {
-        }
```
bwaidelich: Nice one, introduced with 1759cdf ;)

PR #1053, opened 2017-08-12
BUGFIX: Make CLI subprocesses work on Windows

albe:
On Windows, the PHP_BINDIR constant will resolve to "C:\php" no matter what the actual php directory is, because this constant is decided at compile time. However, the PHP_BINARY constant will resolve to the current executable that runs the script at runtime, which at least for CLI will be the PHP executable.

This is a backport of #1010

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/CoreCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/CoreCommandController.php
@@ -364,7 +364,7 @@ class CoreCommandController extends CommandController
      */
     protected function launchSubProcess()
     {
-        $systemCommand = 'FLOW_ROOTPATH=' . FLOW_PATH_ROOT . ' FLOW_PATH_TEMPORARY_BASE=' . FLOW_PATH_TEMPORARY_BASE . ' ' . 'FLOW_CONTEXT=' . $this->bootstrap->getContext() . ' ' . PHP_BINDIR . '/php -c ' . php_ini_loaded_file() . ' ' . FLOW_PATH_FLOW . 'Scripts/flow.php' . ' --start-slave';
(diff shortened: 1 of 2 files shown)


albe:
@kaystrobach

kaystrobach:
works like a charm now -- thank you!

bwaidelich:
Thanks!

kaystrobach:
can you please tag a new release? - would be awesome

kaystrobach:
hey guys,

walking through the code i found another interesting place:

Packages/Framework/TYPO3.Flow/Configuration/Settings.yaml:48

maybe the default should be changed as well ...

```
      # Path and filename of the PHP binary
      # The constant PHP_BINDIR usually contains the path, but on Windows this doesn't work reliably
      phpBinaryPathAndFilename: '%PHP_BINDIR%/php'
```

albe:
Yeah, I found that as well, but:

- that one setting is used for Web requests, when spawning the commands
- in Web context, PHP_BINARY isn't safe to use any more (e.g. on apache you end up with the path to the apache process binary)

I started to write a more sophisticated PHP binary detection, but that would need some more work to be used at the right places and would also involve spawning a couple `exec()` calls to "test" the probable executables. Not sure that isn't too much work on a per request basis, so probably needs some compile-time/static caching of the result

kaystrobach:
mhmm strange, may also be interesting when using php-fpm ... it's really bad, that php does not provide a save way to get the cli binary.

Maybe it makes sense to get in touch with the php devs?

kaystrobach:
otherwise the subprocess are used for compilation / reflection.

maybe it makes sense to use https://github.com/nikic/PHP-Parser instead of real subprocesses.

albe:
There is an idea to use Roave Better-Reflection for this, which would make the subprocess unnecessary. But that will be a hell lot of work. Anyway, let's better discuss that somewhere else.

kaystrobach:
Looks even better and seems to nicely integrate with composer, what is awesome.
Sure - should I open a meta ticket for that, or is there one already?

kitsunet:
The Roave thing is also waaaay slower, we had a go in some project with it...

PR #1056, opened 2017-08-15
BUGFIX: Support proxies forwarding the port in the Forwarded-Host header

### LINKED ISSUE #1054: Apache as Proxy provides `X-forwarded-Host` with host + port
The `TrustedProxiesComponent` only expects  a hostname in `X-forwarded-Host` while Apache with mod_proxy for example provides Host:Port in that header effectively causing an exception in Flow/Neos. The Component should check that and act accordingly if it finds the Host:Port pattern.

albe:
Some proxies like apache's mod_proxy set the port from the forwarded
connection in the `X-Forwarded-Host` header, e.g. `acme.com:443`.
This change adjusts the TrustedProxy component to allow setting the
forwarded port inside the Host header.
If the `X-Forwarded-Port` header is also present, that one will take
precedence though.

Fixes #1054

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
@@ -50,14 +50,21 @@ class TrustedProxiesComponent implements ComponentInterface
         }
 
         $hostHeader = $this->getFirstTrustedProxyHeaderValue(self::HEADER_HOST, $trustedRequest);
+        $portFromHost = null;
         if ($hostHeader !== null) {
+            $portSeparator = strrpos($hostHeader, ':');
+            if ($portSeparator > 0) {
+                $portFromHost = substr($hostHeader, $portSeparator + 1);
+                $trustedRequest->getUri()->setPort($portFromHost);
+                $hostHeader = substr($hostHeader, 0, $portSeparator);
+            }
             $trustedRequest->getUri()->setHost($hostHeader);
         }
 
         $portHeader = $this->getFirstTrustedProxyHeaderValue(self::HEADER_PORT, $trustedRequest);
         if ($portHeader !== null) {
             $trustedRequest->getUri()->setPort($portHeader);
-        } elseif ($protocolHeader !== null) {
+        } elseif ($protocolHeader !== null && $portFromHost === null) {
             $trustedRequest->getUri()->setPort(strtolower($protocolHeader) === 'https' ? 443 : 80);
         }
 
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
@@ -53,7 +53,7 @@ class TrustedProxiesComponent implements ComponentInterface
         $portFromHost = null;
         if ($hostHeader !== null) {
             $portSeparatorIndex = strrpos($hostHeader, ':');
-            if ($portSeparatorIndex > 0) {
+            if ($portSeparatorIndex !== false) {
                 $portFromHost = substr($hostHeader, $portSeparatorIndex + 1);
                 $trustedRequest->getUri()->setPort($portFromHost);
                 $hostHeader = substr($hostHeader, 0, $portSeparatorIndex);
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
@@ -52,11 +52,11 @@ class TrustedProxiesComponent implements ComponentInterface
         $hostHeader = $this->getFirstTrustedProxyHeaderValue(self::HEADER_HOST, $trustedRequest);
         $portFromHost = null;
         if ($hostHeader !== null) {
-            $portSeparator = strrpos($hostHeader, ':');
-            if ($portSeparator > 0) {
-                $portFromHost = substr($hostHeader, $portSeparator + 1);
+            $portSeparatorIndex = strrpos($hostHeader, ':');
+            if ($portSeparatorIndex > 0) {
+                $portFromHost = substr($hostHeader, $portSeparatorIndex + 1);
                 $trustedRequest->getUri()->setPort($portFromHost);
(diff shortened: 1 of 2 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
```diff
@@ -50,14 +50,21 @@ public function handle(ComponentContext $componentContext)
         }
 
         $hostHeader = $this->getFirstTrustedProxyHeaderValue(self::HEADER_HOST, $trustedRequest);
+        $portFromHost = null;
         if ($hostHeader !== null) {
+            $portSeparator = strrpos($hostHeader, ':');
+            if ($portSeparator > 0) {
```
bwaidelich: strrpos returns `FALSE` if the needle doesn't exist in the haystack string. Also &lt;nitpick&gt;that variable won't contain the port separator but it's position&lt;/nitpick&gt;
I suggest the following code:
```php
list($hostHeaderWithoutPort, $portFromHost) = explode(':', $hostHeader);
$portSeparator = strrpos($hostHeader, ':');
if ($portSeparator !== false) {
    $trustedRequest->getUri()->setPort($portFromHost);
}
$trustedRequest->getUri()->setHost($hostHeaderWithoutPort);
```
That would also allow a host of ":<port>" (even though I'm not sure if that's valid tbh).
If you agree I can push that as follow-up with a corresponding test

bwaidelich: Sorry, wrong snippet. Just:
```php
list($hostHeaderWithoutPort, $portFromHost) = explode(':', $hostHeader);
if ($portFromHost !== null) {
    $trustedRequest->getUri()->setPort($portFromHost);
}
$trustedRequest->getUri()->setHost($hostHeaderWithoutPort);
```

TYPO3.Flow/Tests/Unit/Http/Component/TrustedProxiesComponentTest.php
```diff
@@ -339,6 +339,46 @@ public function hostIsNotOverridenIfTheHeaderIsNotTrusted()
...
+    {
+        $request = Request::create(new Uri('http://acme.com'), 'GET', array(), array(), array('HTTP_X_FORWARDED_HOST' => 'neos.io:443', 'HTTP_X_FORWARDED_PROTO' => 'http'));
+        $trustedRequest = $this->callWithRequest($request);
+        $this->assertEquals(443, $trustedRequest->getUri()->getPort());
+    }
+
+    /**
+     * @test
+     */
+    public function portFromHostHeaderIsOverriddenByPortHeader()
```
bwaidelich: Some more substantial feedback: Is that really the intended behavior that the port header overrides a port from the host header?
(not a rhetoric question, I'm sure that makes sense but I wonder if that's defined somewhere)

TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
albe: Hm, yeah. I had that exact solution with explode first, but thought it would lead to wrong results if there are two `:` in the host name. Not sure if that would ever be the case, as that would at most be valid when specifying `user:password@host` and that's nothing that should ever occur in a host header, no? Anyway, I'm open to change this in good faith :)

albe: Ok, actually the host header might include an IP address instead of the domain name and now think what happens when that is an IPv6 address... lots of `:` involved...

TYPO3.Flow/Tests/Unit/Http/Component/TrustedProxiesComponentTest.php
albe: There is no defined standard on the `X-Forwarded-*` headers, as they just became a defacto-standard by usage. So there is no rule on how this should work, but it just makes total sense if any weird proxy ever decided to specify the port in both headers.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
bwaidelich: I doubt that we support IPv6 at all, as we'd have to strip of the square brackets around these to get the port (see https://stackoverflow.com/questions/186829/how-do-ports-work-with-ipv6).

Anyways, good call, so this should work even for `:<just-the-port>`:
```php
$portSeparatorIndex = strrpos($hostHeader, ':');
if ($portSeparatorIndex !== false) {
    $portFromHost = substr($hostHeader, $portSeparatorIndex + 1);
    $trustedRequest->getUri()->setPort($portFromHost);
    $hostHeader = substr($hostHeader, 0, $portSeparatorIndex);
}
$trustedRequest->getUri()->setHost($hostHeader);
```
IMO https://github.com/neos/flow-development-collection/pull/1056#pullrequestreview-56697249 is more important though

TYPO3.Flow/Tests/Unit/Http/Component/TrustedProxiesComponentTest.php
bwaidelich: I could even imagine a server error if `X-Forwarded-For` and `X-Forwarded-Port` are not in sync. But, well, at least the current behavior is covered with a test and I agree that it does make sense!

albe: Here is one theoretic scenario in which the two might be different:

Suppose you have two proxies, one that specifies the port via `X-Forwarded-Port`, the other via the `X-Forwarded-Host`. You'd end up with either `X-Forwarded-Host: acme.com, foo-bar.org:8080; X-Forwarded-Port: 443` or `X-Forwarded-Host: acme.com:443, foo-bar.org; X-Forwarded-Port: 8080`.

Now of course that also doesn't make clear which is "correct" generally (I'm not even sure that would be handled sanely - curse you `mod_proxy`!).

bwaidelich:
Sorry, I can't look into this again before Monday so I just removed my review for now. I fully trust you guys ;)

albe:
ping @bwaidelich can you check again or remove your block, since you said you trust us :)

bwaidelich:
@albe I already removed my -1!?
I can check again tomorrow

albe:
Thanks!

Issue #849, opened 2017-02-02
Response::createFromRaw() chokes on Protocol Upgrade


kdambekalns:
When a raw response like this is encountered:

````
HTTP/1.0 200 Connection established

HTTP/1.1 200 OK
Server: openresty
Content-Type: application/json; charset=utf-8
Content-Length: 3202
Vary: Accept-Encoding
Date: Thu, 02 Feb 2017 15:17:42 GMT
Connection: Keep-alive

{"id":"ecfb276920529fc9b9…
````

the resulting response contains the raw "2nd response", which e.g. breaks decoding it as JSON.

This is due to the fact that after an empty line header parsing stops and the remaining lines are considered the response content.

**Comment by kdambekalns** [comment]
@robertlemke Interesting, no?

**Comment by albe** [comment]
Interesting indeed that this receives a "HTTP/1.0 200" and then another response with "HTTP/1.1 200"... that is *not* a proper protocol upgrade as to my understanding (the first response code should be anything other than 200). Also, this is not a proper response as to HTTP spec, which clearly states that everything after the first empty line is the body. The Response implementation is following spec, so the issue is CURL specific (see also http://php.net/manual/de/function.curl-setopt.php#103232). I guess we need to fix this in `CurlEngine`.

```php
while (substr($response->getContent(), 0, 5) === 'HTTP/') {
   $response = Http\Response::createFromRaw($response->getContent(), $response);
}
```
should do the trick

albe:
Closing this as of the merged PR.

PR #1057, opened 2017-08-15
BUGFIX: CurlEngine follows chain of Http headers

### LINKED ISSUE #1044: Using a proxy with cURL leads to misbehaving in splitting header and body
### Description

When using a system-wide proxy, TYPO3\Flow\Http fails to correctly split HTTP headers and content. This happens because there is a loop parsing every single response line and stops at the first empty line it finds which usually splits header from content:

Response.php, at line 144
`foreach ($lines as $line) {
if ($parsingHeader) {
if (trim($line) === '') {
$parsingHeader = false;
continue;
} ... } ... } ...
`

In case a system wide proxy is set (ex.: env var http_proxy/HTTPS_PROXY in web server or php-fpm config), there are two header blocks in a cURL response while the first block represents the direct answer from proxy and the second one the forwarded block the proxy server received from request target.

### Steps to Reproduce

1. Configure http(s) proxy to use by setting http_proxy and/or HTTPS_PROXY env var in web server config or php-fpm config, depending on what is used
2. Create a HTTP network request by using TYPO3\Flow\Http framework package
3. Call `$response->getContent()` function which then contains the forwarded headers in the first lines

#### Expected behavior

`$response->getContent()` function should only return the request body. The first, usually unnecessary header lines (directly from proxy connection) should be truncated and not parsed as header lines instead of actually needed response headers from target server.

#### Actual behavior

`$response->getContent()` function returns the forwarded HTTP header lines the proxy received from request target before the actual response content starts.

### Affected Versions

Neos: 2.3.14

Flow: 3.3.15

albe:
Under some circumstances, curl will return all the headers of the request
chain, for example when the CURLOPT_FOLLOWLOCATION option is set.
This change will make the CurlEngine follow the chain of response headers
until the final response.

Fixes #1044,  #849, #570

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
@@ -148,9 +148,10 @@ class CurlEngine implements RequestEngineInterface
         curl_close($curlHandle);
 
         $response = Http\Response::createFromRaw($curlResult);
-        if ($response->getStatusCode() === 100) {
+        while (substr($response->getContent(), 0, 5) === 'HTTP/' || $response->getStatusCode() === 100) {
             $response = Http\Response::createFromRaw($response->getContent(), $response);
         }
+
         return $response;
     }
 }
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
@@ -148,8 +148,11 @@ class CurlEngine implements RequestEngineInterface
         curl_close($curlHandle);
 
         $response = Http\Response::createFromRaw($curlResult);
-        while (substr($response->getContent(), 0, 5) === 'HTTP/' || $response->getStatusCode() === 100) {
-            $response = Http\Response::createFromRaw($response->getContent(), $response);
+        try {
+            while (substr($response->getContent(), 0, 5) === 'HTTP/' || $response->getStatusCode() === 100) {
+                $response = Http\Response::createFromRaw($response->getContent(), $response);
+            }
+        } catch (\InvalidArgumentException $e) {
         }
 
         return $response;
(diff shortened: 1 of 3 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
```diff
@@ -148,9 +148,10 @@ public function sendRequest(Http\Request $request)
         curl_close($curlHandle);
 
         $response = Http\Response::createFromRaw($curlResult);
-        if ($response->getStatusCode() === 100) {
+        while (substr($response->getContent(), 0, 5) === 'HTTP/' || $response->getStatusCode() === 100) {
```
albe: Actually not sure if the `100` code check is still needed with this, but I kept it in place just to be sure.

kdambekalns:
@thelalaman

TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
bwaidelich: Is that safe to check the response _content_ to start with "HTTP/"?
Seems a bit "magic" to me. Wouldn't the `$curlResult` contain more details about this so we can catch that case more "safely"?

albe: I asked that myself too, but there is really no other information in `$curlResult`. It just appends the response headers after each other. We could only apply some more checks on the format of the content to make more sure it is indeed a HTTP Response, like checking for exact line delimiters (CRLF), exact status code match (3 digits) and possible headers in `identifier: value(;value)*?` format and followed by an empty line.

bwaidelich:
@kdambekalns did your comment have any deeper meening or was that a happy accident? ;)

kdambekalns:
It was meant to draw attention to this PR, since I know @thelalaman had that issue and his colleague opened the latest of the issues this should fix.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Client/CurlEngine.php
albe: probably `Response::createFromRaw` should just be more strict on those checks itself and then we catch any error and break the loop.

albe:
This should prevent most cases that somehow respond with seemingly HTTP response header content by making the parsing more strict and CurlEngine catching and stopping on eceptions.

albe:
Yeah, that's a good point for things to go over generally in the tests.

PR #1059, opened 2017-08-16
TASK: Make "type" argument optional in "configuration:show" CLI command

bwaidelich:
During development it's a common requirement to output
the configuration for a given path.
Previously that always required the configuration *type*
to be specified like::

    ./flow configuration:show --type Settings

With this change the type is optional and defaults to "Settings"
which is commonly the most interesting type for this command.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Command/ConfigurationCommandController.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Command/ConfigurationCommandController.php
@@ -53,13 +53,20 @@ class ConfigurationCommandController extends CommandController
      * The command shows the configuration of the current context as it is used by Flow itself.
      * You can specify the configuration type and path if you want to show parts of the configuration.
      *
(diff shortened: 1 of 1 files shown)


aertmann:
have always annoyed me lol 😀

kitsunet:
This totally is a good change, but for the good fo all upmerging, please why not master as it is a feature...

albe:
The PR headline says "TASK:" :P You're right, but I thought in this case it makes sense to have this convenience available in the LTS, since the change is not breaking and straightforward

bwaidelich:
@kitsunet sorry if this caused trouble. I deliberately applied this to our LTS versions because we still have to work with them for a long time and this thing bugged me (and assumably some of you) enough and I wanted to prevent anyone having to think about the branch they're working in when dealing with multiple installations on the CLI.

But much more importantly: If upmerging this one-liner that should apply cleanly to all affected branches (except for the static namespace change) takes more than five minutes, we should seriously improve that process and I'm happy to help with that if I can!

PR #1061, opened 2017-08-17
!!!FEATURE: Configurable FlashMessage containers

bwaidelich:
Introduces a new setting `Neos.Flow.mvc.flashMessages.containers`
that allows to configure separate FlashMessage containers.

It also changes the default behavior to persist FlashMessages in a
session Cookie rather than in a server-side session.

This is a breaking change because it changes the API slightly by
removing the `@api` annotation from the `FlashMessageContainer`.
This is required because that object is no longer stored in the
session by default.
Instead of interacting with the `FlashMessageContainer` directly,
the `addFlashMessage()` method must be used now.

The `FlashMessageContainer` can still be fetched from the
`ControllerContext` but this will internally now use the
`FlashMessageService` to restore the container from the configured
storage implementation.

Background:

This patch changes the default behavior to store FlashMessages in
a (HTTP) session cookie. This should not have any effect to the
PHP code and FlashMessages can still be obtained with the
`<f:flashMessages />` Fluid ViewHelper.
However, the FlashMessages can now also be read from JavaScript
allowing the redirected action to be cached!

Furthermore it's now possible to use different cookies for different
parts of the website so that FlashMessages don't interact.
This is useful for example for FlashMessages in the Neos Backend
and/or within plugins.


bwaidelich:
See motivation for this PR: https://discuss.neos.io/t/rfc-rework-flash-message-implementation/2513/4

albe:
Looking at the upcoming LTS and that we don't have any real major features for it (yet), can we somehow add this in a way that makes it less breaking and eligible for the LTS? Would be a nice thing to have in there

albe:
@bwaidelich would still be cool to get this in now with 6.0. Is there anything left to do technically other than resolving conflicts and writing documentation for the feature?

bwaidelich:
I agree. It would be cool to have this for 6.0 but currently it's not on top of my personal priority list

> Is there anything left to do technically other than resolving conflicts and writing documentation for the feature

No I don't think so

albe:
> > Is there anything left to do technically other than resolving conflicts and writing documentation for the feature
> 
> No I don't think so

Cool, so I think that's something that anyone can do then (myself included) :) Thanks for the quick reply!

albe:
Green 🎉 Only thing left is documentation now. And maybe reconsider the parameter ordering on `addFlashMessage`, because I'm not sure having the code as last parameter is optimal. Especially considering the order of arguments on the `Message` constructor, it would maybe be better to keep them in line.

albe:
Having psalm is awesome! I merged master into this, ran psalm and immediately found a couple of small bugs in this code. Even one that would not even make this work at all, because it still mutated the Http Response and did not replace it in the Http component.

Also fixed a whole couple of typehint issues.

albe:
```
ERROR: DuplicateClass - Packages/Framework/Neos.FluidAdaptor/Migrations/Code/Version20161130112935.php:19:7 - Class Neos\Flow\Core\Migrations\Version20161130112935 has already been defined in /home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Flow/Migrations/Code/Version20190425144900.php
class Version20161130112935 extends AbstractMigration
```

Another one I didn't even notice locally

albe:
Everythings green and merged the PSR-7 stuff in. Looks good now for me.

kdambekalns:
@albe Could you fix the psalm baseline conflict? Thanks!

albe:
The conflict was actually just a mismatch in the psalm-version header.

- strict_types: not sure, but seeing how Daniel already started adding those at various places... I'd probably do that a bit more separately, but no hard objections
- more types: maybe :)
- psalm-baseline: unfortunately yes, because this in fact introduced two new psalm errors that can not be (easily) solved here - namely calling `objectManager->get` with more than one param. It declares only one and uses `func_get_args()`.
See https://github.com/neos/flow-development-collection/pull/1061/commits/7cae16b80504083f46b72583c43e33bafbe157ee#diff-f03d2f6511f11079ec340295b48fd068R117 and https://github.com/neos/flow-development-collection/pull/1061/commits/7cae16b80504083f46b72583c43e33bafbe157ee#diff-d828396aac59073d4c3416476ed8a4bbR1727-R1732
the other baseline changes are resolved errors as a side effect of the above added error

albe:
> But why are strict_types not declared? We should do that for new code!

> bwaidelich commented on 17 Aug 2017

Indeed. The code is just not as new as it looks 😆 I'll take a look later today

kdambekalns:
I figured we'll be cleaning up things anyway, so LGTM.

PR #1063, opened 2017-08-18
BUGFIX: Skip setting value to Controller Argument if uri param is null

### LINKED ISSUE #1062: Controller Arguments Action mapping error
### Description

Look at this example method:
```
/**
 * @param Person $person
 */
public function showAction(Person $person) { ... }
```

I created the simple function with the required argument `$person`, then I configure the error exception handler in the `Settings.ymal` of all 401, 404, 500 as below:
```
TYPO3:
  Flow:
    error:
      exceptionHandler:
        renderingGroups:
          notFoundExceptions:
            matchingStatusCodes: [404]
            ....
          errorExceptions:
            matchingStatusCodes: [500]
            ...
```
Everything looks fine of the error page if there is a wrong typing uri. But it seems an error if the uri param has empty value. eg. `http://...domain...?person=` or `http://...domain...?person`

### Steps to Reproduce

1. Browse url with empty value of `person` param as above example
2. Then we see error as blow instead of the error page what we had configure
```
Uncaught Exception in Flow
Argument 1 passed to ...\Controller\PersonController::showAction() must be an instance of ...\Domain\Model\Persion, null given
TypeError thrown in file ...
```

#### Expected behavior

It would be always shown the error page which we had configured.

### Affected Versions

Neos: n/a

Flow: any

limchivy:
Fixes #1062

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
@@ -13,6 +13,7 @@ namespace Neos\Flow\Mvc\Controller;
 
 use Neos\Flow\Http\Response;
 use Neos\Flow\Mvc\Exception\ForwardException;
+use Neos\Flow\Mvc\Exception\InvalidArgumentValueException;
 use Neos\Flow\Mvc\Exception\RequiredArgumentMissingException;
 use Neos\Flow\Mvc\Exception\StopActionException;
 use Neos\Flow\Mvc\Exception\UnsupportedRequestTypeException;
@@ -357,6 +358,7 @@ abstract class AbstractController implements ControllerInterface
      * Maps arguments delivered by the request object to the local controller arguments.
      *
      * @return void
+     * @throws InvalidArgumentValueException
      * @throws RequiredArgumentMissingException
      * @api
      */
@@ -365,7 +367,11 @@ abstract class AbstractController implements ControllerInterface
         foreach ($this->arguments as $argument) {
             $argumentName = $argument->getName();
             if ($this->request->hasArgument($argumentName)) {
-                $argument->setValue($this->request->getArgument($argumentName));
+                $argumentValue = $this->request->getArgument($argumentName);
+                if ($argument->isRequired() && $argumentValue === '' && class_exists($argument->getDataType())) {
+                    throw new InvalidArgumentValueException('The argument "' . $argumentName  . '" is invalid.', 1505460827);
+                }
+                $argument->setValue($argumentValue);
             } elseif ($argument->isRequired()) {
                 throw new RequiredArgumentMissingException('Required argument "' . $argumentName  . '" is not set.', 1298012500);
             }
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -364,7 +364,7 @@ protected function mapRequestArgumentsToControllerArguments()
     {
         foreach ($this->arguments as $argument) {
             $argumentName = $argument->getName();
-            if ($this->request->hasArgument($argumentName)) {
+            if ($this->request->hasArgument($argumentName) && !empty($this->request->getArgument($argumentName))) {
```
albe: If we really only want to avoid setting `null` we should better explicitly test against null here. `empty` will also avoid empty string, zero values (also `"0"`), false and empty array. That's not what we want.

limchivy: @albe yes, you are right. Thank you :)
I will update it

limchivy:
@albe 
> Edit: One generic question though: How is it supposed to explicitly submit null for an property then, for example when you want to unset a value?

My previous thought:
- If the action param is required, it is not able to reset property value to `null` from the summit form
- If the action param is optional, it will get from the default value 

But not I understand my problem is caused from the Object converting if it is null value.
If it is not a Object param , it would work as your concerned question.

limchivy:
@albe There is a wrong argument validation at `Neos.Flow/Classes/Mvc/Controller/Argument.php:290` when there is a null `$rawValue`.

What do you think if we return `TargetNotFoundError` instead of `null` at `Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php:180`?

albe:
> There is a wrong argument validation at Neos.Flow/Classes/Mvc/Controller/Argument.php:210 when there is a null $rawValue.

Can you elaborate a bit more, maybe with an concrete example, what the issue is you see with that? Is that directly related to your current use-case that you try to resolve?

> What do you think if we return TargetNotFoundError instead of null at Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php:180?

The TypeConverter does not know or care about required/optional semantics, it just tries to convert some source into an instance of a specific persistent object. If the input is empty (string), it will return `null`, which is the closest equal to an empty object/object without identity. Returning `TargetNotFound` would be misleading, since it did not even attempt to find the "empty object" anywhere. Again, the use case here is, that it should be possible to set an object argument/property to `null` explicitly through an empty input like HTML forms commonly do.

limchivy:
@albe I recently updated the code, it would be good if you could help to review. Thanks.

kitsunet:
I would also need an example to understand the issue. I assume it needs to be solved differently. I guess we still do things wrongly in property mapping...

limchivy:
@kdambekalns It is happened when the argument is empty string (we can write uri as `url?object=` or `url?object`)

1. the `hasArgument()` is `true`, and its value is empty string
2. the mapping argument object is `null` if it is set the `setValue` is empty sting (PersistentObjectConverter behavior)

limchivy:
@kitsunet May you have a look the example case at #1062? And please let me know there is any concern.

kitsunet:
See comment in the issue. Feel free to re-open if you think the issue cannot be solved as described.

PR #1068, opened 2017-08-30
BUGFIX: Fix FLOW_VERSION_BRANCH value

gerhard-boden:
Use correct version within Bootstrap.php


gerhard-boden:
Also added this to our Google Docs documentation

PR #1069, opened 2017-08-31
BUGFIX: Don’t use scalar type hint string for RedisBackend

johannessteu:
The scalar type hint for `compress()` forced the return value of uncompress() to an emtpy string if there is no value in redis. This has to be false.


kitsunet:
A comment about why there is not (and cannot be) a return type declaration for now would be great :)

albe:
Thanks. One of those tedious "supposed contracts" = return type annotations, that just don't hold.

PR #1070, opened 2017-09-01
BUGFIX: Allow file uploads without arguments for value

malteriechmann:
I created an image upload for our Flow application. After updating Flow, it stopped working, because two array keys do not exist. To fix this error I suggest making these keys optional.

`$argumentsForValue['originallySubmittedResource']` and `$argumentsForValue['__collectionName']` must not exist. The file upload should not break, if these array keys are missing.

**Checklist**

- [ ] The PR is created against the [lowest maintained branch]

### DIFF
--- a/Neos.Flow/Classes/Http/Request.php
+++ b/Neos.Flow/Classes/Http/Request.php
@@ -653,11 +653,11 @@ class Request extends BaseRequest implements ServerRequestInterface
             $value['type']
         );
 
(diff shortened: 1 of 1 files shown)


albe:
What should happen if they exist but are null/emptyish? Maybe better check `empty(...)` instead? Also needs to be targeted at 4.2, since this is a Bugfix

PR #1071, opened 2017-09-01
BUGFIX: Result type hint leads to wrong function result

fritjofbohm:
The `int` type hint of `strpos(string $haystack, string $needle, int $offset = 0): int` casts a `FALSE` result of `mb_strpos` to `0`, which leads to a wrong result if `$haystack` does not contain `$needle`.

**Checklist**


albe:
Thanks for this!

PR #1073, opened 2017-09-01
BUGFIX: RedisBackend compress() method may return false

albe:


Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -434,7 +434,7 @@ public function setDatabase(int $database)
     /**
      * @param string $password
      */
-    public function setPassword($password)
+    public function setPassword(string $password)
```
albe: We explicitly test against empty string in `getRedisClient()`, so without the typehint this could potentially become null (or any other non-string value) and lead to an exception for failed authentication.

PR #1074, opened 2017-09-04
BUGFIX: Annotation reader breaks partial index creation

fritjofbohm:
The `FlowAnnotationDriver` uses a wrong property of a loaded `Index` annotation object, preventing partial indexes being created on DB platforms supporting it, and breaking schema generation if there are `flags` specified in the `@ORM\Index` annotation, too.

```
/**
 * @Flow\Entity
 * @ORM\Table(indexes={
 *     @ORM\Index(name="foreignrelation_null_idx", columns={"foreignrelation"}, options={"where":"foreignrelation IS NULL"})
 * })
 */
```


albe:
@fritjofbohm I'll rebase this on 3.3 this evening unless you find time to do so before.

fritjofbohm:
@albe Wow, you're fast.
Please go ahead. It's unlikely that I'll find the time today.

albe:
Had to push to my own fork, so created a new PR. Closing this in favor of the other, but I kept the original commit.

PR #1075, opened 2017-09-05
BUGFIX:  Use ``!empty`` instead of ``empty`` in condition

lisakampert:
**What I did**
Change the condition to ``!empty``, so that NULL-values get catched and will not tried to be set, which causes further errors.

### DIFF
--- a/Neos.Flow/Classes/Http/Request.php
+++ b/Neos.Flow/Classes/Http/Request.php
@@ -653,11 +653,11 @@ class Request extends BaseRequest implements ServerRequestInterface
             $value['type']
         );
 
-        if (empty($argumentsForValue['originallySubmittedResource'])) {
+        if (!empty($argumentsForValue['originallySubmittedResource'])) {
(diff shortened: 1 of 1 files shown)


kitsunet:
Thanks a lot! Should be merged as soon as travis is happy.

kitsunet:
(Also note: we apparently need tests for this)

gerhard-boden:
oh boy 😞 That one's on me! 
Note to self: never release in a hurry

kdambekalns:
@gerhard-boden so, does this mean it should go into a released branch, i.e. 4.2!?

albe:
We all could have noticed this @gerhard-boden, so the blame is on us as a team :) And yes, that lesson to take away is probably golden

kitsunet:
Yep 4.2 is correct. That is code touched by the PSR-7 change, so the bug doesn't exist pre 4.2

gerhard-boden:
Can't change the base on Github without getting wrong commits into this one. I could try it locally, gimme a sec

gerhard-boden:
it worked :)

PR #1076, opened 2017-09-05
BUGFIX: Annotation reader breaks partial index creation

albe:
The `FlowAnnotationDriver` uses a wrong property of a loaded `Index` annotation object, preventing partial indexes being created on DB platforms supporting it.

Retargeted follow-up to #1074


albe:
Consider this an approved review from me, so I'll merge this as of the approval from @kdambekalns on #1074

PR #1082, opened 2017-09-20
FEATURE: Make fizzle filter operations capable of dealing with arrays

p-weisk:
This commit amends the implementation for the ``^=``, ``$=`` and ``*=`` fizzle operators by checking if the given property is an array and if so, checking whether the filter matches in the following way:

* ``*=`` matches if the array contains an item that matches the filter string exactly.

* ``^=`` matches if the array's first item matches the filter string exactly.

* ``$=`` matches if the array's last item matches the filter string exactly.

If the given property is not an array the filter operations behave the same way as before.

This commit also contains unit tests for the array-based filter operations with the following test cases:
    * a) one array item matches the filter string with the filter operator -> filter should match
    * b) multiple items match the filter string with the filter operator -> filter should match
    * c) no item matches the filter string with the filter operator -> filter should not match
    * d) a substring of an item would match filter string with the filter operator if it wasn't in an array -> filter should not match
    * e) item would match filter string if different filter operator was used -> filter should not match

### DIFF
--- a/Neos.Eel/Classes/FlowQuery/Operations/Object/FilterOperation.php
+++ b/Neos.Eel/Classes/FlowQuery/Operations/Object/FilterOperation.php
@@ -42,11 +42,11 @@ use Neos\Utility\TypeHandling;
  * >=
  *   Value is greater than or equal to operand
  * $=
- *   Value ends with operand (string-based)
+ *   Value ends with operand (string-based) or value's last element is equal to operand (array-based)
  * ^=
- *   Value starts with operand (string-based)
+ *   Value starts with operand (string-based) or value's first element is equal to operand (array-based)
  * *=
- *   Value contains operand (string-based)
+ *   Value contains operand (string-based) or value contains an element that is equal to operand (array based)
  * instanceof
  *   Checks if the value is an instance of the operand
  * !instanceof
--- a/Neos.Eel/Classes/FlowQuery/Operations/Object/FilterOperation.php
+++ b/Neos.Eel/Classes/FlowQuery/Operations/Object/FilterOperation.php
@@ -234,11 +234,34 @@ class FilterOperation extends AbstractOperation
             case '>=':
                 return $value >= $operand;
             case '$=':
-                return strrpos($value, $operand) === strlen($value) - strlen($operand);
+                if (is_array($value)) {
+                    if ($this->evaluateOperator(end($value), '=', $operand)) {
+                        return true;
+                    }
+                    return false;
+                } else {
+                    return strrpos($value, $operand) === strlen($value) - strlen($operand);
+                }
             case '^=':
-                return strpos($value, $operand) === 0;
+                if (is_array($value)) {
+                    if ($this->evaluateOperator(reset($value), '=', $operand)) {
+                        return true;
+                    }
+                    return false;
+                } else {
+                    return strpos($value, $operand) === 0;
+                }
             case '*=':
-                return strpos($value, $operand) !== false;
+                if (is_array($value)) {
+                    foreach ($value as $item) {
+                        if ($this->evaluateOperator($item, '=', $operand)) {
+                            return true;
+                        }
+                    }
+                    return false;
+                } else {
+                    return strpos($value, $operand) !== false;
+                }
             case 'instanceof':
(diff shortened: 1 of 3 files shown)


daniellienert:
Hey @p-weisk, 
welcome and thanks a lot for your first contribution to the project - its a great pleasure to me personally!
Unfortunately, something went wrong with your PR, looks like you forked your branch from an other branch as you like to merge into. This one is not mergeable.
Also, the change looks more like a feature then a bugfix, as it enables something that was not promised / broken before. I would recommend to target it as a feature on master.

Cheers, 
Daniel

p-weisk:
Hey @daniellienert ,

thanks for your comment and sorry for the broken pull request. I changed the branch to master now. Since I changed it from bugfix to feature now there is no need to merge it into the 2.3 branch anymore, right?

Greetings
Paul

daniellienert:
Right.

albe:
Thanks for this, looking good from my side. @daniellienert can you revisit your review and possibly approve?

aertmann:
Hey @p-weisk, thanks for your contribution. I can see the use case for this behavior, although doubt it's a very common one. However the behavior is not intuitive if you ask me and there's nothing similar in jQuery as it DOM doesn't have array properties, hence I'm a bit reluctant about adding it.

Another way to achieve what you're doing is to do s.th. like

``*=``: ``${Array.indexOf(q(node).property('arrayProperty')) !== -1)``
``^=``: ``${Array.first(q(node).property('arrayProperty')) === 'abc')``
``$=``: ``${Array.last(q(node).property('arrayProperty')) !== 'abc')``

Would that work for you or does it need to be an EEL filter?

Would like some more opinions on it (cc @skurfuerst @kitsunet). At the very least it needs to be documented.

dfeyer:
I'm with @aertmann here, mixed feeling ... not sure if we want this in the core ... Array.first/last can be use and the synthax is clean an easy to read. 

But there are cases where more advanced operator can make a lots sense, like in postgres your have the `@>` (contain operator) so you can test is JSON array property contains a string like this:

    WHERE info->'food' @> '"carrots"';

This kind of operator can make sense. I think we don't need to be too dogmatic about "it does not exist in DOM query" ... because we don't query the DOM here, but we can deal a lots with JSON style structure. So maybe taking some inspiration from Psql can be good, like:

- `?` Does the key/element string exist within the JSON value?
-  `?|` Do any of these key/element strings exist?
- `?&` Do all of these key/element strings exist?

@p-weisk Where did you take your proposed synthax ?

albe:
I have to say, I like the transport of the known operators "contains" (`*=`), "starts with" (`^=`) and "ends with" (`$=`) to arrays. After all, that's exactly what they do on arrays like they do on strings (which could be considered arrays of characters). The operator is the same there, so it makes sense to use the same syntax and not invent something new (unless we indeed want to extend to something like @dfeyer suggested).

dfeyer:
@daniellienert Can you review again ?

aertmann:
Still not a fan of the proposal I must say. The syntax just doesn't feel natural to me and it can be achieved in other ways as I explained. The point is not that we're not working with the DOM here, but that jQuery struck a good balance between functionality and API surface area. If this is a very common use case that would be difficult to solve in different ways, then the proposal makes sense. I just don't see that being the case here.

Food of thought, this change doesn't include any documentation. I don't think we can really make a decision without seeing how well this proposal fits with the documentation.

Sorry for being the devils advocate here, but we should always be conservative when deciding on a end user API.

albe:
Fair enough :) And you are right regarding documentation. We should always expect documentation for new features before accepting.

albe:
This has somehow been forgotten. Maybe time to accept this finally in for 5.1? @daniellienert ?

jonnitto:
> This commit also contains unit tests for the array-based filter operations with the following test cases:
    1. one array item matches the filter string with the filter operator -> filter should match
    2. multiple items match the filter string with the filter operator -> filter should match
    3. no item matches the filter string with the filter operator -> filter should not match
    4. a substring of an item would match filter string with the filter operator if it wasn't in an array -> filter should not match
    5. item would match filter string if different filter operator was used -> filter should not match

PR #1083, opened 2017-09-21
BUGFIX: Force PropertyMapper to correctly handle bool in constructor

monofone:
**What I did**
Found a bug in the propertymapping when mapping from scalar bool to object. Fixed it in the Typeconverter and added tests 

**How I did it**
Fixed the ScalarTypeToObjectConverter for bool scalar type.

**How to verify it**
A ValueObject with the constructor signature containing `__construct(bool $value)` can be build from a scalar boolean value.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Property/PropertyMapper.php
+++ b/Neos.Flow/Classes/Property/PropertyMapper.php
@@ -260,7 +260,7 @@ class PropertyMapper
     {
         $targetClass = TypeHandling::truncateElementType($targetType);
         if (!class_exists($targetClass) && !interface_exists($targetClass)) {
-            throw new Exception\InvalidTargetException(sprintf('Could not find a suitable type converter for "%s" because no such the class/interface "%s" does not exist.', $targetType, $targetClass), 1297948764);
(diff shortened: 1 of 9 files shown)


robertlemke:
Hey, that looks good now. Let's merge it!

bwaidelich:
FYI (not to blame): This one seems to have broken the Neos export (see https://github.com/neos/neos-development-collection/issues/1732).
I would like to keep this, so I'm looking for an alternative fix. But we might need to revert it as hotfix

kitsunet:
I am still pretty sure the one chnage in the StringConverter has to be chnaged back and instead a normalizeType in another place added. Actually we should try to create a test in the same way but for int because that should show the same issue and fail currently. Maybe that gives a hint where this should actualyl be fixed.

bwaidelich:
@monofone I reverted this for now as a hotfix (see #1109)

kdambekalns:
The was followed by https://github.com/neos/flow-development-collection/pull/1086 and https://github.com/neos/flow-development-collection/pull/1087 - which seem to do what is needed for `StringConverter` according to the discussion at hand (in #1109.)

bwaidelich:
FYI: In the meantime I have fully (hopefully) understood the problem and will re-add this fix with a slight adjustment

PR #1085, opened 2017-09-21
BUGFIX: Make string-based Fizzle filters treat numbers as strings

p-weisk:
This fixes #1048 by making sure that `strpos()` recieves two strings instead of a
string and a number in a Fizzle filter operation (`*=`, `^=`, `$=`), making sure that
it will look for the actual number in the string it is comparing the operand to instead
of looking for the char that is represented by the number as a unicode ordinal number.

It does this by casting the variable in which the operand is saved to string where it
is given to the strpos/strrpos function as a parameter.

### DIFF
--- a/TYPO3.Eel/Classes/TYPO3/Eel/FlowQuery/Operations/Object/FilterOperation.php
+++ b/TYPO3.Eel/Classes/TYPO3/Eel/FlowQuery/Operations/Object/FilterOperation.php
@@ -234,11 +234,11 @@ class FilterOperation extends AbstractOperation
             case '>=':
                 return $value >= $operand;
             case '$=':
-                return strrpos($value, $operand) === strlen($value) - strlen($operand);
+                return strrpos($value, "$operand") === strlen($value) - strlen($operand);
             case '^=':
-                return strpos($value, $operand) === 0;
+                return strpos($value, "$operand") === 0;
             case '*=':
-                return strpos($value, $operand) !== false;
+                return strpos($value, "$operand") !== false;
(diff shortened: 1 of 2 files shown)


albe:
Hey, sorry for being slow on giving feedback. Thank you very much and congrats on your first contribution. Thank you also for providing a test with this!

I'm not that much into the Fizzle Parser, but this sounds and looks reasonable. My only concern would be if the string interpolation has performance implications.
I'll bring this PR up in the team to get feedback and possibly merge it.

kdambekalns:
Seems to work just as well. :)

albe:
Thanks everyone :) I'll merge this then

bwaidelich:
Not to blame just FYI: Apparently the Travis CI don't run the `flow-development-collection` with PHP7.1 so this slipped through, but the unit tests fails on PHP 7.1 (see https://travis-ci.org/neos/neos-development-collection/jobs/289601532).
The line
```php
return strrpos($value, (string)$operand) === strlen($value) - strlen($operand);
```
probably needs to be changed to
```php
return strrpos($value, (string)$operand) === strlen($value) - strlen((string)$operand);
```
(note the additional type cast)

I can't take care of it now, but I'll look into it asap if noone else takes care

kitsunet:
the actual problem is probably ONLY in the test `$this->assertInstanceOf(FlowQuery::class, $query->filter("[stringProperty $= " + $filterString + "]"));` uses `+` for concatenation of strings which is not correct for PHP. Should be `.`. Not a big deal. Can fix later.

PR #1088, opened 2017-09-29
BUGFIX: Respect given locale when checking general translation path

nezaniel:
for label overrides

This seems to have been forgotten in the original implementation and lead to inconsistent behavior when override labels were present in multiple languages


albe:
I didn't even know this feature existed... and it's also not documented (http://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Internationalization.html#file-locations-and-naming).

Serious question: When/why would one place a folder of Translations in the `Data` directory of the Flow installation?

nezaniel:
You would put translations there if they are imported from a 3rd party system and thus cannot go to your package repository.

This behaviour was introduced in 4.2 and as it seems I have been sloppy on documentation... It's definitively a new feature but non-breaking since it doesn't affect the current behaviour if no files are found there.

albe:
Ok, then see my comment as a "please document this feature" request :)

nezaniel:
Will do, as soon as I've written the missing tests :)

albe:
Thanks, that looks good. Now only the `globalTranslationPath` setting needs to be added to the Settings Schema, then the test will be green too and this should be ready to merge :)

albe:
Thanks :) Would merge this once Travis passes this weekend if no one objects.

nezaniel:
Done. There seems to be an issue with the test using the vfs... Is there anything I can do about that?

albe:
I restarted the job. There were also some issues with connecting to PPA repository, so I assume it was a temporary connectivity issue.

albe:
Ok, that wasn't the issue apparently... need to investigate, but the test failure shouldn't really be related to this afais...

nezaniel:
you mean other tests that initialize the i18n stuff will indirectly invoke the service without explicitly invoking the vfs?
In that case I'd have to remove that part of the test config and somehow brute force-inject the path where it is used, correct?

albe:
I had to think for a few moments myself, but yeah, I think that would be a good solution.

mficzel:
I added a pr for this branch https://github.com/CORE4/flow-development-collection/pull/1

Basically three things were done in there: 
- Use localeFallbackChain
- merge current 4.2 branch into the pr 
- fix the tests by placing the globalTranslationPath in one of the mocked packages

mficzel:
I found another small issue during testing https://github.com/CORE4/flow-development-collection/pull/2 

Without this change a global translation that in the local fallback chain would override a
better suiting translation from a package.

kdambekalns:
Do I get it right that despite three approvalsthe PR to the PR by @mficzel needs to be merged first? If that's the case, please "request changes" next time, @mficzel :)

PR #1089, opened 2017-10-04
BUGFIX: Arrays::setValueByPath is too strict

dfeyer:


albe:
Thanks! Another of those pesky little buggers... return types are hard :/

kitsunet:
mmm, I see. Ok that really makes sense. Bäääääh PHP 😆

bwaidelich:
Thanks for this fix, this was actually a quite nasty one that lead to loads of exceptions in a project of ours because this method is called during `__sleep` of all objects(!)

Issue #751, opened 2016-12-02
Documentation wrongly says that Flow only supports PSR-0 autoloading


markwatney2016:
### Description

Documentation about autoloading is out of date.

> Currently Flow only supports PSR-0 autoloading and will only use the first entry given in the manifest.
...
Other autoloading ways (classmap generation and files) are currently not supported by Flow.

### Steps to Reproduce

http://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/PackageManagement.html#using-3rd-party-packages

albe:
Anyone with autoloading insights out to fix up this documentation section? @kitsunet @kdambekalns ? I'm not even sure the thing about "only use the first entry" is correct any more

kdambekalns:
Considering https://github.com/neos/flow-development-collection/blob/8f13a0dfc143e851a0e596e37e8d3f8e44e21276/Neos.Flow/Classes/Core/ClassLoader.php#L28-L46 this is no longer true, indeed…

kitsunet:
I can try to fix it

PR #1090, opened 2017-10-08
TASK: Describe use of 3rd party packages correctly in docs

kitsunet:
Adjusted documentation to current status quo regarding package
management and third party packages.

Fixes: #751


albe:
@aertmann yep, but that's of course not responsibility of this PR :) Thanks Christian!

Issue #1065, opened 2017-08-21
Using scalar return-type declaration 'void' results in an fatal error


MarkusRodler:
### Description

It is not possible to use the scalar type 'void' as an return type. If void is used this error is thrown: "Fatal error: Scalar type declaration 'void' must be unqualified [...]"
Other scalar types as 'string' or 'int' are working as expected.
The problem is within the generated sources because the void return-type will be qualified: \void

Example code:
```php
/**
 * @Flow\Scope("singleton")
 */
class ExampleController extends AbstractController
{
    public function indexAction(): void
    {
    }
}
```

### Steps to Reproduce

1. Make a controller with indexAction
2. Add ': void' as the return-type of that action
3. Open that Action in your browser and error is thrown

#### Expected behavior

No error

#### Actual behavior

Error is thrown: "Fatal error: Scalar type declaration 'void' must be unqualified [...]"

### Affected Versions

all

PR #1091, opened 2017-10-08
BUGFIX: Correct use of `void` return type in proxy classes

kitsunet:
As `void` was not on the list of return types that would be
normalized to include a leading backslash in the reflection
data it would be output wrongly in proxy classes and so break
applications that used it.

Fixes: #1065

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Reflection/ReflectionService.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Reflection/ReflectionService.php
@@ -1389,15 +1389,13 @@ class ReflectionService
         }
 
         $returnType = $method->getDeclaredReturnType();
-        if ($returnType !== null) {
-            if (!TypeHandling::isSimpleType($returnType) && !in_array($returnType, ['self', 'null', 'callable'])) {
-                $returnType = '\\' . $returnType;
-            }
-            if ($method->isDeclaredReturnTypeNullable()) {
-                $returnType = '?' . $returnType;
-            }
-            $this->classReflectionData[$className][self::DATA_CLASS_METHODS][$methodName][self::DATA_METHOD_DECLARED_RETURN_TYPE] = $returnType;
+        if ($returnType !== null && !in_array($returnType, ['self', 'null', 'callable', 'void']) && !TypeHandling::isSimpleType($returnType)) {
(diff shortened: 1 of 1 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Reflection/ReflectionService.php
```diff
@@ -1389,15 +1389,13 @@ protected function reflectClassMethod($className, MethodReflection $method)
...
-                $returnType = '?' . $returnType;
-            }
-            $this->classReflectionData[$className][self::DATA_CLASS_METHODS][$methodName][self::DATA_METHOD_DECLARED_RETURN_TYPE] = $returnType;
+        if ($returnType !== null && !in_array($returnType, ['self', 'null', 'callable', 'void']) && !TypeHandling::isSimpleType($returnType)) {
+            $returnType = '\\' . $returnType;
+        }
+        if ($method->isDeclaredReturnTypeNullable()) {
+            $returnType = '?' . $returnType;
         }
+        $this->classReflectionData[$className][self::DATA_CLASS_METHODS][$methodName][self::DATA_METHOD_DECLARED_RETURN_TYPE] = $returnType;
```
aertmann: just to point it out, now this is set to null

kitsunet: Yep, IMHO that makes it actually more consistent because the key will always be set. In the getter it's an `isset` check anyways.

albe:
I just took a look and there are barely any UnitTests on the ReflectionService. I also have a hard time thinking about how to properly unit test such things in isolation, since the `reflectClassMethod` method only updates internal state, so it's really not easy to test it's behaviour.

So the only thing to do for now would be another functional test, but a full functional test only for this single return type feels like overdoing. I'd still prefer that to not having it covered, but since `void` return type only is available in PHP 7.1, we have the issue again of how to properly test this in a way to work with old PHP versions.

Therefore I'd go with merging this as is (unless someone else comes up with a proper test) for now.

albe:
Any objections @aertmann?

aertmann:
nope @albe

bwaidelich:
FYI: This didn't fix the issue entirely it seems: #1151 
let's create a functional test for this

Issue #1078, opened 2017-09-12
composer create-project use abandoned packages


jkhaled:
### Description

[when creating project with composer create-project it use abandoned packages]

### Steps to Reproduce

1. [go to the flow framework installation page  with [install](https://github.com/neos/flow/blob/1350f25facea561f809527a3baf62c7b1e08ee32/Documentation/TheDefinitiveGuide/PartII/Installation.rst)  ]
2. I followed the steps of project creation using composer create-project
3. [I run the cmd `composer create-project --keep-vcs neos/flow-base-distribution tutorial 3.0.0]`

#### Expected behavior
  
create project with last version of flow framework library

#### Actual behavior

I got this message that said the packages used are abondoned 

> Package typo3/flow is abandoned, you should avoid using it. Use neos/flow instead.
Package typo3/eel is abandoned, you should avoid using it. Use neos/eel instead.
Package flowpack/behat is abandoned, you should avoid using it. Use neos/behat instead.
Package typo3/welcome is abandoned, you should avoid using it. Use neos/welcome instead.
Package typo3/kickstart is abandoned, you should avoid using it. Use neos/kickstarter instead.

### Affected Versions

Neos:

> typo3/buildessentials             3.0.5              TYPO3 Flow Build Toolchain Essentials
typo3/eel                         3.0.15             The Embedded Expression Language (Eel) i...
typo3/flow                        3.0.15             Flow Application Framework
typo3/fluid                       3.0.15             Next-Generation Templating Framework for...
typo3/kickstart                   3.0.15             A simple generator for controller and vi...
typo3/welcome

Flow:

You can get this information by running `composer show`.

bwaidelich:
@jkhaled Thank you for this report.
Version 3.0 of Flow still uses the `typo3/*` packages that have been moved to `neos/*` for the next major version (the currently latest stable version of Flow is 4.2).
You can safely ignore those warnings if you really intended to install 3.0

I'll leave this ticket open though as a reminder to update the documentation accordingly

PR #1092, opened 2017-10-08
TASK: Document installing latest stable instead of fixed version

kitsunet:
Instead of documenting how to install a specific fixed version
just document installing the latest stable version. If people
want to install an older version they can figure that out.
This prevents errors due to forgotten updates to this section
of the documentation.

Fixes: #1078

### DIFF
--- a/Neos.Flow/Documentation/TheDefinitiveGuide/PartII/Installation.rst
+++ b/Neos.Flow/Documentation/TheDefinitiveGuide/PartII/Installation.rst
@@ -20,11 +20,12 @@ which boil down to this in the simplest case:
 
 Then use `Composer`_ in a directory which will be accessible by your web server to download
 and install all packages of the Flow Base Distribution. The following command will
-clone the 3.0.0 version, include development dependencies and keep git metadata for future use:
+clone the latest stable version, include development dependencies and keep git metadata
+for future use:
 
 .. code-block:: sh
 
- composer create-project --keep-vcs neos/flow-base-distribution tutorial 3.0.0
+ composer create-project --keep-vcs neos/flow-base-distribution tutorial
 
 .. note::
 	Throughout this tutorial we assume that you installed the Flow distribution in
@@ -150,7 +151,7 @@ with Flow contains this code already
 		Options -MultiViews
 
 	</IfModule>
-  
+
 Configure server-side scripts
 -----------------------------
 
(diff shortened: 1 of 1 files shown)


bwaidelich:
Why not add a note "This will install the latest stable version of Neos. In order to install a specific version, type..."? (Sry, can't do it myself atm cause on mobile)

Neos.Flow/Documentation/TheDefinitiveGuide/PartII/Installation.rst
```diff
@@ -20,11 +20,19 @@ which boil down to this in the simplest case:
...
 
- composer create-project --keep-vcs neos/flow-base-distribution tutorial 3.0.0
+ composer create-project --keep-vcs neos/flow-base-distribution tutorial
+
+This will install the latest stable version of Neos. In order to install a specific version, type
+
+.. code-block:: sh
+ composer create-project --keep-vcs neos/flow-base-distribution tutorial 3.3
+
+to install the last LTS version for example.
```
bwaidelich: now we'll have to maintain this again and don't forget to update the version with the next LTS. Why not use a placeholder here?

albe: I'm always bad with placeholders, I never find them obvious enough to not make any mistakes or create questions (is this character part of the placeholder or not?). Do you have any suggestion for how this could look?

bwaidelich: But wasn't that the point of this PR to get rid of the fixed version string? ;)

> Do you have any suggestion for how this could look?

Maybe sth like this:

```
The following command will clone the latest stable version, include development dependencies and keep git metadata for future use into a folder `tutorial`:

    composer create-project --keep-vcs neos/flow-base-distribution tutorial

In order to install a *specific version*, type

    composer create-project --keep-vcs neos/flow-base-distribution <target-directory> <version>

And replace `<target-directory>` with the folder name to create the project in and `<version>` with the specific version to install, for example `1.2`.
See [Composer documentation](https://getcomposer.org/doc/03-cli.md#create-project) for further details.
```
?

albe: I couldn't have come up with this somehow :)

albe:
Nothing annoying about being constructive :)

PR #1093, opened 2017-10-12
BUGFIX: Add missing var annotation to Uri::defaultPortsForScheme

bwaidelich:


albe:
Gonna merge this, as it's a real no-brainer.

PR #1094, opened 2017-10-13
BUGFIX: Restore XLIFF translation fallback

fritjofbohm:
This modification fixes the translation fallback mechanism and the ability to read regionalized language variants.

Issues fixed:

* Translation fallback did not work. `Neos.Flow.i18n.fallbackRule.order` was ignored when looking for XLIFF files
* When using a regional `Neos.Flow.i18n.defaultLocale` (e.g. `de_AT` for an Austrian version of `de`), XLIFF files were never read from  a directory with the full `defaultLocale` name, but only from the base language directory. In the example, files in `Packages/Application/Vendor.Package/Resources/Private/Translations/de_AT` were never read, only from`Packages/Application/Vendor.Package/Resources/Private/Translations/de`.

### DIFF
--- a/Neos.Flow/Classes/I18n/Xliff/Service/XliffFileProvider.php
+++ b/Neos.Flow/Classes/I18n/Xliff/Service/XliffFileProvider.php
@@ -94,11 +94,12 @@ class XliffFileProvider
      */
     public function getMergedFileData($fileId, Locale $locale): array
     {
-        if (!isset($this->files[$fileId][$locale->__toString()])) {
+        if (!isset($this->files[$fileId][(string)$locale])) {
             $parsedData = [
                 'fileIdentifier' => $fileId
             ];
             $localeChain = $this->localizationService->getLocaleChain($locale);
+            // Walk locale chain in reverse, so that translations higher in the chain overwrite fallback translations
             foreach (array_reverse($localeChain) as $localeChainItem) {
                 foreach ($this->packageManager->getActivePackages() as $package) {
                     /** @var PackageInterface $package */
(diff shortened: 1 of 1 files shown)


albe:
Thanks, this makes sense! Left two comments. Also, this should probably be targeted lower (3.3 even as it's LTS and still in bugfix support).

albe:
Thanks! Can you also rebase this on 3.3, since simply changing the base won't work in this case.

fritjofbohm:
This file is new in 4.2, so we can't rebase this to 3.3. Before 4.2, Flow used a different approach in loading XLIFF files (see https://github.com/neos/flow-development-collection/pull/894)

albe:
Ah yes, of course! My bad :) Then this just needs a rebase on 4.2, which should be pretty easy.

Neos.Flow/Classes/I18n/Xliff/Service/XliffFileProvider.php
```diff
@@ -87,26 +94,30 @@ public function initializeObject()
      */
     public function getMergedFileData($fileId, Locale $locale): array
     {
-        if (!isset($this->files[$fileId][$locale->getLanguage()])) {
```
aertmann: hmm feels wrong to me to rely on the magic `__toString()` method, maybe add a `getIdentifier` or `getLocaleIdentifier` method and use that instead? `__toString()` could do that same then of course..

but maybe that's just me..

fritjofbohm: I'd say a `getIdentifier()` method in `\Neos\Flow\I18n\Locale` is generally a good thing, but it's also a medium change to Flow if you consequently replace each `Locale` string cast with the new method. This is just a humble bugfix ;)

aertmann: you're right, seems like it's used a bunch of other places in the code so guess someone meant to use that way.. nvm then

fritjofbohm:
OK, I did a rebase of this stuff on 4.2. Please let me know if I did it right

albe:
Seems a couple of unrelated commits sneaked in. Try pulling latest 4.2 branch, then (force) rebase on that and skip all but your two commits (if there are still any other).

fritjofbohm:
Next try. Better this time?

PR #1097, opened 2017-10-17
TASK: Add a default .editorconfig

mstruebing:
This introduces a default ``.editorconfig`` file in the root of the freshly created project.

It ensures that php follows the PSR-2 coding-style, yaml gets 2 spaces indentation and md files doesn't trim trailing whitespace.

### DIFF
--- a/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/.editorconfig
+++ b/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.php]
+indent_type = space
+indent_size = 4
+
+[*.yaml]
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = false
--- a/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/.editorconfig
+++ b/TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/.editorconfig
@@ -5,13 +5,10 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-
-[*.php]
 indent_type = space
 indent_size = 4
 
 [*.yaml]
-indent_type = space
 indent_size = 2
 
 [*.md]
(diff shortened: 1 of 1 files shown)


TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+[*.php]
+indent_type = space
+indent_size = 4
+
+[*.yaml]
+indent_type = space
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = false
```
kdambekalns: Why is this a good idea and/or needed?

kdambekalns:
What about indentation of JS & HTML files? No rules added for those…

TYPO3.Flow/Resources/Private/Installer/Distribution/Defaults/.editorconfig
mstruebing: Because double whitespace at the end of a line in markdown enforces a line break. http://markdown-guide.readthedocs.io/en/latest/basics.html#line-return

mstruebing:
HTML and JS isn't always used in a Flow project so I thought it isn't necessary to add rules for these.
PHP, Markdown and Yaml files are always present.
I can add default rules for HTML and JS but especially in JS it really depends on your set up and taste.

mstruebing:
@kdambekalns would you like if I add `tab` as default indendation(`[*]`)?

albe:
I'd also probably not imply a code style on languages that we do not provide something from, because there is then no friction between framework code and user code. Also, those other languages don't fall under some PSR or other "standard". Let the user decide for those cases themself.

kitsunet:
Every other file type is additionally "feature" IMHO. But IF we add any blanked rule then spaces please, no tabs. Btw. fusion files?

albe:
Fusion!

kdambekalns:
- Double-space at end of line is magic in md? WTF!? 
- Tabs? Hell, no!
- Fusion!

mstruebing:
Now I there are default 4 spaces indentation which is also effective for fusion and php, no need to seperate them in this case.

albe:
Not sure if we should put this into 3.1, since that is only supposed to receive security fixes - but 3.3 should be fine.

mstruebing:
Changed base branch to 3.3.

PR #1099, opened 2017-10-17
FEATURE: Add cookie support on curl request

dhelfer:
`Neos\Flow\Http\Client\CurlEngine` will now attach cookies to an outgoing request.

### DIFF
--- a/Neos.Flow/Classes/Http/Client/CurlEngine.php
+++ b/Neos.Flow/Classes/Http/Client/CurlEngine.php
@@ -142,7 +142,7 @@ class CurlEngine implements RequestEngineInterface
             foreach ($request->getCookies() as $cookie) {
                 $cookies[] = $cookie->getName() . '=' . $cookie->getValue();
             }
-            curl_setopt($curlHandle, CURLOPT_COOKIE, implode(';', $cookies));
+            curl_setopt($curlHandle, CURLOPT_COOKIE, implode('; ', $cookies));
         }
 
         $curlResult = curl_exec($curlHandle);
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Http/Client/CurlEngine.php
```diff
@@ -137,7 +137,13 @@ public function sendRequest(Http\Request $request)
             curl_setopt($curlHandle, CURLOPT_PORT, $requestUri->getPort());
         }
 
-        // CURLOPT_COOKIE
+        if (count($request->getCookies()) > 0) {
+            $cookies = array();
+            foreach ($request->getCookies() as $cookie) {
+                $cookies[] = $cookie->getName() . '=' . $cookie->getValue();
+            }
+            curl_setopt($curlHandle, CURLOPT_COOKIE, implode(';', $cookies));
```
albe: I think the glue needs to be `'; '` (see https://forums.phpfreaks.com/topic/77789-solved-multiple-cookies-in-curl-curlopt-cookies/) and unsure if a delimiting semicolon is required, but the CURL documentation (https://curl.haxx.se/libcurl/c/CURLOPT_COOKIE.html) has it.

dhelfer: you are right on the glue, there has to be a space.
according to the RFC there is no need for a delimiting semicolon.
https://tools.ietf.org/html/rfc6265#section-4.2.1
I hope I will find some time in the near future to provide a test for this feature.

albe:
Ok, so after thinking and playing a bit, I think to test this effectively, we'd need to be able to start a PHP internal webserver from functional tests, that invokes a simple request info dumping script. For this, we'd need to abstract what is done in `SubProcess` one level higher, since we need to be able to stop the PHP server after the test invokation.

This is far beyond the scope of this PR, so I'd just merge this for now and open an issue with the idea to build this testing infrastructure first.

PR #1100, opened 2017-10-19
TASK: Remove some PR template artifacts from changelog

albe:


albe:
Merging this now, since it's just cosmetic on the changelog and the failures are not related to this anyway.

PR #1101, opened 2017-10-19
TASK: Fix FlowQuery Tests for PHP 7.1

albe:
The newly added test in https://github.com/neos/flow-development-collection/pull/1085 made the test fail in PHP 7.1 due to wrongly trying to concat strings with `+`.


albe:
Insight: The PHP 7.1 tests for travis are only active post 3.3 branch, but this and the previous were targeted on 3.3, hence this issue wasn't noticed. Guess we need to make travis runs after upmerges more visible.

PR #1104, opened 2017-10-27
TASK: Improve property mapping exception message

albe:
The exception message was way too technical and in a lot of cases contained the useless addition `, at property path ""`.


albe:
> however I've always felt that the missing path is due to something else as it's always related to some property.

In most cases when the path is missing, it is a more generic issue with property mapping/configuration, yes. But since we can't know exactly what at this level, it's IMO best to just avoid the unnecessary path output and depend on the original exception message to contain helpful information.

That in turn could also be improved, for example for the "is not allowed to be mapped" case, but that's another PR IMO.

Issue #1106, opened 2017-11-07
Normalize scalar type names to the built-in PHP names


bwaidelich:
The `TypeHandling::normalizeType()` utility function converts the short scalar type "int" and "bool" to their longer equivalent (see [source](https://github.com/neos/flow-development-collection/blob/master/Neos.Utility.ObjectHandling/Classes/TypeHandling.php#L76-L90)).

This should be turned around for consistency, as PHP scalar type hints also use the short notation.

Fix version: Lowest supported branch (currently Flow 3.3 LTS)

bwaidelich:
I‘m not 100% certain if it‘s worth the breaking change to be honest and one could argue that the current behavior is in sync with PHPs ˋgettype()ˋ return values.
In any case, to fix https://github.com/neos/neos-development-collection/issues/1732 we‘ll need a different solution that works for the older branches

kitsunet:
This is not declared public API in any case. And mostly used in conjunction with tags/annotations which again obviously are closely related to type declarations. Those again use the short form, eg. currently 

```
/**
 * @param boolean $foo
 */
function bar(bool $foo) {
```

this is our guideline. First of all we should suggest using short forms in the param tag as well. Second internally the param is currently always normalised to `boolean` whereas the type declaration is not because we need the original (and only working) string `bool` to be used in proxy classes. That's why I suggested since a while to change internal defaults to use short forms everywhere so we are consistent with the type declarations. Obviously that means a `gettype` result needs to be run through `normalizeType` as well, but we don't use gettype a lot and I think it's ok to do that.

kitsunet:
And first of all I would strongly suggest to revert the change made tot he string converter in here:
https://github.com/neos/flow-development-collection/pull/1083

I guess somewhere another normalize call is missing if it doesn't work with `boolean`. That would be the right fix for now and then we can take our time finding a good solution for the mess...

kdambekalns:
> as PHP scalar type hints also use the short notation

From #1109: Long and short form still work in PHP for type declarations:

    > $f = function (int $i): bool {};
    > $f = function (integer $i): boolean {};

Also, `is_integr()` still exists, even if it's not in the documentatoin anymore:

    > var_dump(is_int(1));
    bool(true)
    > var_dump(is_integer(1));
    bool(true)

The `settype()` function accepts both as well:
    
    > $var='1'; settype($var, 'int'); var_dump($var);
    int(1)
[...]

PR #1107, opened 2017-11-07
BUGFIX: Normalize scalar type names to the built-in PHP names

bwaidelich:
This adjusts `TypeHandling::normalizeType()` to return the short
notation (`bool` and `int`) for booleans and integers.
Previously it was the other way around and the short names were
converted to the longer equivalents.

Background:
Because the types were used inconsistently the `PropertyMapper` could
not find the correct type converters. Leading to errors like this one:
https://github.com/neos/neos-development-collection/issues/1732

Fixes: #1106


bwaidelich:
Closing this in order to avoid that this gets merged by accident.
I'll open a new PR against master for older branches we need a B/C fix in Neos (see https://neos-project.slack.com/archives/C050C8FEK/p1510073417000270)

PR #1109, opened 2017-11-08
Revert "BUGFIX: Force PropertyMapper to correctly handle bool in constructor"

bwaidelich:
Reverts neos/flow-development-collection#1083

Background: That PR broke the Neos export for boolean properties (see https://github.com/neos/neos-development-collection/issues/1732).
Also it was merged into 4.0 and not into the lowest supported branch (3.3) as stated.

I revert this now as a hotfix because I failed to completely comprehend the PR (I changed the `StringConverter` back and all tests were still green).


Neos.Flow/Classes/Property/TypeConverter/StringConverter.php
```diff
@@ -76,7 +76,7 @@ class StringConverter extends AbstractTypeConverter
     /**
      * @var array<string>
      */
-    protected $sourceTypes = ['string', 'integer', 'float', 'bool', 'array', \DateTimeInterface::class];
+    protected $sourceTypes = ['string', 'integer', 'float', 'boolean', 'array', \DateTimeInterface::class];
```
albe: I think only this was the actual breaking part of the original change, because the TypeConverter map will look up by the normalized type of the source, and hence this TypeConverter would no longer match for `boolean`. On top, the above `canConvertFrom()` method probably also would need to be changed to sth. like `TypeHandling::normalizeType($methodParameter['type']) === TypeHandling::normalizeType(gettype($source));`.

bwaidelich: Yes, I think so too. But I couldn't confirm because no test failed so I'd rather revert this completely and then re-apply it cleanly (and possibly to a different branch)

gerhard-boden:
Thanks for taking care guys!

gerhard-boden:
Want to run a bugfix release this week, since many people seem to be affected by https://github.com/neos/neos-development-collection/issues/1732 according to the daily slack messages

kdambekalns:
For the record, @robertlemke said in Slack:

---

Sorry, I don't have much time right now, but for the record: The fix was merged for 4.0 because the bug was introduce in 4.0 (through the adjustments for PHP 7.x if I remember correctly). The patch did fix the described bug (using @var bool as type hint, for example in action methods). There was no failing test before this fix, because the tests were wrong. But of course it shouldn't break the importer, so we'll have to find another fix. However, just reverting the patch will break the @bool handling again.

---

And I was saying: `bool` must be used in PHP, like `int` for type declarations now. "We" don't like abbreviations, thus we always went for `boolean` and `integer` in "anything Flow". To avoid confusion, we should be using `bool` and `int` consistently now, and make sure anyone using `boolean` and `integer` is still getting the expected results.

Some checking revealed not all of this is true. Still long and short form work in PHP for type declarations:

    > $f = function (int $i): bool {};
    > $f = function (integer $i): boolean {};

Also, `is_integr()` still exists, even if it's not in the documentatoin anymore:

    > var_dump(is_int(1));
    bool(true)
    > var_dump(is_integer(1));
    bool(true)

The `settype()` function accepts both as well:
    
    > $var='1'; settype($var, 'int'); var_dump($var);
    int(1)
    > $var='1'; settype($var, 'integer'); var_dump($var);
    int(1)
    
And last but not least there is `gettype()`:

    > var_dump(gettype(1));
    string(7) "integer"

That returns the long form.

kitsunet:
Problem identified, @bwaidelich is preparing something.

PR #1111, opened 2017-11-09
BUGFIX: No longer pick up too much YAML in migration

kdambekalns:
The migration to adjust Role persistence reads YAML files to learn
about existing roles. It picked up "too much" sometimes, which would
lead to errors, e.g. if some YAML could not be parsed but was not at
all relevant for the task.

This change makes the check on the path of found files stricter to
avoid that.

### DIFF
--- a/TYPO3.Flow/Migrations/Mysql/Version20130319131400.php
+++ b/TYPO3.Flow/Migrations/Mysql/Version20130319131400.php
@@ -167,7 +167,7 @@ class Version20130319131400 extends AbstractMigration
 
         $yamlSource = new \TYPO3\Flow\Configuration\Source\YamlSource();
         foreach ($configurationPathsAndFilenames as $pathAndFilename) {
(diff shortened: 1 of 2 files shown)


bwaidelich:
@kdambekalns Sorry for being thick, but I don't understand the changed RegEx. Could you quickly elaborate what "too much" meant and why adding `Policy` to `Development|Production`  fixes this?

kdambekalns:
First of all, only `Policy.yaml` is of interest here, as the code needs to know roles, nothing else. But what the change does is remove that from the regex. Now the regex still reads "all files", but only in a folder `Configuration` directly in a package folder (`%Packages/[^/]+/([^/]+)/Configuration/`). Later a check for `roles` is done anway. It would be even better, though, to only load `Policy.yaml`…

And the error I received was about some invalid YAML file that was in fact located in a test or documentation folder (I don't fully remember)…

albe:
wait, the Migration is older, doesn't that need to go into 3.3 then?

bwaidelich:
@kdambekalns what about @albe's comment? apart from that fine to be merged IMO

kdambekalns:
Erm, well. So, probably yes. Feel free to do, if you find the time!

PR #1112, opened 2017-11-09
BUGFIX: Allow TypeConverters to use the short name for bool and int

bwaidelich:
This re-applies the original fix (#1083) with some adjustments
in order to prevent regressions:

* The `StringConverter` is no longer incorrectly adjusted
* The `PropertyMapper` now normalizes source and target types
  to build the "typeConverterMap"

As a result, TypeConverters can now specify short and long version
in the source/target type definition.

Background:

The original fix was mainly an adjustment to `ScalarTypeToObjectConverter`
to normalize the constructor param type in order to allow simple types
to be mapped to classes with a single `bool` or `int` constructor argument.
But it also adjusted the `StringConverter` breaking the Neos export (see
neos/neos-development-collection#1732).

Credits to @monofone for the original patch!

Related: #1083
Related: #1109

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/ScalarTypeToObjectConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/ScalarTypeToObjectConverter.php
@@ -15,6 +15,7 @@ use Doctrine\ORM\Mapping\Entity;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Property\PropertyMappingConfigurationInterface;
 use Neos\Flow\Reflection\ReflectionService;
+use Neos\Utility\TypeHandling;
 
 /**
  * A type converter which converts a scalar type (string, boolean, float or integer) to an object by instantiating
@@ -70,7 +71,7 @@ class ScalarTypeToObjectConverter extends AbstractTypeConverter
             return false;
         }
         $methodParameter = array_shift($methodParameters);
-        return $methodParameter['type'] === gettype($source);
+        return TypeHandling::normalizeType($methodParameter['type']) === gettype($source);
     }
 
     /**
--- a/Neos.Flow/Classes/Property/PropertyMapper.php
+++ b/Neos.Flow/Classes/Property/PropertyMapper.php
@@ -229,8 +229,8 @@ class PropertyMapper
         $truncatedTargetType = TypeHandling::truncateElementType($normalizedTargetType);
         $converter = null;
 
-        $sourceTypes = $this->determineSourceTypes($source);
-        foreach ($sourceTypes as $sourceType) {
+        $normalizedSourceTypes = $this->determineSourceTypes($source);
+        foreach ($normalizedSourceTypes as $sourceType) {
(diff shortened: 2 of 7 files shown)


Neos.Flow/Classes/Property/TypeConverter/ScalarTypeToObjectConverter.php
```diff
@@ -70,7 +71,7 @@ public function canConvertFrom($source, $targetType)
             return false;
         }
         $methodParameter = array_shift($methodParameters);
-        return $methodParameter['type'] === gettype($source);
+        return TypeHandling::normalizeType($methodParameter['type']) === gettype($source);
```
albe: I think, the gettype still should also be normalized (possibly also at other similar places), mainly for the following reason:

http://php.net/manual/en/function.gettype.php
`for historical reasons "double" is returned in case of a float, and not simply "float"`

albe:
Looks good, but I think there is another buggy behaviour lurking, which could be fixed with this together. Ofc that can also be done separately though.

Neos.Flow/Classes/Property/TypeConverter/ScalarTypeToObjectConverter.php
bwaidelich: Woot, PHP never stops to surprise me ;)
Anyways, in that case we should probably fix `TypeHandling::normalizeType()` to be in sync with `gettype()`. No need for double conversion IMO

bwaidelich: And looking at it again, `ReflectionService::getMethodParameters()` should probably already normalize the parameter types so that we don't have to do the above in all cases

albe: But then we would again introduce breaky change in behaviour :/ IMO we shouldn't cater to "for historical reasons" crap and rather work around it

kitsunet: >And looking at it again, `ReflectionService::getMethodParameters()` should probably already normalize the parameter types so that we don't have to do the above in all cases

This is IMHO a dangerous change and we should think very carefully about it, although I agree in general. Especially we would need to check if let's say the following happens:

```
class OriginalClass {
    __construct(bool $foo)
```

now reflection service returns `boolean` which would probably end up somewhere in proxy building, generating a proxy like:

```
class OriginalClass {
    _construct(boolean $foo)
```

According to @kdambekalns this should still work syntactically. The question is, if this results in an error because the signature doesn't match or if PHP internally normalises this as well and comes to the conclusion that this is in fact the same.

Otherwise we have a problem and should provide a way to get the _raw_ type as hinted.

gerhard-boden:
@albe Are you okay with me merging this, so I can get the bugfix release out the door?

albe:
Yes sure. Go ahead :)

PR #1113, opened 2017-11-09
BUGFIX: Get current request for ``StandaloneView`` from ``RequestHandler``

gerhard-boden:
If no request is passed to a freshly created ``StandaloneView`` it is now retrieved from the  ``RequestHandler`` instead of the old method which called ``createFromEnvironment`` (now used as fallback if ``RequestHandler`` is not an instance of ``HttpRequestHandlerInterface``).

The old implementation can lead to faulty behaviour if the current Request is needed within the view, especially when running Flow behind proxies.

### DIFF
--- a/TYPO3.Fluid/Classes/TYPO3/Fluid/View/StandaloneView.php
+++ b/TYPO3.Fluid/Classes/TYPO3/Fluid/View/StandaloneView.php
@@ -20,6 +20,8 @@ use TYPO3\Flow\Mvc\Controller\ControllerContext;
 use TYPO3\Flow\Mvc\Routing\UriBuilder;
 use TYPO3\Flow\Utility\Files;
 use TYPO3\Fluid\View\Exception\InvalidTemplateResourceException;
+use TYPO3\Flow\Core\Bootstrap;
+use TYPO3\Flow\Http\HttpRequestHandlerInterface;
 
 /**
  * A standalone template view.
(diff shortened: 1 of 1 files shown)


gerhard-boden:
I wanted to keep the fix as non-breaky as possible. Would it make sense to return a CLI Request? But that would not help if you want to create valid URIs for example. 
Throwing an exception if we can not get an HTTP request from the RequestHandler on the other hand would change the current behaviour.

gerhard-boden:
Another thing comes to mind: Did someone skim through the `Request::createFromEnvironment()` usages throughout the core? There might be other places that need fixing as well

albe:
Only other relevant usage I see is `AbstractExceptionHandler`.

I like this, as it is the fully b/c way to fix the issue for now. I'm open to go for a breaking solution that gives a good compromise of convenience (StandaloneView inside a HttpRequest should render URIs correctly) and being clear about other cases. I could imagine the `else` part getting dropped for 5.0 and then throwing an exception inside the `getRequest()` method, if it is not set, explaining that one needs to create and set a senseful Request object for this StandaloneView to be able to work in that case (there might be usecases where the Request is totally not needed, so I wouldn't error out earlier).

gerhard-boden:
do we add markers within the code so we remember which parts to deprecate for 5.0 as part of a breaking change?

albe:
Generally yes, especially if you're not sure when this will happen and if it will be remembered. In the current scenario just creating a PR with that change targeted for master already and labelling it `next major!!!` would also work out fine.

albe:
I'd merge this by this weekend if no one objects till then.

PR #1114, opened 2017-11-10
TASK: ``translateByShortHandString`` return ID key if empty

putzwasser:
`translateByShortHandString` return the source label / ID key if no translation was found.

### DIFF
--- a/Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
+++ b/Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
@@ -135,10 +135,11 @@ class TranslationHelper implements ProtectedContextAwareInterface
         $shortHandStringParts = explode(':', $shortHandString);
         if (count($shortHandStringParts) === 3) {
             list($package, $source, $id) = $shortHandStringParts;
-            return $this->createTranslationParameterToken($id)
+            $translation = $this->createTranslationParameterToken($id)
                 ->package($package)
                 ->source(str_replace('.', '/', $source))
                 ->translate();
+            return empty($translation) ? $id : $translation;
         }
 
         throw new \InvalidArgumentException(sprintf('The translation shorthand string "%s" has the wrong format', $shortHandString), 1436865829);


aertmann:
A test for this behavior should be added to https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Tests/Unit/I18n/EelHelper/TranslationHelperTest.php

aertmann:
Since this is a breaking change, it probably can't go into 4.3, but needs to wait until 5.0 coming beginning next year.

putzwasser:
@aertmann Actually I consider this a bug and not a breaking change.

The `@return` tag defines a string as return type of `translateByShortHandString` and `translate`. As of now the return type can be string _and_ null which is a behaviour that doesn't match the documentation and phpdoc definition, so a user can't really rely on it.

albe:
Is this really breaking? Its a protected method and the docblock already stated `@return string Translated label or source label / ID key` - looks more like a BUGFIX to me

aertmann:
the resulting behavior is different, before if a label wasn't translated it wouldn't output anything but with the change it will output the translation ID.. that's pretty breaking

this change doesn't just ensure a string is returned instead of null, to do that it should return an empty string instead of the translation ID.. such a change would be a bugfix

aertmann:
and yes the comment is incorrect as that's not how it behaves, so the comment should probably be adjusted to reflect the actual behavior..

I know that the Neos translation view helper returns the translation ID if no translation is found, however this is a custom implementation done there..

with all that being said, I personally would welcome returning the translation ID instead of an empty string just like the comment states as it makes it easier to notice/fix, however it's a change in behavior and could break existing solutions

aertmann:
another point is that this change would mean that the behavior depends on how the translation helper is used as it would only return the ID for missing translations if only the $id attribute was passed since that calls the `translateByShortHandString` method..

thus if this should be change in general, it would make sense to change the behavior in the `translate` method of `TranslationParameterToken`

bwaidelich (changes requested):
I agree to @aertmann - even though your change makes sense it would be a fundamental change in behavior that would be very hard to adjust to: If any existing implementation relies on the current behavior they would not break but display a different (maybe incorrect) result.

albe:
So, is there any scenario where this change, which is generally wanted, but was rejected for it's rather silent breakiness, can be included? I imagine following options, but maybe you have some more @bwaidelich @aertmann 

- very clearly document the changing behaviour and include it with the upcoming major
- add the changed behaviour behind a feature-flag switch, which is disabled by default. Then sometime in the future switch the default to true and ultimately maybe remove the feature flag again.

If nothing is deemed feasible, can we just close this then instead of keeping it open endlessly? Let's at least decide and move forward, no matter how hard the decision ;)

bwaidelich:
@albe Thanks for being constructive! :)
Both solutions (documentation/feature flag) have their drawbacks of course.

IMO we should start with an issue that describes exactly the problem that this PR is trying to solve. Maybe there are other solutions

Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
```diff
@@ -135,10 +135,11 @@ protected function translateByShortHandString($shortHandString)
         $shortHandStringParts = explode(':', $shortHandString);
         if (count($shortHandStringParts) === 3) {
             list($package, $source, $id) = $shortHandStringParts;
-            return $this->createTranslationParameterToken($id)
+            $translation = $this->createTranslationParameterToken($id)
                 ->package($package)
                 ->source(str_replace('.', '/', $source))
                 ->translate();
+            return empty($translation) ? $id : $translation;
```
kdambekalns:
```suggestion
            return $translation ?? $id;
```

kdambekalns (changes requested):
A tiny suggestion – and the test as requested by @aertmann in 2017 should be added.

albe:
@bwaidelich There you go: #1559

sorenmalling:
@kdambekalns Can this go into next major? It's mentioned as something that should go into a major

kdambekalns:
> Can this go into next major?

Sure, it targets master already. But skimming the comments at least some test is missing – anything else?

albe:
Nothing from my side, though @bwaidelich was aiming for a potential alternative solution.

Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
albe: empty is not equal to `??` null coalescing, as it will also capture empty strings:
https://3v4l.org/bnFlZ
I guess an empty string can be considered a missing translation?

PR #1115, opened 2017-11-15
BUGFIX: Allow using a unix socket for mysql connection

liwo:
When specifying a unix socket for mysql connection, no host parameter must be given. As the code enforces a default value for host connection parameter, this simply adds a check to skip this if a unix socket is configured.


kitsunet:
Sounds sensible to me... @kdambekalns ? We might probalby just want to remove this fallback because it should be in our default settings anyway.

albe:
Let's play it safe and remove the fallback with next major and keep this fix until then. Wdyt?

kdambekalns:
This should go against 3.3

liwo:
Sorry for the wrong branch. I just recreated the PR against 3.3 in #1118.

albe:
Thanks! Would you be up to create another PR targeting master that just removes the fallback? We can then merge that once we have upcoming LTS branched off

PR #1117, opened 2017-11-15
BUGFIX: Handle non-string input in StringHelper

### LINKED ISSUE #1116: Eel StringHelper breaks on null values
### Description

In Eel, using a null value as input to the `StringHelper` always just worked. Since the introduction of type declarations in 4.2 it throws an error.

### Steps to Reproduce

1. Use e.g. `String.crop(null, 10, '…')`
2. Enjoy!

#### Expected behavior

The method should return an empty string.

#### Actual behavior

`Argument 1 passed to Neos\Utility\Unicode\Functions::strlen() must be of the type string, null given`

### Affected Versions

Neos: 3.2
Flow: 4.2

kdambekalns:
With the introduction of literal type declarations in the unicode
functions class in Flow the behaviour of the StringHelper changed in
subtle ways. Previously, when a null value was used as input for e.g.
the crop function, it would silently cast to string and return an
empty string. Now an error is raised, breaking e.g. the use of node
properties that are lacking an empty string as default value (which
probably is true for 99% of string properties on nodes.)

This changes makes the StringHelper be tolerant towards non-string
input by adding explicit typecasts. Even in places where the implicit
casts are still in effect, simply to be very clear on the intention.

Fixes #1116

### DIFF
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -11,9 +11,9 @@ namespace Neos\Eel\Helper;
  * source code.
  */
 
-use Neos\Flow\Annotations as Flow;
 use Neos\Eel\EvaluationException;
 use Neos\Eel\ProtectedContextAwareInterface;
+use Neos\Flow\Annotations as Flow;
 use Neos\Utility\Unicode\Functions as UnicodeFunctions;
 use Neos\Utility\Unicode\TextIterator;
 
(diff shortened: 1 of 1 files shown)


gerhard-boden:
tests failed again with that strange composer 404 error (this time for different machines). Restarting again
 🤞

gerhard-boden:
I think we already had the typehint in combination with ``null``  problem lately. It went something like this:
``null`` is not accepted for ``string $string`` unless it's ``string $string = null`` or ``?string`` in PHP 7.1

albe:
Right!

PR #1118, opened 2017-11-16
BUGFIX: Allow using a unix socket for mysql connection

liwo:
When specifying a unix socket for mysql connection, no host parameter must be given. As the code enforces a default value for host connection parameter, this simply adds a check to skip this if a unix socket is configured.


albe:
Just stripped the following from the PR message to keep the generated changelog clean:

> This is the same change as #1115 but against 3.3 rather than 4.0.

Issue #1120, opened 2017-11-16
Prepare Routing for multi-domains / add extension points


bwaidelich:
Extend the Router as follows:

## 1. Extend RouterInterface

As a first step we should change the `RouterInterface` from:
```php
interface RouterInterface
{
    public function route(Request $httpRequest);

    public function resolve(array $routeValues): string
}
```
to sth like:
```php
interface RouterInterface
{
    public function route(RouteContext $context): RouteResult

    public function resolve(ResolveContext $context): ResolveResult
}
```
* `RouteContext` contains some optional `routeValues` in addition to the current
* `Http\Request` (so that we can pre-set route values before the routing kicks in).
* `RouteResult` wraps the resolved `routeValues` and/or contains a flag whether a route matched at all.
* `ResolveContext` contains the `routeValues` in addition to the current `Http\Request` and (in v2) some flags like `forceAbsoluteUrl`, `forceScheme`
* `ResolveResult` contains the resolved `Http\Uri` and/or a flag whether a route resolved

## 2. Adjust Router Cache

The cache identifiers must be built from the `RouteContext` / `ResolveContext` (i.e. they should probably implement the `CacheAwareInterface`)

## 3. Allow hooking into RouteContext

I.e. it should be possible to write a `Http\Component` that can set `routeValues` for the current `RouteContext` (example: extract language from the `Accept-Language` HTTP header and set a routeValue correspondingly)

## 4. Allow hooking into ResolveResult

Introduce a signal that allows 3rd party code to hook into the `ResolveResult` just before it is put into the Cache. It should receive the `ResolveContext` and the current `ResolveResult` and might return a new result (or modify the existing one somehow)

## To be discussed

* Do we need a b/c layer for the breaking Router(Interface) signature change?
* Are there better names for the new classes (see above)?

## Background:

This will allow 3rd party packages (and Neos) to hook into the routing process and is a first step for a fully fledged multi-domain routing in Flow as described in #614 

Neos: 3.3 LTS
Flow: 4.3 LTS

bwaidelich:
### Intermediate status

I managed to get a [first version](https://github.com/bwaidelich/flow-development-collection/commit/03ca79c15c2820e871d77dffc129d37f69bfd68e) running (without adjusted tests & caching) that allows for custom RoutePart implementations to react to/influence parts of the URI "outside" of their respective route part.

I mostly followed the steps outlined above, with the following deviations:

#### 2. Adjust Router Cache

[...]

PR #1126, opened 2017-11-28
!!! FEATURE: More extensible Routing

bwaidelich:
An overhauled Router implementation that allows for more flexible
routing.

### Routing Parameters

Routing `Parameters` can be defined globally (via HTTP component) in order
to allow custom RoutePart handler to react to influences that are outside of
the incoming URI path (example: The requested host name or scheme)

For a RoutePart handler to access the parameters they have to implement
the new `ParameterAwareRoutePartInterface`.
The `DynamicRoutePart` already implements the interface. For custom implementations
extending `DynamicRoutePart` the parameters will be accessible via `$this->parameters`.

### Extended URI matching

RoutePart handlers can now return an instance of `MatchResult` when mapping
incoming requests.
This allows the handler to specify *Tags*  to be associated with the route.

#### Example:

*Before:*

```php
protected function matchValue($value)
{
     // custom logic, returning false if the $value doesn't match
     $this->value = $matchedValue;
     return true;
}
```

*Now:*

```php
protected function matchValue($value)
{
     // custom logic, returning false if the $value doesn't match
     return new MatchResult($matchedValue, RouteTags::createFromTag('some-tag'));
}
```
*Note:* The RouteTags argument is optional
*Note:* For backwards compatibility the above still works as before

### Extended URI resolving

Similar to the above, RoutePart handlers can now specify tags for *resolved* routes
by returning an instance of `ResolveResult`.
This also allows RouteParts to specify `UriConstraints` that will affect the resolved
URI.
UriConstraints allow to pre-set the following attributes of the resulting URI:

* Scheme  (for example "https")
* Host (for example "www.somedomain.tld")
* Host prefix (for example "en.")
* Host suffix (for example "co.uk")
* Port (for example 443)
* Path (for example "some/path")
* Path prefix (for example "en/")
* Path suffix (for example ".html")

#### Example:

*Before:*

```php
protected function resolveValue($value)
{
     // custom logic, returning false if the $value doesn't resolved
     $this->value = $resolvedPathSegment;
     return true;
}
```

*Now:*

```php
protected function resolveValue($value)
{
     // custom logic, returning false if the $value doesn't resolve
     return new ResolveResult($resolvedPathSegment, UriConstraints::create()->withHost('some-domain.tld'), RouteTags::createFromTag('some-tag'));
}
```

*Note:* The UriConstraints and RouteTags arguments are optional
*Note:* For backwards compatibility the above still works as before

## Breaking Change

This isn't a breaking feature in the strict sense because it doesn't affect the public API.
But it introduces some profound changes that potentially requires 3rd party code to be
adjusted:

#### UriBuilder return type

`UriBuilder::build()` (and thus also `UriBuilder::uriFor()`) no longer returns a `string` but instead
an instance of `UriInterface`.
In most places this won't be a problem, because it can be casted to a string but sometimes you
need to manually convert the result: `$uri = (string)$uri;`

#### RouterInterface signature

In case you implemented a custom router:
Congratulations for being brave!
You'll need to slightly adjust the code due to the modified signature of the two methods `route`
and `resolve`:

*Previously:*
```php
/**
 * @return array|null
 **/
public function route(Request $httpRequest);

/**
 * @return string
 **/
public function resolve(array $routeValues);
```

*Now:*
```php
public function route(RouteContext $routeContext): array;

public function resolve(ResolveContext $resolveContext): UriInterface;
```

#### RouterCachingService

Some public methods of the `RouterCachingService` have been changed and you might need
to adjust code that accesses those (directly or via AOP).

Resolves: #1120


Neos.Flow/Classes/Mvc/Routing/Dto/Parameters.php
```diff
@@ -0,0 +1,79 @@
...
+    public function withParameter(string $parameterName, $parameterValue): self
+    {
+        if (!TypeHandling::isLiteral(gettype($parameterValue)) && (!$parameterValue instanceof CacheAwareInterface)) {
+            throw new \InvalidArgumentException(sprintf('Parameter values must be literal types or implement the CacheAwareInterface, given: "%s"', is_object($parameterValue) ? get_class($parameterValue) : gettype($parameterValue)), 1511194273);
+        }
+        $newParameters = $this->parameters;
+        $newParameters[$parameterName] = $parameterValue;
+        return new static($newParameters);
+    }
+
```
albe: Docblock

Neos.Flow/Classes/Mvc/Routing/Route.php
```diff
@@ -396,12 +437,13 @@ public function matches(Request $httpRequest)
      * protocol and host).
      *
      * @param array $routeValues An array containing key/value pairs to be resolved to uri segments
-     * @return boolean TRUE if this Route corresponds to the given $routeValues, otherwise FALSE
+     * @return bool An instance of ResolvesResult if this Route corresponds to the given $routeValues, otherwise FALSE
```
albe: bool|ResolveResult

albe:
> **Yes, we use this in Neos**

Ah ok, so from Flow POV it is the applications duty to flush routing caches to the needs.

Maybe this can be somehow improved by setting sane default tags for the standard routePartHandlers, I could imagine the host parts to be good candidates for flushing decisions.

Neos.Flow/Classes/Mvc/Routing/Dto/Parameters.php
```diff
@@ -0,0 +1,79 @@
...
+    private $parameters = [];
+
+    /**
+     * @param string[] $parameters
+     */
+    private function __construct(array $parameters)
+    {
+        $this->parameters = $parameters;
+    }
+
```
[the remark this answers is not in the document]
bwaidelich: done

Neos.Flow/Classes/Mvc/Routing/Dto/Parameters.php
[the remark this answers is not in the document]
bwaidelich: done

Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
```diff
@@ -0,0 +1,199 @@
...
+    {
+        $newConstraints = $this->constraints;
+        $newConstraints[self::CONSTRAINT_PATH] = $path;
+        return new static($newConstraints);
+    }
+
+    public function withPathPrefix(string $pathPrefix): self
+    {
+        if ($pathPrefix === '') {
+            return $this;
```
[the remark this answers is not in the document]
bwaidelich: Good catch because `withPathPrefix` always *adds* the prefix and won't replace existing ones. But I don't think it would be necessary to *remove* any previously added prefixes.
The way this works currently is that the (Noes Frontend) Route Part adds a domain specific prefix like `de/` and afterwards the Router prefixes `index.php/` if URL rewriting is not available. I could imagine other Route Parts adding more prefixes but they should not unset/overwrite previous ones as that would be highly confusing.. Anyways, if the need arises we can always add this

bwaidelich:
FYI: I removed the `WIP` prefix. This still needs documentation (and I'll work on that asap) but that _could_ be done in a separate PR if this was merged first

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * @Flow\Proxy(false)
+ */
+final class RouteTags
+{
+
+    /**
+     * Pattern a tag must match. @see \Neos\Cache\Frontend\FrontendInterface::PATTERN_TAG
+     */
+    const PATTERN_TAG = '/^[a-zA-Z0-9_%\-&]{1,250}$/';
```
dfeyer: Why duplicated this ? Why not a small object responsible to validate a tag ?

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
```diff
@@ -0,0 +1,145 @@
...
+        return new static([]);
+    }
+
+    /**
+     * Creates an instance with one given tag
+     *
+     * @param string $tag Tag value satisfying the PATTERN_TAG regex
+     * @return RouteTags
+     */
+    public static function createFromTag(string $tag): self
```
dfeyer: createFromString ? Sounds strange Tag::createFromTag ?

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
```diff
@@ -0,0 +1,145 @@
...
+        return new static($newTags);
+    }
+
+    /**
+     * Checks the format of a given $tag string and throws an exception if it does not conform to the PATTERN_TAG regex
+     *
+     * @param string $tag
+     * @throws \InvalidArgumentException
+     */
+    private static function validateTag($tag)
```
dfeyer: RouteTags look like a collection, why not creating RouteTag object to encapsulate the validation logic ?

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Creates a new instance with the given $tag added
+     * If the $tag has been added already, this instance is returned
+     *
+     * @param string $tag
+     * @return RouteTags
+     */
+    public function withTag(string $tag): self
```
dfeyer: RouteTags::add can be more in sync with method like RouteTags::has

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
kdambekalns: Indeed, `createFromString()` seems way better.

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
kdambekalns: Well… It returns `self`, for `add()` I'd expect a `bool` or `void` return value. It's in line with e.g. how PSR-7 handles mutators.

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
kdambekalns: True. Validation should be done in a tag object, not in tags. Then again, no blocker for me, as it's private and can be changed any time later…

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
kdambekalns: Well. Tags may have different requirements, this one here could be less strict than the one from the cache. Then again, if the tags here are to be passed to caching, why not fully use the validation from there?

bwaidelich: Yes, we could just use the pattern from `\Neos\Cache\Frontend\FrontendInterface` but I wanted to avoid the coupling to another package..
We do store tags in the cache (by default) so they have to be compatible (as in: the pattern needs to be at least as strict). But that's not the only usecase. So unless we introduce a Tag ValueObject in the framework (which would be a good idea IMO) I'd rather duplicate that line of code

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
bwaidelich: A tag is a string in our case, so I think it is more explicit actually: Create a new tag collection from a single tag.
`createFromString` would mean: Create a tag collection from an arbitrary string wich would IMO suggest that you could do sth like `RouteTags::createFromString("tag1, "tag2")` which you can't right now.. But, true, I could change that

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
bwaidelich: I agree, that a tag would make a good Value Object and in fact in an earlier version I had that. But if we go that path we should introduce a generic `Tag` value object IMO, maybe in a package `neos/tags` and refer to that from all places where we currently use strings for tags (like in `neos/cache`).
I could kickstart that project to be included with the next Flow/Neos versions but as Karsten wrote it won't really change the API of this one

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
albe: I'm with Bastian here, `createFromString` kind of implies that it generates a collection with possibly multiple entries from a single (or multiple, but then I'd rather call it `createFromStrings`) string. The method is supposed to just convert a single Tag (guess we just really need a VO/TypeClass for that) into a collection.

bwaidelich:
FYI: I finally managed to write some documentation. I also marked the change breaking so that we don't forget to mention the changes in the release notes

bwaidelich:
Thanks a lot for the reviews and comments.. Merging this with agreement of @mficzel @skurfuerst @nezaniel. Let's keep our fingers crossed that this won't break anything

bwaidelich:
FYI: As discussed I'll push a follow up that reverts the changed `UriBuilder::uriFor()` return type change so that it returns a `string` again.
For the next major version we could re-consider but returning an instance of `UriInterface` lead to follow-up bugs

Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php
kitsunet: Reading this now I would actually say a Tag should probably be it's own object created from a string. :D

bwaidelich: @kitsunet yes, that's what I suggested with https://github.com/neos/flow-development-collection/pull/1126#discussion_r156880274 - I'll take care of kickstarting that to make up for todays fuckups ;)

PR #1128, opened 2017-11-30
FEATURE: Add PHPs json_encode options to stringify eel helper

daniellienert:
This feature adds the possibility to pass option constants to the Json.stringify Eel helper. This is needed for example if you want to have unescaped UTF8 characters in the output.

The usage looks like this:

    @process.jsonStringify = ${Json.stringify(value, ['JSON_UNESCAPED_UNICODE', 'JSON_FORCE_OBJECT'])}

It may look a bit weird on the first look to pass PHP options to a Javascript-like function. But its a PHP method in the end that does the encoding, so the PHP constants are the right options here.

### DIFF
--- a/Neos.Eel/Classes/Helper/JsonHelper.php
+++ b/Neos.Eel/Classes/Helper/JsonHelper.php
@@ -24,6 +24,10 @@ class JsonHelper implements ProtectedContextAwareInterface
     /**
      * JSON encode the given value
      *
+     * Usage example for options:
+     *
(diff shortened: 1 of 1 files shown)


albe:
I'm undecided and tbh. have too little Eel experience. How would the usage look effectively?

daniellienert:
The usage look like this:

`@process.jsonStringify = ${Json.stringify(value, ['JSON_UNESCAPED_UNICODE'])}`

albe:
Thanks :) I guess this opens up a can of possible typo worms, which will only emit warnings and lead to unexpected behaviour otherwise. I'm really undecided and let the others make the decision.

PR #1130, opened 2017-12-04
TASK: Add some documentation to array:indexOf Eel helper

daniellienert:


Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -143,12 +143,15 @@ public function last(array $array)
     }
 
     /**
-     * @param array $array
-     * @param mixed $searchElement
-     * @param integer $fromIndex
-     * @return mixed
+     * Returns the index of an array value, starting at 0.
+     * Returns -1 if the value was not found.
```
aertmann: not exactly, it returns the index of a an array value matching the supplied search element..

Maybe just copy the description from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf
"Returns the first index at which a given element can be found in the array, or -1 if it is not present"

daniellienert:
@aertmann: Any further objections?

daniellienert:
@kdambekalns - Thought about that. But I think it is an improvement not a bugfix, as the old documentation wasn't wrong at that point :)

PR #1132, opened 2017-12-06
BUGFIX: Check for valid session identifier before accessing the cache

Torsten85:
Currently it is possible to change the cookie value of `Neos_Flow_Session` (`TYPO3_Flow_Session`) to an invalid cache identifier. This leads to an `InvalidArgumentException` and an error code 500. 
This pull requests checks the validity of the session identifier before attempting to access the `metaDataCache`.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Session/Session.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Session/Session.php
@@ -320,7 +320,11 @@ class Session implements SessionInterface
         if ($this->sessionCookie === null || $this->request === null || $this->started === true) {
             return false;
         }
-        $sessionMetaData = $this->metaDataCache->get($this->sessionCookie->getValue());
+        $sessionIdentifier = $this->sessionCookie->getValue();
+        if ($this->metaDataCache->isValidEntryIdentifier($sessionIdentifier) === false) {
+            return false;
+        }
+        $sessionMetaData = $this->metaDataCache->get($sessionIdentifier);
         if ($sessionMetaData === false) {
             return false;
         }


albe:
Makes sense. I'd more consider this a BUGFIX though, in which case it would need to be targeted to 3.3.

aertmann:
@Torsten85: seems there's conflicts with the change, make sure your own repository is synced to upstream..

you  can either re-target this pull request to 3.3 instead of master or create a new one..

thank you

kitsunet:
The question is though, woudln't you want an exception on an invalid session identifier?

Torsten85:
@aertmann Thanks. If re-targeted this pr to 3.3 and renamed it to indicate that this is a bugfix.

@kitsunet I think in most cases it makes sense to throw an exception. But the session identifier can be changed by everyone directly in the browser... It shouldn't be possible for everyone to trigger an exception (and a 500 error) by simply changing a cookie value.

aertmann:
in these cases we sometimes resort to logging the exception but not throwing it..

albe:
Yeah, we could inject the security logger and log a warning there.

sorenmalling:
@Torsten85 The loggerInterface is already injected. So we can use this for logging a error in the general application log with a error state.

@Torsten85 Will you push a new change?

bwaidelich:
@sorenmalling You added this PR to the 6.0/7.0 project board but we decided only to work with issues there (in order to keep it easier to maintain and to have a nice changelog that links to the higher level issue descriptions)

albe:
Plus Bugfixes can probably go without fixed planning for a specific release, since most of the time they target a lower version anyway and can be released anytime. Unless of course they are bugfixes that target a bug that was introduced into master lately and needs to land before the release.

albe:
Closing for #2411

PR #1137, opened 2017-12-14
TASK: Add PostgreSQL JSONB query support for DQL

dfeyer:


dfeyer:
> and if all the names need to be that horrible.

Postgres create this fancy operator, but dont named them ;) so look it's common to name them, like we writte them ;) 

> Also, how does this behave on PostgreSQL 9.4?

https://www.postgresql.org/docs/9.4/static/functions-json.html all operator supported by version 9.4

About the documentation ... hum, it's pure DQL stuff ... but yes at least this PR can have a message ;)

dfeyer:
This PR use the new operators 
https://github.com/neos/neos-development-collection/pull/1801

PR #1138, opened 2017-12-14
BUGFIX: Cast objects to string in HTTP headers

bwaidelich:
This adjusts `Headers::set()` so that it accepts arbitrary
objects and casts them to string when rendering the header.

Previously all values where casted to an `array` (unless
they are an instance of `\DateTimeInterface`) leading to
PHP errors when passing in other objects.

### DIFF
--- a/Neos.Flow/Classes/Http/Headers.php
+++ b/Neos.Flow/Classes/Http/Headers.php
@@ -110,8 +110,8 @@ class Headers
             $date = clone $values;
             $date->setTimezone(new \DateTimeZone('GMT'));
             $values = [$date->format('D, d M Y H:i:s') . ' GMT'];
-        } else {
-            $values = (array) $values;
+        } elseif (!is_array($values)) {
+            $values = [$values];
         }
 
         switch ($name) {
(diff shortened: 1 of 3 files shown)


bwaidelich:
Note: This fix is targeted to 4.2 because the issue has probably been introduced with cedb834137c51c4ed682e9fa71f4f62d876c117d (which is not part of older releases)

kitsunet:
Why would you set an object as header though? 

I am fine to add it ofc, but isn't this exactly too much magic exception?

albe:
Generally makes sense, but...

> targeted to 4.2 because the issue has probably been introduced with cedb834

Are you sure about this, because the `set(..)` method hasn't been touched there and I can find the `$values = (array) $values;` code back until 3.3?

bwaidelich:
> Why would you set an object as header though?

For example to avoid fixes like this one:

https://github.com/neos/flow-development-collection/pull/1126/commits/0ae3283b18cd60f1faf31d6d4bbf237e5fa459b7

> isn't this exactly too much magic exception?

I think it's less magic than arbitrarily casting the object to an array.
OTOH the [PSR-7 interface](http://www.php-fig.org/psr/psr-7/) only allows `string|string[]` for the header value.. But since we allow DateTimes.. not sure

> Are you sure about this, because the set(..) method hasn't been touched

No, not sure and you're right Flow 3.3 is probably affected, too

bwaidelich:
What do you guys think about changing
```php
@param array|string|\DateTime
```
to 
```php
@param string|string[]|\DateTime
```
in `master` and throwing an exception for other values to make this easier to debug?

kitsunet:
I like!

bwaidelich:
Closing in favor of #1140

PR #1139, opened 2017-12-15
BUGFIX: Logger::logThrowable will return a message

dfeyer:


kitsunet:
You could probably also just really return the message from the storage.

kitsunet:
Problem is that only the storage knows how to find the stored item. So it must give some info back for the logger.

dfeyer:
@kitsunet Ok, Interface updated the Storage now return the message, make sense

dfeyer:
@kdambekalns Done

kdambekalns:
BTW… why is this in Flow, not in Flow.Log?

kitsunet:
Because for now I wanted to have storing of throwables separated from logging. It's a different concern (and my evil secret plan to replace logging with monolog at some point would be much harder that way, because we want to keep storing exception details I guess :D)

PR #1140, opened 2017-12-15
BUGFIX: Throw exception when passing objects to `Headers::set()`

bwaidelich:
According to PSR-7 (http://www.php-fig.org/psr/psr-7/) HTTP headers
can only accept values of `string` or `string[]`.

This adjusts our `Headers` implementation to throw an exception
if an object is passed to the `set()` method in order to make
error handling easier.

Exceptions prove the rule so we still do allow instances of
`\DateTimeInterface` to be set as header value.


bwaidelich:
See #1138 for the original approach to this

bwaidelich:
**NOTE:** Right now `Headers::set()` won't even accept values of type `integer` (or any other simple type other than `string`). I'm not sure if that's maybe too strict?

kdambekalns:
With `\DateTimeInterface` I think it's good to accept it (because it must be "cast" to string in a specific way. Anything else can be cast to string without trouble by the "user of the API", so… fine with me.

Neos.Flow/Tests/Unit/Http/HeadersTest.php
```diff
@@ -471,4 +481,48 @@ public function getCacheControlDirectiveReturnsTheSpecifiedDirectiveValueIfPrese
...
+        $this->assertSame($expectedResult, $headers->getPreparedValues());
+    }
+
+    /**
+     * @test
+     */
+    public function getPreparedValuesRendersOneHeaderPerArrayItem()
+    {
+        $headers = new Headers();
+        $headers->set('X-Array', ['some', 'array', 123]);
```
kdambekalns: Well, the `123` violates the `string[]` declared in the docblock, and should be rejected as well?

bwaidelich: "Eagle Eye Karsten" ;)
You're right.. we need to check every array item for its type. mäh

Neos.Flow/Classes/Http/Headers.php
```diff
@@ -110,8 +110,10 @@ public function set($name, $values, $replaceExistingHeader = true)
             $date = clone $values;
             $date->setTimezone(new \DateTimeZone('GMT'));
             $values = [$date->format('D, d M Y H:i:s') . ' GMT'];
-        } else {
+        } elseif (is_string($values)) {
             $values = (array) $values;
```
albe: I have to say, I prefer the `$values = [$values];` version better, because it is more explicit. I really had to check if `(array) $string;` would actually end up in an array with the single string item, or for some php-ish reasons create an array of characters or something else bizarre.

Neos.Flow/Tests/Unit/Http/HeadersTest.php
albe: Could be done with an `elseif ($values !== array_filter( $values, 'is_string' ))` check after the other type checks.

bwaidelich:
@albe @kdambekalns Thanks a lot for your feedback and sorry for the long delay.. I just rebased this and incorporated your suggestions

bwaidelich:
And immediately I got bitten. By the `StandardsComplianceComponent` of all things, that sets the `Content-Length` to an integer..
I still think it would be a good idea to be strict here - at least starting with the next major

kitsunet:
I will revisit this for next release but we need to get there anyway.

albe:
Probably related to #1552 so depending on how we implement PSR-7 (with or without our own Headers implementation), this might be relevant or obsolete. Linking here to not lose track.

kitsunet:
I guess this can be closed now....

PR #1141, opened 2017-12-18
TASK: Revert changed return type of `UriBuilder`-resolved URIs

bwaidelich:
This partly reverts #1126 by casting the resolved URI so that
`UriBuilder::uriFor()` and `UriBuilder::build()` returns a
`string` again rather than an instance of `UriInterface`.

Background:
Even though `UriInterface` is mostly casted to a string implicitly,
this change was considered "too dangerous" for a minor release
because it breaks code that relies on the previous return type
(i.e. when serializing the resolved URI).

For the next major release we can re-consider this change.

Related: #1120


nezaniel:
Yep, looks good

PR #1142, opened 2017-12-18
TASK: Revert some accidental doc comment changes

bwaidelich:
This is a (cosmetic) follow-up to #1126 reverting some
doc comment changes that were introduced by accident.

Related: #1120


bwaidelich:
Sorry for this, the diff View in Github is not the best to discover these gremlins in a large PR...

PR #1143, opened 2017-12-18
TASK: Remove submodule file that should not exist

kdambekalns:
This was added during reverting some unhappy change and should not be
here at all.


kdambekalns:
See 7a896f08424c065569124d2b2035db180e46f77b

PR #1144, opened 2017-12-18
BUGFIX: Adjust order of constant initializing in bootstrap process

mficzel:
The definition of the constants is moved to the run() method in the bootstrap
to allow predefinition of constants after creating the bootstrap instance.
This is needed  wich since the autoloader is created in bootstrap  constructor.
The autoloader in turn is needed for the calculation of constants in the DevelopmentServerRouter.


mficzel:
Closed in favor of #1145

PR #1145, opened 2017-12-18
BUGFIX: Flow server init should not rely on Files utility

kitsunet:
Removes usage of Files utility in ``PhpDevelopmentServerRouter.php``
in favor of the same code to determine the root of the installation
which is used in the command line init ``flow.php``.


kitsunet:
Rerunning one jenkins job. There was a fun little 1sec overlap problem in some time thing.

PR #1149, opened 2017-12-19
BUGFIX: Tag routes with UUIDs of entities in route values

bwaidelich:
This is a follow up to #1126 fixing the automatic tagging of
routes in `RouterCachingService::storeResolvedUriConstraints()` that
no longer included the UUIDs of entities in route values.

Related: #1120


albe:
Right.

PR #1154, opened 2017-12-24
!!!TASK: Only manage packages of a `Neos` type

kitsunet:
Our package management should never touch third party packages ever
and should only have to deal with packages made for Flow and Neos.
Those packages are the only ones expected to have Configuration,
Resources and a Package file. Everything else should be handled as
black box.
This serves two purposes, one it shortens the bootstrap time by not
having as many packages to boot. Basically third party packages now
do not matter anymore in terms of boot performance.
Secondly it further reduces the scope of the package manager which
should ultimately be minimal.

This modifies the package manager to only care for packages which
have a `neos-*` package type. Everything else will not be seen anymore.

This is marked breaking just because it is a big change, technically
your packages should always be `neos-*` type if you wanted them handled
by Flow and Neos.


kitsunet:
The moving of objects configuation is a consequence of no longer booting library packages. The rest is code improvements along the way.

bwaidelich:
FYI: This is a breaking change (see #1277). Maybe we can "fix" it somehow. Otherwise we should at least adjust the documentation

sorenmalling:
Can I just throw in something :-) AFAIK, I'm able to enter my own "package type" in composer.json and have Flow install it in a separate folder

```
Neos:
  Flow:
    package:
      packagesPathByType:
        'myproject-plugin': 'Plugin'
```

These packages will still need the Flow object magic, even though it's not "named" as a `neos-*` package

kitsunet:
Nope, per definition they will not, should not, cannot. You should define them as `neos-myproject-plugin` then. Also that configuration is fairly useless now. It's mostly handled by composer and if you need this configuration for anything right now you should go over to mono repos because then you won't need it ;)

PR #1156, opened 2017-12-25
!!!TASK: Remove deactivating packages from the package manager

kitsunet:
Packages no longer have state, they are either available or not
available entirely depending on the fact if they are installed.
The API is changed accordingly to reflect this change.

The implemented ``PackageManager`` still supports the
``getActivePackages`` and ``isPackageActive`` methods for
backwards compatibility but both are deprecated and removed from
the interface.

This is breaking if you rely on the functionality and requires
future changes if you use the deprecated methods.

### DIFF
--- a/Neos.Flow/Classes/Package/PackageManager.php
+++ b/Neos.Flow/Classes/Package/PackageManager.php
@@ -769,6 +769,7 @@ class PackageManager implements PackageManagerInterface
 
         $packageStatesCode = "<?php\n" . $fileDescription . "\nreturn " . var_export($orderedPackageStates, true) . ';';
 
+        Files::createDirectoryRecursively(dirname($this->packageInformationCacheFilePath));
         $result = @file_put_contents($this->packageInformationCacheFilePath, $packageStatesCode);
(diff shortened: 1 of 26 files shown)


Neos.Flow/Classes/Package/PackageManager.php
```diff
@@ -518,7 +405,7 @@ public function freezePackage($packageKey)
             throw new \LogicException('Package freezing is only supported in Development context.', 1338810870);
         }
 
-        if (!$this->isPackageActive($packageKey)) {
+        if (!$this->isPackageAvailable($packageKey)) {
             throw new Exception\UnknownPackageException('Package "' . $packageKey . '" is not available or active.', 1331715956);
```
kdambekalns: _… or active._

kitsunet:
You are right, we can as well fully break it at this point :D I somehow though this might be nicer to upgrade for people. I guess I would still keep ignoring the "Inactive" folder though.

kitsunet:
@kdambekalns maybe re-evaluate the review? :)

PR #1157, opened 2017-12-26
BUGFIX: catch NoMatchingRouteException to fix the RedirectHandler

dimaip:
**What I did**
When no route is found the NoMatchingRouteException is now thrown, which breaks up the http component chain, so the RedirectHandlerComponent is never reached.
This change swallows the exception.

**How to verify it**

Install the redirect handler extension and make sure the redirects work.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/RoutingComponent.php
+++ b/Neos.Flow/Classes/Mvc/Routing/RoutingComponent.php
@@ -14,6 +14,7 @@ namespace Neos\Flow\Mvc\Routing;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Http\Component\ComponentContext;
 use Neos\Flow\Http\Component\ComponentInterface;
+use Neos\Flow\Mvc\Exception\NoMatchingRouteException;
 use Neos\Flow\Mvc\Routing\Dto\RouteParameters;
 use Neos\Flow\Mvc\Routing\Dto\RouteContext;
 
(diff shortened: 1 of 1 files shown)


kitsunet:
Shouldn't the redirect handler come BEFORE the routing component though?

mficzel:
@kitsunet it is a feature that redirects are handled after other routes. This really avoids redirects bering send for valid urls. That way the redirect component does not have to be perfect.

dimaip:
Yep, what Martin said, and also it won't slow down normal requests.

dimaip:
@kitsunet Another I problem, a redirect would still get an error logged:
```
NOTICE   : Router route(): No route matched the route path "rss-feed".
```

This is annoying. I analyze those kind of logs, and it feels wrong to see them for valid redirects...

kitsunet:
Technically we should disable notices from production context.

kitsunet:
(And you can do so yourself via settings I think. Look for Flows log settings)

dimaip:
I like to log all error levels up to debug, helps with analyzing production issues faster and it doesn't cost me much with graylog anyways.

dfeyer:
> This is annoying. I analyze those kind of logs, and it feels wrong to see them for valid redirects...

Annoying, but hard to avoid, basically the redirect module and the routing part are decoupled, so no way for the router to know that something will happens later in the request for this URL. The dark side of decoupling ;)

PR #1158, opened 2017-12-27
BUGFIX: routing:routepath command does not use routing DTOs

fritjofbohm:
Before this fix, calling `./flow routing:routepath /somePath` will result in something like

```
Argument 1 passed to Neos\Flow\Mvc\Routing\Route_Original::matches() must be an instance of Neos\Flow\Mvc\Routing\Dto\RouteContext, instance of Neos\Flow\Http\Request given, called in /flowRootPath/Data/Temporary/Development/SubContextFritjofBohm/Cache/Code/Flow_Object_Classes/Neos_Flow_Command_RoutingCommandController.php on line 179
```

With this fix, the command uses the right DTOs that were introduced with 4.3.


kitsunet:
merged as no-brainer.

PR #1159, opened 2017-12-29
TASK: Update standalone Fluid to 2.3.x

kitsunet:
Pulling in the latest improvements in Fluid.

One major improvement is the addition of an escaping toggle: 

```
{escaping off}
```

In a template, layout or partial. When toggled off it causes the template parser to skip escaping of all outputs from ViewHelpers, variable outputs etc.

For legacy reasons the feature is added with backwards compatibility for the original escaping modifier. Therefore, the following values are all equally possible and mean the same:

```
{escaping off}
{escapingEnabled off}
{escaping false}
{escaping=false}
{escaping=off}
{escaping = off}
{escapingEnabled = off}
{escapingEnabled = false}
{escapingEnabled off}
{escapingEnabled false}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,10 +21,13 @@
         "symfony/dom-crawler": "~2.8.0",
         "symfony/console": "~2.8",
         "neos/composer-plugin": "^2.0.0",
-        "typo3fluid/fluid": "~2.1.3",
+        "typo3fluid/fluid": "~2.3.4",
         "ext-mbstring": "*"
     },
     "replace": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -103,6 +106,7 @@
     "require-dev": {
         "mikey179/vfsstream": "~1.6",
         "phpunit/phpunit": "~6.0",
+        "neos/flow": "*",
```
kdambekalns: Where does that come from?!

PR #1160, opened 2017-12-29
!!!TASK: Require PHP 7.1

kitsunet:
This raises the PHP version requirement for framework packages to
PHP 7.1 to be able to use a bunch of useful features for us in the
future.

Eg. iterable pseudo type, nullable types, void returns and more.


kitsunet:
Travis configuration was updated to only test 7.1 and 7.2 now.

PR #1161, opened 2017-12-29
TASK: Update ramsey/uuid dependency version

kitsunet:


kitsunet:
Those are from the merger script. Seem to be forgotten by someone else...
Yes, basically because we can, to avoid having EOL dependencies at some point...

Neos.Flow/composer.json
```diff
@@ -28,7 +28,7 @@
 
         "psr/http-message": "~1.0.0",
 
-        "ramsey/uuid": "^3.0.0",
+        "ramsey/uuid": "~3.7.1",
```
albe: Is there a good reason to constrain this so thightly onto 3.7 all of a sudden?

kitsunet: Good point. Mmm, maybe it's fine after all...

kdambekalns:
Since `^3.0.0` means `>= 3.0.0, < 4.0.0` already, and composer always tries to use the most recent version, this does not _update_ the dependency per se.

It simply raises the lower limit, but since we do not _depend_ on anything in the newer versions, that can be left as is.

OTOH it would not constrain to  `>= 3.7.1, < 3.8.0` as suggested by @bwaidelich, so it would do no harm, really.

I'll close this anyway, since the other changes to the composer manifest seem arbitrary (`neos/flow` in `require-dev`?)

albe:
@kdambekalns Are you sure about your statement regarding `~3.7.1` to not constraint to the minor version? As of the composer docs https://getcomposer.org/doc/articles/versions.md#tilde-version-range-

> ~1.2 is equivalent to >=1.2 <2.0.0, while ~1.2.3 is equivalent to >=1.2.3 <1.3.0

PR #1162, opened 2017-12-31
FEATURE: Add PSR-6 and PSR-16 compatibility

kitsunet:
This implements a cache frontend that is compatible with PSR-6
specification ``http://www.php-fig.org/psr/psr-6``.

When you use that Frontend you should rely only on the PSR-6 interface
methods and disregard anything not defined in the interfaces as that
is subject to change on our side.

Additionally a separate PSR-16 compatible SimpleCache is implemented
with it's own factory as the interfaces are incompatible with our interfaces.


kitsunet:
composer.json is the result of the merger so there seemed to be stuff missing before...

kitsunet:
As this has some overlappings I will add PSR-16 support as well...

kitsunet:
Replace by #1168

PR #1163, opened 2018-01-01
FEATURE: Make ObjectManager compatible to PSR-11

kitsunet:
PSR-11 is the container interface and it is rather easy to make the
ObjectManager compatible with it, so here we are.


kitsunet:
The question is, do we want to add `has` to the ObjectManagerInterface and then let it extend the ContainerInterface? Would be even nicer that way but more breaking (if someone actually implemented ObjectManagerInterface that is).

dfeyer:
> I would also gladly accept a PR for that version.

Me too ;) the chances to have custom ObjectManager out there are smallish and even easy to solve

kitsunet:
Ok, merging this and then create a PR for that on top with breaking marker.

PR #1164, opened 2018-01-02
BUGFIX: Fix duplicate configuration keys in schema files

daniellienert:
The latest symfony yaml parser is much more strict. This PR fixes duplicate keys in schema files.


daniellienert:
Style CI error is unrelated to this change.

PR #1165, opened 2018-01-03
BUGFIX: Add doctrine mapping type only once

daniellienert:
If you instantiate an additional entityManager in order to connect
to a second database the mappingTypes are added again
which leads to an error.


kitsunet:
Maybe we should take the type registration out of there and do that in a separate step once.

daniellienert:
@kitsunet: Any Idea where it should be done? Would like to see that as a bugfix first before refactoring.

kitsunet:
Oh yeah, we can definitely merge this now 👍 I guess as we have a Factory that would be a great place. I have to read the full thing but I don't quite get why we have this class at all and don't do it in the entity manager factory.

PR #1167, opened 2018-01-04
TASK: Cache expressions as strings

kitsunet:
This might seem counter-intuitive at first but serves two purposes:

1) Incremental change to the expressions is possible now, before
a change in policy protected classes would usually lead to an exception
and required you to flush the full cache, this prevents the issue and
therefore makes developing much smoother.

2) Caches can be moved to redis or any other backend as it's only
strings now.

The second part also makes a further separation of EEL and Flow
possible in the future, eventually allowing EEL to be come a package
that could be used independently from Flow.


Neos.Eel/Configuration/Objects.yaml
```diff
@@ -21,6 +22,16 @@ Neos\Eel\CompilingEvaluator:
 
 Neos\Eel\EelEvaluatorInterface:
   className: Neos\Eel\CompilingEvaluator
+  # FIXME: This configuration shouldn't be necessary. Check in ObjectConfiguration.
```
kdambekalns: Well, the interface does not describe the constructor, could that be the reason?

Also, if doing this for the interface, the constructor would _have_ to be included there, otherwise this would not work if the implementation deviates. Which is may well do, unless the constructor is in the interface. Which it isn't.

You see? ;)

kitsunet: Right, very good point. I will move it to setter.

kitsunet:
Not a big fan of property injection anymore 😆 But let's go for it, for now.

PR #1168, opened 2018-01-06
FEATURE: Add PSR-6 and PSR-16 support to cache framework

kitsunet:
This implements a PSR-6 compatible cache pool ``http://www.php-fig.org/psr/psr-6`` and
a factory for those caches.

Additionally a separate PSR-16 compatible SimpleCache is implemented
with it's own factory as the interfaces are incompatible with our interfaces.

Important: Both new cache variants are not integrated into Flows cache management at all,
you need to take care of getting and flushing those caches, they are not flushed on
`./flow flow:cache:flush`.


composer.json
```diff
@@ -25,6 +27,9 @@
         "ext-mbstring": "*"
     },
     "replace": {
+        "typo3/eel": "self.version",
```
dfeyer: It is really required ?

composer.json
```diff
@@ -103,6 +108,7 @@
     "require-dev": {
         "mikey179/vfsstream": "~1.6",
         "phpunit/phpunit": "~6.0",
+        "neos/flow": "*",
```
dfeyer: It is really required ?

composer.json
kitsunet: Just a result of the composer manifest merger script that probably wasn't executed before by someone... We can and should check these but that's not the responsibility of this change.

composer.json
kitsunet: same as above.

Neos.Cache/Classes/Frontend/PsrFrontend.php
```diff
@@ -0,0 +1,176 @@
...
+        $value = ($this->useIgBinary === true) ? igbinary_unserialize($rawResult) : unserialize($rawResult);
+        return new PsrCacheItem($key, true, $value);
+    }
+
+    /**
+     * Returns a traversable set of cache items.
+     *
+     * @param string[] $keys
+     * @return array
+     * @throws PsrInvalidArgumentException
```
daniellienert: This method actually doesn't throw an exception. I know the interface tells otherwise, but the comment block seems wrong that way.

Neos.Cache/Classes/Frontend/PsrFrontend.php
```diff
@@ -0,0 +1,176 @@
...
+
+        return $this->remove($key);
+    }
+
+    /**
+     * Removes multiple items from the pool.
+     *
+     * @param string[] $keys
+     * @return bool
+     * @throws PsrInvalidArgumentException
```
daniellienert: No Exception is thrown actually.

Neos.Cache/Classes/Psr/SimpleCacheFrontend.php
```diff
@@ -0,0 +1,205 @@
...
+    {
+        $this->backend->flush();
+        return true;
+    }
+
+    /**
+     * @param iterable $keys
+     * @param mixed $default
+     * @return iterable
+     * @throws Exception\PsrInvalidArgumentException
```
daniellienert: No Exception is thrown actually.

Neos.Cache/Classes/Psr/SimpleCacheFrontend.php
```diff
@@ -0,0 +1,205 @@
...
+        }
+        return $result;
+    }
+
+    /**
+     * @param iterable $values
+     * @param integer $ttl
+     * @return bool
+     * @throws Exception
+     * @throws Exception\PsrInvalidArgumentException
```
daniellienert: No Exception is thrown actually.

Neos.Cache/Classes/Psr/SimpleCacheFrontend.php
```diff
@@ -0,0 +1,205 @@
...
+            $allSet = $this->set($key, $value, $ttl) && $allSet;
+        }
+
+        return $allSet;
+    }
+
+    /**
+     * @param iterable $keys
+     * @return bool
+     * @throws Exception\PsrInvalidArgumentException
```
daniellienert: No Exception is thrown actually.

Neos.Cache/Classes/Psr/SimpleCacheFrontend.php
```diff
@@ -0,0 +1,205 @@
...
+            $this->delete($key);
+        };
+
+        return true;
+    }
+
+    /**
+     * @param string $key
+     * @return bool
+     * @throws Exception\PsrInvalidArgumentException
```
daniellienert: No Exception is thrown by this method.

kitsunet:
I know the exception tags are not in line with how we do it. But all these methods CAN throw those exceptions because they call methods internally that do. I wanted to stay in line with the interface by adding those. But I can obviously remove them if you all disagree. @daniellienert

Neos.Cache/Classes/Exception/PsrInvalidArgumentException.php
```diff
@@ -0,0 +1,22 @@
...
+ * source code.
+ */
+
+use Psr\Cache\InvalidArgumentException as Psr6InvalidArgumentException;
+use Psr\SimpleCache\InvalidArgumentException as Psr16InvalidArgumentException;
+
+/**
+ * An invalid argument (usually an inacceptable cache key) was given to a PSR cache.
+ */
+class PsrInvalidArgumentException extends \InvalidArgumentException implements Psr6InvalidArgumentException, Psr16InvalidArgumentException
```
mficzel: Why did you choose PsrInvalidArgumentException as class name instead of InvalidArgumentException?

mficzel:
I would more like it much more to see all the psr classes in a special folderstructure. To help people to actually use it. Basically all the code is there so that would be just moving stuff around.
```
Classes
  Psr
    SimpleCache
       SimpleCache.php
       SimpleCacheException.php
    Cache
       CacheException.php
       CacheItem.php
       CachePool.php
```

kitsunet:
@mficzel probably a good idea. Will move stuff around!

mficzel:
I also miss some documentation that explains how i instantiate a PSR Cache or SimpleCache and configure the Backend via Settings. Offcourse it makes sense to finish the code first. 

I like this feature and look forward using it for working on standalone eel.

kitsunet:
Rebased and moved everything around. Please note that the `Neos\Cache\Exception` is a generic exception shared by all parts of the implementation, not only the PSR caches.

Neos.Cache/Classes/Psr/Cache/CachePool.php
```diff
@@ -0,0 +1,176 @@
...
+use Neos\Cache\Frontend\VariableFrontend;
+use Neos\Cache\Psr\InvalidArgumentException;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+
+/**
+ * A frontend that implements the CacheItemPoolInterface from the PSR-6 specification.
+ */
+class CachePool extends VariableFrontend implements CacheItemPoolInterface
+{
```
mficzel: Not sure why this extends the VariableFrontend other than sharing the cacheFactory. Imho it would be much clearer to have a separate CachePoolFactory as for the simpleCache.

Neos.Cache/composer.json
```diff
@@ -6,6 +6,8 @@
     "license": ["MIT"],
```
mficzel: Here a provide-section should be added
```
"provide": {
        "psr/cache-implementation": "1.0",
        "psr/simple-cache-implementation": "1.0"
    },
```

Neos.Cache/Classes/Psr/Cache/CachePool.php
kitsunet: I wanted to have this implement the Flow CacheInterface as well, so you can configure it like a frontend. So PSR-6  caches can be instantiated via CacheManager and Objects configuration.
Of course we can change that. Would like to hear more opinions.

Neos.Cache/composer.json
kitsunet: mmm, good point, will do.

Neos.Cache/Classes/Psr/Cache/CachePool.php
mficzel: Imho this sounds hacky. I would try to achieve that with two separate getPsr*Cache methods in the cacheManager. That way you can inject a PSR-cache via objects.yaml and still use the backendConfiguration from Caches.yaml. If another class still uses the another Cache-Frontend it can get this injected as well and we end up having two frontends that share a single backend.

Neos.Cache/Readme.md
```diff
@@ -0,0 +1,67 @@
...
+        'myCache', 
+        \Neos\Cache\Backend\SimpleFileBackend::class
+    );
+    
+The first argument given to either factory is a unique identifier for the specific cache instance.
+If you need different caches you should give them separate identifiers.
+
+## Documenatation
+
+The PSR-6 CachePool is a special implementation of a frontend as described in the documentation below.
```
mficzel: I think this line not correct any mode

Neos.Cache/composer.json
mficzel: The `provide` section is still missing

kitsunet:
@mficzel Updated composer.json (finally) and adjusted comments and documentation.

kitsunet:
@mficzel I think it's clear that this should be your responsibility. But sure I can add some documenation about that tomorrow. I don't want to hastily add something to the CacheManager without thinking it through. I might add a very slim way to flush those caches though and if possible a way to integrate that into CacheManger via Package class. I would then like to merge this tomorrow either way, just to be able to get on with this.

kitsunet:
Agreed 👍

PR #1171, opened 2018-01-09
!!! FEATURE: PSR-3 Logging

kitsunet:
This change accomplishes two things. On the one hand it
provides PSR-3 compatibility for the logger package.
On the other hand it lays the ground work to allow  any
PSR-3 compatible logger to be used in Flow and applications.

This is breaking in case you implemented the ``Neos\Flow\Log\LoggerInterface`` 
yourself, you should switch to the PSR-3 logger interface (should be easy).


kitsunet:
This is the first round with this. I just wanted to share it to have some more opinions. UnitTests obviously need to be added.

I tried to make this as non breaking as humanly possible. The idea is to keep the old logging interface(s) around until the major after this one to give people plenty of time to migrate as I think a code migration for this is too complicated and wouldn't cover all cases anyway.

In any default configuration case this should work without much difference to before.

Following cases are covered:

- Classic configuration of old style logger. This will be routed through a PSR-3 compatible logger now but end up about the same.
- Custom backend for our logging package: Will just work out of the box, as the new PSR-3 compatible logger feeds into anything that is BackendInterface compatible.

What very probably doesn't work well is a custom `\Neos\Flow\Log\LoggerInterface` itself, that cannot really made non breaking as we can't know what is happening in there. In this case people need to adapt to the PSR-3 LoggerInterface OR turn their Logger into a backend for our logging package.

What automatically works already in this version is using any PSR-3 compatible package as logger for anything in Flow and applications by defining a `loggerFactory` compatible with the respective, simple interface, from which Flow can retrieve PSR-3 loggers. Configuration for those loggers will reside under a key that matches the LoggerFactoryInterface implementation and implementations can be configured any way needed.

Next steps:
- Use the new PsrSystemLoggerInterface and PsrSecurityLoggerInterface instead of the deprecated SystemLoggerInterface and SecurityLoggerInterface everywhere in this collection to avoid the roundtrip through the old logger interface.
- Tests
- Example implementation of factory for Monolog
- cleanup (comments, api annotations, documentation)

Neos.Flow.Log/Classes/PsrLogger.php
```diff
@@ -0,0 +1,111 @@
...
+
+use Neos\Flow\Log\Backend\BackendInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerTrait;
+use Psr\Log\LogLevel;
+
+/**
+ *
+ */
+class PsrLogger implements LoggerInterface
```
daniellienert: Hey @kitsunet, wouldn't it possible to make this PSR compatible logger the default one. And if so could we remove the PSR in the naming?

kitsunet:
No way, it would clash with the Logger in Flow package and doing that would be really breaking. I don't think it makes much sense to do it that hard. @daniellienert

kitsunet:
No clue why travis can't checkout the branch currently :(

kdambekalns:
FYI: I cleared the PR build cache on Travis, maybe that helped. Either way, the tests "properly fail" now.

kdambekalns:
That last commit undoes quote a few updates to dependencies…

kitsunet:
YAY only behat now.

kitsunet:
Removed WIP label as the only thing missing is a monolog example implementation. But that would be separate from this PR anyway.

kitsunet:
Master branch of this package is now adapted to this change and provides the necessary Factory and configuration. Ideally install via composer  (using custom repository entry).

Neos.Flow.Log/Classes/DefaultLogger.php
```diff
@@ -99,14 +135,20 @@ public function addBackend(Backend\BackendInterface $backend)
...
+        if ($this->wasPsrLoggerInjected) {
+            throw new Exception('A PSR-3 logger was injected so removing backends is not possible. Create a new instance.', 1515343007859);
+        }
+
         if (!$this->backends->contains($backend)) {
             throw new NoSuchBackendException('Backend is unknown to this logger.', 1229430381);
         }
         $backend->close();
         $this->backends->detach($backend);
+        $this->psrLogger = null;
```
kdambekalns: When can this ever happen to be needed? Ah, `createDefaultPsrLogger()` sets a logger, but not the inject flag. Maybe add a comment here?

Neos.Flow.Log/Classes/DefaultLogger.php
```diff
@@ -123,17 +165,31 @@ public function removeBackend(Backend\BackendInterface $backend)
...
-        foreach ($this->backends as $backend) {
-            $backend->append($message, $severity, $additionalData, $packageKey, $className, $methodName);
+
+        $psrLogLevel = self::LOGLEVEL_MAPPING[$severity] ?? LOG_INFO;
+
+        if ($additionalData === null) {
+            $additionalData = [];
+        }
+
+        $additionalData = is_array($additionalData) ? $additionalData : ['additionalData' => $additionalData];
```
kdambekalns: So… if I pass an array having the key `className` and pass the `$className` argument, the latter will override my additional data entry. I'd say `$additionalData = ['additionalData' => $additionalData]` unconditionally.

Neos.Flow.Log/Classes/DefaultLogger.php
```diff
@@ -20,6 +21,27 @@
...
+        LOG_WARNING => LogLevel::WARNING,
+        LOG_ERR => LogLevel::ERROR,
+        LOG_CRIT => LogLevel::CRITICAL,
+        LOG_ALERT => LogLevel::ALERT
+    ];
+
+    /**
+     * @var \Psr\Log\LoggerInterface
+     */
+    protected $psrLogger;
```
kdambekalns: Applicable to every place this name is used: `psrLogger` or `PsrLogger` is a really bad name IMHO. It's a logger according to PSR-3, and it does not log PSRs, so it should be `psrThreeCompatibleLogger`, but that is rather long.

Not a blocker for me, but still a tad bit weird.

Neos.Flow.Log/Classes/PsrLogger.php
```diff
@@ -0,0 +1,114 @@
...
+use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerTrait;
+use Psr\Log\LogLevel;
+
+/**
+ * PSR-3 supporting logger.
+ *
+ * @api
+ */
+class PsrLogger implements LoggerInterface
```
kdambekalns: Said it earlier, but here it's even worse. This is not a logger for PSRs, but a PSR-3 compatibe logger.

Neos.Flow.Log/Classes/PsrLogger.php
```diff
@@ -0,0 +1,114 @@
...
+            if (!$backend instanceof BackendInterface) {
+                throw new \InvalidArgumentException(sprintf('The provided backend was not an instance of the "%s" interface', BackendInterface::class), 1515338089289);
+            }
+            $backend->open();
+            $this->backends[] = $backend;
+        }
+    }
+
+    /**
+     * @param mixed $level
```
kdambekalns: Is it really mixed? Weird…

Neos.Flow.Log/Classes/PsrLogger.php
```diff
@@ -0,0 +1,114 @@
...
+        foreach ($this->backends as $backend) {
+            $backend->append($message, $backendLogLevel, $additionalData, $packageKey, $className, $methodName);
+        }
+    }
+
+    /**
+     * @param array $context
+     * @return array list of packageKey, className and methodName either string or null
+     */
+    protected function extractlegacyDataFromContext(array $context): array
```
kdambekalns: Shpuld be `extractLegacyDataFromContext`

Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -114,12 +114,15 @@ public function injectReflectionService(ReflectionService $reflectionService)
     }
 
     /**
-     * @param SystemLoggerInterface $systemLogger
+     * Injects the (system) logger based on PSR-3.
+     *
+     * @param LoggerInterface $logger
      * @return void
+     * @Flow\Autowiring(false)
```
kdambekalns: Why is this done here, but not elsewhere?

Neos.Flow/Classes/Cli/SlaveRequestHandler.php
```diff
@@ -75,16 +76,16 @@ public function handleRequest()
...
-        $systemLogger->log('Running sub process loop.', LOG_DEBUG);
+        $systemLogger->log(LogLevel::DEBUG, 'Running sub process loop.');
         echo "\nREADY\n";
 
         try {
             while (true) {
                 $commandLine = trim(fgets(STDIN));
                 $trimmedCommandLine = trim($commandLine);
-                $systemLogger->log(sprintf('Received command "%s".', $trimmedCommandLine), LOG_INFO);
+                $systemLogger->log(LogLevel::INFO, sprintf('Received command "%s".', $trimmedCommandLine));
```
kdambekalns: Why not use `info()` like elsewhere?

Neos.Flow/Classes/Cli/SlaveRequestHandler.php
```diff
@@ -75,16 +76,16 @@ public function handleRequest()
         $sequence->invoke($this->bootstrap);
 
         $objectManager = $this->bootstrap->getObjectManager();
-        $systemLogger = $objectManager->get(SystemLoggerInterface::class);
+        $systemLogger = $objectManager->get(PsrLoggerFactoryInterface::class)->get('systemLogger');
```
kdambekalns: Why is `systemLogger` kept here, and not changed to `logger` like elsewhere?

Neos.Flow/Classes/Command/CoreCommandController.php
```diff
@@ -189,9 +190,13 @@ public function compileCommand($force = false)
 
         /** @var PhpFrontend $classesCache */
         $classesCache = $this->cacheManager->getCache('Flow_Object_Classes');
+        $systemLogger = $this->objectManager->get(PsrLoggerFactoryInterface::class)->get('systemLogger');
```
kdambekalns: `systemLogger` vs `logger` again…

Neos.Flow/Classes/Command/DoctrineCommandController.php
```diff
@@ -17,11 +17,13 @@
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Cli\CommandController;
 use Neos\Flow\Error\Debugger;
-use Neos\Flow\Log\SystemLoggerInterface;
+use Neos\Flow\Log\ThrowableStorageInterface;
```
kdambekalns: This is still weird. What has a storage to do with logging?

Neos.Flow/Classes/Command/DoctrineCommandController.php
```diff
@@ -64,6 +71,14 @@ public function injectSettings(array $settings)
         $this->settings = $settings['persistence'];
     }
 
+    /**
+     * @param LoggerInterface $systemLogger
+     */
+    public function injectcSystemLogger(LoggerInterface $systemLogger)
```
kdambekalns: The method name should not work (at least not for auto-wiring), since it is misspelt. And why not `logger` (again)?

Neos.Flow/Classes/Log/LoggerBackendConfigurationHelper.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $this->legacyConfiguration = $logConfiguration;
+    }
+
+    /**
+     * Normalize a backend configuration to a unified format.
+     *
+     * @return array
+     */
+    public function getNormalizedLegacyConfiguration()
```
kdambekalns: Methods should have return type annotations here and below.

Neos.Flow/Classes/Log/LoggerBackendConfigurationHelper.php
```diff
@@ -0,0 +1,81 @@
...
+            $backendObjectNames = $configuration['backend'];
+            $backendOptions = $configuration['backendOptions'] ?? [];
+            $normalizedConfiguration[$logIdentifier] = $this->mapLoggerConfiguration($backendObjectNames, $backendOptions);
+        }
+
+        return $normalizedConfiguration;
+    }
+
+    /**
+     * @param $backendObjectNames
```
kdambekalns: Parameter types and type hints missing

Neos.Flow/Classes/Log/LoggerBackendConfigurationHelper.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Map a backend object name and it's options into an array with defined keys.
+     *
+     * @param string $backendObjectName
+     * @param array $backendOptions
+     * @return array
+     */
+    protected function mapBackendInformation($backendObjectName, $backendOptions)
```
kdambekalns: Type hints missing

Neos.Flow/Classes/Log/LoggerFactory.php
```diff
@@ -77,12 +85,31 @@ public function __construct()
...
 
         return $this->logInstanceCache[$identifier];
     }
 
+    /**
+     * @param string $identifier
+     * @param string $loggerObjectName
+     * @return LoggerInterface
+     */
+    protected function createPsrBasedLogger($identifier, $loggerObjectName)
```
kdambekalns: (Return) type hints missing…

Neos.Flow/Classes/Log/LoggerFactory.php
```diff
@@ -106,6 +133,16 @@ protected function instantiateLogger($loggerObjectName, $backendObjectNames, arr
...
 
+        $logger = $this->injectAdditionalDependencies($logger);
+        return $logger;
+    }
+
+    /**
+     * @param LoggerInterface $logger
+     * @return LoggerInterface
+     */
+    protected function injectAdditionalDependencies(LoggerInterface $logger)
```
kdambekalns: Return type missing

Neos.Flow/Classes/Log/PsrLoggerFactory.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Get the logger configured with given identifier.
+     * This implementation treats the logger as singleton.
+     *
+     * @param string $identifier
+     * @return \Psr\Log\LoggerInterface
+     * @throws \Exception
+     */
+    public function get(string $identifier)
```
kdambekalns: Return type hint(s) missing here…

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -702,7 +723,7 @@ protected function getFlattenedValidationErrorMessage()
                 $logMessage .= 'Error for ' . $propertyPath . ':  ' . $error->render() . PHP_EOL;
             }
         }
-        $this->systemLogger->log($logMessage, LOG_ERR);
+        $this->logger->log(LogLevel::ERROR, $logMessage);
```
kdambekalns: Why not `logger->error()`?

kitsunet: It was easier to search replace to this format 😆

kitsunet:
@kdambekalns my main problem is the breakiness. I can remove a lot of cruft by removing adapter layers and stuff, but that makes it way more breaking. Right now it is relatively non breaking while still allowing pretty much full PSR-3 usage. I am happy to walk someone through it, but currently I feel  due to the way our logger works and PSR-3 works those layers and factories are necessary for the time being.

Neos.Flow.Log/Classes/DefaultLogger.php
kitsunet: Yeah, happy to have a better name, but as you said, the correct name is a bit unwieldy. Not that that should stop us... Idk, I can change that.

Neos.Flow.Log/Classes/DefaultLogger.php
kitsunet: Again part of backwards compatibility. I would agree to what you say, but that will mean some legacy places will probably not log the className anymore.

Neos.Flow.Log/Classes/PsrLogger.php
kitsunet: I think the interface declares it as mixed, that's where this comes from...

Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
kitsunet: I think because some issues with this being used in compiletime.

kitsunet:
@kdambekalns Comprehensive update posted... Documentation obviously is still needed, but I would like to wait until we are sure this is merge ready to not create stale documentation.

kitsunet:
and obviously I forgot something...

kitsunet:
Done now. This should fix the comments you left and follow our discussion yesterday. @kdambekalns

kitsunet:
Note: About 700 lines of code are deprecated until the major after this one. So at that point we are at about +300 lines of code to add the flexibility of PSR-3 deep in Flow (and a bunch of those lines are actually added configuration for using the factory via `Objects.yaml`)

kitsunet:
And resolved conflicts again...

kitsunet:
and rebased again...

Neos.Flow.Log/Tests/Unit/LoggerTest.php
```diff
@@ -28,7 +28,7 @@ public function logPassesItsArgumentsToTheBackendsAppendMethod()
         $mockBackend = $this->getMockBuilder(BackendInterface::class)->setMethods(['open', 'append', 'close'])->getMock();
         $mockBackend->expects($this->once())->method('append')->with('theMessage', 2, ['foo'], 'Foo', 'Bar', 'Baz');
 
-        $logger = new Logger();
+        $logger = new DefaultLogger();
```
dimaip: Why do you use deprecated class here instead of `Psr\Logger`? For the ease of migration?

dimaip:
Resolved conflicts.
Wow I really like throwing away parts of code in favor of 3rd party code.
But my brain doesn't have enough RAM to review 1.5K LOC PRs... Somebody teach me how you do that...

Neos.Flow.Log/Tests/Unit/LoggerTest.php
kitsunet: These are the logger unit tests, so IMHO they should just continue testing the original logger even if it's deprecated.

dimaip: Oh doh, sorry, didn't see it was a test

PR #1176, opened 2018-01-10
!!! TASK: Make ObjectManagerInterface a superset of PSR-11

kitsunet:
Makes the ``ObjectManagerInterface`` fully compatible with
the PSR-11 ``ContainerInterface``.

This is a breaking change in case you implemented
``ObjectManagerInterface`` because you would have to implement the
``has`` method now in your implementation.


kitsunet:
After discussion in: https://github.com/neos/flow-development-collection/pull/1163

@albe and @dfeyer maybe have a look :)

kitsunet:
Nope, no drawbacks AFAIK. We might want to look at how we can (in appropriate places) say we want the Containerinterface and get the object manager. But that can be done later.

PR #1177, opened 2018-01-11
BUGFIX: Pass signal information once, not once more for each slot

kdambekalns:
Signals can pass arguments to slots. These are amended with signal information,
if desired.

Because slots are hadled in a loop, and the passed signal arguments array was
used in that loop, the addition of the signal information was done mutliple
times, if requested.

This change makes sure that the signal information is only added once.


kdambekalns:
That was a nice one. :)

PR #1178, opened 2018-01-13
DOCUMENTATION: Removed dev composer option since it's deprecated

sebastiansommer:
The option --dev is deprecated. Dev packages are installed by default.


kitsunet:
Perfect, thanks!

PR #1179, opened 2018-01-14
TASK: Update symfony and doctrine dependencies

daniellienert:
This change updates the following packages to recent versions:

- doctrine/migrations: ~1.3.0 -> ~1.6.0
- doctrine/dbal: ~2.5.0 -> ~2.6.0
- symfony/yaml: ~2.8.0 -> ~4.0.0
- symfony/dom-crawler: ~2.8.0 -> ~4.0.0
- symfony/console: ~2.8 -> ~4.0.0


kitsunet:
The dependency changes were only done in the global composer.json but should have been don inside the packages and then merged unto the global one. The current state of this is broken if you would use the split packages of master. (and if anyone runs the merge script for the package composer.jsons because it deletes the version updates there).
I will push and merge a change that corrects this.

kdambekalns:
Gnah, I should have noticed. Sorry!

PR #1181, opened 2018-01-15
TASK: Allow debugging of "silent command failures"

kdambekalns:
Sometimes a subcommand execution fails without a proper error message. And
it may happen that even the PHP error is empty at this point.

The error may be "Argument list too long", but this cannot be debugged
easily.

This change dumps the command that was to be executed into a file, if the
length of the command exceeds a threshhold. One can then run it manually
and hopefully see a helpful error message.

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
@@ -634,6 +634,24 @@ class Scripts
                 $exceptionMessage = implode(PHP_EOL, $output);
             } else {
                 $exceptionMessage = sprintf('Execution of subprocess failed with exit code %d without any further output. (Please check your PHP error log for possible Fatal errors)', $result);
+
+                // if the command is too long, it'll just produce /usr/bin/php: Argument list too long but this will be invisible
+                // the length is a guesstimate, there is no easy/clear way to find the limit being applied.
+                if (strlen($command) > 10 * 1024) {
+                    $exceptionMessage .= ' Try to run the command manually, it may just be too long.';
+
+                    if (!file_exists(FLOW_PATH_DATA . 'Logs/Exceptions')) {
+                        mkdir(FLOW_PATH_DATA . 'Logs/Exceptions');
+                    }
+                    if (file_exists(FLOW_PATH_DATA . 'Logs/Exceptions') && is_dir(FLOW_PATH_DATA . 'Logs/Exceptions') && is_writable(FLOW_PATH_DATA . 'Logs/Exceptions')) {
+                        $referenceCode = date('YmdHis', $_SERVER['REQUEST_TIME']) . substr(md5(rand()), 0, 6);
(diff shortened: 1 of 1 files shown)


albe:
Good call, had the issue already that I couldn't properly debug the cause and needed to add a dump to find the actual command executed to manually run it. Anything speaking against always logging the full comand in the case it fails without output?

kdambekalns:
I think we could always log it, good idea.

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -634,6 +634,23 @@ public static function executeCommand($commandIdentifier, array $settings, $outp
...
             } else {
                 $exceptionMessage = sprintf('Execution of subprocess failed with exit code %d without any further output. (Please check your PHP error log for possible Fatal errors)', $result);
+
+                // If the command is too long, it'll just produce /usr/bin/php: Argument list too long but this will be invisible
+                // If anything else goes wrong, it may as well not produce any $output, but might do so when run on an interactive
+                // shell. Thus we dump the command next to the exception dumps.
+                $exceptionMessage .= ' Try to run the command manually, to hopefully get some hint on the actual error.';
+
+                if (!file_exists(FLOW_PATH_DATA . 'Logs/Exceptions')) {
+                    mkdir(FLOW_PATH_DATA . 'Logs/Exceptions');
```
daniellienert: We already use the Files Utility in this class. So we could use `Files::createDirectoryRecursively()`here. (Otherweise add a recursive flag)

TYPO3.Flow/Classes/TYPO3/Flow/Core/Booting/Scripts.php
```diff
@@ -634,6 +634,23 @@ public static function executeCommand($commandIdentifier, array $settings, $outp
...
+                }
+                if (file_exists(FLOW_PATH_DATA . 'Logs/Exceptions') && is_dir(FLOW_PATH_DATA . 'Logs/Exceptions') && is_writable(FLOW_PATH_DATA . 'Logs/Exceptions')) {
```
daniellienert: Shouldn't `(is_dir(FLOW_PATH_DATA . 'Logs/Exceptions') && is_writable(FLOW_PATH_DATA . 'Logs/Exceptions')` enough?

kdambekalns: I was "careful" and took that block from the exception handling code we already have. Not sure TBH about the intricacies of files vs. dirs vs. symlinks here…

kitsunet:
Why didn't you use the ThrowableStorage? Would be much less code duplication.

kdambekalns:
@kitsunet Because I wasn't aware of it. And it's not used in our code, at least not where I looked. Doh?

kdambekalns:
Ah, admittedly I developed this while debugging things on a Neos 3.0 instance…

kitsunet:
Yep, that makes sense ofc. Well, can be refactored, so not a big deal.

PR #1182, opened 2018-01-15
TASK: Fix code format in docs

DanielSiepmann:
- Add code-block syntax to format php code as php code.
- Fix indentation in code example 
- Indent using spaces instead of tabs.
- Fix position of curly braces.

### DIFF
--- a/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Validation.rst
+++ b/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Validation.rst
@@ -1,4 +1,4 @@
-﻿==========
+==========
 Validation
 ==========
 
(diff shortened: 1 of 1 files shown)


Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Validation.rst
```diff
@@ -80,16 +85,17 @@ itself possibly fail the validation and try to redirect to previous action, endi
 Furthermore, it is also possible to execute *additional validators* only for specific action
 arguments using ``@Flow\Validate`` inside a controller action::
 
-	class CommentController extends \Neos\Flow\Mvc\Controller\ActionController {
+    class CommentController extends \Neos\Flow\Mvc\Controller\ActionController {
```
daniellienert: Why did you not wrap this one and the following bigger blocks into a php code block?

DanielSiepmann: It's already a code block due to `::` at the end of the line before.

That this should be highlighted as PHP is defined globally as default at https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Documentation/conf.py#L314
```
# Use PHP syntax highlighting in code examples by default
highlight_language='php'
```
Which does not work at github, but using sphinx while rendering the docs.

DanielSiepmann: The mixture exists as I didn't wanted to add a `:` at the text when there was a `.` before.
`::` ads a `:` in rendered HTML.

Issue #1184, opened 2018-01-16
Resolved URIs must respect configured BaseUri


bwaidelich:
### Description

The "More extensible Routing" feature, added with #1126, introduced a regression that leads to the configured `baseUri` to be ignored when building URIs with the `UriBuilder`.

### Steps to Reproduce

1. configure a `Neos.Flow.http.baseUri` of `http://base-uri.localhost:8082`
2. Create an action that renders a link, for example using a Fluid ViewHelper: `<f:link.action action="someAction">Link</f:link.action>`
3. Render the action using *a different domain* than the one configured above, for example: `http://default.localhost:8081/your.package/someaction`

#### Expected behavior

The rendered link should be prefixed with the configured `baseUri`:
```
http://base-uri.localhost:8082/your.package/someaction
```

#### Actual behavior

The configured `baseUri` is ignored, and instead the *currently requested URI* is considered as prefix:
```
http://default.localhost:8081/your.package/someaction
```

### Affected Versions

Flow:

4.3 (4.3.0 - 4.3.2)

Related: #1120

PR #1185, opened 2018-01-16
BUGFIX: Respect configured baseUri when resolving URIs

bwaidelich:
This fixes a regression intdroduced with #1126 that prevented
the configured `baseUri` from being respected when building
URIs using the `UriBuilder`.

Background:

Instead of using the *Request URI* as base for generated URIs
this fix uses the *Request Base URI*.
Usually the result is the same, but not if a different `baseUri` is
configured from the one being requested.

Fixes: #1184

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
@@ -27,7 +27,7 @@ final class ResolveContext
      *
      * @var UriInterface
      */
-    private $requestUri;
+    private $baseUri;
 
     /**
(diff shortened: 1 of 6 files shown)


robertlemke:
@bwaidelich correct that this fix does not have to be backported to earlier versions? I wondered if it would make sense to at least back port the unit test …

Either way, to me this looks like ready to merge.

bwaidelich:
@robertlemke The bug affects 4.3 only (so 4.3.0, 4.3.1 and 4.3.2 currently) to which versions do you want to backport it?

robertlemke:
Yes, the _existing_ bug only affects 4.3 – what I meant was to backport the unit test in order to have some coverage for a possible future bug in earlier versions. But well, nevermind.

Issue #1186, opened 2018-01-21
Use neos.io as domain for the XSD files


markusguenther:
### Description

The command controller for the creation of the XSD files still uses the typo3.org domain.
This will replace the domain with neos.io

### Steps to Reproduce

1. execute `./flow documentation:generatexsd Neos\\Neos\\ViewHelpers --target-file neos.xsd`
2. take a look at the neos.xsd file header

#### Expected behavior

uses neos.io

#### Actual behavior

uses typo3.org

jonnitto:
As there is a PR for this (https://github.com/neos/neos-development-collection/pull/1860 and https://github.com/neos/flow-development-collection/pull/1187) I'll close this issue.

Thank you @markusguenther, we'll discuss and decide on the PR thread.

daniellienert:
I reopened the issue. The issue is not closed until the PR is merged - if the PR gets merged the issue will be closed automatically.

PR #1187, opened 2018-01-21
TASK: Adjust XSD domain to neos.io

markusguenther:
The XSD file generator still uses typo3.org instead of neos.io

Resolves: #1186

### DIFF
--- a/Neos.FluidAdaptor/Classes/Command/DocumentationCommandController.php
+++ b/Neos.FluidAdaptor/Classes/Command/DocumentationCommandController.php
@@ -45,8 +45,7 @@ class DocumentationCommandController extends CommandController
      */
     public function generateXsdCommand(
         $phpNamespace, $xsdNamespace = null, $targetFile = null, $xsdDomain = ''
-    )
-    {
+    ) {
         $xsdDomain = trim($xsdDomain);
         $parsedDomain = parse_url($xsdDomain);
         if (empty($xsdDomain) || !isset($parsedDomain['host'], $parsedDomain['scheme'])) {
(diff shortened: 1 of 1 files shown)


kitsunet:
But that is a breaking change, right?

markusguenther:
It depends ... but when you generate new xdd files... as uses the old one via typo3.org. You need to adjust your templates to get the new one.

So for upgrading people it is breaking. Ask myself why it is still TYPO3 ;)

kdambekalns:
IMHO that is just a namespace URL. Just like `com.sun.java.…` - does it really make sense to change that? But making it an option sounds good, so the XSD can be generated for your own stuff and the be used alongside the default one…

Then again, for our "own Neos" viewhelpers, it should be probably be `neos.io` indeed.

Breaking: is this actually in widespread use?

jonnitto:
Can we merge this one?

markusguenther:
Think I will adjust it first to have the domain as parameter. Then you can have a own domain for you own package ViewHelpers for instance :)

markusguenther:
Also added the parameter xsd-domain. So now everybody is able to generate xsd files with the own domain. If this parameter is not used or something in the URL is missing we are falling back to the neos.io URL.

kitsunet:
I wanted to think this through since a while. Shouldn't we actually version those URIs? At least major versions, as breaking chnages to VHs may occur effectively making a separate "description" necessary?

And even though that's neither required nor expected it might be neat to actually provide a description of said VHs at the given URI at some point in the future, so we might want to choose the URI in a way that may easily allow us to do just that?

Some random thoughts I had earlier on this topic.

Also this is relevant:
https://discuss.neos.io/t/fluid-syntax-and-xmlns-namespaces/1903

PR #1192, opened 2018-01-24
TASK: Move object validation out of PersistenceManager

kdambekalns:
This moves the code used to validate objects using the onFlush event
of Doctrine out of the PersistenceManager and into it's own class.

The listener is then added using Flow settings, instead of PHP code
inside PersistenceManager.initializeObject().

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationListener.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationListener.php
@@ -1,0 +1,127 @@
+<?php
+namespace Neos\Flow\Persistence\Doctrine;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\ORM\Event\OnFlushEventArgs;
+use Doctrine\ORM\UnitOfWork;
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Persistence\Exception\ObjectValidationFailedException;
+use Neos\Flow\Persistence\PersistenceManagerInterface;
+use Neos\Flow\Reflection\ClassSchema;
+use Neos\Utility\ObjectAccess;
+use Neos\Flow\Reflection\ReflectionService;
+use Neos\Utility\TypeHandling;
+use Neos\Flow\Validation\ValidatorResolver;
+
+/**
+ * Flow's Doctrine PersistenceManager
+ *
+ * @Flow\Scope("singleton")
+ * @api
+ */
+class ObjectValidationListener
+{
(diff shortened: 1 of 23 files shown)


Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationListener.php
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * @Flow\Inject
+     * @var ReflectionService
+     */
+    protected $reflectionService;
+
+    /**
+     * @Flow\Inject
+     * @var PersistenceManagerInterface
```
kitsunet: Might as well annotate the doctrine persistence manager implementation as it won't work with any other persistence manager anyway.

Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationListener.php
```diff
@@ -0,0 +1,127 @@
...
+                    unset($entityInsertions[spl_object_hash($entity)]);
+                    continue;
+                }
+
+                $knownValueObjects[$className][$identifier] = true;
+            }
+            $this->validateObject($entity, $validatedInstancesContainer);
+        }
+
+        ObjectAccess::setProperty($unitOfWork, 'entityInsertions', $entityInsertions, true);
```
kitsunet: TIL. Didn't know we did this kind of nasty magic there... But probably no other way currently?

Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationListener.php
```diff
@@ -0,0 +1,127 @@
...
+use Neos\Flow\Persistence\Exception\ObjectValidationFailedException;
+use Neos\Flow\Persistence\PersistenceManagerInterface;
+use Neos\Flow\Reflection\ClassSchema;
+use Neos\Utility\ObjectAccess;
+use Neos\Flow\Reflection\ReflectionService;
+use Neos\Utility\TypeHandling;
+use Neos\Flow\Validation\ValidatorResolver;
+
+/**
+ * Flow's Doctrine PersistenceManager
```
kitsunet: Not quite

kdambekalns:
Thanks for the review… "found" this branch while doing upmerges and decided to push it into a PR. :)

Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationListener.php
kdambekalns: TBH I don't know. We sure did it for a reson back then…

kitsunet:
Generally yes, I just wonder if we shouldn't clal the listener class something more insightful, eg. `ObjectValidationListener` or `ObjectValidationOnFlushListener`.

kdambekalns:
Because it is used

> to de-duplicate value objects and validate […]

That _could_ be split in two, of course…

kdambekalns:
THB I don't get the test failure. And I am sure it's unrelated to this PR, so… 🤷‍♂️

kdambekalns:
_That_ should work now… 🙄

Issue #1180, opened 2018-01-15
Split configuration sources do not work after cache flush


fnkr:
### Description

[Split configuration sources][1] do not work for us since we upgraded from Flow 3 to 4.3.

- If the first Flow call after cache flush is an http request, split configuration sources do not work
- If the first Flow call after cache flush is from the command line (the `./flow` tool), split configuration sources do not work, but do work for subsequent Flow calls

### Steps to Reproduce

**1)** Set up split configuration sources as described [in the docs][1].

This is how our `Packages/Vendor.Site/Classes/Package.php` looks like:

```php
namespace Vendor\Site;

use Neos\Flow\Configuration\ConfigurationManager;
use Neos\Flow\Core\Booting\Sequence;
use Neos\Flow\Core\Bootstrap;
use Neos\Flow\Package\Package as BasePackage;

class Package extends BasePackage
{
    public function boot(Bootstrap $bootstrap)
    {
        $dispatcher = $bootstrap->getSignalSlotDispatcher();

        $dispatcher->connect(ConfigurationManager::class, 'configurationManagerReady',
            function (ConfigurationManager $configurationManager) {
                $configurationManager->registerConfigurationType(
                    ConfigurationManager::CONFIGURATION_TYPE_SETTINGS,
                    ConfigurationManager::CONFIGURATION_PROCESSING_TYPE_SETTINGS,
                    true
                );
            }
        );
    }
}
```

**2)** Add `Packages/Vendor.Site/Configuration/Settings.Foo.yaml`:

```yaml
Vendor:
  Site:
    foo:
      bar: baz
```

**3)** Call Flow via http request or command line and check whether the extra configuration entries are available.

#### Expected behavior

```
$ ./flow flow:cache:flush
Flushed all caches for "Development/Vendor/Local" context.
$ ./flow configuration:show --path Vendor.Site.foo
Configuration "Settings: Vendor.Site.foo":

bar: baz
$ ./flow configuration:show --path Vendor.Site.foo
Configuration "Settings: Vendor.Site.foo":

bar: baz
```

#### Actual behavior

```
$ ./flow flow:cache:flush
Flushed all caches for "Development/Vendor/Local" context.
$ ./flow configuration:show --path Vendor.Site.foo
Configuration "Settings: Vendor.Site.foo" was empty! <========
$ ./flow configuration:show --path Vendor.Site.foo
Configuration "Settings: Vendor.Site.foo":

bar: baz
```

### Affected Versions

Neos: -

[...]

PR #1193, opened 2018-01-24
FEATURE: Enable split source for settings files

kitsunet:
Settings can now be split source.
The Flow Settings now come split to make finding something easier.

Additionally files with the same name as a configuration source but with ``yml``
extension will result in an exception to avoid confusion.

Fixes: #1180


bwaidelich:
@kitsunet you're the man! I didn't check it myself yet, but did you have a look whether we have any hard-coded checks for `Settings.yaml` in the PHP source somewhere? At least Karstens [comment](https://github.com/neos/flow-development-collection/issues/1180#issuecomment-359833041) suggests that there were some dependencies (just checking if you looked for that already, otherwise I can do so tomorrow)

Neos.Flow/Classes/Configuration/Source/YamlSource.php
```diff
@@ -53,6 +53,7 @@ public function has($pathAndFilename, $allowSplitSource = false)
     {
         if ($allowSplitSource === true) {
             $pathsAndFileNames = glob($pathAndFilename . '.*.yaml');
+            $pathsAndFileNames = array_merge($pathsAndFileNames, glob($pathAndFilename . '.*.yml'));
```
bwaidelich: why `array_merge` wouldn't
```php
$pathsAndFileNames = [glob($pathAndFIlename . '.*yaml'), glob($pathAndFilename . '.*.yml')];
```
work the same?

Neos.Flow/Classes/Configuration/Source/YamlSource.php
```diff
@@ -79,9 +81,10 @@ public function has($pathAndFilename, $allowSplitSource = false)
      */
     public function load($pathAndFilename, $allowSplitSource = false)
     {
-        $pathsAndFileNames = [$pathAndFilename . '.yaml'];
+        $pathsAndFileNames = [$pathAndFilename . '.yaml', $pathAndFilename . '.yml'];
         if ($allowSplitSource === true) {
             $splitSourcePathsAndFileNames = glob($pathAndFilename . '.*.yaml');
+            $splitSourcePathsAndFileNames = array_merge($splitSourcePathsAndFileNames, glob($pathAndFilename . '.*.yml'));
```
bwaidelich: again, is the `array_merge` necessary? (just a nitpick ofc)

Neos.Flow/Classes/Configuration/Source/YamlSource.php
kitsunet: Two globs in an array would mean two dimensional array, so that would then need to be treated differently.

Basically [[a,b], [c]] vs. [a, b, c] (array_merge).

kitsunet:
There is a whole bunch of references for `Settings.yaml` but hat is all documentation and error messages. I think it's fine if the error says `You have to set Foo.Bar in your Settings.yaml`.

We should indeed probably mention split settings in the documentation.
I might also need to add a test I guess^^

We can debate if we want to error on yml. But it seems to be not too difficult to fix this.

As for order it will be `Settings.yaml` first and then `Settings.Something.yaml`. The split files are in no particular order. I think filesystem order, whatever that may be at any point in time.

bwaidelich:
> [CM] I think it's fine if the error says You have to set Foo.Bar in your Settings.yaml.

Yes, I think so too!

> We can debate if we want to error on yml. But it seems to be not too difficult to fix this.

Great that you implemented support for both. My concern was less about the difficulty of the implementation though but about the implications: Do we *want* to allow both file suffixes? And what happens if I have both, `Settings.yaml` and `Settings.yml` (I assume both are respected with precendence to `*.yml`but it's not explicit) – but, granted, this is more of an edge case and maybe not important


> documentation! 📚 -> but it already exists [...]

Yes, it just wasn't mentioned yet that this is enabled for Settings too now. Now it is ;)

kitsunet:
Ok, Ok 😆

kitsunet:
@kdambekalns Now with more exception.

aertmann:
yay nice one!

PR #1194, opened 2018-01-26
BUGFIX: Add expire to cache tags

nlx-sascha:
The tags are never garbage collected from redis when the cache entry expires. So we need to set the same cache lifetime to the cache tags.

**What I did**
Add the same cache lifetime of the entry to the cache tags.

**How to verify it**
Configure a redis cache with a very low cache lifetime (60s).
Use `redis-cli info` to check if the database is empty.
Add a entry to redis and use the same command to verify that the entry and tags are present. This should look the `db1:keys=5,expires=5,avg_ttl=59225`. Wait 60s and and all entries should removed from the database.

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -272,9 +272,8 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
         $ttl = $this->redis->ttl($tag);
         if ($ttl === -1 || $lifetime === self::UNLIMITED_LIFETIME) {
             return self::UNLIMITED_LIFETIME;
-        } else {
-            return max($ttl, $lifetime);
         }
+        return max($ttl, $lifetime);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


kitsunet:
Now that sounds very sensible, good catch!

nlx-sascha:
I know that the calculation of the expires is not 100% atomic, but i have no idea how to calculate the max expires for a tag during the `$redis->multi`. Maybe a inline script can handle this, but I'm not sure if this would make it better.

aertmann:
Hey. Any specific reason this is only targeted master when it's a bugfix?

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -251,6 +260,23 @@ private function buildKey(string $identifier): string
...
+     * @return int
+     */
+    private function calculateExpires($tag, $lifetime)
+    {
+        $ttl = $this->redis->ttl($tag);
+        if ($ttl === -1 || $lifetime === self::UNLIMITED_LIFETIME) {
+            return self::UNLIMITED_LIFETIME;
+        } else {
+            return max($ttl, $lifetime);
+        }
```
aertmann: could have a typed return, also else statement is unnecessary

nlx-sascha: Sorry I don't get it. Why do you think the else statement is unnecessary.
Would you prefer `return ($ttl === -1 || $lifetime === self::UNLIMITED_LIFETIME) ? self::UNLIMITED_LIFETIME : max($ttl, $lifetime)`. Sorry for using the wrong branch will change that.

aertmann: just meant the `} else {` part is not needed.. it returns early

PR #1195, opened 2018-01-28
BUGFIX: Fix deprecated Symfony console helper implementations

gerhard-boden:
Due to the upgrade to symfony/console 4.0 some implementations for utilized CLI helpers changed. 
Unit tests were added for all console output helpers to prevent similar breakdowns in the future.

regression https://github.com/neos/flow-development-collection/pull/1179

### DIFF
--- a/Neos.Flow/Classes/Cli/ConsoleOutput.php
+++ b/Neos.Flow/Classes/Cli/ConsoleOutput.php
@@ -162,10 +162,7 @@ class ConsoleOutput
      */
     public function select($question, array $choices, bool $default = null, bool $multiSelect = false, $attempts = null)
     {
-        if (is_array($question)) {
-            $question = $this->combineQuestion($question);
-        }
-        $question = new ChoiceQuestion($question, $choices, $default);
+        $question = new ChoiceQuestion($this->combineQuestion($question), $choices, $default);
         $question
             ->setMaxAttempts($attempts)
             ->setMultiselect($multiSelect)
@@ -184,10 +181,7 @@ class ConsoleOutput
      */
     public function ask($question, string $default = null): string
     {
-        if (is_array($question)) {
-            $question = $this->combineQuestion($question);
-        }
-        $question = new Question($question, $default);
+        $question = new Question($this->combineQuestion($question), $default);
 
         return $this->getQuestionHelper()->ask($this->input, $this->output, $question);
     }
@@ -203,10 +197,7 @@ class ConsoleOutput
      */
     public function askConfirmation($question, bool $default = true): bool
     {
-        if (is_array($question)) {
-            $question = $this->combineQuestion($question);
-        }
-        $question = new ConfirmationQuestion($question, $default);
+        $question = new ConfirmationQuestion($this->combineQuestion($question), $default);
 
         return $this->getQuestionHelper()->ask($this->input, $this->output, $question);
     }
@@ -221,10 +212,7 @@ class ConsoleOutput
      */
     public function askHiddenResponse($question, bool $fallback = true): string
     {
-        if (is_array($question)) {
-            $question = $this->combineQuestion($question);
-        }
-        $question = new Question($question);
+        $question = new Question($this->combineQuestion($question));
         $question
             ->setHidden(true)
             ->setHiddenFallback($fallback);
@@ -248,10 +236,7 @@ class ConsoleOutput
      */
     public function askAndValidate($question, callable $validator, int $attempts = null, string $default = null): bool
     {
-        if (is_array($question)) {
-            $question = $this->combineQuestion($question);
-        }
-        $question = new Question($question, $default);
+        $question = new Question($this->combineQuestion($question), $default);
         $question
             ->setValidator($validator)
             ->setMaxAttempts($attempts);
@@ -276,10 +261,7 @@ class ConsoleOutput
      */
     public function askHiddenResponseAndValidate($question, callable $validator, int $attempts = null, bool $fallback = true): string
     {
-        if (is_array($question)) {
-            $question = $this->combineQuestion($question);
-        }
-        $question = new Question($question);
+        $question = new Question($this->combineQuestion($question));
         $question
             ->setHidden(true)
             ->setHiddenFallback($fallback)
@@ -385,12 +367,16 @@ class ConsoleOutput
     /**
      * If question is an array, split it into multi-line string
      *
-     * @param array $question
+     * @param string|array $question
      * @return string
      */
-    protected function combineQuestion(array $question): string
+    protected function combineQuestion($question): string
     {
-        return implode(PHP_EOL, $question);
+        if (is_array($question)) {
+            return implode(PHP_EOL, $question);
+        }
+
+        return $question;
     }
 
     /**
--- a/Neos.Flow/Classes/Cli/ConsoleOutput.php
+++ b/Neos.Flow/Classes/Cli/ConsoleOutput.php
@@ -92,7 +92,7 @@ class ConsoleOutput
      * @param array $arguments Optional arguments to use for sprintf
      * @return void
      */
-    public function output($text, array $arguments = []): void
+    public function output(string $text, array $arguments = []): void
     {
         if ($arguments !== []) {
             $text = vsprintf($text, $arguments);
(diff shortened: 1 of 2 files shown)


gerhard-boden:
Got to run now, will deliver tests shortly

gerhard-boden:
Started writing Unit Tests for console ouputs, but can't finish and corrected a couple of minor mistakes, please don't waste your time on reviewing, since some lines are subject to change.

gerhard-boden:
This is ready for reviews now! I also added unit tests with a coverage of 86 %, since Flow loses major CLI functionality when these helpers do not work.
Two of 15 output helpers are note easily testable, because they provide hidden input while typing within the console.

gerhard-boden:
While working on this I noticed this line:
https://github.com/neos/flow-development-collection/pull/1195/files#diff-3e19a78838a95d41e9e6247ed3c1e49cL178

Why do we want to confirm an input automatically if the user enters nothing? Isn't that kinda dangerous? Or is empty input for a y/n questions widely considered as `y` when it comes to CLI input?

aertmann:
I feel like they keep changing the API for that package, this is definitely gonna be breaking for many applications out there. Wonder if it's really worth it, but guess it should be done at some point and now is a good time due to next major. Any thoughts?

gerhard-boden:
We upgraded our Symfony dependencies from 2.3 to 4.0, so we missed the deprecation phase in 3.0 

The usage within Flow should stay almost the same as it was before. (some argument default value changes from boolean to null, but even that could be taken care of if we wanted to). So the breakiness for packages using our the Flow internal `ConsoleOutput` wrappper should be quite low. 

Unless you called the Symfony output helpers directly ofc, then you have some work to do (but from my experience it's pretty easy to update the helpers).

kdambekalns:
> Why do we want to confirm an input automatically if the user enters nothing?

That is the default the dialog will show, so it is just usual behaviour: the default is spelt uppercase, and pressing enter accepts it. It's up to the user of the method to change to no, if that seems to be a snaer default.

Neos.Flow/Classes/Cli/ConsoleOutput.php
```diff
@@ -222,15 +268,25 @@ public function askAndValidate($question, $validator, $attempts = false, $defaul
      *
      * @param string|array $question The question to ask. If an array each array item is turned into one line of a multi-line question
      * @param callable $validator A PHP callback that gets a value and is expected to return the (transformed) value or throw an exception if it wasn't valid
-     * @param integer|boolean $attempts Max number of times to ask before giving up (false by default, which means infinite)
+     * @param integer|null $attempts Max number of times to ask before giving up (false by default, which means infinite)
```
kdambekalns: Should read "null by default", no?

Neos.Flow/Classes/Cli/ConsoleOutput.php
```diff
@@ -204,13 +243,20 @@ public function askHiddenResponse($question, $fallback = true)
      * @param callable $validator A PHP callback that gets a value and is expected to return the (transformed) value or throw an exception if it wasn't valid
      * @param integer|boolean $attempts Max number of times to ask before giving up (false by default, which means infinite)
```
kdambekalns: Should read "integer" or "integer|null" and "null by default", no?

Neos.Flow/Classes/Cli/ConsoleOutput.php
```diff
@@ -143,27 +156,40 @@ public function outputTable($rows, $headers = null)
      * @param array $choices List of choices to pick from
      * @param boolean $default The default answer if the user enters nothing
      * @param boolean $multiSelect If TRUE the result will be an array with the selected options. Multiple options can be given separated by commas
-     * @param boolean|integer $attempts Max number of times to ask before giving up (false by default, which means infinite)
+     * @param boolean|null $attempts Max number of times to ask before giving up (false by default, which means infinite)
```
kdambekalns: Should read "null by default", no?

gerhard-boden:
Since I almost touched each method, I went a ahead and also added type hints for alle arguments. I usually try to avoid this in bug fixes, since it blurs the actual changes, but this is almost a rewrite anyways :)

Neos.Flow/Classes/Cli/ConsoleOutput.php
```diff
@@ -239,85 +295,127 @@ public function askHiddenResponseAndValidate($question, $validator, $attempts =
...
+        return $this->questionHelper;
+    }
+
+    /**
+     * If question is an array, split it into multi-line string
+     *
+     * @param array $question
+     * @return string
+     */
+    protected function splitQuestion(array $question): string
```
daniellienert: nitpick: The naming is strange - a split method that implodes. Would be more a `combineQuestion`?

Neos.Flow/Tests/Unit/Cli/ConsoleOutputTest.php
```diff
@@ -0,0 +1,299 @@
...
+use Neos\Flow\Cli;
+use Neos\Flow\Cli\ConsoleOutput;
+use Neos\Flow\Tests\UnitTestCase;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\StreamOutput;
+
+/**
+ * Test cases for CLI console output helpers
+ */
+class ConsoleOutputTest extends UnitTestCase
```
daniellienert: Great to have tests for these methods. As they are testing external state (php://memory) I would classify them as functional tests though.

daniellienert:
Thanks a lot @gerhard-boden for taking care!

Neos.Flow/Tests/Unit/Cli/ConsoleOutputTest.php
kdambekalns: I disagree. They test a single unit, it's just that the result isn't returned directly. Our functional tests fire up a full framework, more or less, which isn't the case here.

Neos.Flow/Classes/Cli/ConsoleOutput.php
kdambekalns: Fully agree.

Neos.Flow/Classes/Cli/ConsoleOutput.php
kdambekalns: Still says `false by default`

Neos.Flow/Classes/Cli/ConsoleOutput.php
kdambekalns: No change here…

Neos.Flow/Classes/Cli/ConsoleOutput.php
kdambekalns: Still says `false by default`

gerhard-boden: I think this view is not in sync, when i look at the latest commit (https://github.com/neos/flow-development-collection/pull/1195/commits/e096711f9de8ccfda6c2c8438281b139225cb722) I see the change

kdambekalns: Ah, that's how they mean "Show outdated" - not only the comment, but the outdated diff as well. Bummer. :)

kdambekalns:
@gerhard-boden Thanks, indeed. And almost done, but for the `splitQuestion` thing…

Neos.Flow/Classes/Cli/ConsoleOutput.php
gerhard-boden: yea, it's about splitting the array into multiple lines, but i see the confusion.  Will go with `combineQuestion` instead

kdambekalns: I'd go even further and embed the `is_array` into the method as well, then unconditionally call it where it's used. This way it seems half-baked, not only because the docblock says "If question is an array"… Comes with the price of not being able to use the `array` type hint, but…

kdambekalns:
Just to make this is seen, I posted it after the most recent commit…

> I'd go even further and embed the is_array into the method as well, then unconditionally call it where it's used. This way it seems half-baked, not only because the docblock says "If question is an array"… Comes with the price of not being able to use the array type hint, but…

gerhard-boden:
Thanks for the feedback guys! Latest changes should implement everything we agreed upon

PR #1196, opened 2018-01-28
TASK: Small boot performance improvements

kitsunet:
These changes are meant to defer initializations of low level classes as much
as possible.

ConfigurationManager had some unnecessary code and route configuration
processing was separated out to make the class smaller and better to maintain.

ReflectionService is now build via factory at runtime which means if it's not
needed it will never do its costly initialisation.

Same for the AuthenticationProviderManager internals.


Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
```diff
@@ -92,15 +100,15 @@ public function __construct(AuthenticationProviderResolver $providerResolver, Re
      * Inject the settings and does a fresh build of tokens based on the injected settings
      *
      * @param array $settings The settings
-     * @return void
+     * @return voidt
```
bwaidelich: typo

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
```diff
@@ -206,6 +206,8 @@ protected function setupSecurity()
         $this->accountRepository = $this->objectManager->get(Security\AccountRepository::class);
 
         $this->authenticationManager = $this->objectManager->get(AuthenticationProviderManager::class);
+        // Making sure providers and tokens were actually build, so the singleton TestingProvider exists.
+        $this->authenticationManager->getProviders();
```
bwaidelich: Mh.. Requiring the call of a getter is a bit smelly, no?

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
```diff
@@ -377,5 +395,7 @@ protected function buildProvidersAndTokensFromConfiguration(array $providerConfi
                 $tokenInstance->setAuthenticationEntryPoint($entryPoint);
             }
         }
+
+        $this->providerConfigurations = null;
```
bwaidelich: So `buildProvidersAndTokensFromConfiguration()` sets `$this->providerConfigurations` to `null`? That's kind of twisted..
Couldn't this be changed to sth like
```php
protected function buildProvidersAndTokensFromConfiguration()
{
    if ($this->tokens !== null) {
        return;
    }
    // ...
}
```

(and don't initialize `$tokens` at declaration time)

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
kitsunet: No because actually the provider shouldn't be a singleton. It's just the tests that do nasty things, because they manipulate the internal state of the (test) provider (which is not possible if it was not build at that time).

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
kitsunet: Will have a look, could work.

kitsunet:
@bwaidelich I settled for a `isInitialized` property to make this clear. Just checking providers or tokens seemed difficult as both are interdepednent as the tests show. So a property indicating the status seemed like the best idea.

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -305,6 +305,15 @@ Neos\Flow\Property\PropertyMapper:
           1:
             value: Flow_PropertyMapper
 
+#
+# Reflection
+#
+Neos\Flow\Reflection\ReflectionService:
+  factoryObjectName: Neos\Flow\Reflection\ReflectionServiceFactory
+
+Neos\Flow\Reflection\ReflectionServiceFactory:
```
daniellienert: The ReflectionServiceFactory already has a singleton annotation. What is that configuration for?

daniellienert:
@kitsunet - Sadly there are some merge conflicts in the meantime. Would you solve them?

kdambekalns:
@bwaidelich Could you reconsider your vote on this one?

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
bwaidelich: 👍 I probably missed out that this was for tests only the first time around

PR #1207, opened 2018-01-31
FEATURE: Introduce findAllIterator() method

daniellienert:
> If the amount of the stored data increases, receiving all objects using a ``findAll()`` may
> consume a lot more memory than available. In this cases, you should use the ``findAllIterator()``.
> This method returns an ``IterableResult``over which you can iterate, getting only one object at a time.

As this pattern is used a lot in Neos repositories, it would make sense to move it to the base repository class.

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Repository.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Repository.php
@@ -1,5 +1,4 @@
 <?php
-
 namespace Neos\Flow\Persistence\Doctrine;
 
 /*
@@ -166,7 +165,6 @@ abstract class Repository extends EntityRepository implements RepositoryInterfac
             $iteration++;
         }
     }
-
 
     /**
      * Finds an object matching the given identifier.
--- a/Neos.Flow/Classes/Persistence/Doctrine/Repository.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Repository.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Neos\Flow\Persistence\Doctrine;
 
 /*
@@ -11,6 +12,9 @@ namespace Neos\Flow\Persistence\Doctrine;
  * source code.
  */
 
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityRepository;
 use Doctrine\ORM\Internal\Hydration\IterableResult;
 use Neos\Flow\Annotations as Flow;
@@ -32,6 +36,11 @@ abstract class Repository extends EntityRepository implements RepositoryInterfac
     protected $persistenceManager;
 
     /**
+     * @var EntityManager
+     */
+    protected $entityManager;
+
+    /**
      * Warning: if you think you want to set this,
      * look at RepositoryInterface::ENTITY_CLASSNAME first!
      *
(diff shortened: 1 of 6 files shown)


kdambekalns:
This `Repository` is persistence-layer agnostic, so the new stuff must go into the Doctrine-specific `Repository`…

kdambekalns:
Haha, what @kitsunet just wrote. That being said, I'd be fine with getting rid of non-Doctrine persistence for Flow 5.0 😎

kitsunet:
I am not sure of the implications. I like the idea of having hte option of non doctrine persistence. but \o/ neos won't work with anyting else at the moment anyway.

daniellienert:
Interesting - I did't know that this one exists. But that would mean, that e.g. `NodeDataRepository` inherits the wrong `Neos\Flow\Persistence\Repository` atm?

albe:
Regarding removing the non-doctrine persistence:
https://twitter.com/ih8nickfinding/status/912601641068752899

daniellienert:
@kitsunet - I moved the methods. Would you have another look?

Neos.Flow/Classes/Persistence/Doctrine/Repository.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
kdambekalns: We don't do empty lines before `namespace`.

Neos.Flow/Classes/Persistence/Repository.php
```diff
@@ -27,6 +28,15 @@
...
     protected $persistenceManager;
 
+    /**
+     * Doctrine's Entity Manager. Note that "ObjectManager" is the name of the related
+     * interface ...
+     *
+     * @Flow\Inject
+     * @var ObjectManager
+     */
+    protected $entityManager;
```
kdambekalns: This is not needed (same for the related use statement)

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
```diff
@@ -196,6 +196,7 @@ From Evans, the rules we need to enforce include:
 * When a change to any object within the Aggregate boundary is committed, all invariants
   of the whole Aggregate must be satisfied.
 
+
```
kdambekalns: A specific reason for this empty line?

Neos.Flow/Tests/Functional/Persistence/Fixtures/SubSubEntityRepository.php
```diff
@@ -11,7 +11,7 @@
  * source code.
  */
 
-use Neos\Flow\Persistence\Repository;
+use Neos\Flow\Persistence\Doctrine\Repository;
```
kdambekalns: Why? Is the fixture specific for a test related to the Doctrine repository?

Neos.Flow/Tests/Functional/Persistence/Fixtures/SuperEntityRepository.php
```diff
@@ -11,7 +11,7 @@
  * source code.
  */
 
-use Neos\Flow\Persistence\Repository;
+use Neos\Flow\Persistence\Doctrine\Repository;
```
kdambekalns: Why? Is the fixture specific for a test related to the Doctrine repository?

Neos.Flow/Classes/Persistence/Doctrine/Repository.php
```diff
@@ -109,6 +129,45 @@ public function findAll()
...
+            yield $object;
+            if ($callback !== null) {
+                call_user_func($callback, $iteration, $object);
+            }
+
+            $iteration++;
+        }
+    }
+
+
```
kdambekalns: One empty line between methods.

Neos.Flow/Tests/Functional/Persistence/Fixtures/SubSubEntityRepository.php
daniellienert: I think so. As they are only used in the `Neos\Flow\Tests\Functional\Persistence\**Doctrine**\RepositoryTest` they should be of kind doctrine repository.

Neos.Flow/Tests/Functional/Persistence/Fixtures/SuperEntityRepository.php
daniellienert: Same here

Neos.Flow/Tests/Functional/Persistence/Fixtures/SubSubEntityRepository.php
kdambekalns: 👍

Neos.Flow/Tests/Functional/Persistence/Fixtures/SuperEntityRepository.php
kdambekalns: 👍

PR #1209, opened 2018-02-02
BUGFIX: Behat subprocess should escape paths in command

kitsunet:


kitsunet:
With a path that includes spaces I need this to run behat tests.

kdambekalns:
Ah, @kitsunet – maybe put this into the 4.0 branch?

kitsunet:
Good point

kitsunet:
Replaced by #1210

PR #1210, opened 2018-02-03
BUGFIX: Behat subprocess should escape paths in command

kitsunet:


kitsunet:
Cherry pick of the master change.

PR #1211, opened 2018-02-05
BUGFIX: Avoid migration failure if files are not existing in FS

kaystrobach:
**What I did**

This fix migrates non existing files into a separate table, so that they can be fixed lateron.
Having this change creates a new table, named
typo3_flow_resource_resource_backupfiles, which contains file entries, which do not comply
with the contraint rules.

**How I did it**

Adjust migration

**How to verify it**

* add a file into the resources table, remove the actual file from the FS, then run migrations
* run migrations esp. *20141015125841*

**Checklist**

### DIFF
--- a/TYPO3.Flow/Migrations/Mysql/Version20141015125841.php
+++ b/TYPO3.Flow/Migrations/Mysql/Version20141015125841.php
@@ -44,11 +44,14 @@ class Version20141015125841 extends AbstractMigration
             return;
         }
         $resourcesResult = $this->connection->executeQuery('SELECT persistence_object_identifier, sha1, filename FROM typo3_flow_resource_resource');
+
         while ($resourceInfo = $resourcesResult->fetch(\PDO::FETCH_ASSOC)) {
             $resourcePathAndFilename = FLOW_PATH_DATA . 'Persistent/Resources/' . $resourceInfo['sha1'];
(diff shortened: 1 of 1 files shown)


kdambekalns:
@kaystrobach Can't you fix this by running `resource:clean` before doing the migration?

kaystrobach:
hey @kdambekalns this has sadly not worked out.

TYPO3.Flow/Migrations/Mysql/Version20141015125841.php
```diff
@@ -67,6 +70,8 @@ public function postUp(Schema $schema)
                 continue;
             }
 
+            $this->write(print_r(array('persistent', $mediaType, $md5, $filesize, $resourceInfo['persistence_object_identifier']), true));
```
kdambekalns: This looks like debug output, is that intentionally left here?

kaystrobach: Can be removed, i used it during the migration.

kaystrobach: Not creating the table is also possible. One solution would be to check if the migration can be applied before it starts and do not fail at 95%

kaystrobach:
Then we could also add the hint how to remove data, which is not valid anymore ...

PR #1213, opened 2018-02-09
!!! TASK: Add factory method to ``AuthenticationProviderInterface``

kitsunet:
There was a long standing comment about having the constructor be part
of the interface but that was deactivated back in the day as the proxy
constructor was incompatible with the interface then. A interface
defined way of creating an ``AuthenticationProvider`` makes sense though
therefore we introduce a static ``create`` method that returns a fresh
instance of the class in question. This removes the implicit knowledge
about the constructor in the code.

This is breaking if you implemented your own
``AuthenticationProviderInterface`` without making use of the
``AbstractProvider`` which provides a basic ``create`` method.
You may need to add this create method in your implementation.

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/Provider/AbstractProvider.php
+++ b/Neos.Flow/Classes/Security/Authentication/Provider/AbstractProvider.php
@@ -11,7 +11,6 @@ namespace Neos\Flow\Security\Authentication\Provider;
  * source code.
  */
 
-use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Security\Authentication\AuthenticationProviderInterface;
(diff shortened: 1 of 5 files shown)


Neos.Flow/Classes/Security/Authentication/Provider/AbstractProvider.php
```diff
@@ -30,6 +29,16 @@
...
     protected $options = [];
 
+    /**
+     * @param string $name
+     * @param array $options
+     * @return AuthenticationProviderInterface
+     */
+    public static function create(string $name, array $options)
+    {
+        return new static($name, $options);
```
albe: Could we now even make the constructor protected, so we don't have to care about unintentional instanciation?

PR #1217, opened 2018-02-20
BUGFIX: Avoid useless calls to verifyRedisVersionIsSupported()

kdambekalns:
The `set()` call returns a boolean, so the check `instanceof \Redis` always fails.
That means on every set, the "supported version check" is run.


kdambekalns:
Checking if the set succeeded is a good thing, I guess. And unless it fails, there's not much overhead. But yeah, it is a bit weird.

PR #1219, opened 2018-02-22
FEATURE: Add `remove` flowQuery operation

mficzel:
The operation removes the given items from the current FlowQuery context. To do so it accepts a single argument that may be an array, an object or a FlowQuery.

### DIFF
--- a/Neos.Eel/Classes/FlowQuery/Operations/RemoveOperation.php
+++ b/Neos.Eel/Classes/FlowQuery/Operations/RemoveOperation.php
@@ -1,0 +1,59 @@
+<?php
+namespace Neos\Eel\FlowQuery\Operations;
+
+/*
+ * This file is part of the Neos.Eel package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Eel\FlowQuery\FlowQuery;
+use Neos\Flow\Annotations as Flow;
+
+/**
+ * Removes the given items from the current context.
+ * The operation accepts one argument that may be an Array, a FlowQuery
+ * or an Object.
+ */
+class RemoveOperation extends AbstractOperation
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @var string
+     */
+    protected static $shortName = 'remove';
+
+    /**
+     * {@inheritdoc}
+     *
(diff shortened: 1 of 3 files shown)


mficzel:
Possible use cases are thinks like:

- Remove the documentNode from the list of documents like -
- Get all siblings of a node `${q(node).parent().children().remove(node).get()}`
- Remove manually selected documents from a query result to avoid showing items twice in a list

Neos.Eel/Classes/FlowQuery/Operations/RemoveOperation.php
```diff
@@ -0,0 +1,59 @@
...
+        $valuesToRemove = [];
+        if (isset($arguments[0])) {
+            if (is_array($arguments[0]) || $arguments[0] instanceof \Traversable) {
+                foreach ($arguments[0] as $item) {
+                    $valuesToRemove[] = $item;
+                }
+            } else {
+                $valuesToRemove[] = $arguments[0];
+            }
+        }
```
aertmann: Hey it looks fine and this is just a suggestion as personally I try to avoid creating loops if possible..

    $valuesToRemove = is_array($arguments[0]) ? $arguments[0] : [$arguments[0]];
     if ($arguments[0] instanceof \Traversable) {
        $valuesToRemove = iterator_to_array($arguments[0]);
    }

Neos.Eel/Classes/FlowQuery/Operations/RemoveOperation.php
```diff
@@ -0,0 +1,59 @@
...
+        $context = $flowQuery->getContext();
+        $newContext = [];
+        foreach ($context as $item) {
+            if (in_array($item, $valuesToRemove, true) === false) {
+                $newContext[] = $item;
+            }
+        }
```
aertmann: same here

    $filteredContext = array_filter($context, function($item) {
        return in_array($item, $valuesToRemove, true) === false;
    }, []);

Neos.Eel/Classes/FlowQuery/Operations/RemoveOperation.php
mficzel: I adjusted the core pretty much as suggested. Thanks for the feedback.

Neos.Eel/Classes/FlowQuery/Operations/RemoveOperation.php
mficzel: I adjusted the core pretty much as suggested. Thanks for the feedback.

PR #1228, opened 2018-03-07
TASK: Use RouterInterface instead of concrete implementation

Torsten85:
Currently the concrete `Router` Implementation is used instead of the more flexible `RouterInterface`.

This PR ensures the `RouterInterface` is used.

### DIFF
--- a/Neos.Flow/Classes/Command/RoutingCommandController.php
+++ b/Neos.Flow/Classes/Command/RoutingCommandController.php
@@ -19,7 +19,7 @@ use Neos\Flow\Mvc\Routing\Dto\RouteContext;
 use Neos\Flow\Mvc\Routing\Dto\RouteParameters;
 use Neos\Flow\Mvc\Routing\Exception\InvalidControllerException;
 use Neos\Flow\Mvc\Routing\Route;
-use Neos\Flow\Mvc\Routing\RouterInterface;
+use Neos\Flow\Mvc\Routing\Router;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 
 /**
@@ -37,7 +37,7 @@ class RoutingCommandController extends CommandController
 
     /**
      * @Flow\Inject
-     * @var RouterInterface
+     * @var Router
      */
     protected $router;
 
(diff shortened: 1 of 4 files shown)


Neos.Flow/Classes/Command/RoutingCommandController.php
```diff
@@ -37,7 +37,7 @@ class RoutingCommandController extends CommandController
 
     /**
      * @Flow\Inject
-     * @var Router
+     * @var RouterInterface
```
bwaidelich: The RoutingCommandController invokes methods on the router that are not part of the interface, namely `getRoutes()`

Neos.Flow/Classes/Http/Client/InternalRequestEngine.php
```diff
@@ -49,7 +49,7 @@ class InternalRequestEngine implements RequestEngineInterface
 
     /**
      * @Flow\Inject(lazy = false)
-     * @var Router
+     * @var RouterInterface
      */
     protected $router;
```
bwaidelich: You didn't adjust the return type hint of the respective getter (https://github.com/Torsten85/flow-development-collection/blob/39e601e0312fe5bc3a8ef0e9f4a937063e3cda4d/Neos.Flow/Classes/Http/Client/InternalRequestEngine.php#L138)

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -292,6 +292,9 @@ Neos\Flow\Property\PropertyMapper:
 # ResourceManagement                                                       #
 #                                                                          #
 
+Neos\Flow\Mvc\Routing\RouterInterface:
+  className: Neos\Flow\Mvc\Routing\Router
+
```
bwaidelich: You changed this in the section "ResourceManagement", There is already one for "MVC"

bwaidelich:
Makes sense in general, but there are some minor issues with the current change

Neos.Flow/Classes/Http/Client/InternalRequestEngine.php
Torsten85: Sorry for that, was a bit in a hurry ;)

PR #1230, opened 2018-03-07
BUGFIX: Detect checked state for bound object collections

### LINKED ISSUE #1229: Using collection property with CheckboxViewHelper never checks checked automatically
### Description

The checkbox VH can be bound to an object property. If that property contains a collection of objects, the automatic setting of the `checked` attribute based on the `value` attribute does not work.

### Steps to Reproduce

1. Assign a `User` object to a form, it has a `groups` property.
2. Assign an array of `Group` objects to the form.
3. Render a list of checkboxes like
    ```
    <f:for each="{groups}" as="group">
        <f:form.checkbox property="groups" value="{group}"/>`
    </f:for>
    ```

#### Expected behavior

For any group the user is assigned, the `checked` attribute should be checked.

#### Actual behavior

No checkbox is checked.

### Affected Versions

Flow: 4.0 and up, maybe older versions

kdambekalns:
When a collection property with objects is bound to a checkbox VH,
this change makes sure to check object identifiers against the passed
value when automatically determining the `checked` state.

Fixes #1229

### DIFF
--- a/Neos.FluidAdaptor/Classes/ViewHelpers/Form/CheckboxViewHelper.php
+++ b/Neos.FluidAdaptor/Classes/ViewHelpers/Form/CheckboxViewHelper.php
@@ -93,21 +93,17 @@ class CheckboxViewHelper extends AbstractFormFieldViewHelper
         }
         if (is_array($propertyValue)) {
             if ($checked === null) {
-                if (TypeHandling::isSimpleType(TypeHandling::getTypeForValue(current($propertyValue))) === false || Arrays::containsMultipleTypes($propertyValue)) {
-                    $checked = false;
-                    foreach ($propertyValue as $value) {
-                        if (TypeHandling::isSimpleType(TypeHandling::getTypeForValue($value))) {
-                            $checked = $valueAttribute === $value;
-                        } else {
-                            // assume an entity
-                            $checked = $valueAttribute === $this->persistenceManager->getIdentifierByObject($value);
-                        }
-                        if ($checked === true) {
-                            break;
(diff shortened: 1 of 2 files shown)


bwaidelich:
as a bugfix shouldn't this target a lower branch?

kdambekalns:
Absolutely, just didn't select the correct branch…

kdambekalns:
Ah, there is Travis…

albe:
Looks good, but yeah, a test would be cool

Neos.FluidAdaptor/Classes/ViewHelpers/Form/CheckboxViewHelper.php
```diff
@@ -89,7 +93,22 @@ public function render($checked = null, $multiple = null)
         }
         if (is_array($propertyValue)) {
             if ($checked === null) {
-                $checked = in_array($valueAttribute, $propertyValue, true);
+                if (TypeHandling::isSimpleType(TypeHandling::getTypeForValue(current($propertyValue))) === false || Arrays::containsMultipleTypes($propertyValue)) {
```
bwaidelich: I think we can just change this to always walk through the array. `Arrays::containsMultipleTypes()` iterates through it anyways.

bwaidelich:
@kdambekalns sorry, I messed with Git..

kdambekalns:
Regarding 3.3… well, I was just too lazy, AFAIR

bwaidelich:
I think 4.0 as target is fine!

Issue #1232, opened 2018-03-13
Available Locale Detection is finding too many invalid locales


albe:
### Description

The scanning for available Locales, which is scanning Folders inside `Resources/*` for filenames that contain a locale-like prefix before the file extension, is currently also taking a lot of invalid locales into account. Therefore things like "cli", "peg", "min", "css" and others are considered "available" Locales (for example when scanning minified css/js files inside Public Resources).

### Affected Versions

Flow: all

### Strategies to solve the issue:

We could further restrict scanning of folders for next major version (see also https://github.com/neos/flow-development-collection/pull/385) so that really only the Translation folders are scanned by default. This would be easy to implement, but needs some documentation.

Or we allow to somehow also specify, which packages are scanned at all, so that for example only the application package can be scanned and hence the avilable locales be controlled very directly. This would require to change the `includePaths` and/or `excludePatterns` options for the `i18n.scan` setting.

kitsunet:
TBH, I would scrap this scanning and make it something you configure?

albe:
I do think it has some value when scanning translation files of the application package, so you don't have to manually configure a list of locales you support, but instead just drop another translation file in.

But a configuration option to define a list of "available" Locales sounds reasonable.

cwenzelg:
we are also currently working on translations. i would love the idea to only define the list of available locals via a configuration as we have some langue files already there but they are still work in progress and not considered "available" yet.

albe:
Okay, so it's now (as of the upcoming 5.0) possible to set an explicit list of available Locales via configuration, which will prevent the scanning from happening. Also, the scanning will now only go through `Public` and `Private/Translations` folders by default. We are thereby now free to deprecate the scanning if we decide to, since there is an alternative.

For now, I see this issue as resolved, so closing this.

PR #1234, opened 2018-03-16
!!!TASK: Only scan Private/Translations for available locales

albe:
Before the full `Resources/Private` folder was scanned for available locales, which also included
for example the CLDR, which ended up filling the available locales with much more locales than
are actually considered "available" in a normal Flow application.
This will therefore allow applications to define available locales easily from the Translations
provided.

This is breaking, because it will end up with less available locales by default, since only the
locales of Flow Translations are considered available, instead of all of CLDR locales.

Related to #1232

### DIFF
--- a/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Internationalization.rst
+++ b/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Internationalization.rst
@@ -38,8 +38,8 @@ use the i18n service API to obtain these verified ``Locale`` objects.
   the ``Neos.Flow.i18n.scan.includePaths`` setting. This is useful to restrict the scanning
   to specific paths when you have a big file structure in your package ``Resources``.
   You can also blacklist folders through ``Neos.Flow.i18n.scan.excludePatterns``.
-  By default all folders except 'node_modules', 'bower_components' and any folder starting
-  with a dot will be scanned.
+  By default the ``Public`` and ``Private/Translations`` folders, except 'node_modules',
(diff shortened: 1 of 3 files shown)


kdambekalns:
@albe The test failures seem easy to fix, could you have a look?

albe:
Will take a look

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Internationalization.rst
```diff
@@ -38,8 +38,8 @@ use the i18n service API to obtain these verified ``Locale`` objects.
   the ``Neos.Flow.i18n.scan.includePaths`` setting. This is useful to restrict the scanning
   to specific paths when you have a big file structure in your package ``Resources``.
   You can also blacklist folders through ``Neos.Flow.i18n.scan.excludePatterns``.
-  By default all folders except 'node_modules', 'bower_components' and any folder starting
-  with a dot will be scanned.
+  By default the ``Public`` and ``Private/Translations`` folders, except 'node_modules',
```
daniellienert: The node_modules and bower_components folders are normally subfolders of the Private folder, which is nor scanned anymore, so excludePatterns and the hint could be removed also?

albe: Sometimes people seem to do have node_modules and other such folders in `Resources\Public`, at least according to Aske as of https://github.com/neos/flow-development-collection/pull/385#issuecomment-226050606

PR #1237, opened 2018-03-19
Feature changing loggers via settings

cwenzelg:
fix to read the settings correctly for the SqlLogger backend.
move setting all lockers in the settings like it already is for the SystemLogger so loggers can be changed via the settings.


kitsunet:
I would really like to get https://github.com/neos/flow-development-collection/pull/1171 in first...

kitsunet:
I think this is moot with the logging changes as the Logger is deprecated and you should use the PSR LoggerInterface everywhere.

PR #1240, opened 2018-03-22
BUGFIX: IPv6 support in Uri, parse_url and TrustedProxiesComponent

### LINKED ISSUE #1238: TrustedProxiesComponent: errors with IPv6
### Description

When using IPv6 addresses, an error can be triggered.

#### Actual behavior

Using headers like these, coming from Varnish in this case:

```
X-Forwarded-For: 3707:ff10:c5:509a::2
X-Forwarded-Proto: http
X-Forwarded-Port: 80
X-Real-Ip: 3707:ff10:c5:509a::2
X-Forwarded-Server: [3b00:f48:1008::212:183:20]
X-Forwarded-Host: [3b00:f48:1008::212:183:20]
Host: [3b00:f48:1008::212:183:20]
```
leads to

```
Exception #1184071241 in line 58 of /…/Neos_Flow_Http_Component_TrustedProxiesComponent.php: "20]" is not valid port number as part of a URI.

6 Neos\Flow\Http\Uri::setPort("20]")
5 Neos\Flow\Http\Component\TrustedProxiesComponent_Original::handle(Neos\Flow\Http\Component\ComponentContext)
4 Neos\Flow\Http\Component\ComponentChain_Original::handle(Neos\Flow\Http\Component\ComponentContext)
3 Neos\Flow\Http\Component\ComponentChain_Original::handle(Neos\Flow\Http\Component\ComponentContext)
2 Neos\Flow\Http\RequestHandler::handleRequest()
1 Neos\Flow\Core\Bootstrap::run()
```

### Affected Versions

Flow: all versions having `TrustedProxiesComponent`

kdambekalns:
Fixes #1238

### DIFF
--- a/Neos.Utility.Unicode/Classes/TYPO3/Flow/Utility/Unicode/Functions.php
+++ b/Neos.Utility.Unicode/Classes/TYPO3/Flow/Utility/Unicode/Functions.php
@@ -200,7 +200,9 @@ abstract class Functions
         }
 
         // the host and port must be used as is, to allow IPv6 syntax, e.g.: [3b00:f59:1008::212:183:20]:8080
-        $components['host'] = $componentsFromUrl['host];
+        if (array_key_exists('host', $componentsFromUrl)) {
+            $components['host'] = $componentsFromUrl['host'];
+        }
         if (array_key_exists('port', $componentsFromUrl)) {
             $components['port'] = (integer)$componentsFromUrl['port'];
         } else {
--- a/Neos.Utility.Unicode/Classes/TYPO3/Flow/Utility/Unicode/Functions.php
+++ b/Neos.Utility.Unicode/Classes/TYPO3/Flow/Utility/Unicode/Functions.php
@@ -179,6 +179,15 @@ abstract class Functions
      */
     public static function parse_url($url, $component = -1)
     {
+        $hostFromUrl = parse_url($url, PHP_URL_HOST);
+        if ($hostFromUrl === false) {
+            return false;
+        }
+        $portFromUrl = parse_url($url, PHP_URL_PORT);
+        if ($portFromUrl === false) {
+            return false;
+        }
+
         $encodedUrl = preg_replace_callback('%[^:@/?#&=\.]+%usD', function ($matches) {
             return urlencode($matches[0]);
         }, $url);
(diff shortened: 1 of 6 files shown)


Neos.Utility.Unicode/Classes/TYPO3/Flow/Utility/Unicode/Functions.php
```diff
@@ -179,6 +179,15 @@ public static function pathinfo($path, $options = null)
...
     public static function parse_url($url, $component = -1)
     {
+        $hostFromUrl = parse_url($url, PHP_URL_HOST);
+        if ($hostFromUrl === false) {
+            return false;
+        }
+        $portFromUrl = parse_url($url, PHP_URL_PORT);
+        if ($portFromUrl === false) {
+            return false;
+        }
```
bwaidelich: we now call `parse_url()` three times. I guess for a good reason, but just to make sure: In which cases would the above checks return false and shouldn't there be a test covering these?

kdambekalns: Ok, removed one of the calls. As for when this may fail: just as in the call that was there before this change… if something goes wrong, handle it.

TYPO3.Flow/Classes/TYPO3/Flow/Http/Component/TrustedProxiesComponent.php
```diff
@@ -52,11 +52,20 @@ public function handle(ComponentContext $componentContext)
...
         $portFromHost = null;
         if ($hostHeader !== null) {
-            $portSeparatorIndex = strrpos($hostHeader, ':');
-            if ($portSeparatorIndex !== false) {
-                $portFromHost = substr($hostHeader, $portSeparatorIndex + 1);
-                $trustedRequest->getUri()->setPort($portFromHost);
-                $hostHeader = substr($hostHeader, 0, $portSeparatorIndex);
+            if (strpos($hostHeader, '[') === 0 && strrpos($hostHeader, ']') !== false) {
+                $portSeparatorIndex = strrpos($hostHeader, ':', -strrpos($hostHeader, ']'));
+                if ($portSeparatorIndex !== false) {
```
albe: This block could be moved outside of the brackets check to calculate the `$portSeparatorIndex`.

PR #1243, opened 2018-03-28
FEATURE: Add a method to destroy remote sessions

### LINKED ISSUE #1051: Allow "remote" logout for other accounts' sessions
### Description

Currently, there is no clean way to remove sessions that belong to another account. Deactivating an account (e.g. by setting a past expirationDate) will leave the sessions intact.

A way around this is to mimic the behavior of the AuthenticationProviderManager when creating sessions:

`$sessionManager->destroySessionsByTag( 'Neos-Flow-Security-Account-' . md5($account->getAccountIdentifier()) );`

which will effectively "log out" the given `$account`.

#### Wanted Behavior

Some API that does not involve calculating / guessing session tags, perhaps `$sessionManager->destroySessionsByAccount($account)`.

### Affected Versions

Flow 4.1.5, maybe earlier.

maltemuth:
**What I did**

Expose a public method to destroy sessions for a given Account instance, effectively "logging out" the account. Resolves #1051 

**How I did it**

* add a method getSessionTagForAccount to Security\Context
* use this method instead of duplicate session tag knowledge in existing code
* add a method destroySessionsForAccount to Security\Context that
  destroys all sessions belonging to the given account

**How to verify it**

@TODO: write a tests that shows the behavior


maltemuth:
This still needs tests:

- Verify that the code in AccountRepository is still working
- Verify that the new method actually "logs out"
- Verify that the new methods leaves the "local" session intact

I can't add labels - i'd like to have "WIP" and "Help needed", if possible:
- Should those be functional or behavioral tests?
- Can somebody point me to "create a session for $account" test code?

Neos.Flow/Classes/Security/AccountRepository.php
```diff
@@ -57,9 +58,9 @@ class AccountRepository extends Repository
     public function remove($object)
     {
         parent::remove($object);
-        /** @var Account $object */
-        $tag = 'Neos-Flow-Security-Account-' . md5($object->getAccountIdentifier());
-        $this->sessionManager->destroySessionsByTag($tag, sprintf('The account %s (%s) was deleted', $object->getAccountIdentifier(), $object->getAuthenticationProviderName()));
+        
+        // destroy the sessions for the account to be removed
+        (new SecurityContext)->destroySessionsForAccount($object);
```
albe: Not sure about this `(new SecurityContext)->` idiom, we use this nowhere and this also makes testing this method harder. I'd go with dependency injection. Also, the "reason" string is missing now `sprintf('The account %s (%s) was deleted', $object->getAccountIdentifier(), $object->getAuthenticationProviderName())`

maltemuth: Many thanks for taking time to comment on this. I'm also unsure about this idiom; from my point-of-view it highlights a perceived architectural issue: dependency injection would inject _the_ security context of the current request, which is never needed; rather, _any_ security context will do, and is only needed in this one method.

What do you think: am I just "seeing things" or is this actually somewhat fine?

albe: You're right, this hints at an architectural smell. SecurityContext is specific to one session, but those methods are generic. This can of course be solved by introducing another layer of abstraction, but is that worth the added conceptual complexity?

maltemuth: IMHO, this PR has already fixed one architectural smell: If you need to destroy sessions for accounts, you need to know exactly how to do that. So it is replacing one architectural smell for another, which is not necessarily bad.

Right now, the architecture in the bigger picture has no concept of a "remote session"; any notion of relations between Account and Session objects are strictly request-local and handled by the Security Context internally. I think something like a derived "Remote Security Context" might be what I'm looking for - but architecture on this scale is way beyond my skills for now.

Regarding worth, in the context of Neos as a product, there might be one if the Backend actually makes use of this (e.g. "User XXX is currently logged in" or "User XXX just updated a node with path /foo/bar") but right now, I don't see any good examples. In the context of Flow, it would be possible to implement functionality like "You are currently logged in on the following devices: …" with a remote-logout option (for example like WhatsApp Web, or Netflix).

In the context of #1051 there is no need to complicate the situation further.

albe: Oh I do see the use of this feature within Flow. I rather questioned if adding another LoA just to make the solution more architectural clean is worth it.

I'd probably go with injecting the SecurityContext and adding a documentation block, that explains why it's done in this place and that it's not for the current Context, but because the Context is the best place to hold the necessary knowledge to destroy arbitrary sessions by account.

albe:
I think all this needs is tests to cover the feature and otherwise this looks good. Not sure I find time to go for those, but would be cool to have this in.

dfeyer:
Nice change, I all for it to be merged with a bunch of test.

maltemuth:
I'd love to take a stab at the tests, but I need a pointer on how to set up sessions during tests.

bwaidelich:
*ping* last call to get this into 5.2 (although it could even be considered a bug)
@maltemuth do you need any more help with this, I'm happy to provide some if I can

albe:
After looking again at this, I'm inclined to merge this without a functional test, for following reason:
The actual functionality here is inside the already existing `sessionManager->destroySessionsByTag()` method, which should either be tested already - or if not, can as well be solved in an own PR. The only thing this PR provides, is move the logic for generating Tags into a dedicated function (which could be unit tested as above) and opening the above call publicly.

maltemuth:
Unfortunately, I can't make time right now to work on this. I think I will be able to around New Year's Eve - I think I can provide the Unit Test without further assistance.

jonnitto:
Should we merge this even if there is a StyleCI issue?

daniellienert:
@jonnitto - that would mean that master is broken afterwards. But we could fix it there before new branches are branched out.

Neos.Flow/Classes/Security/Context.php
```diff
@@ -867,4 +867,26 @@ public function getContextHash()
...
+        return 'Neos-Flow-Security-Account-' . md5($account->getAccountIdentifier());
+    }
+
+    /**
+     * destroys all sessions belonging to the given $account
+     *
+     * @param Account $account
+     * @param string $reason
+     */
+    public function destroySessionsForAccount(Account $account, $reason = '')
```
daniellienert:
```suggestion
    public function destroySessionsForAccount(Account $account, string $reason = ''): void
```

albe:
Christian just gave the solution: The imported namespace was the same as in the file, so the import was not neccessary. Should be green now!

Neos.Flow/Classes/Security/Context.php
albe: Good catch.

jonnitto:
@albeme @daniellienert 👍 Thank you for your effort! Great

If it is ok, we should merge this one asap

albe:
Ready to merge from my side

PR #1252, opened 2018-03-30
TASK: Remove references to excludeClasses from object management

kitsunet:
The setting ``Neos.Flow.excludeClasses`` was deprecated since a while
but the code handling it was not removed until now.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
@@ -236,77 +236,37 @@ class CompileTimeObjectManager extends ObjectManager
 
     /**
      * Given an array of class names by package key this filters out classes that
-     * have been configured to be excluded or included by object management.
+     * have been configured to be included by object management.
      *
      * @param array $classNames 2-level array - key of first level is package key, value of second level is classname (FQN)
      * @param array $includeClassesConfiguration array of includeClasses configurations
-     * @return array The input array with all configured to be excluded from object management filtered out
+     * @return array The input array with all configured to be included in object management added in
      * @throws InvalidConfigurationTypeException
      * @throws \Neos\Flow\Configuration\Exception\NoSuchOptionException
      */
     protected function filterClassNamesFromConfiguration(array $classNames, $includeClassesConfiguration)
     {
-        if (isset($this->allSettings['Neos']['Flow']['object']['excludeClasses'])) {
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
```diff
@@ -236,77 +236,37 @@ protected function registerClassFiles(array $packages)
...
-    protected function applyClassFilterConfiguration($classNames, $filterConfiguration, $includeOrExclude = 'include')
+    protected function applyClassFilterConfiguration($classNames, $filterConfiguration)
     {
-        if (!in_array($includeOrExclude, ['include', 'exclude'])) {
-            throw new \InvalidArgumentException('The argument $includeOrExclude must be one of "include" or "exclude", the given value was not allowed.', 1423726253);
-        }
         foreach ($filterConfiguration as $packageKey => $filterExpressions) {
             if (!array_key_exists($packageKey, $classNames)) {
-                $this->systemLogger->log('The package "' . $packageKey . '" specified in the setting "Neos.Flow.object.' . $includeOrExclude . 'Classes" was either excluded or is not loaded.', LOG_DEBUG);
+                $this->systemLogger->log('The package "' . $packageKey . '" specified in the setting "Neos.Flow.object.includeClasses" was either excluded or is not loaded.', LOG_DEBUG);
```
albe: Is `was either excluded` still correct? Rather `was either not included`, no?

kdambekalns:
I did not check but… is there still `excludeClasses` in settings and/or documentation?

kdambekalns:
@albe Fine with you  now?

kaystrobach:
is there any other way of excluding classes from the proxy class generation?
we used that for perfomance optimization and to exclude azti generated coded stubs from being parsed.

kdambekalns:
https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartV/AnnotationReference.html#proxy

kaystrobach:
won't work if you auto generate code for wsdl etc. 

excludeClasses was the perfect solution.

we now moved these classes out of the Classes folder to avoid the problem

albe:
That's a valid option, or maybe use a non-neos package for those classes.

kaystrobach:
mhmm, but why should i introduce a new package for highly coupled components?

In theory i could use a different package - agree

albe:
The whole Flow Framework (and Neos Distribution) is a bunch of highly coupled packages ;)

Issue #1233, opened 2018-03-13
Add support for basic localization through routing and request parameters


albe:
### Description

Flow contains a very useful localization functionality through translations and localized resources. However, there is currently no way to switch locales per request in a straightforward way. For this, one needs to implement some mixture of custom route part handling, and a HTTP component to somehow make locales work inside an application.

I suggest to provide some easily configurable and extensible way to make use of locale selection through routing, request params and headers.

I already implemented a HTTP component to allow detecting and setting a current locale from different request parameters, like a configurable URI path part (ugh), a configurable cookie value and the Accept-Language headers (see https://gist.github.com/albe/2ffea7959d546b2b81d6a1bb37e24693). Even though that component works pretty well in itself, it does not work with the routing component, which should be responsible for deciding where in the URI the locale should be encoded and it's also not easily possible to generate URIs that encapsulate the currently selected (or a differently specified) locale.

I therefore suggest to move this further and make Locales a first class citizen in the HTTP module, so that common locale resolution methods can be enabled easily by configuration.

### Known possible ways to specify a locale in the request

1. in the hostname (subdomain or TLD, see https://github.com/neos/flow-development-collection/issues/1120)

2. in the URI (uri part, position needs to be configurable)

3. as request parameter (GET/POST, parameter name needs to be configurable)

4. as a cookie (cookie name needs to be configurable)

5. as a generic HTTP header (header name needs to be configurable)

6. from the "Accept-Language" header

### Possible implementation methods

1. Use an HTTP component like above, which is configured to follow after the routing component, and will take a locale from an matched `@locale` routing value. On the other side, extend the UriBuilder to set a default value (the current locale) for `@locale` if it is not explicitly set via arguments. That way, it's already easily possible to de-/encode a locale into URIs, easily configurable via known routing mechanics.

In order to allow the locale route parts to be optional, this would probably require a `LocaleRoutePart` handler, to prevent other route parts from being taken as the locale and routes possibly not resolving any more.

2. We could also make the Locale a first class parameter of the (Action)Request object, similar to controller and action names and the format.

3. Does the `@locale` parameter need to go into the `__referrer` on Fluid Forms?

albe:
@bwaidelich I'd especially be interested in your opinion regarding the routing part of the solution, and how that could look together with the new routing mechanism

bwaidelich:
I can't really dig into this right now, but it's an interesting topic!
In general I'd be against a "first class locale" if we can solve it with a generic implementation.

[...]

PR #1264, opened 2018-04-05
FEATURE: Support locale in routing from the request

albe:
This change is supposed to allow to define a RoutePart `@locale` in routing, that will be resolved and then in turn copied to generated URIs. This will allow to easily create a HTTP component, that will set the application locale from the routing information and/or other request parameters, like cookie or headers, which can then be used to localize the application.

Related to #1233


albe:
Ignore the test failures for now. They should be easy to fix. I'd like to get feedback, if this is something that would get support from the rest of the team.

I decided to keep everything locale to just be a string. Resolving to an actual Locale object or detecting the best matching locale via the i18n services is something that should be handled by the HTTP component.

sorenmalling:
I like the concept, of being able to bind a routing parameter to a locale. I can see a great usecase for when building "static pages" with localized content via XLF and similar (i do that in a project).

One thing that I don't know if this change will cover, is the usage in other "locale aware" context, ex. "translation services" like "TranslateViewHelper", "Or i18nServices" for when rendering a Fluid template for email etc. just to mention a few

sorenmalling:
My thought is whether we should cast it to a Locale object instead of a string, to keep the future implementation easier and non-breaking?

albe:
I intentionally avoided that to prevent coupling the MVC package to the i18n package, but I'm open to that. I would otherwise handle the conversion to `Locale` object inside the HTTP component, which also sets the i18n current locale.

sorenmalling:
Sound legit, all though the i18n package is a part of the "core" Flow package, so I don't see what we benefit from not turning it to a object right away - but I can also see the good idea of keeping it clean and lowlevel (data-wise) at that point

cwenzelg:
I was searching for exactly this solution. i would love to see this getting implemented. we would start getting this in our dev systems right away.

albe:
@sorenmalling it's just an architectural cleanness thing, because even though all packages belong to the core, we try to keep them as decoupled as possible, in order to be able to split them away when we see fit. See the few packages (Cache, Logging, etc.) already split by Christian. So if at some point we might decide to split off the i18n package, it's good to have as few interdependencies as needed.

@cwenzelg you can solve that by putting the locale optional part into your main Routes.yaml and then include your application routes with another optional part:

Routes.yaml:
```
-
  name: 'Acme Labs with locale'
  uriPattern: '({@locale}/)acme/<AcmeSubroutes>'
  routeParts:
    '@locale':
      handler: 'Flowpack\Localize\Http\LocaleRoutePart'
  subRoutes:
    'AcmeSubroutes':
      package: 'Acme.Labs'
```

Acme.Labs Routes.yaml:
```
-
  name: 'default'
  uriPattern: '{@controller}(/{@action})'
  defaults:
    '@package':    'Acme.Labs'
    '@subpackage': ''
    '@action':     'index'
    '@format':     'html'
  appendExceedingArguments: true
```

Normally, the locale is a global route parameter anyway, so that even makes sense.

albe:
So my main questions remaining are these:

- should the locale argument in the ActionRequest be an `Locale` object

- where should the LocaleRoutePart be placed? Inside i18n in the core? Without a LocaleRoutePart the above route configuration will end up with lots of invalid routings, because `@locale` will just be matched with the first path element.

- should the LocaleRoutePart be applied automatically to `@locale` route part?

@kitsunet @kdambekalns @bwaidelich maybe?

kdambekalns:
@albe What about WIP label?

bwaidelich:
I still stick to my comments at https://github.com/neos/flow-development-collection/issues/1233#issuecomment-372676767 but I won't block this of course

albe:
Yes, feedback is welcome and highly appreciated. Other than that it's mostly fixing/adding tests what is keeping this "WIP". If we can land this in 5.1 I'm happy, if not I'm not unhappy ;)

bwaidelich:
OK, a little more feedback nevertheless (: 
* This still needs documentation and tests, right?
* How will this work together with Neos? i.e. shouldn't we somehow respect the new `@locale` argument when selecting the backend/frontend language?

albe:
- Docs&Tests: yes!
- Neos: As long as it is not configured in the Routes.yaml, nothing changes really. Even then, an HttpComponent needs to fetch the `@locale` from Routing and "apply" it (make it the current i18n Locale, etc. - such a Component would be provided with a follow-up). I don't really know how Neos currently deals with the language in routes apart from it being a ContentDimension. So in my (not so deep) thinking, Neos could change it's Routes config and then use a HttpComponent to map the `@locale` to a ContentDimension. This might be totally off though, so I'd gladly work it out in a way that works for Neos

albe:
Don't close. Still a topic for me, though prioritized lower than e.g. PSR-7

albe:
Reopening to keep this on the radar - still would like to have some basic support of locale from routing in Flow

kdambekalns (commented):
Could this be updated to match the current codebase? And what is the status?

albe:
Would love to sometime rework this and finish it up. I still do think (4+ years later) it would be a worthwhile feature. But currently I don't think I can pull it off unfortunately.

PR #1266, opened 2018-04-09
BUGFIX: if cached configuration does not contain all configuration types, ensure on next run the configuration is cached again

skurfuerst:
## Problem Description

the following happens in Production context, if letting the CLI compile request run through, but abort the web request.

In this case:
- the cached Configurations.php object does *not* contain information for NodeTypes etc.
- thus, node types are loaded on every request
- this leads to a slow system which can only be fixed by removing the cached Configuration file.

To solve this issue, we now check whether the unprocessed configuration contains the right infos before serializing them.


aertmann:
Hey I've actually run into this issue recently using Neos 2.3 and found out it works fine in Neos 2.0, 2.1 & 2.2 so only broken from Flow 3.3 and up.. any chance this could be targeted at 3.3 instead of 4.0?

aertmann:
Never mind, the configuration manager was refactored for 4.0 so doesn't seem to be related.

Issue #1060, opened 2017-08-16
Support `Forwarded` Header in TrustedProxiesComponent


albe:
The TrustedProxiesComponent currently supports all forms of direct `key: value` headers for specifying the forwarded host, port and IP. Those headers, most frequently `X-Forwarded-*`, have become a defacto-standard, but a formal standard has been defined:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

The TrustedProxiesComponent should support that header.

PR #1269, opened 2018-04-09
FEATURE: Add Forwarded Header support

albe:
This adds support for setting the standardized `Forwarded` Header as described in RFC 7239 Section 4 (https://tools.ietf.org/html/rfc7239#section-4), as the `headers` trusted proxy setting.
Also, this change allows to set a single header value for the `headers`, so that working with the single `Forwarded` header is more convenient:

```
Neos:
  Flow:
    http:
      trustedProxies:
        headers: 'Forwarded'
```

Resolves: #1060


kitsunet:
Can you look at the scheme validation error? I guess otherwise @mficzel might have an idea.

albe:
Done :)

PR #1273, opened 2018-04-12
!!!TASK: Safe trusted proxies default value

albe:
By default, all proxies were trusted beforehand, but this is an usafe setting in most setups.
This change switches the `trustedProxies.proxies` setting to `%env:FLOW_HTTP_TRUSTED_PROXIES`, which means no proxies are trusted by default unless something is specified via the environment variable and hence the client IP address, port and host values for the request can not be overridden by any of the `Forwarded-*` headers.

This is breaking if you use a CDN or reverse proxy on your server and relied on the previous unsafe
default setting. In that case you should instead provide a list of the IP ranges of your proxy/CDN
servers, either directly or through the `FLOW_HTTP_TRUSTED_PROXIES` environment variable or explicitly switch back to trust all proxy servers by setting the value to `'*'`.


robertlemke:
That's a good point - we should take the opportunity to switch this behavior for 4.0.

However, I guess that there will be some hassle for certain hosting setups which because the trusted proxies need to be set by a setting. We could make this migration easier by using an environment variable as a fallback, similar to the `FLOW_CONTEXT` variable. Then a server administrator could just set this configuration for all sites and there wouldn't be any action necessary by Neos site integrators.

My suggestion would be this:

- if the setting `proxies` is set to "*", all proxies are trusted
- if the setting `proxies` is set to an empty value (null, empty array, false), `FLOW_HTTP_TRUSTED_PROXIES` will be used, and if that does not exist, no proxies are trusted
- if the setting `proxies` is set to a non-empty array, only these values are used

The default value of `proxies` would be changed to `[]`.

The new `FLOW_HTTP_TRUSTED_PROXIES` variable would be retrieved via `Bootstrap::getEnvironmentConfigurationSetting()`, like `FLOW_CONTEXT`.

I looked into creating that change myself, but I guess that including test setup and adjusting tests that would take about an hour, which I can't invest today or tomorrow. So if anyone likes the idea, please go ahead!

kitsunet:
Why wouldn't we just make the default `proxies: %env:FLOW_HTTP_TRUSTED_PROXIES%` ? That doesn't need further processing at any other point.

albe:
I like the idea to set via environment variable, though I'd prefer having it explicit instead of "empty value means use a hardcoded ENV VAR". Also fine with making it the default value.

How would that behave though, if `FLOW_HTTP_TRUSTED_PROXIES` is not set? Then that value would be emptyish, so that would need to be another option for specifying "Trust no proxies", right?

Also, ist it possible to set an array value via environment variable, or would that need further string-to-array processing (`explode`/TypeConverter)?

I can take care if the above questions are answered.

kitsunet:
According to php docs `getenv` returns FALSE if the env was not set. So that could work out well?

albe:
Yeah. Just need to change the `=== []` check against `empty()`. What about the array through ENV?

albe:
k, then I'll go with comma separated for now, as that matches what is possible with the headers settings already.

PR #1275, opened 2018-04-14
FEATURE: Allow specifying validation groups for Collection and Entity validators

### LINKED ISSUE #1008: Partial validation not working as described
A question regarding expected behavior of partial validation. According to https://flowframework.readthedocs.io/en/4.0/TheDefinitiveGuide/PartIII/Validation.html#advanced-feature-partial-validation it is possible to restrict validation of a property to a specific group.

I was trying to do what the example for `$prop4` shows, but for a `Collection`. That didn't work, because… https://github.com/neos/flow-development-collection/blob/72537e845d8dde10f7d9ff5a8b270c234ba1ab07/Neos.Flow/Classes/Validation/ValidatorResolver.php#L305-L313 kicks in unconditionally, before any `Validation` annotation is even considered in https://github.com/neos/flow-development-collection/blob/72537e845d8dde10f7d9ff5a8b270c234ba1ab07/Neos.Flow/Classes/Validation/ValidatorResolver.php#L315-L326

Q: Is this really consciously intended behavior?

albe:
This makes it possible to explicitly specify a `Collection` or `GenericObject` type validator on according properties in order to override the options, like validation groups. This is useful for example, when you want to prevent a collection of entities or a whole model to be validated in some validation groups. Previously, you would have had to exclude all the properties of the related entities via according validate annotations with specified validation groups.

Resolves #1008


daniellienert:
@albe: Had a look through the code. As fa as I understood it (never have been in this corner of flow before) it looks good. But the tests failing seem related. Would you have a look?

albe:
Yes, I was already at it. I stumbled over some oddities and will need to take another look.

kdambekalns:
@albe, did you have that other look you wanted to take in May?

kdambekalns:
FYI: Looking into the test failure right now…

kdambekalns:
The validation does not fail, because the collection argument is lost during transformation from raw value to mapped value in `Argument::setValue()`. Investigating…

albe:
Thanks for looking into this. That's a good catch. I'll try to look into it and find a solution to this.

kdambekalns:
Good luck. I tried all sorts of property mapping configuration in the test controller, but had no luck.

albe:
Found the problem. It is not the PropertyMapper. It's in the CollectionValidator where Results are merged and I have little clue what is really going on there. Actually the fix is so weird that I start believing there is something else wrong with the implementation of merging results. Will push the fix later.

albe:
Okay, this is bigger than I thought and here is the deal what's going on:

- in `CollectionValidator` we currently do this:
```php
$this->result->forProperty($index)->merge($collectionElementValidator->validate($collectionElement));
```
  This leads to the SubResult for the index being created **before** calling the validator on the element. The Element itself in this case is of the same type as the base class (cyclic relation), hence it will call **the same instance of the Validator** again (see below for explanation), overwriting `$this->result` and boom - the result of the failing validation is lost.

- this can be worked around by instead doing:
```php
$propertyResult = $collectionElementValidator->validate($collectionElement);
$this->result->forProperty($index)->merge($propertyResult);
```
This will validate the element first, and then merge results into the subProperty result. However, this only works, because there are no other element results that are overwritten/lost in this test beforehand. This only hides the bug for this specific case.

So the root issue is, that the CollectionValidator is reused (or more correctly, that Validators just overwrite the `$this->result` instance variable on every call) - even though it is not of singleton scope. Why is this? The answer is simply: because we build Validator chains for classes, not instances.

So what are the possible solutions?

- we completely change the way we build Validator chains to be based on instances (I think there was already an issue for that way back). However, this is a major (most likely breaking) change and far out of scope.

- we avoid using and overwriting instance variables for storing validator results (at least for Collection and Object Validators - anything that can be called in a cycle). Either by using function scope variables inside the `validate()` method (how to deal with all those functions depending on `$this->result` though?) or keeping a stack of results and pushing/poping (more b/c since we can keep `$this->result` as a pointer to the head of the stack). This goes down back to `AbstractValidator` though and might still probably break dozens of custom validator implementations.

- we just run with the workaround for now and solve the issue at hand separately with an API change with deprecation now and final fix in next major.

Further advice highly required @kdambekalns

albe:
See #1369 for a possible fix with the stack solution. If it's accepted, this should depend (be rebased) on it.

kdambekalns:
Wait, so it passes now, just because the element type annotation was added? But… I tried that locally, and it did _not_ help…

albe:
It was failing for me with those applied too, but for the reasons mentioned/fixed with #1369 so... huh? :confused:

kdambekalns:
But the tests pass for me now, too… 😕🤷‍♂️

albe:
IIRC that were the tests that I used in the end (also the ones I use in the bugfix PR). Maybe I didn't push those, but I'm not entirely sure why those should fail and the previous not.

albe:
So this just needs a rebase once up-merges have been done and then tests should pass again.

albe:
Yay, tests passing :tada:

kdambekalns:
Hm… I'd say we can change the target branch for this one to be 5.1 (which I branched while the tests for this one were still running…)

PR #1276, opened 2018-04-15
FEATURE: Arrow function expressions in Eel

hlubek:
Adds support to parse and evaluate arrow function expressions in Eel.
This is only implemented for the CompilingEelEvaluator, since
expression evaluation need to be deferred in functions.

Arrow function arguments can be specified with or without parentheses.
The body must be an expression and not a block.

Example:

    map(items, (x) => x * x)

### DIFF
--- a/Neos.Eel/Resources/Private/Grammar/Eel.peg.inc
+++ b/Neos.Eel/Resources/Private/Grammar/Eel.peg.inc
@@ -33,7 +33,11 @@ Term: term:BooleanLiteral !Identifier | term:NumberLiteral | term:StringLiteral 
 
 /*!* CombinedExpressions
 Expression: exp:ConditionalExpression
-SimpleExpression: term:WrappedExpression | term:NotExpression | term:ArrayLiteral | term:ObjectLiteral | term:Term
+SimpleExpression: term:ArrowFunction | term:WrappedExpression | term:NotExpression | term:ArrayLiteral | term:ObjectLiteral | term:Term
+ArrowFunction: arguments:MethodArguments < '=>' > exp:Expression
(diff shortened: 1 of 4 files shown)


skurfuerst:
hahaha really cool :+1: 👍

dfeyer:
Nice one we need to have this documented somewhere, right ?

Working a lots with Elixir recently and they have a feature name "function capturing" for a more compact (and alternative, they support also the full function declaration) syntax:

     map(items, &(&1 * &1))

Looks geeky ;) But it's really nice to work with. 

Next step

   items |> map(&(&1 * &1))

And we have a new shiny functional language ;)

Neos.Eel/Resources/Private/Grammar/Eel.peg.inc
```diff
@@ -33,7 +33,11 @@ Term: term:BooleanLiteral !Identifier | term:NumberLiteral | term:StringLiteral
 
 /*!* CombinedExpressions
 Expression: exp:ConditionalExpression
-SimpleExpression: term:WrappedExpression | term:NotExpression | term:ArrayLiteral | term:ObjectLiteral | term:Term
+SimpleExpression: term:ArrowFunction | term:WrappedExpression | term:NotExpression | term:ArrayLiteral | term:ObjectLiteral | term:Term
+ArrowFunction: arguments:MethodArguments < '=>' > exp:Expression
+MethodArguments: arguments:MethodArgumentsWithParens | arguments:MethodArgumentsWithoutParens
+MethodArgumentsWithParens: '(' < Identifier? (< ',' > Identifier)* > ')'
+MethodArgumentsWithoutParens: Identifier
```
daniellienert: "Parens" sounds a bit strange, I would write the full word

PR #1279, opened 2018-04-18
BUGFIX: Object Management for classes with underscores

bwaidelich:
This is a new try to fix support for Underscore_Classes in Flow.
It is a follow-up to #655, this time targeted to master because
of its risk to break things.

Resolves: #654


bwaidelich:
Unfortunately tests still fail with this one:
```
Required class "Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\SubNamespace\SubNamespace_Underscores" could not be loaded properly for reflection.
```

kitsunet:
Pushed some update, BUT it still totally fails locally. And I haven't got a clue yet why.

kitsunet:
And found it :) This should hopefully make travis happy and should finally fix this

kitsunet:
Ok, next step, now the class name and file should be in the right place according to PSR-0 finally. But the reflection service does something fishy and tries to load `Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\SubNamespace\Underscores` which is not surprising after all. We are _almost_ back to square one. But it shows again, that looking at files and guessing what class is in there is a bad idea for these edge cases...

kitsunet:
Oh and obviously even IF everything worked right (which I doubt) we coudln't do this in our tests as the tests paths are registered as PSR-4 so underscores are not even allowed like in PSR-0

kitsunet:
Ok, I will revert the test class to be a PSR-4 class with underscores that are not special anymore. That _might_ maybe work. We probably won't get PSR-0 classes with underscore to pass our reflection and proxy building due to the way we collect class files...

kitsunet:
Ok, this seems to work fine locally...

kitsunet:
I don't get it. I don't have the errors travis gets.

bwaidelich:
Mh, this is weird. I do get the same errors as travis when running the Flow functional tests

bwaidelich:
This is a candidate for the next major IMO

bwaidelich:
I'll take another look now!

bwaidelich:
OK, I rebased the branch to the current `master` but still get the following error when running functional tests:
```
Required class "Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\SubNamespace_Underscores" could not be loaded properly for reflection.
```

And the following doesn't work, either:
```
composer require google/apiclient:"^2.0"
```
with the following `Settings.yaml`:
```yaml
Neos:
  Flow:
    object:
      includeClasses:
        'google.apiclient': ['.*']
```

```php
class SomeClass
{
    /**
     * @Flow\Inject
     * @var \Google_Client
     */
    protected $apiClient;
}
```

=>

```
The object "Google_Client" which was specified as a property in the object configuration of object "Wwwision\CR\Command\TestCommandController" (automatically registered class) does not exist.
```

@kitsunet I'm a little confused now: Would you say the class schema of https://github.com/googleapis/google-api-php-client is wrong (according to PSR-0) or do we just not support this kind of "esotheric" schema?

albe:
Formally, google-api-client is not PSR-0 compliant, since that states that every class MUST have a toplevel namespace ("Vendor Name") - see https://www.php-fig.org/psr/psr-0/
Our own ClassLoader currently has no case for a class without a namespace part, so it will only attempt the default load locations from `fallbackClassPaths`, which are only filled with PSR-4 paths.

Also `Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\SubNamespace_Underscores` should live somewhere in `{psr-basefolder}/SubNamespace/Underscores.php` according to PSR-0, but in `{psr-basefolder}/SubNamespace_Underscores.php` according to PSR-4. Since we provide the `Neos\Flow\Tests` Namespace base in `autoload-dev` it should work?

However, our own ClassLoader always splits the classname by `_` and adds those parts to the namespace parts (PSR-0 like), see https://github.com/bwaidelich/flow-development-collection/blob/654-objectmanagement-for-classes-with-underscores/Neos.Flow/Classes/Core/ClassLoader.php#L122

kdambekalns:
I am confused, TBH. Do we still need/want this? What is the status?

bwaidelich:
> I am confused, TBH.

Me too :)

> Do we still need/want this? What is the status?

I still think that it would be good not to reserve the underscore. But I couldn't get it to work (at least not as easily as I had hoped for) so the status is: stale.. at the moment

albe:
With our own class loader being technically deprecated (needs to be effectively deprecated still - see https://github.com/neos/flow-development-collection/pull/925 and a missing follow-up), I guess we can close this, since the composer autoloader should not have this issue, right?

kdambekalns:
LGTC ;)

PR #1280, opened 2018-04-18
!!! TASK: Cleanup in package management

kitsunet:
This is the next step towards a leaner package management,
the essential part is that packages are now separated into
third party packages and Flow (enabled) packages.
All packages are available for object management but Resources
and Configuration as well as booting are only expected and
managed in Flow (enabled) packages.

The ``Package`` class is still a fully Flow enabled package and
no adaption should be necessary to packages.
``GenericPackage`` is the low level class for describing any
other package in the system.
According to that change a couple of interfaces where added:

* ``BootableInterface`` describes a bootable package
* ``FlowPackageInterface`` extension of the ``PackageInterface``
  Flow specifics are now moved over to the ``FlowPackageInterface``
* ``PackageKeyAwareInterface`` defines that the package has a
  package key. Currently that is implemented by all packages but
  we might change that at a later point in time.

The notion of ``protected`` and ``objectManagementEnabled`` is gone from
package classes and the interfaces as both are no longer needed.

The ``PackageManager`` and interface no longer support deleting of
packages, this should happen through composer now.
In order the ``package:delete`` command is removed as well.

``PackageManager::getPackageVersion`` was moved over to
``ComposerUtility::getPackageVersion`` where it should have been in the
first place. It was not part of the interface nor marked api.

This change is breaking if you use the ``PackageManager`` to get
all packages and expect them to be Flow packages. You must now
check for instanceof ``FlowPackageInterface`` if you expect Flow
specific functionality from a package object.

This is also a preparation to drop the ``PackageManagerInterface`` as
overriding the implementation is neither possible nor sensible.
You can directly use the ``PackageManager`` object from now.


kitsunet:
Grew bigger than expected... But this is the halfway compromise to removing all third party packages from object management. Probably much nicer for the future of slow refactoring of this. Most of it is shuffling code around. Not much has been added, a bit removed though.

Neos.Flow/Classes/Package/BootableInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Core\Bootstrap;
+
+/**
+ * An interface to indicate a bootable package
+ */
+interface BootableInterface
```
bwaidelich: &lt;nitpick&gt;I'd go for `BootablePackageInterface`&lt;/nitpick&gt;

Neos.Flow/Classes/Package/GenericPackage.php
```diff
@@ -0,0 +1,321 @@
...
+ */
+
+use Neos\Flow\Composer\ComposerUtility;
+use Neos\Flow\Core\ClassLoader;
+use Neos\Utility\Files;
+
+/**
+ * The generic base package that represents third party packages
+ */
+class GenericPackage implements PackageInterface, PackageKeyAwareInterface
```
bwaidelich: I'm not so sure about the name `GenericPackage` to be honest.. But `ComposerPackage` is not better, is it?
Then again, this is pretty low-level so it doesn't _really_ matter I guess

Neos.Flow/Classes/Command/PackageCommandController.php
```diff
@@ -99,27 +99,6 @@ public function createCommand(string $packageKey, string $packageType = PackageI
...
-        if (!$this->packageManager->isPackageAvailable($packageKey)) {
-            $this->outputLine('The package "%s" does not exist.', [$packageKey]);
-            $this->quit(1);
-        }
-        $this->packageManager->deletePackage($packageKey);
-        $this->outputLine('Deleted package "%s".', [$packageKey]);
-        Scripts::executeCommand('neos.flow:cache:flush', $this->settings, false);
-        $this->sendAndExit(0);
-    }
-
```
bwaidelich: The `PackageCommandController` now complains about missing methods in `createCommand()` and `listCommand()`

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -477,12 +480,12 @@ public static function initializeSystemFileMonitor(Bootstrap $bootstrap)
...
 
         $context = $bootstrap->getContext();
-        /** @var PackageManagerInterface $packageManager */
+        /** @var PackageManager $packageManager */
         $packageManager = $bootstrap->getEarlyInstance(PackageManagerInterface::class);
         $packagesWithConfiguredObjects = static::getListOfPackagesWithConfiguredObjects($bootstrap);
 
         /** @var PackageInterface $package */
-        foreach ($packageManager->getAvailablePackages() as $packageKey => $package) {
+        foreach ($packageManager->getFlowPackages() as $packageKey => $package) {
```
bwaidelich: In the following lines we refer to `getConfigurationPath()`, `getResourcesPath()` and `getAutoloadPaths()` but they are not part of the `PackageInterface` any longer. The latter isn't even part of the `FlowPackageInterface`

Neos.Flow/Classes/I18n/Service.php
```diff
@@ -293,7 +293,7 @@ protected function generateAvailableLocalesCollectionByScanningFilesystem()
         $blacklistPattern = $this->getScanBlacklistPattern();
 
         /** @var PackageInterface $activePackage */
-        foreach ($this->packageManager->getAvailablePackages() as $activePackage) {
+        foreach ($this->packageManager->getFlowPackages() as $activePackage) {
             $packageResourcesPath = Files::getNormalizedPath($activePackage->getResourcesPath());
```
bwaidelich: `getResourcesPath()` is not part of the `PackageInterface` any longer

Neos.Flow/Classes/I18n/Xliff/Service/XliffFileProvider.php
```diff
@@ -18,7 +18,7 @@
 use Neos\Flow\I18n\Xliff\Model\FileAdapter;
 use Neos\Flow\I18n\Xliff\V12\XliffParser as V12XliffParser;
 use Neos\Flow\Package\PackageInterface;
```
bwaidelich: &lt;nitpick&gt;this import is not used any longer&gt;/nitpick&gt;

Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
```diff
@@ -211,15 +212,18 @@ protected function registerClassFiles(array $packages)
...
-            if ($package->isObjectManagementEnabled() && (ComposerUtility::isFlowPackageType($package->getComposerManifest('type')) || isset($includeClassesConfiguration[$packageKey]))) {
+            if (ComposerUtility::isFlowPackageType($package->getComposerManifest('type')) || isset($includeClassesConfiguration[$packageKey])) {
                 foreach ($package->getClassFiles() as $fullClassName => $path) {
                     if (substr($fullClassName, -9, 9) !== 'Exception') {
                         $availableClassNames[$packageKey][] = $fullClassName;
                     }
                 }
-                if (isset($this->allSettings['Neos']['Flow']['object']['registerFunctionalTestClasses']) && $this->allSettings['Neos']['Flow']['object']['registerFunctionalTestClasses'] === true) {
+                if ($package instanceof FlowPackageInterface && $shouldRegisterFunctionalTestClasses) {
                     foreach ($package->getFunctionalTestsClassFiles() as $fullClassName => $path) {
```
bwaidelich: `getFunctionalTestsClassFiles()` is no longer part of the `PackageInterface`

Neos.Flow/Classes/Package/PackageManager.php
```diff
@@ -654,28 +650,6 @@ protected function preparePackageStateConfiguration($packageKey, $packagePath, $
...
-                return preg_replace('/^v([0-9])/', '$1', $composerLockData['version'], 1);
-            }
-        }
-
-        return '';
-    }
-
     /**
      * Requires and registers all packages which were defined in packageStatesConfiguration
      *
```
bwaidelich: The `PackageManager` refers to the no longer existing `PackageInterface::getPackageKey()`  in `unregisterPackage()` and `registerPackageFromStateConfiguration()`

Neos.Flow/Classes/ResourceManagement/Storage/PackageStorage.php
```diff
@@ -159,8 +158,7 @@ public function getStreamByResource(PersistentResource $resource)
     public function getPublicResourcePaths()
     {
         $paths = [];
-        $packages = $this->packageManager->getAvailablePackages();
-        foreach ($packages as $packageKey => $package) {
+        foreach ($this->packageManager->getFlowPackages() as $packageKey => $package) {
             /** @var PackageInterface $package */
             $publicResourcesPath = Files::concatenatePaths([$package->getResourcesPath(), 'Public']);
```
bwaidelich: `getResourcesPath()` is no longer part of the `PackageInterface`

Neos.Flow/Classes/Command/SchemaCommandController.php
```diff
@@ -49,7 +49,7 @@ class SchemaCommandController extends CommandController
      * @param boolean $verbose if TRUE, output more verbose information on the schema files which were used
      * @return void
      */
-    public function validateCommand(string $configurationFile = null, string $schemaFile = 'resource://Neos.Utility.Schema/Private/Schema/Schema.schema.yaml', bool $verbose = false)
+    public function validateCommand(string $configurationFile = null, string $schemaFile = 'resource://Neos.Flow/Private/Schema/Schema.schema.yaml', bool $verbose = false)
```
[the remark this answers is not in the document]
kitsunet: Ofc :) The Schema utility is a third party package as far as Flow is concerned, so the Resource file had to be moved to Flow (it's only used in Flow anyway).

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -173,6 +173,7 @@ public static function initializePackageManagement(Bootstrap $bootstrap)
     {
         $packageManager = new PackageManager(PackageManager::DEFAULT_PACKAGE_INFORMATION_CACHE_FILEPATH, FLOW_PATH_PACKAGES);
         $bootstrap->setEarlyInstance(PackageManagerInterface::class, $packageManager);
+        $bootstrap->setEarlyInstance(PackageManager::class, $packageManager);
```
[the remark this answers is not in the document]
kitsunet: Until the follow up change that removes the interface, yes. Didn't want to do that in this change as well, otherwise it would've been even bigger.

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -199,12 +200,13 @@ public static function initializeConfiguration(Bootstrap $bootstrap)
         $environment->setTemporaryDirectoryBase(FLOW_PATH_TEMPORARY_BASE);
         $bootstrap->setEarlyInstance(Environment::class, $environment);
 
+        /** @var PackageManager $packageManager */
```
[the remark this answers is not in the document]
kitsunet: Mmmm, I would do that in the change that removes the interface. Again I wanted to scope this change as much as possible to the actual package stuff.

Neos.Flow/Classes/Package/BootableInterface.php
kitsunet: Yeah, can change if you want.

Neos.Flow/Classes/Package/GenericPackage.php
kitsunet: I was almost creating `ComposerPackage` but then I would want to introdce something in between as well, because a composer package shouldn't know about package keys IMHO.

Neos.Flow/Classes/Command/PackageCommandController.php
kitsunet: oh? ok, will have a look.

Neos.Flow/Classes/Core/Booting/Scripts.php
kitsunet: True, the docblock is bad.

Neos.Flow/Classes/I18n/Service.php
kitsunet: True, the docblock is bad.

Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
kitsunet: `if ($package instanceof FlowPackageInterface && `

Neos.Flow/Classes/Package/PackageManager.php
kitsunet: Will fix.

bwaidelich:
Still reviewing this one and I can't fully wrap my head around the new package inheritance

image
(http://yuml.me/edit/b9c53533)

kitsunet:
The idea is, that GenericPackage is `PackageKeyAwareInterface` and `PackageInterface`, while `Package` is all the interfaces....

bwaidelich:
But in that case, wouldn't this make more sense:
image

And since all Flow packages are PackageKeyAware and Bootable it could be simplified to

image

or am I missing something here?

kitsunet:
No, that could work. Will add it to the rest of the changes.

bwaidelich:
Recap (after talking to Christian):
My diagram wasn't entirely correct. The current structure is more like this:

image

And it makes sense because

1. We might want to have "non-bootable" Flow packages at some point
2. We want to get rid of `GenericPackage::getPackageKey()`, removing it from the `PackageInterface` is a first step in that direction

Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
kitsunet: Added it to the FlowPackageInterface for now as `@internal`

Neos.Flow/Classes/Package/PackageManager.php
kitsunet: Turns out `unregisterPackage` was unused and fixed the other by adjusting the return value of the PackageFactory.

kitsunet:
@bwaidelich should have fixed the comments that I could.

bwaidelich:
@kitsunet awesome. I removed my -1 and will continue the review/test tomorrow!

kitsunet:
@bwaidelich as discussed earlier I would merge this now and we can see if anything is missing. IMHO it should be all fine...

kaystrobach:
this BC actually made it impossible to have generic fluid template packages, which could be used in typo3 and flow.

is there any way to have a package of type library working with a Views.yaml like:

```yaml
-
  requestFilter: 'isPackage("Vendor.Package") && isFormat("html") && !isController("Controller")'
  options:
    layoutRootPaths:
      'vendor.layout/Layouts':       'resource://vendor.layout/Private/Layouts/'
      'Vendor.Package/Layouts':   'resource://Vendor.Package/Private/Layouts/'
```

PR #1282, opened 2018-04-19
FEATURE: Allow specifying a list of available Locales via settings

### LINKED ISSUE #1232: Available Locale Detection is finding too many invalid locales

albe:
With this, it is possible to specify a list of available Locales via the
`Neos.Flow.i18n.availableLocales` setting, which will then avoid triggering
the scanning process.

Resolves #1232


albe:
Simply rebased to get rid of the "Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes" test failure

PR #1283, opened 2018-04-19
TASK: Object configuration skips objects with autowiring false

kitsunet:
As constructor arguments and injections should not be automatically
resolved when either the whole class or the respective method has
autowiring false, we should skip any processing of those as early
as possible to avoid any unnecessary or even wrong checks for
possible autowiring

This should improve compile performance slightly and prevent errors
with constructors of singletons with autowiring false.
Those would previously throw an exception due to non autowired
constructor arguments..


kitsunet:
The error here: https://travis-ci.org/neos/flow-development-collection/jobs/365256489
is fixed by this if we disable autowiring for the bootstrap (which should be done anyway).

kitsunet:
@kdambekalns we talked about this.

PR #1286, opened 2018-04-24
BUGFIX: Render debugger-styles directly with first debug-output

mficzel:
The styles for rendering the debug-output are now rendered with the
first output itself instead of being echoed.

The direct echo solution caused trouble whenever non html-data like
json was rendered because the styles were rendered outside the json-structure 
and thus caused invalid json.


beheist:
Yes, love it!

PR #1288, opened 2018-04-24
Add setArguments function

dimitriadisg:
Right now, it's only possible to update the URI query with a string, through setQuery. This change does the respective changes to the query and arguments by setting the arguments instead.

**What I did**
Added a setter for the Uri arguments property

**How I did it**
In the same manner as the setQuery method

**How to verify it**
Use setArguments instead of setQuery, the result will be the same

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Http/Uri.php
+++ b/Neos.Flow/Classes/Http/Uri.php
@@ -337,6 +337,19 @@ class Uri implements UriInterface
     {
         return $this->arguments;
     }
+    
+    /**
+     * Sets the URI's arguments. Updates (= overwrites) the query accordingly!
+     *
+     * @param array $arguments The query arguments.
+     * @return void
+     * @api
+     */
+    public function setArguments($arguments)
+    {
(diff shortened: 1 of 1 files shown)


kitsunet:
Thanks for your first contribution! I can totally understand the wish for this method, and maybe we might add something similar. But I would refrain from adding more mutating methods as the general direction would be to follow PSR-7 only and work with non mutating methods only (so that you return a new changed instance - see withQuery method for example). Therefore downvoting this for now to avoid merge without good disucssion.

albe:
I'm with Christian there. Make this a non-mutating API.

albe:
So in light of the further PSR-7 adoption, which does not include this method in the UriInterface, I'd close this. We won't have this method at this place and also deprecated `getArguments()`. Sorry for this @dimitriadisg, but that decision was not easy as you can see in the according PR and the discuss post. We'd prefer to stick with our own HTTP Api, but there is too much to gain with adopting the standard.

PR #1292, opened 2018-04-24
BUGFIX: !!!Reduce varchar size from 255 to 191 for utf8mb4 support

gerhard-boden:
This is not a serious proposal for solution, but just a hack. 
This change here would be the worst case scenario how we fix it.

Here's a brief description of the problem:
https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes

Alternatives are very welcome 😃


kitsunet:
Nope, that's why we increased the minimum required MySql and MariaDB versions because from maria 10.2.2 and Mysql 5.7.something this is fixed on the DB side.

PR #1293, opened 2018-04-25
Revert "!!!TASK: Implicit joins in a query make result distinct"

kitsunet:


kitsunet:
Will merge this latest in about 1 hour from now unless I hear better alternatives...

PR #1300, opened 2018-04-26
BUGFIX: Fix broken cache get/set in RedisBackend

kdambekalns:
The type hints broke correct detection of hits/misses in the cache
backend.

### DIFF
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -456,22 +456,20 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
     }
 
     /**
-     * TODO: No return type declaration for now, as it needs to return false as well.
-     * @param string $value
-     * @return mixed
+     * @param string|bool $value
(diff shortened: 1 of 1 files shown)


Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -456,22 +456,20 @@ public function setRedis(\Redis $redis = null)
...
-     * @param string $value
-     * @return mixed
+     * @param string|bool $value
+     * @return string|bool
      */
-    private function uncompress(string $value): string
+    private function uncompress($value)
     {
-        if (empty($value)) {
+        if ($value === false || empty($value)) {
```
kitsunet: `empty(false) === true` therefore this addtional check is unnecessary.

kdambekalns: Yes, but it clearly transports the intention and does away with implicit type handling. Thinking about it, this should probably even be `if ($value === false || $value === '') {} elseif (!is_string($value) { throw new \InvalidArgumentException(); }`

PR #1301, opened 2018-04-28
FEATURE: Extends Eel DateHelper with machine-readable formats

putzwasser:
Adds methods which format a DateTime into machine-readable formats to use it e.g. in the `datetime` attribute of a `<time>` tag.

* htmlDateTime formats a DateTime into the W3C format "Y-m-d\TH:i:sP"
* htmlDate formats the date part of a DateTime into "Y-m-D"
* htmlTime formats the time part of a DateTime into "H:i:sP"

### DIFF
--- a/Neos.Eel/Classes/Helper/DateHelper.php
+++ b/Neos.Eel/Classes/Helper/DateHelper.php
@@ -250,6 +250,39 @@ class DateHelper implements ProtectedContextAwareInterface
     }
 
     /**
+     * Format a given datetime in a machine-readable way to use it in a <time> tag
+     *
+     * @param \DateTimeInterface $dateTime
+     * @return string
+     */
+    public function htmlDateTime(\DateTimeInterface $dateTime)
+    {
+        return $dateTime->format(\DateTime::W3C);
+    }
+
+    /**
+     * Format a given date in a machine-readable way to use it in a <time> tag
+     *
+     * @param \DateTimeInterface $dateTime
+     * @return string
+     */
+    public function htmlDate(\DateTimeInterface $date)
+    {
+        return $date->format('Y-m-d');
+    }
+
+    /**
+     * Format a given time in a machine-readable way to use it in a <time> tag
+     *
+     * @param \DateTimeInterface $dateTime
+     * @return string
+     */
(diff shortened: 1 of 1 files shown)


Neos.Eel/Classes/Helper/DateHelper.php
```diff
@@ -249,6 +249,39 @@ public function second(\DateTimeInterface $dateTime)
...
 
+    /**
+     * Format a given datetime in a machine-readable way to use it in a <time> tag
+     *
+     * @param \DateTimeInterface $dateTime
+     * @return string
+     */
+    public function htmlDateTime(\DateTimeInterface $dateTime)
+    {
+        return $dateTime->format(\DateTime::W3C);
```
bwaidelich: Is there a particular reason why you chose the `\DateTime:W3C` format?
I know that we use this as default in the `DateTimeConverter` but this was a mistake IMO because it lacks some crucial features.

bwaidelich:
@putzwasser Thanks a lot and congratulations to your first Pull Request!
I'm a bit skeptical though to be honest because this seem to use rather random formats (unless I miss something here).
Did you know that you can achieve the above already by using the 2nd argument of the `format` helper?
```
timeTag = Neos.Fusion:Tag {
  tagName = 'time'
  attributes.datetime = ${Date.format(someDate, 'Y-m-d\TH:i:sO')}
  content = ${Date.format(someDate, 'Y-m-d')}
}
```
(untested)

putzwasser:
Thanks for your comment.
Actually, the formats are not random. The output can be used in [time tags](https://www.w3schools.com/tags/tag_time.asp)—or the [datetime attribute](https://www.w3schools.com/tags/att_time_datetime.asp) to be more precise. Basically they are shorthands to get machine readable datetime/date/time formats.
So, yes one could simply use `Date.format` but since there are shorthands for `hour/minute/second` i figured this could be handy because `Y-m-d\TH:i:sO` is easy to be confused with `d/m/Y\TH:i:sO` and other possibilities to format a date.
At least it's easier to confuse than e.g. `Date.format('H')` and this has the `Date.hour(datetime)` shortcut ;)

PR #1308, opened 2018-05-02
TASK: Improve functional test persistence tearDown()

kdambekalns:
This adds `forgetInstance` calls to the `tearDown()` in the Flow
`FunctionalTestCase` that avoid side effects between tests. One sympto this
fixes is failing tests caused by "The entity manager is closed" when a
preceeding test produced an exception during `persistAll()`.

### DIFF
--- a/Neos.Flow/Tests/FunctionalTestCase.php
+++ b/Neos.Flow/Tests/FunctionalTestCase.php
@@ -11,6 +11,7 @@ namespace Neos\Flow\Tests;
  * source code.
  */
 
+use Doctrine\ORM\EntityManagerInterface;
 use Neos\Flow\Configuration\ConfigurationManager;
 use Neos\Flow\Core\Bootstrap;
 use Neos\Flow\Annotations as Flow;
@@ -20,6 +21,7 @@ use Neos\Flow\Mvc\ActionRequest;
 use Neos\Flow\Mvc\Routing\Dto\RouteParameters;
 use Neos\Flow\Mvc\Routing\Dto\RouteContext;
 use Neos\Flow\Mvc\Routing\Route;
+use Neos\Flow\Persistence\PersistenceManagerInterface;
 use Neos\Utility\Arrays;
 use Neos\Utility\Files;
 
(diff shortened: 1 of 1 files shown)


kdambekalns:
What I'd love to have feedback/input on:

- can we remove more/other cleanup with this?
- should this go into this, earlier or the master branch only?
- why didn't we do it like this from the start?

kdambekalns:
Huh, failures… Two things may be needed: One of the failures might actually rely on the EntityManager not being cleaned up properly. And the "too many connections" could be caused by a new connection per test, but existing ones are not properly closed because the "reset" is a bit too simplistic.

kitsunet:
Sounds like a good idea to me. I guess closing the connection is a good idea though. Maybe it was just not done for performance reasons?

kdambekalns:
I assume no closing was needed since the EM was not thrown away, and thus no new connections were created…

kitsunet:
Right, but with this change it's needed or do I miss something?

kdambekalns:
It seems as if it's needed, yes. :)

albe:
> - can we remove more/other cleanup with this?

I couldn't find anything on first look. All persistenceManager clearing is done within tests to simulate two requests.

> - should this go into this, earlier or the master branch only?

IMO earlier, cleaning up properly is a bugfix.

> - why didn't we do it like this from the start?

No clue.

Torsten85:
Correct me if I'm wrong but a lot of objects hold a reference to the `PersistenceManager` and relying on the singleton pattern to ensure that all references pointing to the same object. "Forgetting" the `PersistenceManager` (or any singleton) would mean, that in every test a new instance would be created and you end up with multiple PersistenceManagers. Due to the statefulness this could lead to a lot of other problems.

albe:
Stumbling upon this again and I think it's still a good thing to get more robust functional tests. Regarding the valid concerns Torsten raised:
Indeed that would be an issue and break more things than this solves. The case here basically is some Singleton that has the PersistenceManager/EntityManager injected, and hence ends up holding a reference to another instance than the one supposed for the current test run.

So I think we need to be more radical here and do an `forgetAll`. That way only custom Singletons that themself hold a reference to one of the Flow Singletons would break, but then again, this is really a bad use case that I have no issue with.

Only downside is the (highly?) decreased performance for the functional tests.

PR #1312, opened 2018-05-14
BUGFIX: Fix distinction between PSR and legacy loggers in the LoggerFactory

beheist:
As discussed[1] - set third parameter to "true" to do the correct comparison.
If this parameter is set to true, a class name string as "object" is allowed.

[1] https://neos-project.slack.com/archives/C04PYL8H3/p1526288487000147


bwaidelich:
Can you add some more background information or clearify what "as discussed" means in this case? ;)

PR #1317, opened 2018-05-23
FEATURE: Add pkpass as media type

lennartzellmer:
I added the pkpass (Apple Wallet file) media type to be recognized by Neos.
This is especially necessary if you push content to S3 Storage / Object Storage have it delivered directly.

To achive this the file ending and media type have been added to the Neos.Utility.MediaTypes mappings.


aertmann:
looks good, however could possibly go into a stable release as a TASK instead to make it useful immediately

kitsunet:
Agreed, that can totally go into an older version IMHO.

lennartzellmer:
Yes! That would be very useful.
Should I make another pull request for the 5.0 branch then? Or do you think it should go back further?

albe:
Needs a new PR and it should IMO target 4.3 as of now.

How about adding the format to the script if it's not yet included in the mediatypes downloaded from apache? See #1364 for how I would imagine it.

albe:
I guess we can close this with #1364 merged.

PR #1323, opened 2018-06-12
FEATURE: Add map, reduce, filter, some, every functional helpers to ArrayHelper

hlubek:
**What I did**

Implemented `map`, `reduce`, `filter`, `some` and `every` according to the JavaScript specs as functional Eel helpers in ArrayHelper.

**How to verify it**

Run unit tests for ArrayHelper.

**Checklist**


hlubek:
We should also add `Array.filter`, `Array.some` and `Array.every` for completeness.

hlubek:
_Ping_

Would be nice to get this in to have better functional Eel support in Fusion.

kitsunet:
yep, can do, but I would like to actually NOT add the type hints to all existing methods. We postponed it for last major because it felt to breaking. Via Fusion we figured some stuff might come in that is not exactly type safe so we didn't merge the change that adds type hints to all helpers. But I gladly merge this with just the new code.

albe:
Can we get this done for 5.1?

hlubek:
@kitsunet Alright, let's do it without the type hints (which could lead to BC issues).

hlubek:
Alright, will push a follow-up ASAP

hlubek:
@jonnitto We tried to have Eel helpers in sync with JavaScript standard library (where possible). AFAIK  the `callback` argument is mandatory there. For me `${Array.filter(myValues)}` also is not quite readable in what it means.

Shouldn't we add an additional method to `ArrayHelper` to do just that? (which has nothing to do with the new method literals)

kitsunet:
> Shouldn't we add an additional method to ArrayHelper to do just that? 
yes please, IMHO this can go in as is and then we can add an `Array.filterEmptyValue()` at some point.

jonnitto:
I just thought about the `array_filter` function in PHP. I know, Eel Helpers are close to JS, but many of them are closer to PHP…

hlubek:
Rebased to master and solved conflicts, should be good to go (I kept return type hints where we can assume they are non breaking)

albe:
Since in the JavaScript reference the parameter is also called `accumulator` I'd stick with that name at least for the EEL case.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce#Parameters

mficzel:
Altered my review to approve because @albe made a valid point and EEL is inspired by JS which is not the case for Fusion. 

Altering the Fusion:Reduce from `carryName = 'carry'` to `accumulatorName = 'accumulator'` would be an option but looks ab bit longish to me. So suggest to stick with `carry` there.

kitsunet:
@jonnitto Do we want to bring this into 5.2 still? Then we need to change the base branch before merging. Your call, I was about to hit merge button.

PR #1327, opened 2018-06-18
FEATURE: base64 en-/decoding via Fusion

lisakampert:
**What I did**
An Eel-Helper which uses PHP functions to encode and decode a data-string


kitsunet:
Hey Lisa, thanks, this is indeed a great idea, I would probably avoid creating a separate class for this though and just add the functionality to the string helper.
What do you think?

bwaidelich:
Nice one! I agree to Christian's comment. Let's add it to the existing `StringHelper` and call it `base64encode()`/`base64decode()`

lisakampert:
It's okay for me, too. But I hadn't found any time to change and adjust it ;-) Hope I'll find time on the weekend!

bwaidelich:
I'll take care

kitsunet:
Will merge when travis is done.

PR #1328, opened 2018-06-19
TASK: Add typehint to allosCallOfMethod

mstruebing:
Since 3.3 is supporting PHP 5.5 I make this PR against 4.0. (PHP 5.5. doesn't support TypeHints.


kitsunet:
Can't change the interface signature in a released version. This is a breaking change and can only happen in next major.

PR #1330, opened 2018-06-22
Improve performance of ResourceManager::getResourceBySha1()

bwaidelich:
This adds an index to the `PersistentResource` table drastically
improving the performance of `ResourceManager::getResourceBySha1()`
calls (= public API).

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
+++ b/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
@@ -25,6 +25,7 @@ use Neos\Flow\ResourceManagement\Exception as ResourceException;
  * Model representing a persistable resource
  *
  * @Flow\Entity
+ * @ORM\Table(indexes={@ORM\Index(columns={"sha1"})})
  */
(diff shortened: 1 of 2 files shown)


bwaidelich:
Some performance tests I did locally (MacBookPro with 10k Resources):

`ResourceManager::getResourceBySha1()` 1000 iterations:

Without index:

```First: 0.0597s, last: 0.0376s, average: 0.0441s, total: 44.136s```

With index:

```First: 0.0106s, last: 0.001s, average: 0.0003s, total: 0.330s```

kitsunet:
Shouldn't this go into 3.3 even? But I take it either way 😆 👍

bwaidelich:
Pew.. I started with 3.3 but considered it less of a (critical) bug and more of a performance optimization, but well..
Postgres good point!

kdambekalns:
Well, being 133,7454545 times slower is at least wanting to be considered a bug :)

bwaidelich:
@kitsunet  @kdambekalns right right, there you go: #1331

PR #1331, opened 2018-06-22
Improve performance of ResourceManager::getResourceBySha1()

bwaidelich:
This adds an index to the `PersistentResource` table drastically
improving the performance of `ResourceManager::getResourceBySha1()`
calls (= public API).

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Resource.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Resource/Resource.php
@@ -24,6 +24,7 @@ use TYPO3\Flow\Resource\Exception as ResourceException;
  * Model representing a persistable resource
  *
  * @Flow\Entity
+ * @ORM\Table(indexes={@ORM\Index(columns={"sha1"})})
  */
 class Resource implements ResourceMetaDataInterface, CacheAwareInterface
 {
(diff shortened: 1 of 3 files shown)


TYPO3.Flow/Classes/TYPO3/Flow/Resource/Resource.php
```diff
@@ -63,7 +64,7 @@ class Resource implements ResourceMetaDataInterface, CacheAwareInterface
      * The IANA media type of this resource
      *
      * @var string
-     * @Flow\Validate(type="StringLength", options={ "maximum"=100 })
+     * @Flow\Validate(type="StringLength", options={ "maximum"=100 })Version20180622074421.php
```
bwaidelich: oops, hang on

bwaidelich:
I had to fix some typos.. (= Friday). Feel free to squash if you happen to merge this one

kitsunet:
Mmm, I know I suggested 3.3, but I guess we need ot think about how we deal with upmerges and the migration. Because upmerging this migration will horribly fail ofc. So technically we would need to chdeck for table existence before adding the index and then add another migration after upmerging for the new table name. Or am I thinking too complicated here?

kitsunet:
So, eh, or we just save ourselves the trouble and, uhm, go back again to 4.3? Sorry, just an idea, because I think this is very sensible to do and should go in ASAP.

kdambekalns:
I'll take care of the migration voodoo today.

bwaidelich:
> or we just save ourselves the trouble and, uhm, go back again to 4.3?

That was my thinking when I pushed this against 4.3 :)
But it should not be too much work to adjust this when upmerging (see #1300 for reference).

@kdambekalns Thanks for taking care!

kitsunet:
Again, this is IMHO broken like this if you up merge. We need another migration than this for the higher versions IMHO (for correct upgrades) and then this one should still exist but would fail on existing installations. So it NEEDS to check for existence of the table.

PR #1332, opened 2018-06-24
WIP FEATURE: Watchman prototype for file watching

skurfuerst:
... to discuss on the Neos sprint :)

### DIFF
--- a/Neos.Flow/Classes/Monitor/FileMonitor.php
+++ b/Neos.Flow/Classes/Monitor/FileMonitor.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Neos\Flow\Monitor;
 
 /*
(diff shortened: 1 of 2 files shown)


gerhard-boden:
To install watchman on Linux you have to build it from source:
https://facebook.github.io/watchman/docs/install.html#installing-from-source

While I got it running in the end it CAN be a bit tricky to install depending on the distribution you're using (you also need a bunch of dependencies), I would also suggest to make this feature completely optional.

The watchman build was (seemingly) successful, but I still got 
```array(3) {
["watcher"]=>
  string(7) "inotify"
  ["watch"]=>
  string(29) "/var/www/neos-master/Packages"
  ["version"]=>
  string(5) "4.9.0"
}
Notice: Undefined index: files in
/var/www/neos-master/Packages/Framework/Neos.Flow/Classes/Monitor/FileMonitor.php line 293
```

skurfuerst:
@gerhard-boden hmm, if this is the case, I am unsure whether watchman will be worth integrating. It might improve the watching situation on Win and Mac though...

PR #1337, opened 2018-07-02
BUGFIX: Allow usage of multiple '%env:' replacements per Setting

nlx-lars:
**What I did**
Allowed usage of multiple '%env:NAME%' per Settings. In Flow versions below 4.0, this used to work fine. Since 4.0 the ConfigurationManager uses a different regular expression which
only matches the last environment variable used in a Setting.
**How I did it**
Modified the regex used to parse the Configuration of Environment Variables.
**How to verify it**
Run tests or add Setting that has two Environment Variables in it (`something: '%env:VAR1% something else %env:VAR2%'`).

**Checklist**


daniellienert:
Hey @nlx-lars,
thanks to your first contribution to Flow! 
Looks good to me - somehow the claim in the testset looks familiar :)

PR #1341, opened 2018-07-23
BUGFIX: Do not remove leading slashes from base path

beheist:
Passing a path with a leading slash as base path always fails on unix, as all absolute paths begin with a slash there. the tests do not cover this because they use vfs, whose paths start with `vfs://`.

### DIFF
--- a/Neos.Utility.Files/Classes/TYPO3/Flow/Utility/Files.php
+++ b/Neos.Utility.Files/Classes/TYPO3/Flow/Utility/Files.php
@@ -180,7 +180,7 @@ abstract class Files
     public static function removeEmptyDirectoriesOnPath($path, $basePath = null)
     {
         if ($basePath !== null) {
-            $basePath = trim($basePath, '/');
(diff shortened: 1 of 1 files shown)


johannessteu:
Probably a good idea to cover this issue with a unit test?

beheist:
@stolle can we do that? Since the issue is platform-specific and not reproducible with vfs, would that not make the test rather pointless or even problematic, since you can't have paths starting with / on windows?

johannessteu:
@beheist Hmm, yeah. Actually not that easy, you're right. The trim stuff must be excluded in another function to check the rtrim easy >.>

albe:
By now that should go into 4.3 actually - or should we do another round of upmerges?

beheist:
No, it's not that important since that parameter is never used in the core. I'll retarget to Flow 4.3. Are you OK with merging this?

beheist:
Hm, got conflicts now. I'll close this and resubmit from a 4.3 base.

PR #1343, opened 2018-07-23
FEATURE: Logger backend to write the log in JSON format

daniellienert:
This logger writes the log message as JSON which is easy parsable for log processing
but also readable.

### DIFF
--- a/Neos.Flow.Log/Classes/Backend/JsonFileBackend.php
+++ b/Neos.Flow.Log/Classes/Backend/JsonFileBackend.php
@@ -44,7 +44,7 @@ class JsonFileBackend extends FileBackend
         $severityLabel = (isset($this->severityLabels[$severity])) ? strtolower(trim($this->severityLabels[$severity])) : 'unknown';
 
         $logEntryData = [
-            'timestamp' => time(),
+            'timestamp' => date(\DateTime::ATOM),
             'processId' => $processId,
             'severity' => $severityLabel,
             'remoteIp' => $remoteIp,
--- a/Neos.Flow.Log/Classes/Backend/JsonFileBackend.php
+++ b/Neos.Flow.Log/Classes/Backend/JsonFileBackend.php
@@ -1,0 +1,66 @@
+<?php
+namespace Neos\Flow\Log\Backend;
+
+/*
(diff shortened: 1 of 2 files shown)


daniellienert:
Hey @kitsunet, what do you think about adding this backend to the core?
If it is worth it, Ill take care of docs and stuff

kitsunet:
Why not, seems sensible to me :)

Neos.Flow.Log/Classes/Backend/JsonFileBackend.php
```diff
@@ -0,0 +1,66 @@
...
+            'processId' => $processId,
+            'severity' => $severityLabel,
+            'remoteIp' => $remoteIp,
+            'message' => $message,
+            'origin' => [
+                'packageKey' => $packageKey,
+                'className' => $className,
+                'methodName' => $methodName
+            ],
+            'additionalData' => $additionalData
```
dfeyer: What happens is additionalData are not JSON encodable ?

And maybe an array_filter to avoid null value, lots of value are null by default, so the JSON will contains a lots of key without values ?

daniellienert: > What happens is additionalData are not JSON encodable ?
Good point. Any idea, how to cover this? Testing the encoding before and change the array to an error if it fails? Or build something like https://github.com/neos/flow-development-collection/blob/b84273d1728bc7591ce233b37434aa502b4219c5/Neos.Flow.Log/Classes/PlainTextFormatter.php to sanitize the array?

> And maybe an array_filter to avoid null value
I am not sure about that, If you have the same fields in all you messages, then you could rely on that fact when parsing. My use case is ingesting the log messages into elasticsearch and there we have the empty fields anyway

albe: Let `json_encode` handle that and just look for `json_last_error` if it returns `false`.

kitsunet: technically whoever is logging should take care. Putting that data into a log file also means serialisation of sorts (worst case toString), so I think it's fine to expect the data to be json_encode-able

PR #1344, opened 2018-07-25
[BUGFIX] Interface used is not imported

pankajlele:
This fixes fatal error when running kickstart:model command

When running ./flow kickstart:model Test.Me Event "title:string" found fatal error 
Class 'Neos\Kickstarter\Service\FlowPackageInterface' not found

Imported missing interface to fix this

To verify run kickstart:model and it should work without errors!

**Checklist**


kitsunet:
Good catch! Will merge when travis is done.

Issue #1347, opened 2018-07-31
Support Fluid Feature of optional Partials/Sections


vertexvaar:
### Description

I use ` <f:render partial="[file]" optional="true" />` in my template but when rendering i get the exception 
```
The partial files "resource://Vendor.Package/Private/Partials/[file].html", "resource://Vendor.Package/Private/Partials/[file]" could not be loaded.
```
despite defining it as optional

### Steps to Reproduce

1. Create or edit a Template
2. Try to render a non-existing partial wight the render ViewHelper

#### Expected behavior

If the partial does not exist and optional is flagged there must not be an exception about missing partials

#### Actual behavior

I get the exception "The partial files [...] could not be loaded."

### Affected Versions

Flow: 5.0.5

### Possible solutions

The optional-feature is implemented by catching the Exception `\TYPO3Fluid\Fluid\View\Exception\InvalidTemplateResourceException` which is thrown when a file could not be resolved.

1. Let `\Neos\FluidAdaptor\View\Exception\InvalidTemplateResourceException` extend `\TYPO3Fluid\Fluid\View\Exception\InvalidTemplateResourceException`
2. Throw `\TYPO3Fluid\Fluid\View\Exception\InvalidTemplateResourceException` instead of the own Exception

PR #1348, opened 2018-07-31
BUGFIX: Extend the expected exceptions for missing templates and sections

vertexvaar:
typo3fluid/fluid expects specific exceptions to be thrown to implement
the feature of optional sections and partials. Neos.FluidAdaptor has to
throw these exceptions or derivates of them. Otherwise the exceptions won't
be catched and displayed to the user.

fixes: #1347

**What I did**

I implemented solution 1 from the issue

**How I did it**

I made the exceptions shipped with FluidAdaptor inherit from the expected ones from typo3fluid/fluid

**How to verify it**

Try to reproduce the issue after applying the patch

**Checklist**

### DIFF
--- a/Neos.FluidAdaptor/Classes/View/Exception/InvalidSectionException.php
+++ b/Neos.FluidAdaptor/Classes/View/Exception/InvalidSectionException.php
@@ -11,13 +11,11 @@ namespace Neos\FluidAdaptor\View\Exception;
  * source code.
  */
 
-use Neos\FluidAdaptor\View;
-
 /**
  * An "Invalid Section" exception
  *
  * @api
  */
-class InvalidSectionException extends View\Exception
+class InvalidSectionException extends \TYPO3Fluid\Fluid\View\Exception\InvalidSectionException
 {
 }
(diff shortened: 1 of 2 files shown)


albe:
Generally yes, but this is breaking, because cases where `Neos\FluidAdaptor\View\Exception` is caught and expected to also catch the more concrete `InvalidSectionException` and `InvalidTemplateResourceException` will now no longer catch.

This is a bit pedantic I know, but we should try to stick to semantic versioning as much as possible and this is more a feature than a real bug anyway. So this should probably be targeted at master and land into 5.x from that PoV. What do the others think?

vertexvaar:
I don't think of it as a feature because trying to render optional partials results in an exception, but this is just a personal feeling :smile: 
I'm working on 5.x so i'd be fine with the merge to master.

Until then i will add the requested partial file as an empty file as a workaround.

albe:
Damn, that was targeting the wrong branch by now. Will open another PR that targets 4.3.

aertmann:
@albe: it would get upmerged to 4.3.. unless you didn't intend this to land in 4.0, 4.1 and 4.2 as well? if no this PR needs to be reverted..

albe:
4.0-4.2 are in security only mode, so... yeah, needs to be reverted in 4.0

vertexvaar:
Do i have to create another PR for 5.0/5.1 ?

albe:
No, this is going to be upmerged in all higher branches. Thanks for asking :)

PR #1349, opened 2018-07-31
TASK: Throw a helpful exception when enctype is not set correctly for uploads

### LINKED ISSUE #545: Provide a hint when a form with an upload field does not have enctype set correctly
Jira issue originally created by user @albe:

It would be nice if Fluid would inform the developer that he forgot to add the "enctype" attribute on the form when he is using an upload field, where enctype "multipart/form-data" is required.

Currently, the file upload will then just silently fail.

 Jira-URL: https://jira.neos.io/browse/FLOW-176

albe:
This change will throw an Exception when the Form Upload ViewHelper is used, but the according Form does not specify an enc-type of "multipart/form-data", which is easily forgotten and leads to non-working forms.

Resolves #545


albe:
Travis hickups. Gonna restart jobs.

albe:
:tada: finally, travis let all jobs run through and they pass!

albe:
Damn me - just had to improve those ugly implementation tests :crossed_fingers: travis

PR #1350, opened 2018-08-01
BUGFIX: Allow nullable constructor arguments

lisakampert:
**What I did**
Check if the `$targetType` is nullable and the given source, too. If this is true, return `null`.


albe:
Closed in favor of #1353

PR #1355, opened 2018-08-06
BUGFIX: Correctly build sub process command arguments

kitsunet:
While building sub commands via ``Scripts::buildSubprocessCommand``
the arguments are not build in a syntax that is sensible for parsing
by Flow later. Specifically an argument is build like this:

    '--argumentName' 'argumentValue'

The missing equals sign (=) makes this problematic to parse if the
value contains an equal sign itself as that will then be identified
as separator between argument name and value. With this change those
arguments are now build like this:

    '--argumentName'='argumentValue'

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -689,7 +689,7 @@ class Scripts
         $escapedArguments = '';
         foreach ($commandArguments as $argument => $argumentValue) {
             $argumentValue = trim($argumentValue);
-            $escapedArguments .= ' --' . trim($argument) . ($argumentValue !== '' ? '=' . escapeshellarg($argumentValue) : '');
+            $escapedArguments .= ' ' . escapeshellarg('--' . trim($argument)) . ($argumentValue !== '' ? '=' . escapeshellarg($argumentValue) : '');
         }
 
         $command .= sprintf(' %s %s %s', escapeshellarg(FLOW_PATH_FLOW . 'Scripts/flow.php'), escapeshellarg($commandIdentifier), trim($escapedArguments));
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -687,13 +687,9 @@ class Scripts
         }
 
         $escapedArguments = '';
-        if ($commandArguments !== []) {
-            foreach ($commandArguments as $argument => $argumentValue) {
-                $escapedArguments .= ' ' . escapeshellarg('--' . trim($argument));
-                if (trim($argumentValue) !== '') {
-                    $escapedArguments .= ' ' . escapeshellarg(trim($argumentValue));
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -687,13 +687,9 @@ protected static function buildSubprocessCommand($commandIdentifier, array $sett
...
-        if ($commandArguments !== []) {
-            foreach ($commandArguments as $argument => $argumentValue) {
-                $escapedArguments .= ' ' . escapeshellarg('--' . trim($argument));
-                if (trim($argumentValue) !== '') {
-                    $escapedArguments .= ' ' . escapeshellarg(trim($argumentValue));
-                }
-            }
+        foreach ($commandArguments as $argument => $argumentValue) {
+            $argumentValue = trim($argumentValue);
+            $escapedArguments .= ' --' . trim($argument) . ($argumentValue !== '' ? '=' . escapeshellarg($argumentValue) : '');
```
bwaidelich: With that `escapeshellarg` is no longer applied to the argument name. That might be a problem, no?

kitsunet: See commit message. An argument name can neither contain spaces nor special characters-

bwaidelich: > IMHO the argument name cannot and should not be something that needs escaping

But effectively this would introduce a potential vulnerability if $someone forwards the argument from an untrusted source, wouldn't it?

kitsunet: In theory, yes, but we are looking at a low level method here, if you forward an argument name from an untrusted source to a command line command you made mistakes already.
But if you prefer I can reinstate that :)

kdambekalns: I'd leave it in place, TBH.

bwaidelich: I don't want to split hairs, but it's called by highlevel (API) methods (`executeCommand(Async)()`) and the commandArguments are passed through without validation.

kdambekalns:
This worked in the past:

    '--0' 'some-value'

The new version does not work:

    '--0'='some-value'

What does work, though, is:

    '--0=some-value'

bwaidelich:
Maybe we can detect whether it's an "associative" array and use the old (unnamed) format if it's not?

kdambekalns:
In the end it boils down to

    $ "/opt/local/bin/php" -c '/opt/local/etc/php72/php.ini' '/Users/karsten/…/flow.php' 'flowpack.jobqueue.common:job:execute' '--0=dummy' '--1=4959aa81b9aa43f79b7b8d81b2e6f388fa94db47'
    Please specify the required argument "queue": ^C

So, the names must not be given, unless they are "argument names".

kdambekalns:
See https://github.com/neos/flow-development-collection/pull/1386

PR #1357, opened 2018-08-07
BUGFIX: Add ext-xml as composer dependency

### LINKED ISSUE #1356: Missing ext-xml dependency in composer.json
We do depend on ext-xml inside the i18n XliffParser as far as I understand, so we should actually add `"ext-xml"` as a dependency in composer.json.

albe:
Fixes #1356


kdambekalns:
The dependency must be added to the composer manifest of the Neos.Flow package as well - the one changed in this PR is generated from the individual manifests!

albe:
Argh, should have known that... 🙈

PR #1353, opened 2018-08-02
BUGFIX: Allow nullable constructor arguments

lisakampert:
- Check if the $targetType is nullable and the given source, too. If this is true, return null.

**What I did**

**How I did it**

**How to verify it**

**Checklist**

- [x] Code follows the PSR-2 coding style
- [x] Tests have been created, run and adjusted as needed
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)

PR #1358, opened 2018-08-08
BUGFIX: Allow nullable constructor arguments

### LINKED ISSUE #1361: Reflection of arguments annotated as `|null`
Currently, the ReflectionService does not deal properly with arguments to methods (and properties?) that are annotated as "nullable" via the `|null` pattern.

This leads to following effects:

- for an argument, that is not also typehinted nullable via `= null` default value, the `ALLOW_NULL` reflection property is not set

- such arguments are not properly type expanded and end up as `SomeClass|null` in the reflection data instead of `Acme\Foo\Bar\SomeClass|null` (this in turn prevents proper type conversion from happening - see also #1358)

albe:
In PropertyMapper checks if the $targetType is nullable and the given source, too. If this is true, return null. Also, in ReflectionService, the annotated type is properly expanded when annotated with '|null' or 'null|'.

**Checklist**

Alternative push for #1353

Fixes #1361

### DIFF
--- a/Neos.Flow/Classes/Property/PropertyMapper.php
+++ b/Neos.Flow/Classes/Property/PropertyMapper.php
@@ -155,7 +155,8 @@ class PropertyMapper
     protected function doMapping($source, $targetType, PropertyMappingConfigurationInterface $configuration, &$currentPropertyPath)
     {
         $targetTypeWithoutNull = TypeHandling::stripNullableType($targetType);
-        if ($source === null && $targetType !== $targetTypeWithoutNull) {
+        $isNullableType = $targetType !== $targetTypeWithoutNull;
+        if ($source === null && $isNullableType) {
             return null;
         }
         $targetType = $targetTypeWithoutNull;
--- a/Neos.Flow/Classes/Property/PropertyMapper.php
+++ b/Neos.Flow/Classes/Property/PropertyMapper.php
@@ -154,6 +154,12 @@ class PropertyMapper
      */
     protected function doMapping($source, $targetType, PropertyMappingConfigurationInterface $configuration, &$currentPropertyPath)
     {
+        $targetTypeWithoutNull = TypeHandling::stripNullableType($targetType);
+        if ($source === null && $targetType !== $targetTypeWithoutNull) {
+            return null;
+        }
+        $targetType = $targetTypeWithoutNull;
+        
         if (is_object($source)) {
             $targetClass = TypeHandling::truncateElementType($targetType);
             if ($source instanceof $targetClass) {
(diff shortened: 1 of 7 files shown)


albe:
The code change in ObjectConverter is actually a workaround for the problem described in #1361 

```php
            $typeWithoutNull = TypeHandling::stripNullableType($methodParameters[$propertyName]['type']);
            if ($typeWithoutNull != $methodParameters[$propertyName]['type']) {
                return $methodParameters[$propertyName]['class'] . '|null';
            } else {
                return $methodParameters[$propertyName]['type'];
            }
```

Therefore I'd like to get a bugfix for that issue in instead of merging this work-around.

albe:
Can't approve since I'm PR author because #1353 didn't pass build somehow. Looking good though from my side.

lisakampert:
I tested our nullable-problem against these changes and it works fine!

Neos.Flow/Classes/Property/PropertyMapper.php
```diff
@@ -154,6 +154,12 @@ public function getMessages()
      */
     protected function doMapping($source, $targetType, PropertyMappingConfigurationInterface $configuration, &$currentPropertyPath)
     {
+        $targetTypeWithoutNull = TypeHandling::stripNullableType($targetType);
+        if ($source === null && $targetType !== $targetTypeWithoutNull) {
```
bwaidelich: &lt;nitpick&gt;should we maybe inline this logic to `&& TypeHandling::isNullableType($targetType)`? Would make it a bit more readable and reusable&lt;/nitpick&gt;

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1452,24 +1452,26 @@ protected function reflectClassMethodParameter($className, MethodReflection $met
      */
     protected function expandType(ClassReflection $class, $type)
     {
+        $typeWithoutNull = TypeHandling::stripNullableType($type);
+        $isNullable = $typeWithoutNull !== $type;
```
bwaidelich: ..could then be changed to `TypeHandling::isNullableType($targetType)` in the places below

Neos.Flow/Classes/Property/PropertyMapper.php
albe: Sounds reasonable.

albe: Caveat: We would do the stripping call twice (which is a bit costly if it ends up in the preg_replace) I'd leave it like that for now, since we're inside property mapping which might get called quite a lot and just make the intention more explicit by assigning the $type comparison to an expressive variable (like below).

PR #1360, opened 2018-08-09
TASK: Remove doesNotPerformAssertions from a test

kdambekalns:
… because the test does in fact assert something. So PhpUnit complained:

`This test is annotated with "@doesNotPerformAssertions" but performed 1 assertions`


albe:
Hm, just noticed the same is in https://github.com/neos/flow-development-collection/pull/1358/files#diff-128583b778fa56ba5599af848567d490R539 - but this didn't give me warnings while testing. Is this specific to a PhpUnit version?

kdambekalns:
Yeah, probably "became wrong" with the latest PhpUnit release…

PR #1362, opened 2018-08-09
BUGFIX: Use same instance for injecting Doctrine ObjectManager and EntityManagerInterface

### LINKED ISSUE #1345: Using Doctrine\Common\Persistence\ObjectManager is broken instead of deprecated
Since https://github.com/neos/flow-development-collection/pull/1272 the usage of `ObjectManager` is deprecated in favor of the `EntityManagerInterface`. However, injecting the old `ObjectManager` is breaking behaviour, by no longer persisting data instead of just working correctly still.

Flow: 5.0+

What we need to do, is somehow return the same instance for `EntityManagerInterface` and doctrine `ObjectManager`, either from the `EntityManagerFactory` or the Flow `ObjectManager` directly.

albe:
This fixes the b/c break introduced with the deprecation of the ObjectManager. The fix can be removed with Flow 6.0

Fixes #1345

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/ObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/ObjectManager.php
@@ -191,6 +191,11 @@ class ObjectManager implements ObjectManagerInterface
      */
     public function get($objectName)
     {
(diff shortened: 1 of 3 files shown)


albe:
@DavidSporer @kdambekalns

DavidSporer:
Looks good from what I can judge :-)

PR #1363, opened 2018-08-10
BUGFIX: Do not remove leading slashes from base path

beheist:
Retargetet to Flow 4.3. For discussion see #1341 .


albe:
Test failure was something with "could not download" - restarted the job

PR #1365, opened 2018-08-10
TASK: Lower severity of no route matched log message

daniellienert:
I feel that was discussed before. 
We just moved a big project (from wordpress) to Neos and the system log is flooded with thousands of route not matching log levels.
Do we really need them on Notice level so that it gets log on production servers?

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/Router.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Router.php
@@ -13,6 +13,7 @@ namespace Neos\Flow\Mvc\Routing;
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Configuration\ConfigurationManager;
+use Neos\Flow\Mvc\Exception\InvalidRoutePartValueException;
 use Neos\Flow\Mvc\Exception\InvalidRouteSetupException;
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -128,7 +131,7 @@ public function route(RouteContext $routeContext): array
                 return $matchResults;
             }
         }
-        $this->logger->notice(sprintf('Router route(): No route matched the route path "%s".', $httpRequest->getRelativePath()));
+        $this->logger->debug(sprintf('Router route(): No route matched the route path "%s".', $httpRequest->getRelativePath()));
```
dfeyer: Maybe info is enought, you can filter out info level on production, debug seems a bit like "we really don't care about this in production"

daniellienert: Frankly, I don't care about that in production. It just says, the url could not be resolved, which leads to a 404. I care about the 404s in the web servers access log, but I don't need it in the system log of the application. 

And maybe I do it wrong, but I do care about meaningful info log messages in production. They tell me what is happening / what happened on the system which not necessarily needs to be an error.

PR #1366, opened 2018-08-12
TASK: Deprecating non PSR-7 methods

kitsunet:


Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -111,7 +112,8 @@ public function hasHeader($name)
      * @param boolean $replaceExistingHeader If a header with the same name should be replaced. Default is TRUE.
      * @return self This message, for method chaining
      * @throws \InvalidArgumentException
-     * @api
+     * @deprecated Sicne Flow 5.1, use withHeader instead
```
lorenzulrich: Since

Neos.Flow/Classes/Http/Headers.php
```diff
@@ -16,8 +16,9 @@
 /**
  * Container for HTTP header fields
  *
- * @api
+ * @deprecated Headers will be an array in side of the Message object
```
lorenzulrich: "inside the Message object" would be better IMO

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
```diff
@@ -0,0 +1,126 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * Helper to extract various informations from PSR-7 requests.
+ */
+abstract class RequestInformationHelper
+{
+    /**
+     * Returns the relative path (ie. relative to the web root) and name of the
```
lorenzulrich: i.e.

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
```diff
@@ -0,0 +1,126 @@
+<?php
+namespace Neos\Flow\Http\Helper;
+
+use Neos\Flow\Http\Headers;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * Helper to extract various informations from PSR-7 requests.
```
lorenzulrich: information (no s)

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
```diff
@@ -0,0 +1,126 @@
...
+        }
+        if (isset($server['ORIG_SCRIPT_NAME'])) {
+            return $server['ORIG_SCRIPT_NAME'];
+        }
+
+        return '';
+    }
+
+    /**
+     * Returns the relative path (ie. relative to the web root) to the script as
```
lorenzulrich: i.e.

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
```diff
@@ -0,0 +1,126 @@
...
+        $baseUri = clone $request->getUri();
+        $baseUri = $baseUri->withQuery('');
+        $baseUri = $baseUri->withFragment('');
+        $baseUri = $baseUri->withPath(self::getScriptRequestPath($request));
+
+        return $baseUri;
+    }
+
+    /**
+     * Return the Request-Line of this Request Message, consisting of the method, the uri and the HTTP version
```
lorenzulrich: URI in uppercase

Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php
```diff
@@ -0,0 +1,76 @@
...
+            if (isset($fileInformation['error']) && $fileInformation['error'] !== \UPLOAD_ERR_NO_FILE) {
+                $untangledFiles = Arrays::setValueByPath($untangledFiles, $fieldPath, $fileInformation);
+            }
+        }
+
+        return $untangledFiles;
+    }
+
+    /**
+     * Returns and array of all possibles "field paths" for the given array.
```
lorenzulrich: possible (withous s)

kitsunet:
Ok, this is the minimum package that ideally should go into the release. If travis is fine with it, I might remove the WIP status. At least we have everything deprecated then and have 3 releases to fix that stuff on our side as well...

kitsunet:
This currently leaves a lot of deprecated method usages around BUT it means we get the message out to adjust code now.

kitsunet:
> I'm currently thinking about, if we can somehow wrap the PSR-7 classes into own classes that provide some of the sugar-methods 

That's a big no-go as it automatically invalides the use of the interfaces.

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
```diff
@@ -0,0 +1,248 @@
...
+            } else {
+                $contentLines[] = $line;
+            }
+        }
+        if ($parsingHeader === true) {
+            throw new \InvalidArgumentException('The given raw HTTP message contains no separating empty line between header and body.', 1502984823);
+        }
+        $content = implode(chr(10), $contentLines);
+
+        $response->setHeaders($headers);
```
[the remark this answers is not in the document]
kitsunet: We can run withHeader for each single one, which is the way to go, but for this change I mostly want to deprecate things and make sure alternatives exist.

Neos.Flow/Classes/Http/RequestHandler.php
```diff
@@ -104,9 +105,13 @@ public function handleRequest()
...
         $this->resolveDependencies();
-        $this->addPoweredByHeader($response);
+        $response = $this->addPoweredByHeader($response);
+        $this->componentContext->replaceHttpResponse($response);
         if (isset($this->settings['http']['baseUri'])) {
-            $request->setBaseUri(new Uri($this->settings['http']['baseUri']));
+            $baseUri = new Uri($this->settings['http']['baseUri']);
+            $request = $request->withAttribute(Request::ATTRIBUTE_BASE_URI, $baseUri);
+            $this->componentContext->replaceHttpRequest($request);
+            $request->setBaseUri($baseUri);
```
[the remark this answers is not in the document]
kitsunet: Needs further refactoring before I am confident to remove it.

kitsunet:
@albe see comments/new commit.

Neos.Flow/Classes/Http/RequestHandler.php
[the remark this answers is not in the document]
kitsunet: One of the main problems is that we have baseUri on the BaseRequest while Attributes only exist on ServerRequests, so we cannot replace baseUri with attributes on the BaseRequest. Any change to that would be a breaking one I guess, so I think we got to leave it like this until next major.

kitsunet:
This looks pretty good to me for a quick shot to prepare people for next major. We can iterate on that in the next releases then. 
One major gripe I have is `getHeader()` which according to our implementation can return a single value (string) or an array of multiple values. According to PSR it should always be an array even with one entry (which consistency wise makes sense). Technically IMHO it's not a breaking change to change that behaviour now and let getHeader ALWAYS return an array just like PSR requires us to. BUT even our code often just assumes a singular value and fails hard, so \o/ I am a bit unsure how to deal with that. For now I will probably want to leave it as is, but would be good to discuss this particular part.

albe:
Create a depreciation for the value return type with breakage release note, but still return single values (return type annotation doesn't match actual return types). Then change with next major.

Neos.Flow/Classes/Http/RequestHandler.php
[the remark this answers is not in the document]
albe: Don't we have a ServerRequest here anyway? Otherwise we couldn't `withAttribute` either

Neos.Flow/Classes/Http/Response.php
```diff
@@ -78,13 +78,15 @@ public function __construct(ResponseInterface $parentResponse = null)
...
      * @deprecated Since Flow 5.1, use ResponseInformationHelper::createFromRaw
      * @see ResponseInformationHelper::createFromRaw()
      */
     public static function createFromRaw($rawResponse, Response $parentResponse = null)
     {
-        return ResponseInformationHelper::createFromRaw($rawResponse, $parentResponse);
+        /** @var Response $response */
+        $response = ResponseInformationHelper::createFromRaw($rawResponse);
+        $response->parentResponse = $parentResponse;
     }
```
albe: Missing return

Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -76,8 +78,9 @@ public function getHeaders()
      * Dates are returned as DateTime objects with the timezone set to GMT.
      *
      * @param string $name Name of the header
-     * @return array|string An array of field values if multiple headers of that name exist, a string value if only one value exists and NULL if there is no such header.
+     * @return string[] An array of field values, may be empty if the header was not set (check with hasHeader)
      * @api
+     * @deprecated Not deprecated but from 5.1 this will only return arrays of strings according to the PSR-7 standard.
```
kdambekalns: Not true… In the current code (and thus 5.1) this will still return string or array…

kdambekalns:
Pushed a smallish commit, I wanted to get a feeling of how much we use deprectaed methods and how much work it is to get rid of those usages. Some things are simple, other are not.

Neos.Flow/Classes/Http/AbstractMessage.php
kitsunet: Correct, that should read "from next major..."

kitsunet:
The answer is `a lot` and `not simple` ;)

albe:
Yeah, this is gonna be a big change. And I second that the excessive use of static helpers required is not nice to say the least. I guess standardization is a nice concept here, but the PSR-7 interface is far too basic/generic. It is what it is though. What I also find odd somehow with the immutability, is the absurd creation of needless clones for construction. I guess there should be a Builder class that can mutate the object before returning it, but PHP ain't got no friend classes. Still I see the benefit down the road, which is PSR-15 and plugging in HTTP behavior from 3rd party.

Neos.Flow/Classes/Http/RequestHandler.php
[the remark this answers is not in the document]
kitsunet: huh, why do you mean it#s created above? it gets an attribute, but the way the code is structured makes it at least technically breaking if we do not set the deprecated baseUri as well.

Neos.Flow/Classes/Http/Headers.php
```diff
@@ -186,8 +202,9 @@ public function getAll()
     {
         $fields = $this->fields;
         $cacheControlHeader = $this->getCacheControlHeader();
-        if (!empty($cacheControlHeader)) {
-            $fields['Cache-Control'] = [$cacheControlHeader];
+        $fields['Cache-Control'] = [$cacheControlHeader];
+        if (empty($cacheControlHeader)) {
```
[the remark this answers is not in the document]
kitsunet: I guess this needs a longer explanation.... Eventually we should return ALL headers, right now our Headers implementation handles two headers differently. One is `Set-Cookie` the other is `Cache-Control`. As you can see above I initialise `Cache-Control` empty in the defaults, therefore at this point it could've been empty all along or from setting it, either way, in that case it must be unset. Therefore unconditionally setting and then unsetting if empty, everything else would be more code.

I didn't include Set-Cookie here yet, for sake of backwards compat (seemed riskier than cache control).

Neos.Flow/Classes/Http/Helper/ArgumentsHelper.php
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Helper to unify different HTTP request arguments.
+ */
+abstract class ArgumentsHelper
+{
+    /**
+     * Takes the raw GET & POST arguments and maps them into the request object.
+     * Afterwards all mapped arguments can be retrieved by the getArgument(s) method, no matter if they
+     * have been GET, POST or PUT arguments before.
+     *
```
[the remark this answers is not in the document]
kitsunet: Sure can do, although that is not a comment I wrote, but just moved around.

Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -440,10 +467,15 @@ public function withoutHeader($name)
...
         $streamResource = fopen('php://memory', 'r+');
-        fwrite($streamResource, $this->content);
+        if (is_resource($this->content)) {
+            stream_copy_to_stream($this->content, $streamResource);
+        } else {
+            fwrite($streamResource, $this->content);
+        }
         rewind($streamResource);
 
-        return new ContentStream($streamResource);
```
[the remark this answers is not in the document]
kitsunet: Ah must be a left over from refactoring because I had some other operation inbetween there...

kitsunet:
DO NOT MERGE AT THIS POINT, you can review but there is something broken right now, investigating.

kitsunet:
So the test error is basically a separation problem in functional tests. the injected (current) session in authentication is a different one than the one in the request so the one in the request is never started. Not an issue in a "regular" request flow, but in functional tests it is. Looking at a fix...

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -562,7 +562,6 @@ Neos\Flow\Security\Authentication\Provider\TestingProvider:
 #                                                                          #
 
 Neos\Flow\Session\SessionInterface:
-  scope: singleton
```
albe: Isn't this breaking AF?

kitsunet: Not really, because it is fetched from the SessionManager, just with the difference that now the object manager doesn't hold that reference as well. Check how the object manager internally deals with objects and factories with singleton vs. prototype. It wouldn't have mattered to change that just now in some old version, it just prevents the object manager to hold a reference itself.

Neos.Flow/Classes/Http/Helper/ArgumentsHelper.php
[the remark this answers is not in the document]
kitsunet: This is done.

Neos.Flow/Classes/Http/AbstractMessage.php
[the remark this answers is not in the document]
kitsunet: done

Neos.Flow/Classes/Http/RequestHandler.php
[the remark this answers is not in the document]
albe: Doesn't setBaseUri just set the attribute internally too?

Neos.Flow/Classes/Http/Headers.php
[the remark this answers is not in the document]
kdambekalns: Ah, indeed. Could be this, I think, but fine either way:

        $fields = $this->fields;
        $fields['Cache-Control'] = [$this->getCacheControlHeader()];
        if (empty($fields['Cache-Control'])) {
            unset($fields['Cache-Control']);
        }
        return $fields;

And cookies are simply not returned by `getAll()` now, right? And that is like it is now?

Neos.Flow/Classes/Http/RequestHandler.php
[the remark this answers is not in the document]
kitsunet: Yeah, good point, that is true only because we actually can know that we get a `Request` here and not something else. I was in a bit too deep and thought it may happen that we have an instance of `BaseRequest` (which also has the set/get baseUri methods) but has no attributes, so yes, rather pointless with the overriden baseuri methods in `Request`. Will change.

Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -76,8 +78,9 @@ public function getHeaders()
      * Dates are returned as DateTime objects with the timezone set to GMT.
      *
      * @param string $name Name of the header
-     * @return array|string An array of field values if multiple headers of that name exist, a string value if only one value exists and NULL if there is no such header.
+     * @return string[] An array of field values, may be empty if the header was not set (check with hasHeader)
```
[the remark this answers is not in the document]
kitsunet: Yep, fine for me, just wanted to make sure people actually read it 😆

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -47,6 +49,9 @@ Neos:
 
         'postprocess':
           chain:
+            'prepareSession':
```
albe: Are we still "preparing" the session at this stage? Maybe `'sendSession'` or similar might work better here.

Neos.Flow/Classes/Session/Http/SessionRequestComponent.php
```diff
@@ -0,0 +1,86 @@
...
+     * @var SessionManagerInterface
+     */
+    protected $sessionManager;
+
+    /**
+     * @param ComponentContext $componentContext
+     */
+    public function handle(ComponentContext $componentContext)
+    {
+        if (!$this->sessionManager instanceof SessionManager) {
```
albe: This probably hints at a deficit in the SessionManagerInterface, but out of scope of this PR.

Neos.Flow/Classes/Session/Aspect/LazyLoadingAspect.php
```diff
@@ -34,9 +36,9 @@ class LazyLoadingAspect
 
     /**
      * @Flow\Inject
-     * @var SessionInterface
+     * @var SessionManagerInterface
      */
-    protected $session;
+    protected $sessionManager;
```
albe: What's the reason to inject `SessionManagerInterface` instead of `SessionInterface` here? Just for consistency with the components or is there a technical reason?

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
```diff
@@ -31,10 +31,10 @@
 class AuthenticationProviderManager implements AuthenticationManagerInterface
 {
     /**
-     * @var SessionInterface
+     * @var SessionManagerInterface
      * @Flow\Inject
      */
-    protected $session;
+    protected $sessionManager;
```
albe: Same as below, but I guess the answer will be the same.

Neos.Flow/Classes/Http/Helper/ArgumentsHelper.php
```diff
@@ -0,0 +1,54 @@
...
+     * The order of merging is GET, POST, FILES, with later items overriding earlier ones.
+     *
+     * @param array $getArguments Arguments as found in $_GET
+     * @param array $postArguments Arguments as found in $_POST
+     * @param array $untangledFiles Untangled $_FILES as provided by \Neos\Flow\Http\Helper\UploadedFilesHelper::untangleFilesArray
+     * @return array the unified arguments
+     * @see \Neos\Flow\Http\Helper\UploadedFilesHelper::untangleFilesArray
+     */
+    public static function buildUnifiedArguments(array $getArguments, array $postArguments, array $untangledFiles): array
```
albe: Is this even called anywhere right now?

Neos.Flow/Classes/Session/Aspect/LazyLoadingAspect.php
kitsunet: Technical reason. This just like the AuthenticationProviderManager both are initialised very early in Flow (before the component chain runs), so the currentSession isn't initialised properly at this point, leading to cookies not being set.

Neos.Flow/Classes/Session/Http/SessionRequestComponent.php
kitsunet: Yep it does and correct. IMHO we should actually set the interface to internal or remove it, but that's for another change.

Neos.Flow/Classes/Http/Helper/ArgumentsHelper.php
kdambekalns: in `Http\Request` on line 142 (in `__construct`)

albe: Totally missed that, because that file was collapsed! :D

PR #1369, opened 2018-08-18
!!!BUGFIX: Cyclic validator calls don't lose previous results

albe:
This changes uses a stack inside Validators to store results of previous
calls, so that results are not lost when a validator is called in a cycle.
This may happen for any Object or Collection validators, since the Validator
instances chain is built per class and hence cyclic relations (A -> * -> A)
may lead to incorrect validation results otherwise.

Note: This bugfix is only breaking if you implemented your own Validator, extending `GenericObjectValidator` or `CollectionValidator`, overriding the `validate()` method and setting `$this->result` inside. In that case you are required to change the code to make use of `pushResult()`/`popResult()` like it is done inside `AbstractValidator::validate()`.
This is an exception to the otherwise semantic versioning we strive for. It was decided because the breaking only happens in cases that would otherwise potentially be affected by the buggy behaviour of validation that could lead to invalid data entering the system.

In general, you are advised to create own Validators only by implementing the `isValid()` method instead of overriding the `validate()` method.

### DIFF
--- a/Neos.Flow/Classes/Validation/Validator/AbstractValidator.php
+++ b/Neos.Flow/Classes/Validation/Validator/AbstractValidator.php
@@ -51,9 +51,15 @@ abstract class AbstractValidator implements ValidatorInterface
     protected $options = [];
 
     /**
+     * @deprecated since Flow 5.1. Don't overwrite this and instead use pushResult/popResult
      * @var ErrorResult
      */
     protected $result;
+
+    /**
+     * @var array<ErrorResult>
+     */
+    protected $resultStack;
 
     /**
      * Constructs the validator and sets validation options
@@ -90,6 +96,34 @@ abstract class AbstractValidator implements ValidatorInterface
             ),
             $options
         );
+
+        $this->resultStack = [];
+    }
+
+    /**
+     * Push a new Result onto the Result stack and return it in order to fix cyclic calls to a single validator.
+     * @since Flow 5.1
+     * @see https://github.com/neos/flow-development-collection/pull/1275#issuecomment-414052031
+     * @return ErrorResult
+     */
+    protected function pushResult()
+    {
+        if ($this->result !== null) {
+            \array_push($this->resultStack, $this->result);
+        }
+        $this->result = new ErrorResult();
+        return $this->result;
+    }
+
+    /**
+     * Pop and return the current Result from the stack and make $this->result point to the last Result again.
+     * @return ErrorResult
+     */
+    protected function popResult()
+    {
+        $result = $this->result;
+        $this->result = array_pop($this->resultStack);
+        return $result;
     }
 
     /**
@@ -102,11 +136,11 @@ abstract class AbstractValidator implements ValidatorInterface
      */
     public function validate($value)
     {
-        $this->result = new ErrorResult();
+        $this->pushResult();
         if ($this->acceptsEmptyValues === false || $this->isEmpty($value) === false) {
             $this->isValid($value);
         }
-        return $this->result;
+        return $this->popResult();
     }
 
     /**
--- a/Neos.Flow/Classes/Validation/Validator/AbstractValidator.php
+++ b/Neos.Flow/Classes/Validation/Validator/AbstractValidator.php
@@ -51,7 +51,7 @@ abstract class AbstractValidator implements ValidatorInterface
     protected $options = [];
 
     /**
-     * @deprecated since Flow 5.1. Don't overwrite this and instead use pushResult/popResult
+     * @deprecated since Flow 4.3. Don't overwrite this and instead use pushResult/popResult
      * @var ErrorResult
      */
     protected $result;
@@ -102,14 +102,14 @@ abstract class AbstractValidator implements ValidatorInterface
 
     /**
      * Push a new Result onto the Result stack and return it in order to fix cyclic calls to a single validator.
-     * @since Flow 5.1
+     * @since Flow 4.3
      * @see https://github.com/neos/flow-development-collection/pull/1275#issuecomment-414052031
      * @return ErrorResult
      */
     protected function pushResult()
     {
         if ($this->result !== null) {
-            \array_push($this->resultStack, $this->result);
+            array_push($this->resultStack, $this->result);
         }
         $this->result = new ErrorResult();
         return $this->result;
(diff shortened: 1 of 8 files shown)


Neos.Flow/Tests/Functional/Mvc/ActionControllerTest.php
```diff
@@ -286,6 +286,28 @@ public function validatedGroupObjectArgumentsAreValidated()
         $this->assertEquals($expectedResult, $response->getContent());
     }
 
+    /**
+     * @test
+     */
+    public function cyclicCollectionsAreValidated()
```
albe: This test will fail and let this object through validation, even though the inner e-mail is invalid, without the push/pop semantic applied.

albe:
This is a possible fix for the issue discovered in #1275 

However there a few things to discuss first:

- I "deprecated" the `result` property of the `AbstractValidator` to signal, that it's usage needs to take care and that you should use the methods `pushResult`/`popResult` inside the `validate()` method instead of overwriting `result`. Is this okay or should this be communicated otherwise?

- This just solves the issue at hand in a b/c as possible way - however, it introduces a weird concept of having to push/pop results inside the Validator to work with them. Any validator not adhering to this new mechanic could potentially break this fix again. This would however not worsen things over what is currently the case. There would be possibly other solutions without a stack, but that would probably be breaking, because they would fully get rid of `result` instance variable usage.

- Should this land in 4.3+ or only in master (5.1+)? I feel the bug is bad enough (it let's invalid object hierarchies through validation) to justify getting this into 4.3

albe:
Okay, so apparently this is currently breaking other tests. Specifically, this is breaking, because the `AggregateBoundaryValidator` extends the `ObjectValidator` and doesn't adhere to this new mechanic. This just shows how breaking this could possibly be, though the fix for this case is easy. Basically any Validator extending the `ObjectValidator` or `CollectionValidator` and overwriting the `validate()` method could break things.

Neos.Flow/Classes/Validation/Validator/AbstractValidator.php
```diff
@@ -90,6 +96,34 @@ function ($value) {
...
+    /**
+     * Push a new Result onto the Result stack and return it in order to fix cyclic calls to a single validator.
+     * @since Flow 5.1
+     * @see https://github.com/neos/flow-development-collection/pull/1275#issuecomment-414052031
+     * @return ErrorResult
+     */
+    protected function pushResult()
+    {
+        if ($this->result !== null) {
+            \array_push($this->resultStack, $this->result);
```
kdambekalns: This is namespaced, but usually we don' do this, and other places in this PR also don't do it.

Neos.Flow/Classes/Validation/Validator/AbstractValidator.php
```diff
@@ -90,6 +96,34 @@ function ($value) {
             ),
             $options
         );
+
+        $this->resultStack = [];
+    }
+
+    /**
+     * Push a new Result onto the Result stack and return it in order to fix cyclic calls to a single validator.
+     * @since Flow 5.1
```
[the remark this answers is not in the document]
kdambekalns: Well, since this targets 4.3, it would be 4.3 :)

Neos.Flow/Classes/Validation/Validator/AbstractValidator.php
```diff
@@ -51,10 +51,16 @@
     protected $options = [];
 
     /**
+     * @deprecated since Flow 5.1. Don't overwrite this and instead use pushResult/popResult
```
kdambekalns: Fix version number, see below…

kdambekalns:
Well, it fixes things. And it provides a way for others to fix "their" validators as well. Seems to be the best solution, given the scope of this.

robertlemke:
+1 for the fix in general.

Just a final question to @albe: do you have any remaining "bad gut feeling" regarding that this change might break existing third-party code?

albe:
@robertlemke I guess, the worst case is something like the `AggregateBoundaryValidator`, where it overwrites the result, but also calls the `parent::validate()` method, since that will break the stack completely. That implementation was pretty bad though, looking at it now, but that doesn't mean we should carelessly break such cases. Maybe we can find a solution for that too? Maybe by internally not depending on `$this->result` at all any more. I'll think a bit about it.

albe:
Okay, so after a bit more thought, we have following situation:

- depending on `$this->result` to stay in tact can break the push/pop semantic

- other VH and especially 3rd party depend on `$this->result`

- even the VH API (`addError()`) depends on an instance variable (the `result` it adds to isn't passed)

Here are some possible part-solutions:

- we completely work around `$this->result`, by keeping current result on stack too and making `addError` work on the stack. This will still break VHs that create their own `Result`, do something with it and then call `parent::validate` and/or `addError` (!).

- we keep it as is, which will break only for VHs like above, that overwrite `result` and call `parent::validate`. Instead we clearly communicate in the release notes, that special care needs to be taken for custom Validator implementations, that make use of `parent::validate()`, since those might inhibit buggy behaviour and should get fixed.

- we do a double push and an additional check, if the `result` was changed (compare `$this->result` with top of stack). If so, we handle that case somehow to prevent breaking the calling validator. This adds a lot of ugly workaround code though.

In any case, we should fix our own VH implementations to avoid overwriting `validate()` needlessly and instead keep their logic inside `isValid()` where everything is safe. This is the case for `Conjunction`/`Disjunction`/`GenericObject`/`Collection` and `AggregateBoundary` validators. At least the latter three are easily changed. The `Conjunction`/`Disjunction` fall out of the scope, because they change the basic validation logic (`acceptsEmptyValues`/`isEmpty` check) so it is okay for them to really overwrite `validate`.

kitsunet:
I am very much for option 2. Clearly documenting what is "public" API of this makes sense anyway.

albe:
Added a small warning section to the docs and changed our own VHs for now.

albe:
`$object` it is inside `isValid` instead of `$value` :roll_eyes:

albe:
Marked the PR as breaking and added a note to the commit message that explains the breaking behaviour and why the decision was made to allow this into a bugfix release.

Is this okay @robertlemke or should we go another route with this?

robertlemke:
@albe sounds good to me, thanks for the additional effort! 👍🏻

PR #1373, opened 2018-08-23
TASK: Document SimpleFileBackend in Flow manual

kdambekalns:


bwaidelich:
Nice one, thanks!
It would be a chance to add a note (or even warning) about the `SimpleFileBackend` not supporting a `lifetime` option!?

PR #1374, opened 2018-08-23
TASK: Add migrations for Neos.Cache PdoBackend

kdambekalns:
This makes sure the tables needed for the PdoBackend are set up
automatically and can be used without further steps.

The migrations support MySQL and PostgreSQL, support for SQLite is
unchanged.

If the PdoBackend is already in use with MySQL, you should adjust the
table definition manually to the new version (make content LONGTEXT and
set the engine to InnoDB.)


aertmann:
These migrations would be run for everyone right? So even if you're not using  the PdoBackend for any caches, you'll still have the tables created. Is that desired? I remember seeing some other implementation that created the tables runtime as needed. What are your thoughts?

bwaidelich:
my preference is also not to create tables that are not used tbh..
Creating them "on the fly" sounds a bit dangerous to me though. My favorite solution would be:
* throw a meaningful exception if a cache backend is not properly setup (i.e. table is missing)
* add some `cache:status` CLI command to be able to check that
* add `cache:setup` CLI command to add the tables for you

It would probably require an additional interface, like `SetupableInterface` :) but we did that for the [EventStore tables](https://github.com/neos/Neos.EventSourcing/blob/master/Classes/Command/EventStoreCommandController.php#L50-L102)  and it works quite nicely IMO.

Anyways, won't block this if others don't mind the "dead" tables

kdambekalns:
I kept the tables names unchanged to avoid potentially breaking things. Because changed table names would need the backend code to be adjusted and if someone already has the backend in use, they would need to rename the existing tables to keep caches intact.

As for automating this – our "Beach companion" hooks into `cache:warmup` and does the needed changes, so that might be an option. Better than yet another command?!

daniellienert:
Hey, I also wouldn‘t like to have the cache tables If I don’t use the pro backend.
Just into the blue - the file backend creates the needed directory structure when not present - it is clearly not that simple with a database table - but can’t e somehow mimic that behavior here?

robertlemke:
I understand that, from a developer's / maintainer's standpoint it is not nice to have tables which are not used. Clean, pure, I agree.

Just let's look at it from a user's point of view. If you don't have a persistent file system, where do you store your encryption key? Your OAuth keys? You don't want them to be wiped away on the next deployment or container restart.

Now, there are a lot of good solutions to this problem, mostly dealing with it programatically tailored to your specific environment. Or using other storages, like Google Cloud Storage. PDO is certainly not my favorite storage for that purpose. The reason why I brought this up nevertheless is, that it's a drop-in / dont-let-me-think solution for Neos. We do know that people will have a PDO database when they run Neos, so we can prepare this as a persistent cache out of the box.

Practically, what's so bad about having a _neos_flow_cache_entries_ and _neos_flow_cache_tags_ table which possibly remains empty? You can still use the Flow Cache package without PDO, because you don't have to run the migration.

From experience with our Beach projects I realized that having a magic / automatic method (see https://github.com/flownative/beach-flow-companion/blob/master/Classes/Cache/PdoBackend.php) is error prone. I tried to mimic what we do with the filesystem based cache, but it still will stumble once in a while - you basically need to re-implement what doctrine migrate would do, because you possibly need to migrate from one structure to another (recently we had to).

And how about a *:setup command? Certainly, but – see above – it would basically need to be able to do the same like doctrine:migrate - keep track of which migrations have been applied already, be able to revert and so on. And you would need an extra – custom – step in your deployment.

Coming back to the user's perspective: I don't really want or need a PDO Backend. But users need to store encryption keys and some more. How can we achieve that otherwise without weeks of programming?

bwaidelich:
For me it‘s not so much about being „pure“. I just dislike a growing amount of tables. I always found it very tedious to navigate the myriads of tables in TYPO3 databases. (And I‘m guilty too with the ˋObjectPatMappingˋ entity).
Another option might be to extract the PdoBackend including migrations to a separate package.
But if you guys agree I could take care of the commands. It‘s not too much wirk actually because we have that all in place for the ES package (including schema migrations).

But obviously there are worse things. It‘s not the end of the world to have two additional tables IMO

kitsunet:
I don't really mind either way, the tables are fine for me. Important point though
a) the array_search for `cache` and `tags` seems rather generic, I am pretty sure that might lead to problems with 3rd party tables existing. Worst case one of the two is created and the other isn't...
b) The Neos.Cache package is not a flow package, having migrations in there is questionable in terms of separation of the packages, but if these migrations are picked up I am ok with that.

albe:
The table names `cache` and `tags` where probably really a bad idea, yeah. Much too generic and potential to colllide.

So what do we have?

- there may already be tables existing, that are used with this backend for persistent caching, i.e. we can not allow the data inside to be lost

- there may already be tables existing, but for a whole different purpose/with different schema. We may not mess with those tables

- no tables exist, in which case we do want to create them automatically to make it easier to use this backend (even if they stay empty)

robertlemke:
I don't want to annoy you, but please, let's quickly discuss this again from the user's point of view: does anyone have a good solution for storing encryption keys (and the like) out of the box for environments without a persistent file system? Maybe you know something I haven't thought of and we can skip this PDO backend alltogether.

kitsunet:
Nope, there isn't. Unless we specifically allow setting the ecryption key via environment variable and that's the only thing a storage is needed for. I realise my comment about the array_search is not as big an issue really. If you have those tables that's probably fine. I thought it was a fuzzy search but the tables are actually called like that :/

albe:
I like the idea to have the encryption key from envvar, that is probably a valid use case. That alone won't make this issue invalid though.

Otherwise Christian is right. Better not introduce this migration in a Bugfix release. It's not a bug, it's a feature ;)

aertmann:
For reference, this package https://github.com/Flowpack/jobqueue-doctrine provides a setup command to generate the necessary tables.

bwaidelich:
> For reference, this package https://github.com/Flowpack/jobqueue-doctrine provides a setup command to generate the necessary tables.

yes, it's a similar implementation to the one in the neos/eventsourcing package mentioned above.
I could not resist and started to work on a possible solution: #1384
It needs some more love but works in general and the `cache:list` and `cache:show` commands are pretty useful anyways IMO

bwaidelich:
*ping*

kdambekalns:
Let's go for #1384 instead of this one.

PR #1375, opened 2018-08-23
TASK: Expose target-locale, original and product-name in XliffParser

dfeyer:


dfeyer:
I'm currently working on a community package to help manage translation in sync with NodeType configuration in Flow. Missing property like the target local make it hard/hacky how to define if it's the base translation, or if you need to keep the file in sync with the source locale, ... Product name allow to detect when translation are overriden in a package, ...

PR #1376, opened 2018-08-24
TASK: Tweak docblocks in Neos.Cache classes

kdambekalns:
Non-functional fixes to docblocks, fixing a few typos, adding a few
@throws annotations and doing some cleanup here and there.

### DIFF
--- a/Neos.Cache/Classes/Backend/PdoBackend.php
+++ b/Neos.Cache/Classes/Backend/PdoBackend.php
@@ -10,6 +10,7 @@ namespace Neos\Cache\Backend;
  * information, please view the LICENSE file which was distributed with this
  * source code.
  */
+
 use Neos\Cache\Backend\AbstractBackend as IndependentAbstractBackend;
 use Neos\Cache\Exception;
 use Neos\Cache\Exception\InvalidDataException;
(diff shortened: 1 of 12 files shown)


Neos.Cache/Classes/Backend/PdoBackend.php
```diff
@@ -112,11 +113,12 @@ public function initializeObject()
      * @param string $entryIdentifier An identifier for this specific cache entry
      * @param string $data The data to be stored
      * @param array $tags Tags to associate with this cache entry
-     * @param integer $lifetime Lifetime of this cache entry in seconds. If NULL is specified, the default lifetime is used. "0" means unlimited liftime.
+     * @param integer $lifetime Lifetime of this cache entry in seconds. If NULL is specified, the default lifetime is used. "0" means unlimited lifetime.
      * @return void
      * @throws Exception if no cache frontend has been set.
      * @throws \InvalidArgumentException if the identifier is not valid
      * @throws InvalidDataException if $data is not a string
+     * @throws \Neos\Utility\Exception\FilesException
```
albe: Is this really correct? FilesException in a DB based backend?

kdambekalns: Yes, because it might create an SQLite DB, in the filesystem…

albe: Okay, can we somehow document that then? `@throws ... when using SQLite storage and the according file operation fails`

PR #1377, opened 2018-08-27
TASK: Speed up lookups in IdentityRoutePart

kdambekalns:
This adds a second index to the database table in which the
ObjectPathMapping instances are stored. It speeds up find operations
for which the existing primary key does not help.

With enough data (250k entries for ObjectPathMapping) this speeds up
page delivery by about 60%.

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/ObjectPathMapping.php
+++ b/Neos.Flow/Classes/Mvc/Routing/ObjectPathMapping.php
@@ -19,6 +19,11 @@ use Neos\Flow\Annotations as Flow;
  * This contains the URI representation of an object (pathSegment)
  *
  * @Flow\Entity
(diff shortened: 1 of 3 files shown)


DavidSporer:
Looks good for me as well :-)
I can also confirm from our production environment that it works.

kitsunet:
Yes, looks good, BUT again migration in released feature version :(
And apparently the migrations version we use in 4.3 doesn't know about the AbstractMigration, I assume it should be the now deprecated `Doctrine\DBAL\Migrations\AbstractMigration` still.

kdambekalns:
Fixed the `AbstractMigration`. About schema migrations: this is ok according to our rules at https://www.neos.io/download-and-extend/release-process.html, since no data loss and/or downtime are caused.

Issue #1359, opened 2018-08-08
Non-deterministic test


albe:
```
There was 1 failure:
1) Neos\Flow\Tests\Functional\Security\Authentication\Provider\PersistedUsernamePasswordProviderTest::authenticationWithCorrectCredentialsResetsFailedAuthenticationCount
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'2018-08-08T18:05:10+02:00'
+'2018-08-08T18:05:09+02:00'
/home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Flow/Tests/Functional/Security/Authentication/Provider/PersistedUsernamePasswordProviderTest.php:123
/home/travis/build/neos/flow-development-distribution/Packages/Libraries/phpunit/phpunit/phpunit:53
```

The test is doing a comparison on dates based on the assumption that the test runs quickly enough (within 1s), which as shown above is not always true.

```php
$this->assertEquals((new \DateTime())->format(\DateTime::W3C), $account->getLastSuccessfulAuthenticationDate()->format(\DateTime::W3C));
```

PR #1378, opened 2018-08-29
BUGFIX: Extend the expected exceptions for missing templates and sections

albe:
typo3fluid/fluid expects specific exceptions to be thrown to implement
the feature of optional sections and partials. Neos.FluidAdaptor has to
throw these exceptions or derivates of them. Otherwise the exceptions won't
be catched and displayed to the user.

fixes: #1347 

Follow-Up for #1348 with correct target.


albe:
Merging with previous approvals once travis passes

albe:
Ugh... previous test failure see #1359

PR #1379, opened 2018-08-29
Revert "BUGFIX: Extend the expected exceptions for missing templates and sections"

albe:
Reverts neos/flow-development-collection#1348


albe:
Should have gone into 4.3

PR #1381, opened 2018-08-29
FEATURE: Allow overriding the LockHoldingPage via environment variable

### LINKED ISSUE #585: Make LockManager's template ("LockHoldingStackPage") configurable
Jira issue originally created by user cerlestes:

The attached patch makes the so called 'LockHoldingStackPage', that is displayed by the LockManager when the site is locked during compile time, configurable via Settings.yaml

 Atachements:

Jira Image 

 Jira-URL: https://jira.neos.io/browse/FLOW-364

albe:
You can now specify the environment variable FLOW_LOCKHOLDINGPAGE in order
to override the path to the page shown, when the site is locked, relative
to the FLOW_ROOT_PATH.

Resolves #585


jonnitto:
Can you make this PR against the 5.2 branch? I would like to have this in the next release, as it solves a painpoint. Ping also someone for a review.

albe:
Yep, just found this missing myself. Will push a quick doc update with the feature today.

Neos.Flow/Classes/Core/LockManager.php
```diff
@@ -54,6 +54,7 @@ public function __construct()
         $lockPath = $this->getLockPath();
         $this->lockPathAndFilename = $lockPath . md5(FLOW_PATH_ROOT) . '_Flow.lock';
         $this->lockFlagPathAndFilename = $lockPath . md5(FLOW_PATH_ROOT) . '_FlowIsLocked';
+        $this->lockHoldingPage = defined('FLOW_LOCKHOLDINGPAGE') ? FLOW_PATH_ROOT . FLOW_LOCKHOLDINGPAGE : FLOW_PATH_FLOW . 'Resources/Private/Core/LockHoldingStackPage.html';
```
albe: Note: Maybe make that relative to `FLOW_PATH_PACKAGES` instead? Guess there's no much use in putting the page directly into the flow root.

albe:
Found this to be the best fitting place and since the useable environment variables do not seem to be documented anywhere, I added a whole short section on them all. @daniellienert

Issue #1383, opened 2018-08-31
Automatic Cache Setup


bwaidelich:
Some Cache Backends currently require manual setup.

This is not only tedious but it can lead to exceptions that only occur once a cache is accessed.
And because some caches are required very early in the boot sequence this can lead to a broken CLI.

Especially the PdoBackend is troublesome: It automagically creates required tables when using SQLite but otherwise it leads to a
```
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'cache' doesn't exist
```
PDO Exception as soon as it's accessed.

A solution could be to add tables with the default names via migration, like proposed with #1374 
This is a quick solution but it has the drawback that it creates those tables whether they are needed or not. Also it won't solve the issue with other cache backends that require setup and/or health check

PR #1384, opened 2018-08-31
FEATURE: Setupable Cache Backends

bwaidelich:
Introduces two new interfaces `WithSetupInterface` and `WithStatusInterface`
that can be implemented by cache backends in order to allow them to be
set up / investigated via CLI.

For this the following commands have been added:

* `cache:list`
* `cache:show`
* `cache:setup`
* `cache:setupall`

Resolves: #1383


bwaidelich:
image

image

image

bwaidelich:
Some feedback would be nice :)

albe:
Looking good from the screenshot and a rough look over the code. It does add a new system that the developer has to take care of/be aware of, but at least there's a tool for him.

I'm not soooo sure about the `Setupable` naming, at least it's not in the oxford dictionary: https://en.oxforddictionaries.com/search?utf8=%E2%9C%93&filter=dictionary&query=setupable
Don't have a good alternative at hand though.

I think the status check for the filebackend can be difficult, because it is done from CLI, so even if the check that the cache folder exists and is writeable succeeds there, it does not necessarily mean it will from a web request. Can we reliably do a local web request that does the check and returns the result?

bwaidelich:
@albe Thanks for the feedback!

> It does add a new system that the developer has to take care of/be aware of

Well, frankly it can be ignored completely and everything works as before. But, good point, we could improve the exception messages along the lines of `The cache table "foo" does not exist, did you run the cache:setup command?` to make this easier to use

> I'm not soooo sure about the Setupable naming

Me neither :)
Maybe someone has a better idea?

> Can we reliably do a local web request that does the check and returns the result?

That would probably require a lot of infrastructure (web endpoint, etc.) that can break itself. IIRC the error messages are pretty clear already in case of missing file permissions (otherwise we should improve those). By default the cache directories are created from the CLI anyways

albe:
Just found `Initializable` https://stackoverflow.com/questions/392513/can-something-be-initializable - at least it's already used somewhere, though... one of the comments there has a point: "it sounds like an interface to something that should just happen in the constructor"
Also, for our case it's not so much something that "can/may be setup/initialized", but that "needs to be" done once per installation.

bwaidelich:
> Just found Initializable [...] at least it's already used somewhere, though

So is `Setupable` https://github.com/search?q=setupable&type=Code :D

albe:
```
Languages
440 Swift
152 Java
85 C#
```
looks like the naming comes from a specific school of language(s) :laughing: 

But okay, you got a point there ;)

kitsunet:
I like it, left one comment though. As for the name, I would also prefer something else while not having a real good suggestion. What about `SetupEnabledInterface` or `SetupRequiredInterface` or even `WithSetupInterface`.

bwaidelich:
@kitsunet thanks for the feedback

> left one comment though

Which comment was that? I fail to find it?!

> What about SetupEnabledInterface

I like that. At least it's better than setupable or initializable because it consists of actual words :)
I will adjust the PR asap

bwaidelich:
> I like SetupRequiredInterface - because it shows that the setup is a required step

I thought so too, but it is not required necessarily. i.e. FileBackends should implement the interface but by default they would work anyways

Neos.Cache/Classes/Backend/SetupEnabledBackendInterface.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public function setup(): Result;
+
+    /**
+     * TODO document
+     *
+     * @return Result
+     * @api
+     */
+    public function getStatus(): Result;
```
kitsunet: getStatus outside of the interface (so in the implementation) might be a bit unspecific. Could also be the current status of the cache (or cache content) or any other kind of status. So in that case I would probably name it getSetupStatus ? Even if that sounds a bit awkward.
But I could very well imagine having a getStatus to ask if the cache is available right now (eg. for redis) at a later point.

kitsunet:
Added the comment I failed to submit earlier. Sorry...

Neos.Cache/Classes/Backend/SetupEnabledBackendInterface.php
bwaidelich: I agree but on the other hand:
I played around with the RedisBackend and implemented the `getStatus()` to render details about the server version etc. It's quite useful IMO.
So **suggestion:**
I split the interface into `BackendWithSetupInterface` and `BackendWithStatusInterface`. Some like the PdoBackend can implement both and the ones that can't be set up via PHP (i.e. Redis) can only implement the status interface (and we can at any point extend `getStatus()` to return more infos about the cache).
Do you agree to that approach?

Neos.Cache/Classes/Backend/PdoBackend.php
```diff
@@ -464,4 +506,134 @@ protected function context(): string
...
         return $this->context;
     }
+
+    /**
+     * TODO document
+     *
+     * @return Result
+     * @api
+     */
+    public function getStatus(): Result
```
albe: Thinking again about this, what if we instead of adding all those `BackendWithSetupInterface` and `BackendWithStatusInterface` responsibilities, we go another route and solve this by composition instead? So build a `PdoBackendSetup` class that receives the `PdoBackend` instance and then does the setup for this backend - it only needs access to the table names and the databaseHandle. This way we could clearly separate the resposibilities for setting up and getting the status (monitoring), instead of loading them onto the sometimes already big Backend implementation classes and increasing the list of interfaces a backend implements. WDYT?

Neos.Cache/Classes/Backend/PdoBackend.php
```diff
@@ -98,12 +119,27 @@ public function setPassword(string $password)
...
     /**
-     * Initialize the cache backend.
+     * Sets the name of the "cache" table
      *
+     * @param string $cacheTableName
      * @return void
+     * @api
      */
-    public function initializeObject()
+    public function setCacheTableName(string $cacheTableName)
```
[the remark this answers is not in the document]
bwaidelich: But besides: Why do you think that the setters being public can cause issues?

Neos.Cache/Classes/Backend/PdoBackend.php
bwaidelich: Usually I'm always in favor of composition. In this case it wouldn't replace inheritance though but just add another abstraction. And in this case I think it wouldn't improve things really because IMO it's the responsibility of the backend implementation to know about its status and the code needs to know a lot of the inner workings per definition.

Neos.Cache/Classes/Backend/PdoBackend.php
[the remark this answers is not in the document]
albe: "If it's possible, it will happen" - we just shouldn't make changing such options public, if it is not intended.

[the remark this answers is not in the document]
bwaidelich: I agree. I'm just not sure if there are other use cases for this (i.e. tests) that currently expect the public setters in the Backends. I'm happy to make them all protected and find out but it's out of scope of this PR to change this practice IMO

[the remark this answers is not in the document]
bwaidelich: I just checked again and there are already other backend with protected setters so apparently this is not expected to be public. I'll change the new setters to be protected. too!

Neos.Cache/Classes/Backend/PdoBackend.php
albe: I was more coming from the Single Responsibility Principle side of things. Setting itself up and watching it's status is not the primary responsibility of a Cache Backend. It was just an idea though on how to aproach this in a way, that will prevent adding another ~200 LOCs to an already rather complex backend implementation.

bwaidelich:
OK, here's a new version with the interfaces split up.
**Note:** I left the cache backend implementations as untouched as possible in order to make it easier to review this one and to avoid regressions.
If this gets accepted I'll create a follow-up addressing the following issues:
* make setters protected (as suggested by @albe above)
* initialize the backends lazily to make them more resilient & performant

A comment regarding the latter: In the current implementation the CLI commands don't help with a broken cache configuration because exceptions are thrown all over the place. If we only throw exceptions when *setting* caches we can work around this and finally be able to display a proper warning if redis authentication failed, there are insufficient file permissions, ...

albe:
Thanks, looks good by reading. Will try to take a closer look and test this locally.

kitsunet:
Oh my, I don't know how I could miss this, but this created a direction dependency on Doctrine DBAL in the neos/cache package, which a) is not declared in the compser.json and b) isn't really in line with it being an independent package. :( IMHO we should move that part over to Flow then...

PR #1387, opened 2018-09-05
BUGFIX: correctly handle 410 redirect exceptions

dimaip:
The redirect package throws exceptions on 410, but currently they are not caught, the way 404 exceptions are handled.

**Checklist**

### DIFF
--- a/Neos.Flow/Configuration/Settings.yaml
+++ b/Neos.Flow/Configuration/Settings.yaml
@@ -75,7 +75,7 @@ Neos:
         renderingGroups:
 
           notFoundExceptions:
-            matchingStatusCodes: [404]
+            matchingStatusCodes: [404, 410]
             options:
               logException: FALSE
               templatePathAndFilename: 'resource://Neos.Flow/Private/Templates/Error/Default.html'
@@ -271,7 +271,7 @@ Neos:
         # You can list any paths that you want to include in scanning for locales relative to the "Resources"
         # directory of each package. Setting a path to FALSE will skip scanning this folder.
         # Additionally, you can define patterns that you want to further skip inside the scanned paths.
-        #
+        #/
         # Example:
         #   includePaths:
         #     '/Public/': TRUE


johannessteu:
What we did in our project is:
    ```
    error:
      exceptionHandler:
        renderingGroups:
          goneExceptions:
            matchingStatusCodes: [410]
            options:
              logException: false
              templatePathAndFilename: 'resource://Neos.Flow/Private/Templates/Error/Default.html'
              variables:
                errorDescription: 'Sorry, the page you requested has gone.'
              layoutRootPath: 'resource://Neos.Neos/Private/Layouts/'
              format: html
```

as `notFoundException` is not really correct and the error description might be misleading

**Comment by bwaidelich** [comment]
Good point, but I don't think that this requires a different error message to be honest. The page was not found. That it is _gone_ is a technical detail (that is reflected via the status code)

**Comment by dimaip** [comment]
Don't know how that would affect 3rd-party extensions that don't expect yet one more renderingGroup. E.g. it seems that packages like Moc.NotFound also put it into the same renderingGroup: https://github.com/mocdk/MOC.NotFound/blob/master/Configuration/Settings.yaml#L7

**Comment by johannessteu** [comment]
But a page with a 404 message is just not the same as a 410. So i'm in favor to make it exact and seperate. The status code and the error message is at all not just an idicator for google but also for the user. The page was not found? Do i have a typo?! Page is gone -> ok, it's deleted.

File: Neos.Flow/Configuration/Settings.yaml [hunk 1]
```diff
@@ -271,7 +271,7 @@ Neos:
         # You can list any paths that you want to include in scanning for locales relative to the "Resources"
         # directory of each package. Setting a path to FALSE will skip scanning this folder.
         # Additionally, you can define patterns that you want to further skip inside the scanned paths.
-        #
+        #/
```
aertmann: assuming this is a typo

aertmann:
hmm I personally think showing it together with page not found is what most people would want, as long as it's possible to separate them like @johannessteu prefers

PR #1388, opened 2018-09-05
BUGFIX: Add nullable parameter detection

ComiR:
Nullable paramters (type prepended with `?`) were introduced with PHP 7.1
see https://secure.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types

We need to support a syntax like `functionName(?string $param)`. Without this fix, this function will get extended as `functionName(string $param)`, which is incompatible.

Starting with PHP 7.1 `functionName(string $param = null)` can also be written as `functionName(?string $param = null)` (but not as ~~`functionName(?string $param)`~~).

For the upmerge with Flow >= 5.0, the PHP version check in line 275 can be removed since at least PHP 7.1 is required there.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethod.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethod.php
@@ -250,6 +250,7 @@ class ProxyMethod
     {
         $methodParametersCode = '';
         $methodParameterTypeName = '';
+        $nullableSign = '';
         $defaultValue = '';
         $byReferenceSign = '';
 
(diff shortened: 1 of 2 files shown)


ComiR:
The test used wrong `allowsNull` values for the parameter info arrays. I updated them to the correct value, dumped by
```php
/**
 * @see `eval`d class definition in test method
 */
class TestClassX {
    public function foo(
        $arg1,
        array $arg2,
        \ArrayObject $arg3,
        $arg4= "foo",
        $arg5 = TRUE,
        array $arg6 = array(TRUE, 'foo' => 'bar', NULL, 3 => 1, 2.3)
    )
    {}
}


$reflectedMethod = new \ReflectionMethod(TestClassX::class, 'foo');
foreach ($reflectedMethod->getParameters() as $parameter) {
    echo $parameter->getName() . ': ';
    var_dump($parameter->allowsNull());
}
```
btw: I bet the `@param` annotation should be for the method, not for the class - but it's irrelevant for the test anyway...

PR #1389, opened 2018-09-06
TASK: Security code refactoring

kitsunet:
This is a comprehensive refactoring of internal code, nothing
should change on the outside.

Mostly reducing nested structures and complexity, breaking
up larger methods and avoiding complicated structures.


Neos.Flow/Classes/Package.php
```diff
@@ -132,5 +134,8 @@ public function boot(Core\Bootstrap $bootstrap)
 
         $dispatcher->connect(Persistence\Doctrine\PersistenceManager::class, 'allObjectsPersisted', ResourceRepository::class, 'resetAfterPersistingChanges');
         $dispatcher->connect(Persistence\Generic\PersistenceManager::class, 'allObjectsPersisted', ResourceRepository::class, 'resetAfterPersistingChanges');
+
+        $dispatcher->connect(AuthenticationProviderManager::class, 'successfullyAuthenticated', Context::class, 'refreshRoles');
+        $dispatcher->connect(AuthenticationProviderManager::class, 'loggedOut', Context::class, 'refreshTokens');
```
bwaidelich: This looks like a new behavior at first sight!? could you elaborate?

Neos.Flow/Classes/Security/Aspect/LoggingAspect.php
```diff
@@ -81,17 +85,26 @@ public function logManagerLogout(JoinPointInterface $joinPoint)
...
-        $accountIdentifiers = [];
-        foreach ($securityContext->getAuthenticationTokens() as $token) {
-            /** @var $account Account */
-            $account = $token->getAccount();
-            if ($account !== null) {
-                $accountIdentifiers[] = $account->getAccountIdentifier();
-            }
-        }
+
+        $accountIdentifiers = array_reduce($securityContext->getAuthenticationTokens(), [$this, 'reduceTokenToAccountIdentifier'], []);
```
bwaidelich: Is there a specific reason why you use `array_reduce` here (that needs to copy & mutate the array for every iteration?) `array_map` would be the better fit here IMO. Besides (nitpick) do we really need the protected `reduceTokenToAccountIdentifier` method? It seems to be only needed here so I'd prefer a closure or at least a private method.
This could work and is not less comprehensible IMO:

```
$accountIdentifiers = [];
array_walk($securityContext->getAuthenticationTokens(), function(TokenInterface $token) use (&$accountIdentifiers) {
    $account = $token->getAccount();
    if ($account === null) {
        return;
    }
    $accountIdentifiers[] = $token-> getAccountIdentifier();
});
```

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
```diff
@@ -293,99 +292,12 @@ protected function emitLoggedOut()
...
-     * Builds the provider and token objects based on the given configuration
+     * Signals that authentication commenced and at least one token was authenticated.
      *
      * @return void
-     * @throws Exception\InvalidAuthenticationProviderException
-     * @throws Exception\NoEntryPointFoundException
+     * @Flow\Signal
      */
-    protected function buildProvidersAndTokensFromConfiguration()
+    protected function emitSuccessfullyAuthenticated()
```
bwaidelich: Is that new? If so I'd vote for moving that to a separate PR in order to keep this one a pure refactoring and to make it easer to review, test & debug

Neos.Flow/Classes/Security/Aspect/LoggingAspect.php
kitsunet: I guess it's a matter of personal preference. I detest reference and array_walk, for me they are smells and easily lead to side effects. array_reduce is a purer way of manipulation IMHO. For map/filter to work you would need to map, filter and map again, and I wasn't exactly a fan of that as it felt to be too many operations, but you are right that this would be even nicer. I could change that though if you like.

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
kitsunet: This is needed together with the connections in the `Context` (see your comment from Package.php) in order to separate AuthenticationManager and Context and not have them cross depend on each other. It's the biggest part of this refactoring while everything else is really just moving code around.

Ofc I could try to split that into two PRs, if you really think that's needed.

Neos.Flow/Classes/Security/Aspect/LoggingAspect.php
bwaidelich: OK, let's try to take personal preference out of the game: Out of curiosity I wrote a little benchmark and while the `array_walk` approach is a little faster a good old `foreach` is more than twice as fast and IMO even easier to read:
```
$accountIdentifiers = [];
foreach ($securityContext->getAuthenticationTokens() as $token) {
    $account = $token->getAccount();
    if ($account !== null) {
        $accountIdentifiers[] = $account->getAccountIdentifier();
    }
}
```
You suggested that in an unrelated change (you know where *g) and I think it just makes sense here, as well :)

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
bwaidelich: > I could try to split that into two PRs, if you really think that's needed

I really don't want to decide that for you. It's just a concern I had when doing a first review: mixing behavioral changes into refactorings has introduced a lot of regressions in the past. But if it's a requirement for a clean separation I'm the last to block this :)

kitsunet:
I reduced this to the mere refactoring parts now. Will adjust the above comment ASAP and bring the split of Context and AuthenticationManager up as separate PR soon.

kitsunet:
#1407 is the behaviour change.

PR #1391, opened 2018-09-07
BUGFIX: Flow CLI command warns of mismatching php version

beheist:
If Flow builds a PHP command for a subrequest, it uses the system default if nothing else is configured. With this change, we avoid Flow executing that request if it isn't explicitly configured to use that same PHP version internally too. This should avoid some errors especially in shared hosting scenarios for less experienced users.

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -701,7 +701,6 @@ class Scripts
     /**
      * @param array $settings The Neos.Flow settings
      * @return string A command line command for PHP, which can be extended and then exec()uted
-     *
      * @throws Exception
      */
     public static function buildPhpCommand(array $settings)
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -792,19 +792,19 @@ class Scripts
             return;
         }
 
-        exec($phpCommand . ' -r "echo phpversion();"', $output, $result);
+        exec($phpCommand . ' -r "echo PHP_VERSION;"', $output, $result);
 
         if ($result !== 0) {
             return;
         }
-        $runningPHPVersion = phpversion();
+
         $configuredPHPVersion = $output[0];
-        if ($runningPHPVersion !== $configuredPHPVersion) {
+        if (array_slice(explode('.', $configuredPHPVersion), 0, 2) !== array_slice(explode('.', PHP_VERSION), 0, 2);) {
             throw new Exception(sprintf('You are executing Neos/Flow with a PHP version different from the one Flow is configured to use internally. ' .
                 'Flow is running with with PHP "%s", while the PHP version Flow is configured to use for subrequests is "%s". Make sure to configure Flow to ' .
                 'use the same PHP version by setting the "Neos.Flow.core.phpBinaryPathAndFilename" configuration option to a PHP-CLI binary of the version ' .
                 '%s. Flush the caches by removing the folder Data/Temporary before executing Flow/Neos again.',
-                $runningPHPVersion, $configuredPHPVersion, $runningPHPVersion), 1536563428);
+                PHP_VERSION, $configuredPHPVersion, PHP_VERSION), 1536563428);
         }
     }
 
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -1,5 +1,4 @@
 <?php
-
 namespace Neos\Flow\Core\Booting;
 
 /*
@@ -716,6 +715,8 @@ class Scripts
             $subRequestEnvironmentVariables = array_merge($subRequestEnvironmentVariables, $settings['core']['subRequestEnvironmentVariables']);
         }
 
+        self::ensureCurrentPhpVersionMatchesConfiguredVersion($settings['core']['phpBinaryPathAndFilename']);
+
         $command = '';
         foreach ($subRequestEnvironmentVariables as $argumentKey => $argumentValue) {
             if (DIRECTORY_SEPARATOR === '/') {
@@ -740,29 +741,30 @@ class Scripts
             $command .= ' -c ' . escapeshellarg($useIniFile);
         }
 
-        self::ensureCurrentPhpVersionMatchesConfiguredVersion($command);
 
         return $command;
     }
 
     /**
-     * @param $command
+     * @param string phpBinaryPathAndFilename
      * @throws Exception
      */
-    protected static function ensureCurrentPhpVersionMatchesConfiguredVersion($command)
+    protected static function ensureCurrentPhpVersionMatchesConfiguredVersion($phpBinaryPathAndFilename)
     {
-        $runningPHPVersion = phpversion() . '-' . PHP_SAPI;
-        exec($command . ' -r "echo phpversion() . \'-\' . PHP_SAPI;"', $output, $result);
+        // Resolve any symlinks that the configured php might be pointing to
+        $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
 
-        if ($result === 0) {
-            $configuredPHPVersion = $output[0];
-            if (strcmp($runningPHPVersion, $configuredPHPVersion) !== 0) {
-                throw new Exception(sprintf('You are running the Flow CLI with a PHP version different from your system default. ' .
-                    'Flow has been run with "%s", while the system default is "%s". Make sure to configure Flow to use the same PHP ' .
-                    'binary by setting the Neos.Flow.core.phpBinaryPathAndFilename configuration option to "%s". Flush the caches by '.
-                    'removing the folder Data/Temporary before running ./flow again.',
-                    $runningPHPVersion, $configuredPHPVersion, PHP_BINARY), 1536303119);
-            }
+        // if the configured PHP binary is empty here, the file does not exist. We ignore that here because it is checked later in the script.
+        if (strlen($configuredPhpBinaryPathAndFilename) === 0) {
+            return;
+        }
+
+        if (strcmp(PHP_BINARY, $configuredPhpBinaryPathAndFilename) !== 0) {
+            throw new Exception(sprintf('You are running the Flow CLI with a PHP version different from the one Flow is configured to use. ' .
+                'Flow has been run with "%s", while the PHP version Flow is configured to use is "%s". Make sure to configure Flow to ' .
+                'use the same PHP binary by setting the "Neos.Flow.core.phpBinaryPathAndFilename" configuration option to "%s". Flush the ' .
+                'caches by removing the folder Data/Temporary before running ./flow again.',
+                PHP_BINARY, $configuredPhpBinaryPathAndFilename, PHP_BINARY), 1536303119);
         }
     }
 
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -799,7 +799,7 @@ class Scripts
         }
 
         $configuredPHPVersion = $output[0];
-        if (array_slice(explode('.', $configuredPHPVersion), 0, 2) !== array_slice(explode('.', PHP_VERSION), 0, 2);) {
+        if (array_slice(explode('.', $configuredPHPVersion), 0, 2) !== array_slice(explode('.', PHP_VERSION), 0, 2)) {
             throw new Exception(sprintf('You are executing Neos/Flow with a PHP version different from the one Flow is configured to use internally. ' .
                 'Flow is running with with PHP "%s", while the PHP version Flow is configured to use for subrequests is "%s". Make sure to configure Flow to ' .
                 'use the same PHP version by setting the "Neos.Flow.core.phpBinaryPathAndFilename" configuration option to a PHP-CLI binary of the version ' .
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -715,7 +715,7 @@ class Scripts
             $subRequestEnvironmentVariables = array_merge($subRequestEnvironmentVariables, $settings['core']['subRequestEnvironmentVariables']);
         }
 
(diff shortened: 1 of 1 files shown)


beheist:
This is that it looks like in the console:
image

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
bwaidelich: PhpStorm likes this additional space by default, but we don't (according to our CGL) You can change this in PhpStorm in the Code Style Settings

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -736,9 +740,32 @@ public static function buildPhpCommand(array $settings)
             $command .= ' -c ' . escapeshellarg($useIniFile);
         }
 
+        self::ensureCurrentPhpVersionMatchesConfiguredVersion($command);
+
         return $command;
     }
 
+    /**
+     * @param $command
```
bwaidelich: missing type hint

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -736,9 +740,32 @@ public static function buildPhpCommand(array $settings)
...
+     * @throws Exception
+     */
+    protected static function ensureCurrentPhpVersionMatchesConfiguredVersion($command)
+    {
+        $runningPHPVersion = phpversion() . '-' . PHP_SAPI;
+        exec($command . ' -r "echo phpversion() . \'-\' . PHP_SAPI;"', $output, $result);
```
bwaidelich: This would run an additional exec each time. Didn't you mean we could check against the `PHP_BINARY` to avoid that?

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -736,9 +740,32 @@ public static function buildPhpCommand(array $settings)
...
+
+        if ($result === 0) {
```
bwaidelich: an early return would improve readability IMO:
```
if ($result !== 0) {
    return;
}
```

bwaidelich: But then again - should we ignore that error silently?

beheist: We check that elsewhere in the script:
https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Core/Booting/Scripts.php#L360-L375

so I didn't think it's necessary to duplicate that logic there.

Neos.Flow/Classes/Core/Booting/Scripts.php
beheist: Yes, I can do that too. We can't compare the versions in that case, but that's not necessary I guess.

beheist:
I changed this to compare binary paths instead of versions, and adapted the error message slightly.
This check will only run if Flow executes a subrequest, but that's the only place where it matters, so that is probably OK.
image

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -736,9 +741,33 @@ public static function buildPhpCommand(array $settings)
...
+    {
+        // Resolve any symlinks that the configured php might be pointing to
+        $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
+
+        // if the configured PHP binary is empty here, the file does not exist. We ignore that here because it is checked later in the script.
+        if (strlen($configuredPhpBinaryPathAndFilename) === 0) {
+            return;
+        }
+
+        if (strcmp(PHP_BINARY, $configuredPhpBinaryPathAndFilename) !== 0) {
```
albe: I'll have to check if this works on windows.

beheist: So, yeah - this is what I thought would happen and probably also the entire reason we have that config setting.
image

I see two ways forward here:
1. Do the check only for CLI requests. This would make the error message disappear, but again open the possibility for hard-to-detect configuration errors
2. Do the check as-is for CLI requests, and do the check like before for non-CLI (run exec(), checking the PHP version of the configured PHP-CLI, and compare the actual version number with the running one). 

I prefer the second option, as it makes config errors impossible, gives the user a precise and actionable error message, and has a minimal performance impact (delays the building of proxies by one exec(), but since proxy building takes a few secs at minimum anyway, this should be hardly noticeable).

bwaidelich: I'm not so much concerned about performance. but option 2 spawns an additional process for every sub request. For packages that use that mechanism extensively (JobQueue, EventSourcing, ...) that doubles the amount of processes. I don't know how problematic that is, but we should evaluate that IMO.

beheist: But the process is only spawned if proxies have not been built yet - once proxies are compiled, we don't reach this section of code. Is this still a problem in this case (I don't know how the other packages work, so..)

bwaidelich: Ah OK, sorry, I only watched with one eye :)
The error message is much clearer, great improvement!

beheist: OK, removing WIP. Final reviews please :)

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -736,9 +741,73 @@ public static function buildPhpCommand(array $settings)
...
+
+        // Resolve any symlinks that the configured php might be pointing to
+        $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
+
+        // if the configured PHP binary is empty here, the file does not exist. We ignore that here because it is checked later in the script.
+        if (strlen($configuredPhpBinaryPathAndFilename) === 0) {
+            return;
+        }
+
+        if (strcmp(PHP_BINARY, $configuredPhpBinaryPathAndFilename) !== 0) {
```
helhum: Two questions here:

1. Why not directly always using `PHP_BINARY` for the sub process instead of relying on some config. AFAIK this also works reliably on Windows
1. Isn't `PHP_BINARY === $configuredPhpBinaryPathAndFilename` easier to grasp than using `strcmp`?

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -736,9 +741,73 @@ public static function buildPhpCommand(array $settings)
...
+        }
+
+        exec($phpCommand . ' -r "echo phpversion();"', $output, $result);
+
+        if ($result !== 0) {
+            return;
+        }
+        $runningPHPVersion = phpversion();
+        $configuredPHPVersion = $output[0];
+        if ($runningPHPVersion !== $configuredPHPVersion) {
```
helhum: This could be a quite intrusive limitations for some systems/ hosting environments.
Since PHP for web requests and PHP on cli are completely different binaries, in theory versions can also differ. E.g. PHP 7.1.3 for web requests and PHP 7.1.2 on CLI

With this change, such constellation would throw an exception. Right?

Neos.Flow/Classes/Core/Booting/Scripts.php
kitsunet: From what I remember the PHP_BINARY was unreliable on shared environments with multiple PHP versions or something like that? I don't really remember the case.

Neos.Flow/Classes/Core/Booting/Scripts.php
kitsunet: I agree, I don't think we need to be that strict. I would rather loosen it to same minor or something. Ideally we would go into the direction of https://github.com/kitsunet/flownative-webcompiler at some point to avoid the CLI thing altogether.

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -736,9 +741,73 @@ public static function buildPhpCommand(array $settings)
...
+    /**
+     * Compares the realpath of the configured PHP binary (if any) with the one flow was called with in a CLI request.
+     * This avoids config errors where users forget to set Neos.Flow.core.phpBinaryPathAndFilename in CLI.
+     *
+     * @param string phpBinaryPathAndFilename
+     * @throws Exception
+     */
+    protected static function ensureCLISubrequestsUseCurrentlyRunningPhpBinary($phpBinaryPathAndFilename)
+    {
+        // Do nothing for non-CLI requests
```
bwaidelich: Why not? Wouldn't it make sense to check this *especially* for commands triggered by a Web request?

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -736,9 +741,73 @@ public static function buildPhpCommand(array $settings)
...
+        if (PHP_SAPI !== 'cli') {
+            return;
+        }
+
+        // Resolve any symlinks that the configured php might be pointing to
+        $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
+
+        // if the configured PHP binary is empty here, the file does not exist. We ignore that here because it is checked later in the script.
+        if (strlen($configuredPhpBinaryPathAndFilename) === 0) {
```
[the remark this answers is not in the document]
bwaidelich: Good catch

Neos.Flow/Classes/Core/Booting/Scripts.php
bwaidelich: > I would rather loosen it to same minor or something

Yes, that's what we did in the Setup package. It's not a beauty, but it seems to work:
https://github.com/neos/setup/blob/master/Classes/Core/RequestHandler.php#L173-L271

> Ideally we would go into the direction of https://github.com/kitsunet/flownative-webcompiler

It would be great if compilation worked without additional requests (I hope this works memory wise, too). But CLI "sub requests" are useful anyways (i.e. for "poor mans" job queue, batched CLI commands, ...). So +1 for making this more stable anyways

Neos.Flow/Classes/Core/Booting/Scripts.php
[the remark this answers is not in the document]
helhum: Oh, and while reading [setup package code](https://github.com/neos/setup/blob/master/Classes/Core/RequestHandler.php#L173-L271), if open_basedir is configured, it may well be that the php binary path is not included, thus regular file functions would not work on that path.

So something like [this](https://github.com/neos/setup/blob/master/Classes/Core/RequestHandler.php#L256-L270) would be required to catch this case.

Neos.Flow/Classes/Core/Booting/Scripts.php
albe: PHP_BINARY is not reliable on Windows from Web. It will just point to the process that launched php (e.g. apache/nginx).

helhum: `PHP_BINARY` is *never* to rely on for web requests. It **is** reliable on cli and always points to the path of the currently in use PHP binary. This whole code is [only executed on cli](https://github.com/neos/flow-development-collection/pull/1391/files#diff-3dae3262c03fe101b952964ee70e2ddbR759). So why not just rely on `PHP_BINARY`, because it would just work.

Like that: 

```php
if (PHP_SAPI === 'cli') {
   return PHP_BINARY
}

return $settings['configuredPHPBinaryForWebRequests'];
```

This works reliably in practice across operating systems for me for years on a fairly large user base.

The only complaint I received was from a user that called the script with additional options for the php binary like so: `php -derror_reporting=0`, which is not easily solvable, except such options are configurable as well.

Neos.Flow/Classes/Core/Booting/Scripts.php
albe: See below. The check can not be done realiably from Web request, because PHP_BINARY is not reliable in that context. The best one can do is check if the version of the specified php binary matches the current php version.

Neos.Flow/Classes/Core/Booting/Scripts.php
albe: The two-lane way seems to work good. Just need to fix the strict version constraint.

Neos.Flow/Classes/Core/Booting/Scripts.php
albe: Yes, definitely needs to be less strict. Minor version match is enough.

Neos.Flow/Classes/Core/Booting/Scripts.php
albe: But this method is not responsible for finding the php binary, but rather check if the found (configured) php binary matches the one that this command is currently run with to detect configuration errors :) Also, we probably shouldn't use PHP_BINARY in the one case (CLI) and rely on a configuration setting (`phpBinaryPathAndFilename`) in the other (Web), even if that would make it work in at least one case reliably.

albe:
@bwaidelich @kitsunet Can one of you approve this thing and merge?

bwaidelich:
If this is a bugfix it should go to the lowest supported branch, no?
And if it's considered too risky it could go to master..

Apart from that: It looks good to me by reading but I can't test this really (at least not on Windows) so I don't dare to merge this

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -700,6 +701,8 @@ protected static function buildSubprocessCommand($commandIdentifier, array $sett
     /**
      * @param array $settings The Neos.Flow settings
      * @return string A command line command for PHP, which can be extended and then exec()uted
+     *
```
```suggestion
```

albe:
> If this is a bugfix it should go to the lowest supported branch, no?

Correct, which is 4.3 as of the time now :) Anything lower is security fixes only.

bwaidelich:
ah right.. Flow != Neos :)

albe:
So, anything left keeping this from getting in? I'd really like to get this merged as is to improve the situation. We can still discuss on a better way of solving things in a new issue if needed.

albe:
Ok, I'll take responsibility for this one, if it messes things up. I'll merge and provide follow-up fixes where necessary.

kitsunet:
The import of `use Neos\Flow\Exception;` in this change breaks things because the SubProcessException is no longer found, masking an error in the compile step with a PHP fatal error...

PR #1393, opened 2018-09-12
TASK: Require composer autoloader in entry scripts

helhum:
Instead of requiring the bootstrap class file and
then requiring Composer autoloader in the Bootstrap constructor,
we can require the autoloader early in the two entry scripts.

This makes it possible to set FLOW_CONTEXT through .env files
e.g with packages like helhum/dotenv-connector

### DIFF
--- a/Neos.Flow/Classes/Core/Bootstrap.php
+++ b/Neos.Flow/Classes/Core/Bootstrap.php
@@ -79,11 +79,12 @@ class Bootstrap
      * Constructor
      *
      * @param string $context The application context, for example "Production" or "Development"
+     * @param \Composer\Autoload\ClassLoader $composerAutoloader Composer autoloader
      */
-    public function __construct(string $context)
+    public function __construct(string $context, \Composer\Autoload\ClassLoader $composerAutoloader = null)
     {
-        // Load the composer autoloader first
-        $composerAutoloader = require(__DIR__ . '/../../../../Libraries/autoload.php');
+        // Load the composer autoloader first if not provided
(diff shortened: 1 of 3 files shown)


helhum:
> if that is breaking

Current state is that it would work with an old index.php as well, as the code to include the autoloader in Bootstrap constructor is still there, in case it is not provided by the entry script

helhum:
Is the `Web/index.php` considered "API" in the sense that people are advised to put code in there to overcome certain limitations?

If not, I'd rather generate this file on each `composer dumpautoload`. What would then be possible is calculating the path to the vendor dir and have arbitrary/configurable vendor dir/web directory locations.

I have code for something for that up my sleeve, including removing the need to remove the need to add the [script section](https://github.com/neos/flow-base-distribution/blob/24b98b16e4ca27d17076f8a3233706c41f37060b/composer.json#L47-L51) to the root composer.json.

Both combined would enable users to just `composer require neos/flow` and have a working version of the framework …

So if you want me to work in this area, just let me know.

kitsunet:
> Current state is that it would work with an old index.php as well, as the code to include the autoloader in Bootstrap constructor is still there, in case it is not provided by the entry script

Good point! :)

> Is the Web/index.php considered "API"

TBD. For me no, but I know that in fact people configured things in there sometimes if they couldn't set environment variables.

helhum:
Thanks for the approval! Looking forward to see it merged.

> but I know that in fact people configured things in there sometimes if they couldn't set environment variables

OK, so if you could set (all) env vars easily with helhum/dotenv-connector, there would be no need to modify this file :)

helhum:
Thanks!

PR #1394, opened 2018-09-13
 BUGFIX: doctrine:migrationgenerate won't move file to selected package 

tvt:
Fixes an issue where running doctrine:migrationgenerate would never move the migration-file to the selected package. After doctrine:migrationgenerate has generated a migration, it asks whether the migration-file should be moved to a specific package. No matter what you choose, it would assume you chose "Don't Move".

Also fixes two related issues in the ConsoleOutput's select method:
- Wrong typehint on $default, breaking the default answer functionality
- Wrong phpdoc typehint on $attempts, as it is an integer, not a boolean.

I added a testcase and modified a couple of other testcases for the ConsoleOutput as well.


daniellienert:
Hey @tvt,

thank you very much for your first contribution to this project which is much appreciated!!
Your changes look great and fixes the problem.

Cheers 
Daniel

PR #1396, opened 2018-09-23
FEATURE: Support HTML5 dialog method

niko5:
Update to support 'dialog' method used in HTML5 <dialog> element
eg. `<form method="dialog" ...>`
See form submission point 20.
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-2 

(note I am not sure if any change required  for` protected function renderCsrfTokenField()`

**What I did**

**How I did it**

**How to verify it**

**Checklist**

### DIFF
--- a/Neos.FluidAdaptor/Classes/ViewHelpers/FormViewHelper.php
+++ b/Neos.FluidAdaptor/Classes/ViewHelpers/FormViewHelper.php
@@ -144,7 +144,7 @@ class FormViewHelper extends AbstractFormViewHelper
 
         if (strtolower($this->arguments['method']) === 'get') {
             $this->tag->addAttribute('method', 'get');
-        } else if($this->arguments['method']) === 'dialog') {
+        } elseif(strtolower($this->arguments['method']) === 'dialog') {
              $this->tag->addAttribute('method', 'dialog');
         } else {
             $this->tag->addAttribute('method', 'post');
--- a/Neos.FluidAdaptor/Classes/ViewHelpers/FormViewHelper.php
+++ b/Neos.FluidAdaptor/Classes/ViewHelpers/FormViewHelper.php
@@ -103,7 +103,7 @@ class FormViewHelper extends AbstractFormViewHelper
     public function initializeArguments()
     {
         $this->registerTagAttribute('enctype', 'string', 'MIME type with which the form is submitted');
-        $this->registerTagAttribute('method', 'string', 'Transfer type (GET or POST)');
+        $this->registerTagAttribute('method', 'string', 'Transfer type (GET or POST or dialog)');
         $this->registerTagAttribute('name', 'string', 'Name of form');
         $this->registerTagAttribute('onreset', 'string', 'JavaScript: On reset of the form');
         $this->registerTagAttribute('onsubmit', 'string', 'JavaScript: On submit of the form');
@@ -144,6 +144,8 @@ class FormViewHelper extends AbstractFormViewHelper
 
         if (strtolower($this->arguments['method']) === 'get') {
             $this->tag->addAttribute('method', 'get');
+        } else if($this->arguments['method']) === 'dialog'){
+             $this->tag->addAttribute('method', 'dialog');
         } else {
             $this->tag->addAttribute('method', 'post');
         }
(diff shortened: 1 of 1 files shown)


albe:
Hey and welcome! Thanks a lot for this! I wasn't even aware there was a "dialog" method for forms.

This could probably target 4.3 instead of just master. Also a small test that checks that the "dialog" method is accepted and rendered correctly would be cool. I don't think this needs some change for the CSRF Token rendering, though.

albe:
This needs to be rebased on 4.3 for retargeting. You can create a new PR or just force-push to this, then change the target branch.

niko5:
@albe any examples of similar tests that I could use as a template? I did of course test it by using in my project but don't think that would be appropriate.

albe:
You can take a look at the tests in https://github.com/neos/flow-development-collection/blob/master/Neos.FluidAdaptor/Classes/ViewHelpers/FormViewHelper.php for example.

However, after reading a bit more about method "dialog", this seems to be a purely client-side tech. So the question is, if this is really needed in the Fluid Form ViewHelper, since that is supposed to help with client->server side communication. For all other cases, one can simply just add a normal `<form method="dialog"...>` to the Fluid Template. Or what exactly would the use-case be? Maybe we can solve this with just documenting this instead of adding code?

niko5:
I did just use a normal form element when I discovered that the Form viewhelper doesn’t support it.

However I don’t quite follow your argument as the Fluid elements do have support for JavaScript ie client side technology already. I would think adding support for this keeps Fluid form viewhelper in line with standards? 

Thanks also for link to tests - I should have looked there already - I will take a look and see if I can see how I would add a test for this.

albe:
> However I don’t quite follow your argument as the Fluid elements do have support for JavaScript ie client side technology already. I would think adding support for this keeps Fluid form viewhelper in line with standards?

Well, yes, but the Fluid ViewHelpers are not meant to just duplicate the available standards in a slightly different way. With that aproach, we would just end up chasing behind the living standard that HTML has become and that is an uphill battle we can only lose. So maybe it is time to take a clearer stance on that. Anyway, not trying to devalue your contribution at all, just having some thoughts about the backgrounds and how to maybe approach the issue at hand from a different perspective.

Would love to hear the thoughts of @kitsunet and maybe @bwaidelich on this

bwaidelich:
At some point it was possible to do
```html
<f:form method="DELETE">
```
and it would render a hidden field with `name="__method" value="delete"` but apparently that was "lost in translation" with the move to TYPO3Fluid.

Anyways, for these cases we introduced the `additionalAttributes` argument back then.
So that **should** work:
```html
<f:form additionalAttributes="{method: 'dialog'}">
```
Unfortunately it does **not** and _that's_ the bug IMO.

So instead of adding more specific rules we could just change
```html
if (strtolower($this->arguments['method']) === 'get') {
    $this->tag->addAttribute('method', 'get');
} else {
    $this->tag->addAttribute('method', 'post');
}
```
to s.th. like
```html
if (strtolower($this->arguments['method']) === 'get') {
    $this->tag->addAttribute('method', 'get');
} elseif (!isset($this->arguments['additionalAttributes']['method'])) {
    $this->tag->addAttribute('method', 'post');
}
```

bwaidelich:
> What if we just pass the attribute as is, unless it is empty?

Yes, that would be an option. The reason why this check was added is:
* Unify the casing (no `method="POST"`) to reduce errors when consuming the value
* Render the hidden `__method` field if the method is not supported by Flow

Both not really important (especially because the 2nd was lost anyways apparently)

But I wonder.. Why use a VH at all if you just want to close a dialog?
`<form method="dialog"...` should just work fine!?

niko5:
using the VH with method="dialog" is still useful as you get all the background binding say to the eg. when using  objectName="theFormCommand" .

bwaidelich:
> using the VH with method="dialog" is still
> useful as you get all the background binding

Not to argue against you (I think we should fix the bug) but FYI: You could also use the `formmethod` attribute on the button input

niko5:
> You could also use the formmethod attribute on the button input

thanks I wil take a look at that too and see how that works.

in terms of the change will this be merged? do I need to do anything else?

PR #1399, opened 2018-10-03
FEATURE: Introduce an utility to add log environment from method name

daniellienert:
This utility can be used to easily add the log environment to log
messages using the PSR compatible logger.

Example:

```
$logger->debug('Running sub process loop.', LogEnvironment::fromMethodName(__METHOD__));
```


kitsunet:
I think compiled static methods must be public for reasons. I assume that's what broke the build.

daniellienert:
@kitsunet, thanks. In this case the problem actually was, that I accidentally submitted a changed logger call. In test cases `Bootstrap::$staticObjectManager` is not available and thus the tests failed. I now explicitly test for availability.

IMHO this one is good to be merged so that we can use the method in Log messages of the core packages.

PR #1401, opened 2018-10-04
TASK: Add support for DateTimeImmutable classes mapping with doctrine

albe:
This change correctly maps `DateTimeImmutable` property types with doctrine.

Depends on #1442

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
@@ -732,6 +732,9 @@ class FlowAnnotationDriver implements DoctrineMappingDriverInterface, PointcutFi
                         case 'DateTime':
                             $mapping['type'] = 'datetime';
                             break;
+                        case 'DateTimeImmutable':
+                            $mapping['type'] = 'datetime_immutable';
+                            break;
                         case 'string':
                         case 'integer':
                         case 'boolean':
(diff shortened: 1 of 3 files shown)


albe:
Damn :|

albe:
Apparently, `datetime_immutable` is only available as of Doctrine DBAL 2.6, which we only added in Flow 5.0, since the former requires PHP 7.1

https://github.com/doctrine/dbal/releases/tag/v2.6.0

So I'm going to retarget this and then hopefully the tests pass.

albe:
Argh, I discovered a bug in the TypeHandling. We currently try to parse types with a rather cumbersome regex, that eagerly matches `DateTime` and hence will parse a `DateTimeImmutable` var annotation as `DateTime` :(

albe:
Generally yes, but in this case I'm considering it something between BUGFIX and FEATURE - hence the label "TASK". It should have been there all along since we upgraded to DBAL 2.7 in the 5.0 release IMO.

If the others also regard this as FEATURE, we can of course only merge this into master.

bwaidelich:
> but in this case I'm considering it something between BUGFIX and FEATURE
I know, it's on the edge. But IMO there should be almost no tasks at all. It's almost always a (missing) feature or a bugfix. In this case I'd say it's the former.

Anyways, not to split hairs, and I'm fine with having that added in a patch level release (I thought it was a problem to backport it, but I misread your comment above)

albe:
> I thought it was a problem to backport it

Nope, this change is really easy still, but it can only go into 5.0 minimum. But it depends on #1442 anyway, so we need to resolve that first.

jonnitto:
@albe Can you fix the tests, so we can merge it?

albe:
Dang, tests failed because the tests were run based on 4.3 branch still. Rebased and force pushed.

albe:
Did upmerges to get #1442 into 5.0 and merged the change into this. Tests should now pass :crossed_fingers:

albe:
Merging with previous votes

PR #1402, opened 2018-10-08
TASK: Make ContentStream easier to use

bwaidelich:
This adds a new named constructor `fromContents()` to the `ContentStream` class making
it easier to instantiate it from a given string.

### DIFF
--- a/Neos.Flow/Classes/Http/ContentStream.php
+++ b/Neos.Flow/Classes/Http/ContentStream.php
@@ -39,13 +39,27 @@ class ContentStream implements StreamInterface
     protected $logger;
 
     /**
-     * @param string|resource $stream
+     * @param string|resource $stream a valid PHP resource or a filename supported by fopen (see http://php.net/manual/en/function.fopen.php)
      * @param string $mode Mode with which to open stream
      * @throws \InvalidArgumentException
      */
     public function __construct($stream, $mode = 'r')
     {
         $this->replace($stream, $mode);
+    }
+
+    /**
+     * Creates an instance representing the given $contents string
+     *
+     * @param string $contents
+     * @return self
+     */
+    public static function fromContents(string $contents): self
+    {
+        $handle = fopen('php://memory', 'r+');
+        fwrite($handle, $contents);
+        rewind($handle);
+        return new static($handle);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


kitsunet:
Just as a note, I wanted to avoid Flow specific extensions to the interface implementations to avoid breaking the interfaces promise of interop. Obviously a static constructor doesn'T count because when you construct a new object you already know which one you create (usually). So IMHO this should be fine.

Neos.Flow/Classes/Http/ContentStream.php
```diff
@@ -88,7 +102,7 @@ public function replace($stream, $mode = 'r')
     {
         $this->close();
         if (!is_resource($stream)) {
-            $stream = $this->openStream($stream, $mode);
+            $stream = @fopen($stream, $mode);
```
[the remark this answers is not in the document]
kitsunet: Makes sense.

Neos.Flow/Classes/Http/Helper/ArgumentsHelper.php
```diff
@@ -43,12 +43,10 @@ public static function buildUnifiedArguments(array $getArguments, array $postArg
     /**
      * @param string $content
      * @return ContentStream
+     * @deprecated since Flow 5.2, use ContentStream::fromContents()
```
albe: That one had a short lifetime :smile:

kitsunet: IMHO we could even remove it immediately.It's just in since the last release and the release notes stated that things are in motion atm and people shouldn't rely on stability on these APIs yet.

albe: Yeah, I remember that. Also doubt anyone has really jumped onto this API yet.

bwaidelich: I'd rather leave it in for now. It doesn't really harm and I'm pretty sure someone somewhere (for example we ourself in the core) already use it

Neos.Flow/Classes/Http/ContentStream.php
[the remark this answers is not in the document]
kdambekalns: But didn't we at some point agree to consider the shutup-op as genuinely evil?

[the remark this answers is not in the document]
bwaidelich: @kdambekalns We already us it in the core a lot because sometimes there is no alternative. Or what do you suggest to do instead?

PR #1404, opened 2018-10-09
BUGFIX: Do not join select property paths to embedded objects

### LINKED ISSUE #989: Querying a property of an embedded value object will fail
### Description

It is not possible to query a nested field of an embedded value object like this:
`$query->matching( $query->equals('valueObject.property', 'value')`.

### Steps to Reproduce

1. Create an Entity (`@Flow\Entity`)
2. Create a Embedded Value Object (`@Flow\ValueObject(embedded=true)`)
3. Create a relation ship between Entity and ValueObject
4. Query for a value object property

#### Expected behavior

The query should return a non-empty QueryResult

#### Actual behavior

Currently the doctrine Query implementation assumes that every nested property belongs to a relation and by that (left) joins that entity. 
I think somewhere here https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Persistence/Doctrine/Query.php#L658 should be a check if the property belongs to a value object, and if so the join should be skipped.

vertexvaar:
Instead of assuming that every property path with a dot is a path
to an other entity check if the property path is a mapped field which
is also true for embedded object properties.

Resolves #989

**What I did**
We'll i suppose i fixed it :sweat_smile: 

**How I did it**
I searched the existing class schema for hints about embedded properties and found it in the entityManager. When the path exists in the fieldMappings it is a field embedded in the object's table. Since we're using doctrine in this kind of query anyway i think we're safe to go whit this solution.

**How to verify it**
The description of the original bug should be suffice.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Query.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Query.php
@@ -644,7 +644,9 @@ class Query implements QueryInterface
     {
         $aliases = $this->queryBuilder->getRootAliases();
         $previousJoinAlias = $aliases[0];
-        if (strpos($propertyPath, '.') === false) {
+        if (strpos($propertyPath, '.') === false
+            || $this->entityManager->getClassMetadata($this->entityClassName)->hasField($propertyPath)
+        ) {
             return $previousJoinAlias . '.' . $propertyPath;
         }
 
--- a/Neos.Flow/Tests/Functional/Persistence/Doctrine/QueryTest.php
+++ b/Neos.Flow/Tests/Functional/Persistence/Doctrine/QueryTest.php
@@ -314,7 +314,7 @@ class QueryTest extends FunctionalTestCase
         $this->persistenceManager->persistAll();
 
         $query = new Query(Fixtures\TestEntity::class);
-        $entities = $query->matching($query->equals('embeddedValueObject.value', 'vo'))->setLimit(1)->execute()->toArray();
+        $entities = $query->matching($query->equals('embeddedValueObject.value', 'vo'))->execute()->toArray();
 
         $this->assertEquals(2, count($entities));
     }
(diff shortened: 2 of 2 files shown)


albe:
Woah, that looks too nice to be true :D Thanks for digging into this!

A few things:
- what are the performance implications with the classMetadata lookup for every property?
- doesn't the check also need to happen in the lower path of the code, when walking through a longer path?
- please let's try to provide a test that shows the buggy behaviour and verifies the fix. I can help if needed

vertexvaar:
When i wrote the fix i had the following (and misquoted btw :sweat_smile: ) saying in mind:

> For every complex problem there is an answer that is clear, simple, and wrong

But that one seems to to do the job. Well, we should test that extensively before merging, but i can't invest time in test at this point so i'm counting on you guys :wink: 

> what are the performance implications with the classMetadata lookup for every property

The classMetaData is already fully built (and hold available throughout the request) and the hasField is just an `isset`

> doesn't the check also need to happen in the lower path of the code, when walking through a longer path?

i haven't checked that yet but as it seems any path (e.g. if you embedd an embeddable in an embeddable in your object) is existent in the mappedFields array.

> please let's try to provide a test that shows the buggy behaviour and verifies the fix. I can help if needed

Yes, tests are required and especially tests that check that no other unwanted side effects will pop up. Your help is very appreciated here :relaxed:

albe:
Sorry, I totally lost track of this one. Currently a bit too loaded to get into it unfortunately :( Would like to hear @kdambekalns - should we maybe merge this even without a test in place?

kdambekalns:
TBH I'd rather have a test…

bwaidelich:
I'm with @kdambekalns - this is too critical not to be tested (but unfortunately I currently lack the resources and insight to add those, sorry)

albe:
Assigning to myself for now to hopefully find a bit of time to add a test and get this in sometime soon.

albe:
Finally found the inspiration and will to do it. I'm so sorry it took so absurdly long for this pretty trivial test...

albe:
@bwaidelich @kdambekalns I'd merge this on monday if none of you object

albe:
/rebase

Let's see how it works on latest 4.3...

PR #1405, opened 2018-10-09
TASK: Make HTTP application token static compiled

kitsunet:
This should increase performance in cases version output is configured
(which is the default) as that requires accessing the composer.lock
file and parsing for the respective version.


kitsunet:
Bit of a micro optimisation, but it should make a slight difference.

PR #1406, opened 2018-10-09
TASK: Economize use of settings

kitsunet:
Removes settings injection when the settings are actually not used
and keep only necessary subset of settings and not all of it.


kitsunet:
Bit of a micro optimisation, but technically having compartmentalised settings everywhere will also greatly improve debug output by having less data.

PR #1407, opened 2018-10-10
TASK: Separate security context and authentication manager

kitsunet:
This contains a break up of the cross dependency between
AuthenticationProviderManager and Security context.

First a new TokenAndProviderFactory (with interface) is
introduced to serve both the constructed tokens and providers
from the configuration. Additionally the session persistent
data was moved from the Context to the new SessionDataContainer
(marked internal). This makes the context a simple singleton to
the outside, avoiding duplication (security context injected
before the session was started would create a duplicate instance
without the session data (most notably some SQL security could
have that).

Additionally it fixes ONE_PER_REQUEST CSRF protection tokens
which wouldn't correctly behave.

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/AuthenticationManagerInterface.php
+++ b/Neos.Flow/Classes/Security/Authentication/AuthenticationManagerInterface.php
@@ -12,6 +12,8 @@ namespace Neos\Flow\Security\Authentication;
  */
 
 use Neos\Flow\Security\Context as SecurityContext;
+use Neos\Flow\Security\Exception\AuthenticationRequiredException;
+use Neos\Flow\Security\Exception\NoTokensAuthenticatedException;
 
 /**
  * Contract for an authentication manager.
@@ -26,6 +28,8 @@ interface AuthenticationManagerInterface
      * Note: The order of the tokens in the array is important, as the tokens will be authenticated in the given order.
      *
      * @return array<TokenInterface> An array of tokens this manager is responsible for
+     * @deprecated Use the TokenAndProviderFactory
+     * @see TokenAndProviderFactoryInterface
      */
     public function getTokens();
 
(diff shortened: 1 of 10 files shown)


kitsunet:
@bwaidelich This is the behaviour-changing part of the refactoring broken up :)

Neos.Flow/Classes/Security/Authentication/AuthenticationManagerInterface.php
```diff
@@ -26,13 +28,17 @@
      * Note: The order of the tokens in the array is important, as the tokens will be authenticated in the given order.
      *
      * @return array<TokenInterface> An array of tokens this manager is responsible for
+     * @deprecated Use the TokenAndProviderFactory
```
albe: I'm not so sure about deprecating this API (and the `getProviders()`) - IMO the AuthenticationManager is the main entry point for all Authentication related API calls and it's not quite clear why there is a "TokenAndProviderFactory" (and then why one Factory for both?) that one needs to inject and use. This is a good and useful implementation detail that helps with maintenance, but we should think twice before widening the API surface in such a way.

```suggestion
```

Neos.Flow/Classes/Security/Context.php
```diff
@@ -618,31 +613,64 @@ public function getInterceptedRequest()
      */
     public function clearContext()
     {
+        $sessionDataContainer = $this->objectManager->get(SessionDataContainer::class);
+        $sessionDataContainer->setSecurityTokens([]);
+        $sessionDataContainer->setCsrfProtectionTokens([]);
+        $sessionDataContainer->setInterceptedRequest(null);
```
albe: Maybe the above three lines that leak the knowledge what "clear" means for the `SessionDataContainer` could be abstracted into a `->clear()` or `->reset()` method on the container?

Neos.Flow/Classes/Security/Authentication/AuthenticationManagerInterface.php
kitsunet: is there a reason someone would want that info though? I see it more as an internal thing anyway and actually wanted to reduce the api surface by making the factory non api for now.

kdambekalns: I think that's fine…

PR #1408, opened 2018-10-11
TASK: Enrich log messages with environment

daniellienert:
Enriches log messages of severity > debug with log environment.


daniellienert:
I only extended log messages with severity > debug because the log environment processing is done for every log message independently of if it is actually written or not.
Although the log environment method should not cause a performance problem, IMHO I wouldn't call it for debug messages which are normally not triggered on production servers.

PR #1410, opened 2018-10-12
TASK: correct naming for setting

beardcoder:
Resolves #1410

**Checklist**


albe:
Hey, thanks for this! I think this should be targeting 4.3 instead of master.

beardcoder:
I liked to do that but i think this is a request for all releases.
At the moment you get an error in PHPStorm and other IDEs with validation support for the editorconfig.

kitsunet:
> request for all releases

We follow an upmerge policy. So it must be merged to the lowest (version) branch this problem applies to and will then be upmerged from there.

beardcoder:
Closed because of 

> We follow an upmerge policy. So it must be merged to the lowest (version) branch this problem applies to and will then be upmerged from there.

Correct PR is now https://github.com/neos/flow-development-collection/pull/1413

PR #1416, opened 2018-10-23
FEATURE: Add maxlength to TextArea view helper

robertlemke:
This adds support for the maxlength property to TextArea.

Related to https://github.com/neos/form/issues/87


robertlemke:
@bwaidelich like this? And if I'm not mistaken, this needs to go into master, right?

PR #1422, opened 2018-10-26
!!! BUGFIX: Allow actually reconfiguring the ThrowableStorage

kitsunet:
This fixes the instantiation and configuration of ThrowableStorage
implementations and also fixes stack traces and HTTP request information
in stored throwables.

This change is not breaking but as pre notice that in the next major the
`ThrowableStorageInterface` will have an additional method that is currently
commented and already used to instantiate throwable storages.
If you implement the interface you must already implement that additional 
method despite it being not part of the interface. As the object configuration
for throwable storage doesn't work without this change there cannot be a
working alternative implementation yet, so this shouldn't break any project.


kdambekalns:
The failing build was for a push to origin (instead of the fork) that was deleted while the build was still running… my fault.

PR #1423, opened 2018-10-26
TASK: Exception interfaces with HTTP status and reference code

kitsunet:
Allows implementing `Exceptions` that provide a HTTP status and/or a
reference code for Flow exception handling without extending from the
Flow base Exception implementation.


daniellienert:
Looks good by reading!

PR #1424, opened 2018-10-29
BUGFIX: prevent mutations in PluralsReader

dimaip:
The line `$quantity = fmod($quantity, $subrule['modulo']);` used to mutate the $quantity variable, so if there were two `mod` operations in the same rule the second operation would apply to the result of the previous operation.


kdambekalns:
**How to verify it**

Quantity `11` in Russian local should end up in `MANY`, not `ONE`. See the test.

**Checklist**

dimaip:
Thanks for the reviews, guys!

PR #1433, opened 2018-11-07
FEATURE: Use localDistributionFolders on ``package:create``

mficzel:
If the composer file has a local distribution folder identified by type "path" and
a local folder reference in the "url" new packages are created in the
distributionFolder and are required via 'composer require ... @dev'.

### DIFF
--- a/Neos.Flow/Classes/Package/PackageManager.php
+++ b/Neos.Flow/Classes/Package/PackageManager.php
@@ -340,6 +340,18 @@ class PackageManager implements PackageManagerInterface
             $manifest['type'] = PackageInterface::DEFAULT_COMPOSER_TYPE;
         }
 
+        $runComposerRequireForTheCreatedPackage = false;
+        if ($packagesPath === null) {
+            $composerManifestRepositories = ComposerUtility::getComposerManifest(FLOW_PATH_ROOT, 'repositories');
+            foreach ($composerManifestRepositories as $repository) {
+                if ($repository['type'] == 'path' && substr($repository['url'], 0, 2) == './' && substr($repository['url'], -2) == '/*') {
+                    $packagesPath = Files::getUnixStylePath(Files::concatenatePaths([FLOW_PATH_ROOT, substr($repository['url'], 0, -2)]));
+                    $runComposerRequireForTheCreatedPackage = true;
+                    break;
+                }
+            }
+        }
+
         if ($packagesPath === null) {
             $packagesPath = 'Application';
             if (is_array($this->settings['packagesPathByType']) && isset($this->settings['packagesPathByType'][$manifest['type']])) {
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Command/PackageCommandController.php
```diff
@@ -78,10 +78,11 @@ public function injectBootstrap(Bootstrap $bootstrap)
      * @Flow\FlushesCaches
      * @param string $packageKey The package key of the package to create
      * @param string $packageType The package type of the package to create
+     * @param string $packagesPath The path where new packages are created relative to the flow root path
      * @return void
      * @see neos.kickstarter:kickstart:package
      */
-    public function createCommand(string $packageKey, string $packageType = PackageInterface::DEFAULT_COMPOSER_TYPE)
+    public function createCommand(string $packageKey, string $packageType = PackageInterface::DEFAULT_COMPOSER_TYPE, ?string $packagesPath)
```
daniellienert: What do you think about setting 'DistributionPackages' as default here and use it if it exists? Would be a better default.

Neos.Kickstarter/Classes/Command/KickstartCommandController.php
```diff
@@ -46,10 +46,11 @@ class KickstartCommandController extends CommandController
      *
      * @param string $packageKey The package key, for example "MyCompany.MyPackageName"
      * @param string $packageType Optional package type, e.g. "neos-plugin"
+     * @param string $packagesPath The path where new packages are created relative to the flow root path.
      * @return string
      * @see neos.flow:package:create
      */
-    public function packageCommand($packageKey, $packageType = PackageInterface::DEFAULT_COMPOSER_TYPE)
+    public function packageCommand(string $packageKey, string $packageType = PackageInterface::DEFAULT_COMPOSER_TYPE, ?string $packagesPath)
```
daniellienert:
```suggestion
    public function packageCommand(string $packageKey, string $packageType = PackageInterface::DEFAULT_COMPOSER_TYPE, ?string $packagesPath): void
```

Neos.Kickstarter/Classes/Command/KickstartCommandController.php
```diff
@@ -46,10 +46,11 @@ class KickstartCommandController extends CommandController
      *
      * @param string $packageKey The package key, for example "MyCompany.MyPackageName"
      * @param string $packageType Optional package type, e.g. "neos-plugin"
+     * @param string $packagesPath The path where new packages are created relative to the flow root path.
      * @return string
```
daniellienert:
```suggestion
     * @return void
```

Neos.Flow/Classes/Command/PackageCommandController.php
mficzel: I removed the change in this file and detect the DistributionPackages folder.

kitsunet:
What could ever go wrong.. My bad should've thought this through, but you cannot run unit tests with an installation with distribution packages anymore as it will start composer when the unit test creates a package... I guess we need to fix this ASAP.

mficzel:
@kitsunet is there an option to run the unit tests with a specified packagePath. Since the composer behavior is not wrong per se that should help aswell.

PR #1436, opened 2018-11-09
BUGFIX: Fix console helpers

gerhard-boden:
Fix description for ``select`` method and return type an therefore tests for ``askAndValidate``.


albe:
@bwaidelich because the buggy return type was introduced in 5.0: https://github.com/neos/flow-development-collection/commit/e4beef6a8f15e1f55dfe9f418dcf5a0e472a81e7#diff-3e19a78838a95d41e9e6247ed3c1e49c

bwaidelich:
@albe ah, thanks for clarifying. makes sense of course :)

gerhard-boden:
Thanks guys!

PR #1437, opened 2018-11-12
FEATURE: Utility Function to detect Eel expressions

rkrahnen:
The function was added to provide a function for validity checks without any need to duplicate the code which was previously used in the first if-condition of the function evaluateEelExpression.

As @kitsunet mentioned in https://github.com/neos/eel/pull/2 the variable `matches` is not necessary, but I've used it to avoid calling `preg_match(Package::EelExpressionRecognizer, $expression, $matches)` inside the existing function `evaluateEelExpression` and `preg_match(Package::EelExpressionRecognizer, $expression)` inside the new function `isEelExpression`, because to me this seems like duplicated code.
I'm happy with both solutions, so feel free to change it if you decide to remove the variable.

### DIFF
--- a/Neos.Eel/Classes/Utility.php
+++ b/Neos.Eel/Classes/Utility.php
@@ -21,8 +21,7 @@ class Utility
      * Return the expression if it is an valid EEL expression, otherwise return null.
      *
      * @param string $expression
-     *
-     * @return bool
+     * @return string|null
      */
     public static function parseEelExpression($expression)
     {
--- a/Neos.Eel/Classes/Utility.php
+++ b/Neos.Eel/Classes/Utility.php
@@ -21,13 +21,12 @@ class Utility
      * Check if expression is an Eel expression.
      *
      * @param string $expression
-     * @param array $matches
      *
      * @return bool
      */
-    public static function isEelExpression($expression, &$matches = null)
+    public static function isEelExpression($expression)
     {
-        return preg_match(Package::EelExpressionRecognizer, $expression, $matches);
+        return preg_match(Package::EelExpressionRecognizer, $expression) === 1;
     }
 
     /**
@@ -67,10 +66,11 @@ class Utility
      */
     public static function evaluateEelExpression($expression, EelEvaluatorInterface $eelEvaluator, array $contextVariables, array $defaultContextConfiguration = [])
     {
-        $matches = null;
-        if (!self::isEelExpression($expression, $matches)) {
+        if (!self::isEelExpression($expression)) {
             throw new Exception('The EEL expression "' . $expression . '" was not a valid EEL expression. Perhaps you forgot to wrap it in ${...}?', 1410441849);
         }
+
+        preg_match(Package::EelExpressionRecognizer, $expression, $matches);
 
         $defaultContextVariables = self::getDefaultContextVariables($defaultContextConfiguration);
         $contextVariables = array_merge($defaultContextVariables, $contextVariables);
--- a/Neos.Eel/Classes/Utility.php
+++ b/Neos.Eel/Classes/Utility.php
@@ -24,9 +24,9 @@ class Utility
      *
      * @return bool
      */
-    public static function isEelExpression($expression)
+    public static function parseEelExpression($expression)
     {
-        return preg_match(Package::EelExpressionRecognizer, $expression) === 1;
+        return preg_match(Package::EelExpressionRecognizer, $expression, $matches) === 1 ? $matches : null;
     }
 
     /**
@@ -66,7 +66,8 @@ class Utility
      */
     public static function evaluateEelExpression($expression, EelEvaluatorInterface $eelEvaluator, array $contextVariables, array $defaultContextConfiguration = [])
     {
-        if (!self::isEelExpression($expression)) {
+        $eelExpression = self::parseEelExpression($expression);
+        if ($eelExpression === null) {
             throw new Exception('The EEL expression "' . $expression . '" was not a valid EEL expression. Perhaps you forgot to wrap it in ${...}?', 1410441849);
         }
 
--- a/Neos.Eel/Classes/Utility.php
+++ b/Neos.Eel/Classes/Utility.php
@@ -18,7 +18,7 @@ namespace Neos\Eel;
 class Utility
 {
     /**
-     * Check if expression is an Eel expression.
+     * Return the expression if it is an valid EEL expression, otherwise return null.
      *
      * @param string $expression
      *
@@ -26,7 +26,7 @@ class Utility
      */
     public static function parseEelExpression($expression)
     {
-        return preg_match(Package::EelExpressionRecognizer, $expression, $matches) === 1 ? $matches : null;
+        return preg_match(Package::EelExpressionRecognizer, $expression, $matches) === 1 ? $matches['exp'] : null;
     }
 
     /**
@@ -71,8 +71,6 @@ class Utility
             throw new Exception('The EEL expression "' . $expression . '" was not a valid EEL expression. Perhaps you forgot to wrap it in ${...}?', 1410441849);
         }
 
-        preg_match(Package::EelExpressionRecognizer, $expression, $matches);
-
         $defaultContextVariables = self::getDefaultContextVariables($defaultContextConfiguration);
         $contextVariables = array_merge($defaultContextVariables, $contextVariables);
 
@@ -87,6 +85,6 @@ class Utility
         $context = new ProtectedContext($contextVariables);
         $context->whitelist('q');
 
-        return $eelEvaluator->evaluate($matches['exp'], $context);
+        return $eelEvaluator->evaluate($eelExpression, $context);
     }
 }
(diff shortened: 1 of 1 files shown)


kitsunet:
Yep, I see it. It's rather weird, but it was weird before, so this doesn't really make it worse, I guess 😄

Neos.Eel/Classes/Utility.php
```diff
@@ -17,6 +17,19 @@
...
+     * Check if expression is an Eel expression.
+     *
+     * @param string $expression
+     * @param array $matches
+     *
+     * @return bool
+     */
+    public static function isEelExpression($expression, &$matches = null)
+    {
+        return preg_match(Package::EelExpressionRecognizer, $expression, $matches);
```
bwaidelich: Please lets make this sth like
ˋˋˋ
return preg_match(Package::EelExpressionRecognizer, $expression) === 1);
ˋˋˋ
and get rid of the $matches param

rkrahnen:
@bwaidelich : I've made the requested changes a while ago, but you did not approve until now - have I done something wrong, so that you didn't see my change?

albe:
I guess Bastian is just loaded with work and then the NeosCon last week. TBH I think the now doubly invoked preg_match will have a bad performance impact (I'm not concerned about the code duplication at all) and I think Bastian (and previously Christian) didn't have that in mind when suggesting to get rid of the `$matches` parameter.

So here's my suggestion on how to go with this (please bear with me/us about nagging around 🙏):
- make the `isEelExpression()` actually return the matched 'exp' capture on match, or `null` if no match
- possibly rename the method then `parseEelExpression(string $expression): ?string` to better match the new behaviour
- use the return value in the check for valid Eel expression and the `$eelEvaluator->evaluate()` call

```php
$eelExpression = self::parseEelExpression($expression);
if ($eelExpression === null) {
    throw ...
}
...
return $eelEvaluator->evaluate($eelExpression, $context);
```

wdyt @kitsunet @bwaidelich ?

rkrahnen:
Nice suggestion 👍

albe:
Still missing the adjustment in the `$eelEvaluator->evaluate($eelExpression, $context);` call, that still references `$matches['exp']`. Thanks for keeping up :)

Neos.Eel/Classes/Utility.php
```diff
@@ -17,6 +17,18 @@
...
+    /**
+     * Check if expression is an Eel expression.
+     *
+     * @param string $expression
+     *
+     * @return bool
+     */
+    public static function parseEelExpression($expression)
+    {
+        return preg_match(Package::EelExpressionRecognizer, $expression, $matches) === 1 ? $matches : null;
```
albe: Make that `? $matches['exp'] : null`

rkrahnen:
You're right, previous commit was stupid, fixed it.

Neos.Eel/Classes/Utility.php
```diff
@@ -17,6 +17,18 @@
  */
 class Utility
 {
+    /**
+     * Return the expression if it is an valid EEL expression, otherwise return null.
+     *
+     * @param string $expression
+     *
```
bwaidelich: empty line can be removed

Neos.Eel/Classes/Utility.php
```diff
@@ -17,6 +17,18 @@
...
+     * @return bool
```
bwaidelich: incorrect return type, should be `?string`

Neos.Eel/Classes/Utility.php
```diff
@@ -17,6 +17,18 @@
...
+     */
+    public static function parseEelExpression($expression)
```
bwaidelich: This could have a return type annotation of `?string`, too

bwaidelich:
Forget the last part of my comment - we do have to extract the "inner expression" before sending the expression to the evaluator.

I thought about this once again and IMO this is the problem we are trying to solve here (I think):
We want to be able to use Eel as DSL within applications.
To prevent that everything has to be an Eel expression, there should be some fallback mechanism.

Something like:
```php
if (isEelExpression($string)) {
  return parseEelExpression($string, $variables);
}
return $string;
```

To prevent the regex from being parsed twice we could change it to
```php
try {
    $parsed = parseEelExpression($string, $variables);
} catch(NotAnEelExpression $exception) {
    return $string;
}
```

@kitsunet @albe @rkrahnen what do you think?

Neos.Eel/Classes/Utility.php
rkrahnen: I changed it to `string|null` to keep compatibility with PHP 7.0

Neos.Eel/Classes/Utility.php
rkrahnen: Did not add return type to keep compatibility with PHP 7.0

Did not add parameter type hint because `evaluateEelExpression` has also none, I would prefer to change both if that is fine for you.

albe:
I do think both ways are valid tbh and I don't consider returning `null` to denote a function couldn't return what it was expected generally wrong (or magic). That's why there are nullable types after all. I do think a try/catch is also a little bit slower, but I'd be fine with that solution too. Would be cool to have some numbers though.

bwaidelich:
> I don't consider returning null to denote a function couldn't return what it was expected generally wrong (or magic)

It's not wrong per se, but it is magic in that the code doesn't force you to cater for that special case and that has caused a lot of pain in the past due to null pointer errors.
In this case I'm not concerned about the null pointer - an Eel expression can always yield `null`. But it's exactly _that_ special case that can't be detected anymore with null == error.

I use Eel expressions a lot for (optionally) dynamic configuration.
For example a custom Email Form Finisher:
```yaml
emailFinisher:
  options:
    subject: 'Some static subject'
    replyTo: '${formValues.email}'
```
In this case I don't know in advance whether the given string is an expression or a static value. If I would assume a static value if `evaluateEelExpression()` returns `null` I'd run into trouble here.

> I do think a try/catch is also a little bit slower

What makes you think that?

albe:
> assume a static value if evaluateEelExpression() returns null I'd run into trouble here.

If the method is `evaluateEelExpression` you are totally right, but for the `parseEelExpression` to return either the (well formed) Eel expression `string` or `null` that's a non-issue ;)

> What makes you think that?

try/catch on a lower level leads to branch mispredictions in the catch case and hence slower execution performance. Also, you need to create the Error object which you throw, so you have a memory allocation and potential GC. See also https://stackoverflow.com/a/17684984
One of the reasons why throwing exceptions should really be exceptional. Which in turn is another argument to avoid the try/catch here, because how exceptional is it really to get a non-eel expression passed?

Neos.Eel/Classes/Utility.php
```diff
@@ -54,8 +65,8 @@ public static function getDefaultContextVariables(array $configuration)
      */
     public static function evaluateEelExpression($expression, EelEvaluatorInterface $eelEvaluator, array $contextVariables, array $defaultContextConfiguration = [])
     {
-        $matches = null;
-        if (!preg_match(Package::EelExpressionRecognizer, $expression, $matches)) {
+        $eelExpression = self::parseEelExpression($expression);
+        if ($eelExpression === null) {
```
albe: @bwaidelich what if we just move the `throw` here into the `parseEelExpression` method? Then we get rid of the nullable return type.

Neos.Eel/Classes/Utility.php
bwaidelich: As this targets master we can safely assume PHP 7.1+ (see https://www.neos.io/download-and-extend.html).

But that was a nitpick anyways.. The more I think about this the more I doubt that we can fix that class.. see discussion below

Neos.Eel/Classes/Utility.php
bwaidelich: Yes, that's what I suggested. But we'd need a new exception (e.g. `NoEelExpressionException`) to differentiate it from exceptions that occur during evaluation of the Eel expression

albe: I wouldn't want to change the exception thrown here, since this is more a refactoring PR that simply extracts an internal functionality into a named method. But yes to new exception class for a broader rewrite of this Utility into a real Service.

albe:
> Maybe it's easier to start with a new service class that can be injected and has a really easy-to-use interface like

Sure! But I think that is something much more complex and beyond the scope of this PR. So I'd rather merge this as is, since it improves the current situation a bit, and progress further in a new issue/PR. Are you okay with this?

kdambekalns:
> I took a greater rework to the core team for further discussions

I moved that to https://discuss.neos.io/t/new-utility-class-for-eel-evaluation/4369, it's not as volatile.

rkrahnen:
@bwaidelich No problem, I followed your discussion with excitement! Thanks for all the feedback.

bwaidelich:
> I moved that to https://discuss.neos.io/t/new-utility-class-for-eel-evaluation/4369, it's not as volatile.

@kdambekalns good call, thanks!

albe:
So, merging this as is as agreed. @rkrahnen Thanks a lot again for keeping up and congrats on the successful contribution! If you feel like it, you're invited to take part in the discussion in the thread linked above and possibly implementation of the restructuring of this Eel Utility class.

PR #1438, opened 2018-11-16
BUGFIX: Fix format for json_encoded DateTime conversion

albe:
The format for converting from json_encode'd DateTime was wrong. The format doesn't use the `\T` separator, but a whitespace and as it seems, `v` is not a valid format specifier for `DateTime::createFromFormat` while it is for `date()`.

http://php.net/manual/en/datetime.createfromformat.php
"In **most cases**, the same letters as for the [date()](http://php.net/manual/en/function.date.php) can be used." [emphasis by me]

Also, this provides a test to actually verify the functionality.
Related to #1415


albe:
A good example why covering new behaviour with tests is important, even with code review in place. Blaming myself for not getting the test, which was really easy to do, done earlier.

PR #1439, opened 2018-11-16
FEATURE: Add ``set`` method to ``Array`` Eel-helper

mficzel:
This allows to set keys which are calculated at runtime in Eel. A possible use case for this
is the ``Neos.Fusion:Reduce`` prototype that can be used for grouping with ``set``.

### DIFF
--- a/Neos.Eel/Classes/Helper/ArrayHelper.php
+++ b/Neos.Eel/Classes/Helper/ArrayHelper.php
@@ -367,6 +367,20 @@ class ArrayHelper implements ProtectedContextAwareInterface
     }
 
     /**
+     * Set the specified key in the the array
+     *
+     * @param array $array
+     * @param string|integer $key the Key that will be set
+     * @param mixed $value the value to assign to the key
+     * @return array The modified array.
+     */
+    public function set(array $array, $key, $value)
+    {
+        $array[$key] = $value;
+        return $array;
+    }
+
+    /**
      * All methods are considered safe
      *
      * @param string $methodName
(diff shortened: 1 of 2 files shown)


Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -366,6 +366,20 @@ public function range($start, $end, $step = 1)
         return range($start, $end, $step);
     }
 
+    /**
+     * Set the specified key in the the array
+     *
+     * @param array $array
+     * @param mixed $key the Key that will be set
```
bwaidelich: doesn't the key have to be a `string`?

mficzel: At least integer would be ok aswell

mficzel: @bwaidelich i altered the annotation to `string|integer`

mficzel:
The main purpose of this is to enable grouping via `Neos.Fusion:Reduce` like so:
```
grouped = Neos.Fusion:Reduce {
    items = ${ __array_of_products__ }
    initialValue = ${[]}
    itemReducer = News.Fusion:Value {
        groupName = ${item.group}
        groupCarry = ${carry[ this.groupName ] || []}
        groupValue = ${Array.push(this.groupCarry, item)}
        # the line below is what is actually not possible now
        value = ${ Array.set(carry, this.groupName, this.groupValue) }
    }
}
```

albe:
Looks good by reading. But this will ofc raise demand for an `unset` helper method ;P

PR #1440, opened 2018-11-16
FEATURE: Add chr, ord, sha1 and nl2br to String Eel-Helper

bwaidelich:
Adds the functions `chr()`, `ord()`, `sha1()` and `nl2br()` to
the `StringHelper`, all simple wrappers for the corresponding
PHP functions.


bwaidelich:
Wow, Github considers the test file binary.. I think I should adjust the test cases a little :D

PR #1442, opened 2018-11-16
BUGFIX: Properly parse `DateTimeImmutable` types

albe:
Currently, a type string of `DateTimeImmutable` will be parsed as `DateTime`. This for example leads to any entity properties annotated as `@var DateTimeImmutable` to be hydrated into a `DateTime` class instead.

Note that this is a hotfix at most, because the real issue is, that the regex does not check for a type ending character, like whitespace, line end or other non-word character. Therefore, it eagerly parses `DateTimeImmutable` by matching the `DateTime` prefix and taking that as the parsed type, which is wrong.

I'd rather change the regex to end with `(?:\b|$)`, but I'm unsure how breaking that would be.

### DIFF
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -24,7 +24,7 @@ abstract class TypeHandling
     /**
      * A property type parse pattern.
      */
-    const PARSE_TYPE_PATTERN = '/^\\\\?(?P<type>integer|int|float|double|boolean|bool|string|DateTime|[a-zA-Z0-9\\\\_]+|object|array|ArrayObject|SplObjectStorage|Doctrine\\\\Common\\\\Collections\\\\Collection|Doctrine\\\\Common\\\\Collections\\\\ArrayCollection)(?:<\\\\?(?P<elementType>[a-zA-Z0-9\\\\_]+)>)?/';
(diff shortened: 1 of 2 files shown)


albe:
Uncovered via #1401

bwaidelich:
Could you describe the bug, I don't get it

albe:
Added a test for the wrong behaviour. Still this should be handled more generally rather than just for the single `DateTimeImmutable` case.

ComiR:
Why even define it with all those extra options? `^\\\\?(?P<type>[a-zA-Z0-9\\\\_]+)(?:<\\\\?(?P<elementType>[a-zA-Z0-9\\\\_]+)>)?` should suffice (since `[a-zA-Z0-9\\\\_]+` matches the other options as well...).

albe:
That's my thought too. But let's get this fixed quickly for now like Sebastian suggests and then discuss how to solve this more generally in an own issue/PR.

bwaidelich:
That change seems to break doctrine entities with `DateTimeImmutable` properties:
```
Failure while evaluating property annotations for class "Neos\Media\Domain\Model\ImportedAsset": The attribute 'type' is required for the column description of property Neos\Media\Domain\Model\ImportedAsset::$importedAt
```
(Neos master, doctrine ORM 2.6.3)

albe:
This basically just oncovered the buggy type parser. This in itself led to DateTimeImmutable types no longer being automatically mapped to `DateTime` types in Doctrine. With #1401 being applied this is fixed. Did upmerges so master is not breaking any longer.

ComiR:
Fortunately it's not released yet. Since `doctrine/orm` only introduced `date_immutable` in version 2.6, so we cannot backport #1401 (due to Flow 4.3 only requiring `doctrine/orm ~2.5.0`).
So either we fix much more than "only" this, or implement this fix only in 5.0+...

kdambekalns:
Since #1401 is in 5.0+ only, this one should be reverted for 4.3 (but kept in 5.0+).

kdambekalns:
I created a "revert PR", please have a look. I think I untangled this mess correctly… :)

PR #1444, opened 2018-11-16
TASK: Update Doctrine DBAL dependency to 2.8

albe:
Since DBAL 2.8 has no b/c breaks, it should be safe to update the version.

See https://github.com/doctrine/dbal/releases/tag/v2.8.0

In the longer term, we should also phase out the doctrine/common dependency.
See https://www.doctrine-project.org/2018/07/12/common-2-9-and-dbal-2-8-and-orm-2-6-2.html
AFAIS we only depend on `Doctrine\Common\Util\Debug` which we are suggested to replace with `symfony/var-dumper`.

### DIFF
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -36,7 +36,7 @@
         "doctrine/orm": "~2.6.0",
         "doctrine/migrations": "~1.6.0",
         "doctrine/dbal": "~2.8.0",
-        "doctrine/common": ">=2.4,<2.8-dev",
+        "doctrine/common": "~2.4",
 
         "symfony/yaml": "~4.0.0",
         "symfony/dom-crawler": "~4.0.0",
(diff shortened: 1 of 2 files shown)


albe:
@kdambekalns should we maybe do `~2.7.0|~2.8.0` (or similar) instead?

kitsunet:
Seems some more dependencies need to be updated.

albe:
doctrine/common was too strict. Maybe we could even increase the minimum doctrine/common dependency, since DBAL 2.7 requires `"doctrine/common": "^2.7.1"`. Tests pass now.

jonnitto:
Original PR text:

> Since DBAL 2.8 has no b/c breaks, it should be safe to update the version.
>
> See https://github.com/doctrine/dbal/releases/tag/v2.8.0
> 
> In the longer term, we should also phase out the doctrine/common dependency.
> See https://www.doctrine-project.org/2018/07/12/common-2-9-and-dbal-2-8-and-orm-2-6-2.html
> AFAIS we only depend on `Doctrine\Common\Util\Debug` which we are suggested to replace with > > `symfony/var-dumper`.

PR #1450, opened 2018-11-19
BUGFIX: Support the parser halt fluid token

### LINKED ISSUE #1449: {parsing off} is not supported
### Using `{parsing off}` in fluid will not stop parsing

A new feature was introduced to fluid (commited on 12th Nov. 2016) which allows templates to be printed as-is by remaining unparsed. This is achieved by putting the string `{parsing off}` in the template.
The Package Neos.FluidAdaptor breaks this feature by not registering the template pre-processor required for this feature.

### Steps to Reproduce

1. Create a fluid template beginning with `{parsing off}`
2. Render the fluid template in a Flow or Neos app
3. Search your rendered HTML for `{parsing off}`. You will find it.

#### Expected behavior

Te template must not be parsed. Therefore all used viewhelpers in the template must not be resolved but outputted as-is (as string). Also `{parsing off}` must not occur in the output.

#### Actual behavior

Fluid renders the Template like any other

### Affected Versions

Since: Always
Neos: (Not part of Neos)
Flow: Always (Currently working with 5.1.4)

Pull request will be available in a minute :wink:

vertexvaar:
This adds the missing template pre-processor to the Neos.FluidAdaptor RenderingContext.
(Actually it is not missing but gets removed upon setting the template pre-processors.)

Fixes #1449

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/Rendering/RenderingContext.php
+++ b/Neos.FluidAdaptor/Classes/Core/Rendering/RenderingContext.php
@@ -28,6 +28,7 @@ use TYPO3Fluid\Fluid\Core\Parser\Configuration;
 use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\Expression\CastingExpressionNode;
 use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\Expression\MathExpressionNode;
 use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\Expression\TernaryExpressionNode;
+use TYPO3Fluid\Fluid\Core\Parser\TemplateProcessor\PassthroughSourceModifierTemplateProcessor;
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContext as FluidRenderingContext;
 use TYPO3Fluid\Fluid\View\ViewInterface;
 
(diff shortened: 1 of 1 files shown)


kdambekalns:
Thanks!

I changed the target branch, since 4.1 to 4.2 are in security-fix-only mode by now.

Here is the remaining oridignal PR comment:

---

**How to verify it**

Trying to reproduce will produce the expected behavior (see the issue)

**Checklist**

bwaidelich:
I have to admit that I have no clue how those `templateProcessors` work exactly.
But looks reasonable, especially if it fixes the issue.

From the Neos standpoint this is a new feature though.. I can't judge the "breakability" of this one

vertexvaar:
> I changed the target branch, since 4.1 to 4.2 are in security-fix-only mode by now.
@kdambekalns i always get these wrong, sorry :sweat_smile: 

> I have to admit that I have no clue how those templateProcessors work exactly.
@bwaidelich They should rather be called TemplatePreProcessors, because that's what they actually do. They do stuff like CDATA section protection (in Neos.FluidAdaptor. Fluid standalone would drop them), check registered namespaces as xmlns declaration and/or old style namespace declaration in (curly) braces and remove them from the rendering as well as the html tag if it has the `data-namespace-typo3-fluid` attribute. In the case of the `PassthroughSourceModifierTemplateProcessor` it will throw an exception to prevent fluid parsing at all. All in all no witchcraft at all :wink: 

> Thanks also for looking into the version matching!
@albe You're welcome.

All together thanks for the approvals (and upcoming merge) :+1:

kitsunet:
We should probably make the list of processors a configuration at some point.

bwaidelich:
> We should probably make the list of processors a configuration at some point.

An important feature of Fluid was always that a template was self-contained (i.e. you don't need any context to know how the template works). Pre processors are a nice feature but IMO they should not be configurable. At least not too easily :) otherwise you could break a totally unrelated part by changing the configuration.

PR #1451, opened 2018-11-19
TASK: Translator uses locale chain

ComiR:
This change makes getTranslationById and getTranslationByOriginalLabel use the configured
locale chain.

This is an updated version of #327 and #328. Please see the discussions there. May be retargeted on master.

### DIFF
--- a/Neos.Flow/Classes/I18n/Translator.php
+++ b/Neos.Flow/Classes/I18n/Translator.php
@@ -148,7 +148,10 @@ class Translator
             }
         }
 
-        return $originalLabel;
+        return $arguments === []
+            ? $originalLabel
+            : $this->formatResolver->resolvePlaceholders($originalLabel, $arguments, $locale)
+        ;
     }
 
     /**
--- a/Neos.Flow/Classes/I18n/Translator.php
+++ b/Neos.Flow/Classes/I18n/Translator.php
@@ -126,19 +126,29 @@ class Translator
         if ($locale === null) {
             $locale = $this->localizationService->getConfiguration()->getCurrentLocale();
         }
-        $pluralForm = $this->getPluralForm($quantity, $locale);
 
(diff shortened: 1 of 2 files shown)


ComiR:
Excuse my use of the force (especially because it gets documented now in the PR 👍), but I had to fix the tests with the help of Travis ;)

Neos.Flow/Classes/I18n/Translator.php
```diff
@@ -126,19 +126,29 @@ public function translateByOriginalLabel($originalLabel, array $arguments = [],
...
+                        $translatedMessage,
+                        $arguments,
+                        $localeInChain
+                    )
+                ;
+            }
         }
 
-        return $translatedMessage;
+        return $originalLabel;
```
ComiR: I am currently not sure if possible placeholders in the `$originalLabel` should be resolved as well. Before they were, so maybe now at least resolve them using the given (or current) locale?

This line could then be changed to
```php
    return $arguments === []
        ? $originalLabel
        : $this->formatResolver->resolvePlaceholders($originalLabel, $arguments, $locale)
    ;
```

kdambekalns:
Huh, I created a different solution "by accident", see https://github.com/neos/flow-development-collection/pull/1555

ComiR:
Well, not quite. In the case of the XLIFF source language being the same as the fallback language in the locale chain: yes. But for all other cases: no.
Simple example: Locale chain "de_CH, de_DE, en". Your solution would jump from "de_CH" straight to "en" (supposing it also is the source language).

DavidSporer:
Hi,
I've tested it today and can confirm that it works in the way @ComiR has described it. The fallback uses the locale chain as expected.
Let me know if you need additional information.

Thanks,
David

albe:
Thanks @DavidSporer for the verification and thanks @ComiR for the long lasting effort! I do think this is good to go when the originalLabel interpolation is in place again (which I'll add just now). But technically it's a new feature (changes behaviour) as is, so it should probably be retargeted to master. I'd love to have this in lower versions too, it just makes so much sense after all, but can't play with semver just for this.

PR #1452, opened 2018-11-20
FEATURE: Make cache application identifier configurable

kitsunet:
In multiple permutations we tried to fix problems with cache identifier
uniqueness in cache backends that are shared like apcu or memcache.
In earlier days it included the PHP_SAPI and then in more recent times
the context and root path. With the refactoring of caches these two
became the hardcoded `applicationIdentifier` which can be used by
any backend to add more specificity to cache identifiers.
It turns out that the root path doesn't work well for some environments
and can result in bugs when used with eg. the PdoBackend and a
deployment that changes the root path (typical Surf or Deployer).
The only backwards compatible way to fix this was to make the
applicationIdentifier configurable with a default that matches the
previously hardcoded values. That way nothing changes in existing
installations but if the bug appears it can be easily fixed.

### DIFF
--- a/Neos.Flow/Classes/Cache/CacheFactory.php
+++ b/Neos.Flow/Classes/Cache/CacheFactory.php
@@ -81,15 +81,16 @@ class CacheFactory extends \Neos\Cache\CacheFactory implements CacheFactoryInter
      *
      * @param ApplicationContext $context The current Flow context
      * @param Environment $environment
+     * @param string $applicationIdentifier
      * @Flow\Autowiring(enabled=false)
      */
-    public function __construct(ApplicationContext $context, Environment $environment)
+    public function __construct(ApplicationContext $context, Environment $environment, string $applicationIdentifier)
     {
         $this->context = $context;
         $this->environment = $environment;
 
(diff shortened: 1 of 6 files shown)


aertmann:
This is targeted for master, that's on purpose? Since it will be backwards compatible..

daniellienert:
@kitsunet, If I got it right, if deployed with surf, even the caches marked as persistent get lost on every deployment? No matter which cache backend is used. If that’s right we really should mention to set that identifier manually in some docs as I think surf / deployer setups are a kind of standard out there.

daniellienert:
Additionally I also think that this can / should go as Bugfix into a lower branch.

kitsunet:
Forgot to adapt tests to the third constructor parameter in the factory... Will take care. I wouldn't mind backporting this, just wanted to go the "secure" route. WDYT @kdambekalns ? As the others seemed to like the idea

kitsunet:
Seems there is an agreement on making this a bugfix. I will take care and add you to the new change as soon as it's ready.

kitsunet:
Here we go: https://github.com/neos/flow-development-collection/pull/1457

PR #1454, opened 2018-11-21
BUGFIX: Throw MappingException in loadMetaDataForClass

### LINKED ISSUE #1453: Fix in ORM 2.6.3 breaks query execution
### Description

The fixed parameter value processing for objects with unloaded metadata in ORM 2.6.3 (https://github.com/doctrine/doctrine2/pull/7471/) breaks query execution (in particular cases.)

#### Actual behavior

When no class schema can be found in `loadMetaDataForClass`, a `ClassSchemaNotFoundException` is thrown. This leads to errors like this as of ORM 2.6.3:

```
FlowAnnotationDriver.php: No class schema found for "some-non-mapped-class".

89 …\FlowAnnotationDriver_Original::getClassSchema("some-non-mapped-class")
88 …\FlowAnnotationDriver_Original::loadMetadataForClass("some-non-mapped-class", Neos\Flow\Persistence\Doctrine\Mapping\ClassMetadata)
```

### Affected Versions

Flow: 5.0.0 and up with doctrine/orm 2.6.3 and up.

kdambekalns:
When no class schema can be found in `loadMetaDataForClass`, a Doctrine
`MappingException` is now thrown. This makes our code work nicely with the
change in https://github.com/doctrine/doctrine2/pull/7471/ that otherwise
leads to errors like this as of ORM 2.6.3:

```
FlowAnnotationDriver.php: No class schema found for "some-non-mapped-class".

89 …\FlowAnnotationDriver_Original::getClassSchema("some-non-mapped-class")
88 …\FlowAnnotationDriver_Original::loadMetadataForClass("some-non-mapped-class", Neos\Flow\Persistence\Doctrine\Mapping\ClassMetadata)
```

Fixes #1453

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
@@ -178,9 +178,13 @@ class FlowAnnotationDriver implements DoctrineMappingDriverInterface, PointcutFi
          * @var ORM\ClassMetadata $metadata
          */
 
-        $class = $metadata->getReflectionClass();
-        $classSchema = $this->getClassSchema($class->getName());
-        $classAnnotations = $this->reader->getClassAnnotations($class);
+        try {
+            $class = $metadata->getReflectionClass();
+            $classSchema = $this->getClassSchema($class->getName());
+            $classAnnotations = $this->reader->getClassAnnotations($class);
+        } catch (ClassSchemaNotFoundException $exception) {
+            throw new ORM\MappingException(sprintf('Failure while fetching class schema class "%s": %s', $metadata->getName(), $exception->getMessage()), 1542792708, $exception);
+        }
 
         // Evaluate Entity annotation
         if (isset($classAnnotations[ORM\MappedSuperclass::class])) {


Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
```diff
@@ -178,9 +178,13 @@ public function loadMetadataForClass($className, ClassMetadata $metadata)
...
          */
 
-        $class = $metadata->getReflectionClass();
-        $classSchema = $this->getClassSchema($class->getName());
-        $classAnnotations = $this->reader->getClassAnnotations($class);
+        try {
+            $class = $metadata->getReflectionClass();
+            $classSchema = $this->getClassSchema($class->getName());
+            $classAnnotations = $this->reader->getClassAnnotations($class);
+        } catch (ClassSchemaNotFoundException $exception) {
```
bwaidelich: Just wondering: Can all of the three lines lead to the `ClassSchemaNotFoundException`? Otherwise I would suggest to wrap the `try/catch` only around the "expected" exceptional cases

kitsunet: Should be only the middle one.

kitsunet:
I assume 4.3 doesn't use ORM 2.6.x ?

albe:
> I assume 4.3 doesn't use ORM 2.6.x ?

Correct. 2.5.x due to PHP 7.1 requirement

kitsunet:
Merged due to the urgency of the problem. We can refator the try catch block any time.

Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
kdambekalns: It is the middle one. I still like the grouping, those lines are really closely related. "Splitting" them by a try/catch seemed confusing.

kdambekalns:
Ok, I'll roll patch-level releases then… Thanks!

PR #1457, opened 2018-11-21
BUGFIX: Make cache application identifier configurable

kitsunet:
In multiple permutations we tried to fix problems with cache identifier
uniqueness in cache backends that are shared like apcu or memcache.
In earlier days it included the PHP_SAPI and then in more recent times
the context and root path. With the refactoring of caches these two
became the hardcoded `applicationIdentifier` which can be used by
any backend to add more specifity to cache identifiers.
It turns out that he root path doesn't work well for some enviroments
and can result in bugs when used with eg. the PdoBackend and a
deployment that chnages the root path (typical Surf or Deployer).
The only backwards compatible way to fix this was to make the
applicationIdentifier configurable with a default that matches the
previously hardcoded values. That way nothing changes in existing
installations but if the bug appears it can be easily fixed.

### DIFF
--- a/Neos.Flow/Classes/Cache/CacheFactory.php
+++ b/Neos.Flow/Classes/Cache/CacheFactory.php
@@ -81,15 +81,16 @@ class CacheFactory extends \Neos\Cache\CacheFactory implements CacheFactoryInter
      *
      * @param ApplicationContext $context The current Flow context
      * @param Environment $environment
+     * @param string $applicationIdentifier
      * @Flow\Autowiring(enabled=false)
      */
-    public function __construct(ApplicationContext $context, Environment $environment)
+    public function __construct(ApplicationContext $context, Environment $environment, string $applicationIdentifier)
     {
         $this->context = $context;
         $this->environment = $environment;
 
         $environmentConfiguration = new EnvironmentConfiguration(
-            FLOW_PATH_ROOT . '~' . (string)$environment->getContext(),
+            $applicationIdentifier,
             $environment->getPathToTemporaryDirectory()
         );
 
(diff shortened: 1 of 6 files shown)


jonnitto:
@kitsunet The tests are still failing…

kitsunet:
Yep, yep, I see. during rebasing I coudln't run the tests so I hoped it would work, but I kind of expected that functional tests might also have a problem...

kdambekalns:
Ok, it get's shady here… when changing `Objects.yaml` and instead of `3` using `2` as argument index, the `var_dump()` suddenly emits the correct setup… as in:

```
/…/Packages/Framework/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php:402:
class Neos\Flow\ObjectManagement\Configuration\Configuration#1408 (12) {
  protected $objectName =>
  string(28) "Neos\Flow\Cache\CacheFactory"
  protected $className =>
  string(28) "Neos\Flow\Cache\CacheFactory"
  protected $packageKey =>
  string(9) "Neos.Flow"
  protected $factoryObjectName =>
  string(0) ""
  protected $factoryMethodName =>
  string(6) "create"
  protected $scope =>
  int(2)
  protected $arguments =>
  array(2) {
    [1] =>
    class Neos\Flow\ObjectManagement\Configuration\ConfigurationArgument#1465 (4) {
      protected $index =>
      int(1)
      protected $value =>
      string(17) "Neos.Flow.context"
      protected $type =>
      int(2)
      protected $autowiring =>
      int(1)
    }
    [2] =>
    class Neos\Flow\ObjectManagement\Configuration\ConfigurationArgument#1466 (4) {
      protected $index =>
      int(2)
      protected $value =>
      string(37) "Neos.Flow.cache.applicationIdentifier"
      protected $type =>
      int(2)
      protected $autowiring =>
      int(1)
    }
  }
  protected $properties =>
  array(0) {
  }
  protected $autowiring =>
  int(1)
  protected $lifecycleInitializationMethodName =>
  string(16) "initializeObject"
  protected $lifecycleShutdownMethodName =>
  string(14) "shutdownObject"
  protected $configurationSourceHint =>
  string(88) "configuration of package Neos.Flow, definition for object "Neos\Flow\Cache\CacheFactory""
}
```

PR #1458, opened 2018-11-21
BUGFIX: Tests adapted to new behavior in PackageManger

kitsunet:
The ``PackageManager::createPackage`` method now can create
packages into a possibly existing ``DistributionPackages``
directory "automagically", which then triggers a composer update.
While this is a good idea to push the idea of distribution packages
it also breaks tests that clearly do not want to run composer
update. This can be fixed by providing a defined package path.

Related #1433


kitsunet:
@jonnitto as this is related to another feature that went into the 5.2 branch this definitely should be merged as well.

PR #1460, opened 2018-11-23
BUGFIX: Make session caches persistent by default

### LINKED ISSUE #1390: Make Session caches persistent by default.
They are a prime target for persistent caching, because sessions normally shouldn't get deleted with general cache flushing.

See https://github.com/neos/neos-development-collection/issues/2175

albe:
Previously the sessions would be flushed whenever caches get flushed. This sometimes leads to unexpectedly logging out users, as happens for example in the neos setup in the last step.
This is only breaking in the sense, that you need to take care when deploying code that changes classes or data stored in the session. In those cases you need to flush the session caches manually via the new CLI command `./flow flow:session:destroyAll`. Also see the documentation https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/SessionHandling.html#session-storage

Resolves #1390

### DIFF
--- a/Neos.Flow/Classes/Command/CacheCommandController.php
+++ b/Neos.Flow/Classes/Command/CacheCommandController.php
@@ -128,7 +128,8 @@ class CacheCommandController extends CommandController
      * Flush all caches
      *
      * The flush command flushes all caches (including code caches) which have been
-     * registered with Flow's Cache Manager. It also removes any session data.
+     * registered with Flow's Cache Manager. It will NOT remove any session data, unless
+     * you specifically configure the session caches to not be persistent.
      *
      * If fatal errors caused by a package prevent the compile time bootstrap
      * from running, the removal of any temporary data can be forced by specifying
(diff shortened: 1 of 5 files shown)


albe:
For the record, I don't target this on 4.3 due to the "breakyness". It's still not breaky enough that I would hold this back until the next major. Up to decide for the RM @jonnitto though :)

kitsunet:
Not sure if this kind of change shouldn't go into a major, just for the quite big behavior change.

beheist:
I'd be happy with merging it into a minor. To me, the benefits are big enough that I'm fine with merging this in a minor - that setup bug is really extremely confusing to many users, and I'd love to see that fixed.

albe:
One thing I noticed, but which I can not fully assess: The persistent cache is not Context-specific, i.e. sessions would be shared among all existing contexts that way. This can be considered a good thing as well as a bad thing, depending on the use case. Since there is a workaround, namely configuring the `cacheDirectory` for the session in each context, I'd say it's okay as long as it is properly documented.

albe:
Added a small CLI command controller for destroying sessions more conveniently and added documentation. Also updated the `flow:cache:flush` command documentation to state that sessions are NOT deleted any more, so the Reference section of the docs will get updated when docs are generated from code with the release.

@jonnitto Are you okay with the documentation?

PR #1462, opened 2018-11-27
TASK: Fix duplicate FLOW_VERSION_BRANCH declaration after upmerge

albe:
A duplicate declaration of FLOW_VERSION_BRANCH slipped in while upmerging from 4.3


albe:
Will merge due to obviousness and triviality after CI pass.

PR #1464, opened 2018-11-30
TASK: Optimize parsing of resource uris

kitsunet:
This avoids the costly ``\Neos\Utility\Unicode\Functions::parse_url``
which does a preg_replace_callback. As we know that we don't expect
just any URL in the resource stream wrapper we might as well use
cheaper string functions to extract the necessary information.
Given that evaluateResourcePath is called a lot especially in a
Neos CMS context this can actually have a performance impact.


kitsunet:
If you wonder why, I recently saw a performance trace of a project with a few thousand preg_replace_callback calls (of which 99% originated within the above method) and that definitely had a performance impact.

kitsunet:
mmm, forgot to adjust all tests...

PR #1465, opened 2018-11-30
BUGFIX: Remove return type declaration, ask() can return mixed value

kdambekalns:


gerhard-boden:
https://github.com/neos/flow-development-collection/blob/9c9e134f787a880d7f03ec61d08c146a4a612bf8/Neos.Flow/Classes/Cli/ConsoleOutput.php#L238

should we also remove it here?

Issue #1467, opened 2018-12-04
File uploads with "/" in name doen't work


futape:
### Description

When uploading a file using an HTML `input` element with a `name` attribute containing a HTML array containing "/"s, the `$_FILES` array isn't untangles properly in the HTTP `Request`.

### Steps to Reproduce

1. Add an element `<input type="file" name="foo[bar/baz]">`
2. Upload a file
3. The `Neos\Flow\Http\Request::$uploadedFiles` property containg an empty array

#### Expected behavior

The array should look like this:

```
array(1)
 'foo' (9) => array(1)
   'bar/baz' (8) => Neos\Flow\Http\FlowUploadedFile
     originallySubmittedResource => NULL
     collectionName => NULL
     clientFilename => 'kirill-bubochkin-570219-unsplash.jpg' (36)
     clientMediaType => 'image/jpeg' (10)
     error => integer 0
     file => '/tmp/phpDBhfeI' (14)
     moved => false
     size => integer 1248317
     stream => NULL
```

#### Actual behavior

The array is empty since the `$_FILES` array isn't untangled properly.

### Affected Versions

Neos: -

Flow: 5.1.5

daniellienert:
Fixed with #1469

PR #1468, opened 2018-12-04
BUGFIX: Support "/" in file upload fields

futape:
Adds support for file uploads with "/"s in their names.
Adds a `UploadedFilesHelper::calculateFieldPathsAsArray` method
returning the paths as arrays instead of "/"-separated strings
(later "/"-split again).
Keeps the `calculateFieldPaths` method returning the paths as
strings for backwards-compatibility.

### DIFF
--- a/Neos.Flow/Tests/Unit/Http/RequestTest.php
+++ b/Neos.Flow/Tests/Unit/Http/RequestTest.php
@@ -817,7 +817,24 @@ class RequestTest extends UnitTestCase
                         ]
                     ]
                 ]
-            ]
+            ],
+            'f0' => [
+                'name' => [
+                    'f1/f2' => 'f.txt',
+                ],
+                'type' => [
+                    'f1/f2' => 'text/plain',
+                ],
+                'tmp_name' => [
+                    'f1/f2' => '/private/var/tmp/php5469a0',
+                ],
+                'error' => [
+                    'f1/f2' => 0,
+                ],
+                'size' => [
+                    'f1/f2' => 100,
+                ],
+            ],
         ];
 
         $untangledFiles = [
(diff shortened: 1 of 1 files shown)


futape:
Fixes #1467

daniellienert:
@futape: I guess this can be closed in favor of #1469 ?

Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php
```diff
@@ -68,13 +63,33 @@ function (&$value) {
...
+    {
+        $fieldPaths = self::calculateFieldPathsAsArray($structure, $firstLevelFieldName);
+        array_walk($fieldPaths, function (&$fieldPath) {
+            $fieldPath = implode('/', $fieldPath);
+        });
+        return $fieldPaths;
+    }
+
+    /**
+     * Returns and array of all possible "field paths" for the given array.
```
albe:
```suggestion
     * Returns an array of all possible "field paths" for the given array.
```

futape:
> Let's get #1469 in and upmerge, then rebase this in order to make sure no overlapping breaking happens. I can't see anything from just looking, but better be safe than sorry :)

#1469 is merged. whats the next step? does 5.1 has to be rebased on 4.3?

albe:
Upmerges (ie merging commits from lowest supported version upwards up until master) are done manually occasionally, latest on patch-level releases. I'll try to get them done today or tomorrow if no one else does them before. Afterwards, this branch just needs to be rebased on 5.1 (or 5.1 merged into this branch). Should be without conflicts and then travis just needs to pass again.

futape:
Alright. Thanks for the lesson :D

albe:
Ok, just did upmerges and in that process also basically had to already do all the changes in this PR. So after rebeasing, this will probably be an empty PR. Sorry :D

albe:
Guess those additional test data only dublicate the test in https://github.com/neos/flow-development-collection/pull/1468/files#diff-f49ffe9dac57e03e1ba6da125adf7e66R975, right? If so, this can be closed. Thanks for your effort! :)

futape:
Yes, the test is pretty much the same. I'll close this PR.

PR #1478, opened 2018-12-10
BUGFIX: Support nested fluid variables by using the original fluid method

vertexvaar:
With this patch it's possible to use the fluid variable nesting feature as well as a specific accessor per element on the path.

**How to use it**

Create a fluid template with nested variables that access an array value:
```html
{config.{type}.value.{flavor}}
```
Assign an array and the two keys:
```php
$this->view->assign('type', 'thing');
$this->view->assign('flavor', 'yellow');
$this->view->assign('config', ['thing' => ['value' => ['yellow' => 'Okayish']]]);
```


vertexvaar:
Whoops, there's a commit that should not be there. I gonna remove it.

kdambekalns:
Please provide a proper PR title and import the TYPO3Fluid namespaces (using an alias, if needed).

albe:
Though stale, still relevant.

albe:
Would like to update this and sometime actually get this in.

albe:
🤞 this should pass with the added test and then this is IMO good to go - finally

albe:
I still think this small regex fix could have landed in Fluid - it still doesn't allow deeply nested variables, but the case in the test should be supported.

albe (approved):

albe:
ping @kitsunet @kdambekalns

kdambekalns:
> Merging a higher branch into a lower branch sounds very wrong and dangerous – is this what you intended, @albe?

Ah, probably because the patch comes from a 5.0 branch on th fork but the base was changed. So this is correct, but highly confusing. 👍

albe:
Rebased in https://github.com/neos/flow-development-collection/pull/2444 to keep the history clean. If we had `rebase & merge` enabled we could solve it that way too :)

PR #1480, opened 2018-12-12
TASK: Fix risky unit test

kdambekalns:
The changed test does contain an `expects` but did advertise it
`@doesNotPerformAssertions`, leading PhpUnit to consider the
test as risky.


kitsunet:
Thanks!

PR #1481, opened 2018-12-12
TASK: Update 3rd party dependency constraints

kdambekalns:
This allows semantically non-breaking updates for a number of 3rd
party dependencies.


kdambekalns:
Yeah, and having those pulled in "early" with CI builds might expose issues to us faster. And waiting longer makes the "difference" larger, which in turn raises reluctancy to update things…

PR #1485, opened 2018-12-18
TASK: Tweak LazySplObjectStorage implementation

kdambekalns:
Aliases call their "original" internally now.


albe:
I see this makes sense from a strict DRY PoV, but somehow it feels awkward still. Especially since `offsetGet` is not "redirected" to something internal (so is there no internal for this, or is `offset*` the "internal"?). Would maybe the other way around make more sense?

PR #1487, opened 2018-12-19
TASK: Allow contains expression on entity properties

kitsunet:
This is very limited on purpose but allows to write policies like this:

`'isType("Whatever\Domain\Model\Party") && (!property("accounts").contains("context.securityContext.account"))'`

### DIFF
--- a/Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/PropertyConditionGenerator.php
+++ b/Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/PropertyConditionGenerator.php
@@ -210,6 +210,24 @@ class PropertyConditionGenerator implements SqlGeneratorInterface
     }
 
     /**
+     * @param $operandDefinition
+     * @return $this
+     * @throws InvalidPolicyException
+     */
+    public function contains($operandDefinition)
+    {
+        if (strpos($this->path, '.') !== false) {
+            throw new InvalidPolicyException(sprintf('The "contains" operator does not work on nested property paths (contained a "."). Got: "%s"', $this->path), 1545212769);
+        }
+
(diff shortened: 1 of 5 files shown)


kitsunet:
@kaystrobach That's the PR and branch to watch :)

kitsunet:
@kdambekalns still WIP but adding you already :)

kitsunet:
Alright, this now has a full set of tests for OneToMany and ManyToMany relations.

@kaystrobach I wouldn't refactor that construct in this change to not dilute what this is about. It's complicated enough as is and this code was there before. It can be beautified in a separate refactoring.

kaystrobach:
ok, have you tested this in flow 4.x already, or can i help with testing here?

kaystrobach:
do you need help with the docs?

kaystrobach:
any chance this will be backported to FLOW 4.x?

albe:
Well, this one is practically a feature, so it is supposed to land in master. Personally, I wouldn't mind having it in 4.x, but then, where would we draw the line? I'll let the others decide on this.

kaystrobach:
IMHO no BC, so this can be a new Feature Version (Semver, second number) 😉

daniellienert:
Hey @kitsunet,
Looks good by reading. There is no documentation addition included. Do we have any detailed documentation about the matcher methods at all?

kaystrobach:
There are only examples in the docs

kdambekalns:
Right, the only documentation as at https://neos.readthedocs.io/en/stable/CreatingASite/Security.html, AFAIK. We could definitely need a reference of all shipped "small" and still undocumented operations.

kaystrobach:
@kitsunet will this be backported in 4.x? would be awesome

PR #1488, opened 2018-12-21
BUGFIX: Fix browser to be able to handle relative URLs

skurfuerst:


albe:
Good catch, so this hints at some non-strict behaving web-server answering. I guess, it's still okay if we make up for this in the sense of the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle)

PR #1490, opened 2019-01-07
FEATURE: Make recursion limit for the debugger configurable via Settings

mficzel:
With the default recursion Limit of 50 php often runs into memory-limits when debugging larger data structures. This change allows to define the recursionLimit via Settings. In addition the default recursionLimit is set to 5.


kitsunet:
I am at a point in this where I would backport a bugfix setting the default to 5 even. And then have this.

mficzel:
The pr to lower the general recursion limit in Flow 4.3+ is here #1493

mficzel:
Rebased on current master after #1493 was merged and upmerged

PR #1491, opened 2019-01-07
TASK: Respect type hinting in reflection / support "@param Type[]" array syntax

Torsten85:
Currently the Reflection Service does not work well with the `@param Type[]` array syntax. Also `array` type hints without a `@param` annotation result in a `mixed` type. 

Through this PR the Reflection Service resolves `Type[]` to `array<Type>`. And `array` stays `array` even without annotation.

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1454,7 +1454,7 @@ class ReflectionService
         }
 
         // expand SomeElementType[]" to "array<\ElementTypeNamespace\SomeElementType>"
-        if (strpos($typeWithoutNull, '[]') === strlen($typeWithoutNull) - 2) {
+        if (substr_compare($typeWithoutNull, '[]', -2, 2) === 0) {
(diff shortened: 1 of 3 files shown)


Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1453,6 +1453,12 @@ protected function expandType(ClassReflection $class, $type)
             return $this->expandType($class, $type) . '<' . $this->expandType($class, $elementType) . '>' . ($isNullable ? '|null' : '');
         }
 
+        // expand SomeElementType[]" to "array<\ElementTypeNamespace\SomeElementType>"
+        if (strpos($typeWithoutNull, '[]') === strlen($typeWithoutNull) - 2) {
```
albe:
```suggestion
        if (substr_compare($typeWithoutNull, '[]', -2, 2) === 0) {
```
Is more understandable and also more performant, since it doesn't need to go through all of $typeWithoutNull to possibly find '[]'. Also, `strrpos` would probably be more correct - even though a type with multiple `[]` is invalid anway.

PR #1492, opened 2019-01-08
BUGFIX: Verify the existence of the `repositories` section in the composer manifest before searching for local packages path

mficzel:
In older setups that did not have a `repositories` section in the composer-maifest `package:create` tried to foreach over a null-value which lead to an php-error. This change checks that the `repositories` is actually an array before iterating.


albe:
The "Bug" was introduced https://github.com/neos/flow-development-collection/commit/3224fc194cfe21a6a033e8b8b7566c5d6877a4f2#diff-5f91f1e8fceb51b2dd58287631d5606c so the target branch is correct too. Thanks!

PR #1493, opened 2019-01-09
BUGFIX: Set recursion limit for debugger to 5

mficzel:
With the previous recursion limit of 50 php often ran into memory-limits when debugging larger data structures.


johannessteu:
Would be a great chance to make this configurable?!

mficzel:
@johannessteu yap there is already a feature pr for that https://github.com/neos/flow-development-collection/pull/1490 this here is a bugfix for older neos systems

PR #1495, opened 2019-01-14
TASK: Use proper dummy hash in PersistedUsernamePasswordProvider

kdambekalns:
This replaces the dummy hash used to prevent timing based attacks by
a valid hash for a random password that was never actually stored
somewhere.

This avoids problems with PHP's encryption methods. With the
previous hash, the hashing was sometimes not applied properly
and the method returns early so that the time-based information
disclosure vulnerability still exists.

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
+++ b/Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
@@ -97,7 +97,7 @@ class PersistedUsernamePasswordProvider extends AbstractProvider
         $authenticationToken->setAuthenticationStatus(TokenInterface::WRONG_CREDENTIALS);
 
         if ($account === null) {
-            $this->hashService->validatePassword($credentials['password'], 'bcrypt=>$2a$14$DummySaltToPreventTim,.ingAttacksOnThisProvider');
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
```diff
@@ -97,7 +97,7 @@ public function authenticate(TokenInterface $authenticationToken)
         $authenticationToken->setAuthenticationStatus(TokenInterface::WRONG_CREDENTIALS);
 
         if ($account === null) {
-            $this->hashService->validatePassword($credentials['password'], 'bcrypt=>$2a$14$DummySaltToPreventTim,.ingAttacksOnThisProvider');
+            $this->hashService->validatePassword($credentials['password'], 'bcrypt=>$2a$16$RW.NZM/uP3mC8rsXKJGuN.2pG52thRp5w39NFO.ShmYWV7mJQp0rC');
```
bwaidelich:
```suggestion
            // validate the account anyways (with a dummy salt) in order to prevent timing attacks on this provider
            $this->hashService->validatePassword($credentials['password'], 'bcrypt=>$2a$16$RW.NZM/uP3mC8rsXKJGuN.2pG52thRp5w39NFO.ShmYWV7mJQp0rC');
```
=> Good Catch! But now that the hash does not contain the hint about the intention we might add it as inline comment

kitsunet: 👍

albe:
Looks good, but the doc comment that Bastian suggested should be added.

PR #1501, opened 2019-01-25
FEATURE: Add methods `csrfToken`, `isAuthenticated` and `hasAccess` to `Security` EelHelper

mficzel:
- `csrfToken` returns CSRF token which is required for "unsafe" requests (e.g. POST, PUT, DELETE, ...)
- `isAuthenticated` returns true, if any account is currently authenticated
- `hasAccess` returns true, if access to the given privilege-target is granted

The new methods add features to eel that previously were available as viewHelpers so they can be used in Fusion directly.

### DIFF
--- a/Neos.Eel/Classes/Helper/SecurityHelper.php
+++ b/Neos.Eel/Classes/Helper/SecurityHelper.php
@@ -15,6 +15,8 @@ use Neos\Flow\Annotations as Flow;
 use Neos\Eel\ProtectedContextAwareInterface;
 use Neos\Flow\Security\Account;
 use Neos\Flow\Security\Context as SecurityContext;
+use Neos\Flow\Security\Authorization\PrivilegeManagerInterface;
+use Neos\Flow\Security\Authentication\TokenInterface;
 
 /**
  * Helper for security related information
@@ -29,6 +31,12 @@ class SecurityHelper implements ProtectedContextAwareInterface
     protected $securityContext;
 
     /**
+     * @Flow\Inject
+     * @var PrivilegeManagerInterface
+     */
+    protected $privilegeManager;
+
+    /**
      * Get the account of the first authenticated token.
      *
      * @return Account|NULL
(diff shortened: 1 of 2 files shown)


Neos.Eel/Classes/Helper/SecurityHelper.php
```diff
@@ -42,6 +50,47 @@ public function getAccount()
...
+    }
+
+    /**
+     * Returns true, if access to the given privilege-target is granted
+     *
+     * @param string $privilegeTarget The identifier of the privilege target to decide on
+     * @param array $parameters Optional array of privilege parameters (simple key => value array)
+     * @return boolean true if access is granted, false otherwise
+     */
+    public function hasAccess($privilegeTarget, $parameters = [])
```
daniellienert:
```suggestion
    public function hasAccess(string $privilegeTarget, array $parameters = []): bool
```

Neos.Eel/Classes/Helper/SecurityHelper.php
```diff
@@ -42,6 +50,47 @@ public function getAccount()
...
+    {
+        return $this->securityContext->getCsrfProtectionToken();
+    }
+
+    /**
+     * Returns true, if any account is currently authenticated
+     *
+     * @return boolean true if any account is authenticated
+     */
+    public function isAuthenticated()
```
daniellienert:
```suggestion
    public function isAuthenticated(): bool
```

Neos.Eel/Classes/Helper/SecurityHelper.php
```diff
@@ -42,6 +50,47 @@ public function getAccount()
         return null;
     }
 
+    /**
+     * Returns CSRF token which is required for "unsafe" requests (e.g. POST, PUT, DELETE, ...)
+     *
+     * @return string
+     */
+    public function csrfToken()
```
daniellienert:
```suggestion
    public function csrfToken(): string
```

Neos.Eel/Classes/Helper/SecurityHelper.php
mficzel: Adjusted as suggested, good point

Neos.Eel/Classes/Helper/SecurityHelper.php
mficzel: Adjusted as suggested, good point

PR #1503, opened 2019-01-31
BUGFIX: Set FLOW_VERSION_BRANCH to 5.2

daniellienert:


albe:
Yes, indeed, see the commit history https://github.com/neos/flow-development-collection/commits/7833ad328de91ded4ed9b58e34301a17bf8fe444/Neos.Flow/Classes/Core/Bootstrap.php
Something we might consider for improving the release process.

kdambekalns:
Of course we have an issue for automating that already: https://github.com/neos/flow-development-distribution/issues/30

PR #1507, opened 2019-02-20
BUGFIX: Accessing non existing array keys can cause a PHP Notice.

salvaracer:
Would be better to check the array keys before using them.

INFO: For some langauges like "zh" (china), the property "$this->rulesets" is
not filled properly. So accessing for example "$this->rulesets['zh']" will cause
a PHP Notice: "Undefined index".

**What I did**

In my fluidtemplate, i want to translate a language ID from "zh/Main.xlf" with this code:

```
{f:translate(
  id: 'lang.id.with.plurals.definition',
  package: 'Some.Package',
  source: 'Main.xlf',
  locale: 'zh',
  quantity: 2
)}
```

The definition of this language ID in my "zh/Main.xlf" is this:

```
<group id="lang.id.with.plurals.definition" restype="x-gettext-plurals">
		<trans-unit id="lang.id.with.plurals.definition[0]">
			<source>Acme default Foo</source>
			<target>Acme ZH Foo</target>
		</trans-unit>
		<trans-unit id="lang.id.with.plurals.definition[1]">
			<source>Acme default Foos</source>
			<target>"Acme ZH Foos</target>
		</trans-unit>
</group>
```

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/I18n/Cldr/Reader/PluralsReader.php
+++ b/Neos.Flow/Classes/I18n/Cldr/Reader/PluralsReader.php
@@ -150,7 +150,7 @@ class PluralsReader
             return self::RULE_OTHER;
         }
 
-        $ruleset = $this->rulesets[$locale->getLanguage()][$this->rulesetsIndices[$locale->getLanguage()]];
+        $ruleset = $this->rulesets[$locale->getLanguage()][$this->rulesetsIndices[$locale->getLanguage()]] ?? null;
 
         if ($ruleset === null) {
             return self::RULE_OTHER;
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/I18n/Cldr/Reader/PluralsReader.php
```diff
@@ -150,6 +150,10 @@ public function getPluralForm($quantity, Locale $locale)
             return self::RULE_OTHER;
         }
 
+        if (!isset($this->rulesets[$locale->getLanguage()])) {
+            return self::RULE_OTHER;
+        }
+
         $ruleset = $this->rulesets[$locale->getLanguage()][$this->rulesetsIndices[$locale->getLanguage()]];
```
dlubitz: What about easily:
`$ruleset = $this->rulesets[$locale->getLanguage()][$this->rulesetsIndices[$locale->getLanguage()]]; ?? null`

salvaracer: Thanks, that works fine too. I have updated my commit.

daniellienert:
Hey @salvaracer, 
seems like this is your first commit to the project. Thanks! We really appreciate it!
As this bug also affects previous versions, it should be targeted at the branch 4.3. 
Would you please do the changes again at 4.3 and add another PR (Retargeting the PR to a lower branch is not possible)

Thank you very much!
Daniel

salvaracer:
Yep I'm new to this project 🥇 So, "rebasing" my branch and "force-push" will not work properly?
Ok, tomorrow I can create a new PR based on branch 4.3.

salvaracer:
Okay, the PR was re-created here (https://github.com/neos/flow-development-collection/pull/1511) for branch 4.3.

PR #1510, opened 2019-02-21
TASK: Avoid rendering of meta http-equiv if location headers are sent

mficzel:
Now meta http-equiv is only rendered if a delay is requested.

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
@@ -323,11 +323,12 @@ abstract class AbstractController implements ControllerInterface
      */
     protected function redirectToUri($uri, $delay = 0, $statusCode = 303)
     {
-        $escapedUri = htmlentities($uri, ENT_QUOTES, 'utf-8');
-        $this->response->setContent('<html><head><meta http-equiv="refresh" content="' . intval($delay) . ';url=' . $escapedUri . '"/></head></html>');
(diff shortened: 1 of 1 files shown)


albe:
Looks good, but out of curiosity: did this not work with delay===0 previously?

kdambekalns:
It would have been nice to know why this is needed. Does it fix a bug? I understand with a Location header the content is useless. but still… this even changes the observable behavior of the system. :)

@albe no, the content was set unconditionally before…

albe:
That's what I was aiming at :) Though probably needless (or maybe not for some archaic clients?) I don't see it being disfunctional before the change. Since master will be next major, it's ok either way, but maybe deserves a bit more reasoning explanation.

mficzel:
@kdambekalns @albe As far as i understand the topic meta http-equiv was not a feature for stupid clients. It was for server-environments that were not able to control the headers. I am pretty sure that flow will not run in such environments anyways so those bits are just waste.

kitsunet:
No, I am pretty sure it was for clients that didn't understand headers at some point. Although I cannot imagine what client that would be still these days. IMHO cleanup is good and an empty redirect is definitely preferable.

I did a quick research:
Location is around since HTTP/1.0, the W3C discourages use of http-equiv refresh for redirecting (it can make sense for other uses, eg. periodical content refresh without JS). Finally a search didn't turn up any results for clients that do not support location headers.

PR #1511, opened 2019-02-26
BUGFIX: Accessing non existing array keys can cause a PHP Notice.

salvaracer:
Would be better to check the array keys before using them.

INFO: For some langauges like "zh" (china), the property "$this->rulesets" is
not filled properly. So accessing for example "$this->rulesets['zh']" will cause
a PHP Notice: "Undefined index".

**What I did**

In my fluidtemplate, i want to translate a language ID from "zh/Main.xlf" with this code:

```
{f:translate(
  id: 'lang.id.with.plurals.definition',
  package: 'Some.Package',
  source: 'Main.xlf',
  locale: 'zh',
  quantity: 2
)}
```

The definition of this language ID in my "zh/Main.xlf" is this:

```
<group id="lang.id.with.plurals.definition" restype="x-gettext-plurals">
		<trans-unit id="lang.id.with.plurals.definition[0]">
			<source>Acme default Foo</source>
			<target>Acme ZH Foo</target>
		</trans-unit>
		<trans-unit id="lang.id.with.plurals.definition[1]">
			<source>Acme default Foos</source>
			<target>"Acme ZH Foos</target>
		</trans-unit>
</group>
```

### DIFF
--- a/Neos.Flow/Classes/I18n/Cldr/Reader/PluralsReader.php
+++ b/Neos.Flow/Classes/I18n/Cldr/Reader/PluralsReader.php
@@ -150,7 +150,7 @@ class PluralsReader
             return self::RULE_OTHER;
         }
 
-        $ruleset = $this->rulesets[$locale->getLanguage()][$this->rulesetsIndices[$locale->getLanguage()]];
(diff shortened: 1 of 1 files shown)


daniellienert:
Thank you @salvaracer!

PR #1518, opened 2019-03-17
TASK: add pushResult for Property example

albe:
This is a direct copy of https://github.com/neos/flow/pull/19 - thanks @kaystrobach


kaystrobach:
Thank you for merging

PR #1520, opened 2019-03-19
BUGFIX: Fix usage of SubContexts in ProxyClassLoader

jrenggli:
Previously there were three different implementations of the composition of a temporary directory path name. The one in ProxyClassLoader didn't use SubContexts correctly.
This change unifies all implementations to a central static method `composeTemporaryDirectoryName()`.

Previous error message:
```
include(): Failed opening '/var/www/myproject/Data//Temporary/Development/Local/AvailableProxyClasses.php' for inclusion (include_path='.:/usr/share/php')
```

### DIFF
--- a/Neos.Flow/Classes/Core/Bootstrap.php
+++ b/Neos.Flow/Classes/Core/Bootstrap.php
@@ -18,6 +18,7 @@ use Neos\Flow\Core\Booting\Scripts;
 use Neos\Flow\Exception as FlowException;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 use Neos\Flow\SignalSlot\Dispatcher;
+use Neos\Flow\Utility\Environment;
 use Neos\Utility\Files;
 
 /**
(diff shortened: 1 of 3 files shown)


kitsunet:
Thanks! I am rather confused that this wasn't noticed before though^^
Seems like something that should be noticeable right away.

The only thing I am not 100% sold on is a static method on the Environment class. I rather not want to give this class more power, especially a widely usable static method. I'll have a look for a better place.

Neos.Flow/Classes/Core/Bootstrap.php
```diff
@@ -534,7 +535,7 @@ protected function defineConstants()
 
         if (!defined('FLOW_PATH_TEMPORARY_BASE')) {
             define('FLOW_PATH_TEMPORARY_BASE', self::getEnvironmentConfigurationSetting('FLOW_PATH_TEMPORARY_BASE') ?: FLOW_PATH_DATA . '/Temporary');
-            $temporaryDirectoryPath = Files::concatenatePaths([FLOW_PATH_TEMPORARY_BASE, str_replace('/', '/SubContext', (string)$this->context)]) . '/';
+            $temporaryDirectoryPath = Environment::composeTemporaryDirectoryName(FLOW_PATH_TEMPORARY_BASE, $this->context);
             define('FLOW_PATH_TEMPORARY', $temporaryDirectoryPath);
```
albe: Why don't we use `FLOW_PATH_TEMPORARY` everywhere instead?

albe:
ping @kitsunet should we merge this as is? (after rebasing on 5.3)

albe:
Will merge this now. Not super happy about the static method on Environment either, but don't have a really good place either. It's not `@api` and shouldn't be used as such, but if it is, so be it. We have way worse methods (ab)used in the wild.

PR #1528, opened 2019-03-25
BUGFIX: add `template` and `psalm` to ignoredTags

dimaip:
Fixes: #1529

**What I did**
Added new annotation tags used by Doctrine to ignoredTags list

**How to verify it**
Just run any Flow command.

**Checklist**

- [x] Code follows the PSR-2 coding style
- [ ] Tests have been created, run and adjusted as needed
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)

PR #1530, opened 2019-03-25
BUGFIX: Ignore template/psalm annotations

### LINKED ISSUE #1529: Latest Doctrine Collections break reflection (template / psalm annotation)
### Description

Running `doctrine:migrate` is broken with a recent Doctrine Collections versions.

### Steps to Reproduce

1. Update packages
2. run `flow doctrine:migrate`

#### Expected behavior

Migrations are done as needed.

#### Actual behavior

`[Semantical Error] The annotation "@template" in class Doctrine\Common\Collections\Selectable was never imported.` or `[Semantical Error] The annotation "@psalm" in method Doctrine\Common\Collections\Selectable::matching() was never imported.` depending on how far you fixed the issue already.

See also https://github.com/doctrine/collections/issues/180 - the annotations were added with https://github.com/doctrine/collections/pull/177

### Affected Versions

Flow: (at least) 4.3 and up with Doctrine Collections 1.6.0 and up

kdambekalns:
Fixes #1529 by adjusting the configuration.

### DIFF
--- a/TYPO3.Flow/Configuration/Settings.yaml
+++ b/TYPO3.Flow/Configuration/Settings.yaml
@@ -408,6 +408,8 @@ TYPO3:
         'AfterStep': TRUE
         'WithoutSecurityChecks': TRUE
         'covers': TRUE
+        'template': TRUE
+        'psalm': TRUE
 
       # If enabled, the Reflection Service notes all incorrect or inconsistent usage
       # of @param annotations in the default log.


kdambekalns:
Ok, the same change existed as https://github.com/neos/flow-development-collection/pull/1528 already, but there wasn't an issue, so I did not realize.

kdambekalns:
But, that PR olny goes against Flow 4.3, but this affects earlier versions as well…

dimaip:
@kdambekalns earlier? I thought 4.3 was the oldest maintained, no?

kdambekalns:
Right, security-fixes only for Neos 2.3 to 3.2.

But if I didn't misread the composer dependencies, we (through transitive dependencies) allow Collections 1.6 in those. So If there is ever the need to update (or install from scratch) any of those, it would break.

dimaip:
Ah, security, keep on forgetting about that....

PR #1531, opened 2019-03-25
!!!TASK: Introduce ActionResponse in preparation for clean PSR-7

kitsunet:
This is the continuation to a clear separation between MVC and HTTP stacks.
The introduced ActionResponse offers a very limited interface to work with
on the MVC level, with the following methods:

* setContent()
* setContentType()
* setRedirectUri()
* setStatusCode()
* setComponentParameter()

Everything in the currently used HTTP\Response is still available but deprecated and will be removed in the next major, so make sure to adapt to the above API.

This change is marked breaking due to the deprecations, it should not break any existing code.

setContent() accepts also PSR-7 StreamInterface implementations and will likely only accept those in the next major.

A detailed blog post with the next steps will follow and go into more detail about the final separation and usage.

Special note regarding the `setComponentParameter`, for now this is your extensible portal towards HTTP. You can use it to set component parameters for your own HTTP components to set additional headers.

We are likely to extend the interface slightly for the major release as we are aware that this implementation is very limiting, but we need a clean separation between MVC and HTTP to start with.

Related discuss post:
https://discuss.neos.io/t/rfc-psr-7-continuation/3597/10


kitsunet:
Ok, this is the first shot at implementing something for inbetween. Without risking b/c breaks this is probably the best we can do direction wise. It's not beautiful but it works without changes to userland code.

We need to discuss the dispatcher and appendContent. I suggest in the major change we provide each controller invocation with a separate ActionResponse and merge them in the dispatcher or HTTP component (TBD). Also interesting question if we want to introduce streams for the content (as PSR-7 does) to avoid moving around big strings...

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * @param int $statusCode
+     * @return void
+     * @api
+     */
+    public function setStatusCode(int $statusCode): void
+    {
+        $this->statusCode = $statusCode;
+        // TODO: This can be removed after the full changes are done for next major.
+        $this->setStatusCode($statusCode);
```
bwaidelich: this will die in recursion in recursion

Neos.Flow/Classes/Mvc/ActionResponseInterface.php
```diff
@@ -0,0 +1,48 @@
+<?php
+namespace Neos\Flow\Mvc;
+
+use Psr\Http\Message\UriInterface;
+
+interface ActionResponseInterface
```
bwaidelich: Why this new interface?

kitsunet: So you can dev against the interface, as the implementation right now carries all the bagage of the old HTTP Response.

Neos.Flow/Classes/Mvc/ActionResponse.php
kitsunet: fixed

Neos.Flow/Classes/Mvc/ActionResponseInterface.php
bwaidelich: Mh. I think that's counter productive TBH (but that might be related to my jetlag *g). Lets talk about this tomorrow if you wish!?

kitsunet: Of course, I was hesitant as well, but I felt it helped to point to some (interface) definition that people can code against until the major. I don't think we absolutely need it, but for a dev not as close into the process I am sure it will help. Let's talk tomorrow :)

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -0,0 +1,125 @@
...
+use Psr\Http\Message\UriInterface;
+
+/**
+ * The new minimal MVC response object.
+ * For anything more use a custom HTTP component and set a component parameter.
+ *
+ * @Flow\Proxy(false)
+ * @api
+ */
+final class ActionResponse extends \Neos\Flow\Http\Response implements ActionResponseInterface
```
albe: extends `Http\Response` gives me the shivers, even if it's just for now to be b/c 🙈

Neos.Flow/Classes/Mvc/ActionResponseInterface.php
albe: Yep, see this more as a helper for the time of change. Maybe we can add this in the doc block and already mark it @deprecated?

Basically, the "right" thing would be to `rename` the current `Mvc\ResponseInterface` to `Mvc\ActionResponseInterace` and deprecate it at the same time, but since the signature changes a lot and renaming would be breaking, this is the way to go.

Neos.Flow/Classes/Mvc/ActionResponseInterface.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    /**
+     * Set content mime type for this response.
+     *
+     * @param string $contentType
+     * @return void
+     * @api
+     */
+    public function setContentType(string $contentType): void;
+
```
albe: I'm missing a `addHeader(string $name, string $value)` method, helping with e.g. setting CORS or cache-control headers in the action.
If we want to elaborate on this, we could maybe design `setAllowHeader()` and `setCacheControl()` methods instead of the 100% generic `addHeader()`.

Neos.Flow/Classes/Mvc/Dispatcher.php
```diff
@@ -130,6 +130,8 @@ public function dispatch(RequestInterface $request, ResponseInterface $response)
      * @param RequestInterface $request
      * @param ResponseInterface $response
      * @throws InvalidControllerException|InfiniteLoopException|NoSuchOptionException
+     *
+     * TODO: From next major we might want to create one ActionResponse PER controller invocation and then merge results here
```
albe: 👍

Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -21,6 +21,7 @@
  *
  * @api
  * @Flow\Proxy(false)
+ * @deprecated
```
bwaidelich: Please lets always add the version since when this is deprecated and a short comment on why and how to replace it

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
```diff
@@ -30,9 +30,9 @@ abstract class ResponseInformationHelper
      * @throws \InvalidArgumentException
      * @return ResponseInterface
      */
-    public static function createFromRaw(string $rawResponse): ResponseInterface
+    public static function createFromRaw(string $rawResponse, string $responseClassName = Response::class): ResponseInterface
     {
-        $response = new Response();
+        $response = new $responseClassName();
```
bwaidelich: Shouldn't we check that the result actually is an implementation of `ResponseInterface` here?

Neos.Flow/Classes/Mvc/ActionResponseInterface.php
bwaidelich: OK, it wasn't just the jetlag. I still don't think that we need a new interface here. And an interface that is deprecated right from the start? why?
See main comment for my suggestion on how to approach this

Neos.Flow/Classes/Mvc/ActionResponseInterface.php
bwaidelich: IMO that would be a case of "unplanned extensibility" described in the original [post](https://discuss.neos.io/t/rfc-psr-7-continuation/3597/11)

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -338,12 +341,15 @@ protected function redirectToUri($uri, $delay = 0, $statusCode = 303)
...
      *
+     * TODO: statusMessage argument is deprecated and should be removed in next major.
+     *
      * @param integer $statusCode The HTTP status code
      * @param string $statusMessage A custom HTTP status message
      * @param string $content Body content which further explains the status
      * @throws UnsupportedRequestTypeException If the request is not a web request
      * @throws StopActionException
      * @api
+     * @deprecated This method will no longer accept a status message after next major, but still exist.
```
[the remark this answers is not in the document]
bwaidelich: According to the [spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html) the status message `...MAY be replaced by local equivalents without affecting the protocol` so I don't think that we should change the signature here (especially because it would be a very nasty breaking change because the message would suddenly get the response content

kitsunet:
Note your main comment runs into a big trap. Mvc\Response is currently not even used, so extending it won't help at all. Currently it's a Http\Response so we need to extend that (which I did) but that class currently carries a lot of baggage around with PSR methods AND original Flow HTTP methods. So the small set of methods you should actually use gets drowned... That was the original reason for the interface.
I guess we should discuss this together :)

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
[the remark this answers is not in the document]
kitsunet: I will remove the comment, BUT that would mean we keep the unused param until eternity?

kitsunet:
The new commit is the result of a short discussion with bwaidelich and I really like it. A bit more code but nicely separated.

Neos.Flow/Classes/Mvc/ActionResponseInterface.php
albe: I don't consider setting CORS or caching headers an unplanned extensibility tbqh. It's HTTP basics and highly dependent on the concrete action invoked (this is important, those headers are not global), so moving such decisions to an extra component is bad practice IMO, because it's harder to understand the caching and cross-origin behaviour of an action.

kitsunet:
Alright tests are green, YAY
so @bwaidelich if you would re-asses your review and maybe @albe also does a review.

I have two questions/considerations though:

Do we want to use an PSR-7 UriInterface for the new setRedirect argument? IMHO much nicer as a string or both, but obviously UriInterface is strictly from a foreign domain. I am still fine with it, that's why I left it that way. We could obviously introduce a DTO for this but compatibility in the places were we need it makes the UriInterface a good candidate.

Second consideration that I didn't put in place yet, but important for the interface and IMHO a good idea would be to use a stream or resource for the content so we don't carry strings around.

So either the nicr way would be the PSR-7 StreamInterface, alternatively we could just accept a resource (so an open stream like php://temp). Either is generally be neater than just strings IMHO. The PSR-7 again has the benefit of having good implementations, good compatibility to were we will need this and nice type hinting.

So let me know what you think.

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
bwaidelich: The 2nd parameter is no longer used (is it?) so we could remove it again

Neos.Flow/Classes/Http/Response.php
```diff
@@ -20,6 +20,7 @@
  *
  * @api
  * @Flow\Proxy(false)
+ * @deprecated
```
bwaidelich:
```suggestion
 * @deprecated Since Flow 5.3, to be removed with 6.0. The whole Flow HTTP implementation will use PSR-7 compatible implementations only from then.
```

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -0,0 +1,104 @@
+<?php
+namespace Neos\Flow\Mvc;
+
+use Neos\Flow\Annotations as Flow;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * The new minimal MVC response object.
```
bwaidelich: nitpick: "new" is relative, especially in code that is meant to stick around for a while :)
But if we remember to adjust the comment when removing the `extends ...` for the next major, this is not a problem

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -0,0 +1,104 @@
...
+ * For anything more use a custom HTTP component and set a component parameter.
```
bwaidelich:
```suggestion
 * It allows for simple interactions with the HTTP response from within MVC actions. More specific requirements can be implemented via HTTP Components. @see setComponentParameter() 
```

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -0,0 +1,104 @@
...
+     * @var string
+     */
+    private $contentType = '';
+
+    /**
+     * @param string $content
+     * @return void
+     * @api
+     */
+    public function setContent($content): void
```
bwaidelich:
```suggestion
    public function setContent(string $content): void
```

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -0,0 +1,104 @@
...
+        // TODO: This can be removed after the full changes are done for next major.
+        $this->headers->set('Content-Type', $contentType);
+    }
+
+    /**
+     * Set a redirect URI and according status for this response.
+     *
+     * @param UriInterface $uri
+     * @param int $statusCode
+     * return void
```
bwaidelich:
```suggestion
     * @return void
```

albe:
In favor of UriInterface - we only depend on the interface definition here, so this is fine.

Same for the content, let's go towards PSR-7 StreamInterface, which is generic enough and opens up our stack for a lot of new use-cases. For the simple cases (return a fixed string), we could mabye provide a simple enough implementation and then do something like `return new StringStream("Some constant string");`. Or, we just allow `StreamInterface|string` in `setContent()` and do the necessary wrapping inside, to keep the API more aproachable at the cost of some missing type hinting.

bwaidelich:
> Same for the content, let's go towards PSR-7 StreamInterface

But how would `ActionResponse:: mergeToHttpResponse()` and `ActionResponse:: mergeActionResponse()` work with that?

albe:
@bwaidelich
> how would you merge the action responses if they return streams?

We could create a writeable temp stream and just pipe all streams in order to that, then push it to the Http stack. The main question is more, what is the correct order, but that is the same for strings as for streams.

bwaidelich:
> We could create a writeable temp stream and just pipe all streams in order to that

Wouldn't that be quite a (performance) overhead for the regular(tm) case of rendering a parsed template?

albe:
Don't really know how much overhead a temp/memory stream is over "simple" string concatenation. Or we move the "Stream conversion" down to the merging process. Then as long as we have only strings (which the rendered templates are), we can just concat them and return a single `Stream(concatenatedTemplatesString)` in the `mergeToHttp` part.

ie. `ActionResponse[string]::merge(ActionResponse[string]) : ActionResponse[string]`,
`ActionResponse[stream]::merge(ActionResponse[string]) : ActionResponse[stream]`
`ActionResponse[string]::merge(ActionResponse[stream]) : ActionResponse[stream]`

Does that make sense?

bwaidelich:
> Or we move the "Stream conversion" down to the merging process

Yes, that's what I thought. Since `Http\Reponse::content` is a stream, this has to be done anyways

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
albe: Good point.

mficzel:
Flow 5.3 has been branched. If this gets ready please merge directly to 5.3 and not to master

Neos.Flow/Classes/Mvc/ActionResponse.php
kitsunet: I think we decided to go more for streaminterface here, so not commiting this.

kitsunet:
Alright, this should be the final state for this after todays discussion @bwaidelich @albe 
I will open a separate PR for the component things.

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
albe: @kitsunet ?

kitsunet:
@bwaidelich @albe the parameter actually IS still used in `\Neos\Flow\Mvc\ResponseDeprecationTrait::createFromRaw` and that's also where we need it for. We need to create the correct instance type there. Or did I miss something?

mficzel:
Some documentation and a commit message are needed aswell. I assume this will be done once the details are sorted

mficzel:
Is this pr possibly causing this issue here? https://github.com/neos/neos-development-collection/issues/2437 @albe @kitsunet @bwaidelich

PR #1533, opened 2019-03-31
FEATURE: Add BEM Eel helper

jonnitto:
This two Eel helper generates BEM classes as an array or string. The modifiers property can be a string (for one modifier), an array (e.g. `['one', 'two']`) or an array with keys and values. If you have an array with keys and values  (like a Neos.Fusion:DataStructure) and a value is `true` the name of the key gets used for the modifier.

**Checklist**

### DIFF
--- a/Neos.Eel/Classes/Helper/ArrayHelper.php
+++ b/Neos.Eel/Classes/Helper/ArrayHelper.php
@@ -492,6 +492,48 @@ class ArrayHelper implements ProtectedContextAwareInterface
     }
 
     /**
+     *
+     * Generates BEM classes as an array. The modifiers property can be a string
+     * (for one modifier), an array (e.g. `['one', 'two']`) or an array with keys and values.
+     * If you have an array with keys and values  (like a Neos.Fusion:DataStructure) and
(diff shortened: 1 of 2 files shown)


jonnitto:
@skurfuerst This is the helper we talked about in Dresden. Drupal has a similar helper…

Wdyt?

dimaip:
Hi!
I like BEM, but to me it feels as something opinionated and thus userlandish. So I would not put it into core.
No hard opinions though.

johannessteu:
I'd also say that this is something that should not be part of the core.

bwaidelich:
As just discussed via Slack: I find it interesting and I like BEM as well but I also think that this should be a separate package

davidspiola:
I also think that this will have lots of benefits. How about releasing it as a flowpack?

bwaidelich:
We had similar requirements in a project and created some helpers similar to these but with a slight twist. I just compiled them into a package: https://packagist.org/packages/wwwision/bem
Feedback is more then welcome and I'm happy to port this to the flowpack namespace if that's wanted (though I don't really see the need for that tbh)

PR #1537, opened 2019-04-06
BUGFIX: Avoid type error when a non taggable cache backend gets flushed by tag

albe:
The typehint of the `flushByTag` method expected an `int` return type, but the method inside the `AbstractFrontend` returned void when a non taggable backend was flushed. This was the case for a `SimpleFileBackend` for example and led to an error.


albe:
BTW: This is targeting 5.0 because the issue (typehint) was added with that version

Issue #1077, opened 2017-09-10
Separate hard dependency on Account class and introduce a interface


sorenmalling:
### Description

Based on the topic form Discuss "[Could we build a version of Flow without doctrine orm / database requirement](https://discuss.neos.io/t/could-we-build-a-version-of-flow-without-doctrine-orm-database-requirement/2187)"

This is a umbrella task on the topic "Move the Security account to it's own package"

### Steps to be completed

- [ ] Find out what parts we are going to "free" from the core
    - [ ] Authentication (Provider and token classes)
    - [ ] Authorization
      - [ ] What about the Firewall?
    - [ ] Policy/Roles
    - [ ] Domain related stuff (Model, Repository and Factory)

- [ ] Find out what configurations that will be removed and what effects it have on the core application. 
  - [ ] Find out if they can be completely avoided
  - [ ] Find out what would be appropriate defaults

- [ ] Create the "common" security package, containing a interface
  - [ ] Find out how to replace the current Account (model) implementation, with a interface without

- [ ] Create a security package, implementing the Account domain model as we know it today
  - [ ] Based on the UsernamePassword Provider & Token, recreate the same functionality.

### Some big questions

**How do we handle current migrations that has anything to do with the account table?**

**How do we ensure that a upgrade to ex. Flow 5, where this concept will be introduced, doesn't break any current usage of the Account and security - what migration wizard steps would we have to take?**

**Are we looking at extending the current configuration for ex. ViewHelper  or other places, where the new AccountInterface is being used? For ex. to determine what provider is used**

sorenmalling:
@albe Hey - can we have this added to 6.2 goal? I will put heart and soul in getting this in :)

albe:
We can put it on the track for now, but be aware that following our timeline, there is less than a month until release! I'll happily push this as far as it can get in a fully b/c way into the next release, but I can't promise it will pass through this time.

I will consult with Christian when we do want to declare our feature freeze and if the current global situation should have effect on our schedule.

albe:
We at least have the `PasswordUsernameToken` stuff in 6.2 - and I guess the breaking changes in the AccountInterface change need to be a bit further investigated and tested. Maybe it's still viable for the 6.3 LTS

PR #1538, opened 2019-04-06
!!! FEATURE: Introduce Account Interface

sorenmalling:
Based upon this discuss.neos.io thread

https://discuss.neos.io/t/could-we-build-a-version-of-flow-without-doctrine-orm-database-requirement/2187/16

This change introduces a AccountInterface to serve for handling
with non-persisted Account during authentication as well as
the possibility of a base interface to implement authentication
classes.

**Checklist**

- [x] AccountIdentifier class
- [x] AccountProviderName class
- [x] CredentialsSource class
- [x] Roles collection 
- [x] Replace usage of getAuthenticationProviderName?
- [x] Replace usage of getIdentifier?
- [x] Replace/adjust relations to the `Account` entity with `AccountInterface` 
- [x] ... ?
- ~The PR is created against the lowest maintained branch~

**Breaking change**

This is a breaking change, since the `getCredentialsSource` of the AccountInterface class no longer return `mixed`, but a exact object. This is shown in the `PersistedUsernamePasswordProvider` class, where the password is cated to string, taking a `__toString` method into effect.

Related #1077


sorenmalling:
It’s this that is failing

`$this->mockAccount->expects($this->once())->method('getCredentialsSource')->will($this->returnValue('8bf0abbb93000e2e47f0e0a80721e834,80f117a78cff75f3f73793fd02aa9086'));`
since it returns a object (`CredentialsSource` ) - I either create a `getCredentialsSource()` method inside the `CredentialsSource` class (kind of double-double-naming) or cast it to string (somehow)? What way should I go?

sorenmalling:
Not being remotely into how Unit test and object mocking works, I would some feedback on how to get the Unit tests to pass.

Inside `Neos\Flow\Tests\Unit\Security\Authentication\Provider\PersistedUsernamePasswordProviderTest` the methods `authenticatingAnUsernamePasswordTokenChecksIfTheGivenClearTextPasswordMatchesThePersistedHashedPassword` and `authenticationFailsWithWrongCredentialsInAnUsernamePasswordToken` contains the following line

```
$this->mockAccount->expects($this->once())->method('getCredentialsSource')->will($this->returnValue('8bf0abbb93000e2e47f0e0a80721e834,80f117a78cff75f3f73793fd02aa9086'));
```

But since the `getCredentialsSource` method now returns a Object, how should I go around this? I either create a `getCredentialsSource()` method inside the `CredentialsSource` class (kind of double-double-naming) or cast it to string (somehow)? What way should I go? Or a third way to adjust these Unit tests?

Neos.Flow/Classes/Security/AccountIdentifier.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Neos\Flow\Security;
+
+class AccountIdentifier {
```
daniellienert: Hey @sorenmalling, 
Would be good to have the newly introduced String-value objects look similar. Please have a look at https://github.com/neos/neos-development-collection/blob/master/Neos.Media/Classes/Domain/ValueObject/Configuration/Label.php - I suggest to adjust this one and the others introduced: No getter, to string method, final ...

Neos.Flow/Classes/Security/AccountInterface.php
```diff
@@ -0,0 +1,24 @@
...
+	/**
+	 * @return AuthenticationProviderName
+	 */
+	public function getProviderName() : AuthenticationProviderName;
+
+	/**
+	 * @return mixed
+	 */
+	public function getCredentialsSource() : CredentialsSource;
+}
```
daniellienert: Codestyle demands a newline at the end of files

Neos.Flow/Classes/Security/AccountInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
```
daniellienert: Please add a strict_type declaration to all introduced files (https://discuss.neos.io/t/usage-of-php-strict-typing-by-default/3861)

daniellienert:
About the tests: 

Just replace:

 `$this->mockAccount->expects($this->once())->method('getCredentialsSource')->will($this->returnValue('8bf0abbb93000e2e47f0e0a80721e834,80f117a78cff75f3f73793fd02aa9086'));`

with  

`$this->mockAccount->expects($this->once())->method('getCredentialsSource')->will($this->returnValue(new CredentialSource('8bf0abbb93000e2e47f0e0a80721e834,80f117a78cff75f3f73793fd02aa9086')));`

Didn't have a look where your value objects are used in the code, but please at a `(string)` cast everywhere where they are used as implicit casting will fail once we declare strict typing in these classes.

Neos.Flow/Classes/Security/AccountInterface.php
sorenmalling: This is a PSR-2 thing? So I must change to spaces for indention I guess :)

Neos.Flow/Classes/Security/AccountInterface.php
sorenmalling: I don't know how to mark this is as done so it's not marked as "blocking"?

sorenmalling:
All checks are green 🎉

Neos.Flow/Classes/Security/Account.php
```diff
@@ -177,15 +187,23 @@ public function setAuthenticationProviderName($authenticationProviderName)
         $this->authenticationProviderName = $authenticationProviderName;
     }
 
+    /**
+     * @return AuthenticationProviderName
+     */
+    public function getProviderName(): AuthenticationProviderName
```
daniellienert: Wait, why didn't you change `getAuthenticationProviderName` but introduced a new method? `getAuthenticationProviderName` is used a lot in the authentication code

sorenmalling: A matter of B/C mentioned in https://discuss.neos.io/t/could-we-build-a-version-of-flow-without-doctrine-orm-database-requirement/2187/17?u=sorenmalling by @bwaidelich Try and avoid breaking more than nescessary and maybe deprecate over time.

sorenmalling: I could replace it with the return of a object and everywhere in the core, cast it to a string. And write a !!! about it for the commit message, so everyone building for Flow 6 can adjust there code?

daniellienert: I would prefer a breaking change now over having two methods that return the same(ish) thing and also having the new one being less descriptive than the old. WDYT @bwaidelich ?

sorenmalling: "BC" edition just comitted :)

sorenmalling: We have the same case with getIdentifier / getAccountIdentifier - i will replace and adjust that tomorrow

sorenmalling: Just created a new commit that uses the AccountIdentifier object as return value

sorenmalling:
I suggest that the task

`Replace/adjust relations to the Account entity with AccountInterface`

Is being a follow-up task, once we got the interface in and I will create a separate issue for that

Neos.Flow/Classes/Security/AccountIdentifier.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
```
daniellienert: This blank line should be removed

bwaidelich: nitpick: empty line can be removed

Neos.Flow/Classes/Security/AccountIdentifier.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace Neos\Flow\Security;
```
bwaidelich: Maybe a sub namespace (e.g. "Account") makes sense for the new classes? (not sure about it)

Neos.Flow/Classes/Security/AccountIdentifier.php
```diff
@@ -0,0 +1,30 @@
...
+
```
bwaidelich: A (short) doc comment describing what an `AccountIdentifier` is and does would be good.
Also we might consider marking it with
```
@Flow\Proxy(false)
```
to prevent the creation of a proxy class (until #1539 is resolved).

And, lastly, it would be a nice measure to implement the `\JsonSerializable` interface here and then have
```php
    public function jsonSerialize(): string
    {
        return $this->identifier;
    }
```
so that this behaves correct when being serialized

Neos.Flow/Classes/Security/AccountIdentifier.php
```diff
@@ -0,0 +1,30 @@
...
+final class AccountIdentifier
+{
+
+    /**
+     * @var string
+     */
+    protected $identifier;
```
bwaidelich:
```suggestion
    private $identifier;
```

Neos.Flow/Classes/Security/AccountIdentifier.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    public function __construct(string $identifier)
+    {
+        $this->identifier = $identifier;
+    }
+
+    /**
+     * @return string
+     */
+    public function __toString()
```
bwaidelich:
```suggestion
    public function __toString(): string
```

Neos.Flow/Classes/Security/AccountInterface.php
```diff
@@ -0,0 +1,27 @@
...
+     */
+    public function getAccountIdentifier(): AccountIdentifier;
+
+    /**
+     * @return AuthenticationProviderName
+     */
+    public function getAuthenticationProviderName(): AuthenticationProviderName;
+
+    /**
+     * @return mixed
```
bwaidelich:
```suggestion
     * @return CredentialsSource
```

Neos.Flow/Classes/Security/Aspect/LoggingAspect.php
```diff
@@ -65,7 +65,7 @@ public function logManagerAuthenticate(JoinPointInterface $joinPoint)
         $authenticationManager = $joinPoint->getProxy();
         $logMessage = 'No account authenticated';
         if ($authenticationManager->getSecurityContext()->getAccount() !== null) {
-            $logMessage = sprintf('Successfully re-authenticated tokens for account "%s"', $authenticationManager->getSecurityContext()->getAccount()->getAccountIdentifier());
+            $logMessage = sprintf('Successfully re-authenticated tokens for account "%s"', (string) $authenticationManager->getSecurityContext()->getAccount()->getAccountIdentifier());
```
bwaidelich: nitpick: IIRC our CGL don't add a whitespace between casts and the value
```suggestion
            $logMessage = sprintf('Successfully re-authenticated tokens for account "%s"', (string)$authenticationManager->getSecurityContext()->getAccount()->getAccountIdentifier());
```

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderName.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Neos\Flow\Security\Authentication;
+
+final class AuthenticationProviderName
+{
+
+    /**
+     * @var string
+     */
+    protected $authenticationProviderName;
```
bwaidelich:
```suggestion
    private $authenticationProviderName;
```

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderName.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    public function __construct(string $authenticationProviderName)
+    {
+        $this->authenticationProviderName = $authenticationProviderName;
+    }
+
+    /**
+     * @return string
+     */
+    public function __toString()
```
bwaidelich:
```suggestion
    public function __toString(): string
```

Neos.Flow/Classes/Security/Authentication/CredentialsSource.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Neos\Flow\Security\Authentication;
+
+final class CredentialsSource
+{
+
+    /**
+     * @var string
+     */
+    protected $credentialsSource;
```
bwaidelich:
```suggestion
    private $credentialsSource;
```

Neos.Flow/Classes/Security/Authentication/CredentialsSource.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    public function __construct(string $credentialsSource)
+    {
+        $this->credentialsSource = $credentialsSource;
+    }
+
+    /**
+     * @return string
+     */
+    public function __toString()
```
bwaidelich:
```suggestion
    public function __toString(): string
```

Neos.Flow/Classes/Security/AccountIdentifier.php
bwaidelich: (most of the comments apply to the other new value objects, too)

Neos.Flow/Classes/Security/AccountIdentifier.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    protected $identifier;
+
+    /**
+     * @param string $identifier
+     */
+    public function __construct(string $identifier)
+    {
+        $this->identifier = $identifier;
+    }
```
bwaidelich: I would suggest to keep the constructor private and add a
```
public static fromString(string $identifier): self
{
    return new static($identifier);
}
```
That will allow us to add more constructors lateron

Neos.Flow/Classes/Security/AccountIdentifier.php
daniellienert: Please copy the license header to the new files.

daniellienert:
Hey @sorenmalling, 
IMHO that looks much better now. 
I added some more nitpicks. 

It would further be great if you would squash your commits into one with a good description what have been changed.

And the, we need a PR in the Neos Repo also, changing the usages of the new methods.

Neos.Flow/Classes/Security/AccountIdentifier.php
sorenmalling: The "RoleCollection"/"AccountRoles" is one of the not done parts. I might not completely agree that it's a requirement of a Account. But I see pros/cons for both. But in terms of that - should we introduce a simple Collection class to hold the Roles? To avoud returning a array

sorenmalling:
Thanks for the feedback - new versions coming up this evening :-)

Neos.Flow/Classes/Security/AccountIdentifier.php
bwaidelich: Right, I only looked at the "TODO list" afterwards :blush:
IMO roles absolutely have to be part of the interface or we can't replace the `Account` in the security framework (everything is role based there).
And, yes, it should be value objects too, along the lines of https://gist.github.com/bwaidelich/8266c980ab8c2db817bd4425a817390a (untested)

(And, no, it didn't want to imply that _you_ have to take care of this all *g)

sorenmalling: Arh, good point about the security framework (which might be splitted at one point #evilplan ;) )

Neos.Flow/Classes/Security/AccountIdentifier.php
sorenmalling: But only for this class? And not for the others (AuthenticationPRoviderName, CredentialsSource)?

sorenmalling:
@bwaidelich I get to a point, where my Unit testing knowledge and Securirty framework is "not there yet".
A number of tests fails, due to change from a array of Role objects to the new Roles collection class.

Can you help me adjust the tests to match the new format?

sorenmalling:
ping @bwaidelich 

This is my current output when running the tests

https://gist.github.com/sorenmalling/c3f599a4ebc0f2ee53058cea18c41f18

albe:
Sorry, had to force push after a messed up merge and push beforehand. Did a couple of fixes to the code that prevented tests from running. Still WIP, but getting there.

sorenmalling:
Trying to go over the failing tests, I can't wrap my head around the issue and how to solve it. 

Is this feature considered something that should go in to Flow 6, then I will fight for it (also for my own projects sake). 

If not, I will like to request some help after the Flow 6 release.

bwaidelich:
> Is this feature considered something that should go in to Flow 6 [...]

Depending on the potential "breakiness" I would love to have this in Flow 6.

Let me know if you need help with the tests

sorenmalling:
> > Is this feature considered something that should go in to Flow 6 [...]
> 
> Depending on the potential "breakiness" I would love to have this in Flow 6.

Well, a little bit breaky, since we add some collections classes instead of arrays. 

> Let me know if you need help with the tests

I do need help ;-) I havent been writing tests before and the concept of mocking etc is very new and never-touched ground.

albe:
IMO definitely something for 6.0, otherwise this needs to be postponed to 7.0

bwaidelich:
> I do need help ;-)

@sorenmalling Sorry, I missed your previous comments somehow.. Up for a quick chat/call tomorrow to go through the remaining tasks?

Neos.Flow/Classes/Security/AccountIdentifier.php
bwaidelich: Yes, your're right. lets do it for all new VOs

sorenmalling:
> > I do need help ;-)
> 
> @sorenmalling Sorry, I missed your previous comments somehow.. Up for a quick chat/call tomorrow to go through the remaining tasks?

@bwaidelich I will be offline till sunday, but will give it a try with fresh eyes, and ping you next week with lasting issues. I will love to get this in and planned time for it :) Thanks for helping me

bwaidelich:
@sorenmalling any news here? I'd be so nice to have this for 6.0 but we're running out of time I'm afraid

sorenmalling:
> @sorenmalling any news here? I'd be so nice to have this for 6.0 but we're running out of time I'm afraid

Working on this as we speak :) Just rebased latest master and try to adjust the tests and functionality

sorenmalling:
Seeing a high number of new failed tasks going all the way back to the DispatcherTest and functionaltytest. Failed task that I haven't seen in the former tests here and my git/fork is totally out of sync and I can't commit to any places :-/ Will try and rebuild my setup and see if the issues are local-only.

bwaidelich:
Great! I'm sure I can help with the rebase. Let's get in touch on Monday, if you like

albe:
I merged master into it to get rid of issues when merging later and also gain from the psalm checks. I guess it's not for the worst, as it shows a few type issues already:
https://travis-ci.org/neos/flow-development-collection/jobs/575994528#L741

But the important thing is the failing test. What happens is, that the authenticateSuccess test will persist/update the Account. During persistence, the Validation will be triggered and it builds a validator for the `roles` property, which it expects to be a collection type (annotated array, hence CollectionValidator). Then it calls the getter to retrieve the value, which is `Roles` and tries to validate that, where it fails - because not a Collection type.

So there's two things to notice:
- the validation on roles is triggered, because we do not build specific validation chains for MVC and persistence - they are the same and there is no escape hatch in validation for `@Flow\Transient` properties
- having a property annotated with a type different from the type the getter returns, will lead to weird behaviour like this

To fix this, we'd need to add hardcoded checks for validationGroup 'Persistence' and check against the class schema if the property is transient... for the moment, adding a `@Flow\IgnoreValidation` or naming the property different from the getter should suffice.

sorenmalling:
> Great! I'm sure I can help with the rebase. Let's get in touch on Monday, if you like

I don't know how you handle these things, but is this blocked untill #1734 is merged and we can run the test with that?

albe:
Let's rather fix the tests more pragmatic here - I'm undecided if renaming the property or adding an IgnoreValidation annotation is better. But the other BUGFIX might take much too long to be fully thought through to wait on it.

sorenmalling:
> Let's rather fix the tests more pragmatic here - I'm undecided if renaming the property or adding an IgnoreValidation annotation is better. But the other BUGFIX might take much too long to be fully thought through to wait on it.

My local environment is messed up and I can't push, but can you -  @albe - push a change with `@Flow\IgnoreValidation` on the `$roles` property? That passed my local unittest

albe:
Sorry, that problem is most likely due to my force push earlier :/ never force push to a collaborative branch. You need to checkout to a competely new branch. If you have commits on your old local copy that are not here, you then need to cherry-pick them.

kdambekalns:
@sorenmalling Are all open tasks done? The PR description checklist has at least one open point… :)

Neos.Flow/Classes/Security/Policy/Roles.php
```diff
@@ -0,0 +1,98 @@
...
+* information, please view the LICENSE file which was distributed with this
+* source code.
+*/
+
+/**
+ * A collection class containing all roles for a account
+ *
+ * @Flow\Proxy(false)
+ */
+final class Roles implements \JsonSerializable, \IteratorAggregate, \Countable, \ArrayAccess
```
bwaidelich: I think this class should be immutable - otherwise this could lead to a real mess.
I.e. throw in `offsetSet` and `offsetUnset` and add methods like
```php
public function withRole(Role $role): self

public function withoutRole(string $roleIdentifier): self
```

or similar

Neos.Flow/Classes/Session/Session.php
```diff
@@ -730,7 +730,7 @@ protected function storeAuthenticatedAccountsInfo(array $tokens)
         foreach ($tokens as $token) {
             $account = $token->getAccount();
             if ($token->isAuthenticated() && $account !== null) {
-                $accountProviderAndIdentifierPairs[$account->getAuthenticationProviderName() . ':' . $account->getAccountIdentifier()] = true;
+                $accountProviderAndIdentifierPairs[(string) $account->getAuthenticationProviderName() . ':' . (string) $account->getAccountIdentifier()] = true;
```
bwaidelich:
```suggestion
                $accountProviderAndIdentifierPairs[(string)$account->getAuthenticationProviderName() . ':' . (string)$account->getAccountIdentifier()] = true;
```

Neos.Flow/Classes/Security/Policy/Roles.php
```diff
@@ -0,0 +1,98 @@
...
+        });
+        return new static($processedRoles);
+    }
+
+    public function count(): int
+    {
+        return count($this->roles);
+    }
+
+    public function has(Role $role): bool
```
bwaidelich: `$account->getRoles()->hasRole(new Role('Some.Role'))` is weird
`$account->getRoles()->has('Some.Role')` would be better IMO

Neos.Flow/Tests/Functional/Security/AccountFactoryTest.php
```diff
@@ -41,8 +41,8 @@ public function createAccountWithPasswordCreatesANewAccountWithTheGivenIdentifie
 
         $actualAccount = $factory->createAccountWithPassword('username', 'password', ['Neos.Flow:Administrator', 'Neos.Flow:Customer'], 'OtherProvider');
 
-        self::assertEquals('username', $actualAccount->getAccountIdentifier());
-        self::assertEquals('OtherProvider', $actualAccount->getAuthenticationProviderName());
+        self::assertEquals('username', (string) $actualAccount->getAccountIdentifier());
```
bwaidelich:
```suggestion
        self::assertEquals('username', (string)$actualAccount->getAccountIdentifier());
```

Neos.Flow/Tests/Functional/Security/AccountFactoryTest.php
```diff
@@ -41,8 +41,8 @@ public function createAccountWithPasswordCreatesANewAccountWithTheGivenIdentifie
...
+        self::assertEquals('OtherProvider', (string) $actualAccount->getAuthenticationProviderName());
```
bwaidelich:
```suggestion
        self::assertEquals('OtherProvider', (string)$actualAccount->getAuthenticationProviderName());
```

daniellienert:
Also left some nitpicks. Maybe you could give us access to alter the PR and fix them directly.

Neos.Flow/Classes/Security/Policy/Roles.php
```diff
@@ -0,0 +1,98 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Flow\Security\Policy;
+
+use Neos\Flow\Annotations as Flow;
+
+/*
+* This file is part of the Neos.FluidAdaptor package.
```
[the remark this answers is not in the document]
albe: Missed that one...

Neos.Flow/Classes/Security/Policy/Roles.php
albe: There would be `isset($account->getRoles()['Some.Role'])` but that also feels weird. Changed the method to accept both `Role` and string.

Neos.Flow/Classes/Security/Policy/Roles.php
albe: Changed like that, though I had `withoutRole(Role $role)`, but yeah, maybe makes sense to filter out roles with the plain role string too... though it also feels counter to the idea of having a VO in the first place. The use case **should be** s.th. like:

```
$editorRole = new Role('User.Editor'); // or have some static constructor even `MyRoles::Editor()`
...
$this->roles = $this->roles->withoutRole($editorRole);
```

**should**

sorenmalling:
Sorry for being away while you have been busy bees on this one. I hope to look into it tomorrow. A system crash sundays had put me behind on this weeks schedule

albe:
Won't be able to put more time into this for now, sorry! So if this is still supposed to go into 6.0, someone would need to finish this up soon. We should at least get a couple days of playing around with this post-merge

sorenmalling:
I’ve pushed what should be the last changes for psalm and checked the correct usage of AccountInterface. Checked of the last tasks

Neos.Flow/Classes/Security/Policy/Roles.php
sorenmalling: This is done in the Roles VO as I can see

Neos.Flow/Classes/Security/AccountIdentifier.php
sorenmalling: Fixed in 85614b7

albe:
@sorenmalling can you merge the current 6.0 branch into this and push? I think that should resolve most (if not all) of the psalm issues - see #1775

sorenmalling:
> @sorenmalling can you merge the current 6.0 branch into this and push? I think that should resolve most (if not all) of the psalm issues - see #1775

I will try and do that right away :)

sorenmalling:
From the errors in travis it seems that the commit pointers are not uptodate yet - will wait and request a retry

albe:
Could be due to the target branch switch... ugh... travis 🤦‍♂

sorenmalling:
> Could be due to the target branch switch... ugh... travis 🤦‍♂
> 
> Edit: if this can't be resolved soon, feel free to just create a new PR from that branch. Just be sure to target the `6.0` branch then

How do I target 6.0 - after selecting it I only get a link to this pull requests

image

albe:
Maybe you need to close this one first? If that doesn't help still, push the branch to a new name on your fork

sorenmalling:
> Maybe you need to close this one first? If that doesn't help still, push the branch to a new name on your fork

Opened a new PR but travis breaks instantly
https://travis-ci.org/neos/flow-development-collection/jobs/586601540

albe:
That's the travis job from this PR ;)

sorenmalling:
Closing this PR due to branch issues. Work continues in #1776

PR #1541, opened 2019-04-09
FEATURE: Add release notes

gerhard-boden:


gerhard-boden:
I noticed that some small adaptations are still necessary, but we can handle them in the branch directly I guess before releasing

PR #1549, opened 2019-04-15
BUGFIX: Fix package:create and derived commands when private packagist is used

mficzel:
When private packagist is used the following setting isn added to the repositories section of the composer.json:

```
repositories: [
    {
        "packagist.org": false
    }
]
```

This caused an error because the package:create command tried to access the undefined `type` property of each defined repository.

This change simply checks for the existence of the type key before acessing it.

#fixes https://github.com/neos/neos-development-collection/issues/2448


radmiraal:
I just ran into this issue in Flow 6.1.4 and it seems like related merged fix has been overwritten 2 months after merge in commit #1eb95525a63420e1a35f103c943d36de5a515b75 ?

albe:
D'uh, that's apparently an `MergeConflictManuallyResolvedError`. Gonna fix it once again. Thanks for the heads-up!

PR #1550, opened 2019-04-19
TASK: Better naming for include and exclude paths/patterns

albe:
Get rid of wording "blacklist"/"whitelist" because there's better terms.
Should have been named like this from the start. I'm to blame.


albe:
Damn travis hickups. Restarted the two failed jobs. Hopefully green soon.

Issue #658, opened 2016-10-07
Rework HTTP implementation


bwaidelich:
The current HTTP stack is quite error prone and often leads to nasty bugs that are difficult to dissect.

Some problems I can think of atm:
### No difference between incoming and outgoing Request

We use the same `Http\Request` to create _outgoing_  requests (i.e. when interacting with the `Http\Clien\Browser`) as for processing _incoming_ requests.
This leads to a convoluted API
#### Request::createFromEnvironment()

The [official documentation suggests](http://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Http.html#creating-an-actionrequest) to use `Request::createFromEnvironment()` to fetch the current request, but this will obviously create a completely new instance which is not what we want in this case.
Note: #656 addresses this issue
#### Bootstrap::getActiveRequestHandler

That's the method that the Flow core uses in several parts (`Logger`, `FileSystemTarget`, `Session`, `InternalRequestEngine`):

``` php
$bootstrap = Bootstrap::$staticObjectManager->get(\TYPO3\Flow\Core\Bootstrap::class);
/* @var Bootstrap $bootstrap */
$requestHandler = $bootstrap->getActiveRequestHandler();
if ($requestHandler instanceof HttpRequestHandlerInterface) {
    $request = $requestHandler->getHttpRequest();
    // ...
}
```

...eeehk
#### HttpRequestHandlerInterface::getRequest()

With 3.3+ it's possible to inject the `HttpRequestHandlerInterface` directly and fetch the current request instance.
Still not a very straight-forward way that still has some potential side-effects

bwaidelich:
## Suggestions

### Incorporate PSR-7

I can't really judge how much work this would mean and if we can do it in a way that won't break everything, but we should consider adhering to the [PSR-7](http://www.php-fig.org/psr/psr-7/) interfaces (and maybe even use one of the existing packages for the request/response implementations)

### Replace RequestHandler with HTTP Component

IMO a good solution for interacting with the "current" request/response would be to make the `Http\ComponentContext`* available as early as possible and allow it to be injected.
The request handling could still be done in the already existing `DispatchComponent` but IMO we don't need the whole `RequestHandler` resolving part.. Using a custom component is easier (see Fluid `AjaxWidgetComponent` for example)

* To be discussed: find a better name that communicates that this is the global HTTP context

albe:
> With 3.3+ it's possible to inject the HttpRequestHandlerInterface directly

[...]

PR #1552, opened 2019-04-19
!!!TASK: Remove deprecated Http objects and replace with PSR-7 implementation

albe:
This replaces the HTTP stack of Flow with PSR-7.

Many areas of Flow are affected by this, most notably and breaking:

* All HTTP is now fully PSR-7
* Response in MVC controllers is no longer an HTTP response and has very different methods.
* CLI and MVC use different dispatchers now
* `ActionRequest::getParentRequest()` will return null at the top instead of an HttpRequest, you can still get the HttpRequest via `ActionRequest::getHttpRequest()`
* `ActionRequest::fromHttpRequest(ServerRequestInterface $httpRequest)` introduced
* `ActionRequest::createSubRequest()` introduced
* ActionRequest can longer be created via `new`
* `\Neos\Flow\Mvc\ActionRequestFactory` introduced to correctly merge arguments from the HTTP request
* `Neos.Http.Factories` introduced, implementing PSR-17 HTTP factories, use those to create and fake HTTP requests
* The HTTP process was split into more components to have easier extension points in between. So you can interject between the creation of the top level ActionRequest (after which security is avaliable) and the actual dispatching to a controller

Related: #658

Example API Changes:
- no more `Mvc\Response`, instead `ActionRequest` and `ActionResponse` are the API inside the MVC stack
- to create an `ActionRequest`, use the `ActionRequestFactory->createActionRequest($serverRequest, $arguments)`
- inside a Controller:
    - `$this->response->setHeader('Content-Type', ...)` -> `$this->response->setContentType(...)`
    - `$this->response->setHeader('Location', ...)` -> `$this->response->setRedirectUri(...)`
    - `$this->response->setStatus(...)` -> `$this->response->setStatusCode(...)`
    - `$this->response->setHeader(...)` -> `$this->response->setComponentParameter(SetHeaderComponent::class, ...)`
- `Request::create(...)` -> `new ServerRequest('GET', ...)`
- `$httpRequest->getBaseUri()` -> `$httpRequest->getAttribute(ServerRequestAttributes::BASE_URI)`


Neos.Flow/Classes/Http/Component/ComponentContext.php
```diff
@@ -12,7 +12,7 @@
  */
 
 use Neos\Flow\Annotations as Flow;
-use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ServerRequestInterface;
```
albe: Need an `ServerRequestInterface` here because a couple of places depend on the request from the Context providing `getAttribute(...)` in order to retrieve e.g. the baseUri

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
```diff
@@ -58,6 +58,13 @@ public static function getScriptRequestPath(ServerRequestInterface $request): st
         return implode('/', $requestPathSegments) . '/';
     }
 
+    public static function getRelativeRequestPath(ServerRequestInterface $request): string
```
albe: Ignore missing docblock.

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
```diff
@@ -110,6 +117,13 @@ public static function renderRequestHeaders(RequestInterface $request): string
         return $renderedHeaders;
     }
 
+    public static function renderRequest(RequestInterface $request): string
```
albe: See above.

Neos.Flow/Classes/Http/Request.php
```diff
@@ -95,6 +97,11 @@ class Request extends BaseRequest implements ServerRequestInterface
      */
     const ATTRIBUTE_BASE_URI = 'baseUri';
 
+    /**
+     * PSR-7 Attribute containing the Flow unified arguments (get, post, files) for this request.
+     */
+    const ATTRIBUTE_UNIFIED_ARGUMENTS = 'unifiedArguments';
```
albe: The unified arguments are a core feature of Flows HTTP Stack and we need to make them somehow available. I don't even know, if there's any other way to access the POST params in PSR-7 Request.

Neos.Flow/Classes/Http/Request.php
```diff
@@ -140,7 +147,72 @@ public function __construct(array $get, array $post, array $files, array $server
...
+     *
+     * @param array $get Data similar to that which is typically provided by $_GET
+     * @param array $post Data similar to that which is typically provided by $_POST
+     * @param array $files Data similar to that which is typically provided by $_FILES
+     * @param array $server Data similar to that which is typically provided by $_SERVER
+     * @return ServerRequestInterface
+     * @see create()
+     * @see createFromEnvironment()
+     */
+    private static function _create(array $get, array $post, array $files, array $server): ServerRequestInterface
```
albe: For now solved the main issue (`Request::create[FromEnvironment]()` all around the places) by just creating a factory method that returns an `ServerRequestInterface` instance. The other solution would be, to make our `Http\Request` decorate on an actual exchangeable `ServerRequestInterface` (and provide some convenience helpers on top). I mean, the main reason for us to use PSR-7 (and PSR-15) is to be able to make use of existing middlewares and keep the HTTP Kernel implementation outside of Flow, right? But that is still guaranteed, even if inside Flow we work with an Http Request class, that provides a bit more than plain PSR-7 (e.g. `getBaseUri()`, `getArguments()`, Headers and Cookies objects that provide some useful IDE help). WDYT?

Neos.Flow/Classes/Http/Request.php
```diff
@@ -193,29 +265,28 @@ public static function create(Uri $uri, $method = 'GET', array $arguments = [],
...
      *
-     * @return Request
+     * @return ServerRequestInterface
      * @api
      */
-    public static function createFromEnvironment()
+    public static function createFromEnvironment(): ServerRequestInterface
     {
-        $request = new static($_GET, $_POST, $_FILES, $_SERVER);
-        $request->setContent(null);
```
albe: Why did we initialize Content to null explicitly?

Neos.Flow/Classes/Http/RequestHandler.php
```diff
@@ -90,15 +95,15 @@ public function getPriority()
...
     {
-        // Create the request very early so the ResourceManagement has a chance to grab it:
-        $request = Request::createFromEnvironment();
-        $response = new Response();
-        $this->componentContext = new ComponentContext($request, $response);
-
         $this->boot();
         $this->resolveDependencies();
+
+        $request = Request::createFromEnvironment();
```
albe: Needed to be moved after `resolveDependencies()` and initialization of ObjectManagement. However, I don't really see any issue with that, since no place in the bootstrap does (and should) need an Http Context with Request and/or Response object.

Neos.Flow/Classes/Http/RequestHandler.php
```diff
@@ -90,15 +95,15 @@ public function getPriority()
...
+        $response = $this->responseFactory->createResponse();
         $response = $this->addPoweredByHeader($response);
-        $this->componentContext->replaceHttpResponse($response);
+        $this->componentContext = new ComponentContext($request, $response);
+        $this->bootstrap->getObjectManager()->setInstance(ComponentContext::class, $this->componentContext);
```
albe: Sneaky attempt to make `ComponentContext` injectable.

Neos.Flow/Classes/Mvc/DispatchComponent.php
```diff
@@ -95,7 +97,7 @@ public function handle(ComponentContext $componentContext)
     protected function prepareActionRequest(ComponentContext $componentContext)
     {
         $httpRequest = $componentContext->getHttpRequest();
-        $arguments = $httpRequest->getArguments();
+        $arguments = $httpRequest->getQueryParams();
```
albe: This should be `getAttribute(Http\Request::ATTRIBUTE_UNIFIED_ARGUMENTS)`

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -250,6 +250,21 @@ Neos\Flow\Monitor\FileMonitor:
 # HTTP                                                                     #
 #                                                                          #
 
+Psr\Http\Message\ServerRequestFactoryInterface:
+  className: Tuupola\Http\Factory\ServerRequestFactory
```
albe: We could just provide our own implementations to be independent. The code is not too much.

Neos.FluidAdaptor/Classes/Core/Widget/AjaxWidgetComponent.php
```diff
@@ -75,15 +76,15 @@ public function handle(ComponentContext $componentContext)
...
      * @return WidgetContext
      */
-    protected function extractWidgetContext(Request $httpRequest)
+    protected function extractWidgetContext(ServerRequestInterface $httpRequest)
     {
-        if ($httpRequest->hasArgument('__widgetId')) {
-            return $this->ajaxWidgetContextHolder->get($httpRequest->getArgument('__widgetId'));
-        } elseif ($httpRequest->hasArgument('__widgetContext')) {
-            $serializedWidgetContextWithHmac = $httpRequest->getArgument('__widgetContext');
+        if (isset($httpRequest->getQueryParams()['__widgetId'])) {
```
albe: `QueryParams` should be fine instead of the unified arguments, or is there a use-case to submit the `__widget*` arguments via POST params?

Neos.FluidAdaptor/Classes/Core/Widget/AjaxWidgetComponent.php
```diff
@@ -75,15 +76,15 @@ public function handle(ComponentContext $componentContext)
...
+            return $this->ajaxWidgetContextHolder->get($httpRequest->ggetQueryParams()['__widgetId']);
```
albe: Typo.

kitsunet:
OK, I totally missed this change and basically started the same in parallel. I will try to rebase my work on top of this, as I have a couple of changes that are not in here yet. We might want to discuss the Factories. I currently do not use them.

Neos.FluidAdaptor/Classes/Core/Widget/AjaxWidgetComponent.php
kitsunet: you never know, but I can imagine that might happen...

Neos.Flow/Configuration/Objects.yaml
kitsunet: Yep, need to look into it. I haven't used factories in my change so far, but probably makes sense.

Neos.Flow/Classes/Mvc/DispatchComponent.php
kitsunet: I am currently not fully convinced this is a good idea, I created a little helper and just merge the arguemnts wherever I needed it. TBD...

Neos.Flow/Classes/Http/RequestHandler.php
kitsunet: yeah, no, please. I think we need to think very careful how we want to provide this, but not that way.

Neos.Flow/Classes/Http/RequestHandler.php
kitsunet: In my change I totally got rid of Request and I would rather prefer that. This whole transition will be breaking, so lets at least use it to clean up as well.

Neos.Flow/Classes/Http/Request.php
kitsunet: See below, I made it a helper that is used "as needed" to avoid having all the arguments laying around two times.

Neos.Flow/Classes/Mvc/DispatchComponent.php
albe: That's possible too. Wanted to avoid throwing that Helper around, but it's correct that it's probably less error prone, since it works with changed arguments somewhere inside the http component chain/middleware stack.

Neos.FluidAdaptor/Classes/Core/Widget/AjaxWidgetComponent.php
albe: Yep yep. So we need the ArgumentsHelper mumbo-jumbo here. Or maybe we can decouple this Fluid Widget stuff from Http and only depend on ActionRequest?

albe:
Thanks for the feedback! Yes, let's discuss things. Just hit me on slack sometime this week.

This is a pretty early status and I only got the first tests pass again as easily as possible, by jumping on the Request::create* stuff to avoid rewriting all those tests. This is not how it should be ofc. If you manage to rebase your work on this or merge them together, that would be awesome!

kitsunet:
There we go, tests will not run fully, but should be a step in the right direction. One thing to discuss is default SERVER vars as they know get applied with our request. That breaks a bunch of tests at the moment and I am not 100% sure how to fix this. Potentially could be fixed in the ServerRequestFactory but \o/ not sure

Neos.Flow/Classes/Http/Component/BaseUriComponent.php
```diff
@@ -0,0 +1,51 @@
...
+use Neos\Flow\Http\HttpRequestHandlerInterface;
+use Neos\Flow\ObjectManagement\ObjectManagerInterface;
+
+/**
+ * Defines the baseUri attribute for the current request.
+ *
+ * This is done after the TrustedProxiesComponent to make sure
+ * the proxy headers are evaluated correctly.
+ */
+class BaseUriComponent implements ComponentInterface
```
albe: Like

Neos.Flow/Classes/Http/Client/Browser.php
```diff
@@ -138,11 +140,14 @@ public function request($uri, $method = 'GET', array $arguments = [], array $fil
             throw new \InvalidArgumentException('$uri must be a URI object or a valid string representation of a URI.', 1333443624);
         }
 
-        $request = Request::create($uri, $method, $arguments, $files, $server);
-
-        if ($content !== null) {
-            $request->setContent($content);
+        $request = new ServerRequest($method, $uri, [], $content, '1.1', $server);
```
albe: Having a factory method for "our" needs is preferable I think. Doesn't need to be the old `Request` class, but I don't see a bunch of reasons why not either.

Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
```diff
@@ -46,9 +46,9 @@ public function handle(ComponentContext $componentContext)
     {
         $request = $componentContext->getHttpRequest();
 
-        $trustedRequest = $request->withAttribute(Request::ATTRIBUTE_TRUSTED_PROXY, $this->isFromTrustedProxy($request));
+        $trustedRequest = $request->withAttribute(HttpRequestHandlerInterface::ATTRIBUTE_TRUSTED_PROXY, $this->isFromTrustedProxy($request));
```
albe: Interesting approach on where to have the request attribute constants.

Neos.Flow/Classes/Http/Request.php
```diff
@@ -115,6 +115,7 @@ class Request extends BaseRequest implements ServerRequestInterface
      */
     public function __construct(array $get, array $post, array $files, array $server)
     {
+        die();
```
albe: Haha 👿

Neos.Flow/Classes/Mvc/DispatchComponent.php
```diff
@@ -105,15 +106,16 @@ protected function prepareActionRequest(ComponentContext $componentContext)
             $httpRequest = $httpRequest->withParsedBody($parsedBody);
         }
 
+        $arguments = ArgumentsHelper::buildUnifiedArguments($httpRequest->getQueryParams(), $parsedBody, UploadedFilesHelper::untangleFilesArray($_FILES));
```
albe: Not sure it's good to refer to $_FILES here again. Can we extract that info back from PSR7 UploadedFiles? Otherwise maybe store it in an Request attribute?

Neos.Flow/Classes/Mvc/Routing/RouterCachingService.php
```diff
@@ -116,6 +117,8 @@ public function storeMatchResults(RouteContext $routeContext, array $matchResult
         if ($this->containsObject($matchResults)) {
             return;
         }
+
+        $httpRequest = $routeContext->getHttpRequest();
```
albe: You wanted to do something with the http request here?

Neos.FluidAdaptor/Tests/Functional/View/StandaloneViewTest.php
```diff
@@ -48,7 +48,7 @@ public function runBare(): void
      */
     public function inlineTemplateIsEvaluatedCorrectly()
     {
-        $httpRequest = Request::create(new Uri('http://localhost'));
+        $httpRequest = new Request('GET', new Uri('http://localhost'));
```
albe: ServerRequest then

Neos.Flow/Classes/Http/Client/Browser.php
kitsunet: Sure!

Neos.Flow/Classes/Mvc/Routing/RouterCachingService.php
kitsunet: Merge error, I had originally used it in the next line but you extract it directly, so will be removed.

Neos.Flow/Classes/Mvc/DispatchComponent.php
```diff
@@ -75,90 +37,22 @@ public function __construct(array $options = [])
...
     public function handle(ComponentContext $componentContext)
     {
-        $componentContext = $this->prepareActionRequest($componentContext);
         $actionRequest = $componentContext->getParameter(DispatchComponent::class, 'actionRequest');
-        $this->setDefaultControllerAndActionNameIfNoneSpecified($actionRequest);
         $actionResponse = $this->prepareActionResponse($componentContext->getHttpResponse());
-        $this->dispatcher->dispatch($actionRequest, $actionResponse);
+        try {
+            $this->dispatcher->dispatch($actionRequest, $actionResponse);
+        } catch (AuthenticationRequiredException $exception) {
```
kitsunet: After rethinking, given that the ActionResponse can carry component context parameters, I guess we should move this catch back to the dispatcher and set the respective parameter there.

We just first need to address the merging of ActionResponse to http response and component context. which is still a TODO.

Neos.Flow/Classes/Http/Component/ComponentInterface.php
```diff
@@ -17,7 +17,7 @@
  * An HTTP component
  *
  * A component is one item of the configurable component chain that is processed for every incoming request. A component can change the current HTTP request and response,
- * can communicate with other components and even change the currently processed chain using the ComponentContext that gets passed to its handle() method.
+ * can communicate with other components and even change the currently processed chain using the IntoComponentContext that gets passed to its handle() method.
```
kitsunet: That is obviously a bad fix by PHPStorm, will fix

Neos.Flow/Classes/Http/Component/RequestBodyParsingComponent.php
```diff
@@ -32,6 +32,9 @@ class RequestBodyParsingComponent implements ComponentInterface
     public function handle(ComponentContext $componentContext)
     {
         $httpRequest = $componentContext->getHttpRequest();
+        if (!empty($httpRequest->getParsedBody())) {
```
kitsunet: I have to check how we handled this before, but Guzzle ServerRequset::fromGlobals() sets $_POST as parsedBody, so if it was a "normal" POST request the parsed body will be fine already at this point.

Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
```diff
@@ -26,7 +26,7 @@
  *   ->withHostPrefix('de.', ['en.', 'ch.'])
  *   ->withPort(8080)
  *   ->withPathPrefix('prefix/');
- * $resultingUri = $uriConstraints->applyTo($exampleUri); // https://de.some-domain.tld:8080/prefix/foo
+ * $resultingUri = $uriConstraints->prepareRendering($exampleUri); // https://de.some-domain.tld:8080/prefix/foo
```
kitsunet: Again bad PhpStorm fix, didn't see that, will fix.

Neos.Flow/Classes/Mvc/Routing/RoutingComponent.php
```diff
@@ -53,8 +53,8 @@ public function __construct(array $options = [])
     /**
      * Resolve a route for the request
      *
-     * Stores the resolved route values in the ComponentContext to pass them
-     * to other components. They can be accessed via ComponentContext::getParameter(outingComponent::class, 'matchResults');
+     * Stores the resolved route values in the IntoComponentContext to pass them
```
kitsunet: and another PhpStorm badness...

Neos.Flow/Tests/FunctionalTestRequestHandler.php
```diff
@@ -125,7 +125,7 @@ public function setComponentContext(ComponentContext $context)
     }
 
     /**
-     * Internal getter to ensure an existing ComponentContext.
+     * Internal getter to ensure an existing IntoComponentContext.
```
kitsunet: ...

Neos.FluidAdaptor/Classes/Core/Widget/AbstractWidgetViewHelper.php
```diff
@@ -238,19 +239,15 @@ protected function initiateSubRequest()
 
                 /** @var $parentResponse ActionResponse */
                 $parentResponse = $this->controllerContext->getResponse();
```
albe: The "parentResponse" can be removed from the `ActionResponse` constructor above in Ln 216.

Neos.FluidAdaptor/Classes/Core/Widget/AbstractWidgetViewHelper.php
```diff
@@ -238,19 +239,15 @@ protected function initiateSubRequest()
...
-                }
+                $intoParentResponse = new IntoActionResponse($parentResponse);
+                $parentResponse = $subResponse->prepareRendering($intoParentResponse)->render();
 
                 throw $exception;
             }
         }
+
+        // At this point the sub reponse is already merged to the parent response,
+        // so it's slightly odd to return it here, but for now that's what we do.
```
albe: See above, we should return `$subResponse->getContent()` here, or rather through the `RendererInterface` API `return $subResponse->prepareRender(new Content())->render();`

Neos.FluidAdaptor/Classes/ViewHelpers/Widget/AutocompleteViewHelper.php
```diff
@@ -51,6 +52,7 @@ class AutocompleteViewHelper extends AbstractWidgetViewHelper
      */
     public function render(QueryResultInterface $objects, $for, $searchProperty, array $configuration = ['limit' => 10])
     {
-        return $this->initiateSubRequest();
+        $response = $this->initiateSubRequest();
```
albe: This should be just `return $this->initiateSubRequest();` The decision that only the content of the SubResponse is rendered is not responsibility of a concrete ViewHelper.

Neos.FluidAdaptor/Classes/ViewHelpers/Widget/PaginateViewHelper.php
```diff
@@ -63,6 +64,6 @@ class PaginateViewHelper extends AbstractWidgetViewHelper
     public function render(QueryResultInterface $objects, $as, array $configuration = ['itemsPerPage' => 10, 'insertAbove' => false, 'insertBelow' => true, 'maximumNumberOfLinks' => 99])
     {
         $response = $this->initiateSubRequest();
-        return $response->getContent();
+        return $response->prepareRendering(new Content())->render();
```
albe: See above.

Neos.FluidAdaptor/Tests/Functional/Core/Fixtures/ViewHelpers/RedirectViewHelper.php
```diff
@@ -33,6 +34,7 @@ class RedirectViewHelper extends AbstractWidgetViewHelper
      */
     public function render()
     {
-        return $this->initiateSubRequest();
+        $response = $this->initiateSubRequest();
```
albe: Ouch, we actually have a test that verifies that a widget can cause a redirect... this ofc changes things regarding my feedback on the AbstractWidgetViewHelper in case this is really something we DO WANT.

kitsunet: I still don't like it, probably something to discuss at the sprint, I guess I will leave the stuff as it is now and take care of the other open points more close to the intention of the PR, and we can revisit this. I would prefer to minimise the API for sub requests and really think through what we want to do with them. It just comes to my mind that backend modules are also using sub requests in a quite weird way, might be that they actually need way more than content. Anyway, this kind of works for now and at least makes the transition from main action dispatch to HTTP relatively clear.

Neos.FluidAdaptor/Classes/Core/Widget/AbstractWidgetViewHelper.php
kitsunet: Yep, yep, I kind of agree, see longer comment below.

Neos.Flow/Classes/Mvc/ActionResponseRenderer/IntoComponentContext.php
```diff
@@ -0,0 +1,57 @@
...
+            $httpResponse = $httpResponse->withHeader('Content-Type', $this->contentType);
+        }
+
+        if ($this->redirectUri) {
+            $httpResponse = $httpResponse->withHeader('Location', $this->redirectUri);
+        }
+
+        foreach ($this->componentParameters as $componentParameterGroup) {
+            foreach ($componentParameterGroup as $parameterName => $parameterValue) {
+                $this->componentContext->setParameter($componentParameterGroup, $parameterName, $parameterValue);
```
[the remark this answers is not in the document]
kitsunet: facepalm, ofc

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -104,4 +111,19 @@ public function setComponentParameter(string $componentClassName, string $parame
...
         $this->componentParameters[$componentClassName][$parameterName] = $value;
     }
+
+    /**
+     * @param ActionReponseRendererInterface $renderer
+     * @return ActionReponseRendererInterface
+     */
+    public function prepareRendering(ActionReponseRendererInterface $renderer): ActionReponseRendererInterface
+    {
+        $renderer->setContent($this->content);
```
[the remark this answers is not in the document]
kitsunet: True, I guess why not, there is no sensible use case for re using the renderer but I will have a look.

Neos.Flow/Classes/Mvc/ActionRequestFromHttpTrait.php
```diff
@@ -0,0 +1,53 @@
...
+     * @return ActionRequest
+     * @throws Exception\InvalidArgumentNameException
+     * @throws Exception\InvalidArgumentTypeException
+     */
+    protected function createActionRequest(ServerRequestInterface $httpRequest, array $additionalArguments = [])
+    {
+        // TODO: Move arguments merging into action request as lazy operation to avoid duplicating the arguments?
+        $arguments = ArgumentsHelper::buildUnifiedArguments($httpRequest->getQueryParams(), $httpRequest->getParsedBody(), UploadedFilesHelper::untangleFilesArray($_FILES));
+
+        /** @var $actionRequest ActionRequest */
```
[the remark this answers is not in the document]
kitsunet: probably not

Neos.Flow/Classes/Mvc/ActionRequestFromHttpTrait.php
```diff
@@ -0,0 +1,53 @@
...
+     * @param ServerRequestInterface $httpRequest
+     * @param array $additionalArguments
+     * @return ActionRequest
+     * @throws Exception\InvalidArgumentNameException
+     * @throws Exception\InvalidArgumentTypeException
+     */
+    protected function createActionRequest(ServerRequestInterface $httpRequest, array $additionalArguments = [])
+    {
+        // TODO: Move arguments merging into action request as lazy operation to avoid duplicating the arguments?
+        $arguments = ArgumentsHelper::buildUnifiedArguments($httpRequest->getQueryParams(), $httpRequest->getParsedBody(), UploadedFilesHelper::untangleFilesArray($_FILES));
```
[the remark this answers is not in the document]
kitsunet: yep, yep, see list at the top ;)

Neos.Flow/Classes/Mvc/ActionReponseRendererInterface.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function setStatusCode(int $statusCode): void;
+
+    /**
+     * @param array $componentParameters
+     */
+    public function setComponentParameters(array $componentParameters): void;
+
+    /**
+     * @return mixed
```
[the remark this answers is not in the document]
kitsunet: Agreed, but see implementation, I really don't see how we can do this differently given that we actually want to render in different ways.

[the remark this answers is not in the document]
albe: Yep. Don't have a good idea yet how to deal with this differently.

Neos.Flow/Classes/Mvc/Dispatcher.php
```diff
@@ -70,32 +95,19 @@ public function injectObjectManager(ObjectManagerInterface $objectManager)
...
-            $this->initiateDispatchLoop($request, $response);
-            return;
-        }
-
         // NOTE: The dispatcher is used for both Action- and CLI-Requests. For the latter case dispatching might happen during compile-time, that's why we can't inject the following dependencies
 
-        /** @var Context $securityContext */
-        $securityContext = $this->objectManager->get(Context::class);
-        if ($securityContext->areAuthorizationChecksDisabled()) {
+        if ($this->securityContext->areAuthorizationChecksDisabled()) {
```
albe: Ah, there are the small nice benefits of separating the two dispatchers. I think that shows this is the right direction after all!

kitsunet: Indeed!

kitsunet:
One thing annoying me atm are the firewall/request filter thingies because I feel they should be on HTTP level, which would work for everything but the ControllerObjectName. I am thinking if we want to write that to the ServerRequest as attribute...

Neos.Flow/Classes/Http/Component/RequestBodyParsingComponent.php
albe: Before we just merged $_POST into arguments on Request creation and kept the request body around until we parsed it when mapping arguments for the ActionController. Guzzle is more correct in assigning the POST as parsed Body, if it exists. It's just a PHP inbuilt request body parsing for `multipart/form-data`/`application/x-www-form-urlencoded` after all.

Neos.Flow/Classes/Http/Client/CurlEngine.php
```diff
@@ -137,12 +122,8 @@ public function sendRequest(Http\Request $request)
             curl_setopt($curlHandle, CURLOPT_PORT, $requestUri->getPort());
         }
 
-        if (count($request->getCookies()) > 0) {
-            $cookies = [];
-            foreach ($request->getCookies() as $cookie) {
-                $cookies[] = $cookie->getName() . '=' . $cookie->getValue();
-            }
-            curl_setopt($curlHandle, CURLOPT_COOKIE, implode('; ', $cookies));
+        if (count($request->getHeader("Cookie")) > 0) {
```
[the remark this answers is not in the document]
kitsunet: I think I would leave i like this, IMHO the request we sent out cannot/should not be a ServerRequest

kitsunet:
The argument merging contains some quirks that I fear might be difficult to work with if doing it lazy, so I refrained from that for now. I think we could implemented that at a later stage as well, so I would skip todo no 2 for now.

albe:
Slowly getting there...

kitsunet:
Oh no, I was fixing half the day^^ Will try to rebase...

kitsunet:
Important note for both of us:

check cases regarding special auth headers and stuff in
`\Neos\Flow\Tests\Unit\Security\Authentication\Token\UsernamePasswordHttpBasicTest`

We probably need to make sure that this happens in the RequestHandler as well (which currently with ServerRequest::fromGlobals() is not the case).

kitsunet:
Ready for review was not exactly what I wanted, actually I just wanted travis to do the tests...

kitsunet:
YAY, unit tests work now! Today we have retro day so I might not do much in regards to functional tests today, but we'll see...

kitsunet:
Let's see what travis says. I think this should be green tests now.

kitsunet:
Alright, this has a few test failures still that I couldn't figure yet, but I got rid of the main bulk of old code and overall things still work, so YAY

kitsunet:
Seems Guzzle Uri internally is quite a bit different from our old trust URI... I guess I need to do some digging on the details and somehow provide respective helpers to fix the behavior

kitsunet:
Tried to fix it for now, needs some discussion I guess, but should be fine.

kitsunet:
I guess we need to talk about this behavior change as well. The guzzle Uri implementation (correctly) encodes urls it stringifies. So you may need to decode them in some places. This seems good and correct behavior to me or do we want to decode IN the browser for example?

albe:
> The guzzle Uri implementation (correctly) encodes urls it stringifies.

In fact, guzzle doesn't do anything at all, but our own `parse_url` implementation inside the Utilities did do an urldecode for the components: https://github.com/neos/flow-development-collection/blob/master/Neos.Utility.Unicode/Classes/Functions.php#L199

https://tools.ietf.org/html/rfc3986#section-7.3 also states, that the single components must be decoded at some point when "dereferencing" the URI, so IMHO our implementation was more correct in fact. After all the encoding is just an artifact of the protocol used to transmit the URI (i.e. HTTP). Ofc there's room for an argument, where you might want to access the actual encoded source version of the URI, so maybe this would rather need a different getter method. PSR7 itself explicitly states that the accessors for the single URI components return %-encoded values, but the `__toString` method doesn't talk about that. IMO a missing thing in the PSR7 spec.

So, I'm not entirely sure. I mean, for the actual usage in most scenarios, you want to just work with the human readable/decoded version. So I see a few possiblities:

- Stick to our own (different) Psr7\Uri implenentation (maybe not best idea)
- Provide our own extension class around Guzzle\Uri (`... extends ...`) with an accessor for the decoded string (`asDecodedString()` or s.th.)- the PSR7 even explicitly states that additional functionality can be added on top of the interface (https://www.php-fig.org/psr/psr-7/#35-psrhttpmessageuriinterface)
- Since this is all HTTP-specific, but the primary use for the decoded URI is inside MVC, we could just use a different Uri class inside Mvc, that works mostly like the PSR-7 version, but with decoded values and use that inside ActionRequest/ActionResponse(?)
- stick with this (breaky) change and tell people to just use url_decode all around the place

Personally I'd prefer option 3 or 2.

Neos.Flow/Classes/Mvc/ActionRequest.php
```diff
@@ -47,9 +47,9 @@ class ActionRequest implements RequestInterface
 
     /**
      * Package key of the controller which is supposed to handle this request.
-     * @var string
+     * @var string|null
```
albe: Note: unfortunately, packageKey and controllerName need to stay nullable for now, because we depend on PHP's default method argument behaviour inside `Controller::forward` and `UriBuilder` (a provided `null` value falling back to the default value in the method declaration). On the other hand, we have a lot of places where we just directly use the return values without checking for null and operate on it as a string.

At least, now it's explicit in the type hints and doc blocks.

albe:
🎉 green again, now we need to decide on the `Uri` thing.

kitsunet:
Oh the CLI rename was quick. Given that it's not exactly the scope of this change I wanted to tackle that separately and ddiscuss a shim within the old namespace. Given that we never wanted to touch CLI in the first place I just wanted to make the necesary changes to keep it working. WDYT? @albe

kitsunet:
I would rather not want to extend the implementation as that voids the point of the interface. Suddenly we rely on a specific implementation (again). Also the tests I fixed were functional tests were the URI read from a rendered content part was wrong, IMHO in that case we are more correct now because it should be encoded in the markup there. I would maybe have a look at how other implementations handle it and decide based on that.

albe:
> Given that we never wanted to touch CLI in the first place I just wanted to make the necesary changes to keep it working.

True, but since a `Cli\Request` and `Cli\Response` class never existed before, I think it makes sense to just introduce them with the name we want it to have finally. We can still put a shim in place, though I'm not entirely sure how that work and if, what that would improve.

We anyway need to think about a code migration plan.

> I would maybe have a look at how other implementations handle it and decide based on that.

👍

kitsunet:
> True, but since a Cli\Request and Cli\Response class never existed before

Huh? That's not true, they existed since ever?

albe:
They did?? Oh wait, yes, they just extended Mvc\Request/Response, no?

kitsunet:
The Request jusut implemented the Mvc\RequestInterface and the Response indeed extended the Mvc\Response class

albe:
Fair enough. So yeah, would be cleaner to move that rename to an own PR. I'll take care this evening.

Neos.Flow/Tests/Unit/Mvc/ActionRequestTest.php
```diff
@@ -342,8 +342,8 @@ public function getControllerSubpackageKeyReturnsTheUnknownCasesPackageKeyIfNoCo
     public function invalidControllerNames()
     {
         return [
-            [42],
-            [false],
+            //[42],
```
albe: @kitsunet should we add an explicit test for the controller (and below action) name not being a "numeric" inside the setter to still pass these tests? I mean, a "42Controller" is an invalid class name anyway.

Neos.Flow/Classes/Mvc/ActionRequestFromHttpTrait.php
```diff
@@ -0,0 +1,63 @@
+<?php
+namespace Neos\Flow\Mvc;
+
+use Neos\Flow\Http\Helper\UploadedFilesHelper;
+use Neos\Utility\Arrays;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * FIXME: Make me a factory instead
```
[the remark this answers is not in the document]
kitsunet: Right, this is one point to iterate over before merging this I guess

Neos.Flow/Classes/Cli/Dispatcher.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * This signal is emitted directly before the request is been dispatched to a controller.
+     *
+     * @param Request $request
+     * @param Response $response
+     * @param ControllerInterface $controller
+     * @return void
+     */
+    protected function emitBeforeControllerInvocation(Request $request, Response $response, ControllerInterface $controller)
```
kitsunet: We also need to do a final iteration on these signals and decide how we want to proceed in the future. I think it woudl be nice if there was two distinct signals, one for any controller invocation and one specific to CLI and MVC.

Neos.Flow/Classes/Security/RequestPattern/ControllerObjectName.php
```diff
@@ -11,7 +11,7 @@
  * source code.
  */
 
-use Neos\Flow\Mvc\RequestInterface;
```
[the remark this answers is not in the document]
kitsunet: Nope wouldn't make sense, the whole security stack depends on ActionRequests atm... If it does at some point we need to rethink that anyway.

Neos.Flow/Classes/Package.php
```diff
@@ -56,8 +57,13 @@ public function boot(Core\Bootstrap $bootstrap)
         $bootstrap->registerCompiletimeCommand('neos.flow:package:rescan');
 
         $dispatcher = $bootstrap->getSignalSlotDispatcher();
+
         $dispatcher->connect(Mvc\Dispatcher::class, 'afterControllerInvocation', function ($request) use ($bootstrap) {
-            if ($bootstrap->getObjectManager()->has(Persistence\PersistenceManagerInterface::class)) {
+            // No auto-persistence if there is no PersistenceManager registered or during compile time
+            if (
+                $bootstrap->getObjectManager()->has(Persistence\PersistenceManagerInterface::class)
+                && !($bootstrap->getObjectManager() instanceof CompileTimeObjectManager)
```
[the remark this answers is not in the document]
kitsunet: Actually I think this was needed for the CLI dispatcher and compile time commands. No clue how it worked before, I assume some signal/slot quirks.

Neos.Flow/Classes/Mvc/ActionRequestFromHttpTrait.php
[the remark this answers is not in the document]
kitsunet: I have created a dedicated factory class from the trait for now. I feel the complexities handled in this are too big for a static factory method. I don't know. I think I like this. Might even become a factory at some point...

Neos.Flow/Classes/Cli/Dispatcher.php
kitsunet: Ok, so the dispatchers irk me a bit. I kind of want an interface for dispatchers now, but the handle method arguments are different and not reconcilable in a way (the reason why we have two dispatchers now). I am thinking, the dispatcher is semi internal anyways, it's not api, so I am thinking, what if we declared this interface (not super helpful but at least something):
    
    /**
     * @return mixed some response to the handled request, depending on the implementation
     */
    public function handle();

And then instead of making the dispatcher implementations singleton, just have a constructor which gets the request, response (maybe even optional) and possible dependencies. The constructor would not be part of the interface then but the interface can point to the fact that you need to create implementations with some input (request).

That would allow us to have a signal on this interface and request implementations to emit a signal for the interface additionally to any other signals it emits.

kitsunet:
Mmm, something has changed for the worse again (apart from the tests), because eg. redirecting to the authentication endpoint for Neos doesn't seem to work and URL encoding for the query arguments is wrong. The initial ? is encoded.

kitsunet:
Found the encoding problem.... in ´\Neos\Flow\Mvc\Routing\Dto\UriConstraints::applyTo` the CONSTRAINT_PATH does include a path AND the query string, which then (correctly) gets encoded on `withPath`... (ll 273).

Guess that needs refactoring.

Neos.Flow/Classes/Mvc/Routing/Route.php
```diff
@@ -534,10 +534,9 @@ public function resolves(array $routeValues)
                 $routeValues = $internalArguments;
             }
             $queryString = http_build_query($routeValues, null, '&');
-            if ($queryString !== '') {
-                $resolvedUriPath .= strpos($resolvedUriPath, '?') !== false ? '&' . $queryString : '?' . $queryString;
-            }
         }
+
+        $this->resolvedUriConstraints = $this->resolvedUriConstraints->withQueryString($queryString);
```
bwaidelich: Die Zeile (und damit das klonen der Constraints) wir ja nur benötigt, wenn `$queryString` nicht leer ist.
Darum würde ich vorschlagen, den Check hier drin zu lassen. Oder (vielleicht besser):
`UriConstraints::withQueryString()` so anzupassen, dass es die aktuelle Instanz zurück gibt, wenn der QueryString identisch mit dem aktuellen Wert ist.

kitsunet: Right, yes, I can leave that in, I figured it'S rather unnecessary because it defaults to an empty string in 525 so it would just be set to empty string then and teh overheat to do withQueryString seemed not too big. But the condition is no big deal. if you are ok with the addition of the extra condition

bwaidelich: My point was: Even if it's an empty string, the `UriConstraints` object will be cloned now.
But, sure, that's just a nitpick. The constraint in general makes a lot of sense to me (especially since `path` does _not_ include the query per (PSR) definition)

Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
```diff
@@ -164,6 +166,19 @@ public function withPath(string $path): self
...
     }
 
+    /**
+     * Create a new instance with the query string constraint added
+     *
+     * @param string $queryString
+     * @return UriConstraints
+     */
+    public function withQueryString(string $queryString): self
+    {
```
bwaidelich:
```suggestion
    {
        $currentQueryString = $this->constraints[self::CONSTRAINT_QUERY_STRING] ?? '';
        if ($queryString === $currentQueryString) {
            return $this;
        }
```
that's roughly what I had in mind btw

albe: It's a micro-optimization, but sure, why not?

albe:
This thing is green again :tada:

albe:
Merging the psalm branch was very good and highlighted a couple of still open non-obvious issues with this change.

First of all, the Cli Dispatcher currently handles a ForwardException, which will return an `ActionRequest` as next request. This is wrong of course - so we either need to separate this further or rethink having a common interface.

kitsunet:
I guess we can just NOT handle a forward exception? at least for now. Maybe in the future it could make sense to handle it differently and actually pass on to the "forwarded" to command....

albe:
> I guess we can just NOT handle a forward exception? at least for now. Maybe in the future it could make sense to handle it differently and actually pass on to the "forwarded" to command....

Just noticed, the `CommandController::forward(...)` method doesn't even throw a `ForwardException`, but instead just mutates the current `$this->request`, setting it to `dispatched = false` and throws a `StopActionException`. So we can really just remove this handling of `ForwardException` inside the `CliDispatcher`.

We probably can even deprectate the `Cli\Request::get/isMainRequest()` methods, as they are no longer really useful anywhere.

I already brought that up in the general MVC context once - the idea of throwing an exception to trigger a forward (i.e. control the application flow) is a bad idea anyway. I would rather have the action/command return a meaningful command object (`return new ForwardRequest($command/actionName, $controllerName, $arguments = [], $packageKey = null);`) in order to signal a "I want to stop here and continue by invoking a different action/command". But that's to be discussed/implemented some other time.

kitsunet:
Alright, more tomorrow, I am done for today...

kitsunet:
I restored the branch just in case.... with such a big change you never know...

PR #1553, opened 2019-04-20
BUGFIX: Prevent stability constraint issues with new packages

kitsunet:
By adding the new package directly to the root composer.json instead
of running ``composer require`` we can circumvent issues with the
defined ``minimum-stability`` in said root composer.json as in that
case it will be ignored.

### DIFF
--- a/Neos.Flow/Classes/Package/PackageManager.php
+++ b/Neos.Flow/Classes/Package/PackageManager.php
@@ -17,6 +17,7 @@ use Neos\Flow\Composer\ComposerUtility as ComposerUtility;
 use Neos\Flow\Core\Bootstrap;
 use Neos\Flow\Reflection\ReflectionService;
 use Neos\Flow\SignalSlot\Dispatcher;
+use Neos\Utility\Arrays;
 use Neos\Utility\Files;
 use Neos\Utility\OpcodeCacheHelper;
 use Neos\Flow\Package\Exception as PackageException;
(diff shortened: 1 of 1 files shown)


kitsunet:
Quick draft to get this off my system. This would probably be the fix for the issues people reported in regards to creating new packages. It should be applied to the version that introduced the composer require call in the PackageManager though.

kdambekalns:
Is this still an issue?

Neos.Flow/Classes/Package/PackageManager.php
```diff
@@ -380,7 +381,11 @@ public function createPackage($packageKey, array $manifest = [], $packagesPath =
         $manifest = ComposerUtility::writeComposerManifest($packagePath, $packageKey, $manifest);
 
         if ($runComposerRequireForTheCreatedPackage) {
-            exec('composer require ' . $manifest['name'] . ' @dev');
+            $rootManifestFilename = FLOW_PATH_ROOT . 'composer.json';
+            $rootManifestContent = json_decode(file_get_contents($rootManifestFilename));
+            $rootManifestContent = Arrays::arrayMergeRecursiveOverrule($rootManifestContent, ['require' => [$manifest['name'] => '@dev']]);
+            file_put_contents($rootManifestFilename, $rootManifestContent);
+            exec('cd ' . escapeshellarg(FLOW_PATH_ROOT) . ' && composer update');
```
mhsdesign, line 388: if this is still an issue - i would add this to only update whats needed.
```suggestion
            exec('cd ' . escapeshellarg(FLOW_PATH_ROOT) . ' && composer update ' . $manifest['name']);
```

kdambekalns:
Looking at the merge conflict it seems the code already does roughly that by now…

PR #1557, opened 2019-04-25
BUGFIX: Fix log environment in logging aspects

daniellienert:
As the `'FLOW_LOG_ENVIRONMENT' => []` level was missing in the log data, the log environment data was not set correctly and written to the log by the file writer.


kitsunet:
Yep, this is clearly more in the spirit of the logging stuff. I guess I will prepare two log lines from each change so we can all compare and discuss.

kdambekalns:
Oh well. My change was just from "looking at it", and the data is not even really needed - the package is the same anyway, the class is in the message anyway, so I just added the method name to the log message and removed the additional data. If this is the better way, all the better. :heart:

PR #1567, opened 2019-04-30
[SECURITY] Avoid OpenSSL padding oracle attacks

### LINKED ISSUE #1566: Security openssl_public encrypt ... padding attack
### Description
grafik

grafik

### Steps to Reproduce

https://threatpoint.checkpoint.com/ThreatPortal/threat?threatId=CVE_2016_2107&threatType=protection

#### Expected behavior

use `OPENSSL_PKCS1_OAEP_PADDING` as 4th parameter to harden against the attack

#### Actual behavior

not protecrted

### Affected Versions

Neos:

Flow: master, 5.x, 4.x, 3.x, 2.x

You can get this information by running `composer show`.

kdambekalns:
This avoids OpenSSL Padding Oracle Information Disclosure by
allowing to specify the padding algorithm used in the RSA wallet
service.

Most probably you are not even affected, since only OpenSSL 1.0.1t
and 1.0.2h are vulnerable, but better safe than sorry.

The padding algorithm default is changed to OPENSSL_PKCS1_OAEP_PADDING,
but a fallback decryption is in place for all data that was encrypted with the
previously unsafe padding algorithm.
Therefore you should migrate all your existing encrypted data, by running it through
`decryptWithPrivateKey` and then again through `encryptWithPublicKey` ONCE.

Fixes #1566

### DIFF
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
@@ -46,7 +46,7 @@ class RsaWalletServicePhp implements RsaWalletServiceInterface
      * The padding to use for OpenSSL encryption/decryption
      * @var int
      */
-    protected $paddingAlgorithm = OPENSSL_PKCS1_PADDING;
+    protected $paddingAlgorithm = OPENSSL_PKCS1_OAEP_PADDING;
 
     /**
      * @var boolean
@@ -344,11 +344,17 @@ class RsaWalletServicePhp implements RsaWalletServiceInterface
      * @return string The decrypted plaintext
      * @throws Exception
      */
-    private function decryptWithPrivateKey($cipher, OpenSslRsaKey $privateKey)
+    private function decryptWithPrivateKey($cipher, OpenSslRsaKey $privateKey, $paddingAlgorithm = null)
     {
         $decrypted = '';
         $key = openssl_pkey_get_private($privateKey->getKeyString());
         if (openssl_private_decrypt($cipher, $decrypted, $key, $this->paddingAlgorithm) === false) {
+            // Fallback for data that was encrypted with old default OPENSSL_PKCS1_PADDING
+            if ($this->paddingAlgorithm !== OPENSSL_PKCS1_PADDING) {
+                if (openssl_private_decrypt($cipher, $decrypted, $key, OPENSSL_PKCS1_PADDING) !== false) {
+                    return $decrypted;
+                }
+            }
             $openSslErrors = [];
             while (($errorMessage = openssl_error_string()) !== false) {
                 $openSslErrors[] = $errorMessage;
--- a/TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
+++ b/TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
@@ -13,6 +13,7 @@ namespace TYPO3\Flow\Security\Cryptography;
 
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Security\Exception as SecurityException;
+use TYPO3\Flow\Security\Exception;
 use TYPO3\Flow\Security\Exception\DecryptionNotAllowedException;
 use TYPO3\Flow\Security\Exception\InvalidKeyPairIdException;
 use TYPO3\Flow\Security\Exception\MissingConfigurationException;
@@ -42,6 +43,12 @@ class RsaWalletServicePhp implements RsaWalletServiceInterface
     protected $openSSLConfiguration = [];
 
     /**
+     * The padding to use for OpenSSL encryption/decryption
+     * @var int
+     */
+    protected $paddingAlgorithm = OPENSSL_PKCS1_PADDING;
+
+    /**
      * @var boolean
      */
     protected $saveKeysOnShutdown = true;
@@ -64,6 +71,10 @@ class RsaWalletServicePhp implements RsaWalletServiceInterface
             $this->keystorePathAndFilename = $settings['security']['cryptography']['RSAWalletServicePHP']['keystorePath'];
         } else {
             throw new MissingConfigurationException('The configuration setting TYPO3.Flow.security.cryptography.RSAWalletServicePHP.keystorePath is missing. Please specify it in your Settings.yaml file. Beware: This file must not be accessible by the public!', 1305711354);
+        }
+
+        if (isset($settings['security']['cryptography']['RSAWalletServicePHP']['paddingAlgorithm'])) {
+            $this->paddingAlgorithm = $settings['security']['cryptography']['RSAWalletServicePHP']['paddingAlgorithm'];
         }
     }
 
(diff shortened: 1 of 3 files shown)


kdambekalns:
When decryption fails, `false` is returned, yes. Migration is an issue with my code as of now, yes. But simply trying to decrypt again with the insecure padding on error might still allow for an attack, no?

albe:
I don't think so, unless you can decrypt stuff encrypted with the safe OAEP padding with the unsafe padding method too - but that would be totally weird, especially since it doesn't work the other way around.

kdambekalns:
Feel free to adjust, I need to work on something else now… 🙁

albe:
Done. We could in theory get rid of the paddingAlgorithm setting now again, but it doesn't hurt to have this configurable in case things don't work out or another padding algorithm becomes the defacto secure default.

TYPO3.Flow/Configuration/Settings.yaml
```diff
@@ -626,6 +626,11 @@ TYPO3:
           # Note: This file must not be accessible by the public!
           keystorePath: '%FLOW_PATH_DATA%Persistent/RsaWalletData'
 
+          # The padding to use when encrypting/decrypting data.
+          # The default OPENSSL_PKCS1_PADDING is insecure, you should change it to
```
kdambekalns:
```suggestion
          # PHP's default OPENSSL_PKCS1_PADDING is insecure, thus we change it to
```

kdambekalns:
I like it!

TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
```diff
@@ -65,6 +72,10 @@ public function injectSettings(array $settings)
         } else {
             throw new MissingConfigurationException('The configuration setting TYPO3.Flow.security.cryptography.RSAWalletServicePHP.keystorePath is missing. Please specify it in your Settings.yaml file. Beware: This file must not be accessible by the public!', 1305711354);
         }
+
+        if (isset($settings['security']['cryptography']['RSAWalletServicePHP']['paddingAlgorithm'])) {
```
bwaidelich: This will always yield true even if the `OPENSSL_PKCS1_OAEP_PADDING` constant doesn't exist

TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
```diff
@@ -41,6 +42,12 @@ class RsaWalletServicePhp implements RsaWalletServiceInterface
      */
     protected $openSSLConfiguration = [];
 
+    /**
+     * The padding to use for OpenSSL encryption/decryption
+     * @var int
+     */
+    protected $paddingAlgorithm = OPENSSL_PKCS1_OAEP_PADDING;
```
bwaidelich: Does it really make sense to set a default variable here? I find that a bit misleading i it is usually overridden via `injectSettings`. And wouldn't that lead to a PHP error if the constant does not exist (because openssl is not enabled) - or was that a requirement for Flow?

TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
```diff
@@ -323,12 +342,25 @@ private function getModulus($keyResource)
...
+     * @throws Exception
      */
     private function decryptWithPrivateKey($cipher, OpenSslRsaKey $privateKey)
     {
         $decrypted = '';
         $key = openssl_pkey_get_private($privateKey->getKeyString());
-        openssl_private_decrypt($cipher, $decrypted, $key);
+        if (openssl_private_decrypt($cipher, $decrypted, $key, $this->paddingAlgorithm) === false) {
+            // Fallback for data that was encrypted with old default OPENSSL_PKCS1_PADDING
+            if ($this->paddingAlgorithm !== OPENSSL_PKCS1_PADDING) {
```
bwaidelich: I don't understand that fallback mechanism. It won't be effective by default (i.e. paddingAlgorithm _is_ OPENSSL_PKCS1_PADDING)

TYPO3.Flow/Tests/Unit/Security/Cryptography/RsaWalletServicePhpTest.php
```diff
@@ -42,8 +42,9 @@ public function setUp()
     {
         vfsStream::setup('Foo');
         $settings['security']['cryptography']['RSAWalletServicePHP']['keystorePath'] = 'vfs://Foo/EncryptionKey';
+        $settings['security']['cryptography']['RSAWalletServicePHP']['paddingAlgorithm'] = OPENSSL_PKCS1_OAEP_PADDING;
         $settings['security']['cryptography']['RSAWalletServicePHP']['openSSLConfiguration']['digest_alg'] = 'sha1';
-        $settings['security']['cryptography']['RSAWalletServicePHP']['openSSLConfiguration']['private_key_bits'] = 384;
+        $settings['security']['cryptography']['RSAWalletServicePHP']['openSSLConfiguration']['private_key_bits'] = 1024;
```
bwaidelich: Just being curious: Is that change directly linked to the security vulnerability or a general improvement?

TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
kdambekalns: Having a default seemed like the right thing. But you are right about the undefined constant. Will fix.

TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
kdambekalns: We used to encrypt using `OPENSSL_PKCS1_PADDING` - but change that default to `OPENSSL_PKCS1_OAEP_PADDING`. To be able to still decrypt existing data, we try with the old padding algorithm upon decryption failure, but only if the current padding is not the old default (because then the error cannot be caused by that change.)

TYPO3.Flow/Tests/Unit/Security/Cryptography/RsaWalletServicePhpTest.php
kdambekalns: With the changed padding the key size used for the tests was too small to fit the encrypted data. The size in "real life" is larger by default, at least nowadays.

TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
kdambekalns: Dang, right, Flow will leave the constant unreplaced and the setting will come through as `%WHATEVER_IT_IS%`

bwaidelich: @kdambekalns This should not have been marked resolved then, right?

bwaidelich: Ah, yes it should.. I got confused

TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
bwaidelich: Ah right.. I tend to not read constant names, my bad

TYPO3.Flow/Tests/Unit/Security/Cryptography/RsaWalletServicePhpTest.php
bwaidelich: Makes sense, thanks!

TYPO3.Flow/Classes/TYPO3/Flow/Security/Cryptography/RsaWalletServicePhp.php
```diff
@@ -13,6 +13,7 @@
 
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\Flow\Security\Exception as SecurityException;
+use TYPO3\Flow\Security\Exception;
```
mficzel: Is there a reason why this is used twice?
I cannot say much about the crypto part.

kdambekalns: None, I simply didn't see this and auto-imported what I used.

kdambekalns:
@kaystrobach Do you agree to the solution? Can you verify the fix works?

kaystrobach:
Will check today

PR #1569, opened 2019-05-03
TASK: Fix name of index on PersistentResource.sha1

kdambekalns:
The name IDX_35DC14F03332102A is different from what Doctrine does
auto-generate, but needs to be used due to BC reasons with existing
migrations.

See https://github.com/neos/neos-development-collection/issues/2475

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
+++ b/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
@@ -25,7 +25,7 @@ use Neos\Flow\ResourceManagement\Exception as ResourceException;
  * Model representing a persistable resource
  *
  * @Flow\Entity
- * @ORM\Table(indexes={@ORM\Index(columns={"sha1"})})
+ * @ORM\Table(indexes={@ORM\Index(name="IDX_35DC14F03332102A",columns={"sha1"})})
  */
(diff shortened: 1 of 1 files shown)


kdambekalns:
FYI: This is for 5.3, since the used option is only available as of DBAL 2.9, which we don't allow in earlier versions.

albe:
> FYI: This is for 5.3, since the used option is only available as of DBAL 2.9, which we don't allow in earlier versions.

Wait, `@Index(name="...", ..)` is only available as of DBAL 2.9? No, you mean something else, right?

But what is with versions lower 5.3? They will still have the issue of missing migrations on a clean install?

kdambekalns:
@bwaidelich Well, yes… But we'd need to adjust all the other migrations that rely on this index name as well. The problem is solvable, but I doubt it makes sense with the name given now.

@albe Yes, I mixed that up. That was about the length, see https://github.com/neos/neos-development-collection/pull/2478 Which means we could target lower branches with this one, yay!

But some of the migration differences will still be detected, whatever we do, due to mismatches between DB reporting some values and Doctrine expectations.

albe:
> But some of the migration differences will still be detected, whatever we do

Iff this improves the current situation, let's get it in (lowest branch) anyway.

PR #1573, opened 2019-05-12
BUGFIX: Avoid PHP exception in NamespaceDetectionTemplateProcessor

kdambekalns:
Add error checking when splitting on shorthand syntax.

See https://github.com/neos/neos-development-collection/pull/2484
Related to https://github.com/neos/neos-development-collection/issues/2479


albe:
Travis failing due to #1359, restarting.

kitsunet:
I think we should release this as a bugfix version ASAP because it's affecting a lot of people

PR #1574, opened 2019-05-13
!!! TASK: Remove deprecated legacy logger

daniellienert:
This removes the deprecated logger including all deprecated interfaces.
Please use the PSR-3 interfaces introduced with Flow 5.0 instead.

You must adjust the log settings as well. A code migration is included,
and you can check against the settings in Neos.Flow itself in case you
need to adjust things manually.

The `log(…)` method looked like this with the old `Neos\Flow\Log\LoggerInterface`:

    log(
        string $message, int $severity = LOG_INFO, $additionalData = null,
        string $packageKey = null, string $className = null, string $methodName = null
    );

The new interface still has a `log($level, string $message, array $context)`
method, but be aware of the changed parameters (order)!

You should be using the level-specifc methods instead: `debug(…)`,
`info(…)`, `notice(…)`, `warning(…)`, `error(…)`, `critical(…)`, `alert(…)`,
`emergency(…)`. The signature is the same for all these methods:

    …(string $message, array $context)

To replace the previous passing of information about the place the logging
call was done, use this to pass the `$context`:

    use Neos\Flow\Log\Utility\LogEnvironment;

    $logger->debug('Some log message', LogEnvironment::fromMethodName(__METHOD__));


kdambekalns:
> It would be great to have a core migration for the logging settings.

Working on it…

kdambekalns:
That seems to work…

andrehoffmann30:
Hi @kdambekalns, cool to see your code mirgration. I also tried to write one, while I am at the neos code sprint and I think a search and replace code migration for something like `Systemlogger` to `PsrSystemlogger` and so on could be helpful, but I am not sure whether this is necessary. What do you think @daniellienert , @kdambekalns ? Thank you for your time.

kdambekalns:
Hi @andrehoffmann30 – that sounds like a good idea. You should talk to @kitsunet about the desired result, and there might be quite some edge-cases to this – but the most common case of replacing `SystemLoggerInterface` with `LoggerInterface` should be doable and help most people.

daniellienert:
@andrehoffmann30 - be aware, that changing the used interface / injected or instantiated logger is the easier part - rewriting the log calls from the old interface to the new one could be quite challenging.

kdambekalns:
> rewriting the log calls from the old interface to the new one

Indeed. I added some more instructions to the PR and will merge this after fixing the conflict…

rolandschuetz:
Could you please update the Neos docs to explain the new logger, this is still outdated:
https://docs.neos.io/cms/tutorials/logging-and-debugging-only-neos-3-3

PR #1575, opened 2019-05-13
TASK: Raise PHP minimum dependency to version 7.2

daniellienert:


daniellienert:
No matter if we will require PHP 7.3 or 7.2 for 5.0, PHP 7.1 is not supported for the used PHPUnit 8.1. So I would raise that now.

albe:
Personally I'd probably go for 7.3, I mean 7.4 is around the corner and raising versions is only really possible in majors.

kdambekalns:
> Personally I'd probably go for 7.3

Personally I'd agree, but as we discussed this quickly on the sprint:

- with Neos 5.x we run out of the release cycle of 7.2 _and_ 7.3 anyway
- 7.3 does not have any to-die-for features
- adoption is easier if we don't require 7.3

PR #1578, opened 2019-05-13
!!! TASK: Remove registerRenderMethodArguments() from AbstractViewHelper

kdambekalns:
This removes the deprecated registerRenderMethodArguments() method from
the AbstractViewHelper in the Neos.FluidAdaptor package.

To adjust your code, you need to implement `initializeArguments()` in
your ViewHelper and call `$this->registerArgument(…)` in it for your
former arguments to `render()`.

In `render()` (which must be parameterless now), access the arguments
via `$this->arguments['…']`.


salvaracer:
Hello,
if the `render()` Methods of VH are now parameterless, how can I access the Method Arguments
of the `render()` Method, when I am using AOP (aspects)?

In this advide,
`* @Flow\Around("method(public Foo\Bar\ViewHelpers\CoolViewHelper->render(.*))")`,
this line will output an empty array:
`$whereAreMyArgs= $joinPoint->getMethodArguments()`

albe:
Good question. I guess the best way right now is `$arguments = ObjectAccess::getProperty($joinPoint->getProxy(), 'arguments', true);` until TYPO3\Fluid adds a `getArguments()` method to the ViewHelper base class.

salvaracer:
Yep, I solved it for now with Reflection, your code worked directly:
```
$arguments = \Neos\Utility\ObjectAccess::getProperty($joinPoint->getProxy(), 'arguments', true);
```

PR #1580, opened 2019-05-15
!!! FEATURE: Add function helpers to Eel and remove magic `q`

mficzel:
Function helpers are static functions that are available in Eel without a containing helper. This
change removes the default q variable and instead adds a static method `q` to the flowQuery
class that is used as helper function with the following configuration:

```
Neos:
  Fusion:
    defaultContext:
      q: 'Neos\Eel\FlowQuery\FlowQuery::q'
  ContentRepository:
    labelGenerator:
      eel:
        defaultContext:
          q: Neos\Eel\FlowQuery\FlowQuery::q
```
Note: Nested pathes as identifiers for function-helpers are not allowed and will raise an exception.

This is breaking as it makes it necessary to add the configuration above to Neos.Fusion and Neos.ContentRepository. Also custom code that uses FlowQuery and relies on q beeing always present will have to be adjusted.

Upgrade Instructions:

Only if you are using the EelUtility in to evaluate expressions with `q` AND are  using a custom defaultContextConfiguration you have to make sure that the configuration line `q: 'Neos\Eel\FlowQuery\FlowQuery::q'` is added to this configuration.

### DIFF
--- a/Neos.Eel/Classes/FlowQuery/FlowQuery.php
+++ b/Neos.Eel/Classes/FlowQuery/FlowQuery.php
@@ -122,7 +122,7 @@ class FlowQuery implements ProtectedContextAwareInterface, \IteratorAggregate, \
 
     /**
      * The q function from eel
-     * @param mixed
+     * @param $element mixed
      * @return FlowQuery
      * @throws Exception
      */
--- a/Neos.Eel/Classes/FlowQuery/FlowQuery.php
+++ b/Neos.Eel/Classes/FlowQuery/FlowQuery.php
@@ -122,7 +122,7 @@ class FlowQuery implements ProtectedContextAwareInterface, \IteratorAggregate, \
 
     /**
      * The q function from eel
-     * @param $element mixed
+     * @param mixed $element
      * @return FlowQuery
      * @throws Exception
      */
--- a/Neos.Eel/Classes/FlowQuery/FlowQuery.php
+++ b/Neos.Eel/Classes/FlowQuery/FlowQuery.php
@@ -121,6 +121,17 @@ class FlowQuery implements ProtectedContextAwareInterface, \IteratorAggregate, \
     }
 
     /**
+     * The q function from eel
+     * @param mixed
+     * @return FlowQuery
+     * @throws Exception
+     */
+    public static function q($element): self
+    {
+        return new static(is_array($element) || $element instanceof \Traversable ? $element : [$element]);
+    }
+
+    /**
      * Setter for setting the operation resolver from the outside, only needed
      * to successfully run unit tests (hacky!)
      *
--- a/Neos.Eel/Classes/Utility.php
+++ b/Neos.Eel/Classes/Utility.php
@@ -37,8 +37,8 @@ class Utility
                 $currentPathBase = & $currentPathBase[$pathName];
             }
 
-            if (strpos($objectType, '::')) {
-                if (strpos($variableName, '.')) {
+            if (strpos($objectType, '::') !== false) {
+                if (strpos($variableName, '.') !== false) {
                     throw new Exception(sprintf('Function helpers are only allowed on root level, "%s" was given?', $variableName), 1557911015);
                 }
                 $currentPathBase = self::createClosureFromConfiguration($objectType);
--- a/Neos.Eel/Classes/Utility.php
+++ b/Neos.Eel/Classes/Utility.php
@@ -52,7 +52,7 @@ class Utility
     /**
      * Create a closure to be used as Helper for eel.
      *
-     * @param $objectConfiguration string className followed by two colone and the method name
+     * @param string $objectConfiguration className followed by two colone and the method name
      * @return callable
      */
     private static function createClosureFromConfiguration(string $objectConfiguration): callable
--- a/Neos.Eel/Classes/Utility.php
+++ b/Neos.Eel/Classes/Utility.php
@@ -36,10 +36,34 @@ class Utility
                 }
                 $currentPathBase = & $currentPathBase[$pathName];
             }
-            $currentPathBase = new $objectType();
-        }
 
+            if (strpos($objectType, '::')) {
+                if (strpos($variableName, '.')) {
+                    throw new Exception(sprintf('Function helpers are only allowed on root level, "%s" was given?', $variableName), 1557911015);
+                }
+                $currentPathBase = self::createClosureFromConfiguration($objectType);
+            } else {
+                $currentPathBase = new $objectType();
+            }
+        }
         return $defaultContextVariables;
+    }
+
+    /**
+     * Create a closure to be used as Helper for eel.
+     *
+     * @param $objectConfiguration string className followed by two colone and the method name
+     * @return callable
+     */
+    private static function createClosureFromConfiguration(string $objectConfiguration): callable
+    {
+        list($className, $methodName) = explode('::', $objectConfiguration, 2);
+        return function (...$arguments) use ($className, $methodName) {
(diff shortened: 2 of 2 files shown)


Neos.Eel/Classes/FlowQuery/FlowQuery.php
```diff
@@ -120,6 +120,17 @@ public function __construct($context, array $operations = [])
         $this->operations = $operations;
     }
 
+    /**
+     * The q function from eel
+     * @param mixed
+     * @return FlowQuery
+     * @throws Exception
+     */
+    public static function q($element)
```
bwaidelich:
```suggestion
    public static function q($element): self
```
return type annotation?

Neos.Eel/Classes/Utility.php
```diff
@@ -36,12 +36,31 @@ public static function getDefaultContextVariables(array $configuration)
...
         }
 
         return $defaultContextVariables;
     }
 
+    /**
+     * @param $objectConfiguration
+     * @return callable|object
+     */
+    public static function createContextObjectFromConfiguration($objectConfiguration)
```
bwaidelich:
```suggestion
    private static function createContextObjectFromConfiguration(array $objectConfiguration)
```
can be private and the param is an array, right

Neos.Eel/Classes/Utility.php
```diff
@@ -36,12 +36,31 @@ public static function getDefaultContextVariables(array $configuration)
...
+    {
+        if (strpos($objectConfiguration, '::')) {
```
bwaidelich:
```suggestion
        if (strpos($objectConfiguration, '::') !== false) {
```
(strict comparison)

Neos.Eel/Classes/Utility.php
```diff
@@ -36,12 +36,31 @@ public static function getDefaultContextVariables(array $configuration)
...
+            list($className, $methodName) = explode('::', $objectConfiguration, 2);
+            return function () use ($className, $methodName) {
+                return call_user_func_array(
+                    [$className, $methodName],
+                    func_get_args()
+                );
+            };
+        } else {
```
bwaidelich: The else block is not needed here

Neos.Eel/Classes/Utility.php
```diff
@@ -66,12 +85,16 @@ public static function evaluateEelExpression($expression, EelEvaluatorInterface
...
 
-        $contextVariables['q'] = function ($element) {
-            return new FlowQuery\FlowQuery(is_array($element) || $element instanceof \Traversable ? $element : [$element]);
-        };
-
         $context = new ProtectedContext($contextVariables);
-        $context->whitelist('q');
+
+        /*
+         * @TODO: make this recursive
```
bwaidelich: I don't understand what this TODO means, can you elaborate?

kdambekalns:
I wonder when this would be breaking… If I am just a "regular Neos integrator", this would be transparent to me, no? I mean, `q` exists like before, just through a different mechanism. No?

mficzel:
@kdambekalns q will be available as normal in fusion. It just has to be added to the fusion configuration in a separate pr. Currently cleaning this up

mficzel:
The Neos side of this pr is here: https://github.com/neos/neos-development-collection/pull/2505

Neos.Eel/Classes/FlowQuery/FlowQuery.php
```diff
@@ -120,6 +120,17 @@ public function __construct($context, array $operations = [])
         $this->operations = $operations;
     }
 
+    /**
+     * The q function from eel
+     * @param mixed
```
albe:
```suggestion
     * @param $element mixed
```

Neos.Eel/Classes/Utility.php
```diff
@@ -36,12 +36,36 @@ public static function getDefaultContextVariables(array $configuration)
                 }
                 $currentPathBase = & $currentPathBase[$pathName];
             }
-            $currentPathBase = new $objectType();
-        }
 
+            if (strpos($objectType, '::')) {
+                if (strpos($variableName, '.')) {
```
albe: a `$variableName` of `.foo` would pass this, since strpos returns 0 and this is falsy - is this correct?

```suggestion
                if (strpos($variableName, '.') !== false) {
```

Neos.Eel/Classes/Utility.php
```diff
@@ -36,12 +36,36 @@ public static function getDefaultContextVariables(array $configuration)
...
 
+    /**
+     * Create a closure to be used as Helper for eel.
+     *
+     * @param $objectConfiguration string className followed by two colone and the method name
+     * @return callable
+     */
+    private static function createClosureFromConfiguration(string $objectConfiguration): callable
+    {
+        list($className, $methodName) = explode('::', $objectConfiguration, 2);
```
albe: Why do we destructure the return of `explode` here just to glue it together into an array inside the closure?

```suggestion
        $classAndMethodName = explode('::', $objectConfiguration, 2);
        return function (...$arguments) use ($classAndMethodName) {
            return call_user_func_array($classAndMethodName, $arguments);
        }
```

Neos.Eel/Classes/Utility.php
```diff
@@ -36,12 +36,36 @@ public static function getDefaultContextVariables(array $configuration)
                 }
                 $currentPathBase = & $currentPathBase[$pathName];
             }
-            $currentPathBase = new $objectType();
-        }
 
+            if (strpos($objectType, '::')) {
```
albe:
```suggestion
            if (strpos($objectType, '::') !== false) {
```

Neos.Eel/Classes/Utility.php
mficzel: I simply find it easier to understand that way.

albe: Hm, yeah. It's a bit more explicit indeed. Fine for me, thanks for explaining :)

Neos.Eel/Classes/FlowQuery/FlowQuery.php
```diff
@@ -120,6 +120,17 @@ public function __construct($context, array $operations = [])
         $this->operations = $operations;
     }
 
+    /**
+     * The q function from eel
+     * @param $element mixed
```
kdambekalns:
```suggestion
     * @param mixed $element
```

Neos.Eel/Classes/Utility.php
```diff
@@ -36,12 +36,36 @@ public static function getDefaultContextVariables(array $configuration)
...
+                $currentPathBase = new $objectType();
+            }
+        }
         return $defaultContextVariables;
     }
 
+    /**
+     * Create a closure to be used as Helper for eel.
+     *
+     * @param $objectConfiguration string className followed by two colone and the method name
```
kdambekalns:
```suggestion
     * @param string $objectConfiguration className followed by two colone and the method name
```

Neos.Eel/Classes/FlowQuery/FlowQuery.php
albe: I get this wrong way too often...

mficzel:
@kdambekalns Adjusted as requested. Please reapprove so i can finally merge

kdambekalns:
> Please reapprove so i can finally merge

As I said, if you enabled contributor edits, I'd have fixed it myself right away. ;)

mficzel:
@kdambekalns usually this is always on on my prs, no clue why not here. I probably wurstfingered it. Or github changed the default and i have to look for it explicitly now

PR #1581, opened 2019-05-15
BUGFIX: Remove deprecated PhpUnit functions from tests

theDex:
Replace deprecated PhpUnit functions with supported alternatives.

Fixes https://github.com/neos/neos-development-collection/issues/2498


daniellienert:
@theDex: There are still some tests left that complain about deprecated methods - will you take care of these too?

theDex:
Oops, I missed these. Will take care of it!

PR #1584, opened 2019-05-19
BUGFIX: Remove Doctrine from require-dev

albe:
It's already a require, so the duplication just causes problems, when the versions don't match any more (as they do in current master).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -103,9 +103,7 @@
     },
     "require-dev": {
         "mikey179/vfsstream": "~1.6",
-        "phpunit/phpunit": "~6.0",
-        "doctrine/orm": "~2.5.0",
-        "doctrine/common": ">=2.4,<2.8-dev"
+        "phpunit/phpunit": "~6.0"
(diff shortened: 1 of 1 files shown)


kdambekalns:
Hehe, this change is wrong.

- First of all: `"warning": "AUTOGENERATED FILE, ONLY MODIFY THE .composer.json IN THIS DIRECTORY AND RUN THE MERGE SCRIPT."` is there for a reason. To get rid of anything in here, edit the `.composer.json` file and/or the `*/composer.json` files in the individual packages.
- Next: the requirements come from neos/utility-objecthandling and neos/utility-lock and will be added back to this file whenever the merge script (`Build/BuildEssentials/ComposerManifestMerger.php`) is run again.

And as long as the sub-packages deaclare a require-dev, it should be honoured, as long as it is needed. I'll take care of this now… :)

albe:
My fault, thanks for clarifying ❤️ 😄

PR #1586, opened 2019-05-19
BUGFIX: Trust all proxies by default in Development context

albe:
This fixes an issue when setting up Flow on ddev or similar, where the container acts as a proxy and overrides the request port. That would previously lead to web URLs getting generated with wrong ports, because request headers where not trusted by default, leading to errors in the browser when e.g. `https://flow.ddev.local:80` was opened.

### DIFF
--- a/Neos.Flow/Configuration/Development/Settings.yaml
+++ b/Neos.Flow/Configuration/Development/Settings.yaml
@@ -2,7 +2,7 @@
 # Configuration for the Flow Framework                                   #
 #                                                                        #
 # This file contains additions to the base configuration for the         #
-# Flow Framework when it runs in Testing context.                        #
+# Flow Framework when it runs in Development context.                        #
 #                                                                        #
 # Don't modify this file - instead put your own additions into the       #
 # global Configuration directory and its sub directories.                #
@@ -43,3 +43,7 @@ Neos:
         backendOptions:
           logFileURL: '%FLOW_PATH_DATA%Logs/I18n_Development.log'
           severityThreshold: '%LOG_DEBUG%'
+
+    http:
+      trustedProxies:
+        proxies: "*"


kitsunet:
Yeah, it's difficult. I guess we could also log this clearly ("you are receiving proxy headers that are ignored with the current settings.....") but it can easily cause confusion in beginners with some setups.

robertlemke:
I agree with @kdambekalns here.

albe:
Base intent is to make the development setup more frictionless and logging wouldn't help there (though it would probably make sense generally). After all the effect of this is that the URIs generated are not working, so I see another option of making this isse more prominent and providing better documentation.

see #1587 which I'll adjust to not refer to this change and a pending update to the docs.neos.io ddev setup guide. So I'll close this.

kdambekalns:
❤️

albe:
Okay, after a bit of rethinking, I would like to re-open this PR with a slight adjustment, that should make it tackle the concerns and still improve the development experience, when following our setup guides from the new docs:

https://docs.neos.io/cms/installation-development-setup/docker-and-docker-compose-setup#3-set-up-docker-compose
https://docs.neos.io/cms/installation-development-setup/all-platforms-using-ddev-and-docker#using-custom-settings-only-for-ddev

There we basically suggest creating a context `Development/Docker` or `Development/Ddev`.

What if we just provided a Settings.yaml for those contexts with core, that only included the trustedProxies `*` setting? Or, better yet, adjust the documentations to suggest one more generic context `Development/Container` or similar in all those cases, so we only have to have one single such sub-context Settings.yaml?

wdyt @kdambekalns @kitsunet @robertlemke ?

kdambekalns:
What about including the suggested setting in commented form with the `Settings.yaml.default` we ship? Avoids having another context-specific file to care for, no bias towards a naming, …

albe:
Would be better than nothing. Not frictionless though. But if that's what we can agree upon, let's do that!

PR #1591, opened 2019-05-21
!!! TASK: Remove ObjectManager.getSettingsByPath()

kdambekalns:
Instead of `getSettingsByPath(…)` use settings injection or the
`ConfigurationManager` to get settings.


kdambekalns:
Pushed to the wrong remote, will reopen correctly.

PR #1596, opened 2019-05-21
BUGFIX: Return type hint should reflect nullable

ralfkuehnel:
If no controller could be found for the given arguments RoutingCommandController:getControllerObjectName() returns null. The return type hint should reflect that to avoid a TypeError.


daniellienert:
Hey @ralfkuehnel,
thanks for your first contribution! Absolutely makes sense!

PR #1597, opened 2019-05-22
BUGFIX: Fix TypeError if subpackage is empty

ralfkuehnel:
Sorry, found another one...

if subpackage is empty RoutingCommandController:getControllerObjectName() should be called with an empty string for the subPackageKey argument. Otherwise an TypeError is thrown because the argument is not nullable.


albe:
Nothing to be sorry about at all! 

I do think this is more an error of the type hint in the method argument. SubPackage should be nullable, because there's a difference between "no SubPackage" (empty string) and "use current SubPackage" (null).

ralfkuehnel:
You're right, will change it.

daniellienert:
Thanks @ralfkuehnel for the bugfix!

PR #1606, opened 2019-05-30
FEATURE: Allow mapping whole request body to a single action argument

### LINKED ISSUE #1554: Requesthandler should support json array without object wrapper
### Description

It should be possible to send a json array without object wrapper to a controller action with Content-Type `json/application`

### Steps to Reproduce
`curl --header "Content-Type: application/json" --request POST --data '[{"foo":"bar"},{"bar":"foo"}]' https://myhost/myroute`

#### Expected behavior
There should be the possibility to define a name for this special case in the route. 

#### Actual behavior

```
Neos\Flow\Mvc\Exception\InvalidArgumentNameException:
Invalid argument name (must be a non-empty string). 1210858767
Neos\Flow\Mvc\ActionRequest_Original::setArgument("0", array|1|)
```

Flow wants to map the element keys as request arguments which makes no sense here. The array should be mapped to an argument name which should be configurable in the route.
### Affected Versions
All versions

albe:
This allows to annotate a method action with the new `@Flow\MapRequestBody("$foo")`
Annotation, which will lead to the full parsed request body being mapped
into the action argument `$foo`. Previously this was not possible and
the request body had to be wrapped with an object with a single `foo` property.
This is useful for APIs that want to receive specific payloads that are
defined on third parties or want to accept a list of models as an argument.

```
POST /api/v1/foo [{ "name": "foo1" }, { "name": "foo2" }]
```

```php
/**
 * @param array<Foo> $foos
 * @Flow\MapRequestBody("$foos")
 */
public function fooAction(array $foos)
```

Resolves #1554


albe:
Things green now.

albe:
TODO: Documentation for the feature

Neos.Flow/Classes/Mvc/Controller/Argument.php
```diff
@@ -186,7 +189,7 @@ public function setValidator(ValidatorInterface $validator)
      * @return ValidatorInterface The set validator, NULL if none was set
      * @api
      */
-    public function getValidator()
+    public function getValidator(): ValidatorInterface
```
albe:
```suggestion
    public function getValidator(): ?ValidatorInterface
```

Neos.Flow/Classes/Mvc/Controller/Argument.php
```diff
@@ -186,7 +189,7 @@ public function setValidator(ValidatorInterface $validator)
      * @return ValidatorInterface The set validator, NULL if none was set
```
albe:
```suggestion
      * @return ValidatorInterface|null The set validator, NULL if none was set
```

albe:
Will do if this approach is considered good :) See discussion in #1554 - there are different ways on how to achieve this, maybe also without a new annotation, but I just preferred this route in the end.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/PropertyMapping.rst
```diff
@@ -236,52 +236,92 @@ on large collections::
...
-			$commentConfiguration = $this->arguments['comment']->getPropertyMappingConfiguration();
-			$commentConfiguration->allowAllProperties();
-			$commentConfiguration
-				->setTypeConverterOption(
-				\Neos\Flow\Property\TypeConverter\PersistentObjectConverter::class,
-				\Neos\Flow\Property\TypeConverter\PersistentObjectConverter::CONFIGURATION_CREATION_ALLOWED,
-				TRUE
-			);
-		}
+Property Mapping Configuration in the MVC stack
```
albe: Why did I remove this "admonition" and make it an own headline? Because the section about "Configuring the property mapping process" is already pretty long, and it was not possible to directly link (with a hash) into the pretty elaborate part that covers mapping inside the MVC layer.

Also, this is anyway more of a deeper explanation within a specific context, rather than an "admonition".

albe:
Added functional tests to verify the feature and thereby fixed a bug with the anonymous argument mapping in the annotation.

albe:
Restarted Travis Jobs... should turn green again soon

PR #1607, opened 2019-05-31
TASK: Improve exception when a wrong logger class is given

daniellienert:


albe:
Merging as explained above. If issues should come up still, you may blame it on me

PR #1608, opened 2019-06-04
TASK: Tweak dev dependencies

kdambekalns:
See https://github.com/neos/flow-development-collection/pull/1584

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,8 +22,7 @@
         "symfony/console": "~2.8",
         "neos/composer-plugin": "^2.0.0",
         "typo3fluid/fluid": "~2.1.3",
-        "ext-mbstring": "*",
-        "ext-xml": "*"
+        "ext-mbstring": "*"
     },
     "replace": {
         "neos/cache": "self.version",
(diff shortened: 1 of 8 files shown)


composer.json
```diff
@@ -22,8 +22,7 @@
         "symfony/console": "~2.8",
         "neos/composer-plugin": "^2.0.0",
         "typo3fluid/fluid": "~2.1.3",
-        "ext-mbstring": "*",
-        "ext-xml": "*"
```
albe: is the ext-xml removed intentionally? That was added with #1357

composer.json
```diff
@@ -103,7 +102,9 @@
     },
     "require-dev": {
         "mikey179/vfsstream": "~1.6",
-        "phpunit/phpunit": "~6.0"
+        "phpunit/phpunit": "~6.0",
+        "doctrine/orm": "~2.5.0",
```
albe: Readding doctrine as dev dependencies? See #1584

kdambekalns: Exactly, see #1584 :)

composer.json
kdambekalns: Ok, then it must be added to `Neos.Flow/composer.json` as well!

PR #1587, opened 2019-05-19
TASK: Add section for configuration of trusted proxies in container

albe:
Adds a small note that mentions having to configure the trusted proxies in ddev and similar environments. Also explains that Flow therefore trusts all proxies by default in Development context.

Depends on #1586

PR #1609, opened 2019-06-05
TASK: Fix formatting of note

albe:
Related to #1587


albe:
Looking like this right now:
image

PR #1611, opened 2019-06-07
FEATURE: Allow to skip validation of uninitialized proxies

kdambekalns:
The new option `skipUnInitializedProxies` allows to skip validation
of uninitialized proxy instances, to avoid recursions down into
unchanged hierarchies.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
@@ -11,7 +11,6 @@ namespace Neos\Flow\ObjectManagement\Proxy;
  * source code.
  */
 
-use Doctrine\ORM\Mapping as ORM;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Reflection\ClassReflection;
 use Neos\Flow\Reflection\ReflectionService;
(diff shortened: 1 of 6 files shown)


kitsunet:
The interesting question is, when do we actually ever need to validate uninitialised proxies. Because this must be an existing entity that was not changed, which means it went through validation at the time it was written last...

albe:
As said, theoretically, you could have an aggregate with an entity inside, that is "valid" depending on some external (but still within the aggregate) constraints/state. So even if that entity itself wasn't changed and hence not initialized, the validation is important, because that other state could have changed. I have a hard time imagining a practical example straight away, but it's a possibility.

Basically: you should not partially validate an aggregate - the aggregate is to be considered a single whole. And as long as you build sane (small) Aggregates, that is not an issue.

kdambekalns:
But in that case you'd write a custom validator for said object, no?

albe:
Maybe, but if the validator is for that sub entity object (and not the root entity), it wouldn't be invoked any longer and still lead to partially (in)valid aggregates.

kdambekalns:
I'd say in that case you should write a validator for that root entity. I still think this is a valid case, and I wrote this to fix a real problem a customer had. How do we proceed?

albe:
Wait, still needs proper documentation as it's a new feature.

kdambekalns:
@albe Added some documentation, is that sufficient?

albe:
Looks good! Thanks :)

PR #1615, opened 2019-06-12
!!! BUGFIX: Don't start session for sessionless tokens

### LINKED ISSUE #1614: Session is started for sessionless tokens when an EntryPoint is configured
### Description

Since version 2.1 Flow supports "Sessionless authentication" (see 1f35580054d828afe96acb86cf45798f37fab88f)

However, when an `EntryPoint` is configured, a session is started because the `Dispatcher` invokes `Neos\Flow\Security\Context::setInterceptedRequest()`

### Steps to Reproduce

1. Configure a [sessionless authentication](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Security.html#sessionless-authentication-tokens)
2. Configure an [authentication entry point](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Security.html#authentication-entry-points)
2. Authenticate

#### Expected behavior

No session (cookie) should be created

#### Actual behavior

A session is started and the `Neos_Session` cookie is set

### Affected Versions

Flow:

2.1+

bwaidelich:
This adds a condition to the `Dispatcher` to avoid
`Neos\Flow\Security\Context::setInterceptedRequest()` from
being invoked when authenticating an authentication token
that implements the `SessionlessTokenInterface`.

This is a breaking change if code relies on the fact that the intercepted
request is stored even when using sessionless authentication.

Fixes: #1614


bwaidelich:
Note: I created the PR against master even though it is a bugfix, because it's also a change in behavior.

kitsunet:
In that case I would like to block this and fix it in the PSR-7 PR, because it changes and moves that part anyway.

albe:
Since 5.2 the Context itself is no longer of scope session, so there it could be solved by not setting it in the SessionDataContainer iff the request does only contain sessionless tokens (and hence be b/c), right?

bwaidelich:
> Since 5.2 [...] it could be solved by not setting it in the SessionDataContainer iff the request does only contain sessionless tokens

You mean, we could leave the `$securityContext->setInterceptedRequest()` but then avoid that from being stored in the session? I don't know if that's any less breaking tbh.

Thinking about it again I think that we should include this fact. If someone relies on sessionless authentication to contain the intercepted request, this will break sooner or later.

PS: We should also only store the intercepted request if it's an idempodent method (or maybe just `GET`).. I've seen a lot of exceptions because Flow tried to redirect to an URL that must only be called via `POST` after successful authentication

albe:
> You mean, we could leave the $securityContext->setInterceptedRequest() but then avoid that from being stored in the session?

More or less, yes. Setting the intercepted request is okay and useful. The intention of the PR is why does the intercepted request **need to be** stored **in the session** (for a sessionless authentication method)?

The use-case for it is (as far as I understand):
- you request `/foo/bar` (intercepted request), which is protected
- you are not/no longer authenticated
- the request is intercepted and you are redirected to an authentication page
- after authentication you should be redirected to `/foo/bar` to reduce disruption

IMO the storage of intercepted requests should really be independent of the authentication strategy. You should be able to use sessionless authentication but still store some information (like intercepted request) in a session, if you want. In the concrete case, we probably need to make the storage of the intercepted request configurable, with a default strategy being session storage, an alternative being URL parameter. How the interceptedRequest is stored, should be transparent to the user of SecurityContext though.

> We should also only store the intercepted request if it's an idempodent method

Yes, non-GET requests should not be stored and redirected back to - we'd need to redirect to the "previous" (referrer) URL from where the POST was invoked. The user at most loses his form data, but at least doesn't end up at an arbitrary page after auth.

bwaidelich:
@kitsunet @albe Thanks for your feedback.
I like the idea to extend the whole implementation so that the storage of the intercepted request is configurable.
But that would require a lot more thought (i.e. currently the method is just invoked for every token and the last one wins).

This is just meant as a bugfix for the very real scenario that loads of sessions are generated even though they should not.

I'd prefer to merge this one as such and to create a new issue for a more sophisticated "intercepted request" handling.

albe:
Okay, so the (potentially unwanted) effect of this change as is, is, that we completely skip setting the intercepted request instead of just avoiding the storage in the session.

I'd be fine with this, if we just comment this clearly in the code and have a follow-up issue where we can work out the details of a clean refactoring.

Neos.Flow/Classes/Mvc/Dispatcher.php
```diff
@@ -110,7 +110,9 @@ public function dispatch(RequestInterface $request, ResponseInterface $response)
                 } else {
                     $securityLogger->info(sprintf('Starting authentication with entry point of type "%s"', get_class($entryPoint)), LogEnvironment::fromMethodName(__METHOD__));
                 }
-                $securityContext->setInterceptedRequest($request->getMainRequest());
+                if (!$token instanceof SessionlessTokenInterface) {
```
albe:
```suggestion
                // TODO: We should only prevent storage of intercepted request in the session here, but we don't have a different storage mechanism yet.
                if (!$token instanceof SessionlessTokenInterface) {
```

A quick draft of a code-comment... not very elaborated, sorry :/

albe (approved):
Okay, let's do it. It improves the current situation until it's solved properly

bwaidelich:
@albe Thanks! I fully agree to what you wrote and will take care of creating an issue for the more sophisticated implementation – and another issue/PR to avoid storing the intercepted request for non-GET requests

bwaidelich:
The two follow-up tickets as promised:
* #1693 Extensible "intercepted request" handling (just collecting ideas)
* #1694 Don't intercept unsafe requests (with fix)

@kitsunet would you re-consider your veto please? The "breakiness" of this is really just theoretical – I doubt that anyone relies on the broken behavior

daniellienert:
@bwaidelich: We have merge conflicts due to psr-7 change I guess, would you solve them?

bwaidelich:
yes, I will take care. Thanks

Neos.Flow/Classes/Http/Component/SecurityEntryPointComponent.php
```diff
@@ -39,19 +40,36 @@ public function handle(ComponentContext $componentContext)
         }
 
         $actionRequest = $componentContext->getParameter(DispatchComponent::class, 'actionRequest');
-        $entryPoint = array_reduce($this->securityContext->getAuthenticationTokens(), function ($foundEntryPoint, TokenInterface $currentToken) {
-            return $foundEntryPoint ?? $currentToken->getAuthenticationEntryPoint();
-        }, null);
```
bwaidelich: I extracted this to a method `getFirstTokenWithEntryPoint()` that stops when the first token is found. Before this would always iterate all tokens

Neos.Flow/Classes/Http/Component/SecurityEntryPointComponent.php
```diff
@@ -39,19 +40,36 @@ public function handle(ComponentContext $componentContext)
...
+     *
+     * @return TokenInterface|null
+     */
+    private function getFirstTokenWithEntryPoint(): ?TokenInterface
+    {
+        foreach ($this->securityContext->getAuthenticationTokens() as $token) {
+            if ($token->getAuthenticationEntryPoint() !== null) {
+                return $token;
+            }
+        }
```
albe: Missing return statement, says psalm and I second ;)

bwaidelich: picky bastards :)
This was just a test to see if the checks work *g

albe (approved):
I'm satisfied now :)

PR #1623, opened 2019-06-16
FEATURE: Allow overwriting instead of merging in arrayMergeRecursiveWithCallback

cvette:
This allows to specify a callback that determines whether the value of the first array is unset before merging.

**What I did**
An additional callback parameter is added to the ```Neos\Utility\Arrays:arrayMergeRecursiveOverruleWithCallback``` method. This allows to conditionally  unset the value of the first array before merging the second array value.

Needed for neos/neos-development-collection#2238

**How to verify it**
See the created test case.

**Checklist**

### DIFF
--- a/Neos.Utility.Arrays/Classes/Arrays.php
+++ b/Neos.Utility.Arrays/Classes/Arrays.php
@@ -109,11 +109,20 @@ abstract class Arrays
      *
      * @param array $firstArray First array
      * @param array $secondArray Second array, overruling the first array
-     * @param callable $toArray The given closure will get a value that is not an array and has to return an array. This is to allow custom merging of simple types with (sub) arrays
+     * @param \Closure $toArray The given callable will get a value that is not an array and has to return an array.
+     *                          This is to allow custom merging of simple types with (sub) arrays
+     * @param \Closure|null $overrideFirst The given callable will determine whether the value of the first array should be overridden.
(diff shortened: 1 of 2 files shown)


albe:
Hey, thanks for that. TBQH I'm hesitant to merge this, as it adds an edge-case parameter (unset first element?) that is hard to understand the intention for to a very core method and I feel this would lead down the same path that the PHP language itself suffers.

I see two solutions:
- create a copy of the array merge function inside Neos for this very specific use-case only
- adjust the parameter/function to be more clear to a broad use-case

If the latter, I suggest to consider the following:
- the "unset first array" basically leads to an "override instead of merge" - so basically it's a per-element opt-out of the merge algorithm - let's name it more closely to this
- the closure only needs to be invoked, if the other override conditions (empty/simple value) aren't met already (a function call is costly)
- check if the `instanceof` check moved outside the loop with a default static `return false` closure is more performant
- does the provided test really verify the new behaviour? From what I see, the test would already pass without the new closure parameter, no?

cvette:
Made some adjustments to the code. Still not sure about the closure parameter name. Naming things is hard...

The test fails without the new parameter. k2 gets merged.

albe:
For the record on how this ended up at that implementation:
The idea was, that the callback should be more generic and allow the caller to to influence the merging behaviour on more than just the current element value, as required for the Neos use case. A follow up could now theoretically even merge the two callbacks to make the whole merging process adjustable in the caller, but that ofc would be a b/c break.

Issue #1636, opened 2019-06-21
XSD Schema generator for ViewHelpers improvements


albe:
In order to import a ViewHelper namespace with autocompletion, the option to specify the `xmlns:f="..."` attribute on a tag is available. This allows the IDE to properly parse the `f:...` namespaces in templates without marking them as errors, and, if the according XSD schema is linked correctly, to support the user with autocompletion.

However, since the Typo3Fluid standalone split, it is not easily possible any more, to generate a XSD schema for all `f` namespace ViewHelpers, because the tool only accepts a single PHP namespace to scan.

Also, the documentation for how to make use of that xml namespace import is scarce and the Kickstarter still uses the `{namespace f=...}` syntax for generating templates.

Actionables:
- [ ] schema generator should allow scanning multiple PHP namespaces for a single schema
- [ ] the documentation should be improved, to show how to properly make use of the `xmlns:f` import syntax, generate a schema and link it in common IDEs (PhpStorm)
- [ ] Adjust the kickstarter, to make use of xmlns imports
- [ ] Automatically generate an XSD schema on release for Typo3Fluid+FluidAdaptor and make it available versioned online

PR #1637, opened 2019-06-22
TASK: Include TYPO3Fluid for reflection

albe:
This is needed to be able to generate a XSD schema for the TYPO3 Fluid default ViewHelpers.

Depends on #1638

### DIFF
--- a/Neos.Flow/Configuration/Settings.yaml
+++ b/Neos.Flow/Configuration/Settings.yaml
@@ -322,7 +322,8 @@ Neos:
       # by specifying corresponding regular expressions that don't match classes to exclude.
       #
       # Note: The previous setting "excludeClasses" has been deprecated with Flow 3.0
-      includeClasses: []
+      includeClasses:
+        'typo3fluid.fluid': ['.*']
 
     package:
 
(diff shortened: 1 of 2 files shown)


albe:
Related to #1636 

@kitsunet Are there any classes in TYPO3Fluid that should definitely NOT be reflected? i.e. should we narrow down the `includeClasses`?

albe:
> [Semantical Error] The annotation "@retrurn" in method TYPO3Fluid\Fluid\ViewHelpers\DebugViewHelper::getValuesOfNonScalarVariable() was never imported.
Did you maybe forget to add a "use" statement for this annotation?

:|

albe:
> 1) Neos\Flow\Tests\Functional\Configuration\ConfigurationValidationTest::configurationValidationTests with data set #4 ('Development', 'Settings')
Settings.Neos.Flow.object.includeClasses -> **expected: type=array found: type=array**

Uhm...

ComiR:
https://github.com/neos/flow-development-collection/blob/bab4abb0d96fbb5e0644c10c10f1fa15be2de332/Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.object.schema.yaml#L7 seems wrong.
It looks like it should be
```yaml
    items:
      type: array
      required: true
      items: { type: string, required: true }
```
(no expertise here - please test before usage 😉)

albe:
You were right, but that still didn't fix it :/ I'm baffled at why this fails and how this never failed before?

kdambekalns:
This only turned green, because Travis CI seems to install a newer `typo3fluid/fluid` than declared in our dependencies…

kdambekalns:
https://github.com/neos/flow-development-collection/pull/1638 "fixed" this, but only for the dev-collection tests. "Real" use still required `~2.1.3`. To be fixed with https://github.com/neos/flow-development-collection/pull/1757.

PR #1638, opened 2019-06-22
TASK: Loosen typo3 fluid dependency

albe:
This allows to install any version of TYPO3 Fluid >= 2.1.3, < 2.5.0 instead of the previously limiting to ~2.1.3
Since Flow 5.0+ requires TYPO3 Fluid 2.5.x, this is consistent.

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php
+++ b/Neos.FluidAdaptor/Classes/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php
@@ -47,6 +47,8 @@ class NamespaceDetectionTemplateProcessor extends FluidNamespaceDetectionTemplat
 			)
 		)/xs';
 
+    const SPLIT_PATTERN_TEMPLATE_OPEN_NAMESPACETAG = '/xmlns:([a-z0-9\.]+)=("[^"]+"|\'[^\']+\')*/xi';
+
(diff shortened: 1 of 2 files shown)


albe:
https://github.com/TYPO3/Fluid/compare/2.1.3...2.4 shows the changes since 2.1.3 which we required till now. Nothing really breaking, though there is a behavorial change in 2.4+ with https://github.com/TYPO3/Fluid/commit/ed01d07fb2bfeba72a7d5a58a59b44413d1b515e

I'd still allow < 2.5.0 instead of < 2.4.0 because the next Flow version requires 2.5 minimum, so there'd be a gap. Open to change this though.

albe:
D'oh... breaks because 4.3 FluidAdaptor still depends on `NamespaceDetectionTemplateProcessor::SPLIT_PATTERN_TEMPLATE_OPEN_NAMESPACETAG` which was removed with https://github.com/TYPO3/Fluid/commit/5e296f7128a36ee36114e6d5d189610ddbdc2e9f (Fluid 2.3.0)

So... guess it needs to be < 2.3.0 unless we work around that, e.g. by adding the cons to our own class.

kdambekalns:
I know I ack'ed this but… it broke (again). The `composer.json` is autogenerated, so this needed to change `Neos.FluidAdaptor/composer.json` as well.

kitsunet:
Mmm, maybe we can test the merged composer.json somehow (thought for the future).

PR #1640, opened 2019-06-23
!!! BUGFIX: Simplify and strengthen the type matching pattern

albe:
This pattern was overspecified for common types, but at the same time not closed. Hence this was buggy for any type that started with one of the specified types, like a custom `DateTimeRange` type, which would be matched as `DateTime` only.

The new pattern will now just match any identifier made up of characters, digits, backslashes and underscores up to a whitespace or lineend, which also matches all the previously hard-coded types.

This is not breaking in the normal sense, but the change in a very core regex pattern can cause different behaviour in some edge-cases, hence why this bugfix is not applied to lowest maintained branch.

Followup for #1442


albe:
Had that thought too, but it did allow that previously, so... I'm fine with making the pattern only match valid PHP identifiers.

PR #1643, opened 2019-06-24
BUGFIX: Allow using Flow with PHP wrappers

marcosbc:
**What I did**

Add support for using a fallback to verify whether the `PHP_BINARY` for the currently configured PHP binary file matches the one being used currently.

The current logic only resolves the symlink, which may not always work, e.g. what if the `php` binary is being executed through a wrapper like this?

    #!/bin/sh
    . /path/to/setenv.sh
    exec /path/to/php.bin "$@"

(Where `php.bin` is the binary file and `setenv.sh` a script with sets environment variables - Wrappers like these are heavily used in Bitnami installations.)

**How I did it**

Before Flow compares which PHP binary is being used (and which it is supposedly configured to use), we run a PHP `exec` to print `PHP_BINARY`.

Then, we store the result and if no errors were thrown, use this as the detected PHP binary path to compare with. If any errors were detected (via the "exec" exit code), we use the original logic that resolves any symlink it's pointing to.

If it matches the existing one, it means everything went great, if not an error will be thrown like before.

**How to verify it**

- A correct PHP wrapper pointing to the PHP binary (e.g. php.bin) is allowed for being used for CLI subrequests (method `ensureCLISubrequestsUseCurrentlyRunningPhpBinary`).
- An invalid PHP wrapper fails when being used for CLI subrequests (method `ensureCLISubrequestsUseCurrentlyRunningPhpBinary`).

**Checklist**

- [x] Code follows the PSR-2 coding style - Checked
- [x] Tests have been created, run and adjusted as needed - Couldn't find any tests for this part
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html) - Using 4.3 branch

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -758,6 +758,11 @@ class Scripts
             return;
         }
 
+        // Ensure the actual PHP binary is known before checking if it is correct. If empty, we ignore it because it is checked later in the script.
+        if (strlen($phpBinaryPathAndFilename) === 0) {
+            return;
+        }
+
         // Try to resolve which binary file PHP is pointing to
         exec($phpBinaryPathAndFilename . ' -r "echo PHP_BINARY;"', $output, $result);
         if ($result === 0 && sizeof($output) === 1) {
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -758,8 +758,15 @@ class Scripts
             return;
         }
 
-        // Resolve any symlinks that the configured php might be pointing to
-        $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
+        // Try to resolve which binary file PHP is pointing to
+        exec($phpBinaryPathAndFilename . ' -r "echo PHP_BINARY;"', $output, $result);
+        if ($result === 0 && sizeof($output) === 1) {
+            // Resolve any wrapper
+            $configuredPhpBinaryPathAndFilename = $output[0];
+        } else {
+            // Resolve any symlinks that the configured php might be pointing to
+            $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
+        }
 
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -758,8 +758,15 @@ protected static function ensureCLISubrequestsUseCurrentlyRunningPhpBinary($phpB
             return;
         }
 
-        // Resolve any symlinks that the configured php might be pointing to
-        $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
+        // Try to resolve which binary file PHP is pointing to
+        exec($phpBinaryPathAndFilename . ' -r "echo PHP_BINARY;"', $output, $result);
```
albe: What happens if `$phpBinaryPathAndFilename` is empty?

marcosbc: Sorry for the delay on getting back to you! The command would fail, and the next condition would not be entered (so it would be trying to resolve a symlink). But in any case, it is not pretty so I improved the code to avoid this.

Thanks for checking this!

kdambekalns:
Woah. Sad that having a simply `php` CLI binary in your path is still something that not the norm, it seems… 😢 Why do hosters do that to us?

marcosbc:
@kdambekalns In the case of Bitnami installations, there are a couple of reasons for this:

- Bitnami stacks are self-contained, meaning each stack install its dependencies in a specific location (the installation directory).
- Bitnami packages can be installed at multiple different system paths, meaning you could install one stack at `/opt/bitnami` and then another one at `/opt/neos` (each of those could be a different version of the stack and/or with different library versions).

The downside is that binaries need to find where libraries are located, and the way Bitnami handled this issue was by creating wrappers for these binaries (i.e. the `php` executable is a wrapper to `.php.bin`).

The reason for this PR is that we could use the `php` wrapper, as it was already possible in previous releases of this package. Please let me know if you have any questions on this.

P.S.: Is there any ETA on the release for the release of the `flow-development-collection` package? Thank you!

kdambekalns:
All is well, I am just constantly amazed how compicated things can become in some setups. :)

Well, releases happen every now and then. Patch-level could be soonish. The next major hopefully still in August, as planned.

PR #1644, opened 2019-06-25
BUGFIX: Escape curly braces in CommandController template

daniellienert:
Fixes: #1645


daniellienert:
Thanks @kitsunet - does that also mean, thats the way we fix it?

kitsunet:
I fear, although I guess we could think about switching to the `\Neos\Flow\Mvc\View\SimpleTemplateView` for those templates to avoid this...

PR #1658, opened 2019-06-29
FEATURE: Add psalm static code analysis to travis matrix

albe:
This will provide us with immediate feedback on newly introduced typing errors in the code base.

To update the psalm-baseline, which contains the current set of errors that are simply ignored in order to gradually introduce static analysis to a larger code base, you can just run
`bin/psalm --config=Packages/Framework/psalm.xml --update-baseline`
on an installed distribution.
Note though, that this will only remove fixed errors, not add new errors to the baseline. So we could automate this step.

In order to add new errors to the baseline, you have to run
`bin/psalm --config=Packages/Framework/psalm.xml --set-baseline=Packages/Framework/psalm.xml`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ matrix:
   fast_finish: true
   include:
     - php: 7.2
+      env: STAN=true
+    - php: 7.2
       env: DB=mysql
       addons:
         mariadb: '10.2'
@@ -50,6 +52,7 @@ install:
   - composer update --no-progress --no-interaction
   - php Build/BuildEssentials/TravisCi/ComposerManifestUpdater.php .
   - composer update --no-progress --no-interaction
+  - if [ "$STAN" = "true" ]; then composer require --dev vimeo/psalm; fi
 before_script:
   - echo 'date.timezone = "Africa/Tunis"' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - echo 'opcache.fast_shutdown = 0' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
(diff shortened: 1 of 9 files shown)


albe:
Nice, after merging master into this, it immediately highlights a couple of (new) problems in the codebase. Now we just need a usable exit-strategy for possible false-negatives. Manually managing the psalm-baseline.xml is cumbersome and not to be expected from a contributor.

What would be nice, is a bot that a maintainer could trigger via a comment to the PR, which will then add a commit after running `bin/psalm --config=Packages/Framework/psalm.xml --set-baseline=Packages/Framework/psalm.xml`
Anyone got an idea how to do that? @daniellienert @kdambekalns maybe?

daniellienert:
@albe: How to proceed with that? What could that false-positives be? 
As far as I see the new errors shown here are legit. Tried to fix some in #1687, but a rerun will help to see what is already fixed. Also saw your fixes in #1665 which also improve code quality a lot. 
Would be great to have it in for the upcomming major.

albe:
> What could that false-positives be?

Something like a (formally) wrong typehint, that is given because of our code guidelines or sth. - basically anything that we conciously would say: yeah okay, this is fine and psalm is just a bit too strict here.

Currently this means that a maintainer has to checkout the PR branch on a flow-development-distribution (! not collection) setup, then run `bin/psalm --config=Packages/Framework/psalm.xml --set-baseline=Packages/Framework/psalm.xml` and commit the change to the branch. I mean, maybe that's just fine, because a maintainer has to decide that it's okay to bypass psalm in that case anyway.

> How to proceed with that?

I would suggest we merge your PR, then rebase this and fix the remaining stuff, then merge my other PR after a rebase. Definitely let's try to get this static analysis in ASAP.

I'd even wish to have it back in lower branches, but I think this will lead to some mess with the diverging baseline configs and keeping them in synch on upmerges. Let's get it into master and then try to get rid of all the baseline errors one after the other.

The process for that basically is:
- checkout master (from flow-development-distribution) after this is in
- create a new branch
- run `bin/psalm --config=Packages/Framework/psalm.xml --ignore-baseline`, maybe also use `--report=psalm-report.txt` (or .json or .xml - see https://github.com/ErikBooijCB/psalm/blob/master/src/psalm.php#L172) if the console is flooded with too many errors
- fix some of the errors
- run `bin/psalm --config=Packages/Framework/psalm.xml --update-baseline`
- commit and push to create a PR

albe:
Just found a good example of a "false-negative":

```
ERROR: PossiblyNullReference - Packages\Framework\Neos.Utility.Unicode\Classes\TextIterator.php:263:56 - Cannot call method getValue on possibly null value
            $allValues[] = $this->getCurrentElement()->getValue();
```

```php
        $this->rewind();
        $allValues = [];
        while ($this->valid()) {
            $allValues[] = $this->getCurrentElement()->getValue();
            $this->next();
        }
```

This is a false negative, because as long as `valid()` returns true, `getCurrentElement()` will return a valid entry and not null.

albe:
@daniellienert Tests pass now, so you can merge if you see fit. Afterwards we can incrementally improve the code base with the above method.

albe:
@kitsunet I would even suggest merging this before psr7. I'll take care about rebasing then.

albe:
- `STAN` no issue with that, renaming it to `STATIC_ANALYSIS` - in fact we might to have both psalm AND PHPStan running, maybe even Phan (see https://badootech.badoo.com/php-code-static-analysis-based-on-the-example-of-phpstan-phan-and-psalm-a20654c4011d)
- TODO: - removed, I think we are at a solid base right now and can from there go stricter or less strict how we see fit incrementally

PR #1660, opened 2019-07-02
TASK: Safelist branches for travis builds

albe:
This prevents builds from running doubly on branches created on this repository for PRs, e.g. through the StyleCI bot or by github inline PRs.

See https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: php
+branches:
+  only:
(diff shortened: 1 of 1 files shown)


albe:
Seems to work as the checks for this PR show :)
Next would be to prevent styleci from running twice, though that is probably less problematic.

PR #1662, opened 2019-07-05
BUGFIX: Omit sessionless tokens from session

kdambekalns:
Without this fix, all security tokens – including those which are
implementations of SessionlessTokenInterface – are serialized and
added to the current session. This is a problem for sessionless
tokens, which need to be updated on every request on not just once
per session.

Fixes: #1666
Related: #1614

### DIFF
--- a/Neos.Flow/Classes/Security/Context.php
+++ b/Neos.Flow/Classes/Security/Context.php
@@ -787,10 +787,8 @@ class Context
         $tokensForSession = array_filter(array_merge($this->inactiveTokens, $this->activeTokens), static function (TokenInterface $token) {
             return !$token instanceof SessionlessTokenInterface;
         });
-        if ($tokensForSession !== []) {
-            $sessionDataContainer = $this->objectManager->get(SessionDataContainer::class);
-            $sessionDataContainer->setSecurityTokens($tokensForSession);
-        }
+        $sessionDataContainer = $this->objectManager->get(SessionDataContainer::class);
+        $sessionDataContainer->setSecurityTokens($tokensForSession);
     }
 
     /**
--- a/Neos.Flow/Classes/Security/Context.php
+++ b/Neos.Flow/Classes/Security/Context.php
@@ -16,6 +16,7 @@ use Neos\Cache\CacheAwareInterface;
 use Neos\Flow\Log\PsrSecurityLoggerInterface;
 use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
+use Neos\Flow\Security\Authentication\Token\SessionlessTokenInterface;
 use Neos\Flow\Security\Authentication\TokenAndProviderFactoryInterface;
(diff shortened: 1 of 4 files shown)


kdambekalns:
See https://github.com/flownative/openidconnect-client/blob/62e9b639108170821ce0728b1f4c7ac38a487218/Classes/Authentication/SessionlessTokenBugfixAspect.php

@kitsunet do you know if the previous (pre 5.2) implementation in the security context was affected as well?

kitsunet:
Yes, definitely, that would have been the `$token` property in `Security\Context` which was also not filtered.

kitsunet:
True ofc, but as @kdambekalns question suggests this fix should be made in earlier versions as well.

albe:
Yep, already had a quick look, but in pre 5.2 the tokens were stored in the `Security\Context` which was `@scope("session")` so, there's always a session around... would probably need to be filtered out in `shutdownObject()` (https://github.com/neos/flow-development-collection/blob/4.3/Neos.Flow/Classes/Security/Context.php#L785)

kdambekalns:
Discussed that with @bwaidelich today, and we agreed it would be better to check in the `SessionDataContainer` and throw an exception if a non-session token was passed in. And move the actual filtering to the "user side" of the code.

bwaidelich:
I adjusted the code as discussed and added bug ticket and some tests

Neos.Flow/Classes/Security/Context.php
```diff
@@ -783,8 +784,13 @@ protected function updateTokens(array $tokens)
             $token->updateCredentials($this->request);
         }
 
-        $sessionDataContainer = $this->objectManager->get(SessionDataContainer::class);
-        $sessionDataContainer->setSecurityTokens(array_merge($this->inactiveTokens, $this->activeTokens));
+        $tokensForSession = array_filter(array_merge($this->inactiveTokens, $this->activeTokens), static function(TokenInterface $token) {
+            return !$token instanceof SessionlessTokenInterface;
+        });
+        if ($tokensForSession !== []) {
```
albe: Just to make sure: if there are already tokens in the session data from previously but now no more tokens exist, the session data will no longer be overwrittten with that empty list. Is this potentially an issue?

bwaidelich: Good catch! I was thinking about this one, too - and then discarded the thought.. But maybe it's safer to set the empty array in this case?

What I was wondering (and didn't test yet): Would that start a session even though I don't want that if I only have sessionless tokens?

albe: That's funny, because that is the other taught I had, that I discarded, because a session is started anyway in different places (CSRF for example) 😄

Guess we'd need to have a "getSessionDataContainerIfSessionExists" method.

albe:
Let's play it safe for now and tackle the issue to not create sessions needlessly separately. I think this will be breaking and a much broader effort to get done correctly.

albe:
I'll merge this and do another round of upmerges of the 4.3 fix to this. Let's talk about the "no sessions" in a new issue.

PR #1663, opened 2019-07-05
BUGFIX: Omit sessionless tokens from session

albe:
Without this fix, all security tokens – including those which are
implementations of SessionlessTokenInterface – are serialized and
added to the current session. This is a problem for sessionless
tokens, which need to be updated on every request on not just once
per session.

Backport of #1662
Fixes: #1666


albe:
Note: this needs careful upmerging from 5.1 -> 5.2

albe:
Forgot to merge this first before doing upmerges, d'oh

PR #1665, opened 2019-07-07
TASK: Psalm fixes

albe:
This applies some fixes with psalm alter tool.
See https://psalm.dev/docs/manipulating_code/fixing/

Depends on #1658


sorenmalling:
@albe With all the updates to psalm-baseline.xml the last year, will you say that we can close this issue here?

Issue #1672, opened 2019-07-17
Database mapping issues with DateTimeInternface


markusguenther:
### Description

Today I tried to setup a new development environment for the 5.0 release.
So I cloned the repository and went through the setup.

Neos was not able to create a backend user. First thought it is maybe a setup tool issue but even on cli it is not possible. And when you try to test it with the build in server you also get errors.

### Steps to Reproduce

1. git clone https://github.com/neos/neos-development-distribution.git
2. cd neos-development-distribution && composer install
3. ./flow server run

#### Expected behavior

Server starts and you can setup neos.

#### Actual behavior

Exception:
<img width="900" alt="Bildschirmfoto 2019-07-17 um 11 08 28" src="https://user-images.githubusercontent.com/1014126/61364973-210c0d00-a887-11e9-8f42-0aa828d77bb3.png">

If you don`t use the flow server run and have a local nginx or so.
You will fail in generating users.

```
Failure while evaluating property annotations for class "Neos\ContentRepository\Domain\Model\NodeData": The attribute 'type' is required for the column description of property Neos\ContentRepository\Domain\Model\NodeData::$creationDateTime.
Type: Doctrine\ORM\Mapping\MappingException
  Code: 1382003497
  File:
Data/Temporary/Development/SubContextBeach/SubContextInstance/Cache/Code/Fl

ow_Object_Classes/Neos_Flow_Persistence_Doctrine_Mapping_Driver_FlowAnnotat
        ionDriver.php
  Line: 416

Nested exception:
The attribute 'type' is required for the column description of property
Neos\ContentRepository\Domain\Model\NodeData::$creationDateTime.

  Type: Doctrine\ORM\Mapping\MappingException
  File:
Packages/Libraries/doctrine/orm/lib/Doctrine/ORM/Mapping/MappingException.p
        hp
  Line: 360

  Type: Neos\Flow\Core\Booting\Exception\SubProcessException
  Code: 1355480641
  File: Packages/Framework/Neos.Flow/Classes/Core/Booting/Scripts.php
  Line: 703
```

### Affected Versions

Neos: master

Flow: master

markusguenther:
While chatting with @albe we maybe found already the related change.

https://github.com/neos/flow-development-collection/pull/1640/files#diff-18e35307a68fb093757fbd5382f9a641

markusguenther:
The one issue with the DataMapping has been fixed by @albe 
[...]

PR #1673, opened 2019-07-17
BUGFIX: Automatically map DateTimeInterface to ORM type

albe:
Since #1640 a model with a property annotated like

```php
/**
 * @var \DateTimeInterface
 */
protected $someDateTime;
```

lead to an exception in annotation parsing, because the property does not contain a `@ORM\Column(type="...")` mapping.
This worked before, because the type parser interpreted `DateTimeInterface` as just `DateTime` due to the error prone regex matching. Since the regex fix above, `DateTimeInterface` will now be parsed as `DateTimeInterface` and the FlowAnnotationDriver does not automatically map that to an ORM type. This change works around that, by automatically mapping `DateTimeInterface` to the ORM type `datetime` to be b/c with previous behaviour.

Note though, that this is generally bad magic, because you might want your property to be something different from `DateTime` after reconstitution from the ORM, so the real fix should be that you explicitly specify the DateTime (sub-)class in the `@var` annotation and add a corresponding `@ORM\Column(type="...")`, possibly together with a custom DBAL type.

Related to #1672

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
@@ -729,6 +729,7 @@ class FlowAnnotationDriver implements DoctrineMappingDriverInterface, PointcutFi
 
                 if (!isset($mapping['type'])) {
                     switch ($propertyMetaData['type']) {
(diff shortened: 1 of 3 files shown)


albe:
Only to master due to being a b/c fix to #1640, which only exists in master.

albe:
Cool, I'll provide a (unit) test this evening to make the fix complete.

albe:
Tests also pass. Gonna merge.

PR #1674, opened 2019-07-17
BUGFIX: Avoid unsupported operand types error

albe:
Under some circumstances, the session metadata cache might iterate a non-array value and the following logging attempt failing with an unsupported operand type error for the `+` array concatenation. This change works around this error, by checking the `$sessionInfo` to be of type array and assigning a wrapper array otherwise.

### DIFF
--- a/Neos.Flow/Tests/Unit/Session/SessionTest.php
+++ b/Neos.Flow/Tests/Unit/Session/SessionTest.php
@@ -1049,6 +1049,7 @@ class SessionTest extends UnitTestCase
         $this->inject($session, 'objectManager', $this->mockObjectManager);
         $this->inject($session, 'metaDataCache', $metaDataCache);
         $this->inject($session, 'storageCache', $storageCache);
+        $this->inject($session, 'logger', $this->createMock(LoggerInterface::class));
         $session->injectSettings($settings);
         $session->initializeObject();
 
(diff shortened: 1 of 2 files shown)


albe:
For anyone wondering why the `$sessionInfo['sessionIdentifier']` check doesn't error out before: PHP

https://3v4l.org/XWHpj

Yes, we could move the `!is_array($sessionInfo)` check up, but that wouldn't change anything really. The accessors would still just return `null`. The issue is only at the `+` operator, which is strict about supported operand types.

Neos.Flow/Tests/Unit/Session/SessionTest.php
```diff
@@ -1034,6 +1034,27 @@ public function autoExpireTriggersGarbageCollectionForExpiredSessions()
...
+        $settings = $this->settings;
+
+        $metaDataCache = $this->createCache('Meta');
+        $metaDataCache->set('foo', null);
+        $storageCache = $this->createCache('Storage');
+
+        $session = new Session();
+        $this->inject($session, 'objectManager', $this->mockObjectManager);
+        $this->inject($session, 'metaDataCache', $metaDataCache);
+        $this->inject($session, 'storageCache', $storageCache);
```
daniellienert:
```suggestion
        $this->inject($session, 'storageCache', $storageCache);
        $this->inject($session, 'logger', $this->createMock(LoggerInterface::class));
```

albe:
Thanks, tests pass now! Looks good to merge for me

albe:
@daniellienert please re-approve :)

PR #1676, opened 2019-07-19
TASK: Update documentation about AbstractConditionViewHelper.

DavidSporer:
I tried to create a custom IfViewHelper by extending the AbstractConditionViewHelper and noticed that it was still mentioning to overwrite the render function.
However the render function is not called but rather the evaluateCondition function must be overwritten.
I've basically taken the documentation from the Neos docs and copied it here and made some adjustments.

Let me know if this is ok or not (but current state of the documentation is not correct so it should be changed).

Thanks,
David

### DIFF
--- a/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Templating.rst
+++ b/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Templating.rst
@@ -675,36 +675,40 @@ This attribute is available in all tags that inherit from ``Neos\FluidAdaptor\Co
 AbstractConditionViewHelper
 ---------------------------
 
-If you want to build some kind of ``if/else`` condition, you should base the ViewHelper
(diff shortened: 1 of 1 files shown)


albe:
Awesome! Thanks a lot. Just one thing: can you check at which Typo3 fluid version that changed and then target this at a lower flow version?
If anything lower than Fluid 2.5 we can safely target this at flow 4.3

DavidSporer:
It's been there in Fluid 2.4 as it seems so it should work with Flow 4.3.
But I've just seen this commit: https://github.com/TYPO3/Fluid/commit/a67b31f9e6ecb015d0f47892fce46cf64110fd15

We should keep this in mind but it has been introduced in Fluid 2.6 if I have seen it correctly so that won't work with Flow 4.3 then.

albe:
Good catch.

DavidSporer:
I'll create a new Pull Request for 4.3

PR #1678, opened 2019-07-22
BUGFIX: resource:clean will remove resource from right collection

rkrahnen:
This fix makes sure `resource:clean` will remove broken resources from the right collection.

The problem is that you save the SHA1 for a broken resource. Now think about the following case: You have two resources with the same SHA1, but from two different collections (`persistent`/`temporary`). The resource inside the `temporary`-Collection was removed and now you run the command.

It will detect the missing resource from the `temporary`-Collection and add the SHA1 to `$brokenResources`. Now when iteration over `$brokenResources` to get the `PersistentResource` you are using `$this->resourceRepository->findOneBySha1($resourceSha1)`, which ignores the collection. So you can't be sure to get the `PersistentResource` from the `temporary`-Collection that you actually want, it's possible that you get the one from the `persistent`-Collection. This would result in deleting the wrong `PersistentResource` and not removing the broken resource but creating a new problem.

The fix just saves the identifier of the `PersistentResource` to `$brokenResources` and later detects the correct one agin by using `$this->resourceRepository->findByIdentifier($resourceIdentifier)`.

### DIFF
--- a/Neos.Flow/Classes/Command/ResourceCommandController.php
+++ b/Neos.Flow/Classes/Command/ResourceCommandController.php
@@ -222,7 +222,7 @@ class ResourceCommandController extends CommandController
             /* @var PersistentResource $resource */
             $stream = $resource->getStream();
             if (!is_resource($stream)) {
-                $brokenResources[] = $resource->getSha1();
+                $brokenResources[] = $this->persistenceManager->getIdentifierByObject($resource);
             }
         }
 
@@ -235,8 +235,8 @@ class ResourceCommandController extends CommandController
             /* @var ThumbnailRepository $thumbnailRepository */
             $thumbnailRepository = $this->objectManager->get(ThumbnailRepository::class);
 
-            foreach ($brokenResources as $key => $resourceSha1) {
-                $resource = $this->resourceRepository->findOneBySha1($resourceSha1);
+            foreach ($brokenResources as $key => $resourceIdentifier) {
+                $resource = $this->resourceRepository->findByIdentifier($resourceIdentifier);
                 $brokenResources[$key] = $resource;
                 $assets = $assetRepository->findByResource($resource);
                 if ($assets !== null) {
--- a/Neos.Flow/Classes/Command/ResourceCommandController.php
+++ b/Neos.Flow/Classes/Command/ResourceCommandController.php
@@ -237,9 +237,9 @@ class ResourceCommandController extends CommandController
                 $thumbnailRepository = $this->objectManager->get(ThumbnailRepository::class);
             }
 
-            foreach ($brokenResources as $resourceIdentifier) {
+            foreach ($brokenResources as $key => $resourceIdentifier) {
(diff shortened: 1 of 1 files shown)


rkrahnen:
While reviewing my changes I've detected further bugs:

1. I removed the misleading key `$sha1` as the array `$brokenResources` just has numeric keys.
2. If `Neos.Media` is not present, getting the `PersistentResource` is completely skipped, but it's needed later on to generate the output and delete it from the database.

The whole command could be optimized by using just one if condition for `count($brokenResources) > 0` and only iterating once over `$brokenResources` before asking the user what to do, but I'll wait for your response to the current changes first.

albe:
Good catch - not too deep into the ResourceManagement part, but will take a closer look asap

Neos.Flow/Classes/Command/ResourceCommandController.php
```diff
@@ -222,29 +222,33 @@ public function cleanCommand()
...
+            if ($mediaPackagePresent) {
+                /* @var AssetRepository $assetRepository */
+                $assetRepository = $this->objectManager->get(AssetRepository::class);
+                /* @var ThumbnailRepository $thumbnailRepository */
+                $thumbnailRepository = $this->objectManager->get(ThumbnailRepository::class);
+            }
 
-            foreach ($brokenResources as $key => $resourceSha1) {
-                $resource = $this->resourceRepository->findOneBySha1($resourceSha1);
+            foreach ($brokenResources as $key => $resourceIdentifier) {
```
kdambekalns:
```suggestion
            foreach ($brokenResources as $resourceIdentifier) {
```

Neos.Flow/Classes/Command/ResourceCommandController.php
```diff
@@ -222,29 +222,33 @@ public function cleanCommand()
...
+                $resource = $this->resourceRepository->findByIdentifier($resourceIdentifier);
                 $brokenResources[$key] = $resource;
```
kdambekalns:
```suggestion
                $brokenResources[] = $resource;
```

rkrahnen:
Sorry for the two commits, I should have checked your suggestion more carefully.
The foreach key is used to replace the existing entry (`$resourceIdentifier`) with the resource (`$resource`) that is needed in the following iterations over `$brokenResources`, so we have to leave that foreach key.

Maybe it would be better to not save the resource inside the array for memory reasons, but then we would have to call `$this->resourceRepository->findByIdentifier($resourceIdentifier)` in every foreach iterating over `$brokenResources`, which is three times in this function.

For now I would suggest to leave it as is and first get the bug fixed.

kdambekalns:
> The foreach key is used to replace the existing entry (`$resourceIdentifier`) with the resource (`$resource`) that is needed in the following iterations over `$brokenResources`, so we have to leave that foreach key.

Ok, my fault, sorry. I didn't see it used _after_ that loop, but it's needed to overwrite the correct entry from _before_… now I got it. 👍

PR #1681, opened 2019-07-26
BUGFIX: EntityPrivilege fails on empty IN condition

hhoechtl:
**What I did**
Fixed the behavior of the `\Neos\Flow\Security\Authorization\Privilege\Entity\Doctrine\EntityPrivilege` if you execute an `in` operation on an empty array.

**How to verify it**
```yaml
matcher: 'isType("Neos\Neos\Domain\Model\Site") && !property("nodeName").in("context.userInformationContext.siteNodeNames")'
```

If `getSiteNodeNames()` evaluates to `[]` which is possible and valid, the PropertyGenerator tries to to `getParameter()` on this value which can only be applied to scalar values.

This is caused by only checking on an exact `null`, but an empty array should also be ignored.

### DIFF
--- a/Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/PropertyConditionGenerator.php
+++ b/Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/PropertyConditionGenerator.php
@@ -482,7 +482,7 @@ class PropertyConditionGenerator implements SqlGeneratorInterface
                     }
                 }
                 $parameter = implode(',', $parameters);
-            } elseif ($this->getRawParameterValue($operandDefinition) !== null) {
+            } elseif ($this->getRawParameterValue($operandDefinition) !== null && $this->getRawParameterValue($operandDefinition) !== []) {
                 $parameter = $sqlFilter->getParameter($operandDefinition);
             }
         } catch (\InvalidArgumentException $exception) {
--- a/Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/PropertyConditionGenerator.php
+++ b/Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/PropertyConditionGenerator.php
@@ -482,7 +482,7 @@ class PropertyConditionGenerator implements SqlGeneratorInterface
                     }
                 }
                 $parameter = implode(',', $parameters);
-            } elseif ($this->getRawParameterValue($operandDefinition) !== null && $this->getRawParameterValue($operandDefinition) !== []) {
+            } elseif (!($this->getRawParameterValue($operandDefinition) === null || ($this->operator === 'in' && $this->getRawParameterValue($operandDefinition) === []))) {
                 $parameter = $sqlFilter->getParameter($operandDefinition);
             }
         } catch (\InvalidArgumentException $exception) {


Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/PropertyConditionGenerator.php
```diff
@@ -482,7 +482,7 @@ protected function getConstraintStringForSimpleProperty(SQLFilter $sqlFilter, $p
                     }
                 }
                 $parameter = implode(',', $parameters);
-            } elseif ($this->getRawParameterValue($operandDefinition) !== null) {
+            } elseif ($this->getRawParameterValue($operandDefinition) !== null && $this->getRawParameterValue($operandDefinition) !== []) {
```
albe: Hm, tbh I don't really understand this check. The `rawParameterValue` is checked, but then the `$operandDefinition` is given to the SqlFilter, which only accepts a string with `getParameter` - so shouldn't the check be `is_string($operandDefinition)`?

hhoechtl: I'm not quite sure. The code is pretty old and there aren't any tests for this class. I just added something quite specific that fixes my use case that hopefully doesn't break anything else.

albe: Yep, that's what I'm unsure about, because that affects more cases than just the `in` operator I can't even follow the existing logic. Maybe @kdambekalns or @kitsunet knows a bit more about this logic?

albe:
```suggestion
            } elseif ($this->getRawParameterValue($operandDefinition) !== null && ($this->operator !== 'in' || $this->getRawParameterValue($operandDefinition) !== [])) {
```
I think this is how it needs to be to really only affect the `in` operation, where an empty array as operand will not work as expected. Now, this is really hard to read though. Some more code style love needed desperately.

albe:
Okay, so after reading into this again (and also https://stackoverflow.com/questions/13210233/empty-in-clause-parameter-list-in-mysql), I think I get the issue a bit better and also how our ConditionGenerator works.
So at minimum, since the `[]` is invalid operand only affects the `in` operation, this needs to be restricted further. See my other comment on the code section in question.

Would like to get this in soon though, but needs to be rebased on 5.3 first - should definitely land in lowest maintained branch as a BUGFIX.

Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/PropertyConditionGenerator.php
albe:
```suggestion
            } elseif (!($this->getRawParameterValue($operandDefinition) === null || ($this->operator === 'in' && $this->getRawParameterValue($operandDefinition) === []))) {
```
That's probably better, negating every single condition is hard to grasp.

albe:
Has been long enough. Rebased in #1924, will merge now and we'll see. I think with the constraint on only the `in` operator, this is scoped enough to not cause issues. Thanks a lot @hhoechtl and sorry it took so long to finally grok through this!

Issue #1215, opened 2018-02-15
BaseUri override in Settings doesn't resprect path


nixahnung:
If i override the baseUri in http:baseuri setting it seems that an additional path is not respected,

Example: baseUri: http://example.com/path/...

Resolved Uri by UriBuilder: http://example.com/...
Path is missing.

Flow: 4.3.3

sorenmalling:
Do you install Flow inside a subdirectory or do you want the `/path` as a part of your uri ?

bwaidelich:
We had to revert the original fix (#1682) because it introduced a regression. see #1841

albe:
This is fixed in 6.0+ as of https://github.com/neos/flow-development-collection/commit/c37b556c2d93b5f3e038eef9fe8663c8fe72837f - now the only question is if we should go the effort to also create a fix for 5.3, as that is the current active bugfix release version. The effort would be much bigger, since the fix can not be simply backported. So I'd be inclined to close this as is.

Any oppinions @bwaidelich @nlx-sascha @nlx-lars @skurfuerst ?

skurfuerst:
Agreed @albe 👍

nlx-lars:
@albe yes, this can be closed in my opinion 👍

nlx-sascha:
@albe fine for me too

bwaidelich:
Same here, thanks for taking care. but since @nixahnung opened this ticket: are you fine with this, too - or rather: could you solve this otherwise somehow (duh, this is more than 2 years old already?!)

BTW: I just had a laugh that "nix-lars"¹ and "nix-sascha" comment on an issue from "nixahnung". Are you guys from Gaul? :)


¹ ok, I know that it is "n*L*x", but I read it differently at first

albe:
Closing with the above agreement :)

PR #1682, opened 2019-07-30
BUGFIX: Respect Neos.Flow.http.baseUri path in UriBuilder

paxuclus:
If Neos.Flow.http.baseUri contains a path, it was not respected
during uri building.

See: #1185
Resolves: #1215

**What I did**
Add path of `Neos.Flow.http.baseUri` to ResolveContext's uriPathPrefix.

**How to verify it**
Configure `Neos.Flow.http.baseUri` to be an absolute URI with path, build URI to any Controller.

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
@@ -62,6 +62,15 @@ final class ResolveContext
         $this->routeValues = $routeValues;
         $this->forceAbsoluteUri = $forceAbsoluteUri;
         $this->uriPathPrefix = $uriPathPrefix;
+
+        // Only add base uri path for absolute uri, in case of relative uri the uri has to be relative to the given base uri
+        if ($forceAbsoluteUri) {
+            $this->uriPathPrefix = '/' . ltrim($this->uriPathPrefix, '/');
+
+            if ($baseUri->getPath() !== '') {
+                $this->uriPathPrefix = rtrim($baseUri->getPath(), '/') . $this->uriPathPrefix;
+            }
+        }
     }
 
     /**
--- a/Neos.Flow/Tests/Functional/Mvc/RoutingTest.php
+++ b/Neos.Flow/Tests/Functional/Mvc/RoutingTest.php
@@ -361,6 +361,114 @@ class RoutingTest extends FunctionalTestCase
     /**
      * @test
      */
+    public function uriPathPrefixIsRespectedInRoute()
+    {
+        $routeValues = [
+            '@package' => 'Neos.Flow',
+            '@subpackage' => 'Tests\Functional\Http\Fixtures',
+            '@controller' => 'Foo',
+            '@action' => 'index',
+            '@format' => 'html'
+        ];
+        $baseUri = new Uri('http://localhost');
+        $actualResult = $this->router->resolve(new ResolveContext($baseUri, $routeValues, false, 'index.php/'));
+
+        $this->assertSame('index.php/neos/flow/test/http/foo', (string)$actualResult);
+    }
+
+    /**
+     * @test
+     */
+    public function uriPathPrefixIsRespectedInAbsoluteRoute()
+    {
+        $routeValues = [
+            '@package' => 'Neos.Flow',
+            '@subpackage' => 'Tests\Functional\Http\Fixtures',
+            '@controller' => 'Foo',
+            '@action' => 'index',
+            '@format' => 'html'
+        ];
+        $baseUri = new Uri('http://localhost');
+        $actualResult = $this->router->resolve(new ResolveContext($baseUri, $routeValues, true, 'index.php/'));
+
+        $this->assertSame('http://localhost/index.php/neos/flow/test/http/foo', (string)$actualResult);
+    }
+
+    /**
+     * @test
+     */
+    public function pathOfBaseUriIsRespectedInAbsoluteRoute()
+    {
+        $routeValues = [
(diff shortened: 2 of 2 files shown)


albe (approved):
Looks good by reading, but would like to hear Bastian if there's something non-obvious to be considered.
e.g. is the `$uriPathPrefix` somehow supposed to already contain the baseUri path before construction?

nlx-lars:
@albe maybe https://github.com/neos/flow-development-collection/blob/20e8b51810677ff2e5a18107fed35e56b1eacc8b/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php#L352 might be a better place?

I'm also not sure if `Files::concatenatePaths()` and `Files::getUnixStylePath` should be used here as we're dealing with uris, not files. But I couldn't find any equivalent helper function for uris.

Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
```diff
@@ -62,6 +63,14 @@ public function __construct(UriInterface $baseUri, array $routeValues, bool $for
...
         $this->forceAbsoluteUri = $forceAbsoluteUri;
         $this->uriPathPrefix = $uriPathPrefix;
+
+        // Only add base uri path for absolute uri, in case of relative uri the uri has to be relative to the given base uri
+        if ($forceAbsoluteUri && $baseUri->getPath()) {
+            $this->uriPathPrefix =  Files::getUnixStylePath(Files::concatenatePaths([$baseUri->getPath(), $uriPathPrefix])) . '/';
+        }
+        if ($forceAbsoluteUri) {
+            $this->uriPathPrefix = '/' . ltrim($this->uriPathPrefix, '/');
+        }
```
bwaidelich: You could simplify the two conditions into sth like
```php
if ($forceAbsoluteUri) {
    $this->uriPathPrefix = '/' . ltrim($this->uriPathPrefix, '/');
    if ($baseUri->getPath() !== '') {
        $this->uriPathPrefix = rtrim($baseUri->getPath(), '\/') . $this->uriPathPrefix;
    }
}
```
(untested)

Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
```diff
@@ -62,6 +63,14 @@ public function __construct(UriInterface $baseUri, array $routeValues, bool $for
         $this->routeValues = $routeValues;
         $this->forceAbsoluteUri = $forceAbsoluteUri;
         $this->uriPathPrefix = $uriPathPrefix;
+
+        // Only add base uri path for absolute uri, in case of relative uri the uri has to be relative to the given base uri
+        if ($forceAbsoluteUri && $baseUri->getPath()) {
```
bwaidelich: `$baseUri->getPath()` is not a boolean (see https://neos.readthedocs.io/en/stable/References/CodingGuideLines/PHP.html#php-in-general => "Prefer strict comparisons whenever possible")

bwaidelich (commented):
I'll have to look into it again, but I think this change makes sense. Thanks a lot.
I just added some comments regarding CGL

daniellienert (changes requested):
Hey @nlx-lars, thanks for the PR.
Please have a look at @bwaidelich's comments.

Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
nlx-lars: Done. Works as expected

nlx-lars:
Hey @daniellienert and @bwaidelich,

PR has been adjusted and tested. Still works as expected. 👍

albe:
@daniellienert can you remove your veto?

daniellienert (approved):

albe:
This might have caused a regression - see https://github.com/neos/flow-development-collection/issues/1797

Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
```diff
@@ -62,6 +62,15 @@ public function __construct(UriInterface $baseUri, array $routeValues, bool $for
         $this->routeValues = $routeValues;
         $this->forceAbsoluteUri = $forceAbsoluteUri;
         $this->uriPathPrefix = $uriPathPrefix;
+
+        // Only add base uri path for absolute uri, in case of relative uri the uri has to be relative to the given base uri
+        if ($forceAbsoluteUri) {
+            $this->uriPathPrefix = '/' . ltrim($this->uriPathPrefix, '/');
+
+            if ($baseUri->getPath() !== '') {
+                $this->uriPathPrefix = rtrim($baseUri->getPath(), '/') . $this->uriPathPrefix;
```
kitsunet: Ok, so I don't understand this really and think we should revert this whole change and rethink what happens. This was a simple value object before, now it does some internal magic, that doesn't line up with what happens in `\Neos\Flow\Mvc\Routing\Router::resolve` because there the uriPathPrefix is applied to the baseUri anyway, which will result in duplication in `\Neos\Flow\Mvc\Routing\Dto\UriConstraints::applyTo`. The tests do not expose this (yet) because the failing scenario is an installation in a subdirectory of the webhost, because that will have a baseUri with the subdirectory + a script request path (which then becomes a uriPathPrefix) which is the subdirectory. So that subdirectory gets preprended to the path twice. We should make sure to solve this here in another place.

kitsunet:
Definitely causes that regression as described above, revert here: https://github.com/neos/flow-development-collection/pull/1841

PR #1683, opened 2019-08-01
BUGFIX: Allow a single '*' entry in trustedProxies

albe:
This makes the setting `Neos.Flow.http.trustedProxies.proxies` behave equal for a setting of
`"*"` or `["*"]` or `- "*"`.

### DIFF
--- a/Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
+++ b/Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
@@ -44,7 +44,7 @@ class TrustedProxiesComponent implements ComponentInterface
         if ($this->settings['proxies'] === ['*']) {
             $this->settings['proxies'] = '*';
         }
-        if (!is_array($this->settings['proxies']) && $this->settings['proxies'] !== '*')) {
+        if (!is_array($this->settings['proxies']) && $this->settings['proxies'] !== '*') {
             throw new InvalidConfigurationException('The Neos.Flow.http.trustedProxies.proxies setting may only be an array of IP addresses or the single string "*". Got "' . var_export($this->settings['proxies'], true) . '" instead.', 1564659249);
         }
         if (is_array($this->settings['proxies']) && array_search('*', $this->settings['proxies'], true) !== false) {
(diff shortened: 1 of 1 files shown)


robertlemke:
This bugfix seems to be a regression for CIDR based configuration. I used the following string so far:

```
10.0.0.0/8
```

That string does not contain an asterisk (*) and therefore will be rejected by the new code. However, as far as I can see, this being a valid CIDR should work just fine and is supported by the `IpUtility::cidrMatch()`  function used in the code.

albe:
Hmm, indeed. The setting was always supposed to be an array (of IPs or CIDR ranges) and only a single string in the special case of `*`, but I see how that's a valid use-case and makes this breaking. I'll lead with a b/c follow-up.

robertlemke:
The variable names are a bit confusing, too. I would not expect that `$trustedIpHeader` is an array of IP addresses.

Anyway, [the documentation](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Http.html#trusted-proxies) clearly states that this setting value should be an array, so I'll better update our configuration.

robertlemke:
The reason why I used a string there is that I set the trusted proxies CIDR via the official environment variable `FLOW_HTTP_TRUSTED_PROXIES`. How would we support array for these cases?

robertlemke:
image

The documentation even states, that a string with comma-separated values is supported.

kdambekalns:
Gnah, during review I relied on the settings file, which at some point was declared the official source of truth for all things settings:

Screenshot 2019-09-03 at 08 30 15

kdambekalns:
Basically that change needs to take into account, that since 5.0.0 this value can be a string (as of https://github.com/neos/flow-development-collection/pull/1273).

It might be a good idea, to pull [that code](https://github.com/neos/flow-development-collection/blob/1894dfb5e485fd285fb0a7a2ffe165dba94e250c/Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php#L210-L215) into `injectSettings()`…

albe:
Indeed. Very good catch, I totally forgot that CSV style (also looking at 4.3 code base). Will take care!

robertlemke:
Thanks @albe !

PR #1685, opened 2019-08-06
CLEANUP: Use var_dump return parameter

j6s:
**What I did**
When digging through the code I found this instance of capturing the output of `\Neos\Flow\var_dump` using `ob_get_contents` when `\Neos\Flow\var_dump` has a `$return` parameter itself.

Note: I branched off of `4.0` because I thought that that version was still maintained. Turns out it is not and 5.0 is the lowest maintained version as of this month.

**How I did it**
Using the `$return` parameter of `\Neos\Flow\var_dump`

**Checklist**


j6s:
Closing this PR, reopening it based on the correct source branch. Sorry for the inconvinience.

PR #1686, opened 2019-08-06
TASK: Use var_dump return parameter

j6s:
**What I did**
When digging through the code I found this instance of capturing the output of `\Neos\Flow\var_dump` using `ob_get_contents` when `\Neos\Flow\var_dump` has a `$return` parameter itself.

**How I did it**
Using the `$return` parameter of `\Neos\Flow\var_dump`


albe:
Lowest maintained is 4.3 as of now, but since this is just a cleanup I wouldn't mind having it in 5.0+ only. Thanks for this! Makes totally sense!

albe:
@j6s I saw on slack that you looked into backporting that for 4.3? Are you still up to that? If so, I'd wait for that PR instead to prevent a needless merge conflict when we do upmerges. As said, I'm fine with having that in 5.0+ only

PR #1687, opened 2019-08-08
TASK: Fix some findings of the psalm static code analysis

daniellienert:
See: #1658

### DIFF
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -579,7 +579,7 @@ class StringHelper implements ProtectedContextAwareInterface
      */
     public function htmlSpecialChars($string, $preserveEntities = false)
     {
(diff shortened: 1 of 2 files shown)


Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -579,7 +579,7 @@ public function rawUrlDecode($string)
      */
     public function htmlSpecialChars($string, $preserveEntities = false)
     {
-        return htmlspecialchars((string)$string, null, null, !$preserveEntities);
+        return htmlspecialchars((string)$string, ENT_COMPAT | ENT_HTML401, 'UTF-8', !$preserveEntities);
```
albe: I think we completely overrode the defaults with `null` (=0) beforehand, i.e. ENT_NOQUOTES|ENT_HTML401 (==0)

albe:
```suggestion
        return htmlspecialchars((string)$string, ENT_NOQUOTES | ENT_HTML401, ini_get("default_charset"), !$preserveEntities);
```

PR #1689, opened 2019-08-09
!!! TASK: Enable `subdivideHashPathSegment` and `relativeSymlinks` by default

mficzel:
The old defaults for these settings worked but caused trouble once projects got bigger over time:
- `subdivideHashPathSegment: false` caused having too many symlinks in a single folder for many filesystems 
- `relativeSymlinks: false` did not allow to put the `Web/_Resources` directory into the shared folder for faster deployments

ATTENTION: This alters the default behavior and the published resources will get a url with nested pathes. That is why this is considered a breaking change.

NOTE: After updating you have to empty the `Web/_Resources/Persistent` folder and run `./flow resource:publish`. This us usually all done automatically from the deployment tool you are using.

If you do not want this behavior you can disable the subdivision via configuration for your project with the following configuration.

```
Neos:
  Flow:
    resource:
      targets:
        localWebDirectoryPersistentResourcesTarget:
          targetOptions:
            subdivideHashPathSegment: false
            relativeSymlinks: false
``` 
You probably want to redirect requests to the old urls in your webserver configuration, the following regex search/replace patterns may be used for that:
- searchPattern: `^_Resources\/Persistent\/([a-f0-9]{1})([a-f0-9]{1})([a-f0-9]{1})([a-f0-9]{1})([a-f0-9]{36})\/(.+)$`
- replacePattern: `_Resources/Persistent/$1/$2/$3/$4/$1$2$3$4$5/$6`

### DIFF
--- a/Neos.Flow/Configuration/Settings.Resource.yaml
+++ b/Neos.Flow/Configuration/Settings.Resource.yaml
@@ -76,7 +76,7 @@ Neos:
             extensionBlacklist: *uploadExtensionBlacklist
 
             # If the generated URI path segment containing the sha1 should be divided into multiple segments (recommended if you expect many resources):
-            subdivideHashPathSegment: false
+            subdivideHashPathSegment: true
 
             # If the symlinks should be relative instead of absolute
-            #relativeSymlinks: false
+            relativeSymlinks: true


mficzel:
@robertlemke maybe you have a specific opinion on this change?

mficzel:
@daniellienert included `relativeSymlinks` into the pr and adjusted the description.

albe:
Quickly checked the `relativeSymlinks` setting with windows, as I wasn't sure, but it seems to work. So okay from my side, though two separate PRs would have been a bit nicer ;)

daniellienert:
I still think we need some guidance how to configure webservers to redirect old asset URIs to the new ones. The SEO guys will be very unhappy if all the documents suddenly throw a 404.

mficzel:
@daniellienert i have no example for the resource redirect at hand, if you have one please adjust the description directly.

kdambekalns:
TBH, I don't understand the `$result = preg_replace(…)` example in the commit message. 😬

mficzel:
> I don't understand the $result = preg_replace(…) example
@kdambekalns that is my take on showing how a redirect regex might look like. An nginx and .htaccess snippet would be much better.

kdambekalns:
> that is my take on showing how a redirect regex might look like

Ok, but that expression is invalid and seems wrong anyway. What you need is rather this, no?

Take

    ^_Resources\/Persistent\/([a-f0-9]{8})([a-f0-9]{8})([a-f0-9]{8})([a-f0-9]{8})([a-f0-9]{8})\/(.+)$

and replace with

    /_Resources/Persistent/$1/$2/$3/$4/$1$2$3$4$5

(see https://regex101.com/r/DgKfhN/1 for a "live test")

mficzel:
@kdambekalns your pattern was also wrong but i finally endet up here https://regex101.com/r/DgKfhN/3 and adjusted the comment above.

- search: `^_Resources\/Persistent\/([a-f0-9]{1})([a-f0-9]{1})([a-f0-9]{1})([a-f0-9]{1})([a-f0-9]{36})\/(.+)$`
- replace: `_Resources/Persistent/$1/$2/$3/$4/$1$2$3$4$5/$6`

kdambekalns:
Doh, yes. Wrong assumption about the subdivided URL on my side!

Looks better now: https://regex101.com/r/DgKfhN/5

mficzel:
@daniellienert do the regex examples at the end of the description cover your concerns? If yes this can imho be merged.

Issue #1707, opened 2019-08-19
Logs are not separated for contexts (Development / Production / Testing)


daniellienert:
On current master, the System.log is used in all contexts.

PR #1708, opened 2019-08-19
BUGFIX: Logs properly split per application context

kitsunet:
Fixes: #1707


Neos.Flow/Configuration/Development/Settings.yaml
```diff
@@ -43,3 +43,29 @@ Neos:
         backendOptions:
```
daniellienert: Shouldn't these old settings be removed now?

kitsunet: Duh, ofc 😆 Did that now.

PR #1710, opened 2019-08-20
TASK: Add type constraints to logger backend classes

daniellienert:


daniellienert:
Mhmh, seems eval does "things" with the code. I'll take care

Neos.Flow.Log/Tests/Unit/Backend/AbstractBackendTest.php
```diff
@@ -32,9 +32,9 @@ protected function setUp(): void
         $this->backendClassName = 'ConcreteBackend_' . md5(uniqid(mt_rand(), true));
         eval('
 			class ' . $this->backendClassName . ' extends \Neos\Flow\Log\Backend\AbstractBackend {
```
kitsunet: interesting piece of code, maybe we should replace it with a test implementation of the abstract backend at some point.

Neos.Flow.Log/Classes/Backend/FileBackend.php
```diff
@@ -237,11 +237,11 @@ public function append(string $message, int $severity = LOG_INFO, $additionalDat
...
             $processId = ' ' . str_pad((string)posix_getpid(), 10);
         } else {
-            $processId = ' ';
+            $processId = ' ' . str_pad((string)getmypid(), 10);
         }
-        $ipAddress = ($this->logIpAddress === true) ? str_pad((isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : ''), 15) : '';
-        $severityLabel = (isset($this->severityLabels[$severity])) ? $this->severityLabels[$severity] : 'UNKNOWN  ';
-        $output = strftime('%y-%m-%d %H:%M:%S', time()) . $processId . ' ' . $ipAddress . $severityLabel . ' ' . str_pad($packageKey, 20) . ' ' . $message;
+        $ipAddress = ($this->logIpAddress === true) ? str_pad(($_SERVER['REMOTE_ADDR'] ?? ''), 15) : '';
+        $severityLabel = $this->severityLabels[$severity] ?? 'UNKNOWN  ';
```
kitsunet: I just realise, (nothing new but hey) why do we use fixed width for everything BUT the severity label? That defeats the whole purpose...

albe: They are already padded in the static array: https://github.com/neos/flow-development-collection/pull/1710/files/ebe364cda9f814caf56b3f7e46f3254bdffbcf57..ceae15e0ebdfd3ff794c25865beb483319a568f0#diff-2a09d9b0f2f6c2fac5280639206a0a79R143

kitsunet: Duh, I see. Fine, fine

PR #1716, opened 2019-08-20
FEATURE: Add `getBaseUri()` getter to ActionRequest

albe:
Resolves neos/neos-development-collection#2636

### DIFF
--- a/Neos.Flow/Classes/Mvc/ActionRequest.php
+++ b/Neos.Flow/Classes/Mvc/ActionRequest.php
@@ -12,7 +12,9 @@ namespace Neos\Flow\Mvc;
  */
 
 use Neos\Flow\Annotations as Flow;
+use Psr\Http\Message\UriInterface;
 use Psr\Http\Message\ServerRequestInterface as HttpRequestInterface;
+use Neos\Flow\Http\ServerRequestAttributes;
 use Neos\Flow\ObjectManagement\Exception\UnknownObjectException;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 use Neos\Flow\Package\PackageManager;
@@ -693,6 +695,17 @@ class ActionRequest implements RequestInterface
     public function getFormat(): string
     {
         return $this->format;
+    }
+
+    /**
+     * Returns the base Uri of the request.
+     *
+     * @return UriInterface The base URI of the request
+     * @api
+     */
+    public function getBaseUri(): UriInterface
+    {
+        return $this->getHttpRequest()->getAttribute(ServerRequestAttributes::BASE_URI);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


albe:
Related to #1552

Neos.Flow/Classes/Mvc/ActionRequest.php
```diff
@@ -695,6 +697,17 @@ public function getFormat(): string
...
 
+    /**
+     * Returns the base Uri of the request.
+     *
+     * @return UriInterface The base URI of the request
+     * @api
+     */
+    public function getBaseUri(): UriInterface
+    {
+        return $this->getHttpRequest()->getAttribute(ServerRequestAttributes::BASE_URI);
```
kitsunet: That still means we rely on that attribute. I would very much like to get rid of it at some point.

albe: Generally yes, but that's not easy, as there are currently a couple of places that only have access to the Http Request and still use the baseUri, e.g. the WebRedirect or FileSystemTarget

kitsunet: Sure, I think we should discuss this in more detail with basti as well, IMHO we should figure out the intend in various places and adjust accordingly. Eg. FileSystemTarget is weird anyway. And components could decide between setting and building a baseUri from the request on their own (or with an additional indpeendent helper/service)

bwaidelich: Like it's described right now I'd expect this method to return the base URI of the current HTTP request. But it might not if that is different from the configured `Neos.Flow.http.baseUri`.

albe: I'd expect that probably in the Http Request, but in the ActionRequest, I expect the baseUri that would be used for any rendering of a link for example (which is either the baseUri of the request, or can be overridden by the configuration), so that's exactly what it does. But sure, let's have a chat about that.

daniellienert:
@bwaidelich / @kitsunet / @albe - would yo mind if I fix the redirect service with `$this->getHttpRequest()->getAttribute(ServerRequestAttributes::BASE_URI);`, so I could go one with the tests?

kitsunet:
@daniellienert no, please do! Given that we use the getAttribute all over core and Neos and basically everywhere right now... I jusut want to decide on a path for the future.

albe:
We discussed this today and decided that we want to get rid of the `getBaseUri()` method alltogether. The main use case is inside the UriBuilder and that will receive the logic needed to stay b/c - for all other cases, it is debatable what "BaseURI" even means exactly and then whether the http setting should override it. Hence if it is needed, you should use `ResponseInformationHelper::generateBaseUri(ServerRequestInterface $request)` to generate a baseUri from the current http request, and/or inject the Configuration setting explicitly, if you want that to affect your use case.

Closing this therefore.

PR #1718, opened 2019-08-21
!!! FEATURE: ValueObjects are embedded by default

### LINKED ISSUE #2123: Make ValueObjects embedded by default
Embedded value objects are the preferred storage method for all value objects, since it better reflects true value object semantics.

TBD:
- Is this really viable? (I think so)
- Should we provide a code migration that flips all ValueObjects that are not explicitly embedded=false in order to increase b/c?

See https://github.com/neos/flow-development-collection/pull/1718

albe:
This makes all ValueObjects embedded by default. Embedded value objects are the preferred storage method for all value objects, since it better reflects true value object semantics.
This requires a schema update, so you need to generate a migration for your packages and apply it. Alternatively you can run the code migration provided with this change or manually change all your `@Flow\ValueObject` annotations to `@Flow\ValueObject(embedded=false)` in order to keep your current database schema.

TODO:
- [x] Add migration to make existing ValueObjects without explicit embedding attribute `embedded=false`
- [ ] Update documentation to make clear that ValueObjects are embedded by default

Resolves #2123


albe:
TBD:
- Is this really viable? (I think so)
- Should we provide a code migration that flips all ValueObjects that are not explicitly `embedded=false` in order to increase b/c?

sorenmalling:
@albe We got this exception in last test run

```
1) Neos\Flow\Tests\Functional\Http\Client\BrowserTest::argumentsAreSentAsRequestBodyEvenInGetRequest

Doctrine\ORM\ORMException: Column name `persistence_object_identifier` referenced for relation from Neos\Flow\Tests\Functional\Persistence\Fixtures\Post towards Neos\Flow\Tests\Functional\Persistence\Fixtures\TestValueObject does not exist.
```

when relating to a valueobject. Is this a usecase we can handle in the annotationdriver?

albe:
Yeah, basically need to run the migration on Flow itself :) Will do so this evening or tomorrow

albe:
Adding docs next

albe:
@bwaidelich can we merge this for feature freeze and I promise to add the missing docs before release?

bwaidelich (approved):
Makes sense!

kdambekalns:
Wait, what about

> Adding docs next

sorenmalling:
@kdambekalns 
> @bwaidelich can we merge this for feature freeze and I promise to add the missing docs before release?

albe:
TODO:
- [x] Add migration to make existing ValueObjects without explicit embedding attribute `embedded=false`
- [x] Update documentation to make clear that ValueObjects are embedded by default (#2255)

PR #1719, opened 2019-08-21
TASK: Make use of `hasTitle` method to check message title existence

albe:
See https://github.com/neos/flow-development-collection/pull/1046


albe:
Found a major reason why I don't like merging into instead of rebasing... do'h... how do I find the inital commits to reset to?

PR #1720, opened 2019-08-21
!!! FEATURE: Add strict typehints for error message title and code

albe:
This is breaking, because the Error messages no longer accept `null` for the `$code` and `$title` constructor arguments.

See https://github.com/neos/flow-development-collection/pull/1046 for the PR that could not yet fix this issue.

### DIFF
--- a/Neos.Error.Messages/Classes/Message.php
+++ b/Neos.Error.Messages/Classes/Message.php
@@ -57,15 +57,15 @@ class Message
      * Constructs this error
      *
      * @param string $message An english error message which is used if no other error message can be resolved
-     * @param integer $code A unique error code
+     * @param integer|null $code A unique error code
      * @param array $arguments Array of arguments to be replaced in message
      * @param string $title optional title for the message
      * @api
      */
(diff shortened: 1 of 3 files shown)


kdambekalns:
But it seems the tests need some love, returning NULL instead of void is not accepted…

albe:
Hmm... reading sth. like
```php
return new Error('Wrong credentials.', null, [], $this->actionMethodName);
```
which should now be 
```php
return new Error('Wrong credentials.', 0, [], $this->actionMethodName);
```

makes me think if nullable code is maybe still the better idea, than "0" == no code

bwaidelich:
> makes me think if nullable code is maybe still the better idea, than "0" == no code

Yes, otherwise it would be a hell of a breaking change.
We could do `int? $code = 0` but the Php Inspections (EA Extended) complain about

image
image

But I think we should fix the type in the constructor like:
```php
    public function __construct(string $message, int $code = null, array $arguments = [], string $title = '')
    {
        $this->code = $code ?? 0;
        // ...
```

albe:
Green 🎉

PR #1721, opened 2019-08-21
TASK: Further cleanup test classes

daniellienert:
- Replace \PHPUnit_Framework_MockObject_MockObject hints
- Replace object calls with static calls


albe:
Just found https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Assert/Functions.php
https://phpunit.readthedocs.io/en/8.3/assertions.html
so, maybe we should directly switch to use the global assertion functions, which make all those assertions soooo much easier to read:

`$mock->method('foo')->expects(once())`

because, what is a "self once"? I try to read the code as a sentence describing the test (setup) and those `$this`/`self` references are just in the way.

Also, while we're at it, let's do the following:

Replace `->will($this->returnValue(...))` with `->willReturn(...)` everywhere (it's already done in a few places) and completely remove `->expects($this->any())`, as the any invocation expectation is basically a NOP.

```
$fooMock->method('foo')->expects($this->any())->will($this->returnValue($someFooValue));
```

vs.

```
$fooMock->method('foo')->expects(self::any())->will(self::returnValue($someFooValue));
```

vs.

```
$fooMock->method('foo')->willReturn($someFooValue);
```

Which do you prefer to read and which is more clear?

daniellienert:
Hey @albe:
I am unsure about the `once()`, I don't use globale methods anymore.
About the 3 options:

In 1. we call static methods with `$this` which is not more clear than 2. `self::` but more correct.
3. Is nice but requires a manual refactoring. This PR is mostly find / replace

As I already replaced most of the static calls I would like to merge this and improve later as soon as the tests pass.

daniellienert:
Hey @bwaidelich, 
tests are currently failing with 

```
Neos\Flow\Tests\Unit\Mvc\DispatcherTest::dispatchDoesNotSetInterceptedRequestIfRequestMethodIsNotGET
Expectation failed for method name is equal to 'getAuthenticationTokens' when invoked at least once.
Expected invocation at least once but it never occurred.
``` 

Could you help here, if test or the actual code needs to be adjusted?

bwaidelich:
@daniellienert I'll look into it!

bwaidelich:
@daniellienert Apparently the Dispatcher Fix (#1695) did not got upmerged with 3f305a23fb289fbc59f99ff161e9597f35b52fbb - I added that fix in this branch now and adjusted the test

bwaidelich:
FYI: I just merged #1614 - will resolve the conflicts in a minute

bwaidelich:
image
..I'm afraid this needs some more work

kdambekalns:
I agree with @daniellienert – when I read about those global functions recently, I was shocked this is actually offered.

@albe About `any()` - it's only a NOP when not followed by `with(…)`, so be careful! Also, not in this change, as @daniellienert said.

bwaidelich:
Some tests have been disabled (`@test_disabled`) with 2eb076d6c371bdb7761c5efa0f91e9b7cd3be9ab – and IMO we should urgently re-activate those

bwaidelich:
@daniellienert I'm on it. Are you OK with me pushing more tests and tweaks into this branch?

albe:
> In 1. we call static methods with $this which is not more clear than 2. self:: but more correct.
3. Is nice but requires a manual refactoring. This PR is mostly find / replace

Oh, 3. can perfectly be solved with a find / replace, just with a regex.
`'/->will\(\$this->returnValue\((.*?)\)\)/' => '->willReturn($1)'`

> I agree with @daniellienert – when I read about those global functions recently, I was shocked this is actually offered.

I don't like global functions either, but in the context of writing tests, which are inherently very technical already, I would gladly take anything that makes writing and reading (!) them more straightforward. As said, I have no idea what a "self once" or "self any" is supposed to be. But okay with moving this to another PR where we can decide this separately.

> @albe About `any()` - it's only a NOP when not followed by `with(…)`, so be careful! Also, not in this change, as @daniellienert said.

Huh? Now I'm curious.
`method('foo')->expects(any())->with(...)->willReturn('bar')`
is not the same as
`method('foo')->with(...)->willReturn('bar')`
?
I read them both as *"expect method foo when called with arguments ... (any amount of time - including never) to have it return 'bar'"*
I couldn't find something with a quick seach on the phpunit docs, but I might be missing something.

albe:
Need to revert last commit first, as that was an error due to me wrongly documenting the psalm command on discuss 🤦‍♂

albe:
```
ERROR: UndefinedClass - Packages/Framework/Neos.Flow/Classes/Command/ResourceCommandController.php:234:58 - Class or interface Neos\Media\Domain\Repository\AssetRepository does not exist
            $assetRepository = $this->objectManager->get(AssetRepository::class);
ERROR: UndefinedClass - Packages/Framework/Neos.Flow/Classes/Command/ResourceCommandController.php:236:62 - Class or interface Neos\Media\Domain\Repository\ThumbnailRepository does not exist
            $thumbnailRepository = $this->objectManager->get(ThumbnailRepository::class);ERROR: UndefinedClass - Packages/Framework/Neos.Flow/Classes/Command/ResourceCommandController.php:234:58 - Class or interface Neos\Media\Domain\Repository\AssetRepository does not exist
            $assetRepository = $this->objectManager->get(AssetRepository::class);
ERROR: UndefinedClass - Packages/Framework/Neos.Flow/Classes/Command/ResourceCommandController.php:236:62 - Class or interface Neos\Media\Domain\Repository\ThumbnailRepository does not exist
            $thumbnailRepository = $this->objectManager->get(ThumbnailRepository::class);
```

those are new 🤔

Ah, apparently got removed from the baseline:
https://github.com/neos/flow-development-collection/pull/1721/commits/6224528586c3ce5939608430c27ea514be19d748#diff-d828396aac59073d4c3416476ed8a4bbL775-L778

Guess you had the Media package installed locally when running psalm. Hmm... that's pretty fragile. Definitely need some CI process for this...

daniellienert:
Argh, yeah sure - I work on the Neos base distribution ...

PR #1729, opened 2019-08-23
BUGFIX: Trigger Entry Points of all authenticated tokens

bwaidelich:
Previously the `Dispatcher` invoked the `startAuthentication()` method
of all authenticated tokens.

That behavior was changed by accident with #1552 and now only the first
Entry Point was triggered.

### DIFF
--- a/Neos.Flow/Classes/Http/Component/SecurityEntryPointComponent.php
+++ b/Neos.Flow/Classes/Http/Component/SecurityEntryPointComponent.php
@@ -63,7 +63,7 @@ class SecurityEntryPointComponent implements ComponentInterface
                 $actionRequest = $componentContext->getParameter(DispatchComponent::class, 'actionRequest');
                 $this->securityContext->setInterceptedRequest($actionRequest->getMainRequest());
             }
-            $response = $entryPoint->startAuthentication($componentContext->getHttpRequest(), $componentContext->getHttpResponse());
+            $response = $entryPoint->startAuthentication($componentContext->getHttpRequest(), $response);
         }
         $componentContext->replaceHttpResponse($response);
     }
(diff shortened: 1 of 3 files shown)


Neos.Flow/Classes/Http/Component/SecurityEntryPointComponent.php
```diff
@@ -34,46 +34,37 @@ class SecurityEntryPointComponent implements ComponentInterface
...
-            if ($token->getAuthenticationEntryPoint() !== null) {
-                return $token;
+            // Only store the intercepted request if it is a GET request (otherwise it can't be resumed properly)
+            // We also don't store the request for "sessionless authentications" because that would implicitly start a session
+            // TODO: Adjust when a session-independent storing mechanism is possible (see https://github.com/neos/flow-development-collection/issues/1693)
+            if (!$token instanceof SessionlessTokenInterface && $componentContext->getHttpRequest()->getMethod() === 'GET') {
+                $actionRequest = $componentContext->getParameter(DispatchComponent::class, 'actionRequest');
+                $this->securityContext->setInterceptedRequest($actionRequest->getMainRequest());
             }
+            $response = $entryPoint->startAuthentication($componentContext->getHttpRequest(), $componentContext->getHttpResponse());
```
kitsunet: Something to look at again I guess. I never realised it was done this way. At least by reusing the context response we prevent mixing infos from multiple entry points, but I guess at some point making this a bit more deterministic would be neat.

bwaidelich: Oh, hang on, I messed up here.. This should of course take the (potentially new) `$response` or the last one will win
```suggestion
            $response = $entryPoint->startAuthentication($componentContext->getHttpRequest(), $response);
```

bwaidelich: @kitsunet right?

PR #1740, opened 2019-08-26
TASK: Add type declaration to Neos cache classes

daniellienert:
- Adds missing type declarations, especially void types.
- The remaining methods are implementing the untyped PSR interface

### DIFF
--- a/Neos.Cache/Classes/Backend/MultiBackend.php
+++ b/Neos.Cache/Classes/Backend/MultiBackend.php
@@ -1,4 +1,6 @@
 <?php
+declare(strict_types=1);
+
 namespace Neos\Cache\Backend;
 
 /*
@@ -24,7 +26,7 @@ class MultiBackend extends AbstractBackend implements BackendInterface
     /**
      * Configuration for all sub backends (each with the keys "backend" and "backendOptions")
      *
-     * @var array
+     * @var mixed[]
      */
     protected $backendConfigurations = [];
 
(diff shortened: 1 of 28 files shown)


daniellienert:
@albe: Again psalm delivered great suggestions and found some possible errors!

daniellienert:
Are there any objections to merge this?

Neos.Cache/Classes/Backend/MultiBackend.php
```diff
@@ -221,10 +223,10 @@ public function collectGarbage()
     }
 
     /**
-     * @param array $backendConfigurations
+     * @param mixed[] $backendConfigurations
```
albe: Huh, that's an odd way to denote `array` 🤨

kdambekalns: Yup, wouldn't do that. I get `string[]` to make clear what is expected, but `mixed[]` is in no way better than `array` IMHO.

Plus: currently a type like `string[]` breaks Doctrine, so using this in entities would be bad. So let's keep that notation out for now, please. Avoids confusion.

kdambekalns: Then again… `string[]` really is helpful. (But `mixed[]`… not so much.)

daniellienert:
Got it 😉 I’ll take care. Analyzed the classes with phpcs, and in my config it complains about array annotations

PR #1741, opened 2019-08-29
 BUGFIX: Replacing suffixes appends, if nothing to replace

kdambekalns:
The `UriConstraints` behaves weird with `replaceSuffixes` on hosts: if the suffix to
replace is not found, the replacement is appended to the host - instead of nothing
happening.


Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
```diff
@@ -252,13 +252,17 @@ public function applyTo(UriInterface $templateUri, bool $forceAbsoluteUri): UriI
...
-                if ($this->stringEndsWith($host, $replaceSuffix)) {
-                    $host = substr($host, 0, -strlen($replaceSuffix));
-                    break;
+
+            if ($replaceSuffixes === []) {
+                $host .= $suffix;
+            } else {
+                foreach ($replaceSuffixes as $replaceSuffix) {
+                    if ($this->stringEndsWith($host, $replaceSuffix)) {
+                        $host = preg_replace(sprintf('/%s$/i', $replaceSuffix), $suffix, $host);
```
bwaidelich:
```suggestion
                        $host = preg_replace('/' . $replaceSuffix . '$/i', $suffix, $host);
```
Just a micro-optimization (feel free to ignore), but `sprinf` is rather slow

albe: Could even be optimized much more by building a piped regex matcher for all replace suffixes:
```php
$regex = '/' . implode('|', array_map('preg_quote', $replaceSuffixes)) . '$/';
$host = preg_replace($regex, $suffix, $host);
```
https://3v4l.org/9a8Cs

kdambekalns: Thanks!

kdambekalns:
I need to look at the prefix handling I adjusted as well… _sigh_

albe:
Okay, some words on this weird "fix":

The issue was, that the applyToTest expected that the prefix is applied, when no match for the replacements was found (and the prefix isn't already in the original host).
i.e. a target host of `some-domain.tld` with replacements `['de.', 'ch.']` to prefix `'en.'` is expected to result in `en.some-domain.tld`.

So what I did is, add an "not-matches prefix" regex replacement alternative, which effectively just prepends the prefix if none of the other replacements match and the target string does not start with the prefix.

Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
```diff
@@ -252,13 +254,13 @@ public function applyTo(UriInterface $templateUri, bool $forceAbsoluteUri): UriI
...
-            foreach ($replaceSuffixes as $replaceSuffix) {
-                if ($this->stringEndsWith($host, $replaceSuffix)) {
-                    $host = substr($host, 0, -strlen($replaceSuffix));
-                    break;
-                }
+
+            if ($replaceSuffixes === []) {
+                $host .= $suffix;
+            } else {
+                $regex = '/(' . implode('|', array_map('preg_quote', $replaceSuffixes)) . ')$/';
```
albe: Golden question: as this is not covered by tests, but the same expectation as above should be in place here. Add it?

albe:
So I was curious... and really, let's revert the "optimization", because:

https://3v4l.org/hnMbe

the regex replacement is constantly slower by a factor of 2-5 (even if you remove the "no-match replacement" fix, though only by a factor < 2).

Another good case for "premature" optimization without actual benching.
I guess maybe in the bench the global method `stringStartsWith` is optimized away mostly, so potentially in the case of a class method that's not the case, but still... the readability is worth more than the potentially minimal performance win.

kdambekalns:
Ok, that last revert undid the prefix-replacement I adjusted as well. Granted, I mixed two things in one commit, but…

OTOH, now the PR does what I intended it to do.

Still, I think I'll give it another go. Replacements should only happen if there is a match.

kdambekalns:
BTW: Nice test for the effect of the `S` modifier: https://3v4l.org/OltXJ

kdambekalns:
> Still, I think I'll give it another go. Replacements should only happen if there is a match.

No, fine as is, I guess. I just now saw the comment added in the revert commit.

albe:
Sorry for being sneaky with such important information 😅

I mean, we could discuss this stance and add another (breaking, because test covered behavior) change to bring both cases in line. But after looking at it, it also somehow makes sense as is

PR #1743, opened 2019-08-30
TASK: Add missing trailing % in Configuration manual

dogawaf:


albe:
Ah, wait, this should be targeted lower as it's more of a BUGFIX really. 4.3 would be it

PR #1751, opened 2019-09-03
BUGFIX: Allow string for trusted proxies again (env variable use)

kdambekalns:
Fixes a regression introduced with #1683


kdambekalns:
Wrong branch, gnah.

albe:
Happens to me all the damn time too 😅

kdambekalns:
See #1752 instead.

PR #1752, opened 2019-09-03
BUGFIX: Allow string for trusted proxies again (env variable use)

kdambekalns:
Fixes a regression introduced with #1683

### DIFF
--- a/Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
+++ b/Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
@@ -49,14 +49,17 @@ class TrustedProxiesComponent implements ComponentInterface
 
         if ($this->settings['proxies'] === false) {
             $this->settings['proxies'] = [];
+            return;
         }
 
         if ($this->settings['proxies'] === ['*']) {
             $this->settings['proxies'] = '*';
         }
-        if (is_string($this->settings['proxies'])) {
+
+        if (is_string($this->settings['proxies']) && $this->settings['proxies'] !== '*') {
             $this->settings['proxies'] = array_map('trim', explode(',', $this->settings['proxies']));
         }
+
         if (!is_array($this->settings['proxies']) && $this->settings['proxies'] !== '*') {
             throw new InvalidConfigurationException('The Neos.Flow.http.trustedProxies.proxies setting may only be an array of IP addresses or the single string "*". Got "' . var_export($this->settings['proxies'], true) . '" instead.', 1564659249);
         }
--- a/Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
+++ b/Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
@@ -54,10 +54,13 @@ class TrustedProxiesComponent implements ComponentInterface
         if ($this->settings['proxies'] === ['*']) {
             $this->settings['proxies'] = '*';
         }
+        if (is_string($this->settings['proxies'])) {
+            $this->settings['proxies'] = array_map('trim', explode(',', $this->settings['proxies']));
+        }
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
```diff
@@ -54,10 +54,13 @@ public function injectSettings(array $settings)
...
             $this->settings['proxies'] = '*';
         }
+        if (is_string($this->settings['proxies'])) {
+            $this->settings['proxies'] = array_map('trim', explode(',', $this->settings['proxies']));
+        }
         if (!is_array($this->settings['proxies']) && $this->settings['proxies'] !== '*') {
             throw new InvalidConfigurationException('The Neos.Flow.http.trustedProxies.proxies setting may only be an array of IP addresses or the single string "*". Got "' . var_export($this->settings['proxies'], true) . '" instead.', 1564659249);
         }
-        if (is_array($this->settings['proxies']) && array_search('*', $this->settings['proxies'], true) !== false) {
+        if (is_array($this->settings['proxies']) && in_array('*', $this->settings['proxies'], true)) {
```
kdambekalns: This is simply a "clearer intent" fix

Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
```diff
@@ -54,10 +54,13 @@ public function injectSettings(array $settings)
         if ($this->settings['proxies'] === ['*']) {
             $this->settings['proxies'] = '*';
         }
+        if (is_string($this->settings['proxies'])) {
+            $this->settings['proxies'] = array_map('trim', explode(',', $this->settings['proxies']));
+        }
         if (!is_array($this->settings['proxies']) && $this->settings['proxies'] !== '*') {
```
albe: I guess this check is now even obsolete, since it will not be a single string at this point any more.

Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
```diff
@@ -54,10 +54,13 @@ public function injectSettings(array $settings)
         if ($this->settings['proxies'] === ['*']) {
             $this->settings['proxies'] = '*';
         }
+        if (is_string($this->settings['proxies'])) {
```
albe: Better than below: add a check for not equal "*" here.

Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
kdambekalns: It can stilll be `'*'` (now)…

Neos.Flow/Classes/Http/Component/TrustedProxiesComponent.php
```diff
@@ -49,15 +49,21 @@ public function injectSettings(array $settings)
...
         if ($this->settings['proxies'] === ['*']) {
             $this->settings['proxies'] = '*';
         }
+
+        if (is_string($this->settings['proxies']) && $this->settings['proxies'] !== '*') {
+            $this->settings['proxies'] = array_map('trim', explode(',', $this->settings['proxies']));
+        }
+
         if (!is_array($this->settings['proxies']) && $this->settings['proxies'] !== '*') {
             throw new InvalidConfigurationException('The Neos.Flow.http.trustedProxies.proxies setting may only be an array of IP addresses or the single string "*". Got "' . var_export($this->settings['proxies'], true) . '" instead.', 1564659249);
```
albe:
```suggestion
            throw new InvalidConfigurationException('The Neos.Flow.http.trustedProxies.proxies setting may only be an array or comma separated value of IP addresses or the single string "*". Got "' . var_export($this->settings['proxies'], true) . '" instead.', 1564659249);
```

PR #1754, opened 2019-09-03
BUGFIX: Safeguard ob_end_flush() against empty buffer

### LINKED ISSUE #1753: Warning turned exception in Http\RequestHandler
### Description

I just saw this when logging in to the Neos backend on "everything master":

`19-09-03 09:31:00 9372       CRITICAL                       Exception #1: Notice: ob_end_flush(): failed to delete and flush buffer. No buffer to delete or flush in /application/Packages/Framework/Neos.Flow/Classes/Http/RequestHandler.php line 170 - See also: 201909030930594e28ba.txt`

### Steps to Reproduce

1. Log in to Neos
2. See white screen
3. Check System_Development.log

### Cause

```php
protected function sendResponse()
{
    $response = $this->componentContext->getHttpResponse();
    ob_implicit_flush(1);
    foreach (ResponseInformationHelper::prepareHeaders($response) as $prepareHeader) {
        header($prepareHeader);
    }
    echo $response->getBody()->getContents();
    ob_end_flush();
}
```

### Affected Versions

Neos: n/a
Flow: master, I think

kdambekalns:
Fixes #1753

### DIFF
--- a/Neos.Flow/Classes/Http/RequestHandler.php
+++ b/Neos.Flow/Classes/Http/RequestHandler.php
@@ -167,6 +167,8 @@ class RequestHandler implements HttpRequestHandlerInterface
             header($prepareHeader);
         }
         echo $response->getBody()->getContents();
-        ob_end_flush();
+        while (ob_get_level() > 0) {
+            ob_end_flush();
+        }
     }
 }


Neos.Flow/Classes/Http/RequestHandler.php
```diff
@@ -167,6 +167,8 @@ protected function sendResponse()
             header($prepareHeader);
         }
         echo $response->getBody()->getContents();
-        ob_end_flush();
+        while (ob_get_level() > 0) {
```
kitsunet: The question is if we rather want to flush and start before the output?

kitsunet: Also now looking at it again, this looks more like a quick fix of mine. I think output of a PSR response should probably move to the ResponseInformationHelper ?

albe: I guess this also makes it technically impossible to catch a `sendResponse()` call into an output buffer now. Not that this makes much sense inside the HttpRequestHandler anyway, so fine for me

kitsunet:
@albe see my comments above, any opinions?

Neos.Flow/Classes/Http/RequestHandler.php
kdambekalns: The logic behind that is unknown to me anyway. Why should anything be buffered? That code was added with PSR-7, AFAICS, but why… I don't know. I'll let you decide.

albe: Not sure, feels a bit over the top to have a static method in a Helper to send the PSR response? IMO it should be at the lowest level of a PSR RequestHandler (https://www.php-fig.org/psr/psr-15/#21-psrhttpserverrequesthandlerinterface)

kdambekalns: I can't even remember when I last used anything dealing output buffering…

albe: me neither 🤷‍♂

kitsunet: My point is that this code as is is not reusable at all, and I guess it would be nice to do that. and @kdambekalns I am pretty sure I jusut took inspiration from other code thtat sents psr messages, eg. other frameworks or psr implementations

albe: Hm, still not sure why/where this needs to be reusable. The request handler invokes the middleware/component chain, then outputs the Response it receives to the browser. Why would you do that anywhere else too?

kitsunet: own request handler for example. but yeah fine, lets merge this for now :D

kdambekalns: Thanks, at least logins to Neos are possible again. :)

PR #1755, opened 2019-09-03
TASK: Provide comprehensive solution to get base URI

kitsunet:
This introduces a provider that is a comprehensive source for
a possible base URI.

This is in preparation of getting rid of
ServerRequestAttributes::BASE_URI.


kitsunet:
alright @bwaidelich please have a look at the todo comment I left there. I might need a bit of help to figure out how to get rid of that. the relativeRequestPath thing is mostly used in routing, so we would probably need the BaseUriProvider there then but I wasn't sure if that makes sense and works out.

kitsunet:
Yes obviously unit tests need to be adapted.

kitsunet:
yes, stateful singleton I know. Seemed the best way to provide an extension point to this. I know that @kdambekalns just a few days ago had a case where he needed to adapt base URI via HTTP component. I still feel that just saying "get yourself the configured base URI injected and otherwise build one from the current request or pull it from wherever you want" results in thousand different solutions and behaviours, which would be bad.

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
```diff
@@ -68,6 +68,7 @@ public static function getScriptRequestPath(ServerRequestInterface $request): st
      */
     public static function getRelativeRequestPath(ServerRequestInterface $request): string
     {
+        // TODO: How to fix this? Probably give base URI to method, but how to get it in usages.
```
bwaidelich: Why do we need this method, where is it used?

kitsunet: In Routing ;)
\Neos\Flow\Mvc\Routing\Route::matches
\Neos\Flow\Mvc\Routing\Dto\RouteContext::getCacheEntryIdentifier
\Neos\Flow\Mvc\Routing\RouterCachingService::storeMatchResults

Neos.Flow/Classes/Http/HttpRequestHandlerInterface.php
```diff
@@ -26,15 +27,23 @@ interface HttpRequestHandlerInterface extends RequestHandlerInterface
      * Returns the currently processed HTTP request
      *
      * @return ServerRequestInterface
-     * @api
+     * @deprecated
```
bwaidelich: Please lets always add since when something was deprecated and why/replacement
```suggestion
     * @deprecated since 6.0 use getComponentContext()->getHttpRequest() instead
```

Neos.Flow/Classes/Http/HttpRequestHandlerInterface.php
```diff
@@ -26,15 +27,23 @@ interface HttpRequestHandlerInterface extends RequestHandlerInterface
...
      */
     public function getHttpRequest();
 
     /**
      * Returns the HTTP response corresponding to the currently handled request
      *
      * @return ResponseInterface
-     * @api
+     * @deprecated
```
bwaidelich:
```suggestion
     * @deprecated since 6.0 use getComponentContext()->getHttpResponse() instead
```

kitsunet:
To skip (this) stateful singleton we could also provide an interface for "bestPossible" (because everything else you could do in your own code as well) and let people overwrite that if they need the "runtime" part to adjust the base URI in the component chain...

bwaidelich:
I have to wrap my head around this, but IMO a lot of complexity comes from the potential "relative request path" being different from request->uri->path (i.e. running Flow not in the document root).

The current solution to detect this is:
```php
UriHelper::getRelativePath($baseUri, $request->getUri())
```
where `$baseUri` might not be the current request URI at all – so this is could lead to wrong results, no?

What if we set the "scriptRequestPath" as param to the `ServerRequest` instead?

kitsunet:
> I have to wrap my head around this, but IMO a lot of complexity comes from the potential "relative request path" being different from request->uri->path (i.e. running Flow not in the document root).
> 
> The current solution to detect this is:
> 
> ```
> UriHelper::getRelativePath($baseUri, $request->getUri())
> ```
> 
> where `$baseUri` might not be the current request URI at all – so this is could lead to wrong results, no?
> 
> What if we set the "scriptRequestPath" as param to the `ServerRequest` instead?

That sounds sensible I guess. That's part of my confusion, we cannot just take any base URI, it should be related to the request obviously. I wonder if there isn't a way to  detect the request path via $_SERVER ?

kitsunet:
$_SERVER['SCRIPT_NAME'] seems to be the right thing. This always includes the script, regardless of rewrite or not (which is fine IMHO) but also the path to the script from the webroot. so something like `/index.php` in the default case or `/sub/index.php` in case of a sub directory. Maybe check if that is true for your servers too. In that case this seesm to me the way to get the relative path and nothing else?

kitsunet:
Oh NVM, the method should construct the request path after removing any path segments from the base URI, but IMHO that is weird? it should probably rather be the relative path from `getScriptRequestPath` ?

bwaidelich:
> To skip (this) stateful singleton we could also provide an interface for "bestPossible"

I think "bestPossible" doesn't make the intention any clearer to be honest.
Let me see if I got the different use cases right:
* Get the current requests base URI (potentially including relative script path)
* Get the configured baseUri
* Get one or the other (fallback)

I think this can be greatly simplified, but I might miss something. Can we quickly talk about this once again?

Neos.Flow/Classes/Http/BaseUriProvider.php
```diff
@@ -0,0 +1,124 @@
...
+     * - configured base URI
+     * - generated base URI from request
+     *
+     * To ensure a base URI can always be provided this will throw an
+     * exception if none of the three options yields a result.
+     *
+     * @return UriInterface
+     * @throws Exception
+     */
+    public function getBestPossibleBaseUri(): UriInterface
```
daniellienert: What do you think about just `determineBaseUri()`?

bwaidelich: My point is:
the configured or runtime base Uri is not any "better" than the base URI of the current request.
"determineBaseUri" does not really solve the problem, that we are trying to cover multiple use cases with one name.
IMO something quirky but more explicit like `getCurrentRequestUriOrFallbackToConfiguredBaseUri(...)` wouldn't be the worst

bwaidelich: `getConfiguredBaseUriOrFallbackToCurrentBaseUri()` I meant

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
kitsunet: After our discussion and re-reading the code it makes only sense to build a "base URI" from the given request, everything else is wrong anyway, so that's changed.

Neos.Flow/Classes/Http/BaseUriProvider.php
kitsunet: `getConfiguredBaseUriOrFallbackToCurrentRequest` is what we have now :)

kitsunet: `getConfiguredBaseUriOrFallbackToCurrentRequest` is what I did now...

bwaidelich: > getConfiguredBaseUriOrFallbackToCurrentRequest is what I did now...

uhm, where? :)

kitsunet: Yeah, task switching like... not a pro :D I did git push and switched over to other things, but I forgot to notice that it couldn'T push because I committed your suggestions so it wasn't up to speed with the branch :D

kitsunet: there we go

kitsunet:
If we agree on the concept and code I will adjust tests quickly, should be only a unit test formality about injections.

Neos.Flow/Classes/Http/BaseUriProvider.php
```diff
@@ -0,0 +1,88 @@
...
+     * @var string|null
+     */
+    protected $configuredBaseUri;
+
+    /**
+     * @Flow\Inject
+     * @var Bootstrap
+     */
+    protected $bootstrap;
+    /**
```
bwaidelich:
```suggestion

    /**
``` (nitpick)

Comment (kitsunet) [reply Neos.Flow/Classes/Http/BaseUriProvider.php]: I don't get the suggesion^^

Comment (kdambekalns) [reply Neos.Flow/Classes/Http/BaseUriProvider.php]: Add an empty line after `protected $bootstrap;`

File: Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php [hunk 6]
```diff
@@ -392,23 +397,16 @@ protected function getResourcesBaseUri()
...
-        $requestHandler = $this->bootstrap->getActiveRequestHandler();
-        if ($requestHandler instanceof HttpRequestHandlerInterface) {
-            return $requestHandler->getHttpRequest()->getAttribute(ServerRequestAttributes::BASE_URI) . $this->baseUri;
-        }
-
-        if ($this->httpBaseUri === null) {
-            throw new TargetException(sprintf('The base URI for resources could not be detected. Please specify the "Neos.Flow.http.baseUri" setting or use an absolute "baseUri" option for target "%s".', $this->name), 1438093977);
-        }
-        return $this->httpBaseUri . $this->baseUri;
+        $httpBaseUri = (string)$this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest();
```
bwaidelich: Ah, one more suggestion: Should we catch the `\Neos\Flow\Http\Exception` here and throw a more meaningful here like
```php
catch (HttpException $exception) {
  throw new SomeOtherException(sprintf('Failed to determine base URI for FileSystemTarget "%s", please specify the "baseUri" option', $this->name), 1567668788, $exception);
}
```
?

kitsunet: I am not sure how that is better? I thought about it as well, but then think hiding an exception with another one? What's the point. The necessary action is the same and is already outlined in the first one.

Issue #1761, opened 2019-09-09
Compatibility with TYPO3 Fluid 2.6.2 (and 2.6.6)


kdambekalns:
### Description

A test fails:

```
There was 1 failure:

1) Neos\FluidAdaptor\Tests\Functional\View\StandaloneViewTest::xmlNamespacesCanBeIgnored
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'<foo:bar /><bar:foo></bar:foo><foo.bar:baz />foobar'
+'<foo:bar /><bar:foo></bar:foo><foo.bar:baz /><f:format.raw>foobar</f:format.raw>'

…/Packages/Framework/Neos.FluidAdaptor/Tests/Functional/View/StandaloneViewTest.php:282
…/Packages/Framework/Neos.FluidAdaptor/Tests/Functional/View/StandaloneViewTest.php:42
…/Packages/Libraries/phpunit/phpunit/phpunit:61
```

### Steps to Reproduce

1. Install Neos 5.0.x-dev
2. Run functional tests
3. See it fail

#### Actual behavior

My guess: this is caused by Fluid 2.6.2 - in https://travis-ci.org/neos/flow-development-collection/jobs/581986135 it's still 2.6.1 and the tests pass, in https://travis-ci.org/neos/flow-development-collection/jobs/582499795 it's 2.6.2

Looks like this: https://github.com/TYPO3/Fluid/commit/f85a5a7fff02e43f688e31b34d284efb41caf44f (from https://github.com/TYPO3/Fluid/pull/451), claiming:

> Behavior remains unchanged.

… unless you changed the handling that took place in the TemplateProcessor (like we did, AFAICU)

#### Further observations

Doing this in "our" FluidAdaptor RenderingContext:

```
#use Neos\FluidAdaptor\Core\Parser\TemplateProcessor\NamespaceDetectionTemplateProcessor;
use TYPO3Fluid\Fluid\Core\Parser\TemplateProcessor\NamespaceDetectionTemplateProcessor;
```

sees all funtional tests pass, just the one about ignoring namespaces still fails. ~It seems we do not need that processor (anymore).~ That shows we lack a test, see https://github.com/neos/flow-development-collection/issues/1761#issuecomment-529360334 and https://github.com/TYPO3/Fluid/issues/474

Question: do we need that test? Or is that tested in TYPO3 Fluid by now, since that ignoring of namespaces is core functionality?

[...]

PR #1762, opened 2019-09-09
TASK: FluidAdaptor cleanup

kdambekalns:
See #1761 for the "trigger" to this one.

It attempts to clean up outdated code and assumptions about the code
of TYPO3 Fluid that no longer hold.


kdambekalns:
This can sure be amended with more cleanup!

kdambekalns:
Note: we need our own template processor to fix CDATA handling. **DO NOT MERGE AS IS.**

Issue #1763, opened 2019-09-10
PdoBackend: Error when running cache:setupall after upgrading neos/cache from 5.3.3 to 5.3.4


drlimbo:
### Description
Error when running `cache:setupall` after upgrading neos/cache from 5.3.3 to 5.3.4

`Neos_Media_ImageSize:
Creating database tables "cache" & "tags"...
ERROR: Could not create cache tables with DSN "sqlite:/..../Data/Temporary/Production/Cache/Data/Neos_Media_ImageSize.sqlite". 
PDO error: SQLSTATE[HY000]: General error: 1 table "cache" already exists`

### Steps to Reproduce

1. Setup PDOBackend Cache using an SQLite database
2. Clear all Caches
3. Run `cache:setupall`
4. 💥

#### Expected behavior
Cache should be setup. 

#### Actual behavior
Throws an Error, but table is created. 

### Affected Versions
Neos: 5.2.8, 5.3.4

already discussed with @bwaidelich [in this Slack thread](https://neos-project.slack.com/archives/C050C8FEK/p1568107878288200)

bwaidelich:
@drlimbo Thanks for creating the ticket. I reproduced this locally with the following `Caches.yaml` configuration:
```yaml
Flow_Mvc_Routing_Route:
  backend: Neos\Cache\Backend\PdoBackend
  backendOptions:
    dataSourceName: 'sqlite:Data/cache.db'
```

Result of `./flow cache:setupall`:

```
...
Flow_I18n_Cldr_Reader_CurrencyReaderCache:
OK
Flow_Monitor:
OK
Flow_Mvc_Routing_Route:
Creating database tables "cache" & "tags"...
ERROR: Could not create cache tables with DSN "sqlite:Data/cache.db". PDO error: SQLSTATE[HY000]: General error: 1 table "cache" already exists
...
```

The reason for this is that the SQLite database is [automatically created upon connection](https://github.com/neos/flow-development-collection/blob/master/Neos.Cache/Classes/Backend/PdoBackend.php#L368) so there is no need to set them up explicitly.

So technically the error is correct – it tries to create tables that exist already.

But I agree that the error message might be confusing and I'll take care of a PR

bwaidelich:
Fixed with #1766

PR #1764, opened 2019-09-10
Improve PdoBackend

bwaidelich:
Related: #1763


Sebobo:
This should also replace the table names with the ones from the config.

Neos.Cache/Resources/Private/DDL_pgsql.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS "cache" (
+  "identifier" VARCHAR(250) NOT NULL,
+  "cache" VARCHAR(250) NOT NULL,
+  "context" VARCHAR(150) NOT NULL,
+  "created" INTEGER NOT NULL,
```
Sebobo: I saw a recommendation to use `"created" INTEGER CHECK(created >= 0) NOT NULL,` to replace the `UNSIGNED` config.

Neos.Cache/Resources/Private/DDL_pgsql.sql
```diff
@@ -0,0 +1,22 @@
...
+  "lifetime" INTEGER DEFAULT '0' NOT NULL,
```
Sebobo:
```suggestion
  "lifetime" INTEGER CHECK(lifetime >= 0) DEFAULT '0' NOT NULL,
```

Neos.Cache/Classes/Backend/PdoBackend.php
```diff
@@ -386,8 +385,12 @@ protected function connect()
     protected function createCacheTables()
     {
         $this->connect();
+        $sqlPath = __DIR__ . '/../../Resources/Private/DDL_' . strtolower($this->pdoDriver) . '.sql';
+        if (!file_exists($sqlPath)) {
+            throw new Exception(sprintf('Could not find SQL for pdo driver "%s" (at "%s")', $this->pdoDriver, $sqlPath), 1568133488);
+        }
         try {
-            PdoHelper::importSql($this->databaseHandle, $this->pdoDriver, __DIR__ . '/../../Resources/Private/DDL.sql');
+            $this->databaseHandle->exec(file_get_contents($sqlPath));
```
Sebobo: We could first read the sql and then replace the table names. This way the method works with any number of configured caches.

Neos.Cache/Resources/Private/DDL_pgsql.sql
bwaidelich: Sorry for the (very) late reaction...
what does this do and is this syntax compatible with all supported platforms?

bwaidelich: nevermind – this PR is totally outdated

bwaidelich:
=> #1766

PR #1767, opened 2019-09-11
!!! TASK: Set $result private, to remove it from the interface

daniellienert:
Don't access `$this->result` inside Validators and instead use pushResult/popResult/getResult.


daniellienert:
@albe, rebased it. With your changes, would you reconsider your vote?

kdambekalns:
@albe Could you point me to the psalm docs you wrote some time ago? Then I'll fix that…

PR #1770, opened 2019-09-12
TASK: Remove leftover hints at BASE_URI attribute

kitsunet:


daniellienert:
Psalm failures are unrelated and fixed in current 6.0.

PR #1771, opened 2019-09-12
!!!TASK: Remove Lock package from dev collection

kitsunet:
The Lock package is no longer in use anywhere in the Flow or Neos core since several versions, it is effectively deprecated. This change removes it and the last leftovers from the development collection.

This is marked breaking because the LockManager is no longer initialised after this change and the package would not be installed by default.
If you need the package in your own code you should require it independently and setup the LockManager as you need. We suggest looking into alternative locking solutions though.


daniellienert:
@kitsunet: Would you please provide a sentence, why this is removed? As you removed the configuration? Is there no usage in Flow or Neos where need that configuration?

kitsunet:
@daniellienert you are right, sorry, here we go :)

kitsunet:
Would have to dig up the original change, but AFAIK it was planned and communciated that way. Given that the package was removed from active duty a long while ago.

PR #1774, opened 2019-09-17
TASK: Remove lockStrategyClassName from settings

daniellienert:
Remove the configuration in flow and the according settings schema, 
as the Lock package was removed already in https://github.com/neos/flow-development-collection/pull/1771

### DIFF
--- a/Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.utility.schema.yaml
+++ b/Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.utility.schema.yaml
@@ -1,4 +1,3 @@
 type: dictionary
 additionalProperties: false
-properties:
-
+properties: []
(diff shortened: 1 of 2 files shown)


Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.utility.schema.yaml
```diff
@@ -1,7 +1,4 @@
 type: dictionary
 additionalProperties: false
 properties:
-  'lockStrategyClassName':
-    type: string
-    required: true
-    format: class-name
+
```
kdambekalns: No `[]` or `null` here looks weird…

kitsunet: And the tests hate it :D

daniellienert:
Thanks, Tests run green locally. Lets see if that works.

PR #1775, opened 2019-09-18
TASK: Update Psalm baseline

albe:


psalm-baseline.xml
```diff
@@ -53,6 +53,25 @@
     <PossiblyNullPropertyAssignmentValue occurrences="1">
       <code>null</code>
     </PossiblyNullPropertyAssignmentValue>
+    <UndefinedClass occurrences="2">
```
albe: Not sure where these come from... they definitely weren't in the psalm errors before. Maybe difference local run vs. travis environment?

psalm-baseline.xml
```diff
@@ -3005,6 +3042,16 @@
       <code>$validator-&gt;setValidatedInstancesContainer</code>
     </NoInterfaceProperties>
   </file>
+  <file src="Packages/Framework/Neos.Flow/Classes/Validation/Validator/CollectionValidator.php">
```
albe: These two are the errors from #1767

psalm-baseline.xml
kdambekalns: Maybe because I don't have Redis installed? Pretty stupid check, then… :)

But no… I have ext-redis installed. 🤷‍♂️

albe:
https://travis-ci.org/neos/flow-development-collection/jobs/586487624

travis too as well as another one... don't get that on my machine though

albe:
Ok, tried with a fresh install of the development collection and that seems to have to do with it

albe:
Green now. Okay to merge @kdambekalns ?

PR #1777, opened 2019-09-18
BUGFIX: Avoid changing REQUEST_TIME in functional tests

kdambekalns:
This fixes two symptoms:

- the time reported for functional tests (used to be years…)
- an error in the RequestHandlerTest caused by the PhpUnit Timer
  complaining about not being able to determine the time


kdambekalns:
And I even actively tried to find the cause, but did not manage to do it. This time it was really easy… 🤣

PR #1780, opened 2019-09-19
BUGFIX: Handle nullable property annotations correctly

### LINKED ISSUE #1779: Class properties annotated with multiple possible types (e.g. string|null) will cause exception
```
    /**
     * @var string|null
     * @ORM\Column(type="string")
     */
    protected $articleNumber;
```
The type annotation above will cause flow to fail very early during bootstrap with the exception ```Found an invalid element type declaration.
A type "'string|null'" does not exist.``` 

Full stacktrace:

```
Exception #1315564475 in line 1586 of /var/www/projectXy/Packages/Framework/Neos.Flow/Classes/Reflection/ReflectionService.php: Found an invalid element type declaration. A type "'string|null'" does not exist.

23 Neos\Flow\Reflection\ReflectionService::evaluateClassPropertyAnnotationsForSchema(Neos\Flow\Reflection\ClassSchema, "articleNumber")
22 Neos\Flow\Reflection\ReflectionService::addPropertiesToClassSchema(Neos\Flow\Reflection\ClassSchema)
21 Neos\Flow\Reflection\ReflectionService::buildClassSchema("GesagtGetan\CustomerBackend\Domain\SystemCare\Model\SystemCareSlot")
20 Neos\Flow\Reflection\ReflectionService::buildClassSchemata(array|46|)
19 Neos\Flow\Reflection\ReflectionService::reflectEmergedClasses()
18 Neos\Flow\Reflection\ReflectionService::buildReflectionData(array|46|)
17 Neos\Flow\ObjectManagement\CompileTimeObjectManager::initialize(array|175|)
16 Neos\Flow\Core\Booting\Scripts::initializeObjectManagerCompileTimeFinalize(Neos\Flow\Core\Bootstrap)
15 call_user_func(array|2|, Neos\Flow\Core\Bootstrap)
14 Neos\Flow\Core\Booting\Step::__invoke(Neos\Flow\Core\Bootstrap)
13 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
12 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
11 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
10 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
9 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
8 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
7 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
6 Neos\Flow\Core\Booting\Sequence::invokeStep(Neos\Flow\Core\Booting\Step, Neos\Flow\Core\Bootstrap)
5 Neos\Flow\Core\Booting\Sequence::invoke(Neos\Flow\Core\Bootstrap)
4 Neos\Flow\Cli\CommandRequestHandler::boot("Compiletime")
3 Neos\Flow\Cli\CommandRequestHandler::handleRequest()
2 Neos\Flow\Core\Bootstrap::run()
1 require("/var/www/projectXy/Packages/Framework/Neos.Flow/Scripts/flow.php")
```

### Affected Versions

Flow: 6.0.0

albe:
This fixes a regression introduced with #1640, which caused properties on entities annotated as nullable (`|null`) to throw an exception during reflection time in Flow.

Since #1640 we more strictly parse type annotations. Before, something like `string|null` was parsed as just `string` (and even `stringnull` would have worked, which was obviously wrong). With the bugfix , a `|null` was no longer parsed, because a whitespace or line ending was expected at the end of the type annotation `(?:\s|$)`.

This change fixes this, by explicitly allowing `null|` and `|null` in the parser regex - all other type unions are still not allowed because we can not resolve that to some useful type information.

Resolves #1779


albe:
Note: Basically the whole section at https://github.com/neos/flow-development-collection/blob/albe-nullable-entity-properties/Neos.Flow/Classes/Reflection/ReflectionService.php#L1628-L1632 can be removed by now, because the parseType check is done again (and throws) inside `ClassSchema::addProperty` - I don't dare to do this inside this bugfix though.

albe:
I don't think doctrine can do something with a "mixed" type (see https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/basic-mapping.html#doctrine-mapping-types). For that to work you'd have to add a `@ORM\Column(type="theactualtypeyouwant")` and that was the case even before the stricter matching (did it even allow `mixed` at all?). We could for this specific case throw a more meaningful exception that just states that you need to set an `@ORM\Column` type for this to work.

For the others, it's likewise, just that previously those would always (accidentially) resolve to the first type in the union, so both above cases would be the exact same as just `@var int`.

So, should we allow those union types? If so, what should they resolve to (sensibly, not "first wins")? IMO that's more of a thing for a new feature then if at all.

kdambekalns:
Yes, with Doctrine that doesn't make sense, but does it break reflection? That's my concern here… and I didn't test it. 😬

albe:
Yes it is broken but only for entities with union types. We only do `parseType` there (and in type converters and validator creation, where we expect strict types anyway to resolve to a single validator for the type).

gerhard-boden:
Thanks for taking care! 
I applied this patch to my project and I think we're not quite there yet:
```
    /**
     * @var \DateTimeInterface|null
     * @ORM\Column(type="datetime", nullable=true)
     */
    protected $deletionDateTime;
```

results in `A type "'\\DateTimeInterface|null|null'" does not exist.`

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -51,7 +51,9 @@ public static function parseType(string $type): array
             throw new InvalidTypeException('Found an invalid element type declaration. A type "' . var_export($type, true) . '" does not exist.', 1264093630);
         }
 
-        $type = self::normalizeType($matches['type']);
+        $typeWithoutNull = self::stripNullableType($matches['type']);
+        $isNullable = $typeWithoutNull !== $matches['type'];
```
gerhard-boden: Couldn't spot a usage for this variable within the function

albe: Indeed, that was an attempt to already make the method return nullablity info for the type... but then that would be changing behaviour with possible side-effects in a bugfix which is bad.

albe:
> A type "'\\DateTimeInterface|null|null'" 

Uhmm... no clue how that would happen

gerhard-boden:
> > A type "'\DateTimeInterface|null|null'"
> 
> Uhmm... no clue how that would happen

Maybe it's because the type starts with `\`? It only happens when I apply the code from this PR, so I guess it has to do with this pattern?
https://github.com/neos/flow-development-collection/pull/1780/files#diff-18e35307a68fb093757fbd5382f9a641R27

ComiR:
In https://github.com/neos/flow-development-collection/blob/fbd370f445e0d6eb62f2d62b0354d213fa1e3ad3/Neos.Flow/Classes/Reflection/ReflectionService.php#L1464 (and the line before) there should be `$typeWithoutNull` instead of `$type`.

gerhard-boden:
> In
> 
> https://github.com/neos/flow-development-collection/blob/fbd370f445e0d6eb62f2d62b0354d213fa1e3ad3/Neos.Flow/Classes/Reflection/ReflectionService.php#L1464
> 
> (and the line before) there should be `$typeWithoutNull` instead of `$type`.

Great catch @ComiR, tried the adjustment locally and the `ReflectionService` passes during bootstrap

albe:
Fixed that too, yet I still wonder why tests did not fail with the added nullable property fixtures...

ComiR:
Because none of those used a fully qualified class name.

kdambekalns:
Travis failure is psalm, to be fixed in another PR already.

PR #1782, opened 2019-09-20
TASK: Relax option type strictness in AbstractBackend and RedisBackend

kdambekalns:


Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -396,13 +396,13 @@ public function isFrozen(): bool
...
-     * @param integer $lifetime Default lifetime of this cache backend in seconds. If NULL is specified, the default lifetime is used. 0 means unlimited lifetime.
+     * @param integer|string|null $lifetime Default lifetime of this cache backend in seconds. If NULL is specified, the default lifetime is used. 0 means unlimited lifetime.
      * @return void
      * @api
      */
-    public function setDefaultLifetime(int $lifetime): void
+    public function setDefaultLifetime($lifetime): void
     {
-        $this->defaultLifetime = $lifetime;
+        $this->defaultLifetime = $lifetime === null ? null : (int)$lifetime;
```
robertlemke: How about using `is_numeric()` here, as a safeguard against stupid strings? We could throw an `InvalidArgumentException` if it's not null or numeric.

kdambekalns: Given that so far we didn't safeguard against anything here… and it would probably be cast to `0`, which means unlimited. Fine with me. ;)

kdambekalns:
> like `setUsername` in the pdoBackend when set to null?

Well, but that is easier to solve: `?string`

kdambekalns:
Travis failed in psalm analysis.

Neos.Cache/Classes/Backend/RedisBackend.php
kdambekalns: ✅

kdambekalns:
Ok, I changed the fix, so it applies to all cache backends when it comes to `setDefaultLifetime(…)`

kdambekalns:
Haha, moving this fix to the `AbstractBackend` would need changes to the other backends, where the default lifetime is always expected to be a positive integer. I guess I'll just move it back to the `RedisBackend`…

psalm-baseline.xml
```diff
@@ -49,28 +49,6 @@
       <code>null</code>
     </PossiblyNullPropertyAssignmentValue>
   </file>
-  <file src="Packages/Framework/Neos.Cache/Classes/Backend/RedisBackend.php">
```
albe: Hmm, still wonder why your psalm always removes those errors...

kdambekalns: Because I have the Redis extension, so it's not undefined?

kdambekalns: Let's try with annotations… :)

kdambekalns: Let's try with annotations on class instead of property level…

kdambekalns: Oh, 'cmon, psalm. That sucks.

PR #1787, opened 2019-09-24
BUGFIX: Undo `dirname` simplification in getScriptRequestPath

### LINKED ISSUE #1785: RequestInformationHelper creates wrong base Uri on windows
Original report by Patrice Kaufmann on slack.

### Description

The implementation of generateBaseUri and getScriptResourcePath now leads to `/\/` as a baseUri instead of `/` It looks like the comment in the 5.3 php file `// FIXME: Shouldn't this be a simple dirname` on `getScriptRequestPathAndFilename` was wrong and it should not have been a simple dirname :wink:

### Steps to Reproduce

1. Have a windows machine
2. Generate an absolute URI, which involves generating a BaseURI

#### Expected behavior

A proper URL

#### Actual behavior

`/\/`

### Affected Versions

Flow: 6.0

albe:
This was broken in the PSR7 move, by replacing the explode/array_pop/implode logic with a simple dirname. On Windows, you in most cases end up with a single `\` which in turn results in generating URLs like `/\/`. This change simply changes the code back to the pre-PSR state, which worked fine.

Resolves #1785

### DIFF
--- a/Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
+++ b/Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
@@ -51,8 +51,10 @@ abstract class RequestInformationHelper
      */
     public static function getScriptRequestPath(ServerRequestInterface $request): string
     {
-        $requestPath = dirname(self::getScriptRequestPathAndFilename($request));
-        return rtrim($requestPath, '/') . '/';
+        // This is not a simple `dirname()` because on Windows it will end up with backslashes in the URL
+        $requestPathSegments = explode('/', self::getScriptRequestPathAndFilename($request));
+        array_pop($requestPathSegments);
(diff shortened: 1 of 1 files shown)


albe:
For reference to the old (and new) version of the code, see: https://github.com/neos/flow-development-collection/blob/80c317bd152d97b8171ddd32a0ed6b9e67a615de/Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php#L55-L59

albe:
Everything worked fine before 6.0 and the code is the same again :) Anyway, just verified locally with a fresh installation and `server:run` - with 6.0 I get a 404 
image, with this change everything's fine again

kdambekalns:
The failing psalm check is fixed with https://github.com/neos/flow-development-collection/pull/1782/commits/0045b078da8c412c57d39e0d77f3323cdd0b3213

PR #1788, opened 2019-09-24
TASK: Remove unneeded code section

albe:
The call to `TypeHandling::parse` is already done inside `ClassSchema::addProperty` and throws a very similar exception.
As a side effect, this should have a very small performance improvement for class schema building/compile time.


albe:
See https://github.com/neos/flow-development-collection/pull/1780#issuecomment-533462511

PR #1795, opened 2019-09-26
TASK: Fix documented return type for ViewInterface.render()

kdambekalns:
The `render()` in `ViewInterface` method used to return a string,
but now that is no longer true. This makes the docblock list the
return types that our `ActionController` handles, to lessen the
potential for confusion.

### DIFF
--- a/Neos.Flow/Classes/Mvc/View/ViewInterface.php
+++ b/Neos.Flow/Classes/Mvc/View/ViewInterface.php
@@ -63,7 +63,7 @@ interface ViewInterface
     /**
      * Renders the view
      *
-     * @return string|ActionResponse|ResponseInterface|StreamInterface|object The rendered result; object is only handled of __toString() exists!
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Mvc/View/ViewInterface.php
```diff
@@ -59,7 +63,7 @@ public function canRender(ControllerContext $controllerContext);
     /**
      * Renders the view
      *
-     * @return string The rendered view
+     * @return string|ActionResponse|ResponseInterface|StreamInterface|object The rendered result
```
bwaidelich: the `|object` part scares me a bit :)
wouldn't `string|obect` be sufficient then?

kdambekalns: Well, those all have distinct handling… for a generic object a check against availalability of `__toString()` is done.

bwaidelich: Fair enough

PR #1796, opened 2019-09-26
BUGFIX: Check for AbstractLazyCollection instead of PersistentCollect…

twojtylak:
Use more generic `AbstractLazyCollection` instead of  `PersistentCollection` in `isValid` of `CollectionValidator` to prevent lazy collection of classes extending `AbstractLazyCollection` to get loaded.

### DIFF
--- a/Neos.Flow/Classes/Validation/Validator/CollectionValidator.php
+++ b/Neos.Flow/Classes/Validation/Validator/CollectionValidator.php
@@ -50,7 +50,7 @@ class CollectionValidator extends GenericObjectValidator
      */
     protected function isValid($value)
     {
-        if ($value instanceof \Doctrine\ORM\PersistentCollection && !$value->isInitialized()) {
+        if ($value instanceof \Doctrine\Common\Collections\AbstractLazyCollection && !$value->isInitialized()) {
             return;
         } elseif ((is_object($value) && !TypeHandling::isCollectionType(get_class($value))) && !is_array($value)) {
             $this->addError('The given subject was not a collection.', 1317204797);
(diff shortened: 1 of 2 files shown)


albe:
Could you provide a use case, optimally in the form of a test fixture, that shows why PersistentCollection is not sufficient?

twojtylak:
Hello @albe 
In our case we are extending the `AbstractLazyCollection` to add some additional functionality when handling lazy collections. Due to the  `final` Implementation of `PersistentCollection` we are not able to extend this class. 
During the `isValid` check, all data in the collection is loaded from the database and we run into an out of memory error, because our collection is an instance of `AbstractLazyCollection` and not `PersistentCollection`.
I added an additional unit test `collectionValidatorIsValidEarlyReturnsOnUnitializedDoctrineAbstractLazyCollections´ to the pr.

albe:
Ah, I see. That is a valid use case and I didn't notice the final on PersistentCollection. Thanks a lot also for the test! 
So, the next question is if that is really considered a BUGFIX, in which case it should be targeted at 4.3, or rather a feature and should land in 6.1 (master).

albe:
Ok, I was having the same feeling. That means this change needs to be retargeted to 4.3 - @twojtylak can you take care? You need to close this PR, then cherry pick the change into a new branch based on 4.3 and create a new PR

twojtylak:
@albe @kdambekalns 

> @twojtylak can you take care? You need to close this PR, then cherry pick the change into a new branch based on 4.3 and create a new PR


Sure. I'll take care of it. Thank's for your feedback

twojtylak:
@albe @kdambekalns #1798 is now targeting 4.3

PR #1799, opened 2019-10-01
TASK: Set psalm baseline

kdambekalns:
This corrects some false positives

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/AbstractTypeConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/AbstractTypeConverter.php
@@ -119,7 +119,7 @@ abstract class AbstractTypeConverter implements TypeConverterInterface
      * Returns an empty list of sub property names
      *
      * @param mixed $source
-     * @return array<string>
+     * @return array<mixed>
      * @api
      */
     public function getSourceChildPropertiesToBeConverted($source)
--- a/Neos.Flow/Classes/Aop/Pointcut/PointcutMethodNameFilter.php
+++ b/Neos.Flow/Classes/Aop/Pointcut/PointcutMethodNameFilter.php
@@ -38,9 +38,9 @@ class PointcutMethodNameFilter implements PointcutFilterInterface
     protected $methodNameFilterExpression;
 
     /**
-     * @var string The method visibility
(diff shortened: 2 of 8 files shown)


kdambekalns:
Note: it does make a difference, which psalm version is used. I had 3.5.1 and could not reproduce the failures, with 3.5.3 I could. 🤷‍♂️

Neos.Flow/Classes/Property/TypeConverter/AbstractTypeConverter.php
```diff
@@ -119,7 +119,7 @@ public function canConvertFrom($source, $targetType)
      * Returns an empty list of sub property names
      *
      * @param mixed $source
-     * @return array<string>
+     * @return array<mixed>
```
albe: Is this really better than just `array`?

albe:
> Note: it does make a difference, which psalm version is used. I had 3.5.1 and could not reproduce the failures, with 3.5.3 I could. 🤷‍♂️

Yeah, I think we need to pin psalm versions to make results more reproducible, then update more frequently.

Neos.Flow/Classes/Property/TypeConverter/AbstractTypeConverter.php
kdambekalns: No, but it is what the interface uses: https://github.com/neos/flow-development-collection/blob/50166ec3fa9dc13a92ba6b1babae6f99318d3cc3/Neos.Flow/Classes/Property/TypeConverterInterface.php#L78 🤷‍♂️

kdambekalns:
> > Note: it does make a difference, which psalm version is used. I had 3.5.1 and could not reproduce the failures, with 3.5.3 I could. 🤷‍♂️
> 
> Yeah, I think we need to pin psalm versions to make results more reproducible, then update more frequently.

If we added it to the dev dist, it would update automatically… I just have it installed globally, to avoid changing the composer manifest "for no reason"…

Neos.Flow/Classes/Property/TypeConverter/AbstractTypeConverter.php
albe: Hmm. Okay then 🐑

PR #1801, opened 2019-10-04
BUGFIX: Allow to return PSR7 streams and PHP resources/file handles from action

### LINKED ISSUE #1800: Allow to return streams (or resources) from an action
Since Flow 6.0 we internally have everything needed to make an action return responses as PSR7 streams or anything that can be transformed to one. We even convert a string to a stream internally via `GuzzleHttp\Psr7\stream_for` (https://github.com/guzzle/psr7/blob/master/README.md#function-stream_for).

However, if an action does not return a string or object with `__toString()`, the return value will be ignored. If null/nothing is returned the `ActionController` will render the view and set the result as the content for the response.
So to bypass that and have the action return a stream, one has to do s.th. like this currently:
```php
$this->response->setContent(fopen('/path/to/file', 'r'));
$this->view = null; // D'oh
return;
```

Instead it should just be possible to do this:
```php
return fopen('/path/to/file', 'r');
```
or if you already have a PSR stream, return that directly.

Hence I suggest to lift the strict action result checks and pass anything not null or a "stringifyable object" to `$this->response->setContent()` unchanged.

See https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Mvc/Controller/ActionController.php#L505-L511
and
https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Mvc/ActionResponse.php#L66-L68

albe:
This allows to return PSR7 streams, PHP resources or anything else that guzzle stream_for can deal with from the action.
https://github.com/guzzle/psr7/blob/master/README.md#function-stream_for

With that it is possible to just do this in an action:
```php
return fopen('/path/to/file', 'r');
```

Resolves #1800

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/ActionController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/ActionController.php
@@ -504,10 +504,8 @@ class ActionController extends AbstractController
 
         if ($actionResult === null && $this->view instanceof ViewInterface) {
             $this->renderView();
-        } elseif (is_string($actionResult) && strlen($actionResult) > 0) {
+        } else {
             $this->response->setContent($actionResult);
-        } elseif (is_object($actionResult) && method_exists($actionResult, '__toString')) {
-            $this->response->setContent((string)$actionResult);
         }
     }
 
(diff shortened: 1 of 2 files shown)


albe:
Marked as "BUGFIX" because this should really be in 6.0

Issue #1803, opened 2019-10-04
fix double "//" encoding in some URLs


skurfuerst:
## Problem Description

(this is hard to explain, checked into this with @bwaidelich).

In case all of the following conditions are met::

1) RoutePartHandler with custom hostname
   - you have a custom RoutePartHandler, which...
   - returns a `ResolveResult`, which ...
   - switches to a different hostname (using `->withHost(...)`
2) and you create a link through through this route,

then: the URL contains two slashes after the hostname (i.e. `http://my-host//test`).

## Steps to Reproduce

(see test case)

### Preamble 

With Flow 6.0, a very subtle change to the URL handling happened:

- For Flow < 6.0, we had our own Uri implementation, where the withPath() method looked like this: https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Classes/Http/Uri.php#L577-L600.
- For Flow = 6.0, we switched to `GuzzleHttp\Psr7\Uri`, whose method looks like: https://github.com/guzzle/psr7/blob/1.x/src/Uri.php#L487
- pay special attention to the `validateState()` method, which contains this block:
   https://github.com/guzzle/psr7/blob/2595b33c1c924889b474d324f3d719fa40b6954e/src/Uri.php#L750-L757

```php
if ($this->getAuthority() === '') {
...
} elseif (isset($this->path[0]) && $this->path[0] !== '/') {
            @trigger_error(
                'The path of a URI with an authority must start with a slash "/" or be empty. Automagically fixing the URI ' .
                'by adding a leading slash to the path is deprecated since version 1.4 and will throw an exception instead.',
                E_USER_DEPRECATED
            );
            $this->path = '/'. $this->path;
            //throw new \InvalidArgumentException('The path of a URI with an authority must start with a slash "/" or be empty');
        }
```

**This means Guzzle normalizes all URL parts to start with leading slashes, but only if we have an authority set (i.e. a host).**

### Usage in Flow

In Flow, we had the following logic:

- UriBuilder::build() contains a call to RequestInformationHelper::getScriptRequestPath(), which always generates a path starting with a "/". Thus, `$uriPathPrefix` there contains **at least** a `/`.
- **Additional Gimmick**: In the constructor of `ResolveContext`, some logic has been added with #1215 ( + #1185 ), which lead to the situation that **even if $uriPathPrefix would be empty and $forceAbsoluteUri would be `true`, *the `$uriPathPrefix` would still be set to `/`***
[...]

PR #1804, opened 2019-10-04
BUGFIX: fix double "//" encoding in some URLs

skurfuerst:
## Problem Description

(this is hard to explain, checked into this with @bwaidelich).

In case all of the following conditions are met::

1) RoutePartHandler with custom hostname
   - you have a custom RoutePartHandler, which...
   - returns a `ResolveResult`, which ...
   - switches to a different hostname (using `->withHost(...)`
2) and you create a link through through this route,

then: the URL contains two slashes after the hostname (i.e. `http://my-host//test`).

## Steps to Reproduce

(see test case)

### Preamble

With Flow 6.0, a very subtle change to the URL handling happened:

- For Flow < 6.0, we had our own Uri implementation, where the withPath() method looked like this: https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Classes/Http/Uri.php#L577-L600.
- For Flow = 6.0, we switched to `GuzzleHttp\Psr7\Uri`, whose method looks like: https://github.com/guzzle/psr7/blob/1.x/src/Uri.php#L487
- pay special attention to the `validateState()` method, which contains this block:
   https://github.com/guzzle/psr7/blob/2595b33c1c924889b474d324f3d719fa40b6954e/src/Uri.php#L750-L757

```php
if ($this->getAuthority() === '') {
...
} elseif (isset($this->path[0]) && $this->path[0] !== '/') {
            @trigger_error(
                'The path of a URI with an authority must start with a slash "/" or be empty. Automagically fixing the URI ' .
                'by adding a leading slash to the path is deprecated since version 1.4 and will throw an exception instead.',
                E_USER_DEPRECATED
            );
            $this->path = '/'. $this->path;
            //throw new \InvalidArgumentException('The path of a URI with an authority must start with a slash "/" or be empty');
        }
```

**This means Guzzle normalizes all URL parts to start with leading slashes, but only if we have an authority set (i.e. a host).**

### Usage in Flow

In Flow, we had the following logic:

- UriBuilder::build() contains a call to RequestInformationHelper::getScriptRequestPath(), which always generates a path starting with a "/". Thus, `$uriPathPrefix` there contains **at least** a `/`.
- **Additional Gimmick**: In the constructor of `ResolveContext`, some logic has been added with #1215 ( + #1185 ), which lead to the situation that **even if $uriPathPrefix would be empty and $forceAbsoluteUri would be `true`, *the `$uriPathPrefix` would still be set to `/`***
- This is then used in `Router::resolve()`, where the path prefix (`/`) from the afoementioned ResolveContext is applied to the `$uriConstraints`.
- Now, in `UriConstraints::applyTo()`, the following block is executed:

```php
        if (isset($this->constraints[self::CONSTRAINT_PATH]) && $this->constraints[self::CONSTRAINT_PATH] !== $templateUri->getPath()) {
            $uri = $uri->withPath($this->constraints[self::CONSTRAINT_PATH]);
        }
        if (isset($this->constraints[self::CONSTRAINT_PATH_PREFIX])) {
            $uri = $uri->withPath($this->constraints[self::CONSTRAINT_PATH_PREFIX] . $uri->getPath());
        }
```

In the first `if`-block, the URI path is set. Normally, this is a *non-absolute path*, but **guzzle normalizes this to be an absolute path** (see the section above).

Then comes the second `if`-block, where a `/` is prepended to the absolute path, leading to the `//` situation.

**NOTE: This error can only be triggered when the RoutePartHandler returns a *different* host than the original one; as otherwise, the auto-prefixing is not done in Guzzle.**.

## How to fix this

- We simply trim the `$uriPathPrefix` to not start with `/` in the UriBuilder.
- Additionally, we add safeguards in `ResolveContext` so that the $uriPathPrefix is never starting with a `/`.
- Additionally, we add safeguards in `UriConstraints` so that the `withPathPrefix` is never generating a prefix starting with a `/`.

Finally, we do a **cleanup**; moving code related to base-URLs-with-paths around a bit (see #1215 + #1185 ); so that this is actually done in the `UriConstraints::applyTo` method.

### Affected Versions

Flow 6.0.0

Resolves: #1803

**What I did**

**How I did it**

**How to verify it**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
@@ -54,7 +54,7 @@ final class ResolveContext
      * @param UriInterface $baseUri The base URI, retrieved from the current request URI or from configuration, if specified. Required to fill in parts of the result when resolving absolute URIs
      * @param array $routeValues Route values to build the URI, for example ['@action' => 'index', 'someArgument' => 'foo', ...]
      * @param bool $forceAbsoluteUri Whether or not an absolute URI is to be returned
-     * @param string $uriPathPrefix A prefix to be prepended to any resolved URI
+     * @param string $uriPathPrefix A prefix to be prepended to any resolved URI. Not allowed to start with "/".
      */
     public function __construct(UriInterface $baseUri, array $routeValues, bool $forceAbsoluteUri, string $uriPathPrefix = '')
     {
@@ -63,13 +63,8 @@ final class ResolveContext
         $this->forceAbsoluteUri = $forceAbsoluteUri;
         $this->uriPathPrefix = $uriPathPrefix;
 
-        // Only add base uri path for absolute uri, in case of relative uri the uri has to be relative to the given base uri
-        if ($forceAbsoluteUri) {
-            $this->uriPathPrefix = '/' . ltrim($this->uriPathPrefix, '/');
-
-            if ($baseUri->getPath() !== '') {
-                $this->uriPathPrefix = rtrim($baseUri->getPath(), '/') . $this->uriPathPrefix;
-            }
+        if (strpos($this->uriPathPrefix, '/') === 0) {
+            throw new \InvalidArgumentException('UriPathPrefix "' . $uriPathPrefix . '" is not allowed to start with "/".', 1570187176);
         }
     }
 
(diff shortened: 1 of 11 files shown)


skurfuerst:
## still found a bug, do not merge yet!

skurfuerst:
## Grml. Found another edge case. You can already review, but I'll push another case ;)

skurfuerst:
Follow-Up: https://github.com/neos/flow-development-collection/pull/1839

Issue #1807, opened 2019-10-07
FlashMessages are not displayed with redirect


albe:
## Description

When I `$this->addFlashMessage('foo');` and then `$this->redirect('otherAction');` in Flow 6, there is no flash message anymore.

Apparently, redirecting and setting a cookie at the same time, is buggy at least in Chrome:
https://bugs.chromium.org/p/chromium/issues/detail?id=150066
And hard to reproduce in a way that the Chromium team can fix it:
https://bugs.chromium.org/p/chromium/issues/detail?id=696204#c72 

Technically, that was possible to happen before, so probably we just changed semantics on when a session is started and hence setting a flash message is the first place much more often.

Changing the FlashMessage storage to `SessionStorage` instead of `CookieStorage` doesn't really solve the issue according to feedback.

What we need is more information on the environments where this happens, if it happens every time independent of http/https or has some conditions.

### Affected Versions

Flow: 6.0

bwaidelich:
I doubt that this happens due to a browser bug.
I just tested the following script Chrome and it worked for me:
```php
<?php
$flashMessage = $_COOKIE['flashmessage'] ?? null;
if ($flashMessage === null) {
    setcookie('flashmessage', 'Some Flash Message');
    header('Location: http://localhost:8000', true, 303);
    exit;
}
echo $flashMessage;
```

It would be helpful to see what headers are actually sent with the redirected response

albe:
Yeah, that fits with the assessment of the chrome dev, that it is not easily reproducible. And pretty sure something in Flow 6 causes this to occur out of a sudden.

albe:
After some digging thanks to Mark Kuiphuis, there are two issues at play:
- the flash message http component is not configured by default (the configuration got lost in rebasing)
- the cookie storage (which is default right now) for some reason voids the session cookie, so it only works if the component comes 'before setSessionCookie', which is not intended

I will take care and provide a bugfix PR that:
- adds the missing component configuration
- sets the default flash message storage to `SessionStorage`, which is closer to previous behaviour - we can switch the default back in a later Flow version when cookie storage edge cases have been cleaned up

albe:
Closing as of #1811

SC-cmarquardt:
Hi @albe ,

for me your fix does not solve the problem.
If I redirect during login or logout all flash messages are lost.
Have you any idea how to debug / backtrace this issue?

Kind regards,
Christopher

albe:
[...]

PR #1808, opened 2019-10-07
BUGFIX: Don't overwrite previous Set-Cookie headers in the session component

albe:
This would e.g. overwrite cookies set with the FlashMessage CookieStorage.

Related to #1807


bwaidelich:
Makes sense, but since this is rather critical some (functional) test would be nice.

bwaidelich:
Also this bug affects older versions of Flow as well it seems: https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Classes/Session/Http/SessionResponseComponent.php

albe:
Closing in favor of #1810 targeting 5.1 where we introduced the SessionResponseComponent

PR #1809, opened 2019-10-07
BUGFIX: Don't overwrite previous Set-Cookie headers in the session component

albe:
This would e.g. overwrite cookies set with the FlashMessage CookieStorage.

Related to #1807

Retargeted version of #1808


albe:
Goes even back to 5.1...

PR #1811, opened 2019-10-07
BUGFIX: Add the FlashMessageComponent to http chain

### LINKED ISSUE #1807: FlashMessages are not displayed with redirect

albe:
This unfortunately got lost when rebasing the feature and was unnoticed till now.

Needs to be 'after setSessionCookie' because that component always overrides the 'Set-Cookie' header right now (see #1810). Apparently, flash message `CookieStorage` doesn't play well with this, hence we also need to change the default storage to `SessionStorage` at least until that is resolved.

Resolves #1807

### DIFF
--- a/Neos.Flow/Configuration/Settings.Http.yaml
+++ b/Neos.Flow/Configuration/Settings.Http.yaml
@@ -67,6 +67,9 @@ Neos:
             'setSessionCookie':
               position: 'start'
               component: 'Neos\Flow\Session\Http\SessionResponseComponent'
+            'flashMessages':
+              position: 'after setSessionCookie'
+              component: 'Neos\Flow\Mvc\FlashMessage\FlashMessageComponent'
             'poweredByHeader':
               component: 'Neos\Flow\Http\Component\PoweredByComponent'
             'standardsCompliance':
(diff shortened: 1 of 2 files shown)


albe:
That's a rather urgent bugfix, as it currently completely makes flash messages unusable, without manual (undocumented) adjustments.

PR #1813, opened 2019-10-10
TASK: Clarify explanation of ignoredTables setting

kdambekalns:


kdambekalns:
Well, not really wroing, just overly complicated and not as clear as it could be. I tunneled myself…

PR #1814, opened 2019-10-11
FEATURE: Add optional nonblocking locking

gjwnc:
I added an optional parameter for nonblocking locks. If the lock could not be acquired, the method `applyFlock` throws an Exception.

Before this non blocking feature, it was not possible, that `applyFlock` throws an Exception, since flock always blocked and thus always returned true after successfully acquiring the lock. So this is kind of a bug in the expected interface of the implementation.

### DIFF
--- a/Neos.Utility.Lock/Classes/FlockLockStrategy.php
+++ b/Neos.Utility.Lock/Classes/FlockLockStrategy.php
@@ -138,7 +138,7 @@ class FlockLockStrategy implements LockStrategyInterface
     protected function applyFlock(bool $exclusiveLock, bool $nonblocking)
     {
         $lockOption = $exclusiveLock === true ? LOCK_EX : LOCK_SH;
-        if ($nonblocking){
+        if ($nonblocking) {
             $lockOption |= LOCK_NB;
         }
 
--- a/Neos.Utility.Lock/Classes/FlockLockStrategy.php
+++ b/Neos.Utility.Lock/Classes/FlockLockStrategy.php
@@ -63,17 +63,18 @@ class FlockLockStrategy implements LockStrategyInterface
 
     /**
      * @param string $subject
-     * @param boolean $exclusiveLock TRUE to, acquire an exclusive (write) lock, FALSE for a shared (read) lock.
+     * @param boolean $exclusiveLock true to, acquire an exclusive (write) lock, false for a shared (read) lock.
+     * @param boolean $nonblocking true to, acquire the lock in nonblocking mode, false for a blocking lock lock.
      * @return void
      * @throws LockNotAcquiredException
      */
-    public function acquire($subject, $exclusiveLock)
+    public function acquire(string $subject, bool $exclusiveLock, bool $nonblocking = false)
     {
         $this->lockFileName = Utility\Files::concatenatePaths([$this->temporaryDirectory, md5($subject)]);
         $aquiredLock = false;
         $i = 0;
         while ($aquiredLock === false) {
-            $aquiredLock = $this->tryToAcquireLock($exclusiveLock);
+            $aquiredLock = $this->tryToAcquireLock($exclusiveLock, $nonblocking);
             $i++;
             if ($i > 10000) {
                 throw new LockNotAcquiredException(sprintf('After 10000 attempts a lock could not be aquired for subject "%s".', $subject), 1449829188);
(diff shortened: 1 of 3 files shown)


kitsunet:
Changing an interface in a bugfix is a big no, no, that automatically makes it a breaking change meaning this can only go into a major release, so not 4.3 for sure and 5.0 is already released. But 5.0 also doesn't contain the package anymore. So it would would need to go into the lock package master and someone would need to do a major release of it afterwards.

gjwnc:
@kitsunet I understand your concerns. The reason I added the new parameter `$nonblocking` with a default value of `false` is, that this would change nothing for the code that is built upon those locks.  
So, yeah, I added a new parameter and thus change the interface, but it leaves old code untouched because of the default value and the behaviour also doesnt change. Is this also a no-go here or could this still make it into 4.3? Or did you mean something else with `changing an interface`?

The idea was to provide the possibility to fix code which might need this feature, which I know one example.

kitsunet:
Interfaces are there to be public extensibility points, changing the signature will break implementations of the interface because they no longer use the same method signature, therefore we have a pretty strict policy on interfaces. In the past situations of "probably no one is using this" let to broken projects and I would like to prevent that. So yes, sadly I don't see this. But given that this is no longer part of the development distribution and not maintained basically, we could make the change master of this package and release a new major version that doesn't change the requirements to other packages. Means you should be able to install it into a 4.3 installation.

gjwnc:
Ah, right. Should I create a new PR for the master branch?

albe:
Hey, thanks a lot for your contribution. As Christian said this is to be considered breaking and needs to target master.
Fortunately changing the base branch upwards is easier with our merging strategies. In most cases it just works without further changes. In the given case you need to merge master into this branch (resolve conflicts) and push again. But you can also recreate the change on top of master in a new PR if you prefer :)

gjwnc:
@albe I just wanted to fix the conflicts and noticed that the directory `Neos.Utility.Lock` does not exist in master. Where should my changes go now? The readme from https://github.com/neos/utility-lock mentions that PR's should go here in the flow-development-collection. @kdambekalns mentioned in slack, that the lock package was removed starting with flow 6.0. So where should my PR go?

Here in flow-development-collection where I readd those files? Or in the neos/utility-lock where the readme mentions it's only a read only repo?

kdambekalns:
> But 5.0 also doesn't contain the package anymore.

It was @kitsunet saying that again above, but with the branch being off-by-one…

So, in theory this should go into 5.3 (the latest branch containing the package), but since it's a feature it won't be allowed to be merged there. And we will never release a version 5.4 of Flow.

One option would be to open a new PR against the (formerly read-only) split repo and tag a 5.4 release of the lock package from there. But to be honest, I think we will not maintain that as a team, so… not sure how viable an option that is.

albe:
Ah yes, of course, it's already gone in 6.0+. So yes, if at all, a PR should go to 5.3 (if we'd make an exemption but I'd rather not) or to the "read-only" split https://github.com/neos/utility-lock and we tag a release there. Unsure myself, but I'd be willing to push the "tag a release" button in the read-only one last time (and we should update the README there too)

gjwnc:
I'm a bit confused now. Regarding the code changes itself, if I understand this discussion here, I'll add a new PR for the read only repo neos/utility-lock.

What I don't get is:  
The utility-lock was removed from flow-dev-collection, thus it only resides in its own repo, thus further changes will go there. @albe What do you mean by "push the tag a release button ... one last time"? Why one last time? Am I wrong in the assumption, that locks will still exist in Flow 6.0 or has it been replaced by something else?

albe:
Sorry @gjwnc I totally lost track of this PR!

Yes, my idea was to create a PR in the "read-only" neos/utility-lock repository, because this is a feature which needs to land in master, but we do not have the lock utility any more in master. So if we merge some changes to the neos/utility-lock repository master, we can just tag a new version there, that would then be installable (e.g. as version "6.0" or even "5.4" since it's not breaking).

> Why one last time? Am I wrong in the assumption, that locks will still exist in Flow 6.0 or has it been replaced by something else?

One last time in the sense that the neos/utility-lock package is no longer supported/maintained by the team. It was not used in the core anyway since I think somewhere in 5.x or so. We still use locks in Flow 6.x for the compilation and on filesystem cache backend, but those are implemented without the utility package, which was supposed to be general-purpose. However, locking - especially on PHP - is not easily done in a GP way (as is shown by the missing non-blocking lock till now) and it's definitely not our focus to build and maintain one. I think there are a couple of other 3rd party composer packages around for locking, which we'd advise to use instead.

Having said all that, if you are using neos/utility-lock still and just need this feature to keep it, going with the above suggestion to create a PR in the read-only would be a solution.

gjwnc:
@albe Since neos/utility-lock is not maintained and not used anymore I would suggest to close this PR without merging. My knowledge about neos and flow also evolved since I opened this PR and now I think I could solve the problems I have/had with an external package using a custom lock strategy and/or Objects.yaml.

PR #1817, opened 2019-10-13
TASK: Allow finer control of accepted read and write latency of FileBackends

fcool:
This change raises the default maximal wait time for FileBackend operations to 10 ms, while not affecting the performance on filesystems, which are able to react in the previous latency window.

Additionally it minimizes the random factor of waiting, but maintains the randomized acquiring strategy to let eventual races settle, by repeating the randomized wait time until a fixed amount of time has passed.

The amount of time passing is configurable for read and write latency.

### DIFF
--- a/Neos.Cache/Classes/Backend/SimpleFileBackend.php
+++ b/Neos.Cache/Classes/Backend/SimpleFileBackend.php
@@ -83,6 +83,20 @@ class SimpleFileBackend extends IndependentAbstractBackend implements PhpCapable
     protected $baseDirectory;
 
     /**
+     * Maximal amount of microseconds allowed to spent acquiring a read lock
+     *
+     * @var int
+     */
+    protected $maxAcceptedReadLatencyMicroSeconds = 10000;
+
+    /**
+     * Maximal amount of microseconds allowed to spent acquiring a read lock
+     *
+     * @var int
+     */
+    protected $maxAcceptedWriteLatencyMicroSeconds = 10000;
+
+    /**
      * {@inheritdoc}
      */
     public function __construct(EnvironmentConfiguration $environmentConfiguration, array $options = [])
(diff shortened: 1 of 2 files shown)


Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -520,7 +585,9 @@ protected function readCacheFile(string $cacheEntryPathAndFilename, int $offset
      */
     protected function writeCacheFile(string $cacheEntryPathAndFilename, string $data)
     {
-        for ($i = 0; $i < 3; $i++) {
+        $microtime = microtime(true);
+        $acceptedLatencyInSeconds = $this->maxAcceptedWriteLatencyMicroSeconds / 1000000;
```
albe: We should probably also calculate the max random sleep based on the max latency. As is, this currently does `10.000 / ((500+10) / 2) = 40` attempts on average and 1.000 in the worst case with default config, which feels a bit much. Also, this increases if the max wait is increased and effectively ends in a busy wait cycle. If we limit max to e.g. `maxAcceptedWriteLatencyMicroSeconds / 5` we have on avg. 10 attempts (but still 1.000 in the unlikely worst case). Or we also scale the minimum usleep scaled by `maxAcceptedWriteLatencyMicroSeconds / 20`? Then we have avg. 8 attempts and 20 in the worst case.

Neos.Cache/Classes/Backend/SimpleFileBackend.php
```diff
@@ -484,7 +547,9 @@ protected function configureCacheDirectory()
      */
     protected function readCacheFile(string $cacheEntryPathAndFilename, int $offset = null, int $maxlen = null)
     {
-        for ($i = 0; $i < 3; $i++) {
+        $tries = 10;
+        $waitTime = $this->maxAcceptedReadLatencyMicroSeconds / $tries;
+        for ($i = 0; $i < $tries; $i++) {
```
albe:
```suggestion
        // We do a fixed amount of retries with a fixed sleep because reads do not block each other out unlike writes
        for ($i = 0; $i < $tries; $i++) {
```

albe (commented):
Left a few comments and suggestions. The biggest thing to decide is probably the write sleep min/max, as it is now relatively low in relation to the default max wait, so leads to potentially a lot of cycles.

Neos.Cache/Classes/Backend/SimpleFileBackend.php
fcool: Uh! Completely right!Valid point. What about adjusting both boundaries depending on the maximal values by calculating an expected value to match a given amount of tries?

something like:
```
        $targetedTries = 5;
        $waitTimeScatter = 0.5;
        $expectedWaitTimeAverage = $acceptedLatencyInSeconds / $targetedTries;
        $halfExpectedWaitTimeAverage = $expectedWaitTimeAverage * $waitTimeScatter;
        $minimalWaitTime = $expectedWaitTimeAverage - $halfExpectedWaitTimeAverage;
        $maximalWaitTime = $expectedWaitTimeAverage + $halfExpectedWaitTimeAverage;
```

And as these calculations will be done often, maybe they should go to the setters, to not repeat them.
And of course, we have to decide if tries and scatter is something which should be configurable

albe: Not sure that's straightforward in this probablistic case. Configuring a "targetTries" is probably something that people don't know what to do with or want to know how to configure the allowed deviation (scatter). But let's not overcomplicate it too much (for the user) - being able to configure a max latency is clear, gives perf. guarantees and works. The number of retries can just be picked with a senseful decision and be made explicit.

bwaidelich (changes requested):
Wow, this is one year old and I completely missed it..

IMO it is a *feature* though and thus should go into `master`.
Personally I find the term "latency" a bit misleading in this case.

What do you guys think of continuing this with #1987 (since that already has the correct target branch) or create this change against master and close the other PR?

PR #1818, opened 2019-10-14
TASK: Extract inferDiscriminatorTypeFromClassName() method

kdambekalns:


kdambekalns:
Used in https://github.com/neos/neos-development-collection/pull/2725

PR #1820, opened 2019-10-15
BUGFIX: Fix setting sub-package key in ActionRequest

### LINKED ISSUE #1819: Routing breaks with sub package key being empty string
### Description

When trying to log in to Neos with wrong credentials, an exception is thrown.

### Steps to Reproduce

1. Checkout a Neos dev dist as master
2. Submit wrong credentials in login form

#### Expected behavior

A nice error message about the credentials being wrong.

#### Actual behavior

`Could not resolve a route and its corresponding URI for the given parameters.`

### Affected Versions

Flow: master

kdambekalns:
The ActionRequest needs to set the sub-package key to null when an
empty string is input. This used to be the case earlier and the
handling is effectively just reverted to that behaviour by this.

Fixes #1819

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
+++ b/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
@@ -286,7 +286,7 @@ class UriBuilder
      * @see build()
      * @throws Exception\MissingActionNameException if $actionName parameter is empty
      */
-    public function uriFor(string $actionName, array $controllerArguments = [], string $controllerName = null, string $packageKey = null, string $subPackageKey = null)
+    public function uriFor(?string $actionName, array $controllerArguments = [], ?string $controllerName = null, ?string $packageKey = null, ?string $subPackageKey = null)
     {
         if (empty($actionName)) {
             throw new Exception\MissingActionNameException('The URI Builder could not build a URI linking to an action controller because no action name was specified. Please check the stack trace to see which code or template was requesting the link and check the arguments passed to the URI Builder.', 1354629891);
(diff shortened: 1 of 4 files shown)


Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
```diff
@@ -286,28 +286,25 @@ public function reset()
...
+        if (empty($packageKey) && empty($subPackageKey)) {
             $subPackageKey = $this->request->getControllerSubpackageKey();
         }
-        if ($packageKey === null) {
+        if (empty($packageKey)) {
             $packageKey = $this->request->getControllerPackageKey();
         }
+
+        $controllerArguments['@action'] = strtolower($actionName);
+        $controllerArguments['@controller'] = strtolower(empty($controllerName) ? $this->request->getControllerName() : $controllerName);
```
albe: Maybe replace that with an `if (empty(...))` condition and set controllerName to be in line with the rest of the code

gerhard-boden:
Works IRL but tests fail because of wrong type given here https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Tests/Unit/Mvc/Routing/UriBuilderTest.php#L152 and here https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Tests/Unit/Mvc/Routing/UriBuilderTest.php#L224

albe:
So, allow nullable actionName to stay 100% b/c or break compatiblity by being strict. Guess the former is the "right" way to go. We'll surely (hopefully) revisit things with 7.0

kdambekalns:
Let's break this down:

- `$actionName` must not be empty anyway, so no nullable needed, and it does not have a default
- `$controllerArguments` must be an array if given, null is not a valid value (it is used as an array in the method!)
- `$controllerName` must be a non-empty string if given, null is only allowed to make it optional
- `$packageKey` must be a non-empty string if given, null is only allowed to make it optional
- `$subPackageKey` must be a non-empty string if given, null is only allowed to make it optional

Summary: the tests actually do it wrong, passing an array there has never been right, our tests need to be fixed IMHO.

kdambekalns:
Related to this: https://github.com/neos/neos-development-collection/pull/2733

PR #1821, opened 2019-10-15
BUGFIX: Avoid exception when upcasting errorenous file upload

albe:
Guzzle throws an exception when trying to access the stream of an errored file upload.

### DIFF
--- a/Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php
+++ b/Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php
@@ -57,7 +57,9 @@ abstract class UploadedFilesHelper
      */
     protected static function upcastUploadedFile(UploadedFileInterface $uploadedFile, $originallySubmittedResource = null, string $collectionName = null): FlowUploadedFile
     {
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php
```diff
@@ -57,7 +57,9 @@ public static function upcastUploadedFiles(array $uploadedFiles, array $argument
      */
     protected static function upcastUploadedFile(UploadedFileInterface $uploadedFile, $originallySubmittedResource = null, string $collectionName = null): FlowUploadedFile
     {
-        $flowUploadedFile = new FlowUploadedFile($uploadedFile->getStream(), ($uploadedFile->getSize() ?: 0), $uploadedFile->getError(), $uploadedFile->getClientFilename(), $uploadedFile->getClientMediaType());
+        // If upload failed, just accessing the stream will throwin guzzle
+        $stream = $uploadedFile->getError() === UPLOAD_ERR_OK ? $uploadedFile->getStream() : stream_for(null);
```
gerhard-boden: Is the ``use`` statement for `stream_for` missing?

PR #1825, opened 2019-10-18
FEATURE: Allow split sources for all configurations

### LINKED ISSUE #603: Allow Split Sources for all default configuration types
Jira issue originally created by user @skurfuerst:

I think it would be very helpful to allow Split configuration Sources for Settings, Objects, Policy and Caches -- i.e. everything except Routes (which already allow for inclusion of other files).

This can be changed in ConfigurationManager::$configurationTypes.
### Workaround

In case you need this feature **now** on Flow 2.2, you can do the following:

**create a Settings.SplitSourceMarker.yaml with the following contents:**

```
SettingsSplitSourceActive: TRUE
```

**create the following Package.php**

```
<?php

use TYPO3\Flow\Configuration\ConfigurationManager;
use TYPO3\Flow\Package\Package as BasePackage;

class Package extends BasePackage {

    /****
     * Invokes custom PHP code directly after the package manager has been initialized.
     *
     * @param \TYPO3\Flow\Core\Bootstrap $bootstrap The current bootstrap
     * @return void
     */
    public function boot(\TYPO3\Flow\Core\Bootstrap $bootstrap) {
        $dispatcher = $bootstrap->getSignalSlotDispatcher();

        /****
         * WORKAROUND for now to allow split sources for Settings.
         *
         * - We override the (built-in) configuration type "Settings", enabling "splitSource".
         * - Because Settings are loaded very early in the bootstrap, we need to flush the configuration cache to re-load the Settings.
         *
         * However, we must NOT CLEAR THE CACHE always; thus we have a marker setting "SettingsSplitSourceActive" which is included in Configuration/Settings.SplitSourceMarker.yaml.
         * If this is found in the Settings, we know the split source has been correctly cached; and we do not need to clear it again.
         */
        $dispatcher->connect('TYPO3\Flow\Configuration\ConfigurationManager', 'configurationManagerReady', function(ConfigurationManager $configurationManager) {
            $configurationManager->registerConfigurationType(ConfigurationManager::CONFIGURATION*TYPE_SETTINGS, ConfigurationManager::CONFIGURATION_PROCESSING_TYPE*SETTINGS, TRUE);

            if ($configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION*TYPE*SETTINGS, 'SettingsSplitSourceActive') === NULL) {
                $configurationManager->flushConfigurationCache();
            }
        });
    }
}
```

The solution is slightly inefficient because the settings are parsed twice if they are uncached.

 Jira-URL: https://jira.neos.io/browse/FLOW-42

sorenmalling:
This change sets `$allowSplitSource` to `true` for all configuration types.

In addition it changes the default value for that flag to true in the method
`registerConfigurationType()`.

Resolves #603

### DIFF
--- a/Neos.Flow/Classes/Configuration/ConfigurationManager.php
+++ b/Neos.Flow/Classes/Configuration/ConfigurationManager.php
@@ -117,10 +117,10 @@ class ConfigurationManager
      * @var array
      */
     protected $configurationTypes = [
-        self::CONFIGURATION_TYPE_CACHES => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_DEFAULT, 'allowSplitSource' => false],
-        self::CONFIGURATION_TYPE_OBJECTS => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_OBJECTS, 'allowSplitSource' => false],
-        self::CONFIGURATION_TYPE_ROUTES => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_ROUTES, 'allowSplitSource' => false],
-        self::CONFIGURATION_TYPE_POLICY => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_POLICY, 'allowSplitSource' => false],
+        self::CONFIGURATION_TYPE_CACHES => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_DEFAULT, 'allowSplitSource' => true],
+        self::CONFIGURATION_TYPE_OBJECTS => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_OBJECTS, 'allowSplitSource' => true],
+        self::CONFIGURATION_TYPE_ROUTES => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_ROUTES, 'allowSplitSource' => true],
+        self::CONFIGURATION_TYPE_POLICY => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_POLICY, 'allowSplitSource' => true],
         self::CONFIGURATION_TYPE_SETTINGS => ['processingType' => self::CONFIGURATION_PROCESSING_TYPE_SETTINGS, 'allowSplitSource' => true]
     ];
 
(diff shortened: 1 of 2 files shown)


sorenmalling:
Perhaps worth noticing that the default value of `$allowSplitSource` in `registerConfigurationType` has also been changed to `true`. Is this "bad" in terms of avoiding BC?

sorenmalling:
> Next step, remove that setting completely.

Up for next major?

albe:
As per my comment in https://github.com/neos/flow-development-collection/issues/603#issuecomment-543572038 I'd do it next minor but let's hear what the others are saying

kdambekalns:
How can this be needed with #1193 being in 5.0 already? I am confused…

sorenmalling:
> How can this be needed with #1193 being in 5.0 already? I am confused…

@kdambekalns This is for _all_ configuration Yaml flies (Routes, Objects, Caches) - the
PR you link to is only for Settings.yaml (from my reading)

albe:
Exactly. Sebastian and Bastian too agree that split sources should be available for all configuration types. See comments in the linked issue. But if anyone sees a good reason not to do this I'm happy to hear it!

sorenmalling:
@kdambekalns @albe Can this PR be accepted and merged?

sorenmalling:
> * the 6.0 branch is the wrong target, features in the next minor, which will come off of master - or was 6.0 a conscious decision (and why?)

Branch chosen due to this answer regarding the type of the PR: https://github.com/neos/flow-development-collection/issues/603#issuecomment-543572038

kitsunet:
Right, but the next minor will be 6.1 which is branched off of master (not 6.0) shortly before release. The 6.0 branch is just for 6.0 bugfixes.

I like it. I think the breakiness is manageable even with the default changed, because that would only affect custom configuration types which I guess are rare enough and then people having a custom configuration type would have to have split files around that they didn't want to have parsed for this to become a problem. Seems like a really improbable situation.

albe:
Correct that this needs to target master to land in 6.1 - as said, we always branch off master instead off major version branches. Retargeted this now.

Regarding the Routes config... I didn't realize there could be an issue with making Routes split source, but I can see how this is an issue with manually included sub routes that use the split source naming scheme (not only in Neos). So that basically means we need to keep the setting indefinitely or we hardcode excluding Routes from split sources. I'd favor the latter

albe:
Okay, 6.0 is upmerged to master, so this PR should be clean again after merging master into this and pushing.

sorenmalling:
I’m not clever enough to follow the concept and issues, so I’ll take a break from it all.

sorenmalling:
Too many other things goind on around me, to be able to focus on why this "simple git push" can't work. I seems to have a new experience on each PR. I will look back in next week, if not merged by someone else before than.

albe:
I can take care of fixing/updating the docs later today or tomorrow

davidspiola:
Hi Folks, is this something for the 5.1 relesae?

albe:
IMO yes - just needs @kdambekalns to remove his veto :)

PR #1826, opened 2019-10-18
BUGFIX: Wrap insert statements in a transaction

gerhard-boden:
The PDO backend will run multiple single insert statements for each
cache entry, which can result in high overhead.

When wrapping the statements within a transaction, the cost
is significantly lower depending on the number of generated
cache entries necessary for the current request.

Performance tests show a comparably small impact on the demo page with
a gain of around 100ms - 150ms, and up to 2000ms on a larger product
overview page with a high number of generated cache entries.

### DIFF
--- a/Neos.Cache/Classes/Backend/PdoBackend.php
+++ b/Neos.Cache/Classes/Backend/PdoBackend.php
@@ -142,9 +142,12 @@ class PdoBackend extends IndependentAbstractBackend implements TaggableBackendIn
             $data = bin2hex($data);
         }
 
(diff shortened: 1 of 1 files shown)


gerhard-boden:
@albe Thanks your feedback!
So I went ahead and also wrapped all `DELETE` statements into a transaction, because I also think it makes sense to have consistent behavior when it comes to inserting and deleting entries.

I also used try/catch as you suggested and decided to still throw all exceptions after the rollback is done.

PR #1829, opened 2019-10-23
TASK: Switch to variadic function in object manager

kitsunet:
This uses variadic arguments instead of `func_get_args`
in the object manager.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
@@ -357,12 +357,13 @@ class CompileTimeObjectManager extends ObjectManager
      * defined in the object configuration of the specified object.
      *
      * @param string $objectName The name of the object to return an instance of
+     * @param mixed ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
```diff
@@ -357,12 +357,13 @@ public function getObjectConfigurations()
...
      *
      * @param string $objectName The name of the object to return an instance of
+     * @param mixed ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
      * @return object The object instance
      * @throws \Neos\Flow\ObjectManagement\Exception\CannotBuildObjectException
      * @throws \Neos\Flow\ObjectManagement\Exception\UnresolvedDependenciesException
      * @throws \Neos\Flow\ObjectManagement\Exception\UnknownObjectException
      */
-    public function get($objectName)
+    public function get($objectName, ...$constructorArguments)
```
kitsunet: Never used in there but needed because of extending and signature errors.

albe:
Doesn't that need to be changed in the `ObjectManagerInterface` too? And isn't that considered kind of breaking then?

Issue #1832, opened 2019-10-24
Create new Site-Package in the Web Setup-Tool throws an Error


eWert-Online:
### Description
In the Web Setup tool at Step 3 (Creating a new Site Package) the following error occurs:
`Notice: Undefined index: marktplatz/neos in /var/www/.../Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php line 399`

I dug a little into the files and I think i found the error.
The line `exec('composer require ' . $manifest['name'] . ' @dev');` (https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Package/PackageManager.php#L394) gets executed in the Web-Directory (/var/www/.../Web). So the new Package does not get installed and symlinked. Flow then doesn't know about the package at line 399, resulting in an error. 

It could be a configuration Issue with the Server, but @kitsunet said:
> maybe it's related to the server though because the cwd might be different depending on server settings, but the way you have it here seems not unusual so I woudl rather blame us

### Steps to Reproduce
1. Install a new neos-base-distribution
2. Change the Document Root to `"/your/htdocs/Neos/Web/"`
3. Create a new Site Package with the Setup-Tool

#### Expected behavior
The site package is created successfully

#### Actual behavior
The following error is shown:
`Notice: Undefined index: marktplatz/neos in /var/www/.../Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php line 399`

### Affected Versions
Neos: 5.0.2
Flow: 6.0.2

### Oldest Affected Version
Flow: 5.2.0

### Possible Related Issues:
#1778

eWert-Online:
This was fixed in the latest release and can be closed. 😊

Issue #1778, opened 2019-09-18
Calling flow kickstart:site from a directory different to the base directory 


RobinJespersen:
### Description
If I call ./flow kickstart:site from within the directory containing also the related composer.json  file, everything is fine.

But if I call ./test/flow kickstart:site (from one directory above) it fails with the following message:

```
Notice: Undefined index: package/name in /test/Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php line 391

  Type: Neos\Flow\Error\Exception
  Code: 1
  File: Packages/Framework/Neos.Flow/Classes/Error/ErrorHandler.php
  Line: 81
```
And creates an empty composer.json file in the directory above the ./flow file. 

### Steps to Reproduce

1. composer create-project neos/neos-base-distribution test
2. ./test/flow kickstart:site Package.Name test

#### Expected behavior

` ./test/flow kickstart:site Package.Name test`
should be equivalent to 
` cd ./test && ./flow kickstart:site Package.Name test`

### Affected Versions
site-kickstarter: 4.3.4
Neos: 4.3.4
Flow:5.3.4

davidspiola:
@RobinJespersen we can close this issue, right?

eWert-Online:
I would say this can be closed 😊
It was fixed by #1837

RobinJespersen:
thx :)

PR #1834, opened 2019-10-27
WIP: BUGFIX: Always run composer from Flow root path

eWert-Online:
**What I did**
This bugfix ensures, that the `composer require` command always gets executed in the Flow root path.
It also introduces the `composer/composer` package to the codebase to replace the `exec` command.

If this PR gets merged, it fixes the issues #1832 and #1778.

**How I did it**

- Add the `composer/composer` Package to the composer manifest
- Programatically call composer require with the --working-direcory flag set to `FLOW_PATH_ROOT `

**How to verify it**
1. run the `kickstart:site` command from any other direcory than the Flow root path 
(ex: `./test/flow kickstart:site Package.Name test`)
2. create a site-package with the web setup tool.

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/PdoBackend.php
+++ b/Neos.Cache/Classes/Backend/PdoBackend.php
@@ -168,18 +168,27 @@ class PdoBackend extends IndependentAbstractBackend implements TaggableBackendIn
             $data = bin2hex($data);
         }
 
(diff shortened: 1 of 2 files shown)


kitsunet:
@eWert-Online the problem is that we also need to require composer/composer in the development-collectoin then (basically with your change, go into the development-collection folder and run php ../../Build/BuildEssentials/ComposerManifestMerger.php) that should update the composer.json in that folder with the additional requirement.

Also apparently in the respective test the constant FLOW_PATH_ROOT is not declared, so I guess you need to either do that or mock it for the test.

PR #1836, opened 2019-10-31
Remove JSON_PRETTY_PRINT

kimegede:
This removes saving JSON with pretty print in our Doctrine `JsonArrayType`,
saving the extra bytes.

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/DataTypes/JsonArrayType.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/DataTypes/JsonArrayType.php
@@ -131,7 +131,7 @@ class JsonArrayType extends DoctrineJsonArrayType
         $this->initializeDependencies();
         $this->encodeObjectReferences($array);
 
-        return json_encode($array, JSON_PRETTY_PRINT | JSON_FORCE_OBJECT | JSON_UNESCAPED_UNICODE);
+        return json_encode($array, JSON_FORCE_OBJECT | JSON_UNESCAPED_UNICODE);
     }
 
     /**
--- a/Neos.Flow/Tests/Unit/Persistence/Doctrine/DataTypes/JsonArrayTypeTest.php
+++ b/Neos.Flow/Tests/Unit/Persistence/Doctrine/DataTypes/JsonArrayTypeTest.php
@@ -54,6 +54,6 @@ class JsonArrayTypeTest extends UnitTestCase
     public function passSimpleArrayAndConvertToJson()
     {
         $json = $this->jsonArrayTypeMock->convertToDatabaseValue(['simplestring',1,['nestedArray']], $this->abstractPlatformMock);
-        self::assertEquals("{\n    \"0\": \"simplestring\",\n    \"1\": 1,\n    \"2\": {\n        \"0\": \"nestedArray\"\n    }\n}", $json);
+        self::assertEquals('{"0":"simplestring","1":1,"2":{"0":"nestedArray"}}', $json);
     }
 }


albe:
That one failing test currently depends on the formatting, which is bad. The other failing test and psalm is unrelated.

kdambekalns:
In https://github.com/neos/flow/pull/23 there were some valud comments about potential side affects and this being dependent on the context (to make debugging easier).

albe:
> Eg. development context could be default JSON_PRETTY_PRINT and production context could be without.

I generally like that idea, but if this has any subtle side-effects (like in the test), then it would be bad to have different behaviour between dev and production. 🤔 Still, having that configurable is probably the way to go.

That way we could:
- have the configured default be the same as before for full b/c
- change the default to non-pretty with next major
- users could still choose to have different configs for dev/production if they want

kimegede:
Thanks for the feedback.
I would argue it's bad practice, if this should break anything, assume it's converted to a object or array and then it shouldn't have any effect if it's pretty print or not.

Please let me know, what I need to do, if any. Thanks.

albe:
> I would argue it's bad practice

I agree, but still, causing a breaking change needs to be heavily considered against the benefit. In this case I'd argue that the benefit is relatively small, in relation to potential pain caused by any bad practice in the wild. So the question becomes: should we introduce that break in 6.1 or only in 7.0 (while still providing a feature switch until then).

So, depending on what we agree on:
- should not be configurable, non pretty is inarguably the way to go:
  - is not breaking enough to postpone to 7.0: nothing to do
  - is breaking enough to postponed to 7.0: label as "next major", nothing to do
- should be configurable: discuss/make a suggestion on how that configuration could look, decide if default value should be `PRETTY` until 7.0 for b/c

My favorite would be the last option, though the most effort/work to do. But I'm fine with one of the other options too, if the others have another opinion on this.

/cc @bwaidelich as the senseful "b/c" evangelist ;)

kimegede:
@albe Sounds like some good approaches!
It should be bullet proof, to make it configurable with default `PRETTY_PRINT` in Flow 5 and/or 6 I would say.
And change the default to "non-pretty-print" in 7.x?

Happy to hear, that even such a minor change, isn't just rejected ;)

bwaidelich:
> Isn't it possible to specify column options like [...]

Mh, doesn't seem like it. I'd still prefer a new type then rather then making this configurable globally

albe:
```php
/**
 * @ORM\Column(type="flow_json_array", options={"pretty":false})
 */
protected $someField;
```
Oh my, that would have been a very clean solution. Too sad that the doctrine column types aren't a little bit more flexible.

> I'd still prefer a new type then rather then making this configurable globally

Hm, not sure about a new type just for that. How would that be named and how to explain its existence? What about possibly making `JSON_UNESCAPED_UNICODE` choosable at some point?

Sometime I'd really wish we had some clear way to use feature flags for gradually introducing breaking changes. Like some config file where we put the feature toggles clearly separated and they also disappear with the next major when the toggle is fixated. It would help with getting feedback on those changes in a way that framework users are not bound to opt into whole bundles of features via version pinning.

On another note, by looking around, I just saw this: https://www.doctrine-project.org/projects/doctrine-dbal/en/2.7/reference/types.html#json-array

We need to change our `flow_json_array` base type to `DoctrineJsonType`, to be compatible with DBAL > 2.10

kitsunet:
I don't really think we need a per column control over this. It would also create issues with converting when you change that. (Same goes for global setting btw.) ultimately I am very convinced that the stored representation should not matter at all. I will do a final check on Neos but we should strife for making this a blackbox to the outside. Looking into the future I would hope for DB level json support anyway where the "string" representation is opaque.

albe:
That's what the doctrine `json` type finally abstracts away. Switching we can also get rid of our custom postgre `jsonb` switch. But that's another topic in itself.

kimegede:
I have now fixed the one test, I found possible to fix.
Instead of testing against a string, I would normally have made it to an array and check for the keys.

kimegede:
@daniellienert I can agree with it's not a performance issue, but making debugging easier isn't a case for production IMO.

bwaidelich:
no offense, but I think this might be a case of bike shedding.
on a green field I'd go for the non-pretty version probably, but are the few bytes really worth considering this change?

albe:
So I guess with three vetos we'll conclude that's a `wontfix`. Feel free to bring the discussion up again, but I'll close this for now.

kimegede:
I totally disagree, but thanks for your time guys! :)

PR #1837, opened 2019-10-31
BUGFIX: Always run composer from Flow root path

### LINKED ISSUE #1778: Calling flow kickstart:site from a directory different to the base directory

### LINKED ISSUE #1832: Create new Site-Package in the Web Setup-Tool throws an Error

eWert-Online:
This bugfix ensures, that the `composer require` command always gets executed in the Flow root path (by calling composer require with the --working-direcory flag set to `FLOW_PATH_ROOT `.)

It also introduces the `composer/composer` package to the codebase to replace the `exec` command.

Fixes #1832
Fixes #1778

### DIFF
--- a/Neos.Flow/Classes/Package/PackageManager.php
+++ b/Neos.Flow/Classes/Package/PackageManager.php
@@ -20,6 +20,8 @@ use Neos\Flow\SignalSlot\Dispatcher;
 use Neos\Utility\Files;
 use Neos\Utility\OpcodeCacheHelper;
 use Neos\Flow\Package\Exception as PackageException;
+use Composer\Console\Application as ComposerApplication;
+use Symfony\Component\Console\Input\ArrayInput;
 
 /**
  * The default Flow Package Manager
@@ -383,7 +385,19 @@ class PackageManager implements PackageManagerInterface
         $manifest = ComposerUtility::writeComposerManifest($packagePath, $packageKey, $manifest);
 
         if ($runComposerRequireForTheCreatedPackage) {
-            exec('composer require ' . $manifest['name'] . ' @dev');
+            $composerRequireArguments = new ArrayInput([
+                'command' => 'require',
+                'packages' => [$manifest['name'] . ' @dev'],
+                '--working-dir' => FLOW_PATH_ROOT
+            ]);
+
+            $composerApplication = new ComposerApplication();
+            $composerApplication->setAutoExit(false);
+            $composerErrorCode = $composerApplication->run($composerRequireArguments);
+
+            if ($composerErrorCode !== 0) {
+                throw new Exception("The installation was not successful. Composer returned the error code: $composerErrorCode", 1572187932);
+            }
         }
 
         $refreshedPackageStatesConfiguration = $this->rescanPackages();
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -44,7 +44,9 @@
         "symfony/console": "~4.1.1",
 
         "neos/composer-plugin": "^2.0.0",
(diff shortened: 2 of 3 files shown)


kdambekalns:
From the original PR description:

> **How to verify it**
> 1. run the `kickstart:site` command from any other direcory than the Flow root path 
> (ex: `./test/flow kickstart:site Package.Name test`)
> 2. create a site-package with the web setup tool.
> 
> **Note**
> I did close and reopen the pull request, because i branched off from master and not from 5.2
> Sorry for the inconvenience

kdambekalns:
Will do…

kdambekalns:
From looking at it it seems perfect. The dependency is declared with `neos/flow` itself and the collection manifest has been updated. That should not interfere with anything split-related.

I just wonder why we ever had Flow depend on itself as a dev-dependency… 🤷‍♂️

kitsunet:
This is great, but also shows that now Flow will depend on composer, I guess we should move package creation out of the package manager ASAP and into the flow/kickstarter so you can get rid of that and the composer requirement when you want to. Alas not related to this change which definitely brings things in the right direction.

kdambekalns:
> shows that now Flow will depend on composer

Isn't that an opportunity? I am quite sure we have some dependency-order code that we could delegate, no?

kitsunet:
Hahaha, good idea, will have to look into that and see if we _could_ actually do that to order packages for settings and such.

PR #1839, opened 2019-11-02
BUGFIX: fix linking to root domains (empty path), by generating a "/" URL

skurfuerst:
Note that some URLs might now end with a slash that did not have a slash
before. If your code relies on that, you might need to adjust it!

Resolves: https://github.com/neos/neos-development-collection/issues/2759
Related: https://github.com/neos/flow-development-collection/pull/1804


Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
```diff
@@ -320,9 +320,13 @@ public function applyTo(UriInterface $baseUri, bool $forceAbsoluteUri): UriInter
             $uri = $uri->withPath(trim($baseUri->getPath(), '/') . '/' . ltrim($uri->getPath(), '/'));
         }
 
-        // Ensure the URL always starts with "/" if non-empty.
+        // Ensure the URL always starts with "/" if non-empty; and if path is empty, set it to "/" in order to
+        // enforce it *ALWAYS* starts with a /. The 2nd case is needed to e.g. link to the root node of a Neos
+        // site correctly.
         if (strlen($uri->getPath()) > 0 && $uri->getPath(){0} !== '/') {
```
albe: `If (strlen() === 0 || getPath(){0} != '/') {` maybe to avoid the elseif?

skurfuerst:
hey :-) 

@albe I can feel you - I will add a functional test which is more expressive than the unit tests. These unit tests cannot prove whether everything works correctly or not really (but they are quite exhaustive on the other hand).

All the best,
Sebastian

Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
skurfuerst: changed.

skurfuerst:
> Left a comment. Not sure how I feel about all the test expectation changes :/

I just added a new functional test which actually reproduces the issue (linking to root); in both the absolute and non-absolute case. IMHO this is what "counts" to me; as the DTO was well-tested, but led to the bug nevertheless.

> At least for the absolute cases (where host differs), the / need not be appended I think.

True, although it does not hurt; does it? I would keep it this way for consistency reasons.

All the best,
Sebastian

albe:
Hey Sebastian, thanks a lot for this! :) I agree a functional test better covers the use case at hand.

> True, although it does not hurt; does it?

That's what I'm not 100% sure about - I mean it was covered by unit tests, that the URLs generated do not end with a `/` - and I currently have a project (not Flow/PHP, but anyway) that is riddled with bugs due to expectations of directory and URI separator placements and it is uuuuggghhh :(
What I'm trying to say - this might be "breaky" for some usages, but then again it will already be for the "empty path" case. So maybe it's enough to just clearly mention this breakyness...

skurfuerst:
@albe I *think* that in Neos 5.0 we had the same behavior; only that the "/" was appended at a different place - see the section "Usage in Flow" in https://github.com/neos/flow-development-collection/pull/1804.

All the best,
Sebastian :)

kdambekalns:
The failing Travis CI build is unrelated…

kitsunet:
Ok, great that this is fixed! But, this is highly related to the revert from yesterday, I think we MUST investigate this part of the link building and think it through in a more holistic manner than just fixing tiny parts because it seems it has several (edge case) issues that might even run counter to each other.

skurfuerst:
@kitsunet I am open/around to discuss this, and write functional tests for it :)

kitsunet:
yep, yep, just wanted to create awareness :) Not meaning you in particular, I alos need to read through and get all the things into my head myself before giving some consistent answer :)

PR #1840, opened 2019-11-04
BUGFIX: Set-Cookie headers mustn't be imploded

ComiR:
PSR-7 `getHeaderLine()` defaults to concatenating header values with a comma. This doesn't work for the `Set-Cookie` header, so it must be treated uniquely.

Now having multiple headers of the same name, `header()` must be used without replacement of duplicate headers (which is default).

See:
https://www.php-fig.org/psr/psr-7/#headers-with-multiple-values
https://tools.ietf.org/html/rfc7230#section-3.2.2
https://github.com/guzzle/psr7/issues/298

/cc @kitsunet @albe


albe:
Travis failure is unrelated

PR #1841, opened 2019-11-05
Revert "BUGFIX: Respect Neos.Flow.http.baseUri path in UriBuilder"

kitsunet:
Reverts neos/flow-development-collection#1682


kitsunet:
See comments in original PR on why this should be reverted..

PR #1842, opened 2019-11-06
TASK: Use yield instead of AppendIterator in getObjects

albe:
The method docblock stated that it returns a `Generator`, while in reality for a `PackageStorage` with patterns, it would fall back to an `AppendIterator`.
This change instead makes use of `yield` to return the `StorageObjects` in a memory-efficient manner.

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/Collection.php
+++ b/Neos.Flow/Classes/ResourceManagement/Collection.php
@@ -157,22 +157,15 @@ class Collection implements CollectionInterface
      */
     public function getObjects(callable $callback = null)
     {
-        $objects = [];
         if ($this->storage instanceof PackageStorage && $this->pathPatterns !== []) {
-            $objects = new \AppendIterator();
             foreach ($this->pathPatterns as $pathPattern) {
-                $objects->append($this->storage->getObjectsByPathPattern($pathPattern, $callback));
+                foreach ($this->storage->getObjectsByPathPattern($pathPattern, $callback) as $storageObject) {
+                    yield $storageObject;
+                }
             }
         } else {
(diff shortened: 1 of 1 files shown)


albe:
Not 100% sure about that removed `TODO`, but the example code exists since ages and looks as if that's not going to be implemented like that in the foreseeable future.

Neos.Flow/Classes/ResourceManagement/Collection.php
```diff
@@ -157,22 +157,13 @@ public function publish()
...
             foreach ($this->pathPatterns as $pathPattern) {
-                $objects->append($this->storage->getObjectsByPathPattern($pathPattern, $callback));
+                yield $this->storage->getObjectsByPathPattern($pathPattern, $callback);
             }
         } else {
-            $objects = $this->storage->getObjectsByCollection($this, $callback);
+            return $this->storage->getObjectsByCollection($this, $callback);
         }
-
-        // TODO: Implement filter manipulation here:
```
kitsunet: Yeah, probably fine to remove it :D Maybe _sometime_

albe:
Not good really, because `getObjectsByPathPattern()` already returns a generator - so that's why the AppendIterator was used, to hide having to iterate that result too

kitsunet:
> Not good really, because `getObjectsByPathPattern()` already returns a generator - so that's why the AppendIterator was used, to hide having to iterate that result too

But we could just do that? Like iterate the inner generators? Seems fine to me

albe:
Good now, but probably needs retargeting to 4.3, or at least 6.0?

kdambekalns:
6.0 according to Travis…

kdambekalns:
I'll move that to 6.0

PR #1844, opened 2019-11-08
BUGFIX: fix resource publishing

skurfuerst:
This fixes a regression introduced in #1842:

In this change, a "yield" and a "return" call were introduced in the
same method; so that the early return was not properly executed
because PHP somehow thought this is a fully-suspendable function,
although only one part of the "if"-statement was suspendable.

Related: #1842

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/Collection.php
+++ b/Neos.Flow/Classes/ResourceManagement/Collection.php
@@ -164,8 +164,11 @@ class Collection implements CollectionInterface
                 }
             }
         } else {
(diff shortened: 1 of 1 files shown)


kitsunet:
Known psalm failures, so seems fine!

kdambekalns:
TIL we lack a test for that part of the code, obviously 🙄

kitsunet:
Well it didn't produce any errors, we would have to test if something is iterated, but yeah, probably possible to do that in a functional test.

PR #1845, opened 2019-11-08
BUGFIX: fix linking to root domains (empty path), by generating a "/" URL

skurfuerst:
Also includes a functional test case, which is way more expressive than the Unit Test Case.

Resolves: https://github.com/neos/neos-development-collection/issues/2759
Related: https://github.com/neos/flow-development-collection/pull/1804


kitsunet:
Also here only the known psalm failures that are false positives.

PR #1846, opened 2019-11-09
FEATURE: Enrich default Web/.htaccess

AureliusHogan:
Added some statements for
- Compression
- Browser caching of resource files
- MIME types
- UTF-8 encoding
- Cross Origin
- disable directory listings
- ETag removal

Fixes: #1869

### DIFF
--- a/Neos.Flow/Resources/Private/Installer/Distribution/Defaults/Web/.htaccess
+++ b/Neos.Flow/Resources/Private/Installer/Distribution/Defaults/Web/.htaccess
@@ -5,6 +5,218 @@
 # You can specify a default context by activating this option:
 # SetEnv FLOW_CONTEXT Production
 
+### Begin: Compression ###
+
+<IfModule mod_deflate.c>
+	# Force compression for mangled `Accept-Encoding` request headers
+	<IfModule mod_setenvif.c>
+		<IfModule mod_headers.c>
+			SetEnvIfNoCase ^(Accept-EncodXng|X-cept-Encoding|X{15}|~{15}|-{15})$ ^((gzip|deflate)\s*,?\s*)+|[X~-]{4,13}$ HAVE_Accept-Encoding
+			RequestHeader append Accept-Encoding "gzip,deflate" env=HAVE_Accept-Encoding
+		</IfModule>
+	</IfModule>
+
+	# Compress all output labeled with one of the following media types.
+	#
+	# (!) For Apache versions below version 2.3.7 you don't need to
+	# enable `mod_filter` and can remove the `<IfModule mod_filter.c>`
+	# and `</IfModule>` lines as `AddOutputFilterByType` is still in
+	# the core directives.
+	#
+	# https://httpd.apache.org/docs/current/mod/mod_filter.html#addoutputfilterbytype
+
+	<IfModule mod_filter.c>
+		AddOutputFilterByType DEFLATE application/atom+xml \
+			application/javascript \
+			application/json \
+			application/ld+json \
+			application/manifest+json \
+			application/rdf+xml \
+			application/rss+xml \
+			application/schema+json \
+			application/vnd.geo+json \
+			application/vnd.ms-fontobject \
+			application/x-font-ttf \
+			application/x-javascript \
+			application/x-web-app-manifest+json \
+			application/xhtml+xml \
+			application/xml \
+			font/eot \
+			font/opentype \
+			image/bmp \
+			image/svg+xml \
+			image/vnd.microsoft.icon \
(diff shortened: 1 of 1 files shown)


albe:
Hey, welcome and thanks a lot for your first contribution!

I'm generally greatly in favor of improving the default behaviour of frameworks and make more things work well out of the box. There's definitely room regarding HTTP protocol defaults we set.

Here's a few things for thought though:
- this places a couple decisions that might not be optimal in a wider than assumed range of use-cases (e.g. default caching lifetimes, CORS for all image types and disabled ETag)
- some of those settings effectively would overwrite headers set in the application (e.g. all `Header (un)set` directives) - I'd prefer `setifempty` to at least give the application the possibility to specify those things itself and not be overruled
- all those things would currently only work on an apache webserver environment, so effectively for some things behaviour might be noticably different depending on the webserver. That's a documentation nightmare waiting. Maybe someone can (once we decided on what we want to set) provide an equal set of configs for nginx?

For things that are non-obvious (default cache lifetimes, CORS, ETag removal), I'd at least like to have some arguments collected here on why the chosen values are considered "useful" generically. For example regarding cache lifetimes I have been bitten a few times by overly optimistic chosen values - and it's really non-trivial to get things out of proxy or browser caches once they're inside (yeah, name change/cache busting query params).

AureliusHogan:
Hi albe,

this started with the discussion in: https://discuss.neos.io/t/leverage-browser-caching-with-htaccess/4656/11 

What if we set up different .htaccess template files for different webservers? I just wanted to share the settings that changed response rather dramatically. I'm sure there are a lot more settings that could optimize it even more. And also there is currently no blocking of sensible areas. And that is because I don't know which would be them.

Perhaps it's better to have different tutorials that include best practices.

albe:
Definitely a perfect use-case for a tutorial style (https://docs.neos.io/cms/tutorials) documentation page on docs.neos.io that explains the trade-offs made based on those settings and show some different use-cases with optimized configs.

Also, a lot of the inline comments on 
https://github.com/h5bp/html5-boilerplate/blob/master/dist/.htaccess
already provide answers to reasoning on why those settings are chosen. So really, I don't know why they are stripped in the typo3 .htaccess
For example, there's three very helpful links to the reasoning on the `remove ETags` setting (https://github.com/h5bp/html5-boilerplate/blob/master/dist/.htaccess#L1035-L1037) - that at least I couldn't myself reason for based on my previous knowledge. So let's at minimum keep those and more reflect a picked subset of the html5 boilerplate best practice .htaccess

Things that I see we can safely integrate in our base .htaccess so far:
- Force compression for mangled `Accept-Encoding` request headers
- output compression based on mediatype
- MIME types
- 404 error prevention for non-existing redirected folders (?)
- disable directory listings
- Force IE to render pages in the highest available mode

What I see reasonable, but needs some more thought on impact:
- default UTF-8 encoding: yes, please UTF-8 default, but this could be breaking in some circumstances
- CORS headers: let's use `Header setifempty` as mentioned above
- ETag removal: the reasoning in the links on html5 boilerplate makes sense, at least for multi-server setups, but what if the application itself deliberately sets *useful* ETag headers - this would currently effectively bypass that
- Browser caching of resource files: highest performance impact, but as also mentioned on html5 bp those settings make most sense together with a cache-busting strategy - https://github.com/h5bp/html5-boilerplate/blob/master/dist/.htaccess#L1052-L1054, so I'd argue that we should set the defaults way lower in order to prevent issues with page updates not reaching the users

I'm currently reading up on the "Filename-based cache busting" strategy in the html5 bp template, maybe this could be another useful addition, but we'd need to support generating cache-busted resource URIs in core.

I'd be interested in opinions from @kdambekalns @robertlemke and @daniellienert, they have more experience with running flow/neos on a scale.

davidspiola:
Hi Folk, I'll move this PR in the backlog for now. I you guys think that at least a few settings gonna make it into the release, we can merge that later.

albe:
Addition:
We should maybe base things on 
https://github.com/h5bp/server-configs-apache
and
https://github.com/h5bp/server-configs-nginx
directly. Since nginx config is a manual step, that is probably something we just need to document. Under any circumstance, let's properly reference the source for those configs.

Anyone willing to take this up?

mhsdesign (changes requested):
Hmm this PR is now open for some time - i think this configuration is better kept as an optional and opinionated tutorial on https://docs.neos.io/cms/tutorials instead of actually wandering into the core.

kdambekalns:
I even think this can be found in various tutorials already and doesn't need to be repeated…

PR #1847, opened 2019-11-11
TASK: Add PHP 7.4 to travis test matrix

albe:
PHP 7.4 will be released soon and we should make sure at least Flow 6.1 supports it

Resolves: #1866

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,11 +33,11 @@ matrix:
       env: DB=mysql BEHAT=true
       addons:
         mariadb: '10.2'
-    - php: 7.4
+    - php: 7.4snapshot
       env: DB=mysql
       addons:
         mariadb: '10.2'
-    - php: 7.4
+    - php: 7.4snapshot
       env: DB=mysql BEHAT=true
       addons:
         mariadb: '10.2'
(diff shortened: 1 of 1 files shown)


daniellienert:
Yeah! Great initiative ;) 
But `The command "phpenv global 7.4" failed and exited with 1 during .`
ENV variable should be `7.4snapshot` AFIK.

kdambekalns:
If we did not bork, PHP 7.4 should work with Flow 5.3 and Neos 4.3 and up. So let's get this into 5.3 once we verified it works!

albe:
Well, 7.4 is official now, so we can probably remove the `snapshot` again soon. Also yes, let's check that 5.3 is PHP 7.4 compatible once we figured everything out

kdambekalns:
FYI: A Fluid release is currently planned for December 12th, see https://github.com/TYPO3/Fluid/issues/501

bwaidelich:
I added an issue for this in order to simplify release management and to collect higher-level details for the release notes: #1866

kdambekalns:
FYI: A Fluid release working with PHP 7.4 has been made.

kdambekalns:
I suggest to replace this with https://github.com/neos/flow-development-collection/pull/1872

PR #1849, opened 2019-11-18
FEATURE: Allow StringLengthValidator to ignore HTML tags

bwaidelich:
Adds an option `ignoreHtml` to the `StringLengthValidator`

Usage:

    $validator = new StringLengthValidator(['maximum' => 14, 'ignoreHtml' => true]);
    $validator->validate('Some <b>bold</b> text')->hasErrors(); // false

Resolves: neos/neos-ui#2579

### DIFF
--- a/Neos.Flow/Classes/Validation/Validator/StringLengthValidator.php
+++ b/Neos.Flow/Classes/Validation/Validator/StringLengthValidator.php
@@ -26,7 +26,8 @@ class StringLengthValidator extends AbstractValidator
      */
     protected $supportedOptions = [
         'minimum' => [0, 'Minimum length for a valid string', 'integer'],
-        'maximum' => [PHP_INT_MAX, 'Maximum length for a valid string', 'integer']
+        'maximum' => [PHP_INT_MAX, 'Maximum length for a valid string', 'integer'],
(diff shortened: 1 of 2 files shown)


markusguenther:
Tried it with the neos-demo package and it worked well 👍

albe:
psalm failures unrelated.

bwaidelich:
> Edge-casey [...]

Not really, see https://github.com/neos/neos-ui/issues/2579 (and without the Flow change, we can't extend the UI validator because their options need to be in sync)

albe:
Well, it's edge-casey in the sense of what you consider a "string" and what other specific formatting markup you might want to be able to exclude too. Technically, there should rather be a `HtmlContentLengthValidator` (and a `MarkdownContentLengthValidator`, and a ...) - you get the idea ;)

bwaidelich:
Ah, I see what you mean. You're right but I didn't want to open that pandoras box (actually I was a bit put off that we can't introduce UI validators without having to create a corresponding Flow validator even though they live in a totally different context, but I see the reasoning behind that decision)

davidspiola:
Hi Folks, @albe just told me that this PR is breaking the tests and that it's somehow related to the neos/forms projekt. Could one of you pls check, thx

albe:
Done 😄 ☝️

PR #1850, opened 2019-11-20
BUGFIX: Ensure ActionReponse properties end up in any HttpResponse

kitsunet:
To fix a bug in Neos this ensures that the ActionReponse properties
are applied to a HTTP response rendered
by a View (a new feature in 6.0).

Technically this should not be necessary as any HTTP response rendered
by a View should be understood as being "complete", sadly that is not
always the case with plugins and widgets so we need to take this detour
for now to apply the properties. In the future this should happen in
the respective Views directly and no longer be the responsiblity of
the MVC or HTTP layer.

### DIFF
--- a/Neos.Flow/Classes/Mvc/ActionResponse.php
+++ b/Neos.Flow/Classes/Mvc/ActionResponse.php
@@ -38,9 +38,12 @@ final class ActionResponse
     /**
      * The HTTP status code
      *
-     * @var integer
+     * Note the getter has a default value,
+     * but internally this can be null to signify a status code was never set explicitly.
+     *
+     * @var integer|null
      */
-    protected $statusCode = 200;
+    protected $statusCode;
 
     /**
      * @var string
@@ -185,7 +188,7 @@ final class ActionResponse
      */
     public function getStatusCode(): int
     {
-        return $this->statusCode;
+        return $this->statusCode ?? 200;
     }
 
     /**
@@ -205,6 +208,7 @@ final class ActionResponse
         if ($this->hasContent()) {
             $actionResponse->setContent($this->content);
         }
+
         if ($this->contentType !== null) {
             $actionResponse->setContentType($this->contentType);
         }
@@ -236,8 +240,9 @@ final class ActionResponse
     public function mergeIntoComponentContext(ComponentContext $componentContext): ComponentContext
     {
         $httpResponse = $componentContext->getHttpResponse();
-        $httpResponse = $httpResponse
-            ->withStatus($this->statusCode);
+        if ($this->statusCode !== null) {
+            $httpResponse = $httpResponse->withStatus($this->statusCode);
+        }
 
         if ($this->hasContent()) {
             $httpResponse = $httpResponse->withBody($this->content);
@@ -280,9 +285,8 @@ final class ActionResponse
      */
     public function applyToHttpResponse(ResponseInterface $httpResponse): ResponseInterface
     {
-        if ($this->statusCode !== 200) {
-            $httpResponse = $httpResponse
-                ->withStatus($this->statusCode);
+        if ($this->statusCode !== null) {
+            $httpResponse = $httpResponse->withStatus($this->statusCode);
         }
 
         if ($this->hasContent()) {
--- a/Neos.Flow/Classes/Mvc/ActionResponse.php
+++ b/Neos.Flow/Classes/Mvc/ActionResponse.php
@@ -289,7 +289,7 @@ final class ActionResponse
             $httpResponse = $httpResponse->withBody($this->content);
         }
 
-        if ($this->contentType) {
+        if ($this->contentType !== null) {
             $httpResponse = $httpResponse->withHeader('Content-Type', $this->contentType);
         }
 
--- a/Neos.Flow/Classes/Mvc/ActionResponse.php
+++ b/Neos.Flow/Classes/Mvc/ActionResponse.php
@@ -293,7 +293,7 @@ final class ActionResponse
             $httpResponse = $httpResponse->withHeader('Content-Type', $this->contentType);
         }
 
-        if ($this->redirectUri) {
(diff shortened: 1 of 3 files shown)


kitsunet:
WARNING: We need https://github.com/neos/neos-development-collection/pull/2777 in Neos first, otherwise neos pages with plugins will be broken!

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -264,4 +265,52 @@ public function mergeIntoComponentContext(ComponentContext $componentContext): C
...
+     *
+     * WARNING: Should this ActionResponse contain body content it would replace any content in the given HttpReponse.
+     *
+     * @param ResponseInterface $httpResponse
+     * @return ResponseInterface
+     * @internal
+     */
+    public function applyToHttpResponse(ResponseInterface $httpResponse): ResponseInterface
+    {
+        if ($this->statusCode !== 200) {
```
bwaidelich: IMO we should change the default to `null` here, check for that and only merge the status if it has been set explicitly

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -264,4 +265,52 @@ public function mergeIntoComponentContext(ComponentContext $componentContext): C
...
+            $httpResponse = $httpResponse
+                ->withStatus($this->statusCode);
+        }
+
+        if ($this->hasContent()) {
+            $httpResponse = $httpResponse->withBody($this->content);
+        }
+
+        if ($this->contentType) {
```
bwaidelich: CGL nitpick
```suggestion
        if ($this->contentType !== null) {
```

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -264,4 +265,52 @@ public function mergeIntoComponentContext(ComponentContext $componentContext): C
...
+            $httpResponse = $httpResponse->withHeader('Content-Type', $this->contentType);
+        }
+
+        if ($this->redirectUri) {
```
bwaidelich: CGL nitpick
```suggestion
        if ($this->redirectUri !== null) {
```

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -264,4 +265,52 @@ public function mergeIntoComponentContext(ComponentContext $componentContext): C
...
+    }
+
+    /**
+     * Does this action response have content?
+     *
+     * @return bool
+     */
+    private function hasContent(): bool
+    {
+        return $this->content->getSize() > 0;
```
bwaidelich: can't `$this->content` be null?

bwaidelich: I realize it's instantiated in the constructor, but maybe we should still check for null in `setContent()`

Neos.Flow/Classes/Mvc/ActionResponse.php
kitsunet: I am not sure if that wasn't a breaking change now that we have the default at 200, but I agree to the sentiment. I definitely do not want to do that in this change...

Neos.Flow/Classes/Mvc/ActionResponse.php
kitsunet: Copied from above, but sure, we can make it nicer here than above ;)

Neos.Flow/Classes/Mvc/ActionResponse.php
kitsunet: same as above...

Neos.Flow/Classes/Mvc/ActionResponse.php
kitsunet: Puhh, I would rather want to make sure we definitely never have `null` values in an instance.

Neos.Flow/Classes/Mvc/ActionResponse.php
bwaidelich: Fair enough and I don't want to add even more work. But FYI since the class is final it would be 100% b/c if we'd change
```php
    public function getStatusCode(): int
    {
        return $this->statusCode ?? 200;
    }
```

Neos.Flow/Classes/Mvc/ActionResponse.php
bwaidelich: Sorry for the confusion, that's what I meant. And I just realized that with the existing check of 
```php
        if (!$content instanceof StreamInterface) {
            $content = stream_for($content);
        }
```
it should not be possible to null the field via `$this->response->setContent(null)`, so all good :)

Neos.Flow/Classes/Mvc/ActionResponse.php
kitsunet: Right, that makes sense I guess, seems a bit hidden tricky but I take it ;)

kitsunet: Ah BUT, that actually has another problem... Similar to the content problem. The plugins and widgets WILL probably always set a status code to the response given that it's a full ActionController run, now if the plugin returned 200 it could technically overwrite a redirect (or any other status code in the HTTP response) which would then be unexpected...

bwaidelich: > The plugins and widgets WILL probably always set a status code

Is that so? They should only set that if it's explicitly set (i.e. by a redirect or via `$this->response->setStatusCode()`). If that's not the case, it's probably too big of a change..

kitsunet: You are right, we apparently never set it unless we must :D I expected worse. I guess I will do it and we see where that leaves us.

kitsunet:
Traivs error unrelated and already existing...

PR #1853, opened 2019-11-22
BUGFIX: Check existence and disable new rendering limitations

kitsunet:
The symfony/console package got new limitations for rendering
the progress bar in quick succession with minor version 4.4.

We need to disable these limitations by default for now as they
break our own tests.

We should think about following the defaults in the future and
adjust our tests accordingly.


kitsunet:
This fixes the current test errors regarding console progress bar...

PR #1856, opened 2019-11-28
TASK: Remove version number from ext-json dependency

kdambekalns:
This has been at ^1.2 for ages, basically since Flow exists. Now, that
version is really, really old (from 2006) and the package has long been
moved from PECL into the PHP source (see https://pecl.php.net/package/json)

With PHP 7.4.0RC6 (at least), the version reported for ext-json is the
same as the PHP version, this blocks installation via composer.

Those two facts seems to warrant a simple * for that extension dependency,
like for the other PHP extensions we depend on.

### DIFF
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -10,7 +10,7 @@
 
         "ext-zlib": "*",
         "ext-SPL": "*",
-        "ext-json": "^1.2",
+        "ext-json": "*",
         "ext-reflection": "*",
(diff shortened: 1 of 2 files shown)


kdambekalns:
Fixes `neos/flow x.y.z requires ext-json ^1.2 -> the requested PHP extension json has the wrong version (7.4.0RC6) installed.` during `composer update` (or `install`.)

kdambekalns:
See #1847 as well…

PR #1858, opened 2019-11-29
FEATURE: Add array ksort eel helper

jonnitto:
Resolves: #1864


jonnitto:
As I can say, the failing test has nothing to do with this PR, but has someone a clue, why it fails?

bwaidelich:
I added an issue for this in order to simplify release management and to collect higher-level details for the release notes: #1864

PR #1859, opened 2019-11-29
FEATURE: Add unique array eel helper

jonnitto:


Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -238,6 +238,22 @@ public function shuffle(array $array, $preserveKeys = true): array
...
+     * Removes duplicate values from an array
+     *
+     * @param array $array The input array
+     * @param bool $filter Option if the array should be filtered, defaults to `false`
+     *
+     * @return array The filtered array.
+     */
+    public function unique(array $array, bool $filter = false): array
+    {
+        if ($filter) {
```
Sebobo: As we already have a filter method in this helper shouldn't we recommend that one instead of doing it here too and a bit differently?

jonnitto: Do you mean to remove the option and used it like `Array.filter(Array.unique(ARRAY), item => item ? true : false)` ?

jonnitto: Or do you mean `Array.unique(ARRAY, item => item ? true : false)`

Sebobo: The first one `Array.filter(Array.unique(...`
All our array methods are quite basic and mostly only wrap the php methods.
I know why you like to have the filter in there but I think it's better to just add the unique method and leave all the filter things away.

PR #1860, opened 2019-11-29
FEATURE: Add optional limit parameter to String.pregReplace

jonnitto:
Resolves: #1863

### DIFF
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -184,7 +184,7 @@ class StringHelper implements ProtectedContextAwareInterface
         if ($index === false) {
             return -1;
         }
-        return (integer)$index;;
+        return (integer)$index;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jonnitto:
Ups, it seems my IDE add some spaces to other functions, I'll fix this

bwaidelich:
Nice, but the corresponding ui validator should be kept in sync, right?

jonnitto:
I don‘t know exactly what you mean. Where I‘ll find the code I should adapt?

bwaidelich:
> I don‘t know exactly what you mean. Where I‘ll find the code I should adapt?

Hah, sorry for the confusion. I mixed it up with validators and thought we'd need to adjust https://github.com/neos/neos-ui/blob/master/packages/neos-ui-validators/src/RegularExpression/index.tsx. Nevermind :)

bwaidelich:
I added an issue for this in order to simplify release management and to collect higher-level details for the release notes: #1863

albe:
Remaining psalm errors unrelated.

markusguenther:
@albe Could we fix that somehow? (Psalm stuff)

albe:
Yes, but I'll take care in another PR, so let's not block this on that error.

PR #1861, opened 2019-11-29
FEATURE: Add String.urlize Eel helper

jonnitto:
This Eel helper is great to create id names from headlines (for example)

### DIFF
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -16,6 +16,7 @@ use Neos\Eel\ProtectedContextAwareInterface;
 use Neos\Flow\Annotations as Flow;
 use Neos\Utility\Unicode\Functions as UnicodeFunctions;
 use Neos\Utility\Unicode\TextIterator;
+use Behat\Transliterator\Transliterator;
 
 /**
  * String helpers for Eel contexts
@@ -24,6 +25,22 @@ use Neos\Utility\Unicode\TextIterator;
  */
 class StringHelper implements ProtectedContextAwareInterface
 {
+    /**
+     * Generates a slug of the given string
+     *
+     * Examples::
+     *
+     *     String.urlize('Hello World') == 'hello-world'
+     *     String.urlize('Ä_ÖÜ äöü') == 'ae-oeue-aeoeue'
+     *
+     * @param string $string The string
+     * @return string The converted string
+     */
+    public function urlize($string)
+    {
+        return Transliterator::urlize($string);
+    }
+
     /**
      * Return the characters in a string from start up to the given length
      *
--- a/Neos.Eel/Tests/Unit/Helper/StringHelperTest.php
+++ b/Neos.Eel/Tests/Unit/Helper/StringHelperTest.php
@@ -20,6 +20,25 @@ use Neos\Flow\Tests\UnitTestCase;
  */
 class StringHelperTest extends UnitTestCase
 {
+    public function urlizeExamples()
+    {
+        return [
+            'mixed case with spaces' = ['Hello World' ,'hello-world'],
+            'umlauts, underscores and spaces' = ['Ä_ÖÜ äöü', 'ae-oeue-aeoeue']
+        ];
+    }
+
+    /**
+     * @test
+     * @dataProvider urlizeExamples
+     */
+    public function urlizeWorks($string, $expected)
+    {
+        $helper = new StringHelper();
+        $result = $helper->urlize($string);
+        self::assertSame($expected, $result);
+    }
+
     public function substrExamples()
     {
         return [
--- a/Neos.Eel/composer.json
+++ b/Neos.Eel/composer.json
@@ -8,7 +8,8 @@
         "neos/flow": "*",
         "neos/cache": "*",
         "neos/utility-unicode": "*",
-        "neos/utility-objecthandling": "*"
+        "neos/utility-objecthandling": "*",
+        "behat/transliterator": "^1.2"
     },
     "autoload": {
         "psr-4": {


Sebobo:
Love the feature!

bwaidelich:
I added an issue for this in order to simplify release management and to collect higher-level details for the release notes: #1865

albe:
Generally yes, but some questions:
- why was Behat/Transliterator chosen instead of e.g. `php-intl` + `Transliterator::transliterate("ASCII//TRANSLIT//IGNORE", $string)`? (not saying it's better, just want to have a rationale for the decision)
- what about other non-URL characters or characters that have special meaning, like e.g. "%", "?" or "`" (see https://www.ietf.org/rfc/rfc3986.txt 2.2 + 2.3) - shouldn't those be stripped or at least be urlencoded?

bwaidelich:
Good remarks by @albe (as usual l). We should make sure to get this right from the start and not to introduce unwanted dependencies as ut will be painful to change this lateron

jonnitto:
If this comes to a merge, please squash it 😉

jonnitto:
It would also need `'::Any-Latin;'` before `'::Latin-ASCII;'`, but with this, € and $ get not changed…

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -24,6 +24,43 @@
...
+            'ü' => 'ue',
+            'Ö' => 'Oe',
+            'ö' => 'oe',
+            // Norwegian characters
+            'Å' => 'Aa',
+            'å' => 'aa'
+        ];
+        $string = strtr($string, $specialChars);
+        $string = str_replace('_', ' ', $string);
+        $string = transliterator_transliterate('Any-Latin; Latin-ASCII; [\u0100-\u7fff] remove; [:Punctuation:] Remove; Lower()', $string);
```
bwaidelich: Oh, nice!
Just wondering: is there a specific reason to use procedural style here?
As a little performance optimization I would suggest to go for sth like:

```php
    /**
     * @var \Transliterator
     */
    private static $transliterator;

    private function getTransliterator(): \Transliterator
    {
        if (self::$transliterator === null) {
            self::$transliterator = \Transliterator::createFromRules('...');
        }
        return self::$transliterator;
    }

    // ...

    public function urlize(string $string): string
    {
        $string = $this->getTransliterator()->transliterate($string);
        // ...
```

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -24,6 +24,43 @@
...
+            'Ü' => 'Ue',
+            'ü' => 'ue',
+            'Ö' => 'Oe',
+            'ö' => 'oe',
+            // Norwegian characters
+            'Å' => 'Aa',
+            'å' => 'aa'
+        ];
+        $string = strtr($string, $specialChars);
+        $string = str_replace('_', ' ', $string);
```
bwaidelich: Why that?

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -24,6 +24,43 @@
...
+        $string = transliterator_transliterate('Any-Latin; Latin-ASCII; [\u0100-\u7fff] remove; [:Punctuation:] Remove; Lower()', $string);
+        $string = preg_replace('/[-\s]+/', '-', $string);
```
bwaidelich: Do we really need another regex replacement here?
Can't we just extend the `$specialChars` array above?

jonnitto:
One point about `behat/transliterator`: I know, it would be a new dependency for the Eel package, but not for Neos: It uses the same library for the generation for the `uriPathSegment`.

bwaidelich:
> I know, it would be a new dependency for the Eel package, but not for Neos

You're right, Neos has a dependency declared to that package. But since Eel can be used with Flow alone, it would still add a new dependency.
Do you think the Behat transliterator is more capable than the built-in one?

jonnitto:
At least it would do the same, and all the special characters would be handled correctly. I mean, I would love to have a dependency-less solution, but when I look at the code from the transliterator package, I don't know If we can do it with these few lines. Currently, It works pretty well (I would still have to figure out your optimization), but If we decide not to include the behat package, I would prefer to wait a bit and release it with Neos 5.2. Otherwise, I would have to revert it and then there would be no risk to merge it for Neos 5.1

bwaidelich:
@jonnitto good points, I agree to what you wrote!

jonnitto:
> At least it would do the same, and all the special characters would be handled correctly. I mean, I would love to have a dependency-less solution, but when I look at the code from the transliterator package, I don't know If we can do it with these few lines. Currently, It works pretty well (I would still have to figure out your optimization), but If we decide not to include the behat package, I would prefer to wait a bit and release it with Neos 5.2. Otherwise, I would have to revert it and then there would be no risk to merge it for Neos 5.1

So @albe, @bwaidelich, and @daniellienert: Which option we want to choose?

* 👍 for adding behat
* 👎 for writing our own function and wait for 5.2

Note: If we have one 👎, we will wait for 5.2

bwaidelich:
I think this is a great addition and I don't want to step on your toes, but my tendency is towards postponing because:
* Once decided we'll have to "live" with the decision. For most other Eel helpers the behavior is pretty obvious, but this one is rather special
* As long as this feature is not in the core it's relatively easy to ship in in your own site package, thus the pain of postponing this is not _too_ great.

But I won't block this if you guys decide otherwise

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -24,6 +24,43 @@
...
+     *     String.urlize('Hello World') == 'hello-world'
+     *     String.urlize('Ä_ÖÜßäöüÅåAÆæØEuro:€éñåøå!_И люяюлб PHP! есть. ﬁ  ') == 'ae-oeuessaeoeueaaaaaaeaeoeuroeenaaoaa-i-luaulb-php-est-fi'
+     *
+     * @param string $string The string to be transformed.
+     * @return string The converted string
+     */
+
+    public function urlize(string $string): string
+    {
+        $specialChars = [
```
kitsunet: Why do we do this conversion here and then transliterate, that seems wrong? Shouldn't the Transliteration take care of these?

bwaidelich: True

jonnitto: Nope. The transliteration (with this config) doesn't take care of these chars. But as I said: If we decide not to implement behat, I will further dive deeper into this and optimize the code 😉

jonnitto:
I think it is best to skip this feature for this release and look at what we can do for 5.2

albe:
Yes, let's skip and more closer inspect if we can replace the dependency with Transliterator in a way that doesn't require writing (and maintaining) a couple dozen or more lines of supporting code. If we do find, that the best solution is to pull in behat/transliterate, then that's fine and no issue if done in 5.2

Neos.Eel/Classes/Helper/StringHelper.php
kitsunet: Wait, what do you mean it doesn't take care of that? It doesn't transliterate umlauts to something else?

bwaidelich: I think the example config I posted didn't cover special characters like "€". They could all be added (see http://userguide.icu-project.org/transforms/general).
But the question still remains _what_ rules to apply.. Maybe there's some official mapping for all disallowed URL path characters

kitsunet: Weeell, that's the issue with naming, right. If you want to build a URL with special characters I would argue you urlencode instead of doing some transliteration ;) So whatever is allowed or not there doesn't _really_ matter. This is more of a asciilize I guess? IDK. The problem with transliteration is that there is no universal solution because you loose information. The only way to retain at least part of the information of the original string would be to know it's langauge because the "manual" mapping we do there might be right "for us" but not necessarily for someone else. So either this needs to be given a source language to transliterate for OR must follow basic asciifolding rules, aka ü => u, ú => u etc.

bwaidelich: True that. "disallowed" wasn't the right word.
But there gotta be some standards that define mappings like `ü` => `ue` and similar.. I hoped..

kitsunet: Just to quote from some other text:

> In German the ¨ (umlaut) is always transliterated as `-e`, while in Swedish the diacritic is simply dropped.

There is no universal rule, most diacritics are transliterated differently depending on their source language, so unless you give that only a "flattening" transliteration is sensible.

kitsunet:
General comment on the whole behat translit... Just look at Utility::renderValidNodeName() in CR which uses behat transliterator already. I think back in the  day we had way more troubles with php-intl (which was used there  before) than we ever had with behat translit and I guess btw. it could be nice to use it in routing of custom object properties as well...

Neos.Eel/Classes/Helper/StringHelper.php
bwaidelich: Yes, that's a good call.. It _could_ mean that a helper expects an (optional?) locale or preset but maybe that takes it too far

bwaidelich:
> I think back in the day we had way more troubles with php-intl [...] than we ever had with behat translit 

I guess it has improved quite a bit then. We're using the `\Transliterator` in a project and it is very reliable and much faster than a PHP implementation

bwaidelich:
After letting that sink in a bit I think we should introduce a new service to Flow that internally uses the `\Transliterator` and provides a method like `public function urlize(string $string, Locale $locale): string`.
Because even in one installation it makes sense to generate a different URL depending on the target locale ("Übermäßig" => `/de/uebermaessig`, "На здоровье!" => `/ru/na-zdorove`?)

kitsunet:
Sounds like it should go into neos/utility-unicode or we open a new package ala neos/utility-string or so? I guess there is more that we could/should actually put in there that is now scattered across the code base.

kdambekalns:
Having worked on Flow and Neos for ages I do remember (as @kitsunet mentioned already), that we had massive problems with the Transliterator in PHP. Now `behat/trasliterator` worked "for ages", and still does (if it weren't for PHP 7.4 and the [missing release](https://github.com/Behat/Transliterator/issues/25) of the [fixed version](https://github.com/Behat/Transliterator/pull/29)).

Having just checked out `symfony/strings` as a possible alternative, I must say [trouble](https://github.com/symfony/symfony/issues/35061) is ahead. 

And by all means, please don't make us maintain code for transliteration, ever.

**Bottom line: be careful with this, a wrong turn is taken easily… ;)**

bwaidelich:
> Sounds like it should go into neos/utility-unicode or we open a new package [...]

I agree – for the transliteration service at least. The Eel helper could go into the fusion core IMO (maybe with a soft dependency that only complains about missing dependencies when actually used?)

> we had massive problems with the Transliterator in PHP [...]

I think php-intl got much better recently and I think it would at least be worth re-considering (unless you know that there are still major issues with it of course)

> behat/trasliterator worked "for ages"

But does it really work for "esotheric" languages, too? and is the PHP only implementation fast enough(tm) (i.e. when generating large amounts of strings e.g. for sitemaps). We should at least investigate IMO

> Having just checked out symfony/strings as a possible alternative, I must say trouble is ahead.

Ah, interesting. and nice discussion inbound on the issue :)

> please don't make us maintain code for transliteration, ever.

Absolutely not! The built-in transliterator would only be an option if it works more or less out-of-the box with the default ICU rulesets I guess

> be careful with this, a wrong turn is taken easily… ;)

Agreed!

jonnitto:
Another idea: We have also Eel Helper who depends on Neos. Like `Neos.Caching` for example. What about introducing `Neos.String.urlize` and use the function who generates the `uriPathSegment`? Like that, we would have no new dependency and we would even have support for settings (e.g. `Neos.Neos.transliterationRules`).

PR #1873, opened 2019-12-03
TASK: Remove code dealing with removed class/interface

kdambekalns:
The `FlowSpecificBackendInterface` as well as `Neos\Flow\Cache\Backend`
are gone with Flow 5.0, so this code is never executed.

The `implements CacheFactoryInterface` is redundant, thus removed.


kdambekalns:
> Can you import classes/interfaces that don't exist?

Yes, `use This\Does\Not\Exist` is fine, as is `instanceof \Might\Not\Be\Available`. Which is in fact a good thing!

albe:
Yes, yes it is. Of course. Too much C# work done lately, losing touch with PHP specifics and the benefits...

PR #1875, opened 2019-12-05
BUGFIX: Include entry removal into DB transaction during set

### LINKED ISSUE #1874: PdoBackend: Replacing authenticated accounts should be atomic
### Description

The Neos.UI backend creates a lot of parallel requests for assets property to load previews. Now I’m using the PdoBackend for storing session data.

The request executed Session_Original::storeAuthenticatedAccountsInfo() which executed PdoBackend::set, which then does:

```
$this->remove($entryIdentifier);

...
$statementHandle = $this->databaseHandle->prepare('INSERT INTO "' ...
```
See https://github.com/neos/flow-development-collection/blob/master/Neos.Cache/Classes/Backend/PdoBackend.php#L161

This allows race conditions and causes an exception `Integrity constraint violation: 1062 Duplicate entry `. The current master already introduced a transaction, but I believe the removal should be part of the transaction.

#### Expected behavior

Multiple parallel requests with a valid session all return correct responses.

#### Actual behavior

The ones with race condition errors return 500.

### Affected Versions

Neos: 4.3.3
Flow: 5.3.2

kdambekalns:
This makes sure all DB statements during a `set()` operation are in
one transaction, to avoid race conditions.

Fixes #1874

### DIFF
--- a/Neos.Cache/Classes/Backend/PdoBackend.php
+++ b/Neos.Cache/Classes/Backend/PdoBackend.php
@@ -248,10 +248,10 @@ class PdoBackend extends IndependentAbstractBackend implements TaggableBackendIn
 
         $this->databaseHandle->beginTransaction();
         try {
-            $rowsWhereDeleted = $this->removeWithoutTransaction($entryIdentifier);
+            $rowsWereDeleted = $this->removeWithoutTransaction($entryIdentifier);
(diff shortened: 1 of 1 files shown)


gerhard-boden:
Good catch an nice fix! 👍

PR #1876, opened 2019-12-05
TASK: Some cleanup to the Flow persistence code

kdambekalns:


Neos.Flow/Classes/Persistence/Doctrine/Service.php
```diff
@@ -696,7 +699,7 @@ protected function buildCodeFromSql(Configuration $configuration, array $sql): s
             array_unshift(
                 $code,
                 sprintf(
-                    '$this->abortIf($this->connection->getDatabasePlatform()->getName() !== %s, %s);',
+                    '$this->abortIf($this->connection->getDatabasePlatform()->getName() != %s, %s);',
```
albe: Is that change intentional? Because the variable given could also be `null`?

Neos.Flow/Classes/Persistence/Doctrine/Service.php
```diff
@@ -641,27 +646,25 @@ class $className extends AbstractMigration
     /**
      * @return string
      */
-    public function getDescription(): string
+    public function getDescription()
```
albe: Return type removed? Was this breaking for some cases?

albe:
ping @kdambekalns - would merge if my two questions are answered :)

Neos.Flow/Classes/Persistence/Doctrine/Service.php
kdambekalns: It should match the signature from Doctrine… And in `doctrine/migrations` 1.8 that is without.

Neos.Flow/Classes/Persistence/Doctrine/Service.php
kdambekalns: No, should be  `!==` as per https://github.com/doctrine/migrations/blob/1.8/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php#L147

albe:
```suggestion
                    '$this->abortIf($this->connection->getDatabasePlatform()->getName() !== %s, %s);',
```

kdambekalns:
Those changes are part of #1880 and/or #2231 as well…

PR #1877, opened 2019-12-06
TASK: Add a primary key to "tags" table for MySQL/MariaDB

kdambekalns:
Note: No migrations exist for the PDO cache backend!

To make sure the cache and tags tables are up-to-date, you need to
remove them manually and the run the cache:setup CLI command to have
them recreated with the most recent definition.

Unless you have persistent caches, that should be no problem, but
give that a thought first.

The reasoning behind this change is explained in the documentation
of MySQL and MariaDB.

For MySQL https://dev.mysql.com/doc/refman/8.0/en/innodb-index-types.html
explains:

> If the table has no PRIMARY KEY or suitable UNIQUE index, InnoDB
> internally generates a hidden clustered index named GEN_CLUST_INDEX
> on a synthetic column containing row ID values. The rows are ordered
> by the ID that InnoDB assigns to the rows in such a table. The row
> ID is a 6-byte field that increases monotonically as new rows are
> inserted. Thus, the rows ordered by the row ID are physically in
> insertion order.

For MariaDB it is very similar:

> In XtraDB/InnoDB tables, all indexes contain the primary key as a
> suffix. Thus, when using this storage engine, keeping the primary
> key as small as possible is particularly important. If a primary
> key does not exist and there are no UNIQUE indexes, InnoDB creates
> a 6-bytes clustered index which is invisible to the user.

(see https://mariadb.com/kb/en/library/getting-started-with-indexes/)

The bottom line is, InnoDB actually needs a PK and creates a "hidden"
one, if needed. So it takes the guesswork out of the table setup, to
actually define one.

Even more details can be found online in this article and it's links:
https://federico-razzoli.com/why-mysql-tables-need-a-primary-key

### DIFF
--- a/Neos.Utility.Pdo/Classes/PdoHelper.php
+++ b/Neos.Utility.Pdo/Classes/PdoHelper.php
@@ -20,9 +20,15 @@ abstract class PdoHelper
     /**
      * Pumps the SQL into the database. Use for DDL only.
      *
+     * The given $pathAndFilename is expanded to check for a driver-specific SQL
+     * file to import, so "/some/path/to/mysql.statements.sql" is checked and used
+     * if "/some/path/to/statements.sql" is passed and $pdoDriver is "mysql".
+     * Otherwise the given file is used.
+     *
      * Important: key definitions with length specifiers (needed for MySQL) must
      * be given as "field"(xyz) - no space between double quote and parenthesis -
-     * so they can be removed automatically.
+     * so they can be removed automatically. This is not done for driver-specific
+     * SQL, though.
      *
      * @param \PDO $databaseHandle
      * @param string $pdoDriver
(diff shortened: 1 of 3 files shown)


kitsunet:
Note to self, "context" should be renamed to applicationIdentifier in accordance with how it works inside the cache package.

kdambekalns:
Failure was correct, creating a PK is still a good idea, creating it over all four columns would not be a good idea. Please re-review this solution…

Neos.Utility.Pdo/Classes/PdoHelper.php
```diff
@@ -31,10 +37,17 @@ abstract class PdoHelper
...
-        if ($pdoDriver !== 'mysql') {
-            $sql = preg_replace('/"\([0-9]+\)/', '"', $sql);
+        $path = dirname($pathAndFilename);
+        $filename = basename($pathAndFilename);
+        $dbSpecificPathAndFilename = sprintf('%s/%s.%s', $path, $pdoDriver, $filename);
+        if (file_exists($dbSpecificPathAndFilename)) {
+            $sql = file($dbSpecificPathAndFilename, FILE_IGNORE_NEW_LINES & FILE_SKIP_EMPTY_LINES);
+        } else {
+            $sql = file($pathAndFilename, FILE_IGNORE_NEW_LINES & FILE_SKIP_EMPTY_LINES);
+            // Remove MySQL style key length delimiters (yuck!) if we are not setting up a MySQL db
```
kitsunet: Isn't this e superfluous by now, given that we have a mysql specific file?

kdambekalns: **We** have that file, for our specific case. But the `PdoHelper` can be used by anyone for any case. I'm not going to break that now… ;)

kitsunet: Oh of course! That makes sense :)

bwaidelich: I already broke it with #1764 🤣

kdambekalns: Well, you intend to break it. :)

Issue #1878, opened 2019-12-06
String Keys are converted to Integer in the TransientMemoryBackend


daniellienert:
### Description

When you have:
- IntegerLike string keys for cache entries (eg. "123"), which is allowed when set as string
- The transient Backend the stores that Strings as Key in an array for the tag -> entry mapping
- PHP interpretes that string as integer keys 
- When looping over that array to retrieve the entries, the Frontend complains about the used integer-keys where only strings are allowed

### Affected Versions

> 6.0

PR #1879, opened 2019-12-06
BUGFIX: Cast identifiers to string when calling frontent / backend methods

daniellienert:
Fixes: #1878

### DIFF
--- a/Neos.Cache/Classes/Backend/TransientMemoryBackend.php
+++ b/Neos.Cache/Classes/Backend/TransientMemoryBackend.php
@@ -142,7 +142,7 @@ class TransientMemoryBackend extends IndependentAbstractBackend implements Tagga
     {
         $identifiers = $this->findIdentifiersByTag($tag);
         foreach ($identifiers as $identifier) {
-            $this->remove($identifier);
+            $this->remove((string)$identifier);
         }
         return count($identifiers);
     }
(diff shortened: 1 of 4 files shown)


daniellienert:
If this prevents PHP from doing magic, I also prefer this one.

kdambekalns:
> If this prevents PHP from doing magic

No, it undoes the magic. :)

daniellienert:
@kdambekalns you are a Un-Magician :)
Works, added two tests to prove it.

Neos.Cache/Classes/Backend/TransientMemoryBackend.php
```diff
@@ -142,7 +147,7 @@ public function flushByTag(string $tag): int
     {
         $identifiers = $this->findIdentifiersByTag($tag);
         foreach ($identifiers as $identifier) {
-            $this->remove($identifier);
+            $this->remove((string)$identifier);
```
kdambekalns:
```suggestion
            $this->remove($identifier);
```

daniellienert: absolutely

kitsunet:
Definitely better than before but I guess a “0123” would come back as “123” still?

kdambekalns:
> Definitely better than before but I guess a “0123” would come back as “123” still?

No. The leading zero makes it work:

```
php > var_dump(['123'=>'foo']);
array(1) {
  [123] => string(3) "foo"
}

php > var_dump(['0123'=>'foo']);
array(1) {
  '0123' => string(3) "foo"
}
```

Thankfully it does not cast it as octal like this:

```
php > var_dump([0123=>'foo']);
array(1) {
  [83] => string(3) "foo"
}
```

kitsunet:
Thanks, weird PHP is weird, but good that it works :)

albe:
Yeah, that's what I checked first :)

PR #1880, opened 2019-12-06
!!! TASK: Update Doctrine Migrations to 3.0

### LINKED ISSUE #2122: Update doctrine dependencies
doctrine/common and doctrine/migrations are due for an update for some time. Doctrine/common is even supposed to be phased out (https://www.doctrine-project.org/2018/07/12/common-2-9-and-dbal-2-8-and-orm-2-6-2.html). Doctrine/orm is still okay with 2.7, but 2.8.x is in the making.
see #1880, #2026 & #2021

kdambekalns:
This updated the required version of `doctrine/migrations` from 1.8 to 3.0.

While there are new features in Doctrine Migrations, the reason for us to do
an upgrade is to move forward – the previously used version will not be
maintained forever… This post also gives some background on that:
https://www.doctrine-project.org/2020/04/10/doctrine-migrations-3.0.html

For a Flow user the commands remain unchanged, so far no multi-namespace
migrations are supported and the features to the "official" CLI do not matter,
since we embed the functionality in our own commands.

**Breaking changes**

There are three things that make this upgrade a breaking change:

- `Doctrine\DBAL\Migrations` moved to `Doctrine\Migrations`
- `AbstractMigration` changed method signatures (type delcarations added)

To adjust your PHP code (the migration files), a core migration is provided that
should fix the vast majority of existing migrations. (That core migration is in Flow
and named `Version20201109224100`.)

- The "version" is the FQCN of the migration class (existing entries in the migrations table will be automatically updated)

The needed changes to the DB table where the migration status is stored are done
the first time a command that accesses that table is used. Make sure to have a current
backup and then run `./flow doctrine:migrationstatus --show-migrations`. If all
went well, the migrations should all be listed as a fully-qualified class name, no
longer just a date/time string. If any errors occurred during the command, restore the
backup (the migrations table is sufficient), fix the errors and try again.

See https://github.com/doctrine/migrations/blob/3.0.x/UPGRADE.md#code-bc-breaks
and https://github.com/doctrine/migrations/blob/3.0.x/UPGRADE.md#upgrade-to-20
for a full list of other changes. Most of those are wrapped in Flow code and need no
adjustments in userland code.

Resolves #2122


sorenmalling:
@kdambekalns Would you like me to go over this for doctrine/orm 2.2.* ?
Did you already do any research and know of any blockers?

kdambekalns:
Oh, yes, I did do things… here are three branches that complement each other:

- https://github.com/kdambekalns/neos-development-distribution/tree/task/doctrine-migrations-update
- https://github.com/kdambekalns/flow-development-collection/tree/task/update-doctrine-migrations
- https://github.com/kdambekalns/neos-development-collection/tree/task/update-doctrine-migrations

albe:
D'oh, now we have Doctrine Migrations 3.x - see #2026. We might also need to update doctrine/common with this #2021

kdambekalns:
I'll look into this (again), now that we have 7.0 coming it seems worth it! 💪

albe:
Can we still get this in for 7.0?

kdambekalns:
Good question… sorry, I didn't get around to it. @jonnitto Is the RM to ask, I guess. What are your thoughts, worth working on it today?

kdambekalns:
These command are available and need to be working (again):

- [ ]  `doctrine:validate`                        Validate the class/table mappings
- [ ]  `doctrine:create`                          Create the database schema
- [ ]  `doctrine:update`                          Update the database schema
- [ ]  `doctrine:entitystatus`                    Show the current status of entities
- [ ]  `doctrine:dql`                             Run arbitrary DQL and display
- [ ]  `doctrine:migrationstatus`                 Show the current migration status
- [ ]  `doctrine:migrate`                         Migrate the database schema
- [ ]  `doctrine:migrationexecute`                Execute a single migration
- [ ]  `doctrine:migrationversion`                Mark/unmark migrations as migrated
- [ ]  `doctrine:migrationgenerate`               Generate a new migration

Neos.Flow/Classes/Persistence/Doctrine/Service.php
```diff
@@ -530,21 +528,27 @@ public function markAsMigrated($version, $markAsMigrated)
...
             if ($filterExpression) {
-                $connection->getConfiguration()->setFilterSchemaAssetsExpression($filterExpression);
+                $connection->getConfiguration()->setSchemaAssetsFilter(
+                    static function (string $assetName) use ($filterExpression) {
+                        if ($assetName instanceof AbstractAsset) {
+                            $assetName = $assetName->getName();
+                        }
+
+                        return preg_match($filterExpression, $assetName);
+                    });
```
albe, line 551: CGL 🤷‍♂️
```suggestion
                    }
                );
```

kdambekalns:
Woah, I'm at it… hold your horses. 😎

albe:
Just fixing the obtrusive github thingies :D sorry! I'll hold back now

kdambekalns:
> We might also need to update doctrine/common with this #2021, though it's still supposed to be phased out. So maybe we can even just get rid of it?

    doctrine/orm                      2.7.4        requires  doctrine/common (^2.11 || ^3.0)

We can leave it as is, technically. Or update it…

kdambekalns:
One blocker with this as it is right now: the versions marked as applied in the database have a new "name", and thus are not seen as migrated.

kdambekalns:
If we did not have to care for all the old code… https://www.goetas.com/blog/multi-namespace-migrations-with-doctrinemigrations-30/ sounds cool.

sorenmalling:
> One blocker with this as it is right now: the versions marked as applied in the database have a new "name", and thus are not seen as migrated.

Something a database migration could rename in the doctrine_migration_status table? How did the name change?

> If we did not have to care for all the old code… https://www.goetas.com/blog/multi-namespace-migrations-with-doctrinemigrations-30/ sounds cool.

I believe we can look at that for 7.1 perhaps?

jonnitto:
For flow, @bwaidelich is the RM…

bwaidelich:
I completely missed this one (just saw it now because Jon mentioned me).
I struggle with that one to be honest.. a +752 −734 Change after Feature Freeze that is marked breaking but with not explanation as to why it is breaking.

I'm afraid with the remaining last-minute features I don't have the time to get into this enough to assure a smooth transition for the 7.0 release.. Can someone summarize why we urgently need to upgrade to a new major dependency (and why that wouldn't make this a FEATURE but a TASK) and how the behavior changed such that this could be breaking – and what to do about it?

kdambekalns:
> Something a database migration could rename in the doctrine_migration_status table? How did the name change?

That migration would need to be run using migrations, which would… anyway, it's supposed to get fixed automagically. Debugging that right now…

kdambekalns:
Now all migrations are found, and if there are no PHP errors in any of the migrations (i.e. if core:migrate has been run on all installed packages to fix the method signatures!) the recorded migration entries are now upgraded automagically the first time the `TableMetadataStorage` of Doctrine is used.

albe:
> Can someone summarize why we urgently need to upgrade to a new major dependency

I guess the main reason is this: we still run on a pretty old version of doctrine/migrations (^1.8, current 3.0) and doctrine/commons (^2.4, current 3.0 but is considered deprecated), but upgrading dependencies to a new major version is to be considered breaking. So if we don't do this, we have to wait for another year.

> how the behavior changed such that this could be breaking

Karsten might be able to say a bit more in detail about this, but as seems the change of the migration name in the DB is one thing. The namespace of the abstract base class for a migration changed and some method signatures (type hints) - see https://github.com/neos/flow-development-collection/pull/1880/files#diff-48cebe019a072e0a9340ef4f6a2d81dc24dd3b7faa47490fcb05731dac101669R35-R37
Otherwise it's more of a "breaking by principle" I guess.

> We can leave it as is, technically. Or update it…

@kdambekalns then let's at least raise the minimum version of doctrine/commons to ^2.9 - that will at least help composer find a "good" version

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
```diff
@@ -1069,7 +1069,7 @@ detected between the current schema and the current models in the system:
 
 	namespace Neos\Flow\Persistence\Doctrine\Migrations;
 
-	use Doctrine\DBAL\Migrations\AbstractMigration,
+	use Doctrine\Migrations\AbstractMigration,
```
bwaidelich, line 1072: the method signatures in this example have to be adjusted, too. Right?

bwaidelich:
Thanks for clarifying!
We should explain in the PR description what devs have to do to get their code in sync and/or provide core migrations if possible.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Persistence.rst
kdambekalns: Good point!

kdambekalns:
@albe 
> then let's at least raise the minimum version of doctrine/commons to ^2.9 - that will at least help composer find a "good" version

See https://github.com/neos/flow-development-collection/issues/2122#issuecomment-724515813 - and that clearly allows 2.9 - different topic, though.

@bwaidelich 
> We should explain in the PR description …

Of course. Back when that PR was created, no one knew. Now I do. 😉

kdambekalns:
How to test this:

1. In your dev dist (Neos or Flow, doesn't matter) run `composer require doctrine/migrations:'3.0.1 as 1.8.99'`
2. in `Packages/Framework` check out the PR branch
3. Now the versions should be correct

Then

- run the core migration on anything you have installed besides the Framework itself
- run `doctrine:migrationstatus --show-migrations`
- migrate things up and down a bit
- try `doctrine:migrationgenerate`

kdambekalns:
LMGHTFY: https://github.com/neos/neos-development-collection/search?q=gedmo%2Fdoctrine-extensions

composer.json
```diff
@@ -22,7 +22,7 @@
         "psr/http-server-handler": "^1.0",
         "ramsey/uuid": "^3.0 || ^4.0",
         "doctrine/orm": "^2.6",
-        "doctrine/migrations": "^1.8",
+        "doctrine/migrations": "^3.0",
         "doctrine/dbal": "^2.9",
```
albe, line 26:
```suggestion
        "doctrine/dbal": "^2.11",
```
See comment above

Neos.Flow/composer.json
```diff
@@ -38,7 +38,7 @@
         "ramsey/uuid": "^3.0",
 
         "doctrine/orm": "^2.6",
-        "doctrine/migrations": "^1.8",
+        "doctrine/migrations": "^3.0",
         "doctrine/dbal": "^2.9",
```
albe, line 42:
```suggestion
        "doctrine/dbal": "^2.11",
```
See comment above

Neos.Flow/Classes/Persistence/Doctrine/Service.php
```diff
@@ -11,28 +13,47 @@
  * source code.
  */
 
-use Doctrine\DBAL\DBALException;
-use Doctrine\DBAL\Migrations\Configuration\Configuration;
-use Doctrine\DBAL\Migrations\Migration;
-use Doctrine\DBAL\Migrations\MigrationException;
-use Doctrine\DBAL\Migrations\OutputWriter;
-use Doctrine\DBAL\Migrations\Version;
+use Doctrine\DBAL\Exception as DBALException;
```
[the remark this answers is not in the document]
kdambekalns: `doctrine/dbal` as `^2.11` raises PHP to `^7.3`… Unless that's ok, I'd rather continue to use the deprecated `DBALException`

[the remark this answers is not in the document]
albe: 👍 would be yet another good reason to raise minimum PHP version though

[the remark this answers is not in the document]
kdambekalns: And so be it! 🎉

kdambekalns:
One thing still different from before: output during migrations no longer shows what's happening. Need to figure out where the output goes (probably logger in the dependency factory needs to be populated and then read from)…

kdambekalns:
One more thing: go over CLI output in the documentation and update as needed…

Neos.Flow/composer.json
```diff
@@ -37,10 +37,10 @@
 
         "ramsey/uuid": "^3.0",
```
albe, line 38: Just noticed, that should be `"^3.0 || ^4.0"` to match the collection composer.json, no?

kdambekalns: Yes, obviuosly someone changed only the autogenerated collection manifest in https://github.com/neos/flow-development-collection/commit/99a6c74b193d810ea6274ab51726cc71db48cc13 😎

kdambekalns: Ah, that was a depdendabot PR.

albe:
Regarding the psalm failures left now - guess we need to update the docblocks in our `Doctrine\Query`. We'll need to baseline the `QueryInterface`->`Doctrine\Query` discrepancy though.
Most other look okay, but there's also
```
Docblock-defined class or interface Neos\Flow\Persistence\Doctrine\Migrations\MigrationException does not exist
Docblock-defined class or interface Doctrine\DBAL\Migrations\AbortMigrationException does not exist
```

kdambekalns:
Let's see what is left now…

> We'll need to baseline the `QueryInterface`->`Doctrine\Query` discrepancy though.

Or we adjust that along with #2231 and relax it a bit… 🤷‍♂️

albe:
Yeah, we could do this. I mean at this point it's just a cosmetic change, which can happen any time.

bwaidelich:
Can someone give me an update what needs to be done here?
I'd say we should try to get it in by tomorrow or postpone it to some later release

albe:
AFAIS only a psalm-update-baseline

kdambekalns:
The tests should pass with a psalm baseline update (I'll do that now).

Then we have two items from https://github.com/neos/flow-development-collection/pull/1880#issuecomment-725296651 and finally https://github.com/neos/flow-development-collection/pull/1880#issuecomment-724249509 - these are issues that could be addressed in a follow-up PR, if you agree…

Oh, and of course we need to adjust the DB migrations in Neos and our other packages…

kdambekalns:
Well… that seems to be a mismatch betweeen Psalm versions. @albe Can you check this and maybe do the baseline update for now?

albe:
`<files psalm-version="3.18.2@19aa905f7c3c7350569999a93c40ae91ae4e1626">` yep - you need to install psalm ~3.10.0, or maybe we should finally add that as requirement to the dev-distribution

albe (approved):

kdambekalns:
Thanks for updating the baseline, @albe!

> yep - you need to install psalm ~3.10.0

Hm… what about something new? https://github.com/neos/flow-development-collection/pull/2205 suggests that. Let's track this as https://github.com/neos/flow-development-collection/issues/2243

kdambekalns:
I created a new issue for the known open tasks - so this could go in fast(er) for easier testing. 😬

albe:
> Hm… what about something new?

Yeah, I have that on my radar, but I think we should then try to keep versions across our branches as close as possible (optimally in sync), because upmerging the baseline might otherwise become a hassle.

PR #1883, opened 2019-12-13
TASK: Add some more statistics badges

kdambekalns:
Adds badges on commits sine release, % open issues and resolution
time to Readme.rst.


kdambekalns:
Something I had still lying around since ages ago…

PR #1888, opened 2019-12-17
BUGFIX: Fix error handling for importing resources

j6s:
While working with `ResourceManager` I noticed that importing non-existent resources would fail later in the process than expected: They would only fail when trying to copy the temporary file to it's persistent location instead of failing when fetching the original file.

**What I did**
This PR fixes the incorrect error handling of fetching files when importing resources.

**How I did it**
The code assumed that `copy` would throw an exception if it failed - however, it returns `false` in that case.

**How to verify it**
Try to import a non-existent resource (e.g. http://example.com/this-file-does-not-exist). It should now correctly throw the `Could not copy the file from "..." to temporary file "..."` exception  from `WritableFileSystemStorage::importResource` instead of the much later `The temporary file of the file import could not be moved to the final target "..."` from `WritableFileSystemStorage::moveTemporaryFileToFinalDestination`

**Checklist**


albe:
D'uh, 4.3 is already in security-only mode... so... yeah well, consider this an exceptionally important bugfix :D I'll take care of upmerging now, so this doesn't get lost until 4.x actually receives a security fix and is upmerged.

PR #1891, opened 2019-12-21
BUGFIX: Correctly reflect type of aliased classes in method typehints

albe:
The namespace change introduced with doctrine/persistence 1.3.x makes use of `class_alias` to create a b/c layer over the old class names (see https://github.com/greg0ire/type_deprecation_experiment). However, this broke for a couple of our depending packages that still used the old `Doctrine\Common\Persistence\ObjectMananger` class as a method typehint.
The reason is, that PHP Reflection returns different values for aliased method parameter types when using `ReflectionParameter::getClass()->getName()` vs. `ReflectionParameter::getType()->getName()`.
The former returns the name of the actual class, instead of the alias, while the latter returns the actual name that was used in the typehint, regardless if it is an alias only.

In turn, this led our dependency injection to try to inject classes that it was not aware of and led to issues like https://github.com/yeebase/Yeebase.TwoFactorAuthentication/issues/3 or https://github.com/neos/Neos.EventSourcing/issues/249
It was not an issue for injections that were not typehinted via PHP, e.g. `@Flow\Inject` instead of setter/constructor injection.

This change fixes that retrospectively, by always reflecting the actual class name that is specified in the typehint and hence allowing usage of `class_alias`.

Thanks a lot @alcaeus for helping me figure this out!

### DIFF
--- a/Neos.Flow/Classes/Reflection/ParameterReflection.php
+++ b/Neos.Flow/Classes/Reflection/ParameterReflection.php
@@ -56,13 +56,10 @@ class ParameterReflection extends \ReflectionParameter
      */
     public function getBuiltinType()
     {
-        if (!is_callable([$this, 'getType'])) {
-            return null;
-        }
         $type = $this->getType();
         if ($type === null || !$type->isBuiltin()) {
             return null;
         }
-        return (string)$type;
+        return $type instanceof \ReflectionNamedType ? $type->getName() : (string)$type;
     }
 }
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1766,11 +1766,11 @@ class ReflectionService
                 'position' => $parameterData[self::DATA_PARAMETER_POSITION],
                 'optional' => isset($parameterData[self::DATA_PARAMETER_OPTIONAL]),
                 'type' => $parameterData[self::DATA_PARAMETER_TYPE],
-                'class' => isset($parameterData[self::DATA_PARAMETER_CLASS]) ? $parameterData[self::DATA_PARAMETER_CLASS] : null,
+                'class' => $parameterData[self::DATA_PARAMETER_CLASS] ?? null,
                 'array' => isset($parameterData[self::DATA_PARAMETER_ARRAY]),
                 'byReference' => isset($parameterData[self::DATA_PARAMETER_BY_REFERENCE]),
                 'allowsNull' => isset($parameterData[self::DATA_PARAMETER_ALLOWS_NULL]),
-                'defaultValue' => isset($parameterData[self::DATA_PARAMETER_DEFAULT_VALUE]) ? $parameterData[self::DATA_PARAMETER_DEFAULT_VALUE] : null,
+                'defaultValue' => $parameterData[self::DATA_PARAMETER_DEFAULT_VALUE] ?? null,
                 'scalarDeclaration' => isset($parameterData[self::DATA_PARAMETER_SCALAR_DECLARATION])
             ];
         }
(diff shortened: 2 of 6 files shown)


albe:
Note: The first commit only adds the test that shows the minimal breaking behaviour for verification. Commit with the actual fix coming.

Neos.Flow/Classes/Reflection/ParameterReflection.php
```diff
@@ -56,13 +56,10 @@ public function getClass()
      */
     public function getBuiltinType()
     {
-        if (!is_callable([$this, 'getType'])) {
```
albe: Was only necessary for PHP 5.x compatibility back then. No longer needed.

Neos.Flow/Classes/Reflection/ParameterReflection.php
```diff
@@ -56,13 +56,10 @@ public function getClass()
...
-            return null;
-        }
         $type = $this->getType();
         if ($type === null || !$type->isBuiltin()) {
             return null;
         }
-        return (string)$type;
```
albe: `__toString()` is deprecated on ReflectionType as of PHP 7.1

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1785,7 +1785,7 @@ protected function convertParameterDataToArray(array $parametersInformation)
      * @param MethodReflection $method The parameter's method
      * @return array Parameter information array
      */
-    protected function convertParameterReflectionToArray(ParameterReflection $parameter, MethodReflection $method = null)
+    protected function convertParameterReflectionToArray(ParameterReflection $parameter, MethodReflection $method)
```
albe: We never ever call this protected method without a `MethodReflection` so...

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1803,32 +1803,33 @@ protected function convertParameterReflectionToArray(ParameterReflection $parame
...
-                $explodedParameters = explode(' ', $paramAnnotations[$parameter->getPosition()]);
-                if (count($explodedParameters) >= 2) {
-                    $parameterType = $this->expandType($method->getDeclaringClass(), $explodedParameters[0]);
-                    $parameterInformation[self::DATA_PARAMETER_TYPE] = $this->cleanClassName($parameterType);
-                }
+        $paramAnnotations = $method->isTaggedWith('param') ? $method->getTagValues('param') : [];
+        if (isset($paramAnnotations[$parameter->getPosition()])) {
+            $explodedParameters = explode(' ', $paramAnnotations[$parameter->getPosition()]);
+            if (count($explodedParameters) >= 2) {
+                $parameterType = $this->expandType($method->getDeclaringClass(), $explodedParameters[0]);
```
albe: This is still needed to override the PHP typehint in order to allow generics/collection element typehints.

albe:
Hm, seems PHP 7.4 does somehow not work out.

albe:
Related to https://github.com/neos/Neos.EventSourcing/issues/249

Neos.Flow/Classes/Reflection/ReflectionService.php
bwaidelich: Since `ReflectionService` is not final and this method not private, this could still break things – it's very unlikely though

albe: True. I mean we can postpone this change to master too, not picky here. But I also don't see this super internal method problematic either.

bwaidelich: Yes, I agree and don't think it's necessary to postpone this

albe:
I'd like to merge this soonish @kdambekalns @kitsunet - so if you could take a quick check this is okay as is, that would be nice.

albe:
Merging now to finally get doctrine/persistence 1.3 compatibility in - lost track of this myself 🙈

Issue #1866, opened 2019-11-30
Support for PHP 7.4


bwaidelich:
PHP 7.4 will be released soon and we should make sure at least Flow 6.1 supports it

kdambekalns:
@albe I am trying PHP 6.4 with Flow 5.3 now…

kdambekalns:
Ok, one more dependency we have that is not ready: https://github.com/Behat/Transliterator/pull/29

kdambekalns:
I directly pushed some further changes to `.travis.yml` in Flow, switching to Ubuntu Bionic and stable PHP 7.4 as well as PostgreSQL 9.5 (instead of 9.4).

kdambekalns:
The same adjustments are in the Neos PR now…

kdambekalns:
The PR to `behat/transliterator` was merged some minutes ago…

kdambekalns:
Everything merged, all builds green, release jobs are running. Finally.

jonnitto:
So we could close this issue, right?

PR #1896, opened 2020-01-15
BUGFIX: Fix string offset access syntax with curly braces

kdambekalns:
Related to https://github.com/neos/flow-development-collection/issues/1866


kdambekalns:
Failed in Psalm tests, those are fixed in 6.0 by now. Thus merged using override.

PR #1897, opened 2020-01-15
BUGFIX: PDOBackend doesn't respect configuration

Sebobo:
**What I did**

* Use the table names from the configuration in all places.

### DIFF
--- a/Neos.Cache/Classes/Backend/PdoBackend.php
+++ b/Neos.Cache/Classes/Backend/PdoBackend.php
@@ -170,13 +170,13 @@ class PdoBackend extends IndependentAbstractBackend implements TaggableBackendIn
         try {
             $this->removeWithoutTransaction($entryIdentifier);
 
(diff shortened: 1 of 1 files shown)


Sebobo:
Relates also to #1764

Neos.Cache/Classes/Backend/PdoBackend.php
```diff
@@ -518,6 +518,12 @@ public function valid(): bool
      */
     public function rewind()
     {
+        try {
+            $this->connect();
+        } catch (Exception $e) {
+            return;
+        }
```
bwaidelich: Is that related?

bwaidelich: ah, you mention it in the PR description..
I always struggle a bit when I see a `catch` block that is ignored.

Sebobo: Should maybe have been a separate PR.

Sebobo: Alternative would be to override getIterator and call connect there.

PR #1898, opened 2020-01-16
BUGFIX: Prevent connection error when calling getIterator for PdoBackend

Sebobo:
**What I did**

Call `connect` in `rewind` method to make sure that a connection exists.

### DIFF
--- a/Neos.Cache/Classes/Backend/PdoBackend.php
+++ b/Neos.Cache/Classes/Backend/PdoBackend.php
@@ -518,6 +518,12 @@ class PdoBackend extends IndependentAbstractBackend implements TaggableBackendIn
      */
     public function rewind()
     {
+        try {
+            $this->connect();
+        } catch (Exception $e) {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


Sebobo:
I decided for now to not handle the exception as many methods in this method would be affected and the exception would have to be forwarded.

Another way would be to override `getIterator` instead and verify the connection there.

robertlemke:
Some thoughts:

In general, we should go into the direction of implementing retries for connection issues with databases and other services, like Redis.

But that's a bigger effort which should follow a little plan, so it's implemented consistently. For example, it would be good to have a shared implementation for realizing exponential back-off loops.

Sebobo:
There are two parts to this. On the one hand improving the behaviour of `connect` itself. 
On the other that `connect` is always called before the `databaseHandle` is used.

I would like to fix the second part for now. So we can either call `connect` in all public methods, or during initialise. But the latter might run a connect which we don't even need in some cases.

PR #1901, opened 2020-01-23
BUGFIX: Parse single-line docblock comments correctly

### LINKED ISSUE #1900: Proxy compilation breaks on one-line docblocks
### Description

When using a one-line docblock comment (e.g. if there is only a single thing to comment on) the generated classes contain 2 end comment strings if the method is proxied (e.g. because it is wrapped by an aspect)

Source file:
```
/** @return Foo[] */
public function doSomething() {}
```

Destination File:
```
/**
 * Autogenerated Proxy Method
 * @return Foo[] */
 */
public function doSomething() {}
```

I have not investigated this a whole lot, but this seems to come down to `DocCommentParser::parseDocComment` interpreting the comment end characters as a comment.

### Steps to Reproduce

1. Create a method with a single-line docblock
2. Make sure it gets a proxy method by adding an aspect for it
3. The resulting compiled classes will not run due to syntax errors

#### Expected behavior

The outputted docblock should be a valid comment such as the following:

```
/**
 * Autogenerated Proxy Method
 * @return Foo[]
 */
```

#### Actual behavior

```
/**
 * Autogenerated Proxy Method
 * @return Foo[] */
 */
public function doSomething() {}
```

### Affected Versions

Only tested it with the following versions:
* Neos: 4.3.10
* Flow: 5.3.11

j6s:
Fixes #1900

**What I did**
Single line docblocks would break AOP compilation because the end-comment string `*/` would be interpreted as part of the comment.

**How I did it**
The end-comment string `*/` is now removed from all lines before parsing them. This means that the previous check for the end of the comment `if ($line === '*/`)` no longer works but as far as I could determine this explicit check is not actually needed.

**How to verify it**
There are tests

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Reflection/DocCommentParser.php
+++ b/Neos.Flow/Classes/Reflection/DocCommentParser.php
@@ -47,10 +47,7 @@ class DocCommentParser
 
         $lines = explode(chr(10), $docComment);
         foreach ($lines as $line) {
-            $line = trim($line);
-            if ($line === '*/') {
-                break;
-            }
+            $line = trim(str_replace('*/', '', $line));
             if ($line !== '' && strpos($line, '* @') !== false) {
                 $this->parseTag(substr($line, strpos($line, '@')));
             } elseif (count($this->tags) === 0) {
--- a/Neos.Flow/Classes/Reflection/DocCommentParser.php
+++ b/Neos.Flow/Classes/Reflection/DocCommentParser.php
@@ -47,7 +47,7 @@ class DocCommentParser
 
         $lines = explode(chr(10), $docComment);
         foreach ($lines as $line) {
-            $line = trim(str_replace('*/', '', $line));
+            $line = trim(preg_replace('/\*\/$/', '', $line));
             if ($line !== '' && strpos($line, '* @') !== false) {
                 $this->parseTag(substr($line, strpos($line, '@')));
             } elseif (count($this->tags) === 0) {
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Reflection/DocCommentParser.php
```diff
@@ -47,10 +47,7 @@ public function parseDocComment($docComment)
 
         $lines = explode(chr(10), $docComment);
         foreach ($lines as $line) {
-            $line = trim($line);
-            if ($line === '*/') {
-                break;
-            }
+            $line = trim(str_replace('*/', '', $line));
```
albe: The only thing I fear happening is that some docblocks somehow might contain a `*/` somewhere with a meaning, hence I'd probably have used an approach to only remove it from the end of the line. But then again, how would that be a proper docblock? All I could find googling was https://stackoverflow.com/questions/59688653/i-need-to-use-a-splat-operator-within-a-php-docblock-and-cannot-figure-out-how-t which states that to "encode" a `*/` within a docblock, you should use `{@*}` so that would still be safe... dunno, I'm just overly precautious I guess 🙈

Neos.Flow/Tests/Unit/Reflection/DocCommentParserTest.php
```diff
@@ -38,4 +38,24 @@ public function eolCharacterCanBeNewlineOrCarriageReturn()
...
         $this->assertEquals(['$foo integer', '$bar string'], $parser->getTagValues('var'));
     }
+
+    /**
+     * @test
+     */
+    public function singleLineTagIsParsedCorrectly()
+    {
+        $parser = new DocCommentParser();
+        $parser->parseDocComment('/** @return Foo[] */');
```
albe: Tests! Yay! 🎉 ❤️

Neos.Flow/Classes/Reflection/DocCommentParser.php
j6s: That's a good point - let me `preg_replace` that instead.

albe:
Merging now to finally get this landed - long due. Thanks again!

PR #1903, opened 2020-01-31
!!! FEATURE: Arrays::getValueByPath with escaped dot

gerdemann:
It is now possible to escape the dot in the path of Arrays::getValueByPath.
With this you get array keys, which have a dot.

Example:
```
$foo = ['bar.baz' => 'value'];
$path = 'bar\.baz';
```

**What I did**
It is now possible to escape the dot in the path of Arrays::getValueByPath.

**How I did it**
I have adapted the method Arrays::getValueByPath

**How to verify it**
I wrote a test for it.

**Checklist**


albe:
Technically that's a feature, so should go into next release. But then again, escaping a dot might be an expected enough thing that we could relabel it as "BUGFIX"...

Anyway, one thing I'm unsure about:
What if the `\` is an expected part of the path, i.e. 
```php
$foo = ['bar\\' => ['baz' => 'value']];
```
I know, looks weird, but assume the key comes from a path traversal and stores directory names on a windows machine... (whole lot of edgy cases).
I think in that case this could be breaky, but more importantly, it would need an "escape escape" hatch, which this doesn't cover yet.

gerdemann:
@albe I have adapted the regular expression so that your example can be escaped. This makes it a breaking change, I think.
I am not a regex expert, if someone has a better solution I would be happy.

kitsunet:
Why would it be expected? I don't even know why it should be possible in the first place?

This is an application specific syntax expression (aka path syntax) why does it need escaping support?

gerdemann:
With this change it is possible to use the command 'configuration:show' also for e.g. NodeTypes.
```
./flow configuration:show --type NodeTypes --path Vendor\\Site:My\\NodeType

**Comment by kitsunet** [comment]
`./flow configuration:show --type NodeTypes --path "Vendor.Site:My.NodeType"`

should do fine?

**Comment by kitsunet** [comment]
Note: No it doesn't got it now!

**Comment by gerdemann** [comment]
It would also be nice to do something like this:
```
./flow configuration:show --type NodeTypes --path Vendor\\Site:My\\NodeType.superTypes.Vendor\\Site:My\\OtherNodeType

kitsunet:
Shouldn't we then rather think about a nice syntax extension for the scope? 

Eg. `[Vendor.Site:My.NodeType].properties`

Obviously we need to select the separate well! The brackets look nice but _might_ collide with something ofc.

albe:
> @albe I have adapted the regular expression so that your example can be escaped. This makes it a breaking change, I think.
> I am not a regex expert, if someone has a better solution I would be happy.

Was just fiddling around with the regex myself and was now stuck at `'#(?<!\\\\)((?:\\\\\\\\)*)\\.#'`, which works for any uneven amount of escape characters but needs to use `PREG_SPLIT_NO_EMPTY|PREG_SPLIT_DELIM_CAPTURE` to keep the even number of backslashes in the split. You'd then need to `array_reduce` array elements that only consist of backslashes into the previous. Ugh... getting wild quickly. Your solution only works for the one special case, but I guess that's good enough?

albe (commented):
Two small optimizations

gerdemann:
As alternative here the version with the brackets:
https://github.com/neos/flow-development-collection/compare/master...gerdemann:getvaluebypath-with-escaped-dot-2

kitsunet:
The biggest gripe I have with both variants right now is that it turns a super simple and fast function that is used all over the place into something substantially more complex. And while the increase in processing time might be minimal the amount it gets used will probably make it tangible.

kitsunet:
The path given to utility can already be an array (probably should be split into two functions but separate problem) but I guess I woudl rather like to see the command doing the work and we accept a new argument "pathSegments" or something that is then an array of keys

gerdemann:
Yes, performance is an argument. I would have, under certain circumstances, the functionality not only in the command. What do you think about a second method "getValueByPathString" or something in the Arrays utility?

albe:
> The biggest gripe I have with both variants right now is that it turns a super simple and fast function that is used all over the place into something substantially more complex.

Yep, that's also why I even slightly prefer the "imperfect" solution without the array_reduce.

Guess that's something we can't easily debate or refactor away. Moving the splitting outside the function will strip it of it's core functionality, plus you have to handle all those places where the function is used. I agree though, that the method should be split into two - one that takes a `pathSegments` array and one that takes a path string and calls the former.

So it comes down to a general decision/trade-off of "allowing dots in array keys" vs. code complexity+performance

kitsunet:
IDK, for me the nodetypes are a bit special, do we have any other place where we need this functionality?
it wa sbuild for simple paths like settings and that's what it's great for.

gerdemann:
Then it shouldn't be in a general arrays utility, should it? I use it often with other arrays as well.

kitsunet:
I mean, it fullfills a specific task and that it does fast, but yeah we should certainly talk about that.

albe:
So, how do we proceed? @kitsunet

albe:
Not quite sure what happened to this PR, but it somehow got rebased (?) and a whole lot of other commits pulled in :/ Would like to go over this again tomorrow and see if we can make it mergeable then @kitsunet or rather postpone to 6.3

avency-dev:
@albe I only see one commit, is there anything I can do to help?

gerdemann:
@albe I only see one commit, is there anything I can do to help?

albe:
D'oh - has this resolved itself? :D Thanks!

Neos.Utility.Arrays/Classes/Arrays.php
```diff
@@ -194,7 +194,31 @@ public static function array_reduce(array $array, string $function, $initial = n
     public static function getValueByPath(array &$array, $path)
     {
         if (is_string($path)) {
-            $path = explode('.', $path);
+            if (strpos($path, '\\.') === -1) {
```
bwaidelich, line 197: How can this return -1? According to https://www.php.net/manual/en/function.strpos.php this "Returns FALSE if the needle was not found."

albe: `strpos ( string $haystack , mixed $needle [, int $offset = 0 ] ) : **int**` huh? PHP 🤷‍♂ But you are right.
```suggestion
            if (strpos($path, '\\.') === false) {
```

bwaidelich: Yeah.. One of the reasons for https://wiki.php.net/rfc/union_types_v2#false_pseudo-type

albe:
Sorry this couldn't land in 6.2 yet - let's get this into major soon so it lands in 6.3!

kdambekalns:
Everyone - if we get this sorted out before monday… it could maybe land in 6.3. I skimmed the PR comments, but would love to get a summary of the status from @albe or @kitsunet, as you seem to have thought about this quite a lot. Thanks!

albe:
Sorry I couldn't look into this again yet. I think we still have a decision to make here.

Here's the latest summary of discussion:

> > The biggest gripe I have with both variants right now is that it turns a super simple and fast function that is used all over the place into something substantially more complex.
>
> Yep, that's also why I even slightly prefer the "imperfect" solution without the array_reduce.
>
> Guess that's something we can't easily debate or refactor away. Moving the splitting outside the function will strip it of it's core functionality, plus you have to handle all those places where the function is used. I agree though, that the method should be split into two - one that takes a pathSegments array and one that takes a path string and calls the former.
>
> So it comes down to a general decision/trade-off of "allowing dots in array keys" vs. code complexity+performance 

> IDK, for me the nodetypes are a bit special, do we have any other place where we need this functionality?
it wa sbuild for simple paths like settings and that's what it's great for.

> Then it shouldn't be in a general arrays utility, should it? I use it often with other arrays as well.

So, tl;dr: Should we split this into two methods, in order to keep the simplicity and performance of the original and if so, where to put the new method/move the old?

mhsdesign:
I was once also trying to access an array with dots in paths via this method and i assumed this would work:
`$path = 'something."this.is.one.thing".more.stuff'` but it didnt work (because explode...)
Escaping the dots is something i dindt think of...

btw i think this regex might look better: `(?:\\.)(*SKIP)(*FAIL)|\.`
The above regex can also be modified to select all dots except in quotes as shown here: https://github.com/neos/neos-development-collection/blob/be26b253c8bce69695047fa7799e6c2d4f0cf02f/Neos.Fusion/Classes/Core/Parser.php#L102

albe:
> How about to use stripslashes to unescape the matches? https://3v4l.org/E7mRP#v7.4.28

👍 

> id like this string quoting syntax better: `foo.bar."escaped.dots.likeinfusionpaths".baz`

The drawback with that would be that you'd need to depend on the escape character of the surrounding syntax if this path is within quotes, e.g. `path="foo.bar.\"escaped.dots.likeinfusionpaths\".baz"`

I think backslash as escape character is well known in PHP, so I do think it's a valid thing. I'm open to good arguments for other escape methods.

kdambekalns (commented):
I changed the base to `master`… and AFAICS the regex was about to be simplified to the solution suggested by @mhsdesign? And @albe wanted to do that?

kitsunet:
I still feel this might result in a possibly drastic performance change, just due to the sheer amount of usage this method gets and it now doing a lot of *useless*  things for an edge case. I would rather implement an alternative way for specific use cases even if that means NodeTypes won't work everywhere/with everything.

kdambekalns:
> I still feel this might result in a possibly drastic performance change

So a benchmark for the added `strpos` call would clear this up?

kitsunet:
> > I still feel this might result in a possibly drastic performance change
> 
> So a benchmark for the added `strpos` call would clear this up?

See I tend to overlook that part :D I guess having 50k calls to strpos should be fine...?

mhsdesign:
even though we have the `strpos` inplace, I don't particularly like overloading the code for an edgecase like that....
this function is mainly used for cli arguments, and I think we should just have another function for it.
(I agree that we need to support this functionality somehow, as it is no fun to write such regex by yourself ( only for me 😂))

kdambekalns:
https://3v4l.org/Y37ZN#v8.1.6

mhsdesign:
just tested it, the regex, that i suggested takes less steps and is bullet proof too.

> `(?:\\.)(*SKIP)(*FAIL)|\.`

vs: `(?<!\\)((?:\\\\)*)\.`

mhsdesign:
As i see it its not important where we implement it: may it be directly in `Arrays::getValueByPath` or some other place.

Sure it would be nice if this function can handle it as feature in the future but much more important - at least as i find - is that we have a way to use `./flow configuration:show` with dots inside paths as bugfix. For that case i would be fine with inlining the regex and refactor it later.

For now you can add the above into the `Neos\Flow\Command\ConfigurationCommandController->showCommand`

`Packages/Framework/Neos.Flow/Classes/Command/ConfigurationCommandController.php:75`

```diff
- $configuration = Arrays::getValueByPath($configuration, $path);
+ $configuration = Arrays::getValueByPath(
+     $configuration,
+     array_map('stripslashes', preg_split('#(?:\\\.)(*SKIP)(*FAIL)|\\.#', $path))
+ );
```

use it like:

`./flow configuration:show --type Policy --path "roles.Neos\.Flow:Everybody"`

(ps. dont miss the qoutes around the path, otherwise you need to escape the backslash)

bwaidelich:
> much more important [...] is that we have a way to use ./flow configuration:show with dots inside paths as bugfix

I think, that's a great compromise for an otherwise rather profound change!

mhsdesign:
I will prepare a bugfix-pr for that

kitsunet:
Did the bugfix ever happen, as it's not referenced in here?

kitsunet (approved):
By now I am less sceptical about this, so if we still want, lets do it.

mhsdesign:
> Did the bugfix ever happen, as it's not referenced in here?

Haha caught, no I never did that xD … yet …

mhsdesign:
@kitsunet there was lately a discussion in slack about something similar and a lot of pain involved with escaping things. Since escaping multiple context is always hard to grasp (like when using ddev -> shell -> flow or worse when building up escaped commands in yaml) … I just want to make sure we make the right choice!

After thinking about it i think i like your idea better: https://github.com/neos/flow-development-collection/pull/1903#issuecomment-580658435

mhsdesign:
i think this is so much better 
```
ddev flow configuration:show --type NodeTypes --path [Vendor.Site:My.NodeType].properties
```

than
```
ddev flow configuration:show --type NodeTypes --path '"Vendor\.Site:My.NodeType.properties"'
# or
ddev flow configuration:show --type NodeTypes --path 'Neos\\.Neos:Document'
```
(note due to ddev we need another escaping layer in between, and its common to have to double escape stuff and hard to wrap the head around.)
Of course, when using the `[` as a special char we need to allow escaping it, which will make things more complex but as this is an edgecase (i assume) i think its fine.

Btw its a breaking change ... technically ...

mhsdesign (changes requested):
https://github.com/neos/flow-development-collection/pull/1903#issuecomment-1427434230

aaand target 8.3

mhsdesign:
```php
<?php
function parseInputString(string $input): array
{
    $result = [];
    $current = '';
    $len = \strlen($input);

    for ($i = 0; $i < $len; $i++) {
        $char = $input[$i];

        switch ($char) {
            case "[":
                if ($current !== "") {
                    $result[] = $current;
                    $current = "";
                }
                for ($i++; $i < $len; $i++) {
                    $char = $input[$i];
                    switch ($char) {
                        case "[":
                            throw new \Exception(sprintf('Unexpected char "%s" at position "%s" in "%s"', $char, $i, $input));
                        case "]":
                            if ($current !== "") {
                                $result[] = $current;
                                $current = "";
                            }
                            $nextChar = $input[$i + 1] ?? "";
                            if ($nextChar !== "" && $nextChar !== ".") {
                                throw new Exception('Missing dot after closing');
                            }

                            break 2;
                        default:
                            $current .= $char;
                    }
                }
                break;
            case "]":
                throw new \Exception(sprintf('Unexpected char "%s" at position "%s" in "%s"', $char, $i, $input));
            case ".":
                if ($current !== "") {
                    $result[] = $current;
                    $current = "";
                }
                break;
            default:
                $current .= $char;
        }
    }

    if ($current !== "") {
        $result[] = $current;
    }

    return $result;
}

$input = 'fooo.bar[dots.are.are.lowed].something.else';
$output = parseInputString($input);
print_r($output);
```

i will implement this in the config:show command as bugfix ;)

```
fooo.bar[dots.are.are.lowed].something.else

Array
(
    [0] => fooo
    [1] => bar
    [2] => dots.are.are.lowed
    [3] => something
    [4] => else
)
```

mhsdesign:
https://github.com/neos/flow-development-collection/pull/2973

grüße aus dreden

mhsdesign:
it was discussed on the dd sprint 23 https://github.com/neos/flow-development-collection/pull/2973#issuecomment-1461956084 that we want to use json for complex cli arguments (including dot escaping) and use Array.getValueByPath on php site better with actual arrays instead of paths

PR #1904, opened 2020-01-31
BUGFIX: Allow passing syslog() priorities constants to Logger.log()

fnkr:
**What I did**

`Logger.log()` should accept [syslog() priority constants](https://www.php.net/manual/en/function.syslog.php) (e.g. `LOG_INFO`) instead of strings (e.g. `'info'`) for the `$level` function parameter.

**How to verify it**

```php
$logger->log('Hello, world.', LOG_INFO);
$logger->log('Hello, world.', 'info');
```

**Checklist**

- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html) => Flow 4.x does not use PSR-style loggers and is therefore not affected.

### DIFF
--- a/Neos.Flow.Log/Classes/Psr/Logger.php
+++ b/Neos.Flow.Log/Classes/Psr/Logger.php
@@ -66,13 +66,15 @@ class Logger implements LoggerInterface
      */
     public function log($level, $message, array $context = [])
     {
-        $backendLogLevel = self::LOGLEVEL_MAPPING[$level];
+        if (is_string($level) === true) {
+            $level = self::LOGLEVEL_MAPPING[$level];
+        }
 
         list($packageKey, $className, $methodName) = $this->extractLegacyDataFromContext($context);
         $additionalData = $this->removeLegacyDataFromContext($context);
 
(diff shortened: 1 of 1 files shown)


kitsunet:
https://www.php-fig.org/psr/psr-3/

if anything you should usue the prefered `$logger-> {level}` methods OR the PSR-3 constants, which already work because they are exactly mapped to the magic strings.

fnkr:
@kitsunet I'm aware that the `$logger->{$level}` methods exist. We're processing events and the PHP-compatible log level (integer) is part of the data. We're doing it like this for now:

```php
$severity = array_search($severity, Logger::LOGLEVEL_MAPPING);
$logger->log($severity, $message);
```

kitsunet:
Then that's what you must do, adding above functionality to the logger would make it work different than other PSR loggers which means you can write code that would break when you switch to a different PSR-3 logger which would break the whole point of coding against the PSR LoggerInterface

fnkr:
Sounds reasonable, I'm closing this.

PR #1906, opened 2020-01-31
FEATURE: Allow validation of dynamic argument types

### LINKED ISSUE #1905: Runtime action argument type information should override static type information
Currently, the action method arguments are initialized early inside the controller, before initializing the property mapping and calling `initialize*Action()`. That way the argument only has type information from static reflection. The validation chain is then built based on that argument type information (`$argument->getDataType()`) and in case of an action receiving an interface with dynamic type overrides (`__type` or inside `initialize*Action()`), the validation chain is not correctly built, missing validation information from the actual class property annotations.

The issue in solving this however, is that we need to check runtime type information only after `initialize*Action()` / propertyMappingConfiguration (the latter needs to explicitly allow overriding the type). That in turn expects the action arguments to be already available.

So there are three possible ways to solve this (assuming allowing dynamic typing of action arguments is valid):
- action argument type information and validators are not (fully) available inside `initialize*Action()`, moving `initializeActionMethodValidators()` **after** that
- move `__type` override from the `ObjectConverter` lower down into the Framework stack, so that the argument type is overriden earlier (before property mapping)
- completely rebuilding the valdation chain building to be lazy and only work from actual data types given into the validator

The latter is at least an order of magnitude more complex, but generally desirable - the first is potentially a bit breaky.

albe:
This allows correct validation of action method arguments that are not
statically typed, e.g. have a type hint of an interface.
The type override via `__type` or the property mapping configuration will
now take effect for building the correct validator chain, if the controller has
the property `$enableDynamicTypeValidation` set to `true`.

If you enable this flag, note that the point in time when the validation
chain is built is moved until **after** calling the `initialize*Action` method.
This means that validator information is not available there any more.

Resolves #1905


Neos.Flow/Classes/Mvc/Controller/Argument.php
```diff
@@ -180,6 +181,10 @@ public function getDefaultValue()
     public function setValidator(ValidatorInterface $validator): Argument
     {
         $this->validator = $validator;
+        if ($validator !== null) {
+            $validationMessages = $validator->validate($this->value);
```
albe, line 185:
```suggestion
            $this->validationResults = $this->propertyMapper->getMessages();
            $validationMessages = $validator->validate($this->value);
```

albe:
So unit tests finally pass, but this one functional test 🤔 ...
```
There was 1 failure:
1) Neos\Flow\Tests\Functional\Mvc\ActionControllerTest::optionalObjectArgumentsCanBeOmitted
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'null'
+'Validation failed while trying to call Neos\Flow\Tests\Functional\Mvc\Fixtures\Controller\ActionControllerTestBController->optionalObjectAction().\n
+'
/home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Flow/Tests/Functional/Mvc/ActionControllerTest.php:261
/home/travis/build/neos/flow-development-distribution/Packages/Libraries/phpunit/phpunit/phpunit:61
```

Neos.Flow/Classes/Mvc/Controller/Argument.php
```diff
@@ -212,12 +218,15 @@ public function setValue($rawValue): Argument
             $this->value = $rawValue;
             return $this;
```
albe, line 219: Am I correct that until now a `setValue` without property mapping (because value was already an object) would skip the validation? Was this intended? Why?

Neos.Flow/Classes/Property/TypeConverter/ObjectConverter.php
```diff
@@ -40,14 +40,14 @@
 class ObjectConverter extends AbstractTypeConverter
 {
     /**
-     * @var integer
+     * @var string
      */
-    const CONFIGURATION_TARGET_TYPE = 3;
+    const CONFIGURATION_TARGET_TYPE = 'targetType';
```
albe, line 45: Why this unrelated change? Because Psalm was complaining about the new usage of `getConfigurationValue(..., ::CONFIGURATION_...)` from above and all other converter options already are strings. Thus this type mismatch is fixed, we can reduce psalm baseline and we can theoretically even add an argument typehint to the method.

albe:
Not sure what the StyleCI issues are about all the time :/

albe:
@bwaidelich is this something that could potentially still be considered before branch-off (if I add the missing test)? The issue is still on the board. Otherwise, I'll postpone writing the test until next major where this could be possibly be merged

bwaidelich:
> @bwaidelich is this something that could potentially still be considered before branch-off 

I would rather not to be honest.. I can't really judge but it seems like an invasive change with the potential to break things. And it is missing tests.
And we currently struggle with a lot of other building sites (e.g. stabilizing middleware change #2258, documentation, release notes, ...).
@albe what is your estimation? is this change worth taking the risk or can this be worked around with a custom controller anyways and we postpone it for now?

albe:
As discussed, this has low prio over the other things still open. A suggestion to make this less breaking by @kitsunet was to introduce a boolean property in the controller to switch this behaviour on. Then we could postpone the breakiness to 8.0. I will try to do this, but with that implementation I guess it could even land in 7.1 if it's not gonna make it.

Neos.Flow/Classes/Mvc/Controller/Argument.php
```diff
@@ -212,12 +218,15 @@ public function setValue($rawValue): Argument
...
+            $this->dataType = $configuredType;
+        } elseif (isset($rawValue['__type']) && $configuration->getConfigurationValue(ObjectConverter::class, ObjectConverter::CONFIGURATION_OVERRIDE_TARGET_TYPE_ALLOWED) === true) {
+            $this->dataType = $rawValue['__type'];
+        }
         $this->value = $this->propertyMapper->convert($rawValue, $this->dataType, $this->getPropertyMappingConfiguration());
         $this->validationResults = $this->propertyMapper->getMessages();
-        if ($this->validator !== null) {
-            $validationMessages = $this->validator->validate($this->value);
-            $this->validationResults->merge($validationMessages);
-        }
```
albe, line 220: Maybe we can/should keep those lines. The validator won't be set with the moved `initializeActionMethodValidators()` method in the normal case, so validation will not be executed twice.
However, in case some controller overwrites an argument value for whatever reason, the validation will still be run with that new value, so that's b/c to existing behaviour.

Neos.Flow/Classes/Mvc/Controller/Argument.php
```diff
@@ -180,6 +181,11 @@ public function getDefaultValue()
     public function setValidator(ValidatorInterface $validator): Argument
     {
         $this->validator = $validator;
+        if ($validator !== null && $this->value !== null) {
```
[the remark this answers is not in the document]
albe: Same as with https://github.com/neos/flow-development-collection/pull/1906/files#diff-03d5ab75f1321a95bb4043fbb9ca202d25246c9db5c354664f45c23d4a24bde6L217 - we can just have those lines independent of a feature flag setting in the controller, because it will just work as before depending on the order of the `initializeActionMethodValidators()` method call relative to `mapRequestArgumentsToControllerArguments()`. As a bonus (FEATURE), it makes sure that if a controller manually overwrites the validator of an argument, it will be executed.

Neos.Flow/Classes/Property/TypeConverter/ObjectConverter.php
albe: @bwaidelich You suggested we change this back to make this less breaking. While you're right about reducing breakiness, I do think having those converter options `int` while others are `string` is in fact bad code and should be fixed (though we could move this to another PR). `get/setConfigurationValue(...)` should only accept `string` for the second argument (=config option name)

albe:
This is now no longer breaking (unless we really do consider the `const` change breaking and keep this in).

@bwaidelich @kitsunet looking good from my side now (will still add a test for the specific use case this covers later today or tomorrow)

albe:
There you go. Tests nicely cover the scenario and fail if the `$enableDynamicTypeValidation` property is set to `false` (or not set).

Neos.Flow/Classes/Property/PropertyMapper.php
```diff
@@ -138,7 +138,7 @@ public function convert($source, $targetType, PropertyMappingConfigurationInterf
      */
     public function getMessages()
     {
-        return $this->messages;
+        return $this->messages ?? new Result();
```
bwaidelich, line 141: This could break code that checked for `$propertyMapper->getMessages() === null`

bwaidelich: Instead we could adjust 
```php
$this->validationResults = $this->propertyMapper->getMessages() ?? new Result();
```
in Argument::setValue()

bwaidelich:
As discussed: IMO we should keep the PR as small as possible (i.e. extract changes to PropertyMapper and TypeConverters to separate PRs if they are not necessary for this one to work)

bwaidelich:
As we'd like to branch off 7.0 today I would suggest to merge this into `master` **afterwards** (in it's current version) and to concentrate on #2258 for the release.
I still fear that the change might introduce some side effects and this way we could introduce it for 7.1.
Folks that really require this feature now can implement it in a custom controller already

Neos.Flow/Classes/Mvc/Controller/Argument.php
```diff
@@ -180,6 +181,11 @@ public function getDefaultValue()
...
+            $this->validationResults = $this->propertyMapper->getMessages();
```
albe, line 185:
```suggestion
            $this->validationResults = $this->propertyMapper->getMessages() ?? new Result();
```

Neos.Flow/Classes/Mvc/Controller/Argument.php
```diff
@@ -212,6 +218,13 @@ public function setValue($rawValue): Argument
...
         }
+        $configuration = $this->getPropertyMappingConfiguration();
+        $configuredType = $configuration->getConfigurationValue(ObjectConverter::class, ObjectConverter::CONFIGURATION_TARGET_TYPE);
+        if ($configuredType !== null) {
+            $this->dataType = $configuredType;
+        } elseif (isset($rawValue['__type']) && $configuration->getConfigurationValue(ObjectConverter::class, ObjectConverter::CONFIGURATION_OVERRIDE_TARGET_TYPE_ALLOWED) === true) {
+            $this->dataType = $rawValue['__type'];
+        }
         $this->value = $this->propertyMapper->convert($rawValue, $this->dataType, $this->getPropertyMappingConfiguration());
         $this->validationResults = $this->propertyMapper->getMessages();
```
albe, line 229:
```suggestion
        $this->validationResults = $this->propertyMapper->getMessages() ?? new Result();
```

Neos.Flow/Classes/Property/PropertyMapper.php
albe, line 141:
```suggestion
        return $this->messages;
```

albe: Done

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -149,6 +149,14 @@ class ActionController extends AbstractController
      */
     protected $logger;
 
+    /**
+     * Feature flag to enable the potentially breaking support of validation for dynamic types specified with `__type` argument or in the `PropertyMapperConfiguration`.
+     * Note: This will be enabled by default in a future version.
+     * See https://github.com/neos/flow-development-collection/pull/1906
```
bwaidelich, line 155: IMO we should link to the _issue_ instead (https://github.com/neos/flow-development-collection/issues/1905) since there might be multiple PRs attached to it (at some point). Also we can collect more generic infos in the issue

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -190,7 +198,9 @@ public function processRequest(ActionRequest $request, ActionResponse $response)
         $this->actionMethodName = $this->resolveActionMethodName();
 
         $this->initializeActionMethodArguments();
-        $this->initializeActionMethodValidators();
+        if ($this->enableDynamicTypeValidation !== true) {
```
bwaidelich, line 201: nitpick: any specific reason why you didn't go for
```suggestion
        if (!$this->enableDynamicTypeValidation) {
```

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -200,6 +210,9 @@ public function processRequest(ActionRequest $request, ActionResponse $response)
         $this->mvcPropertyMappingConfigurationService->initializePropertyMappingConfigurationFromRequest($this->request, $this->arguments);
 
         $this->mapRequestArgumentsToControllerArguments();
+        if ($this->enableDynamicTypeValidation === true) {
```
bwaidelich, line 213: and
```suggestion
        if ($this->enableDynamicTypeValidation) {
```
?

(I can totally live with both ofc)

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -361,6 +374,7 @@ protected function initializeActionMethodValidators()
             $ignoredArguments = [];
         }
 
+        /* @var $argument Argument */
```
bwaidelich, line 377: Instead we could annotate the `arguments` field with `Argument[]` to expose the type everywhere?!

Neos.Flow/Classes/Mvc/Controller/Argument.php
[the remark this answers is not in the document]
bwaidelich: Maybe this could be condensed into a short inline comment!?

Neos.Flow/Classes/Property/TypeConverter/ObjectConverter.php
bwaidelich: I still think that it should at least go into a separate PR. I could imagine that folks use this in forms

Neos.Flow/Classes/Mvc/Controller/ActionController.php
albe: Nope, just accustomed to strict boolean checks over the falsy operator.

Neos.Flow/Classes/Mvc/Controller/ActionController.php
albe: True. Needs to be adjusted in AbstractController

Neos.Flow/Classes/Mvc/Controller/Argument.php
albe, line 184:
```suggestion
        // the validation should not be called on null values - for the cases where the value is required, the error will be thrown in Controller::mapRequestArgumentsToControllerArguments()
        if ($validator !== null && $this->value !== null) {
```
Does that sound clear enough or too elaborate @bwaidelich ? I'm pretty bad at condensing information...

albe (commented):
Thanks for the review!

Neos.Flow/Classes/Mvc/Controller/Arguments.php
```diff
@@ -18,6 +18,7 @@
  * A composite of controller arguments
  *
  * @api
+ * @extends \ArrayObject<Argument>
```
albe, line 21: Doesn't change the typehinting in IDE (yet), but psalm should be happy about this.

albe:
Reverted the unrelated change for the configuration consts

Neos.Flow/Classes/Mvc/Controller/Argument.php
kdambekalns: 🤷‍♂️

PR #1909, opened 2020-02-14
TASK: Update psalm-baseline.xml

kdambekalns:
The new errors look like false positives, at least so much I wouldn't
risk a change for them…

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="3.8.2@90d6b73fd8062432030ef39b7b6694b3902daa31">
+<files psalm-version="3.8.5@e6ec5fa22a7b9e61670a24d07b3119aff80dcd89">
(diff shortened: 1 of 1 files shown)


kdambekalns:
How can psalm fail with the very errors this should fix? 🤔

daniellienert:
@kdambekalns, looks like if you updated the baseline with the Neos Dev-Distribution installed.

kdambekalns:
> looks like if you updated the baseline with the Neos Dev-Distribution installed.

Yes, but that's what I always did so far. 🤔

kdambekalns:
So, the trick is… to use PHP 7.2 as the test does on Travis. When updating the baseline with PHP 7.4, the build fails like this.

albe:
> So, the trick is… to use PHP 7.2 as the test does on Travis. When updating the baseline with PHP 7.4, the build fails like this.

D'oh :| Another good reason that we somehow find a solution to "fix" psalm through the pipeline. Maybe another look at GitHub actions could work. Or a job on the jenkins that can be manually triggered?

PR #1912, opened 2020-02-19
BUGFIX: Avoid exception in findIdentifiersByTag()

kdambekalns:
When `substr($metaData, 0, -self::EXPIRYTIME_LENGTH)`on
`$metadata` in the cache `FileBackend.findIdentifiersByTag()`
method fails, `false` is passed to `explode()` causing:

`Exception in line 253 of /…/FileBackend.php: explode() expects
parameter 2 to be string, boolean given`

This fixes that by casting the result to a string.

### DIFF
--- a/Neos.Cache/Classes/Backend/FileBackend.php
+++ b/Neos.Cache/Classes/Backend/FileBackend.php
@@ -250,11 +250,9 @@ class FileBackend extends SimpleFileBackend implements PhpCapableBackendInterfac
             if ($expiryTime !== 0 && $expiryTime < $now) {
                 continue;
             }
-           $extractedTags = substr($metaData, 0, -self::EXPIRYTIME_LENGTH);
-           if ($extractedTags === false) {
-               continue;
-           }
-            if (!in_array($searchedTag, explode(' ', $extractedTags))) {
+
+            $extractedTags = substr($metaData, 0, -self::EXPIRYTIME_LENGTH);
+            if ($extractedTags === false || !in_array($searchedTag, explode(' ', $extractedTags))) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


kitsunet:
Shouldn't we rather extract that before the in_array check and add a separate condition if that was false? Seems sketchy and I am not sure we remember in a year why we would cast that to string.

Neos.Cache/Classes/Backend/FileBackend.php
```diff
@@ -250,7 +250,7 @@ public function findIdentifiersByTag(string $searchedTag): array
             if ($expiryTime !== 0 && $expiryTime < $now) {
                 continue;
             }
-            if (!in_array($searchedTag, explode(' ', substr($metaData, 0, -self::EXPIRYTIME_LENGTH)))) {
+            if (!in_array($searchedTag, explode(' ', (string)substr($metaData, 0, -self::EXPIRYTIME_LENGTH)))) {
```
kitsunet:
```suggestion
           $extractedTags = substr($metaData, 0, -self::EXPIRYTIME_LENGTH);
           if ($extractedTags === false) {
               continue;
           }
            if (!in_array($searchedTag, explode(' ', $extractedTags))) {
```

kdambekalns:
There is probably a better way to fix this, yes. Anything extracted from the read file before that block could probably go wrong in similar ways. See this as a sketch for a solution, intended to be fix and issue in one… ;)

Neos.Cache/Classes/Backend/FileBackend.php
```diff
@@ -250,7 +250,11 @@ public function findIdentifiersByTag(string $searchedTag): array
             if ($expiryTime !== 0 && $expiryTime < $now) {
                 continue;
             }
-            if (!in_array($searchedTag, explode(' ', (string)substr($metaData, 0, -self::EXPIRYTIME_LENGTH)))) {
+           $extractedTags = substr($metaData, 0, -self::EXPIRYTIME_LENGTH);
+           if ($extractedTags === false) {
```
kdambekalns:
```suggestion
           if ($extractedTags === false || !in_array($searchedTag, explode(' ', $extractedTags))) {
```
would be my way of doing it. But that's just style…

kitsunet: Good point! Lets do that. I _think_ though if I commit your suggestion it  contains only that line and not the  extractedTags

albe:
I do think, the `$extractedTags` check is now even unnecessary (in theory), since metaData can only be a string now and unless the file does not contain metadata (but a correct metadata offset = `$index` = 0 - or just the offset read fails). Still good measure to keep both checks.

PR #1913, opened 2020-02-21
BUGFIX: Allow numeric arrays in request body

albe:
When sending e.g. a JSON request with an array as body (`["foo","bar"]`), Flow would throw an exception because it could not map the numeric keys to valid argument names.

This change fixes that by just not merging those bodies into the controller arguments. If you want to access such a request body, you can now use `$this->request->getHttpRequest()->getParsedBody()` from inside your controller.
Associative arrays with valid string keys will still be automatically mapped.

Related to https://github.com/neos/flow-development-collection/pull/1606, which states that mapping an array is possible.

### DIFF
--- a/Neos.Flow/Classes/Mvc/DispatchComponent.php
+++ b/Neos.Flow/Classes/Mvc/DispatchComponent.php
@@ -99,7 +99,9 @@ class DispatchComponent implements ComponentInterface
 
         $parsedBody = $this->parseRequestBody($httpRequest);
         if ($parsedBody !== []) {
-            $arguments = Arrays::arrayMergeRecursiveOverrule($arguments, $parsedBody);
+            if (Arrays::array_all(array_keys($parsedBody), 'is_string')) {
+                $arguments = Arrays::arrayMergeRecursiveOverrule($arguments, $parsedBody);
+            }
             $httpRequest = $httpRequest->withParsedBody($parsedBody);
         }
 
--- a/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/ModelViewController.rst
+++ b/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/ModelViewController.rst
@@ -311,6 +311,14 @@ later sources replace earlier ones
 * body (typically from POST or PUT requests)
 * file uploads (derived from $_FILES)
 
+.. note::
+
+  If you submit a request body that will be parsed into a PHP array, e.g. a JSON
+  array or object, and it contains numeric keys, Flow will not automatically map
+  this into controller arguments, as numbers are no valid PHP variable names.
+  You can access the full request body then via `$this->request->getHttpRequest()->getParsedBody()`
+  inside your controller.
+
 Internal Arguments
 ~~~~~~~~~~~~~~~~~~
 
--- a/Neos.Flow/Tests/Functional/Mvc/ActionControllerTest.php
+++ b/Neos.Flow/Tests/Functional/Mvc/ActionControllerTest.php
@@ -126,6 +126,21 @@ class ActionControllerTest extends FunctionalTestCase
     }
 
     /**
+     * @test
+     */
+    public function arrayRequestBodyIsNotMappedByDefault()
+    {
+        $body = json_encode(['foo']);
+        $request = Request::create(new Uri('http://localhost/test/mvc/actioncontrollertesta/postArrayBody?getArgument=getValue'), 'POST');
+        $request->setContent($body);
+        $request->setHeader('Content-Type', 'application/json');
+        $request->setHeader('Content-Length', count_chars($body));
+
+        $response = $this->browser->sendRequest($request);
+        $this->assertEquals('postArrayBodyAction-getValue-'.$body, $response->getContent());
+    }
+
+    /**
      * RFC 2616 / 10.4.5 (404 Not Found)
      *
      * @test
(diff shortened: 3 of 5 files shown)


kitsunet:
the problem I have with this is, that it's rather arbitrary. I mean given that it's just broken without this, I guess we should do it. But then definitely remove this "feature" completely for next major.

kitsunet:
the problem I have with this is, that it's rather arbitrary. I mean given that it's just broken without this, I guess we should do it. But then definitely remove this "feature" completely for next major.

DavidSporer:
I think it’s fine that way as long as it’s properly documented. 
As @albe mentioned on Slack it’s pretty convenient that Flow maps arguments but it needs a way to support that as well out of the box.

albe:
> the problem I have with this is, that it's rather arbitrary.

I don't think it is too arbitrary at all - we automatically map all request arguments into nameable controller arguments. This is a very convenient feature of Flow since ever and I think removing automatic mapping of request body would be a (breaking) loss. However, numbers are not valid variable names in PHP, so those obviously can not be mapped. Hence it makes sense that Flow just doesn't do it's automagic on those, but instead lets the user explicitly access or map those arguments.

The only "arbitrary" part I see is, if we should not only skip the numeric indexes in a mixed body, e.g. `{ "0": "foo", "bar": "baz" }`. This would as is completely be unmapped, i.e. you can't automatically have a `$bar` controller argument - and I think this makes sense - all or nothing, because we need to assume the whole body is a self contained thing.

Under any circumstance, this is more an edge-case, which is currently not covered and a mixed key body is rather an abomination.

kitsunet:
> all or nothing, because we need to assume the whole body is a self contained thing.
Agreed, BUT a) I feel that alone makes it kind of arbitrary, sure you can explain the rule(s) but it seems to me the feature has a fatal flaw, that you elude to in your comment yourself...

> self contained thing
This is my argument since ever about this feature basically. I think if you get a JSON object that is a self contained thing, it should ALWAYS end up in ONE argument not be split into multiple separate arguments ever, even if it's convenient for some use cases. Eg. I would assume if I get a JSON object describing an event that I can get an event (just like your annotation does) object but never that I can get a eventName, startDate and endDate. That always felt weird to me.

albe:
> I think if you get a JSON object that is a self contained thing, it should ALWAYS end up in ONE argument not be split into multiple separate arguments ever

That's a fair point - though sometimes the "JSON object" is just abused as a DTO container for multiple values. So, maybe we should not map by default any more, but make it opt-in? Also, what about other formats, like form body or some XML? Those are not always a self-contained object either.

kitsunet:
> but make it opt-in? 

That's what I would prefer yes!

> Also, what about other formats, like form body or some XML? Those are not always a self-contained object either.
Yep totally TBD here, for form body that is essentially just POST isn't it? And XML I really have no clue, I am open for suggestions but would probably say we just make it opt in as well for sakes of consisteency.

albe:
Let's move the discussion for the further movement to https://github.com/neos/flow-development-collection/issues/1914 and keep this just about the "bugfix" for current/old behaviour.

Neos.Flow/Classes/Mvc/DispatchComponent.php
```diff
@@ -99,7 +99,9 @@ protected function prepareActionRequest(ComponentContext $componentContext)
 
         $parsedBody = $this->parseRequestBody($httpRequest);
         if ($parsedBody !== []) {
-            $arguments = Arrays::arrayMergeRecursiveOverrule($arguments, $parsedBody);
+            if (Arrays::array_all(array_keys($parsedBody), 'is_string')) {
```
albe: Maybe instead of `is_string` we should even check for actually valid PHP identifiers. Would require a custom predicate function.

kitsunet: Actually this has several problems anyway that I just realise while thinking it through:

$parsedBody !== [] is dangerous already because technically we could have media type json and a body like `"foo"` or even just `14356` which is both valid json, so that will be problematic. Additionally the type converter returns empty array as fallback which seems wrong as well...

IMHO this whole conversion should be rethought.


For JSON the right way to find out what we want to map and what we do not want to map is doing `json_decode($body) instanceof \stdClass` because that is the only case in which we would want to actually map the decoded (then with true as second argument) body to request arguments _maybe_ because I still find that pretty weird and magic.

albe: Interesting idea, will think about it

albe: Okay, so given following:
https://3v4l.org/I5tE6
The `instanceof stdClass` indeed works for all cases except those that contain numeric (or any other non-identifier) keys. I think the bigger problem is indeed that `parseRequestBody` does not handle scalar values (and `null`) correctly and always returns an (at least empty) array. If it did, the check should be `if (is_array($parsedBody) && $parsedBody !== [])`

So I would suggest the following:
- change `parseRequestBody` to return `mixed` (or rather `string|integer|array|null`)
- change the check here to `is_array($parsedBody) && $parsedBody !== []`
- keep the `Arrays::array_all` check (and possibly improve to an `isValidIdentifier` instead of `is_string`)

albe:
Closing in favor of https://github.com/neos/flow-development-collection/pull/1942 after a closer look together with @kitsunet

PR #1915, opened 2020-02-27
BUGFIX: Resolve SqlLogger when it is a dependency proxy

### LINKED ISSUE #1854: Doctrine SqlLogger does not log any queries
### Description
After updating to Flow 6.0 the [Doctrine SqlLogger](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Persistence/Doctrine/Logging/SqlLogger.php) stopped working. 

The issue is that `$this->logger` is not an instance of `LoggerInterface` instead it is a `DependencyProxy`

```
/**
 * @var LoggerInterface
 */
protected $logger;

public function startQuery($sql, array $params = null, array $types = null)
{
    // this is a safeguard for when no logger might be available...
    if ($this->logger instanceof LoggerInterface) {
        ...
    }
}
```

I think this was caused by [BUGFIX: Use PSR-3 logger in SqlLogger](https://github.com/neos/flow-development-collection/commit/4015b65d1827f650faa48def7bd7fe559b113286) which means it only affects Flow 6.0.x

### Affected Versions
Flow: 6.0.x

albe:
The SqlLogger did not log anything since Flow 6.0 due to a instanceof check, when it was a dependency proxy. This change resolves the dependency.

Resolves #1854

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Logging/SqlLogger.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Logging/SqlLogger.php
@@ -37,7 +37,7 @@ class SqlLogger implements \Doctrine\DBAL\Logging\SQLLogger
      */
     public function startQuery($sql, array $params = null, array $types = null)
     {
-        if ($this->logger instanceof DependendcyProxy) {
+        if ($this->logger instanceof DependencyProxy) {
             $this->logger->_activateDependency();
         }
         // this is a safeguard for when no logger might be available...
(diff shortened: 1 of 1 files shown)


kitsunet:
Wanted to get this fixed quickly but I guess we should in such instances switch over to constructor injection maybe, as hte logger is pretty much needed to make this work?

albe:
I'm all in for changing it to constructor injection.

PR #1917, opened 2020-02-28
BUGFIX: Browser correctly builds uploaded files from form

### LINKED ISSUE #1916: Browser submit does not correctly upload files
Since Flow 6.0 (80c317bd152d97b8171ddd32a0ed6b9e67a615de) the `Browser::request()` method expects an `UploadedFile[]` argument for file uploads. However, the `Browser::submit()` method only provides an (untangled) `$_FILES` like array.

albe:
Fixes #1916

### DIFF
--- a/Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php
+++ b/Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php
@@ -42,7 +42,7 @@ abstract class UploadedFilesHelper
                     $value['name'],
                     $value['type']
                 );
-            } else if ($value instanceof UploadedFileInterface) {
+            } elseif ($value instanceof UploadedFileInterface) {
                 $originallySubmittedResourcePath = array_merge($currentPath, ['originallySubmittedResource']);
                 $collectionNamePath = array_merge($currentPath, ['__collectionName']);
                 $upcastedUploads[$key] = self::upcastUploadedFile(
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php
```diff
@@ -21,17 +23,26 @@
 abstract class UploadedFilesHelper
 {
     /**
-     * @param UploadedFileInterface[] $uploadedFiles
+     * @param UploadedFileInterface[]|mixed[][] $uploadedFiles A (deep) array of UploadedFile or an untangled $_FILES array
      * @param array $arguments
      * @param array $currentPath internal argument for recursion
-     * @return array The nested array of paths and uploaded files
+     * @return UploadedFileInterface[] The nested array of paths and uploaded files
```
albe: That's not technically 100% correct but... it's more specific than just `array`?

albe: psalm is with me:
```
ERROR: InvalidReturnType - Packages/Framework/Neos.Flow/Classes/Http/Helper/UploadedFilesHelper.php:29:16 - The declared return type 'array<array-key, Psr\Http\Message\UploadedFileInterface>' for Neos\Flow\Http\Helper\UploadedFilesHelper::upcastUploadedFiles is incorrect, got 'array<array-key, Neos\Http\Factories\FlowUploadedFile|array<array-key, Psr\Http\Message\UploadedFileInterface>>'
     * @return UploadedFileInterface[] The nested array of paths and uploaded files
```

albe: Not sure how this is done correctly in psalm... https://psalm.dev/r/cc875a4a60
Maybe we need to stick with the generic `array` 🤷‍♂

albe: Well, not possible with psalm as of now (and in the foreseeable future):
https://github.com/vimeo/psalm/issues/2777#issuecomment-583891417

albe:
Um, I think this should have landed in 6.0 🤦‍♂

PR #1920, opened 2020-03-02
BUGFIX: Fix output of request and response in throwable storage

daniellienert:
Fixed the checking for valid request and response objects.
Remove superflous configuration of throwablestorage in `Scripts::initializeExceptionStorage`

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -26,8 +26,6 @@ use Neos\Flow\Core\ProxyClassLoader;
 use Neos\Flow\Error\Debugger;
 use Neos\Flow\Error\ErrorHandler;
 use Neos\Flow\Error\ProductionExceptionHandler;
-use Neos\Flow\Http\Helper\RequestInformationHelper;
-use Neos\Flow\Http\HttpRequestHandlerInterface;
 use Neos\Flow\Log\PsrLoggerFactoryInterface;
 use Neos\Flow\Log\ThrowableStorage\FileStorage;
 use Neos\Flow\Log\ThrowableStorageInterface;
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -263,36 +261,7 @@ protected static function initializeExceptionStorage(Bootstrap $bootstrap): Thro
...
-            $response = $requestHandler->getHttpResponse();
-            // TODO: Sensible error output
-            $output .= PHP_EOL . 'HTTP REQUEST:' . PHP_EOL . ($request ? '[request was empty]' : RequestInformationHelper::renderRequestHeaders($request)) . PHP_EOL;
-            $output .= PHP_EOL . 'HTTP RESPONSE:' . PHP_EOL . ($response ? '[response was empty]' : $response->getStatusCode()) . PHP_EOL;
-            $output .= PHP_EOL . 'PHP PROCESS:' . PHP_EOL . 'Inode: ' . getmyinode() . PHP_EOL . 'PID: ' . getmypid() . PHP_EOL . 'UID: ' . getmyuid() . PHP_EOL . 'GID: ' . getmygid() . PHP_EOL . 'User: ' . get_current_user() . PHP_EOL;
-
-            return $output;
-        });
-
-        return $throwableStorage;
```
albe: 😱 all that is totally unnecessary?

daniellienert: Its exactly the same code as in the constructor

kdambekalns:
Just wanted to express some post-merge ❤️!

kitsunet:
Ehhhh, wait, wait, wait! That duplication had a reason :D The FileStorage is just _our_ implementation of the `\Neos\Flow\Log\ThrowableStorageInterface` which has the `setBacktraceRenderer` method. So Flow _should_ set the renderer on initialising the choosen implementation (which might or might not be the FileStorage), BUT given that there might be exceptions BEFORE the init code in Scripts is run the FileStorage would automatically be used as "fallback" exception storage and for that case it had the duplicated code in the constructor. If you have a better idea to solve those cases, feel free, but htis is not ;)

kitsunet:
Please now vote to revert, thanks! 😆

albe:
Oh my, definitely something we should inline document in that duplication. Let's revert quickly then and tag a new release - haven't announced my releases from yesterday yet.

PR #1924, opened 2020-03-03
BUGFIX: EntityPrivilege fails on empty IN condition

albe:
**What I did**
Fixed the behavior of the `\Neos\Flow\Security\Authorization\Privilege\Entity\Doctrine\EntityPrivilege` if you execute an `in` operation on an empty array.

**How to verify it**
```yaml
matcher: 'isType("Neos\Neos\Domain\Model\Site") && !property("nodeName").in("context.userInformationContext.siteNodeNames")'
```

If `getSiteNodeNames()` evaluates to `[]` which is possible and valid, the PropertyGenerator tries to to `getParameter()` on this value which can only be applied to scalar values.

This is caused by only checking on an exact `null`, but an empty array should also be ignored.

Rebased follow-up to #1681


albe:
Merging as of my approval of the original PR

PR #1925, opened 2020-03-04
Revert "BUGFIX: Fix output of request and response in throwable storage"

kitsunet:
Reverts neos/flow-development-collection#1920

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -26,6 +26,8 @@ use Neos\Flow\Core\ProxyClassLoader;
 use Neos\Flow\Error\Debugger;
 use Neos\Flow\Error\ErrorHandler;
 use Neos\Flow\Error\ProductionExceptionHandler;
+use Neos\Flow\Http\Helper\RequestInformationHelper;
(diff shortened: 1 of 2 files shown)


kitsunet:
Would merge in 24h because the change is introducing a new bug.

daniellienert:
Mhm I still don't think it's a good idea to have that code duplicated. If you use another Implementation of the storage, you also should take care of the output rendering yourself as plain text rendering might not fit

kitsunet:
@daniellienert you can choose to ignore the given renderer if needed, but I think it makes sense to inject a default renderer for consistency and to give you the necessary information. The storage shoudln't really be responsible for rendering but mostly for storing the thing...

kitsunet:
Either way removing it is kind of a breaking change, so whatever we want to do in the future, just removing this as a bugfix is IMHO a bad idea.

daniellienert:
Will post a follow-up

kitsunet:
I'll investigate the doubled code, a short look with karsten earlier indicates that we might actually no longer have the use case I described. Will make some tests and find out!

albe:
After following through the code too, it looks more like the code in the constructor is unnecessary, because that will be called anyway. Correct? And our contract is basically "whatever logger is configured, it will get setup with calls to the renderer setters".

PR #1927, opened 2020-03-06
FEATURE: Add EEL Helper File.exists(filename)

markuspfeifenberger:
The EEL helper determines if a file exists. This is eg helpful, if file like templates are generated, and should only be rendered, if the template file exists, for example for generated favicon templates like this:

```
prototype(Neos.Neos:Page) {
head.favicons = Neos.Fusion:Template {
            resource = ${'resource://' + site.context.currentSite.siteResourcesPackageKey + '/Private/Templates/Page/Favicon.html'}
            templatePath = ${this.resource}
            @if.fileExists = ${File.exists(this.resource)}
        }
```

### DIFF
--- a/Neos.Eel/Classes/Helper/FileHelper.php
+++ b/Neos.Eel/Classes/Helper/FileHelper.php
@@ -67,7 +67,7 @@ class FileHelper implements ProtectedContextAwareInterface
      * @param string $filepath
      * @return bool
      */
-    public function exists(string $filepath)
+    public function exists(string $filepath): bool
     {
         return file_exists($filepath);
     }
--- a/Neos.Eel/Classes/Helper/FileHelper.php
+++ b/Neos.Eel/Classes/Helper/FileHelper.php
@@ -62,6 +62,17 @@ class FileHelper implements ProtectedContextAwareInterface
     }
 
     /**
+     * File exists
+     *
+     * @param string $filepath
+     * @return bool
+     */
+    public function exists(string $filepath)
+    {
+        return file_exists($filepath);
+    }
+    
+    /**
      * @param string $methodName
      * @return bool
      */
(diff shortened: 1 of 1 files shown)


Neos.Eel/Classes/Helper/FileHelper.php
```diff
@@ -61,6 +61,17 @@ public function stat(string $filepath)
         return stat($filepath);
     }
 
+    /**
+     * File exists
+     *
+     * @param string $filepath
+     * @return bool
+     */
+    public function exists(string $filepath)
```
daniellienert:
```suggestion
    public function exists(string $filepath): bool
```

daniellienert:
Hey @markuspfeifenberger, thanks for the PR.
I have taken the liberty to remove the unnecessary usage of @context in the example.

albe:
Hey and thank you for this :) Looks fine for me and is simple enough to get in without tests.


If @daniellienert is okay and removes his CR I'd merge. So congrats on your first contribution!

Neos.Eel/Classes/Helper/FileHelper.php
```diff
@@ -61,6 +61,17 @@ public function stat(string $filepath)
         return stat($filepath);
     }
 
+    /**
+     * File exists
```
jonnitto: Congrats on your first contribution! I would suggest adding here a better explanation because this will go into the reference documentation

albe:
```suggestion
     * Check if the given file path exists
```

albe:
friendly ping @daniellienert :)

daniellienert:
@albe looks good now, took the liberty to hit the button

Issue #1889, opened 2019-12-18
FEATURE: Rework ComponentChain to PSR15 middlewares


albe:
Our `ComponentChain` was a good abstraction for an extensible HTTP kernel back in the days. Now, since [PSR-15](https://www.php-fig.org/psr/psr-15/) is a thing and we already adopted [PSR-7](https://github.com/neos/flow-development-collection/issues/658) successfully, the mechanics of the `ComponentChain` don't work out all too well any more, due to the newly immutable nature of HTTP requests and responses.

We basically pass a reference to those two objects (`ComponentContext`) all around the places. When the request/response objects were still mutable, that was not so much of an issue, because you could just get the reference to the current response, alter it and then could be sure that all places accessing the component context would see those changes. While convenient, that was also prone to some oddities in behaviour.
Since the change to PSR-7 one now may not forget to update the context with your altered request/response or a whole different class of weird things happen. This is cumbersome to say the least.

PSR-15 solves that issue at least for the HTTP kernel, by passing instances of the incoming request through the chain and returning instances of response, once the innermost middleware or handler has been called. Outside of that invocation chain, nothing should have access to those objects, hence a `ComponentContext` becomes unnecessary.

So here's some idea on how to adapt to PSR-15 in a b/c as possible way, so we can gradually phase out the `ComponentChain` and the `ComponentContext`.

- introduce a PSR15 middleware chain, that is configured roughly like the current http component chain
- PSR-15 doesn't distinguish incoming (pre-process) and outgoing (post-process) middlewares. Any middleware can decide on it's own to be the one, the other or both. We could still introduce three levels of the chain, like we have for component chains, though that will only be a virtual separation. The order of invocation would be `pre-process -> post-process -> process`, since the response will only be available once the innermost `process` middleware (dispatch for MVC) was invoked
- make the current component chain a PSR-15 middleware itself (or maybe rather a RequestHandler, since the component chain is the response creating innermost middleware)
- gradually replace existing http components with PSR-15 middlewares

t.b.f.d.

Related to #658

albe:
- after some fiddling around, it turns out that wrapping the old component chain in middlewares has one big issue: the `ComponentContext` is supposed to exist in the `HttpRequestHandler` and it is the global entry point to access the current Http Request and Response. This means we need to pass the `ComponentContext` through the middlewares, so it ends up in the final `ComponentChainMiddleware`
[...]

PR #1928, opened 2020-03-08
FEATURE: Add configurable PSR-15 middleware kernel

albe:
This introduces a configurable middleware chain as the new main entry point into HTTP request handling. The old component chain is itself now a middleware "dispatch" that is configured to run at the `end` and other middlewares can be configured to run before, which also means it runs after "dispatch" due to how PSR-15 middlewares work.

Each middleware itself decides if it runs only before, after, or both by invoking the next handler appropriately. See also https://www.php-fig.org/psr/psr-15/meta/

This allows to easily manipulate the incoming HttpRequest or the outgoing HttpResponse with existing middlewares (e.g. https://github.com/middlewares/awesome-psr15-middlewares) and even completely short-cut Flows old component chain handling to handle special cases more performantly.

Depends on #1967
Related to #1889


albe:
Got the welcome page showing with this now.

Neos.Flow/Classes/Http/Middleware/MiddlewaresChainComponent.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Neos\Flow\Http\Middleware;
+
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Http\Component\ComponentContext;
+use Neos\Flow\Http\Component\ComponentInterface;
+
+/**
+ * A http component that invokes a PSR-15 middlewares chain
+ */
+class MiddlewaresChainComponent implements ComponentInterface
```
albe: to be removed

Neos.Flow/Classes/Http/Component/ComponentContext.php
```diff
@@ -54,7 +55,7 @@ class ComponentContext
      * @param ServerRequestInterface $httpRequest
      * @param ResponseInterface $httpResponse
      */
-    public function __construct(ServerRequestInterface $httpRequest, ResponseInterface $httpResponse)
+    public function __construct(ServerRequestInterface $httpRequest = null, ResponseInterface $httpResponse = null)
```
albe: Required to make the ComponentContext injectable

Neos.Flow/Classes/Http/Component/ComponentAfterWrapperMiddleware.php
```diff
@@ -0,0 +1,65 @@
...
+use Neos\Flow\Annotations as Flow;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * A middleware that wraps an old Http Component and invokes it after continuing the middleware chain
+ */
+class ComponentAfterWrapperMiddleware implements MiddlewareInterface
```
albe: Name?

Neos.Flow/Classes/Http/Component/ComponentBeforeWrapperMiddleware.php
```diff
@@ -0,0 +1,64 @@
...
+use Neos\Flow\Annotations as Flow;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * A middleware that wraps an old Http Component and invokes it before continuing the middleware chain
+ */
+class ComponentBeforeWrapperMiddleware implements MiddlewareInterface
```
albe: Name?

Neos.Flow/Classes/Http/Middleware/MiddlewaresChainFactory.php
```diff
@@ -0,0 +1,94 @@
...
+        foreach ($sortedChainConfiguration as $middlewareName => $configuration) {
+            if (isset($configuration['chain'])) {
+                $middleware = $this->create($configuration['chain'], $middlewareName);
+            } else {
+                if (!isset($configuration['middleware'])) {
+                    throw new Exception(sprintf('Middleware chain could not be created because no middleware class name is configured for middleware "%s"', $middlewareName), 1401718283);
+                }
+                // TODO: lazy instantiation of the middlewares?
+                $middleware = $this->objectManager->get($configuration['middleware']);
+                // TODO: b/c layer that allows configuring old Http Components as middlewares
```
albe: Not yet sure how to deal with "around" components

Neos.Flow/Classes/Http/RequestHandler.php
```diff
@@ -92,15 +91,17 @@ public function handleRequest()
     {
         // Create the request very early so the ResourceManagement has a chance to grab it:
         $request = ServerRequest::fromGlobals();
+        // TODO: Required for b/c - can be removed with Flow 7
         $response = new \GuzzleHttp\Psr7\Response();
         $this->componentContext = new ComponentContext($request, $response);
+        $this->bootstrap->setEarlyInstance(ComponentContext::class, $this->componentContext);
```
albe: This makes the Context injectable and is needed to make the `MiddlewaresChainFactory` be instanciable via object management.

Neos.Flow/Classes/Mvc/DispatchMiddleware.php
```diff
@@ -0,0 +1,90 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * A MVC dispatch middleware
+ *
+ * Note, this is a final middleware that will not delegate further.
+ */
+class DispatchMiddleware implements MiddlewareInterface
```
albe: Not yet used - but this is how the dispatch component could look as a middleware

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -225,7 +225,11 @@ protected function parseConfigurationArray($objectName, array $rawConfigurationO
                             } else {
                                 throw new InvalidObjectConfigurationException('Invalid configuration syntax. Expecting "value", "object" or "setting" as value for argument "' . $argumentName . '", instead found "' . (is_array($argumentValue) ? implode(', ', array_keys($argumentValue)) : $argumentValue) . '" (source: ' . $objectConfiguration->getConfigurationSourceHint() . ')', 1230563250);
                             }
-                            $objectConfiguration->setArgument($argument);
+                            if (isset($rawConfigurationOptions['factoryObjectName'])) {
```
albe: BUGFIX for proxies receiving injections of the factory arguments

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
```diff
@@ -44,7 +44,7 @@ public function getContext();
      * Dependency Injection.
      *
      * @param string $objectName The name of the object to return an instance of
-     * @param ...mixed $arguments Any number of arguments that should be passed to the constructor of the object
+     * @param mixed[] ...$arguments Any number of arguments that should be passed to the constructor of the object
```
albe: Apparently this is how jetbrains expects variadic arguments to be annotated

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -257,6 +257,12 @@ Neos\Flow\Http\Component\ComponentChain:
     1:
       setting: Neos.Flow.http.chain
 
+Neos\Flow\Http\Middleware\MiddlewaresChain:
+  factoryObjectName: Neos\Flow\Http\Middleware\MiddlewaresChainFactory
+  arguments:
+    1:
+      setting: Neos.Flow.http.middlewares
```
albe: Maybe we can even use the old component chain config directly, if we handle the order of components correctly

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -24,21 +24,42 @@ Neos:
...
-
+      # Defines the PSR-15 middlewares that a request goes through in a flow of "preprocess", "process" and "postprocess"
+      middlewares:
         'preprocess':
           position: 'before process'
           chain:
             'trustedProxies':
               position: 'start'
-              component: 'Neos\Flow\Http\Component\TrustedProxiesComponent'
+              middleware: 'Neos\Flow\Http\Component\TrustedProxiesComponent'
```
albe: We can configure old components as middlewares, but it currently only works for the preprocess and postprocess chains.

albe:
Travis is green apart from psalm, which I will fix on the go.

albe:
Most important things to decide:
- mostly keep old components (for b/c of relative positioning of custom components) or migrate to middlewares as much as possible already?
- reuse the current `chain` config and make configured components work correctly (hard to test for in every circumstance) or stick with the new config option and deprecate `chain`?
- is there a better way to get the early created `ComponentContext` into the `ComponentChainMiddleware`? Do we maybe not even need to create it that early any more?

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -24,21 +24,42 @@ Neos:
...
             'getSessionCookieFromRequest':
               position: 'after trustedProxies'
-              component: 'Neos\Flow\Session\Http\SessionRequestComponent'
+              middleware: 'Neos\Flow\Session\Http\SessionRequestComponent'
+        'process':
```
albe: Probably this "chain" can be completely removed, since it should only consist of a single middleware - or rather have `DispatchMiddleware` (`ComponentChainMiddleware` be "last" in `preprocess`)

albe:
Currently incorporates #1967 and #1969 out of necessity to make middlewares like https://github.com/middlewares/awesome-psr15-middlewares work without wrapping in factory classes - will rebase once these two are merged

albe:
Thanks for the review!

> I had no success to configure static factory methods or virtual objects

For virtual objects this needs to be rebased. Static factory should work if you just configure an object with just `factoryMethodName: Fully\Qualified\Namespace\Class::method` (and possible additional arguments).

> Is there really a benefit in keeping those chains of chains on the long run.

Nope, not in the longer run. I thought it would help people transition to middlewares, so we can possibly get rid of the component chain easier without hazzle. But maybe it's not really necessary at all?

> I personally think that having one chain of middlewares is a much simpler approach that also is easier to get on the long run.

I do think having two chains is generally not too bad, as it would make the concept of pre- and post- middleware a bit more clear (though only virtually, any middleware at any position can still act in any or both ways). At least I myself had problems with understanding the concept and the fact a middleware that is supposed to change the response needs to still be linked in **before** the final response generating middleware (dispatch).

The biggest issue I have with the multiple chains is with the oddity to have to actually sort "postprocess" **before** "preprocess" or "process" (if the MVC dispatch is last in "preprocess", because having a whole "process" chain for a single middleware is odd - otherwise MVC dispatch would need to be hardcoded, which I'm also not sooo convinced with). 🤷‍♂ Maybe MVC dispatch isn't a middleware at all, but the PSR15 RequestHandler instead and needs to be configured separately?

Or maybe middleware just doesn't play well with such a multi-stage concept and we really need a single chain.

mficzel:
Having two chains for distinct purposes that are internally the same is imho a problem as it leads developers into assuming things that are not there. Also it makes a simple concept that developers may already know more complicated.

I personally would prefer an approach with one middleware-chain on the long run. That would mean for now backwards compatible implementations of the current http-chains as middlewares but on the long run those should be deprecated entirely and be removed with Flow 7 or 8. 

We could even provide a ComponentMiddlewareWrapper and a Migration converts the chain configurations into middleware configurations (based on ComponentMiddlewareWrapper and virtualObjectConfigurations).

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -24,15 +24,21 @@ Neos:
       # request (default).
       baseUri: NULL
 
+      # Defines the PSR-15 middlewares that a request goes through in a flow
+      middlewares:
```
albe, line 28: This is now only a single "chain" out of the box, though it is still possible to configure chains - this could be helpful to group middlewares and sort the whole groups via positional array sorter

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -24,9 +24,15 @@ Neos:
...
+        'dispatch':
```
albe, line 29: Note: This is called `dispatch` because in the future we will end up with an actual `MvcDispatchMiddleware` here, so other middlewares configured in relative position to this one won't have to change.

mficzel: i would be fine with calling it `legacyHttpChain` now and remove it with Flow 7

albe: We could do this, but then again at some point we will need some "final middleware" that creates the response, which IMO is our MVC dispatch. So I'd be in favor of finding a good appropriate name for this future single middleware right now.
`mvcdispatch`/`mvc-dispatch` could be more exact

Neos.Flow/Classes/Http/Component/ComponentContext.php
```diff
@@ -25,6 +25,7 @@
  *
  * @api
  * @Flow\Proxy(false)
+ * @Flow\Scope("singleton")
```
albe, line 28:
```suggestion
 * @Flow\Scope("singleton")
 * @deprecated Will be removed without replacement with next major
```

Neos.Flow/Classes/Http/Middleware/ComponentChainMiddleware.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
albe, line 1:
```suggestion
<?php
declare(strict_types=1);

```

Neos.Flow/Classes/Http/Middleware/MiddlewaresChain.php
```diff
@@ -0,0 +1,92 @@
+<?php
```
albe, line 1:
```suggestion
<?php
declare(strict_types=1);

```

Neos.Flow/Classes/Http/Middleware/MiddlewaresChainFactory.php
```diff
@@ -0,0 +1,82 @@
+<?php
```
albe, line 1:
```suggestion
<?php
declare(strict_types=1);

```

Neos.Flow/Classes/Http/Middleware/MiddlewaresChainFactory.php
```diff
@@ -0,0 +1,84 @@
...
+        if (empty($chainConfiguration)) {
+            return new MiddlewaresChain($parentChain);
+        }
+        $arraySorter = new PositionalArraySorter($chainConfiguration);
+        $sortedChainConfiguration = $arraySorter->toArray();
+
+        $middlewaresChain = [];
+        foreach ($sortedChainConfiguration as $middlewareName => $configuration) {
+            if (isset($configuration['chain'])) {
+                $middleware = $this->create($configuration['chain'], $middlewareName);
```
mficzel, line 66: Is the 'chain' in 'chain' still a thing or is this a leftover?

mficzel: Would be nice to make the middleware nullable. Since some might have to disable a middleware that is added by a package.

Neos.Flow/Configuration/Settings.Http.yaml
mficzel: In the end both would work, i prefer to not name things after future roles as those assumptions may change.

there is no harm if we end up with 

```
middlewares:
   dispatch:
        middleware: SomeClass
   legacyHttpChain: null
```

Neos.Flow/Classes/Http/Middleware/MiddlewaresChainFactory.php
albe: I left the 'chain' in, because I do think there might be a use-case to configuring different sub-chains that describe a processing step.

> Would be nice to make the middleware nullable.

What about just replacing by a `NopMiddleware`? I'm not so much a fan of `'foo': null` as a measure to "disable a middleware"

mficzel: A nop or passtrough middleware would fine aswell.

mficzel: Regarding sub chains. I would like to handle this as middleware once someone needs it and maybe even in a non-core package. Offcourse this would require to pass configuration to the middleware constructor/factory from the settings. I see this as a possible feature for later.

However configuring a middleware via setting may be useful anyways as some middlewares i've seen like the "middlewares/minifier" will be put into the same chain more then once.  Not sure the Objects.yaml with virtual Objects is a good final solution for that.

So this might mean to leave the `chain` out now as it adds api we may not want to support and maybe add middleware configuration via settings as a later feature.

Neos.Flow/Configuration/Settings.Http.yaml
albe: In general I'm with you regarding "assumptions may change". But:
- that we will have an "MVC dispatch" at some point is pretty much a given, unless we're talking about Flow 20 or s.th. where probably a new approach is the base for the framework ;)
- I do think that renaming the middleware config does some harm, because this configuration is based on sorted arrays, which are relative to other *names* - so if someone will configure his own middleware as `before legacyHttpChain` now, he will need to change that later on. Not world-shaking harm, but something we could prevent *most likely* now

```suggestion
        'mvc-dispatch':
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
```diff
@@ -118,9 +121,71 @@ the ``Package`` class of the package containing the request handler::
...
+	}
+
+To activate a middle, it must be configured in the ``Settings.yaml``::
+
+	Neos:
+	  Flow:
+	    http:
+	      middlewares:
+          'custom':
+            position: 'before dispatch'
```
albe, line 171:
```suggestion
            position: 'before mvc-dispatch'
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
```diff
@@ -118,9 +121,71 @@ the ``Package`` class of the package containing the request handler::
...
+the following middleware. A middleware chain basically works like a onion ring, where each middleware is a single layer
+of the onion around the inner core of the application. Each request passes inside through the layer and a response passes
+outside through the layer.
+
 Component Chain
 ---------------
 
+..note::
+
+  The Component Chain is considered deprecated as of Flow 6.2 and will be removed in a later version. All components will
```
albe, line 186:
```suggestion
  The Component Chain is considered deprecated as of Flow 6.3 and will be removed in a later version. All components will
```

Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.http.schema.yaml
```diff
@@ -32,6 +32,31 @@ properties:
...
+            -
+              type: dictionary
+              additionalProperties:
+                type: dictionary
+                additionalProperties: false
+                properties:
+                  'position': [string, number]
+                  'middleware': { type: string, format: class-name }
+                  'chain':
+                    type: dictionary
```
albe, line 59:
```suggestion
```

Remove sub-chain configurability for now

Neos.Flow/Classes/Http/Middleware/MiddlewaresChainFactory.php
```diff
@@ -0,0 +1,84 @@
...
+                    throw new Exception(sprintf('Middleware chain could not be created because no middleware class name is configured for middleware "%s"', $middlewareName), 1401718283);
+                }
+                $middleware = $this->objectManager->get($configuration['middleware']);
+                if (!$middleware instanceof MiddlewareInterface) {
+                    throw new Exception(sprintf('Middleware chain could not be created because the class "%s" does not implement the MiddlewareInterface in middleware "%s"', $configuration['middleware'], $middlewareName), 1401718283);
+                }
+                if (method_exists($middleware, 'setComponentContext')) {
+                    $middleware->setComponentContext($this->componentContext);
+                }
+            }
```
albe, line 78:
```suggestion
            if (!isset($configuration['middleware'])) {
                throw new Exception(sprintf('Middleware chain could not be created because no middleware class name is configured for middleware "%s"', $middlewareName), 1401718283);
            }
            $middleware = $this->objectManager->get($configuration['middleware']);
            if (!$middleware instanceof MiddlewareInterface) {
                throw new Exception(sprintf('Middleware chain could not be created because the class "%s" does not implement the MiddlewareInterface in middleware "%s"', $configuration['middleware'], $middlewareName), 1401718283);
            }
            if (method_exists($middleware, 'setComponentContext')) {
                $middleware->setComponentContext($this->componentContext);
            }
```

albe:
This is now IMO merge ready - I'd like to get it into master ASAP in order to start upgrading the existing components and collect experience with how this works out and how components should be migrated. Target goal is to have PSR-15 in 6.3 and remove the component chain fully in 7 already with a fully (or mostly) transparent migration plan.

At this point I'd also like to get some feedback by @bwaidelich since he originally built the whole component chain base

albe:
Okay, a little further improvement to avoid the nullable arguments in `ComponentContext` (which made psalm rightfully unhappy) - I configured `ComponentContext` via `Objects.yaml` instead, because both arguments can be created with `new` or the `fromGlobals` static guzzle factory that we use anyway. So that now makes the `ComponentContext` injectable generally (though it factually already was with the before change). Also I made Flow compatible with GuzzleHttp 2 (which contains http factories). Build should be green now!

Neos.Flow/Configuration/Settings.Http.yaml
sorenmalling: I love that we remove the chain and have a "single entry point" and define them by position 😍

albe:
It's still in dev, so you need to allow dev stability :) It's basically the current master: https://github.com/guzzle/psr7/blob/master/composer.json#L52

mficzel:
@albe after playing with symfony style controllers and routing in https://github.com/neos/flow-development-collection/pull/2014 it may be a better idea to call the `mcv-dispatch` middleware `dispatch` instead as MVC may actually be an implementation detail here.

Not sure about this yet and the mentioned pr at the moment is also just an experiment now, but a nice looking one.

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -24,9 +24,15 @@ Neos:
       # request (default).
       baseUri: NULL
 
+      # Defines the PSR-15 middlewares that a request goes through in a flow
+      middlewares:
+        'mvc-dispatch':
```
mficzel, line 29: Maybe call this only `dispatch` as mvc is an implementation detail of the controller.

albe: But isn't the fact the request gets dispatched to a **Controller** already part of the MV**C** approach? I would have considered a different approach to be achieved by a different dispatcher - but maybe you are right. OTOH at least it lives in the MVC namespace, so I *guess* the name is okay-ish?

For reference: https://blog.ircmaxell.com/2014/11/alternatives-to-mvc.html
I mean, I never heard of most of those other patterns and following the conclusion, they probably are all the same "MVC" aproach under the coat (a triad of logic, represent, interact), just with little different focus. Iff we ever introduce a really different pattern to handle a request, we might name it `sna-dispatch` (**S**uper **N**ew **A**cronym) - or maybe not even "dispatch" at all?

One possible use-case could be an [Neos.EventSourcing](https://github.com/neos/Neos.EventSourcing) specific "dispatcher" or "commandhandler"?

mficzel: MVC is distinct from dispatching that only is intermingled in the current implementation. We could consider to clean this up in future and extract the dispatching from there. 

I also thought this would be a safe assumption now think that the core is indeed a pure action that is given a request and that returns a response. Wether there is a model or a view or any other pattern involved is nothing the dispatcher has to care about. 

Offcourse this can be implemented in special controller base classes like actionController like it is in flow as all argumentMapping and validation is already done in actionController and not AbstractController. 

Obviously this all does not affect this pr. The only question we have to discuss here is which name would be best and fitting for future developments. I personally think that `dispatch` is best fitting from the perspective of the http-middleware-chain.

mficzel: Maybe the name `controller-dispatch` would be fitting as it only assumes a derived class of AbstractController is called.

albe: > The only question we have to discuss here is which name would be best and fitting for future developments.

Right and I guess you are right. `dispatch` covers the exact case of this very middleware - no matter if it will dispatch to some MVC, command, or other pattern. In fact, if we (or a third party) chooses to switch this, it makes most sense to **replace** this middleware anyway, instead of prepending to it and then short-circuiting it. So the super generic name `dispatch` is fitting for all means and purposes of this "ever last middleware". At least more than `mvc-dispatch`.

albe:
```suggestion
        'dispatch':
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Http.rst
```diff
@@ -118,9 +121,64 @@ the ``Package`` class of the package containing the request handler::
...
+  }
+
+To activate a middleware, it must be configured in the ``Settings.yaml``::
+
+  Neos:
+    Flow:
+      http:
+        middlewares:
+          'custom':
+            position: 'before mvc-dispatch'
```
albe, line 164:
```suggestion
            position: 'before dispatch'
```

albe:
So, let's do this :)

PR #1931, opened 2020-03-08
BUGFIX: Failed validation prevents resource from being persisted

### LINKED ISSUE #769: PersistentResource gets persisted if a attached validation fails
### Description

If a model has a `PersistentResource` as a property and a custom validator attached, validating ex. the Media Type (allow only jpg and png) the PersistentResource is persisted and stored even though the validation fails.

### Steps to Reproduce

1. Create a model with a `PersistentResource` as property
2. Create a validator that returns false
3. Submit a form with a resource
4. See in the log 

```
16-12-09 11:35:14 67587      DEBUG     Flow                 Successfully imported the uploaded file "8196730941.webm" into the resource collection "persistent" (storage: "defaultPersistentResourcesStorage", a Neos\Flow\ResourceManagement\Storage\WritableFileSystemStorage. SHA1: f3fa2e994fb478dfc9eac52c5addc51de1f7fb38)
16-12-09 11:35:14 67587      ERROR     Flow                 Validation failed while trying to call Application\RestApi\Controller\ImageController->uploadAction().
Error for data:  The media type "video/webm" is not allowed for this image.

16-12-09 11:35:14 67587      DEBUG     Flow                 FileSystemSymlinkTarget: Published file. (target: localWebDirectoryPersistentResourcesTarget, file: f3fa2e994fb478dfc9eac52c5addc51de1f7fb38/8196730941.webm)
```

#### Expected behavior

As the validation of the property returns false, the whole object is considered "invalid" and therefore the resource should not be persisted.

As the import is managed inside the `ResourceTypeConverter` and it's applied before the validation, the resource is already persisted at that point. 

It could be argumented that the validator could clean up, but that's not the task for a validator that only returns true/false. 

Somehow a way to interact with the `ResourceTypeConverter` and be able to have validator(s) validate the input

#### Actual behavior

The resource is persisted

### Affected Versions

Neos: * 

Flow: *

albe:
This is achieved by deferring the importing and persisting of resources until after validation.

Fixes #769

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
@@ -12,6 +12,7 @@ namespace Neos\Flow\Mvc\Controller;
  */
 
 use GuzzleHttp\Psr7\Uri;
+use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\ResourceManagement\PersistentResource;
 use Neos\Flow\ResourceManagement\ResourceManager;
 use Neos\Flow\Http\Helper\MediaTypeHelper;
@@ -382,6 +383,7 @@ abstract class AbstractController implements ControllerInterface
      */
     protected function mapRequestArgumentsToControllerArguments()
     {
+        /** @var $argument Argument */
         foreach ($this->arguments as $argument) {
             $argumentName = $argument->getName();
             if ($this->request->hasArgument($argumentName)) {
@@ -391,7 +393,12 @@ abstract class AbstractController implements ControllerInterface
                 // TypeConverter previously, where validation could no longer prevent importing a resource into the system
                 $value = $argument->getValue();
                 if ($value instanceof PersistentResource && !$argument->getValidationResults()->hasErrors()) {
-                    $this->resourceManager->persistResource($value);
+                    try {
+                        $this->resourceManager->persistResource($value);
+                    } catch (\Exception $exception) {
+                        $this->logger->warning('Could not import an uploaded file', ['exception' => $exception] + LogEnvironment::fromMethodName(__METHOD__));
+                        $argument->getValidationResults()->addError(new Error\Error('During import of an uploaded file an error occurred. See log for more details.', 1264517906));
+                    }
                 }
             } elseif ($argument->isRequired()) {
                 throw new RequiredArgumentMissingException('Required argument "' . $argumentName  . '" is not set.', 1298012500);
--- a/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
@@ -12,6 +12,8 @@ namespace Neos\Flow\Mvc\Controller;
  */
 
 use GuzzleHttp\Psr7\Uri;
+use Neos\Flow\ResourceManagement\PersistentResource;
+use Neos\Flow\ResourceManagement\ResourceManager;
 use Neos\Flow\Http\Helper\MediaTypeHelper;
 use Neos\Flow\Mvc\ActionResponse;
 use Neos\Flow\Mvc\Exception\ForwardException;
@@ -88,6 +90,12 @@ abstract class AbstractController implements ControllerInterface
      * @var PersistenceManagerInterface
      */
     protected $persistenceManager;
+
+    /**
+     * @Flow\Inject
+     * @var ResourceManager
+     */
+    protected $resourceManager;
 
     /**
      * A list of IANA media types which are supported by this controller
@@ -378,6 +386,13 @@ abstract class AbstractController implements ControllerInterface
             $argumentName = $argument->getName();
             if ($this->request->hasArgument($argumentName)) {
                 $argument->setValue($this->request->getArgument($argumentName));
+
+                // Needed because we have an "auto-persistence" for mapped Resources since ages, but it was done in the
+                // TypeConverter previously, where validation could no longer prevent importing a resource into the system
+                $value = $argument->getValue();
+                if ($value instanceof PersistentResource && !$argument->getValidationResults()->hasErrors()) {
+                    $this->resourceManager->persistResource($value);
+                }
             } elseif ($argument->isRequired()) {
                 throw new RequiredArgumentMissingException('Required argument "' . $argumentName  . '" is not set.', 1298012500);
             }
--- a/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
+++ b/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
@@ -39,7 +39,7 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     /**
      * The source for the resource when being newly created for deferred persistence
      *
-     * @var string|resource|null
+     * @var string|array|resource|null
      * @Flow\Transient
      */
     protected $source;
@@ -136,7 +136,7 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     /**
      * The source is required for deferred persistence (after validation) in the framework and should otherwise not be set.
      *
-     * @param string|resource|null $source
+     * @param string|array|resource|null $source
      */
     public function __construct($source = null)
     {
@@ -170,7 +170,7 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     }
 
     /**
-     * @return resource|string|null
+     * @return resource|array|string|null
      */
     public function getSource()
     {
@@ -181,7 +181,7 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      * Detach the source from this resource and return it.
      * Used for deferred persistence.
      *
-     * @return resource|string|null
+     * @return resource|array|string|null
      */
     public function detachSource()
     {
@@ -327,6 +327,8 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
         if (!$this->fileSize && $this->source !== null) {
             if (is_string($this->source)) {
                 $this->fileSize = filesize($this->source) ?: 0;
+            } elseif (is_array($this->source)) {
+                $this->fileSize = $this->source['size'] ?? 0;
             } elseif (is_resource($this->source)) {
                 $stat = fstat($this->source);
                 $this->fileSize = $stat['size'] ?? 0;
@@ -346,6 +348,8 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
         if (!$this->sha1 && $this->source !== null) {
             if (is_string($this->source)) {
                 $this->sha1 = sha1_file($this->source);
+            } elseif (is_array($this->source) && isset($this->source['tmp_name'])) {
+                $this->sha1 = sha1_file($this->source['tmp_name']);
             } elseif (is_resource($this->source)) {
                 $pos = ftell($this->source);
                 rewind($this->source);
@@ -383,6 +387,8 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
         if (!$this->md5 && $this->source !== null) {
             if (is_string($this->source)) {
                 $this->md5 = md5_file($this->source);
+            } elseif (is_array($this->source) && isset($this->source['tmp_name'])) {
+                $this->md5 = md5_file($this->source['tmp_name']);
             } elseif (is_resource($this->source)) {
                 $pos = ftell($this->source);
                 rewind($this->source);
--- a/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
+++ b/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
@@ -37,6 +37,14 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     protected $collectionName = ResourceManager::DEFAULT_PERSISTENT_COLLECTION_NAME;
 
     /**
+     * The source for the resource when being newly created for deferred persistence
+     *
+     * @var string|resource|null
+     * @Flow\Transient
+     */
+    protected $source;
+
+    /**
      * Filename which is used when the data of this resource is downloaded as a file or acting as a label
      *
      * @var string
@@ -126,6 +134,14 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     protected $temporaryLocalCopyPathAndFilename;
 
     /**
+     * @param string|resource|null $source
+     */
+    public function __construct($source = null)
+    {
+        $this->source = $source;
+    }
+
+    /**
      * Protects this PersistentResource if it has been persisted already.
      *
      * @param integer $initializationCause
@@ -149,6 +165,27 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     public function getStream()
     {
         return $this->resourceManager->getStreamByResource($this);
+    }
+
+    /**
+     * @return resource|string|null
+     */
+    public function getSource()
+    {
+        return $this->source;
+    }
+
+    /**
+     * Detach the source from this resource and return it.
+     * Used for deferred persistence.
+     *
+     * @return resource|string|null
+     */
+    public function detachSource()
+    {
+        $source = $this->source;
+        $this->source = null;
+        return $source;
     }
 
     /**
--- a/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
+++ b/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
@@ -159,7 +159,7 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      *
      * Note: The caller is responsible to close the returned resource by calling fclose($stream)
      *
-     * @return resource | boolean A stream which points to the data of this resource for read-access or false if the stream could not be obtained
+     * @return resource|boolean A stream which points to the data of this resource for read-access or false if the stream could not be obtained
      * @api
      */
     public function getStream()
@@ -322,6 +322,14 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      */
     public function getFileSize()
     {
+        if (!$this->fileSize && $this->source !== null) {
+            if (is_string($this->source)) {
+                $this->fileSize = filesize($this->source) ?: 0;
+            } elseif (is_resource($this->source)) {
+                $stat = fstat($this->source);
+                $this->fileSize = $stat['size'] ?? 0;
+            }
+        }
         return $this->fileSize;
     }
 
@@ -333,6 +341,16 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      */
     public function getSha1()
     {
+        if (!$this->sha1 && $this->source !== null) {
+            if (is_string($this->source)) {
+                $this->sha1 = sha1_file($this->source);
+            } elseif (is_resource($this->source)) {
+                $pos = ftell($this->source);
+                rewind($this->source);
+                $this->sha1 = sha1(fread($this->source, $this->fileSize));
+                fseek($this->source, $pos);
+            }
+        }
         return $this->sha1;
     }
 
@@ -360,6 +378,16 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      */
     public function getMd5()
     {
+        if (!$this->md5 && $this->source !== null) {
+            if (is_string($this->source)) {
+                $this->md5 = md5_file($this->source);
+            } elseif (is_resource($this->source)) {
+                $pos = ftell($this->source);
+                rewind($this->source);
+                $this->md5 = md5(fread($this->source, $this->fileSize));
+                fseek($this->source, $pos);
+            }
+        }
         return $this->md5;
     }
 
--- a/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
+++ b/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
@@ -134,6 +134,8 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     protected $temporaryLocalCopyPathAndFilename;
 
     /**
+     * The source is required for deferred persistence (after validation) in the framework and should otherwise not be set.
+     *
      * @param string|resource|null $source
      */
     public function __construct($source = null)
--- a/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
+++ b/Neos.Flow/Classes/ResourceManagement/PersistentResource.php
@@ -134,13 +134,16 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     protected $temporaryLocalCopyPathAndFilename;
 
     /**
-     * The source is required for deferred persistence (after validation) in the framework and should otherwise not be set.
+     * Create a resource for deferred persistence (after validation) in the framework.
+     * This is only required internally in the framework and should not be manually used.
      *
-     * @param string|array|resource|null $source
+     * @param string|array|resource $source
      */
-    public function __construct($source = null)
+    public static function createDeferred($source)
     {
-        $this->source = $source;
+        $resource = new self();
+        $resource->source = $source;
+        return $resource;
     }
 
     /**
@@ -170,11 +173,11 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
     }
 
     /**
-     * @return resource|array|string|null
+     * @return bool True if this resource is deferred and has a source attached that needs to be imported first.
      */
-    public function getSource()
+    public function isDeferred(): bool
     {
-        return $this->source;
+        return $this->source !== null;
     }
 
     /**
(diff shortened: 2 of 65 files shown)


albe:
When reviewing, please take a look at the commit https://github.com/neos/flow-development-collection/pull/1931/commits/5120bd76c14fd6bdaf4559ee51f7c50ff147555e first - it only includes the functional changes required for this. All others are only distracting, but useful docblock type changes.

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -12,6 +12,8 @@
  */
 
 use GuzzleHttp\Psr7\Uri;
+use Neos\Flow\ResourceManagement\PersistentResource;
```
albe: This sucks, but is required unless someone else has a good idea how to trigger resource importing **after** validation but before controller invokation.

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -285,6 +322,14 @@ public function setFileSize($fileSize)
      */
     public function getFileSize()
     {
+        if (!$this->fileSize && $this->source !== null) {
```
albe: This (and below lazy evaluations) is maybe not needed, but I'm unsure if the resource properties could be accessed before it being "imported" via the resource manager

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -374,10 +383,23 @@ protected function throwStatus($statusCode, $statusMessage = null, $content = nu
...
                 $argument->setValue($this->request->getArgument($argumentName));
+
+                // Needed because we have an "auto-persistence" for mapped Resources since ages, but it was done in the
+                // TypeConverter previously, where validation could no longer prevent importing a resource into the system
+                $value = $argument->getValue();
+                if ($value instanceof PersistentResource && !$argument->getValidationResults()->hasErrors()) {
+                    try {
+                        $this->resourceManager->persistResource($value);
+                    } catch (\Exception $exception) {
+                        $this->logger->warning('Could not import an uploaded file', ['exception' => $exception] + LogEnvironment::fromMethodName(__METHOD__));
```
albe: Needed to keep the behaviour of import exceptions ending up as validation errors

kdambekalns:
What about `importResource()`, it is obviously left unchanged in this PR. How do those changes affect that? Do they? Why and why not? I think there should be some explanation on that, ideally close to the code involved…

albe:
> What about importResource()?

That's a fair question. It (and sibling methods) are indeed unchanged, because an actual *import* of a Resource should make it persistent (in storage) immediately. Then again, those methods take arbitrary forms of data and convert that into a `PersistentResource`, while the new `persistResource` method takes a (deferred) Resource and only makes it persistent. Also, this makes sure b/c is guaranteed.

Now in regard to where to place this information closer to the code - I'll give it a thought and add something. Thanks for the feedback!

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -125,6 +133,14 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      */
     protected $temporaryLocalCopyPathAndFilename;
 
+    /**
+     * @param string|resource|null $source
```
albe:
```suggestion
     * The source is required for deferred persistence (after validation) in the framework and should otherwise not be set.
     *
     * @param string|resource|null $source
```

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -260,6 +260,40 @@ public function importUploadedResource(array $uploadInfo, $collectionName = self
         return $resource;
     }
 
+    /**
+     * Import a deferred resource into it's destination collection and persist it.
```
albe:
```suggestion
     * Import a deferred resource into it's destination collection and persist it.
     * This will do nothing, if the resource is not deferred (has no source attached).
```

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -260,6 +260,40 @@ public function importUploadedResource(array $uploadInfo, $collectionName = self
...
+            $importedResource = $collection->importResource($resource->detachSource());
+            $resource->setSha1($importedResource->getSha1());
+            $resource->setMd5($importedResource->getMd5());
+            $resource->setFileSize($importedResource->getFileSize());
+        } catch (Exception $exception) {
+            throw new Exception(sprintf('Importing an uploaded file into the resource collection "%s" failed.', $collectionName), 1375197680, $exception);
+        }
+
+        $this->resourceRepository->add($resource);
+        $this->logger->debug(sprintf('Successfully imported the uploaded file "%s" into the resource collection "%s" (storage: "%s", a %s. SHA1: %s)', $resource->getFilename(), $collectionName, $this->collections[$collectionName]->getStorage()->getName(), get_class($this->collections[$collectionName]->getStorage()), $resource->getSha1()));
```
albe:
```suggestion
        $this->logger->debug(sprintf('Successfully imported the uploaded file "%s" into the resource collection "%s" (storage: "%s", a %s. SHA1: %s)', $resource->getFilename(), $collectionName, $collection->getStorage()->getName(), get_class($collection->getStorage()), $resource->getSha1()));
```

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -125,6 +133,16 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      */
     protected $temporaryLocalCopyPathAndFilename;
 
+    /**
+     * The source is required for deferred persistence (after validation) in the framework and should otherwise not be set.
+     *
+     * @param string|resource|null $source
```
albe:
```suggestion
     * @param string|array|resource|null $source
```

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -36,6 +36,14 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      */
     protected $collectionName = ResourceManager::DEFAULT_PERSISTENT_COLLECTION_NAME;
 
+    /**
+     * The source for the resource when being newly created for deferred persistence
+     *
+     * @var string|resource|null
```
albe:
```suggestion
     * @var string|array|resource|null
```

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -143,14 +161,35 @@ public function initializeObject($initializationCause)
...
+     * @return resource|boolean A stream which points to the data of this resource for read-access or false if the stream could not be obtained
      * @api
      */
     public function getStream()
     {
         return $this->resourceManager->getStreamByResource($this);
     }
 
+    /**
+     * @return resource|string|null
```
albe:
```suggestion
     * @return resource|array|string|null
```

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -143,14 +161,35 @@ public function initializeObject($initializationCause)
...
+    public function getSource()
+    {
+        return $this->source;
+    }
+
+    /**
+     * Detach the source from this resource and return it.
+     * Used for deferred persistence.
+     *
+     * @return resource|string|null
```
albe:
```suggestion
     * @return resource|array|string|null
```

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -285,6 +324,14 @@ public function setFileSize($fileSize)
      */
     public function getFileSize()
     {
+        if (!$this->fileSize && $this->source !== null) {
+            if (is_string($this->source)) {
+                $this->fileSize = filesize($this->source) ?: 0;
+            } elseif (is_resource($this->source)) {
```
albe:
```suggestion
            } elseif (is_array($this->source)) {
                $this->fileSize = $this->source['size'] ?? 0;
            } elseif (is_resource($this->source)) {
```

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -296,6 +343,16 @@ public function getFileSize()
      */
     public function getSha1()
     {
+        if (!$this->sha1 && $this->source !== null) {
+            if (is_string($this->source)) {
+                $this->sha1 = sha1_file($this->source);
+            } elseif (is_resource($this->source)) {
```
albe:
```suggestion
            } elseif (is_array($this->source) && isset($this->source['tmp_name'])) {
                $this->sha1 = sha1_file($this->source['tmp_name']);
            } elseif (is_resource($this->source)) {
```

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -323,6 +380,16 @@ public function setSha1($sha1)
      */
     public function getMd5()
     {
+        if (!$this->md5 && $this->source !== null) {
+            if (is_string($this->source)) {
+                $this->md5 = md5_file($this->source);
+            } elseif (is_resource($this->source)) {
```
albe:
```suggestion
            } elseif (is_array($this->source) && isset($this->source['tmp_name'])) {
                $this->md5 = md5_file($this->source['tmp_name']);
            } elseif (is_resource($this->source)) {
```

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -260,6 +260,41 @@ public function importUploadedResource(array $uploadInfo, $collectionName = self
...
+        $collectionName = $resource->getCollectionName();
+        if (!isset($this->collections[$collectionName])) {
+            throw new Exception(sprintf('Tried to import an uploaded file into the resource collection "%s" but no such collection exists. Please check your settings and HTML forms.', $collectionName), 1375197544);
+        }
+
+        /* @var CollectionInterface $collection */
+        $collection = $this->collections[$collectionName];
+        try {
+            // Optimally the collection would just accept a PersistentResource and update it, but this would mean an API change
+            $importedResource = $collection->importResource($resource->detachSource());
```
albe:
```suggestion
            $source = $resource->detachSource();
            if (is_array($source)) {
                $importedResource = $collection->importUploadedResource($source, $collectionName);
            } else {
                $importedResource = $collection->importResource($source, $collectionName);
            }
```

albe:
After thinking around I can't seem to find a good place or wording to describe the non-effect on the import* methods, other then only describing the "deferred" concept better in the `PersistentResource` constructor: it is only intended for framework purposes and should not be manually used (unless you know what you do). So I'm happy for any ideas.

Noticed I missed handling arrays correctly and fixed that, but I think no upload ends up as an array in the converter any more since the PSR-7 change. Have to check that - maybe we can get rid of that for next major.

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
```diff
@@ -292,26 +311,28 @@ protected function handleUploadedFile(UploadedFileInterface $source, PropertyMap
...
+                return new FlowError('An error occurred while uploading. Please try again or contact the administrator if the problem remains', 1340193849);
         }
 
         if (isset($this->convertedResources[spl_object_hash($source)])) {
             return $this->convertedResources[spl_object_hash($source)];
         }
 
         try {
-            $resource = $this->resourceManager->importResource($source->getStream()->detach(), $this->getCollectionName($source, $configuration));
+            $resource = new PersistentResource($source->getStream()->detach());
```
kitsunet: That's a (kind of big) behavior change isn't it? If I use this converter I could so far expect it to create a persisted resource.

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
```diff
@@ -250,38 +260,47 @@ protected function handleHashAndData(array $source, PropertyMappingConfiguration
...
+                fwrite($tempFile, base64_decode($source['data']));
+                fseek($tempFile, 0);
+                $resource = new PersistentResource($tempFile);
             } elseif ($hash !== null) {
-                /** @var PersistentResource $resource */
-                $resource = $this->resourceManager->importResource($configuration->getConfigurationValue(ResourceTypeConverter::class, self::CONFIGURATION_RESOURCE_LOAD_PATH) . '/' . $hash, $collectionName, $givenResourceIdentity);
-                if (is_array($source) && isset($source['filename'])) {
-                    $resource->setFilename($source['filename']);
-                }
+                $resource = new PersistentResource($configuration->getConfigurationValue(ResourceTypeConverter::class, self::CONFIGURATION_RESOURCE_LOAD_PATH) . '/' . $hash);
```
kitsunet: Same as other comment, regarding behavior change.

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
```diff
@@ -207,12 +210,19 @@ protected function handleFileUploads(array $source, PropertyMappingConfiguration
         }
 
         try {
-            $resource = $this->resourceManager->importUploadedResource($source, $this->getCollectionName($source, $configuration));
+            $resource = new PersistentResource($source);
```
kitsunet: and third time ;)

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -125,6 +133,16 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      */
     protected $temporaryLocalCopyPathAndFilename;
 
+    /**
+     * The source is required for deferred persistence (after validation) in the framework and should otherwise not be set.
+     *
+     * @param string|array|resource|null $source
+     */
+    public function __construct($source = null)
```
kitsunet: ooof, especially with the multi type I am not a huge fan of this. ESPECIALLY as this thing is called persisted resource and it is not persisted at all if this is used with the temporary files. In fact someone could come up with the silly idea of actually creating a PersistedResource this way and just handing it to the repository (which obviously fails miserably). I don't like this change and would rather introduce something else for this use case...

kitsunet (changes requested):
I am flagging this as request changes, I am really not happy with this change as is and I hope we find a better solutoin (which might become a feature then...)

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
albe: Well, it is an `"eventually persisted resource"` so to say. I'd be happy to find another solution to allow validation to prevent persisting (and importing) a resource, but currently the property mapping is done *before* validation. And this is not a feature, but actually a bug. A Resource that doesn't pass validation should not end up in the database, let alone on persistent storage. But it's also not easy to just remove the resource from persistence again after validation, as you'd need to know if the resource was newly added or just updated (plus the unnecessary work done). So the converter really should not call `repository->add()` (the `PersistentObjectConverter` doesn't either), nor move the resource to a storage backend.

The only valid other option I see would be to split validation in two separate steps, then only have the first step allow rejecting resources (in their "source" form). But that would be one hell of a change in regards to code required, complexity, breakyness and required documentation to make people understand what the two validations are.

Anyway: Thanks for challenging the solution!

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
albe: I see your point here. What about a static constructor method that creates such a deferred persistent resource? Generally, you should not manually create such a deferred resource (you probably should never create a `PersistentResource` manually at all).

albe:
ping @kitsunet - did you have time to give it another thought as we discussed?

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
albe: PS: This whole converter is the reason I did not introduce a `EventuallyPersistedResource extends PersistentResource` class - because the converter would need to change, which would be super-breaking. With my solution this is more subtle and a `PersistentResource` is differentiated by the `source` property. One with a source has definitely not yet been imported! It of course won't work if you manually create a `new PersistentResource($someSource)` and add that to a repo. But that also didn't work before, you needed to use the `ResourceManager` and if you do that, this solution right now also works.
We could maybe also add a safety-check into the `ResourceRepository` that allows only adding resources that have no source attached, but then again, that would only make it better for those cases where you specify a `$source` - which in turn would be fine in scope of this change.

Neos.Flow/Classes/ResourceManagement/ResourceRepository.php
```diff
@@ -74,6 +74,9 @@ public function __construct()
      */
     public function add($object)
     {
+        if ($object instanceof PersistentResource && $object->isDeferred()) {
```
albe, line 77: @kitsunet see my other comment - I think this is a good measure to prevent the primary line of accidential misuse

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -125,6 +133,19 @@ class PersistentResource implements ResourceMetaDataInterface, CacheAwareInterfa
      */
     protected $temporaryLocalCopyPathAndFilename;
 
+    /**
+     * Create a resource for deferred persistence (after validation) in the framework.
+     * This is only required internally in the framework and should not be manually used.
+     *
+     * @param string|array|resource $source
+     */
+    public static function createDeferred($source)
```
albe, line 142: This also feels better than the optional constructor argument. Should have done that from the start!

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -143,14 +164,35 @@ public function initializeObject($initializationCause)
...
      */
     public function getStream()
     {
         return $this->resourceManager->getStreamByResource($this);
     }
 
+    /**
+     * @return bool True if this resource is deferred and has a source attached that needs to be imported first.
+     */
+    public function isDeferred(): bool
```
albe, line 178: Also better than the getter that was only used to check if source is null

albe:
As said in our unicorn weekly I even think this could be considered a security issue, in which case we would need to target 4.3 - I'd take care if we agree on this.

@kitsunet @kdambekalns @robertlemke @bwaidelich @sorenmalling @daniellienert (sorry for the roundhouse mention)

albe:
ping y'all

Neos.Flow/Classes/ResourceManagement/PersistentResource.php
```diff
@@ -323,6 +387,18 @@ public function setSha1($sha1)
      */
     public function getMd5()
```
sorenmalling, line 388: We removed the md5 part of PersistenceResource #2112

albe: So since this targets 5.3 (as it is a BUGFIX), we'll need to have an immediate follow-up for master (or carefully resolve the removal in the up-merge conflict then). Thanks for the heads-up! :)

sorenmalling:
It's a solution to the original issue, and as you write

> PS: And yes, this is not an optimal solution. Let's keep improving this in master further to avoid the "deffered persisted resource" paradox, but this will inevitably be a breaking change then. 

Did our possibilities change with the introduction of PSR-15 - can we handle a GC of uploaded resources, that is up for deletion due to validation failure once the request "is done"?

albe:
> can we handle a GC of uploaded resources, that is up for deletion due to validation failure once the request "is done"?

The core issue is really that the resource is added for persistence before validation in the first place. It's non-trivial to "un-add" cleanly. And the fact that validation is only run directly before invoking the action method inside the controller, so there is not much room for "in between" layer that does this (after the action is too late, because the action might call `persistAll()`).

Neos.Flow/Classes/ResourceManagement/Storage/FileSystemStorage.php
```diff
@@ -131,12 +131,12 @@ public function getStreamByResourcePath($relativePath)
...
      *
      * @param callable $callback Function called after each iteration
-     * @return \Generator<Object>
+     * @return \Generator<StorageObject>
      */
     public function getObjects(callable $callback = null)
     {
         foreach ($this->resourceManager->getCollectionsByStorage($this) as $collection) {
-            yield $this->getObjectsByCollection($collection, $callback);
+            yield from $this->getObjectsByCollection($collection, $callback);
```
albe, line 139: Note: This is actually another bug looming in the code right now, no?

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
```diff
@@ -237,7 +247,7 @@ protected function handleHashAndData(array $source, PropertyMappingConfiguration
                 return $resource;
             }
 
-            if ($configuration->getConfigurationValue(ResourceTypeConverter::class, self::CONFIGURATION_IDENTITY_CREATION_ALLOWED) !== true) {
+            if ($configuration === null || $configuration->getConfigurationValue(ResourceTypeConverter::class, self::CONFIGURATION_IDENTITY_CREATION_ALLOWED) !== true) {
```
albe, line 250: Note: Maybe a missing configuration should not allow identity creation for the resource?

albe:
Rebased to keep things in sync

albe:
ping @kdambekalns @robertlemke @daniellienert @bwaidelich still would need someone of you to have a close look and approve or reject this. Maybe @kitsunet could remove the veto unless the concerns are still not alleviated

Neos.Flow/Classes/ResourceManagement/Storage/FileSystemStorage.php
albe: See #2167

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
kdambekalns: Right now that is the behaviour, right? So changing it would make this a breaking change?

albe: Well, right now if no configuration is given it just fails miserably. It's not a case that will happen through the property mapper (that one requires a configuration), only if typeconverters are invoked manually.

kdambekalns: Ok, in that case I am all for "deny unless explicitly configured".

albe:
```suggestion
            if ($configuration !== null && $configuration->getConfigurationValue(ResourceTypeConverter::class, self::CONFIGURATION_IDENTITY_CREATION_ALLOWED) !== true) {
```

albe:
Okay everyone - I had an idea for an alternative implementation, that will not introduce a new concept of a `PersistentResource` and be much less breaking. It is partly inspired by @sorenmalling's question here and @bwaidelich's `onStep()` callback implementation we used for PSR-15 middleware chain to hook into the steps to update our RequestHandler's `$request` instance. So props to them both.

Please take a look there - if we can agree on that, this can be closed for good.

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
albe: Fixed that separately with #2349

Neos.Flow/Classes/ResourceManagement/Storage/FileSystemStorage.php
```diff
@@ -158,7 +158,7 @@ public function getObjectsByCollection(CollectionInterface $collection, callable
             $object->setSha1($resource->getSha1());
             $object->setMd5($resource->getMd5());
             $object->setFileSize($resource->getFileSize());
-            $object->setStream(function () use ($resource) {
+            $object->setStream(static function () use ($resource) {
```
albe, line 161:
```suggestion
            $object->setStream(function () use ($resource) {
```

fcool:
Just some general thoughts here: We had this in several Pen-Tests reported to be an issue (importing a file into local storage, even if not validated).
Our approach has been to write a validating ResourceConverter, which can be extended in some way, to make the testers happy. 

In general I really dislike the behavior that the database gets modified by a TypeConverter - persisting changes should be a task the controller has full control of - so I would love a solution, which do not persist the resource already. But I understand why this seems "desired", when it comes to resources. But it leads to all sort of "unexpected" behavior. For example just receiving a "file object" you will only work with ONCE in your controllers action will be saved in storage and database and rot there unrelated. Something we should probably at least address in documentation.

albe:
Yeah, the best solution would be if a resource is not persisted at all (neither storage, nor DB) until the controller decides it wants to. The storage persistence could be solved by a onFlush listener and the controller would need to do `$this->persistenceManager->persist($uploadedResource)` (or persist the parent entity). That would be a breaking change though, but maybe something for Flow 8?

Anyway, regarding this BUGFIX - as mentioned in the other PR I'm leaning a bit more towards this solution right now, but with a few adjustments. And I agree to documenting the behaviour - maybe with a suggestion on how to deal with "temporary" file uploads properly. Hopefully after the 7.2 release I can find some time to continue the work on this.

kdambekalns (changes requested):
I favour the other PR as a fix!

PR #1935, opened 2020-03-14
FEATURE: Emit a signal when view is resolved

### LINKED ISSUE #1902: Fluid: Set view variables via signal/configuration
### Description

Similar to `{settings}` (set via processRequest in ActionController) I will like to open a discussion about how we could make it possible to add global variables to the view. A usecase could be as follows (and is in my current case):

 * Add the authenticated account to the view to have it available.

It can be done with a abstractcontroller and the initializeView method. Downside is, the creation of dependency and required inheritance of a single controller inside the code.

I suggest a `emitViewResolved` signal/slot inside the ActionController method, that pass the view object and give the possibility to work with the view.

Please bring feedback and ideas to this issue for a good and solid solution and descission :-)

sorenmalling:
In order for a package to interact with the rendering views available variables
a signal is being emitted upon resolving of the view.

The variables will be available in the view in the MVC context only

Resolves #1902

**What I did**

I added a Signal/Slot once the view has been resolved

**How to verify it**

Test functionality with the following package

https://github.com/sorenmalling/SorenMalling.EmitViewResolved

### DIFF
--- a/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Templating.rst
+++ b/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Templating.rst
@@ -393,6 +393,51 @@ You can pass arbitrary objects to the view, using ``$this->view->assign($identif
 from within the controller. See the above paragraphs about Object Accessors for details
 how to use the passed data.
 
+Passing data to the view from outside a controller
+--------------------------------------------------
+
+You can also pass data to the view from outside a controller. This can be useful for
+general data, that you want to be available without having to assign it in each action.
+
+Once the view is resolved inside the ``ActionController``, the signal ``viewResolved``
+is being emitted and you can add data.
+
+This is possible with the Signal/Slot dispatcher from your ``Package.php`` file::
+
+    <?php
+    namespace Vendor\Namespace;
+
+    use Neos\Flow\Core\Bootstrap;
+    use Neos\Flow\Mvc\Controller\ActionController;
+    use Neos\Flow\Mvc\View\ViewInterface;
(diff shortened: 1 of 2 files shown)


sorenmalling:
psalm errors is not related to this change

albe:
Psalm errors will be fixed with https://github.com/neos/flow-development-collection/pull/1937

sorenmalling:
@albe Can you restart the tracis-ci checks on this (#1937 is merged)

albe:
One thing we should probably add: a mention of this new signal in the docs other than the auto-generated technical reference. Maybe a `note` here: https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Templating.html

sorenmalling:
> One thing we should probably add: a mention of this new signal in the docs other than the auto-generated technical reference.

You got it - great idea :-)

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Templating.rst
```diff
@@ -393,6 +393,51 @@ You can pass arbitrary objects to the view, using ``$this->view->assign($identif
...
+         *
+         * @param Bootstrap $bootstrap The current bootstrap
+         * @return void
+         */
+        public function boot(Bootstrap $bootstrap)
+        {
+
+            $dispatcher = $bootstrap->getSignalSlotDispatcher();
+
+            $dispatcher->connect(ActionController::class, 'viewResolved', function (ViewInterface $view) {
```
albe:
```suggestion
            $dispatcher->connect(ActionController::class, 'viewResolved', static function (ViewInterface $view) {
```

PR #1937, opened 2020-03-15
TASK: Fix psalm errors

albe:
Make psalm happy again


Neos.Cache/Classes/Backend/PdoBackend.php
```diff
@@ -206,11 +206,12 @@ public function get(string $entryIdentifier)
 
         $statementHandle = $this->databaseHandle->prepare('SELECT "content" FROM "' . $this->cacheTableName . '" WHERE "identifier"=? AND "context"=? AND "cache"=?' . $this->getNotExpiredStatement());
         $statementHandle->execute([$entryIdentifier, $this->context(), $this->cacheIdentifier]);
+        /** @var false|string|null $fetchedColumn */
```
albe: The column `content` is nullable in the schema, so this can return null.

albe:
Having to use `psalm-*` annotations is so-so, but after all it provides a semantic benefit. Would prefer to have just `@immutable` but let's embrace that tool fully where it makes sense (mutability and maybe generic semantics are such cases).

albe:
Not needed, but I think in this case it is more benefit in adding this one annotation than silencing the error. Psalm can this way detect ill-usage of `Now` in the places it is used.

We can of course baseline this case instead too.

daniellienert:
@albe maybe add that explanation to the commit that introduces it

albe:
Done @daniellienert :)

PR #1939, opened 2020-03-18
!!! FEATURE: Introduce AccountInterface, Roles API and Account ValueObjects

### LINKED ISSUE #1077: Separate hard dependency on Account class and introduce a interface

sorenmalling:
**What I did**
Introducing a AccountInterface to avoid hardcoding authentication to only a Neos.Flow defined model.

This also opened for a streamlining of the AccountInterface object, so the following ValueObjects was introduced

 * AccountIdentifier
 * AuthenticationProviderName
 * CredentialsSource

**How I did it**
Created a AccountInterface, have the current Account model implement it to keep the backward compatibility.

All possible places we refere to AccountInterface and `(string)` cast the new ValueObjects

**How to verify it**

Test it together with the `task/adjust-to-account-interface` branch of Neos CMS and look at the changes made to the codebase

Resolves #1077


Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
```diff
@@ -104,14 +104,11 @@ public function authenticate(TokenInterface $authenticationToken)
             return;
         }
 
-        if ($this->hashService->validatePassword($credentials['password'], $account->getCredentialsSource())) {
-            $account->authenticationAttempted(TokenInterface::AUTHENTICATION_SUCCESSFUL);
+        $authenticationToken->setAccount($account);
```
albe: This is the important subtle change I guess. I don't think it is a problem to move this before the `validatePassword()`, but I'd like to hear the others' opinion.

Neos.Flow/Classes/Security/Policy/Roles.php
```diff
@@ -0,0 +1,94 @@
...
+ * @Flow\Proxy(false)
+ */
+final class Roles implements \JsonSerializable, \IteratorAggregate, \Countable
+{
+
+    /**
+     * @var Role[]
+     */
+    private $roles;
+
```
albe: Needs docblocks in all the methods

albe (commented):
Left a few comments, but this is going in very good direction

Neos.Flow/Classes/Package.php
```diff
@@ -122,6 +124,20 @@ public function boot(Core\Bootstrap $bootstrap)
             if ($session->isStarted() && !$token instanceof SessionlessTokenInterface) {
                 $session->renewId();
             }
+
+            if ($token->getAccount() instanceof Account) {
+                $token->getAccount()->authenticationAttempted(TokenInterface::AUTHENTICATION_SUCCESSFUL);
```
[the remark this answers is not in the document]
bwaidelich: No please don't. I think it was a big mistake to track authentication attempts on the account entity itself (leading to a huge amount of mutations even if it wasn't used at all)

Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
bwaidelich: Why is this change required?

bwaidelich: This is a dangerous change IMO because it sets the account even if the token didn't validate.
Why not just move it into the `if ($this->hashService->validatePassword(...` block?

Neos.Flow/Classes/Package.php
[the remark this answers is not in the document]
bwaidelich: I would even suggest to deprecate the respective methods in `Account` and suggest to store those attempts in a separate table, if needed

bwaidelich (commented):
I did a first quick round of review and overall I really really like the change and I hope that we can get it in the upcoming release!
Apart from the issues Alex and me commented upon, we should probably check the docs if they are still accurate and add a phrase or two about the new extensibility.
Also the PR description should explain the new feature and especially why it is breaking – since that ends up in the change log

Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
sorenmalling: @bwaidelich It is require to pass the account (already found by the given accountIdentifier) to the signal for failedAuthentication. This is to remove the hardcoded "authenticationFailed" functionality of the Neos.Flow Account object and move it to a signal, so other providers is also able to handle the case of a authentication failing.

Neos.Flow/Classes/Package.php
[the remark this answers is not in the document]
sorenmalling: @bwaidelich Your comment here, is why it is needed to pass the failed account to the token, even though it fails (described in comment below :) )

sorenmalling:
@bwaidelich I will focus on updating the docs once we have the functionality straight . And provide a test package :)

Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
bwaidelich: Ah sure, I see, thanks for the explanation

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
```diff
@@ -242,21 +242,23 @@ public function logout(): void
      * Signals that the specified token has been successfully authenticated.
      *
      * @param TokenInterface $token The token which has been authenticated
+     * @param AuthenticationProviderInterface $provider The provider that was involved authenticating the token
      * @return void
      * @Flow\Signal
      */
-    protected function emitAuthenticatedToken(TokenInterface $token): void
+    protected function emitAuthenticatedToken(TokenInterface $token, AuthenticationProviderInterface $provider): void
```
sorenmalling, line 249: Will be breaking other implementation if we dont set default to `null` for $provider?

Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
```diff
@@ -100,10 +110,9 @@ public function authenticate(TokenInterface $authenticationToken)
             return;
         }
 
-        $authenticationToken->setAccount($account);
```
sorenmalling: ^^If we do not set the account here, we wil not have it, if the validatePassword method doesn't return true.

Neos.Flow/Classes/Security/TransientAccount.php
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * An account that is not persisted in the database
+ *
+ * This can be used for scenarios where no corresponding entity is stored in the local database, for
+ * example when using a Single Sign-On service.
+ *
+ * @Flow\Proxy(false)
+ * @api
+ */
+final class TransientAccount implements AccountInterface
```
sorenmalling, line 29: @bwaidelich We can/should use this inside the FileBasedSimplekeyProvider, where we create a Account object - that will create fine example for, how to use it :-D

Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
```diff
@@ -12,46 +12,46 @@
...
+use Neos\Flow\Security\Exception as SecurityException;
 use Neos\Flow\Security\Exception\UnsupportedAuthenticationTokenException;
 
 /**
  * An authentication provider that authenticates
  * Neos\Flow\Security\Authentication\Token\UsernamePassword tokens.
  * The accounts are stored in the Content Repository.
  */
-class PersistedUsernamePasswordProvider extends AbstractProvider
+final class PersistedUsernamePasswordProvider extends AbstractProvider
```
albe, line 35: So, breaking it real hard if we break things, eh? :)

albe (approved):
Looks good but breaking. What if we remove the `final` and keep the `accountRepository` property so it's not breaking and merge it into master now?

PS: Talking about the change in 3f27a7efeb31ae599f1de74dc9f9d51c56809fa4 here

Neos.Flow/Classes/Security/Authentication/AuthenticationProviderManager.php
bwaidelich: But it's protected so only the `AuthenticationProviderManager` can call it!?

Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
bwaidelich: Right. But that's exactly the intention, no? => We don't want the token to contain the account if it could validate the password

Neos.Flow/Classes/Security/TransientAccount.php
bwaidelich: Yes, good point, added it with 2e2a7e07a6475ce49d492af4978cee12b5d25ed8 (not tested yet)

Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
bwaidelich: Oops, that must have been a reflex => aabb8f6dbed5a56e4f3a49b8f8c38523872edd4e

albe:
Ok, so one big point is still open: How do we deal with the major breakyness of this change right now?

bwaidelich (changes requested):
Really annoying but the `TransientAccount` is not serializable and I'm not sure how to solve this without making the change even bigger..

The issue is that `Role` is used as a Value Object (representing the assigned role) _and_ entity (knowing its parent roles etc) throughout the system.
A clean solution would be to add a `RoleIdentifier` VO and rename the `Roles` VO to `RoleIdentifiers` and to move the part that checks if an account has a certain role (including parent roles) out of the `Account`.

`Account` still works because it has the `PolicyService` injected and internally stores everything as simple types, only converting to the Value Objects in the getters..
But even if we skip the `TransientAccount`, people will have the serialization issue when creating custom implementations..

Neos.Flow/Classes/Security/Policy/RoleIdentifiers.php
```diff
@@ -0,0 +1,116 @@
...
+     * @param string[] $roleIdentifiers
+     * @return static
+     */
+    public static function fromArray(array $roleIdentifiers): self
+    {
+        $processedRoleIdentifiers = [];
+        array_walk($roleIdentifiers, static function ($roleIdentifier) use (&$processedRoleIdentifiers) {
+            $processedRoleIdentifiers[(string)$roleIdentifier] = $roleIdentifier;
+        });
+        return new static($processedRoleIdentifiers);
```
bwaidelich, line 54: I would suggest to turn this into

```suggestion
        foreach ($roleIdentifiers as $roleIdentifier) {
            if (!is_string($roleIdentifier)) {
                throw new \InvalidArgumentException(sprintf('Expected an array of strings, got: %s', gettype($roleIdentifier)), 1602174016);
            }
            $processedRoleIdentifiers[$roleIdentifier] = $roleIdentifier;
        }
        return new static($processedRoleIdentifiers);
```

..easier to read (IMO) and type safe

Neos.Flow/Classes/Security/Policy/RoleIdentifiers.php
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * @param $offset
+     * @return mixed|Role|null
+     */
+    public function offsetGet($offset)
+    {
+        return $this->roles[$offset] ?? null;
+    }
+
```
bwaidelich, line 98: `offsetGet()` is not part of any of the interfaces, I would suggest to remove it
```suggestion
```

Neos.Flow/Classes/Security/Policy/RoleIdentifiers.php
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * @param string $roleIdentifier
+     * @return RoleIdentifiers
+     */
+    public function withRoleIdentifier(string $roleIdentifier): RoleIdentifiers
+    {
+        return new self(array_merge($this->roleIdentifiers, [(string)$roleIdentifier => $roleIdentifier]));
```
bwaidelich, line 105: This is a tiny bit simpler and faster:

```suggestion
        $newRoleIdentifiers = $this->roleIdentifiers;
        $newRoleIdentifiers[$roleIdentifier] = $roleIdentifier;
        return new self($newRoleIdentifiers);
```

In addition I would suggest to move those two methods underneath the (static) constructors since they are also a special kind of constructor

Neos.Flow/Classes/Security/Policy/RoleIdentifiers.php
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * @param string $roleIdentifier
+     * @return RoleIdentifiers
+     */
+    public function withoutRoleIdentifier(string $roleIdentifier): RoleIdentifiers
+    {
+        return new self(array_diff($this->roleIdentifiers, [(string)$roleIdentifier => $roleIdentifier]));
```
bwaidelich, line 114: similar to above:
```suggestion
        $newRoleIdentifiers = $this->roleIdentifiers;
        unset($newRoleIdentifiers[$roleIdentifier]);
        return new self($newRoleIdentifiers);
```

bwaidelich (commented):
@sorenmalling I finally got around taking another look at your last addition (I was sure I already left some comments – I must have been distracted and closed the browser).

Looking great already, I just left some nitpick comments on the new `RoleIdentifiers` VO.

This PR has some conflicts in the meantime. Would you be up for rebasing it to current master?

sorenmalling:
> This PR has some conflicts in the meantime. Would you be up for rebasing it to current master?

I will bring it up to date 👍

bwaidelich:
@sorenmalling anything I can help to bring this one up to date?
It would be cool to have the breaking changes in as soon as possible in order to given them some time to "ripe" :)

sorenmalling:
@bwaidelich Let me spend some time right now to get the conflicts removed and tests green :) Update coming in 1 hour 👍

sorenmalling:
Conflicts solved via the web ui - will get my local environment up to date again and go through test results from travis and compare to local results 👍

robertlemke:
Does the `CredentialsSource` value actually _have_ to be a string?

Right now, there are some authentication providers which set it to something else. For example, the OIDC package uses a JWT object as its credentials source ([see code](https://github.com/flownative/flow-openidconnect-client/blob/master/Classes/Authentication/OpenIdConnectProvider.php#L165)). Sure, that could be serialized, but then would cause some hassle at other ends.

If there's a good reason to force this to a string, let's do that. But let's be sure about that …

bwaidelich:
> Does the CredentialsSource value actually have to be a string?

It has to be _representable_ as a string at least or it could not be persisted as is.
`$account->setCredentialsSource($someObject)` actually only works because the setter is missing a corresponding type hint, the field itself is a [string](https://github.com/neos/flow-development-collection/blob/7de086efc956ee8de523aeb40ff2cca72410e103/Neos.Flow/Classes/Security/Account.php#L46-L50), and since it maps 1:1 to a column in the database, a complex type would break persistence.

Then again, we could of course introduce an interface for the `CredentialSource` but what methods would you expect there?

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -34,7 +34,7 @@ abstract class TypeHandling
     /**
      * @var array
      */
-    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class];
+    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class, \IteratorAggregate::class];
```
bwaidelich, line 37: If we change this we should not add the specific `IteratorAggregate` interface but the `Traversable` interface that is implemented by all the other classes already:
```suggestion
    protected static $collectionTypes = ['array', \Traversable::class];
```

I'm not 100% sure about possible side effects of changing this here.
A more b/c alternative might be to change
```php
} elseif ((is_object($value) && !TypeHandling::isCollectionType(get_class($value))) && !is_array($value)) {
```
to
```php
} elseif (!is_iterable($value)) {
```
in the [CollectionValidator](https://github.com/neos/flow-development-collection/blob/8b69724089ed4a55a37b2fe383c7f510f6272a5d/Neos.Flow/Classes/Validation/Validator/CollectionValidator.php#L55)

Neos.Flow/Tests/Functional/Security/Authorization/Privilege/Entity/Doctrine/ContentSecurityTest.php
```diff
@@ -81,7 +81,9 @@ protected function setUp(): void
      */
     public function administratorsAreAllowedToSeeHiddenRestrictableEntities()
     {
-        $this->authenticateRoles(['Neos.Flow:Administrator']);
+        $account = $this->authenticateRoles(['Neos.Flow:Administrator']);
+        $account->setAccountIdentifier('Administrator');
+        $account->setAuthenticationProviderName('SomeProvider');
```
bwaidelich, line 86: Why is this required now?

Neos.Flow/Tests/Functional/Security/Authorization/Privilege/Entity/Doctrine/ContentSecurityTest.php
```diff
@@ -235,7 +243,7 @@ public function customersCannotSeeOthersRestrictableEntites()
         $otherAccount = new Security\Account();
         $otherAccount->setAccountIdentifier('othersAccount');
         $otherAccount->setAuthenticationProviderName('SomeProvider');
-        $this->persistenceManager->add($ownAccount);
+        $this->persistenceManager->update($ownAccount);
```
bwaidelich, line 246: Why did this change?
Since `$ownAccount` is a new instance `$this->persistenceManager->add($ownAccount)` makes more sense and it this needs to be adjusted it's probably due to some breaking change in behavior

sorenmalling: Because it's alreadyt added in the call for `authenticateRoles` - try and change it back to `add` and you will see the exception of "this is not a new object - use update instead"

Neos.Flow/Tests/Functional/Security/Authorization/Privilege/Entity/Doctrine/ContentSecurityTest.php
sorenmalling: With the getAccountIdentifier method that returns a AccountIdentifier object, we create it from the string property `$accountIdentifier` - but it's "null" in this case. Causing the persistence to throw a exception like "argument must be string, null given"
What is wrong here in the test cases, is that we try to persist a model, with required properties (NotEmpty) that has no values. But since the "validation" is using the `get`method to create a value, the before mentioned (null given) exception, is causing the issue

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
sorenmalling: The changes in TypeHandling is only untill this PR is pushed 
https://github.com/neos/flow-development-collection/pull/2202

albe: @bwaidelich see https://github.com/neos/flow-development-collection/pull/2202#discussion_r515609751 - had the exact same idea/concern :) Changing the `CollectionValidator` to `is_iterable` might be a sane alternative, if we are unsure about how many places uses `isCollection()` to implicitly also check for `Countable` and/or `ArrayAccess`

bwaidelich:
As discussed with @sorenmalling we'll try the following next steps:
* remove `hasRole()` from the `AccountInterface` and maybe introduce some method to the `PolicyService` that "expands" an instance of `RoleIdentifiers` to retrieve all role ids including parent roles
* Adjust core and tests so that they refer to the `AccountInterface` (or `TransientAccount` when creating new instances) where possible.
* Mark `Account` deprecate in favor of a new `PersistedAccount` that just implements the interface and has proper constructors

bwaidelich:
Some input from todays session at the Sprint:

* Updated AccountInterface PR coming up for Flow 8
* WIP on moving route mapping to persistent cache instead of database
* Question: Account model without ORM, but DBAL? (or a 3rd one?)
  * Idea: keep the existing Account as is, and create a DBAL "repository" to save/load the immutable one (=> rename it from TransientAccount to s.th. more general)
  * !!! maybe de-importantize the ORM (explain why we want that)

PR #1942, opened 2020-03-19
BUGFIX: Allow numeric array body in request

albe:
When sending e.g. a JSON request with an array as body (["foo","bar"]), Flow would throw an exception "Invalid argument name (must be a non-empty string)." when visiting the array key "0" due to the `empty()` check.
This change fixes that by backporting a fix that is already in place in Flow 6.0+ which only checks for empty string.

Such an array can not be mapped to a controller argument, but you can still technically access the argument from the controller via `$this->arguments->getArgument("0")` or the whole body via
`$this->request->getHttpRequest()->getParsedBody()`.

Replaces #1913

### DIFF
--- a/Neos.Flow/Classes/Mvc/DispatchComponent.php
+++ b/Neos.Flow/Classes/Mvc/DispatchComponent.php
@@ -98,7 +98,9 @@ class DispatchComponent implements ComponentInterface
         $arguments = $httpRequest->getArguments();
 
         $parsedBody = $this->parseRequestBody($httpRequest);
-        $httpRequest = $httpRequest->withParsedBody($parsedBody);
+        if ($parsedBody !== null) {
+            $httpRequest = $httpRequest->withParsedBody($parsedBody);
+        }
         if (is_array($parsedBody) && $parsedBody !== []) {
             $arguments = Arrays::arrayMergeRecursiveOverrule($arguments, $parsedBody);
         }
--- a/Neos.Flow/Classes/Mvc/DispatchComponent.php
+++ b/Neos.Flow/Classes/Mvc/DispatchComponent.php
@@ -98,11 +98,10 @@ class DispatchComponent implements ComponentInterface
         $arguments = $httpRequest->getArguments();
 
         $parsedBody = $this->parseRequestBody($httpRequest);
-        if ($parsedBody !== []) {
+        $httpRequest = $httpRequest->withParsedBody($parsedBody);
+        if (is_array($parsedBody) && $parsedBody !== []) {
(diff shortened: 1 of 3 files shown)


Neos.Flow/Classes/Mvc/DispatchComponent.php
```diff
@@ -98,12 +98,11 @@ protected function prepareActionRequest(ComponentContext $componentContext)
         $arguments = $httpRequest->getArguments();
 
         $parsedBody = $this->parseRequestBody($httpRequest);
-        if ($parsedBody !== []) {
+        $httpRequest = $httpRequest->withParsedBody($parsedBody);
```
albe: We always want to set the parsed body, even if it is an empty array, a simple string or even null.

Neos.Flow/Classes/Mvc/DispatchComponent.php
```diff
@@ -125,19 +124,20 @@ protected function prepareActionRequest(ComponentContext $componentContext)
...
      * @param HttpRequest $httpRequest
-     * @return array
+     * @return null|array|string|integer
      */
     protected function parseRequestBody(HttpRequest $httpRequest)
     {
         $requestBody = $httpRequest->getContent();
         if ($requestBody === null || $requestBody === '') {
-            return [];
+            return $requestBody;
```
albe: Now, this is a slight behavorial change. But it is more correct. Then again, the MediaTypeConverter still behaves wrongly by treating anything he can't convert as `[]` empty array (see code comment below).

Neos.Flow/Classes/Property/TypeConverter/MediaTypeConverter.php
```diff
@@ -50,7 +50,7 @@ class MediaTypeConverter extends AbstractTypeConverter implements MediaTypeConve
      * @param string $targetType must be "array"
      * @param array $convertedChildProperties
      * @param PropertyMappingConfigurationInterface $configuration
-     * @return array
+     * @return array|string|integer Note that this TypeConverter may return a non-array in case of JSON media type, even though he declares to only convert to array
```
albe: Will be "fixed" in master by deprecating this "TypeConverter" and putting something else (a "Serializer"?) in place, that is not restricted to one target type.

Neos.Flow/Classes/Mvc/DispatchComponent.php
albe:
```suggestion
        if ($parsedBody !== null) {
            $httpRequest = $httpRequest->withParsedBody($parsedBody);
        }
```
Or maybe not if it is `null`

albe:
Green!

Issue #1944, opened 2020-03-21
Feature: Allow for easy overriding of extracting credentials in UsernamePassword token


sorenmalling:
When implementing own username/password token based authentication, I have different naming of fields or requirements for arguments. This causes me to reimplement (copy/paste) large parts of the UsernamePassword token, since the attributes are hardcoded inside the updateCredentials method.

The extracting of credentials could be moved to a separate method, that can then be overridden by a own implementation

DavidSporer:
I'm not sure if this is the right place but I just noticed another issue that might be related to that:
When you want to use the same authentication provider for regular authentication with a login form and also use that for Basic Auth then the regular WebRedirect won't work since Basic Auth should return the WWW-Authenticate basic header.
Is this the right place to ask or has someone solved such an issue already?

albe:
@bwaidelich any idea on what David said? I have no clue t.b.h.

bwaidelich:
@DavidSporer @albe that's another (IMO unrelated) issue that EntryPoints are always evaluated so you need two provider configurations in that case. It would be nice to be able to use the "request patterns" for those, too, like:
```yaml
security:
  authentication:
    providers:
      DefaultProvider:
        provider: PersistedUsernamePasswordProvider
        entryPoint: 'WebRedirect'
        entryPointOptions: # ...
        filters:
          'Some.Package:CustomPattern':
            pattern:  'Acme\MyPackage\Security\MyOwnRequestPattern'
            patternOptions: # ...
```

I thought there was already a ticket for that, but I can't find it.. Would you create a new one?

DavidSporer:
@bwaidelich @aberl yes but that wouldn't solve the issue I have, I assume?
Currently the name of a provider name is saved in the authenticationprovidername field of an account.
If I understand that correct it means I can set up two different Authentication providers but the need to have different names and will therefore result in two different credentialssources/logins.

Just one idea I had right now: would it be possible to create a Provider that uses the existing credentials of another provider?
Something like this:
`
security:
  authentication:
    providers:
      DefaultProvider:
        provider: PersistedUsernamePasswordProvider
      BasicAuthProvider:
        provider: MyHttpBasicAuthProviderThatValidatesDefaultProviderCredentials
`

If I'm not mistaken I just need to use something like this in the authenticate function of the custom provider:
$account = $accountRepository->findActiveByAccountIdentifierAndAuthenticationProviderName($credentials['username'],
                'DefaultProvider');
[...]

PR #1945, opened 2020-03-21
FEATURE: Allow for overriding of credentials in UsernamePassword token

sorenmalling:
To avoid reimplementing the whole UsernamePassword token
to get credentials from different fieldnames or request
arguments, the extraction are moved to a own method

Resolves: #1944

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
+++ b/Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
@@ -45,15 +45,35 @@ class UsernamePassword extends AbstractToken
             return;
         }
 
-        $arguments = $actionRequest->getInternalArguments();
+        $credentials = $this->extractCredentialsFromRequest($actionRequest);
+
+        if (!empty($credentials['username']) && !empty($credentials['password'])) {
+            $this->credentials['username'] = $credentials['username'];
+            $this->credentials['password'] = $credentials['password'];
+            $this->setAuthenticationStatus(self::AUTHENTICATION_NEEDED);
+        }
+    }
+
+    /**
+     * Extract credentials from the request and pass it on to the token
+     *
+     * This method, makes it convenient for your own implementation to
+     * override the parameters, either based on field name or any other
+     * request arguments or requirements you might have
+     *
+     * @param ActionRequest $request
+     * @return array
+     */
+    protected function extractCredentialsFromRequest(ActionRequest $request)
+    {
+        $arguments = $request->getInternalArguments();
         $username = ObjectAccess::getPropertyPath($arguments, '__authentication.Neos.Flow.Security.Authentication.Token.UsernamePassword.username');
         $password = ObjectAccess::getPropertyPath($arguments, '__authentication.Neos.Flow.Security.Authentication.Token.UsernamePassword.password');
 
-        if (!empty($username) && !empty($password)) {
-            $this->credentials['username'] = $username;
-            $this->credentials['password'] = $password;
-            $this->setAuthenticationStatus(self::AUTHENTICATION_NEEDED);
-        }
+        return [
+            'username' => $username,
+            'password' => $password
+        ];
     }
 
     /**


bwaidelich:
I don't really understand the need of this change to be honest..
Probably that's because personally I really despise the use of inheritance where it can be avoided.

Is

```php
class MyToken extends UsernamePassword
{
    protected function extractCredentialsFromRequest(ActionRequest $request)
    {
        $arguments = $request->getInternalArguments();
        $username = ObjectAccess::getPropertyPath($arguments, '__authentication.My.Custom.username');
        $password = ObjectAccess::getPropertyPath($arguments, '__authentication.My.Custom.password');

        return [
            'username' => $username,
            'password' => $password
        ];
    }
}
```

so much easier than:

```php
class MyToken extends AbstractToken
{
    /**
     * @Flow\Transient
     */
    protected $credentials = ['username' => '', 'password' => ''];

    public function updateCredentials(ActionRequest $actionRequest): void
    {
        $httpRequest = $actionRequest->getHttpRequest();
        if ($httpRequest->getMethod() !== 'POST') {
            return;
        }

        $arguments = $actionRequest->getInternalArguments();
        $username = ObjectAccess::getPropertyPath($arguments, '__authentication.My.Custom.username');
        $password = ObjectAccess::getPropertyPath($arguments, '__authentication.My.Custom.password');

        if (!empty($username) && !empty($password)) {
            $this->credentials = ['username' => $username, 'password' => $password];
            $this->setAuthenticationStatus(self::AUTHENTICATION_NEEDED);
        }
    }
}
```

?
(not a rhetoric question. I see the increased complexity, but I doubt that it's worth that change)

sorenmalling:
> Is [...] so much easier than [...]

I will of course say yes, or I wouldn't have taken the time to create it 😄 

I've found my self do this a number of times, since PersistedUsernamePasswordProvider requires it to be a UsernamePasswordToken instance.
And therefore, I can not extends AbstractToken without also creating a copy of my own PersistedUsernamePasswordProvider

bwaidelich:
> I've found my self do this a number of times

Allow me to challenge this (not to question your efforts, but to understand the intentions):
Why do you need to change the field names at all?

bwaidelich:
..and to follow-up on this:
I'm totally up for making this extensible because I can think of a couple of reasons why you would want to change this (i.e. for API requests).
I just doubt that inheritance is the best option for this.
It might be the only _viable_ but maybe we could do something along the way of:
```
final class UsernamePassword extends AbstractToken
{
    public function __construct(string $postFieldPrefix)
    {
        // ...
    }
    // ...
}
``` 

(or an `array $options` constructor argument instead)

sorenmalling:
>  Why do you need to change the field names at all?

I dislike the usage of internalarguments (a Flow specific thing) instead of plain argument name. Also in terms of API authentication, i can reuse the token

albe:
I totally get why one wants to change the super-verbose internal argument field names, but I'm too inclined to say let's just make those hardcoded strings easier to change rather than make users extend this class.

@bwaidelich We can't `final` the `UsernamePassword` because we have `class UsernamePasswordHttpBasic extends UsernamePassword` ;)
Also, a constructor argument will require us to change the token creation inside `TokenAndProviderFactory`, where it's currently just a `new $tokenClassName()`. So either we make this part a whole lot more complex, or we use settings injection on the token.

bwaidelich:
> We can't final
We could if we would not make `UsernamePasswordHttpBasic` extend from `UsernamePassword` (it overrides the only method anyways). But that wasn't my point, I didn't mean to suggest to make it `final` (that was just the power of habit :)

> a constructor argument will require us to change the token creation inside TokenAndProviderFactory

Good point. That would be my favorite solution in fact:

in `TokenAndProviderFactory`:

```diff
--- a/Neos.Flow/Classes/Security/Authentication/TokenAndProviderFactory.php
+++ b/Neos.Flow/Classes/Security/Authentication/TokenAndProviderFactory.php
@@ -156,7 +156,7 @@ class TokenAndProviderFactory implements TokenAndProviderFactoryInterface
                     continue;
                 }

-                $tokenInstance = new $tokenClassName();
+                $tokenInstance = new $tokenClassName($providerConfiguration['tokenOptions'] ?? []);
                 $tokenInstance->setAuthenticationProviderName($providerName);
                 $this->tokens[] = $tokenInstance;
                 break;
```

in the token implementation:

```php
class UsernamePassword extends AbstractToken
{
    public function __construct(array $options)
    {
        $this->usernamePostField = $options['usernamePostField'] ?? '__authentication.Neos.Flow.Security.Authentication.Token.UsernamePassword.username';
        $this->passwordPostField = $options['passwordPostField'] ?? '__authentication.Neos.Flow.Security.Authentication.Token.UsernamePassword.password';
    }
    // ...
}

and in the `Settings.yaml`:

```yaml
Neos:
  Flow:
  security:
    authentication:
      providers:
        'My.Package:SomeProvider':
          provider: 'PersistedUsernamePasswordProvider'
          token: 'UsernamePassword'
          tokenOptions:
            usernamePostField: 'my.custom.username'
            passwordPostField: 'my.custom.username'
```

Wouldn't that work?

> or we use settings injection on the token

Since tokens are serialized into the user session I would not suggest to do that!

bwaidelich:
FYI: Just to make clear: I won't vote this down or even block it of course. Feel free to merge if you agree to the change!

albe:
> Since tokens are serialized into the user session I would not suggest to do that!

Also a good point :) So definitely no injected config stuff

> Wouldn't that work?

I guess it would. Would just make it more complicated to create such a token yourself, because you'd need to somehow get hold of the provider config then.

bwaidelich:
> more complicated to create such a token yourself

What would be a scenario for this and wouldn't that even make more sense that options could be overridden in those cases? i.e. for testing. Or where else would someone create a token instance?

albe:
> > more complicated to create such a token yourself
> 
> What would be a scenario for this and wouldn't that even make more sense that options could be overridden in those cases? i.e. for testing. Or where else would someone create a token instance?

Yeah, was mostly thinking about tests, though possibly there are also other edge-case scenarios where you would want to create one. I guess the biggest I issue I have then, is that the options are only specified in the settings.yaml, though this can probably be helped with clearly documenting the options in the constructor. I could live well with that.

@sorenmalling Would you be able to adjust to the suggestion of Bastian with the `TokenAndProviderFactory` change and settings? I'd then still add that to 6.2 release :)

bwaidelich:
@sorenmalling @albe I didn't mean to ask for a rework. Let me test if my approach really works – and if it does, I'll push it to a new PR and we can decide on that

sorenmalling:
@bwaidelich @albe 

**Question**
Current implementation uses `getInternalArguments` for arguments - but this is not the case in all tokens - would this also need a "toggle", or can we reach the internal arguments?

I find the Settings.yaml idea conceptual "wrong direciton". `usernamePostField` is not "correct", as the implementation can come in many forms: GET method, a property of a json object, API token, etc.  You might have fallback, depending on a header etc.
The overriding was to be able to get the arguments needed by the request, if you needed something different than the default arguments name :) 

Personal note: This PR is going far away from what I wanted to achieve and seems more complicated for the usecase I was trying to solve :) I find no value in the Settings.yaml implementation for above mentioned usecases, which was what was tried to be solved at first.

bwaidelich:
@sorenmalling 

I gave my idea a quick shot: #1973 

It's not a lot of code and works nicely.. Only caveat: The options are serialized into the session (if not using a sessionless token). So if you change the options configuration, those serialized tokens will still contain the old options.. But this is true for all provider configuration.

> would this also need a "toggle", or can we reach the internal arguments?

Those internal arguments are a pain indeed, I just merged them since we want to allow property paths anyways.

> usernamePostField is not "correct", as the implementation can come in many forms: GET method, a property of a json object, API token, etc

I think you misunderstood my suggestion. The idea wasn't to introduce a global setting that has to be valid for all tokens, but to be able to change the argument names for a _specific_ case.
It would basically just be:

```yaml
 #...
        'My.Package:SomeProvider':
          provider: 'PersistedUsernamePasswordProvider'
          token: 'UsernamePassword'
          tokenOptions:
            usernamePostField: 'my.custom.username'
            passwordPostField: 'my.custom.username'
```
instead of having to write a custom implementation and do
```yaml
        'My.Package:SomeProvider':
          provider: 'PersistedUsernamePasswordProvider'
          token: 'My\Custom\Implementation'
```

And since the `UsernamePassword` token only respects `POST` requests, the option names would be fine here IMO.

> I find no value in the Settings.yaml implementation for above mentioned usecases

I can't really follow you to be honest..
I think it's totally fine to create your custom token implementations and I do so a lot. But having to create a custom class only to change the argument names seems overcomplicated to me..

But there's a more profound issue with the "extends approach":
It basically makes the `extractCredentialsFromRequest()` method part of the public API and makes the class more susceptible to breaking changes

sorenmalling:
> > I find no value in the Settings.yaml implementation for above mentioned usecases
> 
> I can't really follow you to be honest..

You miss the important part of my quote: What was I trying to solve with this patch :)
The arguments can come in different forms. GET Method, a payload you have to extract from or similar (see my previous comment with the examples). 
The token's only responsible is to hold the credentials that is passed on to a provider supporting the token.

I don't see the UsernamePassword token as a token that **only supports POST method**. I see it a a token that gathers username and password credentials, to pass on to a supporting provider. And how I get those credentials can differ from project to project. 

That is the reason why I started this patch, to make a simple entry point for creating own token, with same credentials structure and using the same existing provider.

A different solution could be a mix of both. Being able to override for extending the extraction of arguments. And have the Settings.yaml with a sane default, that can be overridden.

How does that sound to you? 

> But there's a more profound issue with the "extends approach":
It basically makes the extractCredentialsFromRequest() method part of the public API and makes the class more susceptible to breaking changes

IMO Every new change or feature we bring makes it a candidate  for future breaking change, but it also give us possibilities to make greater solution in the even longer run :-)

bwaidelich:
@sorenmalling I'm afraid we talk cross purpose. Are you up for a quick call sometimes today? We also have some more input on the `AccountInterface` change (#1939) we'd like to dicsuss with you briefly

albe:
I guess we can close this in favor of #1975 ?

bwaidelich:
yep

bwaidelich:
@sorenmalling sorry, I didn't mean to close your PR – I confused it with my one..

PR #1948, opened 2020-03-23
BUGFIX: Add samesite cookie configuration

nlx-sascha:
**What I did**
Adds new cookie setting allow to configure SameSite to 'strict', 'lax' and 'none'.
Be careful with SameSite=none as there is a webkit bug (https://bugs.webkit.org/show_bug.cgi?id=198181) inadvertently treated this as SameSite=strict
This change does not effect the current behavior. When you don't set the configuration `samesite` in `settings.yaml` the cookie attribute is ignored.

- [ X ] Code follows the PSR-2 coding style
- [ X ] Tests have been created, run and adjusted as needed
- [ X ] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)


nlx-sascha:
Some additional resources to the topic:
https://chromestatus.com/feature/5088147346030592
https://web.dev/samesite-cookies-explained/
https://bugs.webkit.org/show_bug.cgi?id=198181

mficzel:
Thanks for this, 

i like it a lot aswell but i have a question: Is there really a benefit in defining the possible values for samesite hard and in connecting the logic with the secure attribute?

I understand that this reduces the adjustments that have to be done to a neos-update but it adds hidden magic. I see no way of this doing harm so it may be justified when there is a good reason behind.

nlx-sascha:
> i like it a lot aswell but i have a question: Is there really a benefit in defining the possible values for samesite hard and in connecting the logic with the secure attribute?
Yes, Cookies with SameSite=None must also specify Secure, meaning they require a secure context.
I run into this issue in development and was wondering why it doesn't work, until I found the article on web.dev and found the not that secure is required.
For me it would be find to throw an Exception instead of adding this kind of magic.

Neos.Flow/Tests/Unit/Http/CookieTest.php
```diff
@@ -255,6 +255,60 @@ public function isHttpOnlyReturnsHttpOnlyFlag()
...
+        $this->assertSame(Cookie::SAMESITE_NONE, $cookie->getSameSite());
+    }
+
+    /**
+     * @test
+     */
+    public function SameSiteNoneEnablesSecure()
+    {
+        $cookie = new Cookie('foo', 'bar', 0, null, 'neos.io', '/', false, false, Cookie::SAMESITE_NONE);
+        $this->assertTrue($cookie->isSecure());
```
nlx-sascha, line 282: @mficzel here is the test for the case you haven't test.

Neos.Flow/Classes/Http/Cookie.php
```diff
@@ -86,6 +90,23 @@ class Cookie
...
+     *
+     * sameSite=strict
+     *   Cookie will only be sent in a first-party context, cookie is ignored on the initial request to your site.
+     *   This is a good setting when you have cookies relating to functionality, such as changing a password
+     * sameSite=strict
+     *   Cookie will only be sent in a top-level navigations.
+     *   This is a good setting when you need the cookie on the initial request, such as session login via cookie
+     * sameSite=none
+     *   Cookie will sent in a third-party context.
+     *   This is a good setting when you need the cookie in cors ajax request, such as providing an api with session login via cookie
```
jonnitto, line 104: Shouldn't one description be for `lax`?

albe: Indeed! Eagle-eye :)

Neos.Flow/Classes/Http/Cookie.php
```diff
@@ -86,6 +90,23 @@ class Cookie
      */
     protected $httpOnly;
 
+    /**
+     * Possible values: none, lax, or strict (RFC 6265bis-05, 8.8)
+     *
+     * sameSite=strict
+     *   Cookie will only be sent in a first-party context, cookie is ignored on the initial request to your site.
+     *   This is a good setting when you have cookies relating to functionality, such as changing a password
+     * sameSite=strict
```
albe, line 99:
```suggestion
     * sameSite=lax
```

Neos.Flow/Classes/Http/Cookie.php
nlx-sascha: Yes that was a nice catch. Will fix tomorrow

Neos.Flow/Classes/Http/Cookie.php
```diff
@@ -86,6 +90,23 @@ class Cookie
...
+     *   Cookie will only be sent in a top-level navigations.
+     *   This is a good setting when you need the cookie on the initial request, such as session login via cookie
+     * sameSite=none
+     *   Cookie will sent in a third-party context.
```
albe, line 103:
```suggestion
     *   Cookie will be sent in a third-party context.
```

Neos.Flow/Tests/Unit/Http/CookieTest.php
mficzel: I know and i am confident that it works. Just wanted to state that i could not verify 100% on my setuo

PR #1964, opened 2020-04-01
FEATURE: Virtual Object Configuration

bwaidelich:
The Flow Object Management has been built with the idea of virtual "object names".
But so far it was not possible to have an object configuration with a name that is not equal to the class name.
With this feature it's possible to create configurations for "virtual objects":
```
# the colon ":" makes this a virtual object
'Some.Package:SystemLogger':
  className: Psr\Log\LoggerInterface
  scope: singleton
  factoryObjectName: Neos\Flow\Log\PsrLoggerFactoryInterface
  factoryMethodName: get
  arguments:
    1:
      value: systemLogger

'Some.Package:SecurityLogger':
  className: Psr\Log\LoggerInterface
  scope: singleton
  factoryObjectName: Neos\Flow\Log\PsrLoggerFactoryInterface
  factoryMethodName: get
  arguments:
    1:
      value: securityLogger
```

..and to inject them:

```php
/**
 * @Flow\Inject(name="Some.Package:SystemLogger")
 * @var LoggerInterface
 */
protected $systemLogger;

/**
 * @Flow\Inject(name="Some.Package:SecurityLogger")
 * @var LoggerInterface
 */
protected $securityLogger;
```

Or configure them in factories:

```yaml
Some:
  Package:
    someLoggerClassName: 'Some.Package:SystemLogger'
```

Resolves: #671


Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -353,6 +373,51 @@ protected function parseArgumentOfTypeObject($argumentName, $objectNameOrConfigu
...
+     *     1:
+     *       object:
+     *         factoryObjectName: 'Some\Factory\Class'
+     *
+     * into:
+     *
+     * 'Some\Class\Name':
+     *   arguments:
+     *     1:
+     *       object: 'Some\Class\Name::argument:1'
```
bwaidelich, line 390:
```suggestion
     *       object: 'Some\Class\Name:argument:1'
```

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -353,6 +373,51 @@ protected function parseArgumentOfTypeObject($argumentName, $objectNameOrConfigu
...
+     * @param array &$objectConfigurations
+     * @return void
+     * @throws InvalidObjectConfigurationException
+     */
+    protected function wireObjectArguments(array &$objectConfigurations)
+    {
+        /** @var Configuration $objectConfiguration */
+        foreach ($objectConfigurations as $objectConfiguration) {
+            /** @var ConfigurationArgument $argument */
+            foreach ($objectConfiguration->getArguments() as $index => &$argument) {
```
albe, line 405: `&$argument` needed?

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -353,6 +373,51 @@ protected function parseArgumentOfTypeObject($argumentName, $objectNameOrConfigu
...
+     *         factoryObjectName: 'Some\Factory\Class'
+     *
+     * into:
+     *
+     * 'Some\Class\Name':
+     *   arguments:
+     *     1:
+     *       object: 'Some\Class\Name:argument:1'
+     *
+     * 'Some\Class\Name::argument:1':
```
albe, line 392:
```suggestion
     * 'Some\Class\Name:argument:1':
```

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -353,6 +373,51 @@ protected function parseArgumentOfTypeObject($argumentName, $objectNameOrConfigu
         return $argument;
     }
 
+    /**
+     * Creates a "virtual object configuration" for object arguments, turning:
```
albe, line 377: Why do we need to convert all those old object configurations into virtual object configurations?

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -118,14 +119,32 @@ public function buildObjectConfigurations(array $availableClassAndInterfaceNames
                 if (isset($rawObjectConfiguration['className'])) {
                     $rawObjectConfiguration = $this->enhanceRawConfigurationWithAnnotationOptions($rawObjectConfiguration['className'], $rawObjectConfiguration);
                 }
-                $newObjectConfiguration = $this->parseConfigurationArray($objectName, $rawObjectConfiguration, 'configuration of package ' . $packageKey . ', definition for object "' . $objectName . '"', $existingObjectConfiguration);
+                if (strpos($objectName, ':') !== false) {
+                    if (!isset($rawObjectConfiguration['factoryObjectName'])) {
```
albe, line 123: Clever approach. I do like it even though it's hard to understand what is actually done here. Maybe a few doc comments can help that.

albe:
Needs to be rebased or master merged into again to fix the test failures.

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
bwaidelich: Yes, I'll extract that to a method and add tests for it

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
bwaidelich: good point, probably not!

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
bwaidelich: We only convert those that didn't work before (i.e. when there are nested factories configured):
```yaml
Some\Class:
  factoryObjectName: Some\Factory
  arguments:
    1:
      object:
        factoryObjectName: Some\Other\Factory
```
This configuration would trigger a warning:
```
Warning: Illegal offset type in Packages/Framework/Neos.Flow/Classes/ObjectManagement/ObjectManager.php line 198
```
because in the `ObjectManager::buildObjectByFactory()` the `$argumentInformation['v']` for the first argument contains an instance of `ObjectConfiguration`.. so the part `$this->get($argumentInformation['v'])` fails..

TL;DR: For existing installations we actually don't replace anything here because it is only triggered for cases that didn't work before

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -353,6 +373,51 @@ protected function parseArgumentOfTypeObject($argumentName, $objectNameOrConfigu
...
+                    continue;
+                }
+                $argumentValue = $argument->getValue();
+                if (!$argumentValue instanceof Configuration) {
+                    continue;
+                }
+                $argumentObjectName = $objectConfiguration->getObjectName() . ':argument:' . $index;
+                $argumentValue->setObjectName($argumentObjectName);
+                $objectConfigurations[$argumentObjectName] = $argument->getValue();
+                $argument->set($argument->getIndex(), $argumentObjectName, $argument->getType());
```
albe, line 418: Ah, psalm now doing the good job again:
```
Argument 1 of Neos\Flow\ObjectManagement\Configuration\ConfigurationArgument::set expects int, string provided
$argument->set($argument->getIndex(), $argumentObjectName, $argument->getType());
```
https://travis-ci.org/github/neos/flow-development-collection/jobs/670022547#L796
:)

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -442,13 +442,6 @@ protected function buildPropertyInjectionCodeByConfiguration(Configuration $obje
     public function buildPropertyInjectionCodeByString(Configuration $objectConfiguration, ConfigurationProperty $propertyConfiguration, $propertyName, $propertyObjectName)
     {
         $className = $objectConfiguration->getClassName();
-
-        if (strpos($propertyObjectName, '.') !== false) {
-            $settingPath = explode('.', $propertyObjectName);
-            $settings = Arrays::getValueByPath($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS), array_shift($settingPath));
-            $propertyObjectName = Arrays::getValueByPath($settings, $settingPath);
-        }
```
bwaidelich, line 450: This is ancient (~7 year old) magic that turned an inject class name into a setting if it contained a "."..
With the introduction of the `InjectConfiguration` annotation this should have been removed, but it was probably forgotton

bwaidelich:
@albe @kitsunet This is still missing docs and tests. I would add them asap if you agree to the general feature

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
bwaidelich: Ah right, I forgot about that one.
It's great, that `getIndex()` returns a string and `set` expects an int.. but that's not the only weird part about that ancient code :)
fixed

albe:
Travis is green now. So, as soon as you push documentation and one test I will hit merge button ;)

PR #1967, opened 2020-04-01
BUGFIX: Objects configured with factory don't receive factory arguments

### LINKED ISSUE #1933: Object configured with factory will receive supposed factory arguments
Given object configuration like this:
```
Neos\Flow\Http\Middleware\MiddlewaresChain:
  factoryObjectName: Neos\Flow\Http\Middleware\MiddlewaresChainFactory
  factoryMethodName: create
  arguments:
    1:
      setting: Neos.Flow.http.middlewares
```

we expect that an object `MiddlewaresChain` is created through the factory `MiddlewaresChainFactory` which in turn receives the setting `Neos.Flow.http.middlewares` as it's constructor argument. Otherwise, the `MiddlewaresChain` construction and it's arguments is only a concern of the factory.

**Expected:**
A `MiddlewaresChain` can be instanciated without constructor arguments.

**Actual:**
The `MiddlewaresChain` proxy object will fetch the `Neos.Flow.http.middlewares` setting in it's constructor, if no constructor argument is given.

**Explanation:**
The `Objects.yaml` setting `arguments` is used for the class constructor arguments **and** for the factory arguments. If a factory is configured, the `arguments` should only affect that and not the object constructor.

albe:
Until now a class configured to be constructed by a factory with arguments would end up giving those arguments to both the factory and the class constructor. This can lead to errors when the class constructor expects different arguments.
This change fixes that by separating the arguments for the factory from the class constructor arguments. The `arguments` setting in `Objects.yaml` will refer to the factory method arguments when `factoryObjectName` is given, otherwise to the class constructor arguments.

Fixes #1933

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
@@ -325,7 +325,7 @@ class CompileTimeObjectManager extends ObjectManager
                 ];
 
                 $objects[$objectName]['fa'] = [];
-                $factoryMethodArguments = $objectConfiguration->getArguments();
+                $factoryMethodArguments = $objectConfiguration->getFactoryArguments();
                 if (count($factoryMethodArguments) > 0) {
                     foreach ($factoryMethodArguments as $index => $argument) {
                         $objects[$objectName]['fa'][$index] = [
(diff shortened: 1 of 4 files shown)


albe:
This popped up while working on #1928 where every single middleware would receive the chain configuration as argument.

albe:
Need this merged to get it out of the PSR-15 PR. I'll take the responsibility if it should break something.

PR #1969, opened 2020-04-01
FEATURE: Allow configuring static factory methods in Objects.yaml

albe:
This change allows to configure static factory methods in `Objects.yaml` by only specifying a `factoryMethodName` and leaving out `factoryObjectName`.

Example:
```
Acme\My\Class:
  factoryMethodName: Acme\My\Class::fromStatic
  arguments:
    1:
      setting: Acme.My.Class.ConfigurableValue
```

Before this would have required to create a non-static factory method inside a dedicated factory class (to avoid cyclic instanciation).

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
@@ -192,6 +192,8 @@ class Configuration
     {
         $this->factoryObjectName = $objectName;
         if ($this->factoryMethodName === '') {
+            // Needed for b/c because all configured factory objects should default to 'create' method, but not having
+            // a factory object should not lead to a global static 'create' factory method
             $this->factoryMethodName = 'create';
         }
     }
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
@@ -53,10 +53,10 @@ class Configuration
     protected $factoryObjectName = '';
 
     /**
-     * Name of the factory method. Only used if $factoryObjectName is set.
+     * Name of the factory method.
      * @var string
      */
-    protected $factoryMethodName = 'create';
+    protected $factoryMethodName = '';
 
     /**
      * @var string
@@ -191,6 +191,9 @@ class Configuration
     public function setFactoryObjectName($objectName)
     {
         $this->factoryObjectName = $objectName;
+        if ($this->factoryMethodName === '') {
+            $this->factoryMethodName = 'create';
+        }
     }
 
     /**
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
@@ -229,13 +229,13 @@ class Configuration
     }
 
     /**
-     * Returns true if factoryObjectName and factoryMethodName are defined.
+     * Returns true if factoryObjectName or factoryMethodName are defined.
      *
      * @return boolean
      */
     public function isCreatedByFactory()
     {
-        return ($this->factoryObjectName !== '' && $this->factoryMethodName !== '');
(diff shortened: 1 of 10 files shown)


Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
```diff
@@ -435,45 +435,13 @@ public static function buildSetterMethodName(string $propertyName): string
...
-            case 6:
-                $object = new $className($arguments[0], $arguments[1], $arguments[2], $arguments[3], $arguments[4], $arguments[5]);
-            break;
-            default:
-                $class = new \ReflectionClass($className);
-                $object = $class->newInstanceArgs($arguments);
-        }
-
-        return $object;
+        return new $className(...$arguments);
```
albe, line 445: Not directly related to the change at hand, but why didn't we do this earlier? It's also faster https://gist.github.com/nikic/6390366

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -538,7 +538,7 @@ protected function instantiateClass($className, array $arguments)
         }
 
         try {
-            $object = ObjectAccess::instantiateClass($className, $arguments);
+            $object = new $className(...$arguments);
```
albe, line 541: See below...

Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
```diff
@@ -191,6 +191,9 @@ public function getPackageKey()
     public function setFactoryObjectName($objectName)
     {
         $this->factoryObjectName = $objectName;
+        if ($this->factoryMethodName === '') {
+            $this->factoryMethodName = 'create';
```
bwaidelich, line 195: This seems weird, why is this change required?

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
bwaidelich: I think this change makes a lot of sense, but I would prefer a separate PR for this

bwaidelich:
Also we have to keep in mind to adjust #1964 once this is merged

Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
albe: Because `factoryMethodName` was sets with a default value in the property declaration, which led to ALL classes being handled as if they should be created by factory.
`Error: Method "create" does not exist.`

albe:
Ah, keeping me busy by being pedantic on the details :D ❤️ You are totally right though. Will extract the optimization (which should actually end up in 5.3+ tbf) and add some docs and a test.

Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
albe, line 195:
```suggestion
            // Needed for b/c because all configured factory objects should default to 'create' method, but not having a factory object should not lead to a global static 'create' factory method
            $this->factoryMethodName = 'create';
```

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
albe: See #1972

Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
albe: Would this help explain this odd line of code @bwaidelich ?

bwaidelich: Yes, thanks! :)

bwaidelich:
> Ah, keeping me busy by being pedantic on the details :D

I'll just say "Edge-cases be edgy" 😛

albe:
Okay, that should be it @bwaidelich

PR #1972, opened 2020-04-02
TASK: Use argument unpacking instead of call_user_func_array

albe:
This is supported since PHP 5.6 and is actually faster than call_user_func_array at least since PHP 7.
Just be sure to not use it in places where the method is static or a Closure.
This also deprecates `ObjectAccess::instantiateClass()` (retrospectively). It will be removed with Flow 7.

### DIFF
--- a/Neos.Flow/Classes/Aop/AdvicesTrait.php
+++ b/Neos.Flow/Classes/Aop/AdvicesTrait.php
@@ -54,7 +54,7 @@ trait AdvicesTrait
         }
         $methodName = $joinPoint->getMethodName();
         if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$methodName])) {
-            return call_user_func_array(['self', $joinPoint->getMethodName()], $joinPoint->getMethodArguments());
+            return self::$methodName(...$joinPoint->getMethodArguments());
         }
     }
 }
--- a/Neos.Flow/Classes/Aop/AdvicesTrait.php
+++ b/Neos.Flow/Classes/Aop/AdvicesTrait.php
@@ -52,7 +52,8 @@ trait AdvicesTrait
         if (__CLASS__ !== $joinPoint->getClassName()) {
             return parent::Flow_Aop_Proxy_invokeJoinPoint($joinPoint);
         }
-        if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$joinPoint->getMethodName()])) {
+        $methodName = $joinPoint->getMethodName();
+        if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$methodName])) {
             return call_user_func_array(['self', $joinPoint->getMethodName()], $joinPoint->getMethodArguments());
         }
     }
--- a/Neos.Flow/Classes/Aop/AdvicesTrait.php
+++ b/Neos.Flow/Classes/Aop/AdvicesTrait.php
@@ -54,7 +54,8 @@ trait AdvicesTrait
         }
         $methodName = $joinPoint->getMethodName();
         if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$methodName])) {
-            return self::$methodName(...$joinPoint->getMethodArguments());
+            $arguments = array_values($joinPoint->getMethodArguments());
+            return self::$methodName(...$arguments);
         }
     }
 }
--- a/Neos.Flow/Classes/Mvc/Controller/ActionController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/ActionController.php
@@ -462,7 +462,7 @@ class ActionController extends AbstractController
         $validationResult = $this->arguments->getValidationResults();
 
         if (!$validationResult->hasErrors()) {
-            $actionResult = call_user_func_array([$this, $this->actionMethodName], $preparedArguments);
+            $actionResult = $this->{$this->actionMethodName}(...$preparedArguments);
         } else {
             $actionIgnoredArguments = static::getActionIgnoredValidationArguments($this->objectManager);
             if (isset($actionIgnoredArguments[$this->actionMethodName])) {
@@ -487,9 +487,9 @@ class ActionController extends AbstractController
             }
 
             if ($shouldCallActionMethod) {
-                $actionResult = call_user_func_array([$this, $this->actionMethodName], $preparedArguments);
+                $actionResult = $this->{$this->actionMethodName}(...$preparedArguments);
             } else {
-                $actionResult = call_user_func([$this, $this->errorMethodName]);
+                $actionResult = $this->{$this->errorMethodName}();
             }
         }
 
--- a/Neos.Flow/Classes/Session/Aspect/LazyLoadingAspect.php
+++ b/Neos.Flow/Classes/Session/Aspect/LazyLoadingAspect.php
@@ -118,7 +118,7 @@ class LazyLoadingAspect
         if ($this->sessionOriginalInstances[$objectName] === $proxy) {
             return $joinPoint->getAdviceChain()->proceed($joinPoint);
         } else {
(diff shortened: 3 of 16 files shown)


Neos.Flow/Classes/Aop/AdvicesTrait.php
```diff
@@ -52,7 +52,8 @@ public function Flow_Aop_Proxy_invokeJoinPoint(\Neos\Flow\Aop\JoinPointInterface
         if (__CLASS__ !== $joinPoint->getClassName()) {
             return parent::Flow_Aop_Proxy_invokeJoinPoint($joinPoint);
         }
-        if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$joinPoint->getMethodName()])) {
+        $methodName = $joinPoint->getMethodName();
+        if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$methodName])) {
```
albe, line 56:
```suggestion
        if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$joinPoint->getMethodName()])) {
```
Let's maybe not do this here?

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -462,7 +462,7 @@ protected function callActionMethod()
         $validationResult = $this->arguments->getValidationResults();
 
         if (!$validationResult->hasErrors()) {
-            $actionResult = call_user_func_array([$this, $this->actionMethodName], $preparedArguments);
+            $actionResult = $this->{$this->actionMethodName}(...$preparedArguments);
```
albe, line 465: These cases are debatable... the `->{$this->*MethodName}()` syntax is not super nice. I still slightly prefer it over the `call_user_func_array`, but I'm not gonna fight for it.

kitsunet:
Good call, somehow I thought I opened a PR for that before, but that might still sit around here somewehere, so good to have it!

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
```diff
@@ -435,45 +435,13 @@ public static function buildSetterMethodName(string $propertyName): string
...
-     * For less than 7 arguments `new` is used, for more a `ReflectionClass` is created
-     * and `newInstanceArgs` is used.
-     *
-     * Note: this should be used sparingly, just calling `new` yourself or using Dependency
-     * Injection are most probably better alternatives.
-     *
      * @param string $className
      * @param array $arguments
      * @return object
+     * @deprecated directly use "new $className(...$arguments)" instead
```
albe, line 441: We could postpone the deprecation to 5.2 if wanted. Does it hurt though?

Neos.Flow/Classes/Aop/AdvicesTrait.php
kitsunet: Right, how did that sneak in

albe: I was about to replace the following line with `return self::$methodName(...$joinPoint->getMethodArguments());` but then was unsure and it turned out in a half-change.

Neos.Flow/Classes/Mvc/Controller/ActionController.php
kitsunet: Mmm, IDK, we could do something like: 

```
$actionMethodName = $this->actionMethodName;
$this->$actionMethodName(...$preparedArguments);
```

But I don't think that's much easier readable... I think it's fine, it's odd anywys and might need a revisit at some point.

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
kitsunet: nah, let's deprecatae this right away. We can always decide to leave it in for another round of major...

albe:
```
1) Neos\Flow\Tests\Unit\SignalSlot\DispatcherTest::dispatchPassesTheSignalArgumentsToTheSlotMethod
Error: Cannot unpack array with string keys
```
Ah-ha! Edge-cases be edgy. Well... wrapping all arguments everywhere in an `array_values()` before unpacking is counter-effective so I guess... we have to revert back to *some* call_user_func_array?

Neos.Flow/Classes/Aop/AdvicesTrait.php
kitsunet: sure!

kitsunet:
We could just do array_values in that particular place?

Neos.Flow/Classes/Aop/AdvicesTrait.php
albe: Note: because the `$methodName` might be a static method (?)

albe:
> We could just do array_values in that particular place?

Was about to do that, then noticed that `$object` might be the class name in case of static slot method. Reverted to `call_user_func_array` to avoid an additional `if (is_string($object)) {` condition

kitsunet:
right, fair enough....

Neos.Flow/Classes/Aop/AdvicesTrait.php
```diff
@@ -52,7 +52,8 @@ public function Flow_Aop_Proxy_invokeJoinPoint(\Neos\Flow\Aop\JoinPointInterface
...
             return call_user_func_array(['self', $joinPoint->getMethodName()], $joinPoint->getMethodArguments());
```
albe, line 57:
```suggestion
            return self::$methodName(...$joinPoint->getMethodArguments());
```
Alternative to above revert suggestion. You decide @kitsunet ;P

albe:
IMO looks good to merge

Neos.Flow/Classes/Session/Aspect/LazyLoadingAspect.php
```diff
@@ -118,7 +118,7 @@ public function callMethodOnOriginalSessionObject(JoinPointInterface $joinPoint)
         if ($this->sessionOriginalInstances[$objectName] === $proxy) {
             return $joinPoint->getAdviceChain()->proceed($joinPoint);
         } else {
-            return call_user_func_array([$this->sessionOriginalInstances[$objectName], $methodName], $joinPoint->getMethodArguments());
+            return $this->sessionOriginalInstances[$objectName]->$methodName(...$joinPoint->getMethodArguments());
```
albe, line 121:
```suggestion
            $arguments = array_values($joinPoint->getMethodArguments());
            return $this->sessionOriginalInstances[$objectName]->$methodName(...$arguments);
```

Neos.Flow/Classes/Aop/AdvicesTrait.php
```diff
@@ -52,8 +52,9 @@ public function Flow_Aop_Proxy_invokeJoinPoint(\Neos\Flow\Aop\JoinPointInterface
         if (__CLASS__ !== $joinPoint->getClassName()) {
             return parent::Flow_Aop_Proxy_invokeJoinPoint($joinPoint);
         }
-        if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$joinPoint->getMethodName()])) {
-            return call_user_func_array(['self', $joinPoint->getMethodName()], $joinPoint->getMethodArguments());
+        $methodName = $joinPoint->getMethodName();
+        if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode[$methodName])) {
+            return self::$methodName(...$joinPoint->getMethodArguments());
```
albe, line 57:
```suggestion
            $arguments = array_values($joinPoint->getMethodArguments());
            return self::$methodName(...$arguments);
```

albe:
@kitsunet wdyt? good?

albe:
Let's do this... what can possibly go wrong? 😄 If to does go wrong, I'll be ready to hit revert.

PR #1973, opened 2020-04-02
FEATURE: Configurable Authentication Token options

bwaidelich:
Resolves: #1944

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
+++ b/Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
@@ -13,6 +13,7 @@ namespace Neos\Flow\Security\Authentication\Token;
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Mvc\ActionRequest;
+use Neos\Flow\Security\Exception\InvalidAuthenticationStatusException;
 use Neos\Utility\ObjectAccess;
 
 /**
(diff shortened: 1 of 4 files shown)


Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
```diff
@@ -13,13 +13,18 @@
...
 use Neos\Flow\Mvc\ActionRequest;
+use Neos\Flow\Security\Exception\InvalidAuthenticationStatusException;
 use Neos\Utility\ObjectAccess;
 
 /**
  * An authentication token used for simple username and password authentication.
  */
 class UsernamePassword extends AbstractToken
 {
+    private const DEFAULT_USERNAME_POST_FIELD = '__authentication.Neos.Flow.Security.Authentication.Token.UsernamePassword.username';
```
albe, line 24: *maybe* make that public so third parties can easily use the field path? Also see the code example here: https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Security.rst#simple-usernamepassword-authentication

bwaidelich:
Theres one caveat: The options are serialized into the session (if not using a sessionless token). So if you change the options configuration, those serialized tokens will still contain the old options.. But this is true for all provider configuration.

Also this is missing tests and docs yet

bwaidelich:
I'll add tests & docs once I get a positive feedback from @sorenmalling (#1945)

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
bwaidelich: I don't think it should be part of the API really.. I just introduced the constants to avoid the magic strings really

bwaidelich:
Closing in favor of #1975

PR #1975, opened 2020-04-03
FEATURE: Configurable UsernamePassword fields and UsernamePasswordTokeninterface

sorenmalling:
Resolves: #1944


albe:
Thinking about it - don't all token somehow need to `extractCredentialsFromRequest()`? I mean, of course some might fetch some data from elsewhere, but which data they need to fetch is probably somehow inside the http request.

So what I'm asking - should we maybe rename the new interface to `CredentialsFromRequestInterface` and at some later point (Flow 7) merge that into `TokenInterface`?
Just thinking out loud here, might be a stupid idea.

Neos.Flow/Classes/Security/Authentication/Token/UsernamePasswordInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+use Neos\Flow\Mvc\ActionRequest;
+
+/**
+ * Marker interface for authentication tokens which don't need a session.
+ */
+interface UsernamePasswordInterface
+{
+    public function extractCredentialsFromRequest(ActionRequest $request): array;
```
bwaidelich, line 21: The interface should be
```php
public function getUsername(): string;

public function getPassword(): string;
```

IMO since this is not about the extraction mechanism (which might not even be related to the `ActionRequest`) but about the fact that tokens implementing it, hold username & password

albe: Well, that also makes sense if we keep it as a `UsernamePasswordInterface` :D

sorenmalling: Will you make the action request available as a property for the entire Token?

sorenmalling: I got the idea know. New version pushed

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
```diff
@@ -62,27 +68,27 @@ public function updateCredentials(ActionRequest $actionRequest)
             return;
         }
 
-        $credentials = $this->extractCredentialsFromRequest($actionRequest);
+        $username = $this->getUsername();
```
albe, line 71: `$this->actionRequest` is not set before calling those two methods. What about passing the ActionRequest as argument instead of relying on the property?

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
```diff
@@ -13,30 +13,53 @@
...
      * @var array
      * @Flow\Transient
      */
     protected $credentials = ['username' => '', 'password' => ''];
 
+    /**
+     * The current ActionRequest
+     * @var ActionRequest
+     */
+    protected $actionRequest;
```
bwaidelich, line 38: That's dangerous, it will serialize the action request into the token!
(You could add a `@Flow\Transient` annotation, but IMO it would be better not to keep the request in the object state at all)

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
```diff
@@ -56,6 +78,18 @@ public function updateCredentials(ActionRequest $actionRequest)
         }
     }
 
+    public function getUsername(): string
+    {
```
bwaidelich, line 82: I think this should just be
```php
return $this->credentials['username'] ?? '';
```

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
```diff
@@ -56,6 +78,18 @@ public function updateCredentials(ActionRequest $actionRequest)
...
+        $arguments = array_merge($this->actionRequest->getArguments(), $this->actionRequest->getInternalArguments());
+        return ObjectAccess::getPropertyPath($arguments, $this->options['usernamePostField'] ?? self::DEFAULT_USERNAME_POST_FIELD);
+    }
+
+    public function getPassword(): string
+    {
+        $arguments = array_merge($this->actionRequest->getArguments(), $this->actionRequest->getInternalArguments());
```
bwaidelich, line 89: see above

bwaidelich: also: merging all arguments was my hack to mix internal and regular args, but we should not do it twice :)

Neos.Flow/Classes/Security/Authentication/Token/UsernamePasswordInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+/**
+ * Marker interface for authentication tokens which hold a username and password
```
bwaidelich, line 15: A marker interface is an interface without methods

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
sorenmalling: I lost it here.. how are we suppose to set the values then? If we do not have a place, where we take the request into account, for extractig the username and password?

Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
```diff
@@ -74,8 +75,8 @@ public function getTokenClassNames()
      */
     public function authenticate(TokenInterface $authenticationToken)
     {
-        if (!($authenticationToken instanceof UsernamePassword)) {
-            throw new UnsupportedAuthenticationTokenException('This provider cannot authenticate the given token.', 1217339840);
+        if (!($authenticationToken instanceof UsernamePasswordInterface)) {
+            throw new UnsupportedAuthenticationTokenException(sprintf('This provider cannot authenticate the given token. The token must implement %s', UsernamePasswordInterface::class), 1217339840);
```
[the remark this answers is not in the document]
sorenmalling: The instantiation of a interface (with `new $className()` in the tokenProviderFactory class) automatically creates a instance of what is defined in OBjects.yaml ?

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
albe: yeah, that kind of misses the point of the interface then to easily be able to overwrite from where of the request the credentials are taken. What if we change the signature to `getUsername(ActionRequest $request): string` as I suggested above?

bwaidelich: We have the credentials anyways (since we have to satisfy the `TokenInterface` that provides a getter for them).

I'll make a suggestion

bwaidelich:
@sorenmalling I didn't want to mess with your PR and make it bigger once again :)
but IMO this is required in order to support the new interfaces: https://github.com/sorenmalling/flow-development-collection/pull/1 (it's a PR against _this_ PR!)

If that grows out of hand, we could still implement the minimal version I suggested with #1973 and introduce the interfaces lateron (i.e. next release).
But if we introduce them they should be embraced fully IMO

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
```diff
@@ -13,49 +13,88 @@
...
      * @Flow\Transient
      */
     protected $credentials = ['username' => '', 'password' => ''];
 
+    /**
+     * The current ActionRequest
+     * @var ActionRequest
+     */
+    protected $actionRequest;
+
```
bwaidelich, line 39:
```suggestion
```

bwaidelich: I forgot to remove that one apparently

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
kitsunet: why would we want to merge here or in general?

Neos.Flow/Classes/Security/Authentication/Token/PasswordToken.php
```diff
@@ -42,22 +39,30 @@ class PasswordToken extends AbstractToken
...
     public function updateCredentials(ActionRequest $actionRequest)
     {
         if ($actionRequest->getHttpRequest()->getMethod() !== 'POST') {
             return;
         }
-
-        $postArguments = $actionRequest->getInternalArguments();
-        $password = ObjectAccess::getPropertyPath($postArguments, '__authentication.Neos.Flow.Security.Authentication.Token.PasswordToken.password');
-
+        $allArguments = array_merge($actionRequest->getArguments(), $actionRequest->getInternalArguments());
```
kitsunet, line 49: Why would we want to look into internal arguments here? The method states it looks into POST, which would be satisfied by getArguments() alone?!

Neos.Flow/Classes/Security/Authentication/Token/UsernamePassword.php
bwaidelich: because of the weird way Flow separates internal and default arguments.
```php
ObjectAccess::getPropertyPath($this->actionRequest->getArguments(), '__authentication.Neos.Flow.Security.Authentication.Token.UsernamePassword.username');
```
does not work

Neos.Flow/Classes/Security/Authentication/Token/PasswordToken.php
bwaidelich: `__authentication.Neos.Flow.Security.Authentication.Token.UsernamePassword.username` _is_ considered an internal argument by Flow...

kitsunet: Oh yeah, ofc 🤦 mmmm, tricky that

bwaidelich:
Oh dear.. the tests.. I'll check

albe:
Ok, so let's merge this and use the next week to watch and iron out things on this one, okay?

PR #1977, opened 2020-04-04
TASK: Add release notes for 6.1

albe:
This is an (mostly empty) stub for the till now missing 6.1 release notes

TODO: Update with most important changes - see https://github.com/neos/flow-development-collection/blob/6.1/Neos.Flow/Documentation/TheDefinitiveGuide/PartV/ChangeLogs/610.rst
and https://www.neos.io/blog/neos-5-1-white-knight-and-flow-6-1-released.html


markusguenther:
Oh then @davidspiola and I missed something in the release of 6.1 and 5.1 right?

albe:
Yes, but you're not the only ones :)

PR #1978, opened 2020-04-05
FEATURE: Improve 304 response handling

mficzel:
For responses that match the following conditions an 304 status is returned:
- The request method is GET or HEAD.
- The response status is 200 status.
- The response has has ETag header that matches one of the If-None-Match headers of the request.

!!! The generation of the Etag header is not part of this change and is the job of the application.

Other changes: 
- 304 responses now have an empty body
- Only GET and HEAD Requests will receive 304 status based on ETag and Last-Modified headers

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
+++ b/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
@@ -150,7 +150,17 @@ abstract class ResponseInformationHelper
     public static function makeStandardsCompliant(ResponseInterface $response, RequestInterface $request): ResponseInterface
     {
         $statusCode = $response->getStatusCode();
-        if ($request->hasHeader('If-Modified-Since') && $response->hasHeader('Last-Modified') && $statusCode === 200) {
+        if ($request->hasHeader('If-None-Match') && in_array($request->getMethod(), ['HEAD', 'GET'])
+            && $response->hasHeader('ETag') && $statusCode === 200 ) {
+            $ifNoneMatchHeaders = $request->getHeader('If-None-Match');
+            $eTagHeader = $response->getHeader('ETag')[0];
+            foreach ($ifNoneMatchHeaders as $ifNoneMatchHeader) {
+                if (ltrim($ifNoneMatchHeader, 'W/') == ltrim($eTagHeader, 'W/')) {
+                    $response = $response->withStatus(304);
+                    break;
+                }
+            }
+        } else if ($request->hasHeader('If-Modified-Since') && $response->hasHeader('Last-Modified') && $statusCode === 200) {
             $ifModifiedSince = $request->getHeader('If-Modified-Since')[0];
             $ifModifiedSinceDate = \DateTime::createFromFormat(DATE_RFC2822, $ifModifiedSince);
             $lastModified = $response->getHeader('Last-Modified')[0];
--- a/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
+++ b/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
@@ -151,7 +151,7 @@ abstract class ResponseInformationHelper
     {
         $statusCode = $response->getStatusCode();
         if ($request->hasHeader('If-None-Match') && in_array($request->getMethod(), ['HEAD', 'GET'])
-            && $response->hasHeader('ETag') && $statusCode === 200 ) {
+            && $response->hasHeader('ETag') && $statusCode === 200) {
             $ifNoneMatchHeaders = $request->getHeader('If-None-Match');
             $eTagHeader = $response->getHeader('ETag')[0];
             foreach ($ifNoneMatchHeaders as $ifNoneMatchHeader) {
@@ -160,7 +160,8 @@ abstract class ResponseInformationHelper
                     break;
                 }
             }
-        } else if ($request->hasHeader('If-Modified-Since') && $response->hasHeader('Last-Modified') && $statusCode === 200) {
+        } elseif ($request->hasHeader('If-Modified-Since') && in_array($request->getMethod(), ['HEAD', 'GET'])
+            && $response->hasHeader('Last-Modified') && $statusCode === 200) {
             $ifModifiedSince = $request->getHeader('If-Modified-Since')[0];
             $ifModifiedSinceDate = \DateTime::createFromFormat(DATE_RFC2822, $ifModifiedSince);
             $lastModified = $response->getHeader('Last-Modified')[0];
--- a/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
+++ b/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
@@ -156,7 +156,9 @@ abstract class ResponseInformationHelper
             $eTagHeader = $response->getHeader('ETag')[0];
             foreach ($ifNoneMatchHeaders as $ifNoneMatchHeader) {
                 if (ltrim($ifNoneMatchHeader, 'W/') == ltrim($eTagHeader, 'W/')) {
-                    $response = $response->withStatus(304);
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
```diff
@@ -149,7 +150,18 @@ public static function prepareHeaders(ResponseInterface $response): array
     public static function makeStandardsCompliant(ResponseInterface $response, RequestInterface $request): ResponseInterface
     {
         $statusCode = $response->getStatusCode();
-        if ($request->hasHeader('If-Modified-Since') && $response->hasHeader('Last-Modified') && $statusCode === 200) {
+        if ($request->hasHeader('If-None-Match') && in_array($request->getMethod(), ['HEAD', 'GET'])
+            && $response->hasHeader('ETag') && $statusCode === 200) {
```
albe, line 154: Should we also handle `If-None-Match` with another request method to return a `412` status? (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match)

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
```diff
@@ -194,6 +206,21 @@ public static function makeStandardsCompliant(ResponseInterface $response, Reque
             $response = $response->withoutHeader('Content-Length');
         }
 
+        #
+        # Create a 304 response with an empty body and copy only the required headers
+        # from the original response as is required by RFC 7232 section 4.1
+        #
+        if ($response->getStatusCode() == 304) {
+            $keepHeaders = ['Cache-Control', 'Content-Location', 'Date', 'ETag', 'Expires', 'Vary'];
```
albe, line 214: Is this really supposed to be a safe-list and strip all other headers?

> The server generating a 304 response MUST generate any of the
   following header fields that would have been sent in a 200 (OK)
   response to the same request: Cache-Control, Content-Location, Date,
   ETag, Expires, and Vary.

I read that those headers MUST at least be added from the original 200 response. I don't read that others should be removed, though?
In case I'm wrong: At least, should we maybe also allow custom headers `X-*`?

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
```diff
@@ -149,7 +150,18 @@ public static function prepareHeaders(ResponseInterface $response): array
...
+            && $response->hasHeader('ETag') && $statusCode === 200) {
+            $ifNoneMatchHeaders = $request->getHeader('If-None-Match');
+            $eTagHeader = $response->getHeader('ETag')[0];
+            foreach ($ifNoneMatchHeaders as $ifNoneMatchHeader) {
+                if (ltrim($ifNoneMatchHeader, 'W/') == ltrim($eTagHeader, 'W/')) {
+                    $response = $response->withStatus(304);
+                    break;
+                }
+            }
+        } elseif ($request->hasHeader('If-Modified-Since') && in_array($request->getMethod(), ['HEAD', 'GET'])
```
albe, line 163: At some point we should also add `If-Unmodified-Since` ;) (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Unmodified-Since)

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
mficzel: > Since the goal of a 304 response is to minimize information transfer
   when the recipient already has one or more cached representations, a
   sender SHOULD NOT generate representation metadata other than the
   above listed fields unless said metadata exists for the purpose of
   guiding cache updates (e.g., Last-Modified might be useful if the
   response does not have an ETag field).

I read that other header SHOULD be removed.

mficzel: Maybe we could add a configuration for that

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
mficzel: Not sure we can do that in the standard compliance component as this is 
 used to prevent conflicting write operations see: https://tools.ietf.org/html/rfc7232#section-3.2. 

That can imho only be solved by the application itself.

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
mficzel: There is a section for that here https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php#L160-L169 but that has the same issues that would apply to handling POST requests. 

The If-Match and If-Unmodified-Since headers are specified to prevent two concurrent write operations beeing applied. The status 412 only indicates that the application could not perform the  requested operation because a precondition was not matched. 

Adding only the 412 header as is currently done without preventing the execution of the code could/should be considered a bug.

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
mficzel: We could also ignore the header part and only empty the response body. That would avoid the need of having a configuration for that.

mficzel: @albe I decided to leave the headers in and only empty the response body for 304 responses. 
SHOULD allows this and this cause less problems and discussions then a whitelisting approach.

mficzel:
@albe if you like this i can rebase this for 5.2

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
albe: Well, right, that would need to happen in the cache handling component to check the match condition and the request handling short-circuited at that place. So, something for FullPageCache ;)

mficzel: Full Page Cache only needs to ensure that the ETag header is sent after this change went in. As  the 304 status will be sent from here.

Nice thing is that every Flow Application that creates proper ETags benefits. Regardless wether the same result was calculated again or cached.

While recalculating and later removing the body may look like a pointless approach it makes sense from the client perspective as the client can reuse the already parsed result from the cache and can safe transfer and parsing time.

mficzel: I added a separate pr to remove the 412 status rendering https://github.com/neos/flow-development-collection/pull/1979  

Will probably take some time and maybe even have to wait for flow 7

Neos.Flow/Resources/Private/Translations/ja/ValidationErrors.xlf
```diff
@@ -26,14 +26,14 @@
 				<source>The given value is not a valid date</source>
 			<target state="translated">指定された値は有効な日付ではありません</target></trans-unit>
       <trans-unit id="1325615630" xml:space="preserve">
-				<source>The given date must be between {0,datetime,datetime} and {1,datetime,datetime}</source>
-			<target state="needs-translation">The given date must be between {0,datetime,datetime} and {1,datetime,datetime}</target></trans-unit>
+				<source>The given date must be between {0} and {1}</source>
```
albe, line 29: Huh...? I think this commit has sneaked in accidentially.

mficzel: @albe this is the result of changing the targetBranch from master to 6.2  without rebase.
See no problem with that as this is just an automatically updated translation.

mficzel: Rebased the commits onto 6.2 so this commit is no longer part of this pr

PR #1979, opened 2020-04-06
!!! BUGFIX: Do not create 412 Status 'Precondition Failed' in standard compliance helper

mficzel:
As defined in RFC 7232 section 4.2 the http status 412 indicates that a precondition for a request failed
and that the requested operation could not be executed which only makes sense for non safe requests.

The bug in here is that this is checked after the fact so Flow would still perform the operation but the standard compliance component would add a precondition failed header and thus not prevent the in flight collision but make the problem harder to understand.

This change removes the code for the 412 status from the standard compliance helper. If any application would want to use the status it should evaluate the precondition headers and throw an exception with 412 status if needed.

Fixes: #2062

### DIFF
--- a/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
+++ b/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
@@ -157,15 +157,6 @@ abstract class ResponseInformationHelper
             if ($lastModifiedDate <= $ifModifiedSinceDate) {
                 $response = $response->withStatus(304);
             }
-        } elseif ($request->hasHeader('If-Unmodified-Since') && $response->hasHeader('Last-Modified')
-            && (($statusCode >= 200 && $statusCode <= 299) || $statusCode === 412)) {
-            $unmodifiedSince = $request->getHeader('If-Unmodified-Since')[0];
-            $unmodifiedSinceDate = \DateTime::createFromFormat(DATE_RFC2822, $unmodifiedSince);
-            $lastModified = $response->getHeader('Last-Modified')[0];
-            $lastModifiedDate = \DateTime::createFromFormat(DATE_RFC2822, $lastModified);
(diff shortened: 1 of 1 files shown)


mficzel:
Not sure wether this should go in as bugfix and into which branches or wether this is a behavior change for the next major.

albe:
Uh, that one's odd. Alone for the fact that we still keep the `If-Modified-Since` check in place, which makes sense, but is just soooo unintuitive why `If-Unmodified-Since` doesn't. But yes, we need to fix this for semantic correctness.

So the breaking condition is:
- the application sets a `Last-Modified` header in it's responses
- the client requests with an `If-Unmodified-Since` header
- the application is not idempotent in that case and running the action again changes the state in a meaningful way

Sounds very edge-casey, so maybe we could target this BUGFIX lower - but then again I'm happy to not introduce possibly breaking changes in old versions to fix rare edge-cases. So if you ask me I'd keep it in master.

mficzel:
@albe i agree to put this into next major.

It still makes sense to handle if-none-match and if-modífied-since in the standard compliance component as although we obviously cannot prevent the recalculation of the result there but we still can avoid transferring and parsing content the client already has cached.

Ideally also the application layer would look at those tags and if possible avoid the calculation but it still makes sense to have a fallback there.

Although this approach allows FullPageCache to not deal with 304 status and instead only cache the responses.

jonnitto:
@mficzel Could you create an issue for that? We use the issues for tracking the progress on the release project.

bwaidelich:
I created an issue for that (#2062) and replaced the PR with the feature on the Flow 7.0 Release board

bwaidelich:
@mficzel FYI you added this PR to the 6.0/7.0 project board again, but we decided only to work with issues there. I already created and linked an issue (#2062)

PR #1980, opened 2020-04-08
FEATURE: Array Debugger tolerates iterables

mficzel:
This change lets the debug-array-renderer use the pseudo type iterable (Array or Traversable) instead of array. That way all foreachachble objects are rendered as an array. 

This replaces the previous exceptions for ArrayObjects and DoctrineCollections and will also enable to properly debug other Traversable objects aswell.

### DIFF
--- a/Neos.Flow/Classes/Error/Debugger.php
+++ b/Neos.Flow/Classes/Error/Debugger.php
@@ -140,7 +140,7 @@ class Debugger
             }
         } elseif (is_numeric($variable)) {
             $dump = sprintf('%s %s', gettype($variable), self::ansiEscapeWrap($variable, '35', $ansiColors));
-        } elseif (is_array($variable)) {
+        } elseif (is_iterable($variable)) {
             $dump = self::renderArrayDump($variable, $level + 1, $plaintext, $ansiColors);
         } elseif (is_object($variable)) {
             $dump = self::renderObjectDump($variable, $level + 1, true, $plaintext, $ansiColors);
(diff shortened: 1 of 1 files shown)


mficzel:
For the review it should be reasoned about wether removing the specific code for doctrine collections is really a good idea. I am not that deep into this to fully exclude unwanted side effects here.

Neos.Flow/Classes/Error/Debugger.php
```diff
@@ -197,9 +197,8 @@ protected static function renderObjectDump($object, int $level, bool $renderProp
         $scope = '';
         $additionalAttributes = '';
 
-        if ($object instanceof \Doctrine\Common\Collections\Collection || $object instanceof \ArrayObject) {
-            // The doctrine Debug utility usually returns a \stdClass object that we need to cast to array.
-            return self::renderArrayDump((array)\Doctrine\Common\Util\Debug::export($object, 3), $level, $plaintext, $ansiColors);
+        if (is_iterable($object)) {
+            return self::renderArrayDump($object, $level, $plaintext, $ansiColors);
```
albe, line 201: Had to check myself too, but apparently `Doctrine\Common\Util\Debug` is deprecated anyway in favor of symfony/var-dumper. Also, the only special thing that Doctrine debug util did in case of Collections, was convert them to an array via `->toArray()`, which should not be necessary. It had a couple of special additions for DateTime and Proxy classes though, but those also didn't effect this code path AFAIS.

PR #1987, opened 2020-04-20
FEATURE: Suppport configuration of max lock wait time

### LINKED ISSUE #1911: Lost session with FileBackend and slower latency storage (e.g. network storage like Ceph)
- Concurrent access to the session can cause a re-login (the session looks expired to the user) when using a slower latency filesystem (e.g. network storage like Ceph RBD)
- We experienced this behaviour in a Neos installation with a persistent filesystem backed by a Ceph RBD persistent volume in a Kubernetes installation
- It looks like the file locking and wait operation (see https://github.com/neos/flow-development-collection/blob/fdb7b00603b3982025047bd84ad1b115bfe566a9/Neos.Cache/Classes/Backend/SimpleFileBackend.php#L495) is reporting the session cache entry as `false` after 3 retries with varying wait time - Neos presents a login screen if a single AJAX request was not authenticated
- Using a ephemeral, local storage, does solve the issue as a workaround
- It would be better if the wait time and retry count could be configurable as options, so it can be extended (retry count for a session should be unlimited or the entry should not be reported as missing but an exception could be thrown)

hlubek:
Implement a new option `maxLockWaitMicroseconds` for file cache backends. Acquiring a lock now repeats after failure and waits at least this amount of time before failing. As a sensible default we wait a millisecond which is roughly the same as the previous hard-coded loop with 3 iterations.

Important cache backends like session should use a higher value by default. It would be best to actually cause an exception instead of silently not finding a cache entry for these scenarios.

Up for discussion: should we make the delay interval / range configurable? E.g. we could add a base delay and jitter.

Fixes #1911


albe:
Hi @hlubek , can you take a look at https://github.com/neos/flow-development-collection/pull/1817 and possibly clean that up? @fcool already invested quite some time into this and it was already prettyy far, but unfortunately never finished (also due to my lack of further feedback at some point).

hlubek:
Well, I didn't see #1817 before writing this. Sadly I cannot contribute to that PR (not a maintainer anymore). But I could merge the existing PR in my branch so we could close #1817 and continue here.

fcool:
Let me know, if I can be of any help. It would be very nice to have this settled, finally :)

hlubek:
@fcool Thanks for the offer, I'll try to merge your PR with some ideas of this branch if it's okay for you.

fcool:
Of course. Your aim is my aim ;) We both want to better cope with low latency file systems

sorenmalling:
Hi @fcool  and @hlubek :) How did this move along, did you merge your ideas into one pull request, that we can merge?

bwaidelich:
I missed the two PRs until now. I took the freedom to turn this into a `FEATURE` since it is clearly one :)
@hlubek @fcool did you see the comment from @sorenmalling?

fcool:
Well... I have a different opinion on that. Because it is just failing on high latency file systems. Which at least is unexpected and in current state there is no way to recover but to use better performing file systes.
But yes - the proposed solution is a feature. My original commit in #1817 raised the default waiting time as well, so addressing the problematic behavior directly.

bwaidelich:
@fcool Ah I see, thanks for the explanation.
I would say it's _two_ things in that case:
a potential bugfix that raises the waiting time and a feature that allows fine grained control over it. Don't you think?

hlubek:
@fcool Yes! Especially with network filesystems (e.g. in K8s) this is an issue without configuring special lower latency cache backends.

> It would be best to actually cause an exception instead of silently not finding a cache entry for these scenarios.

This would be something to be addressed to not fail silently but could cause a behaviour change, so it needs additional configuration etc.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Caching.rst
```diff
@@ -355,20 +355,25 @@ Options
...
+|                         | * /tmp/my-cache-directory/             |           |        |         |
++-------------------------+----------------------------------------+-----------+--------+---------+
+| defaultLifeTime         | Cache entry lifetime is **not          | No        |        |         |
+|                         | supported** in this backend. Entries   |           |        |         |
+|                         | never expire!                          |           |        |         |
++-------------------------+----------------------------------------+-----------+--------+---------+
+| maxLockWaitMicroseconds | Maximum time to wait for getting a     | No        | int    | 1000    |
+|                         | file lock for read/write operations in |           |        |         |
+|                         | microseconds. This should be increased |           |        |         |
+|                         | for lower latency storages.            |           |        |         |
```
kdambekalns, line 375:
```suggestion
|                         | microseconds. This should be increased |           |        |         |
|                         | for higher latency storages.           |           |        |         |
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Caching.rst
```diff
@@ -406,16 +411,21 @@ Options
...
+|                         | directory.                             |           |        |         |
+|                         |                                        |           |        |         |
+|                         | :title:`Example:`                      |           |        |         |
+|                         |                                        |           |        |         |
+|                         | * /tmp/my-cache-directory/             |           |        |         |
++-------------------------+----------------------------------------+-----------+--------+---------+
+| maxLockWaitMicroseconds | Maximum time to wait for getting a     | No        | int    | 1000    |
+|                         | file lock for read/write operations in |           |        |         |
+|                         | microseconds. This should be increased |           |        |         |
+|                         | for lower latency storages.            |           |        |         |
```
kdambekalns, line 427:
```suggestion
|                         | microseconds. This should be increased |           |        |         |
|                         | for higher latency storages.           |           |        |         |
```

kdambekalns (changes requested):
I think the PR is fine, except a tiny detail in the added documentation – I think that needs to be reversed, no?

kdambekalns:
(Re-trigger) TravisCI…

albe:
Would like to get this into 7.1 as a feature. Can someone merge master into this again and maybe adjust the docs as Karsten suggested?

kitsunet (approved):

kdambekalns:
Trying to fix CI…

kdambekalns (approved):

kdambekalns:
Wait, does this need to go into 8.3, strictly speaking?

kitsunet:
Yes, you are right, it should.

kdambekalns:
Changing base in GH pulls in way more commits, need to be adjusted manually.

PR #1989, opened 2020-04-21
BUGFIX: Adapt virtual objects change to factory arguments bugfix

albe:
The original code only attempted to fix newly introduced factory arguments, but since #1967 those factory arguments are separated from object arguments. Hence this "fix" didn't work any more.


albe:
Psalm errors will be resolved with #1966

albe:
tbqh I trust the tests more :D Things are more green than before, so 🎉

albe:
Correction: Fully green now 💚

PR #1992, opened 2020-04-24
TASK: Trigger GitHub rebase action only on /rebase

baschny:
Idea is that a `/rebase` comment rebases the current PR on top of the destination branch, so that we are sure that latest state is included. Current implementation tries the rebase on every comment (see results in https://github.com/neos/flow-development-collection/actions).

In 6.1/6.2 the action was already converted to "v2" style (see https://github.com/neos/flow-development-collection/pull/1765).

So in 5.3 first cherry-picked the conversion from 
After that changed the workflow definition so it looks exactly like the documented:
https://github.com/cirrus-actions/rebase

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,9 +1,0 @@
-workflow "Automatic Rebase" {
-  on = "issue_comment"
-  resolves = "Rebase"
-}
-
-action "Rebase" {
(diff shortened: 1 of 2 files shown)


albe:
If that does indeed fix the `/rebase` comment use-case then I love you :)
Not sure how to test/verify it other than merging this and trying it out - so, I'd do just that?

PR #1993, opened 2020-04-29
FEATURE: Configurable PersistedUsernamePasswordProvider lookup name

bwaidelich:
Adds an option `lookupProviderName` to the `PersistedUsernamePasswordProvider` that allows the provider to be reused for multiple authentication types.

Example:

```yaml
Neos:
  Flow:
    security:
      authentication:
        providers:
          'Acme.SomePackage:Default':
            provider: PersistedUsernamePasswordProvider
            token: UsernamePassword
          'Acme.SomePackage:Default.HttpBasic':
            provider: PersistedUsernamePasswordProvider
            providerOptions:
              lookupProviderName: 'Acme.SomePackage:Default'
            token: UsernamePasswordHttpBasic
            entryPoint: HttpBasic
```

Resolves: #1996

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
+++ b/Neos.Flow/Classes/Security/Authentication/Provider/PersistedUsernamePasswordProvider.php
@@ -91,7 +91,7 @@ class PersistedUsernamePasswordProvider extends AbstractProvider
             return;
         }
 
-        $providerName = $this->options['lookupName'] ?? $this->name;
+        $providerName = $this->options['lookupProviderName'] ?? $this->name;
         $this->securityContext->withoutAuthorizationChecks(function () use ($username, &$account, $providerName) {
             $account = $this->accountRepository->findActiveByAccountIdentifierAndAuthenticationProviderName($username, $providerName);
         });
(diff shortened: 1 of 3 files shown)


bwaidelich:
TODO:

- [ ] Documentation
- [ ] Better name for the "lookupName" option?

DavidSporer:
> * [ ]  Better name for the "lookupName" option?

What about something like "accountProviderName"? I mean that's what it does in the end, isn't it? It changes the Provider Name that is used to look for accounts based on the username. lookupName would be fine for me as well, though. I assume this feature is something that only people will use that have quite some in-depth knowledge of Flow so they will have read the documentation at least once. We just need to make sure that it's properly described what that option does.
Thanks for taking care of it!

bwaidelich:
Thanks for the feedback.
I think it's a good idea to emphasize that this is just about the "account lookup". After letting this sink for a while, I'd go for `lookupProviderName` and will adjust the code accordingly.
@DavidSporer would you be up for creating a ticket for that?

bwaidelich:
> Shouldn't this be authenticatingAndUsernamePasswordTokenRespectsTheConfiguredLookupProviderName?

oops, yes, fixed. BTW: You can suggest fixes directly in the code

> Btw. THANK YOU :-)

You're welcome. Thank you for the ticket and sorry that this didn't make it for 6.2..

PR #1998, opened 2020-05-04
fix log file size

rkrahnen:
Hey guys, it's just a small fix for the log file size of the `i18nLogger` as it seems to me like there is just a zero missing.


albe:
Ah! Very good catch, indeed that is off! Would be good to fix this in 5.3 though, as it's clearly a BUG:
https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Configuration/Settings.Log.yaml#L47

rkrahnen:
Will you do this or should I create a new pull request for the 5.3 branch?

albe:
Needs a new PR, so would be cool if you could do this :) Thanks!

albe:
friendly ping @rkrahnen :)

rkrahnen:
I will do the new PR today evening, sorry for the delay!

PR #2002, opened 2020-05-15
BUGFIX: Avoid trying to autowire constructor arguments for factory created objects

albe:
This bug was uncovered with the previous fix to not map arguments to the object when a factory is configured (see #1967).
That way (singleton) objects which expected a (non-optional/nullable) constructor argument that could not be autowired (e.g. because it was a scalar) would lead to the exception "Could not autowire required constructor argument ... for singleton class ...".
This change fixes that, by completely skipping autowiring for objects configured with a factory. The factory has full responsibility to create and wire the constructor arguments correctly.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
@@ -377,6 +377,9 @@ class ConfigurationBuilder
                 continue;
             }
 
+            if ($objectConfiguration->isCreatedByFactory()) {
+                continue;
+            }
 
             $className = $objectConfiguration->getClassName();
             if (!$this->reflectionService->hasMethod($className, '__construct')) {
--- a/Neos.Flow/Tests/Unit/ObjectManagement/Configuration/ConfigurationBuilderTest.php
+++ b/Neos.Flow/Tests/Unit/ObjectManagement/Configuration/ConfigurationBuilderTest.php
@@ -16,6 +16,7 @@ use Neos\Flow\ObjectManagement\Configuration\Configuration;
 use Neos\Flow\ObjectManagement\Configuration\ConfigurationArgument;
 use Neos\Flow\ObjectManagement\Configuration\ConfigurationBuilder;
 use Neos\Flow\ObjectManagement\Configuration\ConfigurationProperty;
+use Neos\Flow\ObjectManagement\Exception\UnresolvedDependenciesException;
 use Neos\Flow\Reflection\ReflectionService;
 use Neos\Flow\Tests\UnitTestCase;
 use Neos\Flow\Annotations as Flow;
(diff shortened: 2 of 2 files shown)


bwaidelich:
I have some concerns and questions:
1. This is a bugfix, but it's targeted for `5.3`. By intention?
2. I think this is quite a breaking change (if I got it right)

For the Neos.EventSourcing Package I prepared the following:

Objects.yaml:
```yaml
Neos\EventSourcing\EventStore\Storage\Doctrine\DoctrineEventStorage:
  arguments:
    1:
      value: []
    2:
      object: 'Neos\EventSourcing\EventStore\EventNormalizer'
    3:
      object:
        factoryObjectName: 'Neos\EventSourcing\EventStore\Storage\Doctrine\Factory\ConnectionFactory'
```

In the Settings.yaml you could write:

```yaml
Neos:
  EventSourcing:
    EventStore:
      stores:
        'Wwwision.Test:EventStore':
          storage: 'Neos\EventSourcing\EventStore\Storage\Doctrine\DoctrineEventStorage'
          storageOptions:
            eventTableName: 'some_package_custom_events'
```

in the EventStoreFactory we have:
```php
$storageInstance = $this->objectManager->get($storageClassName, $storageOptions);
```

This case would no longer work with this change, right?

albe:
> 1. This is a bugfix, but it's targeted for `5.3`. By intention?

Yep, Flow is 5.3+ bugfix support :)

> 2. I think this is quite a breaking change (if I got it right)

This could be, so any input from possibly breaking use-cases helps understand this better!
Though, Christian and me agree that from an objective POV if a factory is configured, it should have full responsibility and partially autowiring would only lead to weird/hard to fix edge-cases.

But let's look:

> For the Neos.EventSourcing Package I prepared the following:
> 
> Objects.yaml:
> 
> ```yaml
> Neos\EventSourcing\EventStore\Storage\Doctrine\DoctrineEventStorage:
>   arguments:
>     1:
>       value: []
>     2:
>       object: 'Neos\EventSourcing\EventStore\EventNormalizer'
>     3:
>       object:
>         factoryObjectName: 'Neos\EventSourcing\EventStore\Storage\Doctrine\Factory\ConnectionFactory'
> ```

So, the `DoctrineEventStorage` itself is instanciated by the ObjectManager directly, but the 3rd argument is instanciated through a factory. Hence this change only affects the 3rd argument, i.e. the connection.

> This case would no longer work with this change, right?

As far as I can see this should still work perfectly fine. Only iff the Connection class instanciated from the ConnectionFactory would expect constructor arguments that the Factory itself doesn't provide (are expected to be autowired), then this would possibly cause an error now.

rkrahnen:
Hey guys, any news? I'd really appreciate it if this could be fixed soon, because I really need it. :-)

albe:
Any approvals appreciated :)

bwaidelich:
Hi guys and sorry for the delay (vacation).
I need to get into this one again, but by reading the change makes sense of course. But a test would be crucial IMO

fcool:
What exactly would you expect from a test, which not yet is covered? I'm currently preparing a test case for the broken scenario - but I do not see, what else is missing, if I search trough `\Neos\Flow\Tests\Functional\ObjectManagement\DependencyInjectionTest`

bwaidelich:
@fcool Probably I miss something, but this change only contains 3 lines that fix the bug. I was missing a corresponding test that fails without that fix

fcool:
Ah fine! This one will be ready in a couple of minutes.

fcool:
Providing a testcase which already would break compile team would have been easier, but quite ugly I guess. So best I can provide is a test case for the ConfigurationBuilder which will exactly fail with the observed exception and will not, if Albes fix is in place.
#2029

albe:
friendly ping @bwaidelich ;)

PR #2003, opened 2020-05-16
BUGFIX: Fix typo

albe:
Well... sorry 😲


albe:
PS: For reference, introduced with https://github.com/neos/flow-development-collection/commit/2438e9993e9f9120e98a2f0804b10e2b6245d618#diff-67f427fb2f1f4b763c2e27a4a4029d4c cc @bwaidelich ;P
But I also totally missed that on review

PR #2004, opened 2020-05-16
BUGFIX: Array.indexOf() returns an integer with asssociative arrays

daniellienert:


albe:
Hm, in JavaScript `indexOf` does not work on an object (which is the rough equivalent of an assoc array in PHP). Also, what information does it carry that 'foo' is at the position 3 in an associative array, which I can't index into with that position? So... I'm a bit undecided on just pretending the array is not assoc or throwing.

daniellienert:
Unlike javascript, in PHP array values can be addressed by is associative key and its position. So IMHO passing it through `array_values` leads to more consistent result.
The php implementation here uses `array_search` which returns the key and not the position, so if you pass an associative array, it just fails due to the ìnt` type constraint. That should be improved.

> Also, what information does it carry that 'foo' is at the position 3 in an associative array, which I can't index into with that position?

AFAIK we have no other method currently to check if a value is contained in an array. This is what I tried to use it for.

daniellienert:
Gna, the array_values would change the behavior ;(

Neos.Eel/Tests/Unit/Helper/ArrayHelperTest.php
```diff
@@ -165,7 +165,8 @@ public function indexOfExamples()
         return [
             'empty array' => [[], 42, null, -1],
             'array with values' => [['a', 'b', 'c', 'b'], 'b', null, 1],
-            'with offset' => [['a', 'b', 'c', 'b'], 'b', 2, 3]
+            'with offset' => [['a', 'b', 'c', 'b'], 'b', 2, 3],
+            'associative' => [['a' => 'el1', 'b' => 'el2'], 'el2', null, 1]
```
albe, line 169:
```suggestion
            'associative' => [['a' => 'el1', 'b' => 'el2'], 'el2', null, 1],
            'associative with offset' => [['a' => 'el1', 'b' => 'el2'], 'el2', 1, 1]
```
This should then also hold to keep in line with non-assoc behaviour.

Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -157,7 +157,7 @@ public function indexOf(array $array, $searchElement, $fromIndex = null): int
         if ($fromIndex !== null) {
             $array = array_slice($array, $fromIndex, null, true);
         }
-        $result = array_search($searchElement, $array, true);
+        $result = array_search($searchElement, array_values($array), true);
```
albe, line 160:
```suggestion
        $result = array_search($searchElement, $array, true);
        if (is_string($result)) {
            return array_search($result, array_keys($array), true) + (int)$fromIndex;
        }
```

PR #2005, opened 2020-05-16
BUGFIX: ActionResponse contains negotiated media type as content-type

albe:
This sets the negotiated media type from the Controller in the `ActionResponse`.

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/AbstractController.php
@@ -113,6 +113,7 @@ abstract class AbstractController implements ControllerInterface
         if ($mediaType === null) {
             $this->throwStatus(406);
         }
+        $this->response->setContentType($mediaType);
         if ($request->getFormat() === '') {
             $this->request->setFormat(MediaTypes::getFilenameExtensionFromMediaType($mediaType));
         }
--- a/Neos.Flow/Tests/Functional/Mvc/AbstractControllerTest.php
+++ b/Neos.Flow/Tests/Functional/Mvc/AbstractControllerTest.php
@@ -75,4 +75,13 @@ class AbstractControllerTest extends FunctionalTestCase
         $response = $this->browser->request('http://localhost/test/mvc/abstractcontrollertesta/forward?actionName=fourth&passSomeObjectArguments=1&arguments[nonObject1]=First&arguments[nonObject2]=42');
         self::assertEquals('fourthAction-First-42-Neos\Error\Messages\Message', $response->getBody()->getContents());
     }
+
+    /**
+     * @test
+     */
+    public function responseContainsNegotiatedContentType()
+    {
+        $response = $this->browser->request('http://localhost/test/mvc/abstractcontrollertesta/second');
+        self::assertEquals('text/plain', $response->getHeaderLine('Content-Type'));
+    }
 }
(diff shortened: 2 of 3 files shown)


bwaidelich:
Can we figure out _where_ this was lost and add a test case to the `AbstractControllerTest`?

albe:
I can not find any trace of that ever being in place, at least it's also not there in 4.3:
https://github.com/neos/flow-development-collection/blob/4.3/Neos.Flow/Classes/Mvc/Controller/AbstractController.php#L122-L128

I'm just wondering why that was never the case/never noticed. Guess because the webservers in use just do a very good job at providing sane defaults for the ContentType?

As to why this is targeted at 6.0: That's because the MVC response only carries the method to set a content-type since 6.0 and the PSR refactoring. Before the MVC response was a plain dumb container for a string content: https://github.com/neos/flow-development-collection/blob/4.3/Neos.Flow/Classes/Mvc/Response.php - so it was not even easily possible to set the Content-Type at this point.

I'll add a test case to https://github.com/neos/flow-development-collection/blob/6.0/Neos.Flow/Tests/Unit/Mvc/Controller/AbstractControllerTest.php to check the negotiated type is set in the response

bwaidelich:
Thanks for double checking!

> I'm just wondering why that was never the case/never noticed. Guess because the webservers in use just do a very good job at providing sane defaults for the ContentType? 

Yes, probably.. weird.

I would suggest to make it a BUGFIX anyways (it can still be targeted for 6.0)

> I'll add a test case to https://github.com/neos/flow-development-collection/blob/6.0/Neos.Flow/Tests/Unit/Mvc/Controller/AbstractControllerTest.php to check the negotiated type is set in the response

Aweseome, thanks!

bwaidelich:
..thinking about it once again: The negotiated content type refers to the _Request_ and it should not automatically set the response content type IMO

albe:
It's the negotiated media type of the accepted response types requested against the supported types of the controller. So that's exactly what the response should return, no?

```php
$mediaType = MediaTypeHelper::negotiateMediaType(MediaTypeHelper::determineAcceptedMediaTypes($request->getHttpRequest()), $this->supportedMediaTypes);
```

bwaidelich:
> It's the negotiated media type of the accepted response

ah, yes of course. You're right!

albe:
Note: The test nicely fails with an empty content-type without the fix in the first commit.

bwaidelich (approved):
Makes sense (by reading). thanks for taking care!

kitsunet:
This introduced a serious regression with the way responses are merged. With this a ContentType will ALWAYS be set in the ActionResponse which means the if conditions while merging inside the ActionResponse will always be true so no overwriting of the content type can happen from repsonses generated in a view for example. Eg. Fusion responses that override the content type will not be able to do that.

bwaidelich:
> This introduced a serious regression with the way responses are merged

Damn, I had a somewhat weird feeling and shouldn't have added my +1 – even if just "by reading"

kitsunet:
If we want to do that we should do it AFTER the view is rendered and then check if we have a content type set and only if not do this.

albe:
Ugh, didn't predict this :( Yeah, guess the solution is to set it after controller invocation if no content-type is set on the response yet.

albe:
Okay, this is not so simple because:
- `ActionResponse::$contentType` is by default uninitialized
- the getter is typehinted to `: string` (which is wrong, because it could be null)
- the `mergeIntoParentResponse()` method explicitly checks for `null` (also `applyToHttpResponse()`, but `mergeIntoComponentContext()` only checks for "truthy")

I suggest fixing the getter with null coalescing operator and just checking for empty string after invoking the action. Then maybe revisit this class in master

PR #2008, opened 2020-05-21
BUGFIX: reflectionService if more than one implementation class is available

Kleisli:
To be honest, I don't know exactly what I did, but it seems reasonable, when I look at the code, and (more important) it fixed my issue I had in NEOS:
Besides the Neos-Backend-Users I derived a custom Frontend-User-Class by extending Neos/Neos/../User, therefore my class is also implementing the Neos\ContentRepository\Domain\Model\UserInterface, but then the ReflectionService failed to fetch the correct implementation class when I tried to log into the backend. (Error "Class '' does not exist") because there was no value in $this->classReflectionData[ProxyInterface::class][self::DATA_INTERFACE_IMPLEMENTATIONS]

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -19,7 +19,6 @@ use Neos\Cache\Frontend\FrontendInterface;
 use Neos\Cache\Frontend\StringFrontend;
 use Neos\Cache\Frontend\VariableFrontend;
 use Neos\Flow\Core\ApplicationContext;
-use Neos\Flow\ObjectManagement\Proxy\ProxyInterface;
 use Neos\Flow\Package;
 use Neos\Flow\Package\PackageManager;
 use Neos\Flow\Persistence\RepositoryInterface;
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -435,14 +435,14 @@ class ReflectionService
         if (count($classNamesFound) === 1) {
             return $classNamesFound[0];
         }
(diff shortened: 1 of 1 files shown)


albe:
> To be honest, I don't know exactly what I did

To be honest, me neither 😄 and I also don't know what the `ProxyInterface::class` in that code is supposed to handle.

> Besides the Neos-Backend-Users I derived a custom Frontend-User-Class by extending

I'm not sure that is the way things are supposed to work. Maybe someone else from the team can explain better or correct me, but I think the point of the `UserInterface` is to be able to fully replace the BE `Neos\User` via `Objects.yaml` configuration. But your case sounds as if both diverging implementations of it should exist at the same time.
So AFAIS the correct way to do what you want would be to fully create your own FE User that **does not implement** `UserInterface`. If you need to, it would be better to just c&p the existing neos User class.

/cc @bwaidelich @kitsunet

Kleisli:
Thank you for your response. :)

I was actually trying to just use my custom User class, as I always did when I was working with Flow only, but then I got an error when using the default PersistedUsernamePasswordProvider
`An account with the identifier "till@kleisli.ch" exists, but the corresponding party is not a Neos User.`
So I thought I'm supposed to extend /Neos/Neos/../User when I'm coding in the neos context and want to create frontend users.

But may be I did it wrong, and it is still a bug. ;)

bwaidelich:
Could someone create an issue for that that explains _expected_ vs_ actual_ behavior?

kdambekalns (changes requested):
Please create an issue describing the problem (as suggested in an earlier comment). I doubt the change in this PR is a proper solution for the problem you experience.

kitsunet:
As there wasn't an update here for quite some time and it's still unclear what the actual issue is I am closing this for now as the change can't be merged as is. The implementation is correct, as it needs to provide the proxied implmenetation of the interface if there is one. The change would lead to randomly returning the proxied class or it#s original. I assume the (unclear) issue should have been solved by configuring the implementation for the interface that is to be used in the project via Objects.yaml.

Issue #2009, opened 2020-05-22
Serialization is not repeatable


fcool:
### Description

DependencyInjection build by the `\Neos\Flow\ObjectManagement\DependencyInjection\ProxyClassBuilder` leads to unrepeatable serialization.
In the second call all related entities are lost in the serialized form.
This espacially happens if you use an object in some kind of deferred workflow (for example using the functionality of `Flowpack.JobQueue.Common` and having the same object in a session, where it gets serialized again)
Those errors are tremendous hard to debug, as its only missing in the serialized form, so it will cause trouble in the disconnected process, which uses unserialize - and now misses all related data. As session savement is done after the response is done it gets worse, as there is no possibility to tell the user, that something gone wrong.

### Steps to Reproduce
(See the provided test case #2010)

1. Create an object having a property with an already persisted entity
2. serialize it twice
3. unserialize the second serializedData 

#### Expected behavior

I expect serialize($object) to be repeatable without leading to different results

### Affected Versions

Flow: 4.3 - 6.2

fcool:
Let me know, if I should retarget for version 5.3. As the change is mergeable in all majors I targeted 4.3 as problems in serialization often leads to quite unexpected attac vectors.

sorenmalling:
@albe This issue can be closed now thanks your merging?

PR #2010, opened 2020-05-22
BUGFIX: Allow objects to be serialized twice without loosing their relations to persisted properties

fcool:
This should solve the issue described in #2009


albe:
wait, needs to target 5.3...

fcool:
Is there a a way to modify this pull request? Or is it as simple as create a new one?

albe:
ah, damn travis... why can't you just rebuild correctly after base branch change?

fcool:
One thing, which might be a reason to keep it in 4.3 is, that this way it might get possible to remove objects from the session. Which might have impacts on the authentication processes - even if I can only think of "logging somebody out" and not gaining additional privileges. Some kind of "logout CSRF" ;)

albe:
Another test with /rebase

albe:
If this was a security issue, we should probably handle it a bit differently ;) I'd rather target 5.3 with this

Issue #2011, opened 2020-05-22
Serialization changes the original object


fcool:
### Description

If you serialize an object containing related entities in a multivalued property (array or similar), bringing this object to serialization it will remove those entities from the object during serialization, which then will stop you from using the properties value after the serialization.

### Steps to Reproduce

See testcase in #2012 

### Affected Versions

Flow: 4.3 - 6.2

fcool:
Let me know, if I should retarget for version 5.3. As the change is mergeable in all majors I targeted 4.3 as problems in serialization often leads to quite unexpected attac vectors.

albe:
Yeah, I guess this also should be retargeted to 5.3 unless someone can find a security issue in overwriting the property?

fcool:
I guess retargeting would be the easiest - if someone has concerns - he has still 6 months to find a reason for a backport.

I will come with a more detailed answer to your concern in the new pull request

fcool:
It's this time again... as the fix has been reverted, it was a matter of time, when I will again run into this.
This is obviously today. :(

PR #2012, opened 2020-05-22
BUGFIX: Do not remove related entities during persistence

fcool:
Expected behavior by serialize is not to change the object to serialize but to provide
a textual representation. If your object contained persisted entities in a multivalued
property it had been silently removed.

Solves #2011


Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -131,7 +131,6 @@ private function Flow_searchForEntitiesAndStoreIdentifierArray($path, $propertyV
                 'identifier' => $identifier,
                 'entityPath' => $path
             ];
-            $this->$originalPropertyName = Arrays::setValueByPath($this->$originalPropertyName, $path, null);
```
albe, line 134: TBH I'm super unsure about the (side) effects of removing this line. I do see the test change below covers a sensible case though...

albe: /cc @bwaidelich @kdambekalns does one of you have a gut-feeling about this, or even an informed opinion?

fcool: Argh! Sorry, I did promise to provide a rebased patch. But building one, I testet, if there would be differences. And yes there are.
Arrays containing Entities on some level would not get serialized correctly any longer, if my observation is correct. But I did not yet find the time to find a solution I would acceppt myself.

fcool: My next try would be to create some new array structure, referencing the array and a property path to temporarily save the identifiers and use this information during wakeup. This would allow the object not to be "changed" (but to extend) during serialization and at least allows that access after serialization is possible, without breaking, what we already have.

albe: Thanks for the clarification :)

kdambekalns:
This should be updated to target 5.3, as 4.3 is EOL. Also, what is the status?

sorenmalling:
Hi @fcool 

Will you provide a update on this - how did you test go? Can you provide a updated PR if needed?

albe:
ping @fcool - last chance to get this into 5.3-6.2 before they go into security-only

fcool:
Thanks for the warning. ;)

How much time is left?

albe:
I'd like to do the release sometime this week. Latest on friday

albe:
D'uh, guess needs a rebase... can take care next week or so, but right now the release has priority

albe:
Closing in favor of #2517 which is rebased and only contains the actual fix commit :)

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
kdambekalns: I _think_ the unset was in place to make sure persisting worked. The actual entities need to be encoded, but keeping them means they _get serialized as well_, doubling data and potentially having bad side-effects. I can imagine this even dates back to our own persistence layer before Doctrine 2… So, it might not be needed anymore.

That being said, the whole concept of `Flow_searchForEntitiesAndStoreIdentifierArray` might be useless by now?!

PR #2013, opened 2020-05-23
BUGFIX: Allow objects to be serialized twice without loosing their relations to persisted properties

### LINKED ISSUE #2009: Serialization is not repeatable

albe:
Resolves #2009

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
@@ -228,6 +228,7 @@ class ProxyClassBuilder
                 $propertyVarTags[$propertyName] = isset($varTagValues[0]) ? $varTagValues[0] : null;
             }
             $code = "        \$this->Flow_Object_PropertiesToSerialize = array();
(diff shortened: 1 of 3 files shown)


albe:
Retargeted follow-up for #2010

albe:
Merging as of my previous approval and build passing.

PR #2014, opened 2020-05-23
FEATURE: Add HttpController that uses requests and response as action arguments

mficzel:
Each controller action that is configured via routing will get the current ActionRequest passed
which allows to acces the arguments from the routing. The result is expected to be either a
PSR7 Response or an ActionResponse. A fallback for string responses is in place to make
debugging easier.

This allows to work with flow routing and controlers in a very similar way to symfony routing and controllers. Also this allows to avoid property mapping and validation for controller actions if one likes to work on the raw requests. Also no more view is instantiated by default.

Questions for now:
1. Should it be supported that Actions recieve PSR Requesrts or ActionRequets?
2. Is it a wise idea to Support Psr7/Response and ActionResponse?
3. Does it make sense that this still is in the MVC part of Flow or should be moved?
4. Should the HttpController become the base class for the ActionController eventually?

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/HttpController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/HttpController.php
@@ -1,0 +1,120 @@
+<?php
+namespace Neos\Flow\Mvc\Controller;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Http\Component\ReplaceHttpResponseComponent;
+use Neos\Flow\Mvc\ActionRequest;
+use Neos\Flow\Mvc\ActionResponse;
+use Neos\Flow\Mvc\Exception\InvalidActionVisibilityException;
+use Neos\Flow\Mvc\Exception\NoSuchActionException;
+use Neos\Flow\ObjectManagement\ObjectManagerInterface;
+use Neos\Flow\Reflection\ReflectionService;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * An HTTP based multi-action controller.
+ *
+ * The action specified in the given ActionRequest is dispatched to a method in
+ * the concrete controller whose name ends with "*Action". If no matching action
+ * method is found, the action specified in $errorMethodName is invoked.
+ *
+ * This controller does not apply any mapping and istead passes the request as argument
+ * to the action and expects an ActionResponse or a Psr7/ResponseInterface in return.
+ */
+class HttpController extends AbstractController
+{
+    /**
+     * @Flow\Inject
+     * @var ObjectManagerInterface
+     */
+    protected $objectManager;
+
+    /**
+     * @Flow\Inject
+     * @var ReflectionService
+     */
+    protected $reflectionService;
+
+    /**
+     * Name of the action method
+     *
+     * @var string
+     */
+    protected $actionMethodName;
+
+    /**
+     * Handles a request. The result output is returned by altering the given response.
+     *
+     * @param ActionRequest $httpRequest The request object
+     * @param ActionResponse $response The response, modified by this handler
+     * @return void
+     * @api
+     */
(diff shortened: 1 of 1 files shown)


albe:
I see the idea there, but I have a few concerns introducing this concept in this specific way:
- we have two different controller types, making things at that level a bit more complex - when use which controller for which scenario?
- the controller doesn't really do much, other than invoking controller methods to handle a PSR request
- the controller extends `Mvc\AbstractController` which is just wrong unless there is an underlying abstract concept we'd need to extract
- the concept of `doSomething(Request): Response` is exactly what PSR middlewares (and up to now Http Components) do - so you could achieve something similar by creating one of those and then having some simple logic inside to resolve the request to a handler = controller action method. So it is not clear why/if this should happen inside a base controller class. At minimum we need to discuss this further
- do we really want to support this low-level handling (PSR req/res) at a seemingly abstracted level (controller) at all? I would probably rather rethink the MVC stack a bit, to better support cases where the given requirements are lacking - as you stated "avoid property mapping and validation"

mficzel:
@albe I do not mean to merge this in this structure. Indeed this does not belong into the mvc 
part. To do this properly routing and dispatching should be extracted from mvc and work with .

Most things can be done psr middlewares but not everything as convenient. Especially we cannot use the uri-builder to create uris for controller actions without routing.

In the end i can imagine having a router/dispacher component in the middleware chain that mainly resolves the controller and action to pass the request to. Then the Mvc/ActionController could build upon this to add property mapping, validation, view resolving and view-rendering.

I can only say that when i worked with symfony the simple routing and request/response controllers feels very nice and really easy to get to concept wise. I would love to say "You can use Flow with property Mapping or with plain Requests" 

Possible also the actionController could be extended to detect when method has an Psr\Request or ActionRequest as argument and simply populate this in that case direct return of Response objects by actions is already there.

sorenmalling:
> In the end i can imagine having a router/dispacher component in the middleware chain that mainly resolves the controller and action to pass the request to. Then the Mvc/ActionController could build upon this to add property mapping, validation, view resolving and view-rendering.

I love this idea and can see it fit in the PSR-15 middleware movement of Flow.

albe:
> Possible also the actionController could be extended to detect when method has an Psr\Request or ActionRequest as argument and simply populate this in that case direct return of Response objects by actions is already there.

That could indeed work. But then again that probably mixes things up a bit too much, so a dedicated `HttpController` is more explicit about expectations probably.

> Especially we cannot use the uri-builder to create uris for controller actions without routing.

Yep. We should try to make UriBuilding/Routing fully independent of MVC stack. MVC should only have convenience accessors to that.
Same would be nice with PropertyMapping, but that one is a bit more complicated, because it depends on type-reflection of the controller, so is bound to that.

> In the end i can imagine having a router/dispacher component in the middleware chain that mainly resolves the controller and action to pass the request to. Then the Mvc/ActionController could build upon this to add property mapping, validation, view resolving and view-rendering.

Right - my current vision for how the stack should look is something like that:

```
Bootstrap 
-> (Http)RequestHandler 
-> PSR-15 Middlewares 
-> Dispatch middleware (resolve controller and invoke it's dispatch loop) 
-> build ActionRequest + invoke: 
-> "MVC" middlewares handle(ActionRequest request, MvcMiddleware next): ActionResponse [including property mapping, validation, resolve view etc.] 
-> Controller.Action(...) 
-> back into RequestHandler and sendResponse()
```
That way, you could just swap out/reconfigure the MVC "middlewares" to have different convenience/abstraction features. Using your HttpController would still work, by simply routing to a Controller that does not invoke the MVC middlewares.

mficzel:
@albe almost 100% agree. How about having a `route` and a `dispatch` middleware. That would allow to put a middleware in between and modify the generated route if needed. Especially if we 
communicate trough the chain with http-headers. I see `mvc` more as a layer that is called by dispatch and not an own middleware. 

I will close this WIP pr now and  prepare a community package ideas. If this proves useful i will bring up this concept again after we converted the http chains to middlewares and maybe managed to extract routing and dispatching alongside.

BTW: I will go with the following action signature that also adds some type clarity for that:
```
public function exampleAction (ActionRequest $request, ActionResponse $response): ?ResponseInterface
```

mficzel:
For anyone interested, the code from this experiment endet in a community package for now to gain real world experience: https://github.com/PackageFactory/PackageFactory.BaseJump

PR #2015, opened 2020-05-25
TASK: Test against PHP 8

kdambekalns:


composer.json
```diff
@@ -6,7 +6,7 @@
     ],
     "type": "neos-package-collection",
     "require": {
-        "php": "^7.2",
+        "php": "^7.2 || ^8.0",
```
albe, line 9: needs to be changed in all the single packages too.

albe:
Guess we need to do one of two things to get this run tests:
- merge as is into master, then fix things failing ASAP
- adjust our travis script, so it does not install dev-master first (https://github.com/neos/flow-development-collection/blob/master/.travis.yml#L65-L67)

I'd go with #1 - wdyt @kdambekalns ?

kdambekalns:
Yeah, let's go with the merge… after 6.3 is branched. Then fix everything that needs fixing—and if viable we can backport compatibility as far as possible.

composer.json
kdambekalns: It is changed, no? What did I miss?

albe: Wasn't the case in the first commit :)

kdambekalns:
Haha! So, there is no version of Doctrine ORM that supports PHP 8 (yet).

kdambekalns:
- `doctrine/dbal` runs with PHP 8 as of 2.12.0: https://github.com/doctrine/dbal/pull/4347
- `doctrine/orm` is on the way: https://github.com/doctrine/orm/pull/8303

PR #2016, opened 2020-05-27
TASK: Change used variable TRUE into true

kaystrobach:
**What I did**

**How I did it**

**How to verify it**

**Checklist**


kaystrobach:
a general recomendation is to provide a way to directly change files from readthedocs. the link points to the readonly repository ... this makes no sense ...

kaystrobach:
> > provide a way to directly change files from readthedocs. the link points to the readonly repository
> 
> Oh, indeed - never noticed that and that's bad! But not sure how to fix this, since RTD is generated from the readonly repos... maybe we need to change that?

isn't there a setting for that? 

https://github.com/readthedocs/sphinx_rtd_theme/issues/314#issuecomment-244648533

albe:
Nice! That could work indeed - we should try out :) Thanks!

PR #2020, opened 2020-05-30
TASK: Rename Doctrine\Common\Persistence -> Doctrine\Persistence

albe:
This fixes psalm errors due to the namespace renaming of various doctrine classes in doctrine/orm 2.7.3 release. The old class name was an alias since 2.7, so this should not break anything. 🤞


albe:
See also https://github.com/doctrine/orm/commit/8a9954e46c1a0d93679bc2c19339f680c5e5affd#r39497749

PR #2029, opened 2020-06-19
TASK: Adding a test case for breaking object configuration if argumen…

fcool:
…ts of object constructed by factory cannot be autowired


albe:
Thanks! :)

PR #2032, opened 2020-06-19
BUGFIX: Make php binary setting check use realpath

albe:
The check if the current executing php binary path as set in `PHP_BINARY` matches the given setting in `phpBinaryPathAndFilename` was case-sensitive. This check is only done when the setting is specified and should guarantee that the path specified refers to the same binary as the one executing the current script. On Windows, the filesystem is case-insensitive by default though (Windows 10 has options to make it case-sensitive) and the chance that someone accidentially gets the casing of the path wrong on windows is (arguably) much more likely than someone having different versions of php installed on a case-sensitive filesystem in pathes `/foo/PHP/php` and `/foo/php/php` respectively.
~Even if that would be the case, if the two versions do not match the check at `ensureWebSubrequestsUseCurrentlyRunningPhpVersion` would still throw an error with an even more specific message.~

So this will prevent irritating errors on Windows in the form of 
image

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -855,7 +855,7 @@ class Scripts
             return;
         }
 
-        if (strcmp(PHP_BINARY, $configuredPhpBinaryPathAndFilename) !== 0) {
+        if (strcasecmp(PHP_BINARY, $configuredPhpBinaryPathAndFilename) !== 0) {
             throw new FlowException(sprintf('You are running the Flow CLI with a PHP binary different from the one Flow is configured to use internally. ' .
                 'Flow has been run with "%s", while the PHP version Flow is configured to use for subrequests is "%s". Make sure to configure Flow to ' .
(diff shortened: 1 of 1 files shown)


albe:
> On a case-sensitive FS the check would now pass even when the case is wrong, so…

Correct, but:

> Even if that would be the case, if the two versions do not match the check at ensureWebSubrequestsUseCurrentlyRunningPhpVersion would still throw an error with an even more specific message.

So the change (for case-sensitive FS) in behavior would be:
- if the two paths have different PHP versions the error message talks about version difference rather than path name difference
- if the two paths have same PHP versions, no error is thrown any more

The latter should be fine IMO, because what we actually *need* is same PHP version to guarantee consistent behaviour?

albe:
> if you configure the path just consider the case whether it matters or not.

Yeah, though at this point the issue is on windows this specifically basically never matters, but suddenly appears to (though in reality it still doesn't).

> Relying on a subsequent more strict check seems "dangerous" to me since the setting could be used elsewhere

It's not about the setting, just about the check :) But as noticed above in the edit, the "subsequent" check is conditionally orthogonal to the one in question. I'll take another look, why we don't even check the PHP version for CLI requests, but only the `PHP_BINARY` path. I think it has to do with the fact we actively use `PHP_BINARY` at some place.

PR #2034, opened 2020-06-24
BUGFIX: Inject settings into the PackageManager during bootstrap

### LINKED ISSUE #1981: While kickstarting a new package: Trying to access array offset on value of type null in PackageManager
### Description

When I kickstart a new package using the kickstarter it fails.

### Steps to Reproduce

1. Clean Flow development distro with collection on master
2. Run ./flow package:create --package-key Foo.Bar

#### Expected behavior

The notification the package is created

#### Actual behavior

Notice: Trying to access array offset on value of type null in /.../Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php line 373

  Code: 1
  File: Packages/Framework/Neos.Flow/Classes/Error/ErrorHandler.php
  Line: 81

### Affected Versions

Flow: 6.0, 6.1, 6.2 and master

I tried to find what happens, it seems like the PackageManager instance which is injected using \Neos\Flow\Command\PackageCommandController::injectPackageManager has $settings with value NULL.

radmiraal:
The injection of the Flow settings into the PackageManager was accidentally
removed from the Scripts::initializeConfiguration() method. This caused
various issues in the PackageManager for example during the kickstart of
new packages using the kickstarter commands.

The accidental removal happened in b71e1924e423b73814b09552d34291002e4a18f1

Fixes #1981

**What I did**

**How I did it**

**How to verify it**
One of the commands failing is the kickstart package command:

    ./flow kickstart:package Acme.Package

Without this change applied the command will fail and injectSettings of the PackageManager will not be called.

**Checklist**


albe:
Great catch!

Seems we need to fix some tests though:

> Notice: Undefined index: package in /home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php on line 120
PHP Notice:  Undefined index: log in /home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Flow/Classes/Core/Booting/Scripts.php on line 239
Notice: Undefined index: log in /home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Flow/Classes/Core/Booting/Scripts.php on line 239
PHP Fatal error:  Uncaught Error: Class name must be a valid object or a string in /home/travis/build/neos/flow-development-distribution/Packages/Framework/Neos.Flow/Classes/Core/Booting/Scripts.php:241

radmiraal:
that's weird, on my machine I only get 4 warnings about phpunit deprecations...

radmiraal:
ah, at least the 'undefined index package' makes sense, will have another look

albe (approved):

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -211,6 +211,12 @@ public static function initializeConfiguration(Bootstrap $bootstrap)
             $configurationManager->refreshConfiguration();
         }
 
+        // Manually inject settings into the PackageManager as the package manager is excluded from the proxy class building
+        $flowSettings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'Neos.Flow');
+        if (is_array($flowSettings)) {
```
sorenmalling, line 216: If setting a for some reason is empty and we end up not injecting anything, will that leave the packageManager in a similar state with no settings, since it's excluded (as commented above this code)?
So if the settings is not an array, should we pass a empty array to ensure that eh injectSettings method is being called so the packagemanager get its setting a property set no matter what?

sorenmalling (changes requested):
See comment about the manual injection of settings

radmiraal:
Hi Soren,

That seems true at first, but keep in mind that this change is reintroducing code which was removed that did not do that if I remember correctly.

Injecting an empty array would fix that $settings is an array no matter what, which would imho better be solved by giving it a default value in the PackageManager itself I think.




Outlook for Android<https://aka.ms/ghei36> downloaden

________________________________

albe:
Agree to what @radmiraal says - making `settings` always an array would be useful, but is not strictly what this PR is about. So @sorenmalling do you agree and can revisit your review?

sorenmalling (approved):

PR #2035, opened 2020-06-24
TASK: Throw exception when accidentially trying to advice a private method

### LINKED ISSUE #1982: Pointcut method designator does not allow to intercept private methods
### Description

Pointcut method designator does not allow to intercept private methods,
only protected or public methods can be intercepted.

This is actually somewhat documented at
https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/AspectOrientedProgramming.html#method

since it only mentions public and protected methods but not private methods.

### Steps to Reproduce

1. create a class with a private method that is called.
2. Create an Aspect with an annotation like `@Flow\AfterReturning("method(Foo\Bar\Service\MyService->getResponse())")` pointing to the private method.

#### Expected behavior

The Aspect to be called when the private method returns something.

#### Actual behavior

Nothing happens.

albe:
Related to #1982

### DIFF
--- a/Neos.Flow/Classes/Aop/Pointcut/PointcutMethodNameFilter.php
+++ b/Neos.Flow/Classes/Aop/Pointcut/PointcutMethodNameFilter.php
@@ -126,6 +126,10 @@ class PointcutMethodNameFilter implements PointcutFilterInterface
                     return false;
                 }
                 break;
+            default:
+                if ($methodDeclaringClassName !== null && $this->reflectionService->isMethodPrivate($methodDeclaringClassName, $methodName)) {
+                    throw new Aop\Exception('The method "' + $methodDeclaringClassName + ':' + $methodName + '" adviced by "' + $this->aspectClassName + ':' + $this->pointcutMethodName + '" is private. Private methods can not be adviced. Please change the scope to at least protected or explicitly specify the visibility you want to match.', 1593013922);
+                }
         }
 
         $methodArguments = ($methodDeclaringClassName === null ? [] : $this->reflectionService->getMethodParameters($methodDeclaringClassName, $methodName));


bwaidelich:
image

oops, I'm sorry.. reverting

Neos.Flow/Classes/Aop/Pointcut/PointcutMethodNameFilter.php
```diff
@@ -126,6 +126,10 @@ public function matches($className, $methodName, $methodDeclaringClassName, $poi
                     return false;
                 }
                 break;
+            default:
+                if ($methodDeclaringClassName !== null && $this->reflectionService->isMethodPrivate($methodDeclaringClassName, $methodName)) {
+                    throw new Aop\Exception('The method "' + $methodDeclaringClassName + ':' + $methodName + '" adviced by "' + $this->aspectClassName + ':' + $this->pointcutMethodName + '" is private. Private methods can not be adviced. Please change the scope to at least protected or explicitly specify the visibility you want to match.', 1593013922);
```
bwaidelich, line 131: `Aop\Exception` does not resolve to a FQCN, shouldn't this just be `Exception`?

Also `$this-> aspectClassName` and `$this-> pointcutMethodName` are not defined in this class..!?

bwaidelich (changes requested):
Maybe I'm doing something wrong (or I messed up while trying to rebase your branch) but I can't get this to run.

Also a test would be nice ofc.. I'll try to add one (only if it works out easily)

Neos.Flow/Classes/Aop/Pointcut/PointcutMethodNameFilter.php
albe: D'oh, yeah - I had this inside the `Pointcut` first, then thought that to be the wrong place and just moved it into the `PointcutMethodNameFilter`. Looking at it again, I'm not sure anymore about this. Exception should probably be thrown in the constructor instead of on match, then rethrown in the `PointcutExpressionParser` with the `sourceHint`?

bwaidelich:
It also fails for private constructors currently, even if they are not adviced explicitly

Neos.Flow/Classes/Aop/Pointcut/PointcutMethodNameFilter.php
bwaidelich: Mh, not sure, that whole proxy building is still Greek to me..
I was about to ditch my version for yours but #2131 seems to do the trick just fine – shall we just go for that?

albe: Looks a bit cleaner than the throw+catch/rethrow that would be necessary here (to get the source hints), so sure. Let's ditch this aproach

albe:
Closing in favor of #2131

PR #2036, opened 2020-06-25
BUGFIX: Disallow advising of private methods

bwaidelich:
Intercepting private methods does not work since the original method
is no longer accessible if it is overridden in the proxy class.

But it was still possible to construct this scenario that would then
fail without exception.

With this change an exception is thrown whenever a private method
is target of a AOP designator.

Resolves: #1982


bwaidelich:
Draft because it misses some doc comments and tests

albe:
@bwaidelich Should we go on with this route and close #2035 or the other way around?

bwaidelich:
@albe Thanks for the reminder, I added the ticket to the board and will look into it again asap

bwaidelich:
Closing in favor of #2131

Issue #2039, opened 2020-07-01
PdoBackend::setup() will throw exception if tables exist already


nlx-lars:
# Description

When running either `./flow cache:setupall` while having more than one PdoBackend configured with the same dataSourceName or `./flow cache:setup` against a PdoBackend whose tables already exist, an exception is thrown.

### Steps to Reproduce

1. Configure Cache with PdoBackend
2. Run `./flow cache:setupall`
3. Run `./flow cache:setupall` again

#### Expected behavior

The second `cache:setup` or `cache:setupall` command should have 0 as exit code and maybe a notice that the tables exist already.

#### Actual behavior

The PdoBackend tries to create the cache tables twice, leading to an error. This is problematic when using `cache:setupall` in a deployment process.

### Affected Versions

Neos: 4.2+

Flow: 5.2+

nlx-lars:
@bwaidelich, @albe and @kitsunet what do you think about this?

The PdoBackend has a `tableExists` method that could be used to determine if tables have to be created in the setup. If the do already exist, a Notice could be added to the Result.

bwaidelich:
Thanks for the input. In a previous version I used the schema manager to get the tables in sync with the SQL, but that was reverted because it introduced a new dependency. Checking if the table exists at all seems sensible to me

nlx-lars:
@bwaidelich nice, I'll submit a PR then! 👍

PR #2040, opened 2020-07-01
BUGFIX: Check if PdoBackend cache tables exist in setup

nlx-lars:
Running the setup for a PdoBackend multiple times should
not lead to an error, similar to other Backends.

Resolves: #2039

**What I did**
Allow PdoBackend::setup() to be called multiple times.

**How I did it**
Check if cache tables exist in PdoBackend::setup() before creating them.

**How to verify it**
See #2039

### DIFF
--- a/Neos.Cache/Classes/Backend/PdoBackend.php
+++ b/Neos.Cache/Classes/Backend/PdoBackend.php
@@ -569,6 +569,10 @@ class PdoBackend extends IndependentAbstractBackend implements TaggableBackendIn
             $result->addNotice(new Notice('SQLite database tables are created automatically and don\'t need to be set up'));
             return $result;
         }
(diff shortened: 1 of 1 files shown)


albe:
Reopened just to get checks to run again. Can be merged once they are green.

daniellienert:
Transferred here from original description by @nlx-lars:

**Checklist**

- [ ✅  ] Code follows the PSR-2 coding style
- [ ❗  ] Tests have been created, run and adjusted as needed. - See below
- [ ✅  ] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)

There aren't any tests for the `setup()` method yet, so I am not entirely sure how to tacke this. The tests run on sqlite and the setup method won't do anything in that case.

PR #2047, opened 2020-07-04
 BUGFIX: Set session cookie after flash message storage

### LINKED ISSUE #1807: FlashMessages are not displayed with redirect

albe:
The FlashMessages component by default uses the SessionStorage, which depends on sending the session cookie through the SessionResponse component. However, the FlashMessages component was configured to be run **after** the session component. That way, unless a session was (still) open, the `Set-Cookie` was not sent and hence the flash message got lost.
This was primarily visible after logging out and redirecting the user with a flash message.

Thanks @astehlik for finding the root cause!

Resolves #1807

### DIFF
--- a/Neos.Flow/Classes/Http/Client/InternalRequestEngine.php
+++ b/Neos.Flow/Classes/Http/Client/InternalRequestEngine.php
@@ -20,10 +20,12 @@ use Neos\Flow\Http\Component\ComponentChain;
 use Neos\Flow\Http\Component\ComponentContext;
 use Neos\Flow\Http;
 use Neos\Flow\Mvc\Dispatcher;
+use Neos\Flow\Mvc\FlashMessage\FlashMessageService;
 use Neos\Flow\Mvc\Routing\RouterInterface;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
 use Neos\Flow\Security\Context;
 use Neos\Flow\Session\SessionInterface;
+use Neos\Flow\Session\SessionManager;
 use Neos\Flow\Tests\FunctionalTestRequestHandler;
 use Neos\Flow\Validation\ValidatorResolver;
 use Psr\Http\Message\ResponseFactoryInterface;
@@ -124,10 +126,13 @@ class InternalRequestEngine implements RequestEngineInterface
         } catch (\Throwable $throwable) {
             $componentContext->replaceHttpResponse($this->prepareErrorResponse($throwable, $componentContext->getHttpResponse()));
         }
-        $session = $this->bootstrap->getObjectManager()->get(SessionInterface::class);
+        $session = $objectManager->get(SessionInterface::class);
         if ($session->isStarted()) {
             $session->close();
         }
+        // FIXME: ObjectManager should forget all instances created during the request
+        $objectManager->forgetInstance(SessionManager::class);
+        $objectManager->forgetInstance(FlashMessageService::class);
         $this->persistenceManager->clearState();
         return $componentContext->getHttpResponse();
     }
--- a/Neos.Flow/Configuration/Settings.Http.yaml
+++ b/Neos.Flow/Configuration/Settings.Http.yaml
@@ -68,7 +68,7 @@ Neos:
               position: 'start'
               component: 'Neos\Flow\Session\Http\SessionResponseComponent'
             'flashMessages':
-              position: 'after setSessionCookie'
+              position: 'before setSessionCookie'
               component: 'Neos\Flow\Mvc\FlashMessage\FlashMessageComponent'
             'poweredByHeader':
               component: 'Neos\Flow\Http\Component\PoweredByComponent'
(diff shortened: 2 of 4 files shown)


Neos.Flow/Classes/Http/Client/InternalRequestEngine.php
```diff
@@ -124,10 +126,13 @@ public function sendRequest(ServerRequestInterface $httpRequest): ResponseInterf
         } catch (\Throwable $throwable) {
             $componentContext->replaceHttpResponse($this->prepareErrorResponse($throwable, $componentContext->getHttpResponse()));
         }
-        $session = $this->bootstrap->getObjectManager()->get(SessionInterface::class);
+        $session = $objectManager->get(SessionInterface::class);
         if ($session->isStarted()) {
             $session->close();
         }
+        // FIXME: ObjectManager should forget all instances created during the request
```
albe, line 133: This case really showed an issue we have with the internal request engine: In case a test is supposed to verify state is correctly rebuilt from an incoming request (e.g. session), the shared ObjectManager currently hides problems.
Probably some other issues are hidden that way and optimally the IRE should completely reset the ObjectManager to a clean state before sending the request.

There is currently no facility for this though, so as a narrow work-around we just forget the SessionManager (to fully get rid of the session so it will be rebuilt from the cookie) and FlashMessageService (to get rid of the cached storage instance, which holds a reference to the FlashMessageContainer).

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -68,7 +68,7 @@ Neos:
               position: 'start'
               component: 'Neos\Flow\Session\Http\SessionResponseComponent'
             'flashMessages':
-              position: 'after setSessionCookie'
+              position: 'before setSessionCookie'
```
albe, line 71: We could make this `start` and `setSessionCookie: position: 'after flashMessages'` but I'm unsure on the potential effects. Considering this works and all components are supposed to be replaced by PSR-15 middlewares anyway, guess this is the safer approach.

Neos.Flow/Tests/Functional/Mvc/ActionControllerTest.php
```diff
@@ -520,4 +530,34 @@ public function trustedPropertiesConfigurationDoesNotIgnoreWildcardConfiguration
...
+
+    /**
+     * @test
+     */
+    public function flashMessagesGetRenderedAfterRedirect()
+    {
+        $request = $this->serverRequestFactory->createServerRequest('GET', new Uri('http://localhost/test/mvc/actioncontrollertest/redirectWithFlashMessage'));
+        $response = $this->browser->sendRequest($request);
+
+        $sessionCookies = array_map(static function ($cookie) {
```
albe, line 542: Optimally the browser would take care of storing cookies and sending them with the following requests...

Neos.Flow/Tests/Functional/Mvc/ActionControllerTest.php
```diff
@@ -520,4 +530,34 @@ public function trustedPropertiesConfigurationDoesNotIgnoreWildcardConfiguration
...
+            return Cookie::createFromRawSetCookieHeader($cookie);
+        }, $response->getHeader('Set-Cookie'));
+        self::assertNotEmpty($sessionCookies);
+
+        $redirect = $response->getHeaderLine('Location');
+        self::assertNotEmpty($redirect);
+
+        $this->objectManager->forgetInstance(StandardController::class);
```
albe, line 550: Not sure this is really needed, but through the `AbstractController` it holds a reference to `ControllerContext` which holds a reference to `FlashMessageService`, so...

albe:
Note: The first commit contains a test that should show the error. The second commit contains the actual bugfix.
For the test to work, some adjustments to the InternalRequestEngine were necessary. See code comments for more details.

albe:
I'd like to take the right to merge this without approval, since the functional change is a super trivial setting order:
```
     position: 'after setSessionCookie'
=>
     position: 'before setSessionCookie'
```
and is covered by a test, plus it solves a super annoying issue we have laying around since 6.0. I'll take the blame if any.

Neos.Flow/Configuration/Settings.Http.yaml
bwaidelich: relative positioning (obviously) only works if the reference component is active, i.e. `after flashMessages` could behave inconsistent if the flashMessages component is disabled (I guess).
`start 10` and `start 20` could work as well, but I think it's fine the way it is now!

bwaidelich:
@albe Thanks a lot for taking care and sorry for the lack of (earlier) feedback!

albe:
It's a bit saddening sometimes, but I totally understand it and I'm not going to blame anyone for it :) We're all doing this in our free time after all.
In this case I just took it upon myself out of personal responsibility and impatience, as I missed the error in the FlashMessage PR back then, which I myself pushed forward.

PR #2049, opened 2020-07-06
TASK: Remove typo3 namespace replacements

albe:
There never was and never will be a typo3/flow version of 6.3+


albe:
Ah, forgot there are also those "replace" things in all the separate packages... meh. Will take care soon.

PR #2050, opened 2020-07-07
BUGFIX: Browser arguments are the request parsed body

albe:
Since 6.0 the Browser `$arguments` were accidentially moved to the request query parameters, while they should reflect the post body arguments.

The docblock even states that:
`@param array $arguments Arguments to send in the request body`

Note that this slightly changes behaviour if you adapted your usage of Browser since 6.0. If you want to send query parameters, you should instead provide an `Uri` instance with the parameters, as was already the case prior to 6.0.

### DIFF
--- a/Neos.Flow/Classes/Http/Client/Browser.php
+++ b/Neos.Flow/Classes/Http/Client/Browser.php
@@ -160,7 +160,7 @@ class Browser
         }
 
         if (!empty($arguments)) {
-            $request = $request->withQueryParams($arguments);
+            $request = $request->withParsedBody($arguments);
         }
         if (!empty($files)) {
             $files = UploadedFilesHelper::upcastUploadedFiles($files, $arguments);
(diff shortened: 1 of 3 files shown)


albe:
Note to self: I should probably add a test for this.

albe:
Looks good to me now. Test confirms the expected behaviour. Maybe we should mark this BUGFIX as "breaking" to make aware of the changed (=fixed) behaviour?

bwaidelich:
Damn, I think I was a bit too quick with that one..
Is `6.0` the right target branch and should this be marked "breaking"?

If we want to revert this one => #2071

albe:
Well, it was breaking when upgrading from 5.x -> 6.x until now - but unintentionally at that. That's why I think it should land in 6.0+ - but maybe it should be marked as breaking and be clearly mentioned, so that an upgrade from older 6.x is not so unexpectedly changing things again (if you never came from 5.x for example).

kdambekalns:
I'd say we keep it like this now. I adjusted the description slightly and marked it as breaking.

PR #2052, opened 2020-07-14
BUGFIX: Add TTL to tags in RedisBackend

nlx-lars:
Previously the tags did not expire with their key, so they were never removed from Redis. Now the same lifetime as their corresponding entries is added to tags set by the RedisBackend

This is a continuation of #1194. To prevent race conditions when two processes write to the same cache tags at the same time, the list of keys that will be written is [watched](https://redis.io/commands/watch) before and during the transaction. If the key is modified during that time, the transaction `exec` will fail and the TTL is calculated again (up to 4 retries).


nlx-lars:
TODO: The TTL of an existing cache tag needs to be extended when another key is set with the same tag.

nlx-lars:
This ***should*** work as expected now, but I did not try this out in a real project yet.

nlx-lars:
So I found one more problem here:

The following test fails (with and without my patch):
```
/**
 * @test
 */
public function entriesWith()
{
    $this->backend->set('entry1', 'foo', ['bar', 'baz'], 1);
    $this->backend->set('entry2', 'foo', ['baz'], 10);
    sleep(2);
    $this->assertFalse($this->backend->has('entry1'));

    $this->assertEmpty($this->backend->findIdentifiersByTag('bar'));
    $this->assertEquals(['entry2'], $this->backend->findIdentifiersByTag('baz'));
}
```

In the last line, `entry1` and `entry2` will be returned by `findIdentifiersByTag`. This is caused by the whole tag getting the higher lifetime (10) - which is fine. The problem is that the tag is a redis `set` and the individual members of a set cannot have their own ttl.
One way to circumvent this would be to filter the identifiers in the `findIdentifiersByTag()` method:

```
public function findIdentifiersByTag(string $tag): array
{
    $identifiers = $this->redis->sMembers($this->buildKey('tag:' . $tag));

    return array_values(array_filter($identifiers, function(string $entryIdentifier) {
        return $this->has($entryIdentifier);
    }));
}
```

kdambekalns:
Testing instructions from the origonal PR description:

---

**How to verify it**
Run tests or
* Setup Flow Cache that uses RedisBackend
* Set a cacheEntry with cacheTags and lifetime
* Enter `redis-cli`
* `KEYS *`
* Check TTL of key `TTL "Redis_Test:entry:foo"`
* Check TTL of tags `TTL "Redis_Test:tag:tag-1"`

**Checklist**

- [ ✅  ] Code follows the PSR-2 coding style
- [ ✅  ] Tests have been created, run and adjusted as needed
- [ ✅  ] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)

kdambekalns:
@nlx-lars you wrote

> The following test fails (with and without my patch):

But the tests pass… Could you check?

nlx-lars:
@kdambekalns I did not add that test to the code yet.

Do you want me to add it and implement the fix in `findIdentifiersByTag ` I proposed?

kdambekalns:
> @kdambekalns I did not add that test to the code yet.
> 
> Do you want me to add it and implement the fix in `findIdentifiersByTag ` I proposed?

Ah, ok… was a bit confusing (to me, at least.) Yes, please add the test.

Whether the fix is good depends on how expensive it is–after all a `has()` call is now done for every found entry…

nlx-lars:
@kdambekalns I added the test and my proposed fix, however I can't comment on how expensive it is.

nlx-lars:
Hi @kdambekalns, any news on this?

I tried to benchmark this a bit, and it looks like fetching entries by tags takes about twice as long now:

```php
use Neos\Cache\Frontend\StringFrontend;
use Neos\Flow\Utility\Algorithms;

class RedisCommandController extends CommandController
{

    private const COMMON_TAG = 'gh-2052';

    /**
     * @var StringFrontend
     */
    protected $cache;

    /**
     * @param int $count Amount of entries to set
     */
    public function benchmarkCommand(int $count = 20000): void
    {
        $this->cache->flush();

        $this->outputLine('Populating Redis with %d entries...', [$count]);

        $this->outputLine();
        $this->outputLine();

        $before = $this->printTime();

        for ($i = 0; $i < $count; $i++) {
            $this->cache->set(
                Algorithms::generateUUID(),
                'foo data',
                [
                    Algorithms::generateUUID(),
                    self::COMMON_TAG,
                ],
                300
            );

            if ($i % 5000 === 0) {
                $this->output('.');
            }
        }

        $this->outputLine();

        $this->printTime($before);

        $this->outputLine();
        $this->outputLine();

        $this->outputLine('Setup done...');

        $this->outputLine();
        $this->outputLine();

        $before = $this->printTime();
        $this->cache->getByTag(self::COMMON_TAG);
        $this->printTime($before);

        $this->outputLine();
        $this->outputLine();
    }

    private function printTime(?float $diffSince = null): float
    {
        $time = microtime(true);

        $this->outputLine('Microtime: ' . $time);
        if ($diffSince) {
            $this->outputLine('Took ' . ($time - $diffSince));
        }

        return $time;
    }

}
```

Results:

```
# without patch

20k
Microtime: 1612634608.3542
Microtime: 1612634608.7818
Took 0.42759203910828

20k
Microtime: 1612634636.6911
Microtime: 1612634637.094
Took 0.40289902687073


# with patch

20k
Microtime: 1612634923.5356
Microtime: 1612634924.2824
Took 0.74675703048706

20k
Microtime: 1612634947.1861
Microtime: 1612634947.9905
Took 0.804368019104


# without patch

100k
Microtime: 1612635658.2632
Microtime: 1612635660.2921
Took 2.02894282341

100k
Microtime: 1612635842.5964
Microtime: 1612635844.7226
Took 2.1261250972748


# with patch

100k
Microtime: 1612635217.0608
Microtime: 1612635220.849
Took 3.7882039546967

100k
Microtime: 1612635406.5575
Microtime: 1612635410.4321
Took 3.8746068477631
```

Since the problem with `findIdentifiersByTag` is not specific to this patch we could move it to a separate issue/pr?

kdambekalns:
Thanks! I looked at the numbers and the benchmark code, and I dare say the effect is not a real problem. The tested case fetches all cache entries (since they share a common tag), something that is rather articial (I would think.)

kdambekalns (approved):

kdambekalns:
Maybe @kitsunet and @skurfuerst could take another look here…

kaystrobach:
Would love to sponsor that if backported to flow 5.x as we have a serious problem with flow taking down redis in a project right now.

albe (commented):
I have basically zero knowledge about redis internals and no way to test this other than running the tests. Code looks sensible from a syntactic POV and is covered by a lot of new tests, so 🎉

albe (approved):

albe:
@kaystrobach is 5.3 good enough or would you need this in a lower version?

albe:
ping @kitsunet @skurfuerst - would also love to have some feedback by someone that has used the redis backend in some real world production before merging

kaystrobach:
5.3 is totally fine.

kaystrobach:
I think whats really the problem with the current implementation is, that the tag set entries grow way to fast and this takes down redis.

we need a way to clean redis entries and sets async to the normal frontend requests.

nlx-lars:
@kaystrobach IMO the `entries` set should be removed completely from the RedisBackend at some point. It's only there so the backend can implement `IterableBackendInterface`. I've never had to iterate over a cache in any real world projects. In fact, I actually removed the `entries` part in a customer project as that caused redis slowdowns and memory problems as you mentioned.

But let's not get off topic here, the `entries` problem should be it's own issue/PR. :)

kaystrobach:
I totally agree. The rest looks fine and improved the performance.

albe:
Hey everyone, after doing upmerges I noticed that psalm builds failed. After looking, one complaint is that `redis->watch()` takes a string, but we are providing an array.
See https://github.com/phpredis/phpredis#watch-unwatch

According to https://github.com/phpredis/phpredis/blob/develop/redis.c#L1360 it takes a list of strings

So maybe we just need to do `$this->redis->watch(...$keysToWatch)` to fix this? Can someone confirm?

PR #2055, opened 2020-07-18
TASK: Add documentation to model unidirectional OneToMany and ManyToMany

### LINKED ISSUE #2053: Improve "Getting Started" tutorial with modelling Post->Tags and Post->Comments relations
Modelling `OneToMany` relations within a single aggregate is one of the most non-obvious things in Flow (and Doctrine) and we never had a proper example around to link to whenever the question comes up (and it comes up frequently). The naive solution, as typically extrapolated from the existing Blog<->Post example, models this with another bidirectional `@ORM\OneToMany` relation, which will lead to a lot of pain down the road for various reasons.

Properly modelling this would involve a `@ORM\ManyToMany` and possibly a `@ORM\JoinTable(inverseJoinColumns={@ORM\JoinColumn(unique=true)})` to create a unidirectional OneToMany. This avoids having to upgrade the e.g. Tags/Comments to AggregateRoots themself, which does not fit the domain.
In theory, a `@ORM\OneToMany` relation should only ever be used on AggregateRoot to AggregateRoot relations and for other relations it should be elevated to `ManyToMany`. Maybe we could even provide some automatism or safeguard for this in the FlowAnnotationDriver (cc @kdambekalns)?

In a second step, it would be good to document how to build forms to properly deal with CRUDy OneToMany relations. There are some corner-cases, especially around adding/deleting elements to a collection relation.

albe:
This adds basic documentation for how to properly model unidirectional OneToMany relations in Flow and slightly extends the "Getting Started" guide.

Related to #2053

### DIFF
--- a/Neos.Flow/Documentation/TheDefinitiveGuide/PartII/ModelAndRepository.rst
+++ b/Neos.Flow/Documentation/TheDefinitiveGuide/PartII/ModelAndRepository.rst
@@ -702,7 +702,7 @@ Then adjust the post model code as follows:
 		 */
 		protected $blog;
 
-    ...
+		...
 
 		/**
 		 * @Flow\Validate(type="NotEmpty")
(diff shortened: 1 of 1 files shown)


albe:
This could be greatly simplified in master with #2056

albe:
@kdambekalns I'd like to get this in as is for now. I will add the two TODOs I had in the PR message to the issue to keep track and eventually follow up, but this definitely won't happen before the release and likeley not within the next 1-2 months.

kdambekalns (approved):

Issue #1395, opened 2018-09-21
Add documentation section for modelling unidirectional OneToMany relations


albe:
This has come up a lot of times already and it's one of the unintuitive things with Doctrine (and hence Flow) - if you model OneToMany relations in your Model, the Many side automatically becomes the "owning side" and hence needs a repository (so it becomes an Aggregate!) and is the side of the relation that you need to update when the relation changes.

So say you want to add one (or more) childs to a parent, you would need to do this:
```
$child = new Child();
$parent->addChild($child);
$child->setParent($parent); // Normally done inside addChild()
$childRepository->add($child);
```

Doing the more intuitive and from a modelling PoV "correct" `$parentRepository->update($parent)` would just lose the newly created child and not persist it.

The solution is to use a `ManyToMany` relation with a unique constraint.
See https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/unitofwork-associations.html#association-updates-owning-side-and-inverse-side
and https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/association-mapping.html#one-to-many-unidirectional-with-join-table

```
    /**
     * @var Collection<Child>
     * @ORM\ManyToMany
     * @ORM\JoinTable(inverseJoinColumns={@ORM\JoinColumn(unique=true)})
     */
    protected $children;
```

This is not mentioned anywhere and easily becomes a road block you hit, so it should be documented clearly somewhere.

albe:
See also https://stackoverflow.com/questions/7766999/doctrine-2-0-vs-2-1-cascade-remove-onetomany 
http://www.inanzzz.com/index.php/post/ur2x/bidirectional-one-to-many-cascade-remove-and-orphan-removal-operations-in-doctrine

albe:
Closing as of #2056 and #2055

PR #2056, opened 2020-07-18
!!!FEATURE: Allow unidirectional OneToMany relations

### LINKED ISSUE #2054: Make modelling unidirectional OneToMany relations easier
Inside a single aggregate `OneToMany` relations are normally best modelled unidirectionally. Bidirectional relations always are harder to manage correctly and can easily lead to unintentional traversal of entity hierarchies with all the drawbacks.
Since Doctrines `OneToMany` annotation is always bidrectional and also dictates the owning side of the relation (at the unexpected side from a modelling PoV), it is not straightforward to modell this correctly.

In Flow specifically, we try to follow DDD best practices in modelling and this means, that the aggregate root is the entry point and the entity that is sent to a repository to persist it and all its subentities. This can not be achieved with the `OneToMany` annotation when the one side is supposed to be closer to the root.
The correct way to annotate this is as follows:
```php
    /**
     * @ORM\ManyToMany(orphanRemoval=true)
     * @ORM\JoinTable(inverseJoinColumns={@ORM\JoinColumn(unique=true)})
     * @var Collection<Comment>
     */
```
The `orphanRemoval` is needed to correctly delete sub-entities, when they are removed from the relation collection. For `OneToMany` we currently set this by default, but not for `ManyToMany` (when it is used to model `OneToMany`). The `JoinTable` is required to make sure that it's impossible to accidentally relate two parents to the same sub-entity and effectively makes the `ManyToMany` a `OneToMany`.

IMO we should allow the user to annotate such a relation simply as:
```php
    /**
     * @ORM\OneToMany
     * @var Collection<Comment>
     */
```
We would need to "transform" a `OneToMany` into a `ManyToMany` inside the `FlowAnnotationDriver` when we can detect that this is an unidirectional relation. Since a bidirectional `OneToMany` needs to contain a `mappedBy=` attribute, this can be easily inferred (without `mappedBy` and trying to actually map it as `OneToMany` doctrine would throw an exception).

Optimally, Doctrine would just make this transformation for unidirectional `OneToMany`, but it's unlikely this will ever be supported. From an ORM PoV this makes some sense even, since the relation annotation maps to the way the relation is implemented in the relational database. `ManyToMany` is implemented with a join table, while `OneToMany` (or `ManyToOne`) are implemtend with a single foreign key column. Switching this based on directionality of a relation could be considered "unexpected" behaviour for an ORM. However Flow modelling happens on a higher abstraction level and hence this implementation detail is more a hinderance than "expected".

Related to #2053 and #1395

albe:
Inside a single aggregate `OneToMany` relations are normally best modelled unidirectionally. Bidirectional relations always are harder to manage correctly and can easily lead to unintentional traversal of entity hierarchies with all the drawbacks.
Since Doctrines `OneToMany` annotation is always bidrectional and also dictates the owning side of the relation (at the unexpected side from a modelling PoV), it is not straightforward to modell this correctly.

In Flow specifically, we try to follow DDD best practices in modelling and this means, that the aggregate root is the entry point and the entity that is sent to a repository to persist it and all its subentities. This can not be achieved with the standard doctrine `OneToMany` annotation when the one side is supposed to be closer to the root.

This change allows the user to annotate such a relation simply as:
```php
    /**
     * @ORM\OneToMany
     * @var Collection<Comment>
     */
```

This is done by remapping `OneToMany` annotations without a `mappedBy` as `ManyToMany` with a unique constraint.

Resolves #2054 #1395

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
@@ -658,7 +658,36 @@ class FlowAnnotationDriver implements DoctrineMappingDriverInterface, PointcutFi
                     $mapping['orderBy'] = $orderByAnnotation->value;
                 }
 
-                $metadata->mapOneToMany($mapping);
+                if ($oneToManyAnnotation->mappedBy !== null) {
+                    $metadata->mapOneToMany($mapping);
+                } else {
+                    // Transform a unidirectional OneToMany annotation to a ManyToMany with unique constraint
+                    /** @var ORM\JoinTable $joinTableAnnotation */
+                    if ($joinTableAnnotation = $this->reader->getPropertyAnnotation($property, ORM\JoinTable::class)) {
+                        $joinTable = $this->evaluateJoinTableAnnotation($joinTableAnnotation, $property, $className, $mapping);
+                    } else {
+                        $joinColumns = [
+                            [
+                                'name' => null,
+                                'referencedColumnName' => null,
+                            ]
+                        ];
+
+                        $joinTable = [
+                            'name' => $this->inferJoinTableNameFromClassAndPropertyName($className, $property->getName()),
+                            'joinColumns' => $this->buildJoinColumnsIfNeeded($joinColumns, $mapping, $property, self::MAPPING_MM_REGULAR),
+                            'inverseJoinColumns' => $this->buildJoinColumnsIfNeeded($joinColumns, $mapping, $property)
+                        ];
+                    }
+                    foreach ($joinTable['inverseJoinColumns'] as &$inverseJoinColumn) {
+                        if (!isset($inverseJoinColumn['unique'])) {
+                            $inverseJoinColumn['unique'] = true;
+                        }
+                    }
+
+                    $mapping['joinTable'] = $joinTable;
+                    $metadata->mapManyToMany($mapping);
+                }
             } elseif ($manyToOneAnnotation = $this->reader->getPropertyAnnotation($property, ORM\ManyToOne::class)) {
                 if ($this->reader->getPropertyAnnotation($property, ORM\Id::class) !== null) {
                     $mapping['id'] = true;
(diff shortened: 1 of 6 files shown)


bwaidelich:
Visually the PR looks good (follows CGL, contains a test, ...) and the ticket description is great (I'd suggest to copy part of it to the PR description so it ends up in the Changelog though). But since I try to avoid the doctrine relations to model aggregates or entities in general I'm really not into that topic and have to pass on this one.

I just wonder if that shouldn't be marked as a breaking change since it potentially affects the behavior of your entities and will create new migrations (right?)

albe:
> I just wonder if that shouldn't be marked as a breaking change since it potentially affects the behavior of your entities and will create new migrations

It should theoretically not affect behavior of the entities - but it is true that it changes the schema/requires a migration, so this should indeed be considered breaking. Thanks for the heads-up!

> Instead of the mention in the ticket
Yeah, the mention happened when I was still thinking about this theoretically. Then a day later I just quickly tried it out in code and pushed it to have something visible to discuss on.

bwaidelich:
> Question: This should then be targeted on next-major I guess.

Yes, I think so.. But someone that uses the ORM annotations more can hopefully provide some better feedback

> Then a day later I just quickly tried it out in code and pushed it to have something visible to discuss on.

That's the way, thanks for taking care! :)

albe:
I tagged a couple of people that have some good knowledge about the annotation stuff and the 6.3 RMs - so it's up to them to decide if this should go into 6.3 or 7.0; I'm fine with either.

albe:
Thanks to @kdambekalns this will now also resolve #1395 :)

sorenmalling:
> Thanks to @kdambekalns this will now also resolve #1395 :)

Is it possible do describe a migration path or any changes needed to current code to benefit from this? A section like

**What to do to your already running Flow project**

sorenmalling (approved):
This is so awesome and nice! Works flawlessly - try this package for an example

https://github.com/sorenmalling/Vendor.OneToManyUnidirectional

sorenmalling:
Can we get past the styleci error and have this in core?

albe:
> Is it possible do describe a migration path or any changes needed to current code to benefit from this? A section like
> 
> **What to do to your already running Flow project**

That's a good question. I guess it depends a little bit on how you modelled the relation previously. If it was done with an `@ORM\ManyToMany(...)` annotation, then the only change is to replace that annotation with `@ORM\OneToMany` (not necessary, but hey, it's easier to understand).
If the relation was modelled as `OneToMany` however and the related entity was promoted to an AggregateRoot therefore, then quite some code would need to be deleted (the repository, some code in the model and he controller), a migration generated and the database adjusted (needs to insert rows into the new join table to relate the entities).

So, it's not so super straightforward, that it could be covered by a code-migration. But maybe this could be something for a blog post (or recipe on docs.neos.io), where a previous code and the steps to "upgrade" it is shown and discussed.

PR #2064, opened 2020-07-24
FEATURE: Provide method to retrieve all authenticated accounts from security context

### LINKED ISSUE #2063: Add SecurityContext::getAuthenticatedAccounts method
Related to https://github.com/neos/neos-development-collection/pull/2981#issuecomment-634641904

albe:
This is needed when implementing e.g. multiple authentication providers and needing to check for an account with specific properties like in https://github.com/neos/neos-development-collection/pull/2981 without querying tokens manually.

Resolves #2063

TODO: Add tests

### DIFF
--- a/Neos.Flow/Classes/Security/Context.php
+++ b/Neos.Flow/Classes/Security/Context.php
@@ -453,14 +453,12 @@ class Context
     }
 
     /**
-
      * Returns the account of the first authenticated authentication token.
      * Note: There might be a more currently authenticated account in the
-     * remaining tokens. If you need them you'll have to fetch them directly
-     * from the tokens.
-     * (@see getAuthenticationTokens())
+     * remaining tokens. If you need them use getAuthenticatedAccounts() instead.
+     * (@see getAuthenticatedAccounts())
      *
-     * @return Account The authenticated account
+     * @return Account|null The authenticated account
      */
     public function getAccount()
     {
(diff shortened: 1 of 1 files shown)


bwaidelich:
I'm not sure, but this might open up a pandora box..

Having multiple authenticated accounts is possible technically but it doesn't really make sense logically IMO. And it might introduce additional complexity to the userland.

I would really like to go with a proper implementation (as suggested with https://discuss.neos.io/t/rfc-accountinterface-and-authentication-rework/4966) for 7.0 (and potentially go with an intermediate solution without adding new API methods).

In short: The idea would be to introduce *realms*  and probably something like:
```php
$neosAccount = $this->securityContext->getAuthenticatedAccountByRealm(AuthenticationRealm::for('Neos.Neos:Backend');
```

albe:
> without adding new API methods

Well, it's deliberately not annotated `@api` ;)

```php
$neosAccount = $this->securityContext->getAuthenticatedAccountByRealm(AuthenticationRealm::for('Neos.Neos:Backend');
```

That looks sane and if I understand the concept correctly, it would replace the current `getAccountByAuthenticationProviderName()` - so maybe we can indeed skip this addition, adjust the neos side to use `getAccountByAuthenticationProviderName()` right away instead (see the `Edit` in https://github.com/neos/neos-development-collection/pull/2981#issuecomment-663688627)

bwaidelich:
> Well, it's deliberately not annotated @api ;)

True, But `Context::getAccount()` is neither and it is used in a lot of userland code :)

I think, the requirements of neos/neos-development-collection#2981 can be solved with my suggested rework (hopefully for the major coming this december). For the work around I would indeed prefer to use `getAccountByAuthenticationProviderName()` (if that works) or keep the work around in the neos code

PR #2081, opened 2020-08-06
BUGFIX: Correct indention for throwables config

markusguenther:
The configuration of the throwable exceptions should not be part of the `psr3` level. And the default config did not work since 6.0. The method `initializeExceptionStorage` still expects the configuration to be at the path `Neos.Flow.log.throwables`.

Fixes: #2082

### DIFF
--- a/Neos.Flow/Configuration/Settings.Log.yaml
+++ b/Neos.Flow/Configuration/Settings.Log.yaml
@@ -52,8 +52,8 @@ Neos:
                 maximumLogFileSize: 10485760
                 logFilesToKeep: 1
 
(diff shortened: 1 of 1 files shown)


markusguenther:
Ah nice did not even know that we have such tests 👍

kdambekalns:
> But we have a test that should run the schema check - in that case why didn't it fail (or fails now)?

From looking at the schema… it does not *require* `throwables` to exist, so that passes. And the `psr3` part of the schema then sees `throwables` as "just another logger", and that allows basically anything below it, too?

Issue #2112, opened 2020-09-08
Unused MD5 in ResourceMetaDataInterface


rkrahnen:
The `ResourceMetaDataInterface` contains two content hashes, a SHA1 and a MD5. While the SHA1 is used for the whole resource management, the MD5 is used only once when creating a `StorageObject` for a `PersistentResource`, so only for copying it without any further usage.

In my opinion the MD5 should be marked as deprecated in Flow 6.3 and removed in Flow 7.0.

sorenmalling:
@rkrahnen Where in the code have you found the md5 methods/property being used?

Can you try and remove them and run the tests to see if any errors occures?

rkrahnen:
There are three places where `setMd5` is used:
1. \Neos\Flow\ResourceManagement\Storage\FileSystemStorage::getObjectsByCollection
2. \Neos\Flow\ResourceManagement\Storage\PackageStorage::createStorageObject
3. \Neos\Flow\ResourceManagement\Storage\WritableFileSystemStorage::importTemporaryFile

The second and third usage both set a new value by using `md5_file()`, whereas the first usage is the one mentioned in my first comment and is the only place where `getMd5` is used.

So for me it seems like it is set but never used (ignoring the copying one).

Sadly I am currently not able to run the tests, sorry.

sorenmalling:
Try and remove the usage and commit a PR - then the tests will automatically run 👍

rkrahnen:
Hopefully I did it right: https://github.com/neos/flow-development-collection/pull/2113

PR #2113, opened 2020-09-08
!!! BUGFIX: Remove unused MD5 in ResourceMetaDataInterface

rkrahnen:
Removes the `md5` property from persistent resources that is obsolete since
we introduced a `sha1` field.

Note:

This change is marked breaking because it removes the two API methods
`PersistentResource::getMd5()` and `PersistentResource::setMd5()`.
In practice this should never be an issue really.

Instead of `PersistentResource::getMd5()` one could use `PersistentResource::getSha1()`
in order to retrieve a unique hash for a file.
In case a MD5 hash is required, the following one can do `md5(stream_get_contents($resource->getStream()))`

Fixes: #2112

### DIFF
--- a/Neos.Flow/Migrations/Mysql/Version20200908155620.php
+++ b/Neos.Flow/Migrations/Mysql/Version20200908155620.php
@@ -1,8 +1,8 @@
 <?php
 namespace Neos\Flow\Persistence\Doctrine\Migrations;
 
-use Doctrine\DBAL\Migrations\AbstractMigration;
 use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
 use Neos\Flow\Core\Bootstrap;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
 use Neos\Flow\ResourceManagement\PersistentResource;
@@ -52,24 +52,18 @@ class Version20200908155620 extends AbstractMigration
     {
         $resourceRepository = Bootstrap::$staticObjectManager->get(ResourceRepository::class);
         $persistenceManager = Bootstrap::$staticObjectManager->get(PersistenceManagerInterface::class);
-        $filename = FLOW_PATH_DATA . 'tmp_md5_migration';
 
         $iterator = $resourceRepository->findAllIterator();
         foreach ($resourceRepository->iterate($iterator) as $resource) {
             /* @var PersistentResource $resource */
-            $stream = $resource->getStream();
-            if (is_resource($stream)) {
-                $file = fopen($filename, 'w');
-                stream_copy_to_stream($resource->getStream(), $file);
-                fclose($file);
-
-                $this->connection->executeUpdate(
-                    'UPDATE neos_flow_resourcemanagement_persistentresource SET md5 = ? WHERE persistence_object_identifier = ?',
-                    [md5_file($filename), $persistenceManager->getIdentifierByObject($resource)]
-                );
-
-                unlink($filename);
+            if (!is_resource($resource->getStream())) {
+                continue;
             }
+
+            $this->connection->executeUpdate(
+                'UPDATE neos_flow_resourcemanagement_persistentresource SET md5 = ? WHERE persistence_object_identifier = ?',
+                [md5(stream_get_contents($resource->getStream())), $persistenceManager->getIdentifierByObject($resource)]
+            );
         }
     }
 }
--- a/Neos.Flow/Migrations/Mysql/Version20200908155620.php
+++ b/Neos.Flow/Migrations/Mysql/Version20200908155620.php
@@ -1,0 +1,75 @@
+<?php
+namespace Neos\Flow\Persistence\Doctrine\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Neos\Flow\Core\Bootstrap;
+use Neos\Flow\Persistence\PersistenceManagerInterface;
+use Neos\Flow\ResourceManagement\PersistentResource;
+use Neos\Flow\ResourceManagement\ResourceRepository;
+
+class Version20200908155620 extends AbstractMigration
+{
+    /**
+     * @return string
+     */
+    public function getDescription()
+    {
+        return 'Drop "md5" column of the "resource" table';
+    }
+
+    /**
+     * @param Schema $schema
+     * @return void
+     * @throws \Doctrine\DBAL\Migrations\AbortMigrationException
+     */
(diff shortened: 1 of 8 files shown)


rkrahnen:
Okay, I think the test fails because of the remaining `$md5` property and the column not being nullable. So I have to add the migration first, but therefor I need some information.

The up migration is easy, I just drop the column in the right resource table after checking which table exists like `Version20180622074421` for example.
But what about the down migration? I could just add the column again, but it would be empty for existing entries. Would that be fine for you as this is something for the next major, or do you have any other idea?

sorenmalling:
Maybe this can help


https://github.com/neos/flow/blob/master/Migrations/Mysql/Version20141015125841.php

There is a hook after up and down you can use to insert values

sorenmalling:
@kitsunet @bwaidelich Main thing to review is the way of the postDown - the way resources is retrieved, we get's a instance of the resourceRepository. Is it safe to do so at this point, in this way?

Neos.Flow/Migrations/Mysql/Version20200908155620.php
```diff
@@ -0,0 +1,75 @@
...
+     * Add md5 content hash for resources.
+     *
+     * @param Schema $schema
+     * @return void
+     */
+    public function postDown(Schema $schema)
+    {
+        $resourceRepository = Bootstrap::$staticObjectManager->get(ResourceRepository::class);
+        $persistenceManager = Bootstrap::$staticObjectManager->get(PersistenceManagerInterface::class);
+        $filename = FLOW_PATH_DATA . 'tmp_md5_migration';
```
bwaidelich, line 55: This should be a temporary path IMO, see `Environment->getPathToTemporaryDirectory()`

Neos.Flow/Migrations/Mysql/Version20200908155620.php
```diff
@@ -0,0 +1,75 @@
...
+
+        $iterator = $resourceRepository->findAllIterator();
+        foreach ($resourceRepository->iterate($iterator) as $resource) {
+            /* @var PersistentResource $resource */
+            $stream = $resource->getStream();
+            if (is_resource($stream)) {
```
bwaidelich, line 61: not important, but this would make it easier to follow IMO:
```php
if (!is_resource($stream)) {
  continue;
}
```

Neos.Flow/Migrations/Mysql/Version20200908155620.php
```diff
@@ -0,0 +1,75 @@
+<?php
+namespace Neos\Flow\Persistence\Doctrine\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
```
bwaidelich, line 4:
```suggestion
use Doctrine\Migrations\AbstractMigration
```
image

Neos.Flow/Migrations/Mysql/Version20200908155620.php
```diff
@@ -0,0 +1,75 @@
...
+                $file = fopen($filename, 'w');
+                stream_copy_to_stream($resource->getStream(), $file);
+                fclose($file);
+
+                $this->connection->executeUpdate(
+                    'UPDATE neos_flow_resourcemanagement_persistentresource SET md5 = ? WHERE persistence_object_identifier = ?',
+                    [md5_file($filename), $persistenceManager->getIdentifierByObject($resource)]
+                );
+
+                unlink($filename);
```
bwaidelich, line 71: Do we really need a file here?
This seems to work just fine:

```php
/* @var PersistentResource $resource */
foreach ($resourceRepository->iterate($iterator) as $resource) {
    if (!\is_resource($resource->getStream())) {
        continue;
    }
    $this->connection->executeUpdate(
        'UPDATE neos_flow_resourcemanagement_persistentresource SET md5 = ? WHERE persistence_object_identifier = ?',
        [md5(stream_get_contents($resource->getStream())), $persistenceManager->getIdentifierByObject($resource)]
    );
}
```

bwaidelich (changes requested):
Thanks for this!
I added a few suggestions. Apart from those this change should be marked breaking since it (strictly speaking) removes a method that was marked `@api`

Neos.Flow/Migrations/Mysql/Version20200908155620.php
rkrahnen: Removed file.

Neos.Flow/Migrations/Mysql/Version20200908155620.php
rkrahnen: Done.

Neos.Flow/Migrations/Mysql/Version20200908155620.php
rkrahnen: Done.

Neos.Flow/Migrations/Mysql/Version20200908155620.php
rkrahnen: Good catch, changed it.

rkrahnen:
@bwaidelich I agree with you about the change being breaking. If there is anything I can do, let me know.

bwaidelich (approved):
Thanks for the update.
I like the current solution but I wonder if we should change the target branch to `master` since it is potentially a breaking change

PS: I already took the freedom to adjust the PR description to mark it breaking

sorenmalling:
>  I like the current solution but I wonder if we should change the target branch to `master` since it is potentially a breaking change

I support that idea 👍

rkrahnen:
Are we waiting for @kitsunet to approve this as well?

bwaidelich:
> Are we waiting for @kitsunet to approve this as well?

No, not really. But the CI tests fail and IMO you should switch the base branch to `master` (maybe that resolves the broken CI)

You can switch the branch by editing the PR:
image

sorenmalling (approved):

sorenmalling:
Approved, if branch is changed to "master" and we can get the tests running 👍

kdambekalns (approved):
Looks good! I restarted the TravisCI jobs, seem to pass now.

rkrahnen:
@bwaidelich : Are you taking care of marking the required parts as deprecated in 6.3?

bwaidelich:
> Are you taking care of marking the required parts as deprecated in 6.3?

Which parts?

bwaidelich:
Ah, I see what you mean, marking `get/setMd5()` deprecated for 6.3.x!? I have too much on my list right now, but I can take care next week if nobody else steps up

bwaidelich:
Ah, heck.. => #2115

rkrahnen:
Awesome, thanks!

PR #2131, opened 2020-09-15
BUGFIX: Disallow advising of private methods

bwaidelich:
Intercepting private methods does not work since the original method
is no longer accessible if it is overridden in the proxy class.

But it was still possible to construct this scenario that would then
fail without exception.

With this change an exception is thrown whenever a private method
is target of a AOP designator.

Fixes: #1982


albe (approved):
Looks good by reading. A quick (unit) test would be super cool on top - guess we won't allow proxying privates in the foreseeable future.

bwaidelich:
Ah, I hit the merge button before I saw your comment regarding a test.. 
But since it's effectively just a nicer exception message I can't be bothered to write one to be honest ;)

PR #2132, opened 2020-09-16
BUGFIX: Improve EmailAddressValidator

### LINKED ISSUE #1227: EmailAddressValidator does not work with IDN domains
### Description

An e-mail address using an IDN domain name is not considered a valid e-mail address.

### Steps to Reproduce

1. Create a scenario using an EmailAddressValidator, e.g. an e-mail field with `Neos.Form`.
2. Enter an e-mail address with an IDN domain, e.g. `test@neös.io`.
3. Trigger validation

#### Expected behavior

The e-mail address is valid.

#### Actual behavior

A validation error is thrown: `Please specify a valid email address.`

### Affected Versions

Flow: all versions

### Background

The EmailAddressValidator is using `filter_var` as follows:

```filter_var($emailAddress, FILTER_VALIDATE_EMAIL)```

This does not support IDN domain names. As of PHP 7.1, it seems to be possible to set a filter flag `FILTER_FLAG_EMAIL_UNICODE`, but this is not a short-term solution for Flow installations using PHP < 7.1.

### PR DESCRIPTION
**Pull request by kdambekalns**
This no longer uses `filter_var()`, which does a rather mediocre job.

Fixes #1227

### CODE REVIEW

**Comment by albe** [comment]
Love to have this old issue about the e-mail validator tackled, but two things:
- does this change behaviour of the validator? If so, it's breaky, right? Maybe make it a feature for next major instead then?
- this package looks well maintained by a quick glance, but I still have a bit of a hard time pulling in (new) 3rd parties 🙈 is there any way we can make it an optional dependency? Probably not...

**Comment by daniellienert** [comment]
@albe: Regarding the package, looks pretty well maintained! https://packagist.org/packages/egulias/email-validator
@kdambekalns: There is this `EmailAddressValidatorTest` - would be great to see a test with IDN domains

**Review by daniellienert — COMMENTED** [review]
@albe: Regarding the package, looks pretty well maintained! https://packagist.org/packages/egulias/email-validator
@kdambekalns: There is this EmailAddressValidatorTest - would be great to see a test with IDN domains

**Comment by kdambekalns** [comment]
* Breaking: I wouldn't know how. An email address is either valid, or not. And if it is wrongly marked as invalid now, that would be nuisance, so it can only get better.
* Package: Anyone using SwiftMailer has to pass that validation anyway, it's what they use. And with 101.339.878 installs according to Packagist it won't go away anytime soon.
* Test: Absolutely! Thanks for the reminder.

**Comment by albe** [comment]
> And if it is wrongly marked as invalid now, that would be nuisance, so it can only get better.

I was more thinking the other way around, e-mails that passed before (wrongly) that would now be rejected, so e.g. accounts suddenly can't login/update/dunno what

> Anyone using SwiftMailer has to pass that validation anyway, it's what they use.

That's great to know! This alleviates the concerns quite a bit!

**Comment by sorenmalling** [comment]
> I was more thinking the other way around, e-mails that passed before (wrongly) that would now be rejected, so e.g. accounts suddenly can't login/update/dunno what

Mark it as a important (!) note to read, when we create release notes. 

I don't consider that breaking, more than optimizing of code and functionality in general.

File: composer.json [hunk 1]
```diff
@@ -27,6 +27,7 @@
         "symfony/console": "^4.2",
         "neos/composer-plugin": "^2.0",
         "composer/composer": "^1.9",
+        "egulias/email-validator": "^2.1",
```
sorenmalling, line 30: Is the requirement suppose to go both places? Wouldn't the requirement in neos/flow do the job correct?

kdambekalns: Yes, both places. The development collection needs those, too. See `ComposerManifestMerger.php`.

kdambekalns:
Now I am officially entering the grey area… added a `FEATURE` commit. 😬

We can split that, if you like.

albe (approved):
I approve as is, given the dependency is well maintained and a dependency of SwiftMailer which is very likely used for sending mails. About the added "feature" - as per strict convention it should end up in next major only, but then again it's just an optional setting to the implementation we use for this bugfix, so I could turn a blind eye on this.

sorenmalling:
> as per strict convention it should end up in next major only

And let's stick to that :) If you want to use this new emailaddressvalidator in a "older project", you can require the package and create the emailaddressvalidator class yourself and use that one :)

albe:
But then we'd have to retarget this whole PR to master, right?

Neos.Flow/Tests/Unit/Validation/Validator/CollectionValidatorTest.php
```diff
@@ -58,13 +57,13 @@ public function collectionValidatorFailsForAValueNotBeingACollection()
      */
     public function collectionValidatorValidatesEveryElementOfACollectionWithTheGivenElementValidator()
     {
-        $this->validator->_set('options', ['elementValidator' => 'EmailAddress', 'elementValidatorOptions' => []]);
-        $this->mockValidatorResolver->expects($this->exactly(4))->method('createValidator')->with('EmailAddress')->will($this->returnValue(new EmailAddressValidator()));
+        $this->validator->_set('options', ['elementValidator' => 'Integer', 'elementValidatorOptions' => []]);
+        $this->mockValidatorResolver->expects($this->exactly(4))->method('createValidator')->with('Integer')->willReturn(new IntegerValidator());
 
         $arrayOfEmailAddresses = [
```
daniellienert, line 63: Its not an array of MailAdresses anymore

daniellienert (commented):
Looks good (apart from the variable name nitpick). How do we decide on the version this is going to?

albe:
> How do we decide on the version this is going to?

I'd let @kdambekalns decide

kdambekalns:
Let's be correct and target master… What @sorenmalling said in https://github.com/neos/flow-development-collection/pull/2132/#issuecomment-698183162 is a good way for those who need that.

albe:
So, let's do this and have it in 7.0 :)

lorenzulrich:
Thanks Karsten!

PR #2134, opened 2020-09-18
!!!FEATURE: Add virtual object configurations for framework loggers

### LINKED ISSUE #2125: Make injecting the framework loggers more straightforward and document it
With the PSR-3 change, getting a hold of the system logger / security logger has changed and (deprecated) interfaces for b/c have been introduced:

```yml
# Deprecated as of Flow 6.0, you should inject Psr\Log\LoggerInterface
# or use the factory to get the "systemLogger"
Neos\Flow\Log\PsrSystemLoggerInterface:
  scope: singleton
  factoryObjectName: Neos\Flow\Log\PsrLoggerFactoryInterface
  factoryMethodName: get
  arguments:
    1:
      value: systemLogger
```

So, right now, the correct way to get the SystemLogger is:
```
@Flow\Inject
@var \Psr\Log\LoggerInterface
```
And for the security logger, you have to inject the `Neos\Flow\Log\PsrLoggerFactoryInterface` and manually fetch the logger with `->get('securityLogger')`, which is more than suboptimal.

However, virtual objects (#1964)  are a thing now and we have all the tools to make this more straightforward:
```yml
'Neos.Flow:SystemLogger':
  className: Psr\Log\LoggerInterface
  scope: singleton
  factoryObjectName: Neos\Flow\Log\PsrLoggerFactoryInterface
  factoryMethodName: get
  arguments:
    1:
      value: systemLogger

'Neos.Flow:SecurityLogger':
  className: Psr\Log\LoggerInterface
  scope: singleton
  factoryObjectName: Neos\Flow\Log\PsrLoggerFactoryInterface
  factoryMethodName: get
  arguments:
    1:
      value: securityLogger
```
should be added to the `Objects.yaml` configuration and documented on how to inject those:
```php
/**
 * @Flow\Inject(name="Neos.Flow:SystemLogger")
 * @var LoggerInterface
 */
protected $systemLogger;

/**
 * @Flow\Inject(name="Neos.Flow:SecurityLogger")
 * @var LoggerInterface
 */
protected $securityLogger;
```

albe:
With this, it is possible to inject the Flow `systemLogger`, `securityLogger`, `sqlLogger` and `i18nLogger` via the virtual objects `Neos.Flow:SystemLogger`, `Neos.Flow:SecurityLogger`, `Neos.Flow:SqlLogger` and `Neos.Flow:I18nLogger` respectively.

```php
/**
 * @Flow\Inject(name="Neos.Flow:SystemLogger")
 * @var LoggerInterface
 */
protected $systemLogger;
```

**Note:** This also removes the deprecated `PsrSecurityLoggerInterface` and `PsrSystemLoggerInterface`, which should be replaced by injections like above.

Resolves #2125


sorenmalling (approved):
Approved by reading - format is as we discussed in original PR

albe:
Added some docs for the logging part of Flow that was still missing. It's very basic still, but any help is highly welcome!

albe:
Q: Is a code migration possible that replaces
```
@var .*PsrSecurityLoggerInterface
@Flow\Inject
```
and
```
@Flow\Inject
@var .*PsrSecurityLoggerInterface
```
with the above? i.e. multi-line replacements?

albe:
TODO: Maybe add a code-example for logging an exception via `ThrowableStorage`?

sorenmalling:
> TODO: Maybe add a code-example for logging an exception via `ThrowableStorage`?

That will be a great addition to the docs 🤩

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
```diff
@@ -0,0 +1,157 @@
...
+	 */
+	protected $systemLogger;
+	
+	/**
+	 * @Flow\Inject(name="Neos.Flow:SecurityLogger")
+	 * @var LoggerInterface
+	 */
+	protected $securityLogger;
+
+	/**
```
albe, line 40: Should we make the sqlLogger and i18nLogger available like this too? Would somehow make sense for consistency, but then again we only had b/c interfaces for `systemLogger` and `securityLogger`.
@daniellienert @bwaidelich ?

albe:
cc @lorenzulrich - maybe you have something to add/improve? I'm really bad at (technical) writing in a way that it is user-friendly/understandable

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
```diff
@@ -0,0 +1,157 @@
...
+=======
+Logging
+=======
+
+.. sectionauthor:: Alexander Berl <alexander@neos.io>
+
+In Flow logging is implemented according to the `PSR-3 Standard`_. This means you can use any logging facility that implements this interface.
+The concrete implementation can easily be configured with the :doc:`Object Management` of Flow. By default, Flow comes with an PSR-3 implementation
+in the ``Neos.Flow.Log`` package that can be configured with one or more storage backends and supports log rotation. The supported backends currently
+consist of a FileBackend, (Ansi)ConsoleBackend, JsonFileBackend and a NullBackend.
```
bwaidelich, line 12:
```suggestion
consist of a ``FileBackend``, ``(Ansi)ConsoleBackend``, ``JsonFileBackend`` and a ``NullBackend``.
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
```diff
@@ -0,0 +1,157 @@
...
+
+Default loggers
+===============
+
+By default Flow comes with four loggers, the so called "system logger", "security logger", "sql logger" and "i18n logger".
+As the names imply, the first is responsible for logging general system level messages, the second for
```
bwaidelich, line 18:
```suggestion
As the names imply, the first is responsible for logging general system level messages and is used by default if an instance of the ``LoggerInterface`` is injected. The second for
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
```diff
@@ -0,0 +1,157 @@
...
+logging security related information. The SQL logger needs to be enabled first via ``Neos.Flow.persistence.doctrine.sqlLogger`` setting and will
+create a log of all database queries, so this can become a big performance penalty and should only be used for debugging purposes.
+Last but not least is the i18n logger which will log away all messages related to the translation framework, for example when the XLIFF translation
+sources are badly formatted.
+Under the hood they all use a ``FileBackend`` for storing the messages, but that can be configured differently via the settings keys ``Neos.Flow.log.psr3.Neos\Flow\Log\PsrLoggerFactory.*``.
+
+You can use these loggers for example if you extend the system via signals or AOP, or you are debugging
```
bwaidelich, line 25: I don't understand that phrase. Why is this related to signal/slot and AOP?

bwaidelich (commented):
Nice one, thanks for pushing this topic!

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
albe: Well, it's not related strictly speaking, but I tried to come up with use-cases where you wouldn't just inject your own application level logger (which we maybe should also mention in a section in this doc).

Neos.Flow/Classes/Security/Aspect/LoggingAspect.php
```diff
@@ -29,8 +28,8 @@
 class LoggingAspect
 {
     /**
-     * @var PsrSecurityLoggerInterface
-     * @Flow\Inject
+     * @var \Psr\Log\LoggerInterface
```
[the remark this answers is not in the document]
albe: Nope :)

albe:
Also replaced nearly all of the `Objects.yaml` logger injections with an `@Flow\Inject(name="Neos.Flow:*Logger")` annotation. `CacheManager` still needs the setter injection because it's instanciated before the injection is available and two other places still use other class injections via `Objects.yaml`, so not sure it really helps getting rid of the logger part only there.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
```diff
@@ -0,0 +1,157 @@
...
+your code. For that you inject a ``Psr\Log\LoggerInterface`` wherever you need it like this::
+
+	/**
+	 * @Flow\Inject(name="Neos.Flow:SystemLogger")
```
daniellienert, line 29: The line above state, that injecting `LoggerInterface` by default injects the SystemLogger. Why is the name necessary here?

Neos.Flow/Classes/Session/Aspect/LoggingAspect.php
```diff
@@ -25,21 +25,11 @@
 class LoggingAspect
 {
     /**
+     * @Flow\Inject(name="Neos.Flow:SecurityLogger")
```
daniellienert, line 28: This changes the Logger from System to Security, on purpose? Thats pretty breaking.

daniellienert (changes requested):
Looks good, apart from that logger change.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
albe: Technically you can leave it for the `SystemLogger` but I think it's a bad idea having a default for this interface (debatable), it's more explicit to state that you expect the system logger at this point, but more importantly, I do think this better helps transport the concept of injecting via the virtual objects configuration, than having two examples for the default and non-default case.

Neos.Flow/Classes/Session/Aspect/LoggingAspect.php
albe: 👀 very good catch! That's why reviews are important!
```suggestion
     * @Flow\Inject(name="Neos.Flow:SystemLogger")
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
daniellienert: I see the point, that it is much clearer, what to expect here. But I use the system logger in 99,99 percent of the time (maybe because I don't like the concept of having different logfiles ... but you rarely see usages of other loggers, especially in 3rd party code). 

With these virtual objects, injects of this "standard" object is much more bulky now and lacks code completion. So it makes things more complicated in the majority of the cases.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
daniellienert: Have you ever had the use case to inject a sql-logger? Isn't this logger just there for doctrine to automatically log the issued queries?

daniellienert (commented):
Personally I don't see an improvement for the case of the system logger, despite of having "configuration over convention" (🤔). But as long as the LoggerInterface still defaults SystemLogger implementation, I don't want to block this.

albe:
> Personally I don't see an improvement for the case of the system logger

Yeah, it's not an "improvement" for that case. But this is more about making the injection of any logger straight forward through a single documented method. More about consistency than anything else.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
albe: > lacks code completion

That's a valid point, but the only solution I see for this is if we reintroduce specific interfaces for the different framework loggers (`SystemLoggerInterface`, `SecurityLoggerInterface`, ...) that you can inject. Totally plausible and an alternative to the virtual objects. Maybe we should discuss this in the team first, if we want to go virtual objects or interfaces for loggers, as changing it again later would be bad?

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
```diff
@@ -0,0 +1,157 @@
...
+By default Flow comes with four loggers, the so called "system logger", "security logger", "sql logger" and "i18n logger".
+As the names imply, the first is responsible for logging general system level messages and is used by default if an instance of the ``LoggerInterface`` is injected. The second for
+logging security related information. The SQL logger needs to be enabled first via ``Neos.Flow.persistence.doctrine.sqlLogger`` setting and will
+create a log of all database queries, so this can become a big performance penalty and should only be used for debugging purposes.
+Last but not least is the i18n logger which will log away all messages related to the translation framework, for example when the XLIFF translation
+sources are badly formatted.
+Under the hood they all use a ``FileBackend`` for storing the messages, but that can be configured differently via the settings keys ``Neos.Flow.log.psr3.Neos\Flow\Log\PsrLoggerFactory.*``.
+
+You can use these loggers for example if you extend the system via signals or AOP, or you are debugging
+your code. For that you inject a ``Psr\Log\LoggerInterface`` wherever you need it like this::
```
bwaidelich, line 26: ^ I see. I still find the phrase a little bit confusing tbh because signal/slot and AOP is so specific. What about something like
```suggestion
Loggers can be used to record events that happen in an application, for example a failed login attempt or a caught exception.
To make use of a logger, the ``Psr\Log\LoggerInterface`` can be injected, that refers to the ``SystemLogger`` which – by default – persists any log message in the file system::
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
bwaidelich: I would suggest that we use the `Psr\Log\LoggerInterface` as synonym for the System Logger by default (i.e. remove `name="Neos.Flow:SystemLogger"` from the examples).
IMO that doesn't really make it less explicit really. It just means that I want to use some default logger implementation and don't really care which one it is _at this point_.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
bwaidelich: good point, Daniel.. :)

albe: Yes, it is. You could still inject it somewhere to give more details to the following logged SQL expressions for example.
I mean it's not straightforward to use it (or the i18n logger), but not unthinkable and hence why should we exempt those?

I see it as "these are the loggers the framework uses, if you really want, you can also use them, but most likely you should use your own".

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
albe: > that doesn't really make it less explicit really.

Guess that depends on the POV - and I can totally see your point now "don't really care which one".
We should document that clearly with examples for both - both is possible, so it's a decision (or non-decision) to make by the developer.

albe: Does https://github.com/neos/flow-development-collection/pull/2134/files#diff-a206dea4eb5ffcc6b7dc0386821eca9eR58-R65 suffice for you two?
I would leave the "name" attribute in this specific place, since it is meant to list the different (explicit) options, I removed it from the other example for ThrowableStorage below.

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -80,17 +81,6 @@ class Router implements RouterInterface
      */
     protected $lastResolvedRoute;
 
-    /**
-     * Injects the (system) logger based on PSR-3.
-     *
-     * @param LoggerInterface $logger
-     * @return void
-     */
-    public function injectLogger(LoggerInterface $logger)
```
albe, line 89: Note: We have way too many tests depend on this method (and the other `injectLogger`). Would keeping the method interfere with the `@Flow\Inject` annotation?

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
albe: So, according to the poll, we'll stick with virtual objects, even though they won't have autocomplete for the name. Regarding the usage of virtual object system logger injection in core, see my argument above and in the discuss thread. I'm open for counter-arguments though.

bwaidelich (approved):
Looks good (by reading)
Thanks a lot for taking care!

Neos.Flow/Classes/Mvc/Routing/Router.php
albe: I was more questioning if I should maybe undo the removal (to help the tests). Otherwise need to replace all the `injectLogger(...)` calls there with `_set('logger', ...)` "magic"

albe: any thoughts on this @bwaidelich ?

albe: Okay, so I reverted this for now, then replaced the needless cases of `injectLogger()` in tests with `$this->inject()`. So this change is a little bit less breaking (if 3rd party depends on the `injectLogger()` method), yet we have the usage in core streamlined to where it's necessary.

albe:
I reverted the removal of the `injectLogger()` methods for now, then replaced the needless cases in tests with `$this->inject()`.
Would merge this now once tests pass again.

Issue #2133, opened 2020-09-18
Clients may enclose cookie values in double quotes


albe:
According to RFC 6265 https://tools.ietf.org/html/rfc6265#section-4.1.1 a cookie value may be enclosed in double quotes.

Apparently, Android clients for example make use of this:
https://stackoverflow.com/questions/19921514/android-basicclientcookie-cookiestore-adds-quotes-around-cookie-value

Currently, our `Cookie` class implementation does not strip double quotes from cookie values, therefore this may lead to invalid values being received, like this:
```
InvalidArgumentException: ""YsWSZUIQZwH4wO4Ei2BIhJzayT8PxEZl"" is not a valid cache entry identifier.
#11 /application/Packages/Libraries/neos/cache/Classes/Frontend/VariableFrontend.php(87): get
#10 /application/Packages/Framework/Neos.Flow/Classes/ObjectManagement/DependencyInjection/DependencyProxy.php(99): __call
#9 /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Session_SessionManager.php(70): initializeCurrentSessionFromCookie
#8 /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Session_Http_SessionRequestComponent.php(62): handle
#7 /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Http_Component_ComponentChain.php(62): handle
#6 /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Http_Component_ComponentChain.php(62): handle
#5 /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Http_Middleware_ComponentChainMiddleware.php(54): process
#4 /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Http_Middleware_MiddlewaresChain.php(78): process
#3 /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Http_Middleware_MiddlewaresChain.php(92): handle
[...]

PR #2137, opened 2020-09-18
BUGFIX: Take into account that a cookie value may be enclosed in double quotes

DavidSporer:
According to RFC 6265 https://tools.ietf.org/html/rfc6265#section-4.1.1 a cookie value may be enclosed in double quotes.
This change takes this into account by removing the first and last double quote of a value (enclosing double quotes).

Fixes: #2133

**What I did**
Take into account that according to https://tools.ietf.org/html/rfc6265#section-4.1.1 Cookie values may be enclosed in double quotes.

**How I did it**
Stripping enclosing double quotes from the Cookie value.

**How to verify it**
Creating a cookie and adding quotes around it in the dev tools of your browser.
Before the change this leads to an exception, after the change it works as expected.
As an alternative you can use an Android phone that behaves like this but we don't have one in our office.

### DIFF
--- a/Neos.Flow/Classes/Http/Cookie.php
+++ b/Neos.Flow/Classes/Http/Cookie.php
@@ -122,7 +122,7 @@ class Cookie
         }
 
         $this->name = $name;
-        $this->value = $value;
+        $this->value = $value === null ? null : preg_replace('/^"(.*)"$/', '$1', $value);
         $this->expiresTimestamp = $expires;
         $this->maximumAge = $maximumAge;
         $this->domain = $domain;


Neos.Flow/Classes/Http/Cookie.php
```diff
@@ -122,7 +122,7 @@ public function __construct($name, $value = null, $expires = 0, $maximumAge = nu
         }
 
         $this->name = $name;
-        $this->value = $value;
+        $this->value = $value === null ? null : preg_replace('/^"(.*)"$/', '$1', $value);
```
bwaidelich, line 125: I would prefer another constraint check like the other ones above that throws an `\InvalidArguments` exception if the value does not adhere to the rules and then fix the code that _creates_ the instance.
That way the risk of unplanned side effects is limited.

Also, this should be covered with a test :)

bwaidelich (changes requested):

Neos.Flow/Classes/Http/Cookie.php
DavidSporer: Ok but the issue is that the quotes aren't added by Flow but by the Client.
So basically Flow is setting the value without quotes and on the next requests, the client adds quotes around it.
Yep, tests should be there - is there a documentation on how to setup and run the tests for Flow? I never did this. We're using Codeception in our stuff.

bwaidelich: > Ok but the issue is that the quotes aren't added by Flow but by the Client.

But Flow calls the constructor, doesn't it?
I think that part should be using [Cookie::createFromRawSetCookieHeader()](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Http/Cookie.php#L172) (or a new static constructor if that doesn't fit) which should take care of the trimming.

> is there a documentation on how to setup and run the tests for Flow?

Not really..
You can have a look at the existing tests to get an idea.
Executing unit tests is as simple as running:
```
./bin/phpunit --colors -c Build/BuildEssentials/PhpUnit/UnitTests.xml
```

But don't worry, I can help out with the tests!

bwaidelich:
As you can see from the broken CI: Maybe we don't even need to write new tests but only extend the existing ones.
Also as a bugfix this should target the lowest supported branch (`5.3`). I'll give it a shot

bwaidelich:
I didn't want to mess with your PR and created a "minimal invasive" version for `5.3`: #2138

DavidSporer:
Thanks! I guess I can close this one then.

Neos.Flow/Classes/Http/Cookie.php
albe: Note: As of 6.3 **development-distribution** you can also just run `./composer test-unit`, `./composer test-functional` and/or `./composer test-static` inside the flow root directory
See https://github.com/neos/flow-development-distribution/blob/master/composer.json#L40-L42

PR #2138, opened 2020-09-18
BUGFIX: Trim enclosing double quotes from session cookie

bwaidelich:
According to RFC 6265 https://tools.ietf.org/html/rfc6265#section-4.1.1 a cookie
value may be enclosed in double quotes.
This change takes this into account by removing the first and last double quote of a
value (enclosing double quotes) when starting/resuming a session.

Fixes: #2133

### DIFF
--- a/Neos.Flow/Classes/Session/Http/SessionRequestComponent.php
+++ b/Neos.Flow/Classes/Session/Http/SessionRequestComponent.php
@@ -74,7 +74,8 @@ class SessionRequestComponent implements ComponentInterface
     {
         return new Cookie(
             $name,
-            $value,
+            // @see https://github.com/neos/flow-development-collection/issues/2133
+            trim(urldecode($value), '"'),
             0,
             $this->sessionSettings['cookie']['lifetime'],
             $this->sessionSettings['cookie']['domain'],
--- a/Neos.Flow/Tests/Unit/Http/Component/SessionRequestComponentTest.php
+++ b/Neos.Flow/Tests/Unit/Http/Component/SessionRequestComponentTest.php
@@ -1,0 +1,210 @@
+<?php
+namespace Neos\Flow\Tests\Unit\Http\Component;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Http\Component\ComponentContext;
+use Neos\Flow\Http\Cookie;
+use Neos\Flow\Session\Http\SessionRequestComponent;
+use Neos\Flow\Session\SessionInterface;
+use Neos\Flow\Session\SessionManager;
+use Neos\Flow\Tests\UnitTestCase;
+use PHPUnit\Framework\MockObject\MockObject;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Test case for the SessionRequestComponent
+ */
+class SessionRequestComponentTest extends UnitTestCase
+{
+
+    /**
+     * @var SessionRequestComponent
+     */
+    private $sessionRequestComponent;
+
+    /**
+     * @var SessionManager|MockObject
+     */
+    private $mockSessionManager;
+
+    /**
+     * @var ServerRequestInterface|MockObject
+     */
+    private $mockHttpRequest;
+
+    /**
+     * @var ComponentContext|MockObject
+     */
+    private $mockComponentContext;
+
+    /**
+     * @var array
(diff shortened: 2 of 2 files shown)


bwaidelich:
Thanks to @davidspiola & @albe for doing the _hard_ work on this one :)

Neos.Flow/Tests/Unit/Http/Component/SessionRequestComponentTest.php
```diff
@@ -0,0 +1,216 @@
...
+        $this->mockSessionManager->expects($this->once())->method('initializeCurrentSessionFromCookie')->willReturnCallback(static function (Cookie $cookie) use ($expectedCookie) {
+            self::assertSame($expectedCookie, (string)$cookie);
+        });
+
+        $this->sessionRequestComponent->handle($this->mockComponentContext);
+    }
+
+    public function cookieValueDataProvider(): array
+    {
+        return [
```
bwaidelich, line 181: Those are the actual test cases that test the bugfix. The other test cases were added boy-scout style :)

Neos.Flow/Classes/Session/Http/SessionRequestComponent.php
```diff
@@ -74,7 +74,8 @@ protected function prepareCookie(string $name, string $value)
     {
         return new Cookie(
             $name,
-            $value,
+            // @see https://github.com/neos/flow-development-collection/issues/2133
+            trim(urldecode($value), '"'),
```
albe, line 78: So we're okay with potentially removing too many double quotes? (see https://github.com/neos/flow-development-collection/issues/2133#issuecomment-694728094) Just trying to get your opinion on this. I know we already had it that way in `Headers`, but we could as well fix it there and here with this, no?

Neos.Flow/Tests/Unit/Http/Component/SessionRequestComponentTest.php
```diff
@@ -0,0 +1,216 @@
...
+            ['sessionCookieValue' => 123, 'expectedNewCookieValue' => '123'],
+            ['sessionCookieValue' => '', 'expectedNewCookieValue' => ''],
+            ['sessionCookieValue' => 'some String', 'expectedNewCookieValue' => 'some String'],
+            ['sessionCookieValue' => '"leading quote', 'expectedNewCookieValue' => 'leading quote'],
+            ['sessionCookieValue' => 'trailing quote"', 'expectedNewCookieValue' => 'trailing quote'],
+            ['sessionCookieValue' => '"quotes"', 'expectedNewCookieValue' => 'quotes'],
+            ['sessionCookieValue' => '""double quotes"', 'expectedNewCookieValue' => 'double quotes'],
```
albe, line 188: Tests ❤️

See comment above, but I guess this test states that you think we should just do it like this?

albe (approved):
Generally approve apart from the question about whether to fix the `trim()` from `Header` instead of copying it.

And another topic: What about the `Cookie::createFromRawSetCookieHeader(...)` method? I feel like that should be used at some place (probably `Header` and `Session`), but it isn't? Should we deprecate and remove it or try to make actual use of it (could be in a follow-up)?

Neos.Flow/Classes/Session/Http/SessionRequestComponent.php
bwaidelich: In that part of the code `$value` contains the Flow session ID which consists of 32 basic characters (matching regex `/^[a-zA-Z0-9]{32}$/`). Plus, the enclosing quotes in some cases.
I think it's fine to fix just that case for now. If it turns out that other clients do other weird stuff we might have to decide differently

Neos.Flow/Tests/Unit/Http/Component/SessionRequestComponentTest.php
bwaidelich: Yes, for now at least

bwaidelich:
> What about the Cookie::createFromRawSetCookieHeader(...) method?
>[...] Should we deprecate and remove it or try to make actual use of it (could be in a follow-up)?

This should be a minimal invasive change in order to be applied to all supported branches without too much risk.
If `Cookie::createFromRawSetCookieHeader()` is not used I'd be happy about a follow up that deprecates/removes it for `master`

Neos.Flow/Classes/Session/Http/SessionRequestComponent.php
albe: Good point, didn't think about the specific use-case we have in the session component.

DavidSporer: That's actually also the case that threw the exception in our case since double quotes aren't a valid cache entry identifier so removing the double quotes entirely sounds ok to me.

PR #2139, opened 2020-09-19
BUGFIX: Throw helpful exception when collection property has missing element type

### LINKED ISSUE #978: Improve developer experience for property mapping errors
**Use case:**

A DTO model is supposed to be property mapped, containing a collection type property.
The property is correctly annotated `Collection<SubEntity>`, while accidentially the setter (or constructor) is annotated with just `Collection`.

**Expected:**

A useful error message is shown that hints at the missing element type annotation in the setter.

**Actual:**

`Exception while property mapping for target type "Acme\Domain\Model\SomeDto", at property path "collection.0": The target type was no string, but of type "NULL"`

albe:
Fixes #978

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/CollectionConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/CollectionConverter.php
@@ -13,6 +13,7 @@ namespace Neos\Flow\Property\TypeConverter;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Property\Exception\InvalidDataTypeException;
 use Neos\Flow\Property\PropertyMappingConfigurationInterface;
 use Neos\Utility\TypeHandling;
 
@@ -85,6 +86,9 @@ class CollectionConverter extends AbstractTypeConverter
     public function getTypeOfChildProperty($targetType, $propertyName, PropertyMappingConfigurationInterface $configuration)
     {
         $parsedTargetType = TypeHandling::parseType($targetType);
+        if ($parsedTargetType['elementType'] === null) {
+            throw new InvalidDataTypeException(sprintf('The collection type property "%s" is missing an element type (got "%s")', $propertyName, $targetType), 1600530641);
+        }
         return $parsedTargetType['elementType'];
     }
 }
--- a/Neos.Flow/Classes/Property/TypeConverter/CollectionConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/CollectionConverter.php
@@ -87,7 +87,7 @@ class CollectionConverter extends AbstractTypeConverter
     {
         $parsedTargetType = TypeHandling::parseType($targetType);
         if ($parsedTargetType['elementType'] === null) {
-            throw new InvalidDataTypeException(sprintf('The collection type property "%s" is missing an element type (got "%s")', $propertyName, $targetType), 1600530641);
+            throw new InvalidDataTypeException(sprintf('The annotated collection property "%s" is missing an element type. Should be given in the format "Collection<$elementType>" and "$elementType" is the class name of the contained objects.', $propertyName), 1600530641);
         }
         return $parsedTargetType['elementType'];
     }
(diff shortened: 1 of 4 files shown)


Neos.Flow/Classes/Property/TypeConverter/CollectionConverter.php
```diff
@@ -85,6 +86,9 @@ public function getSourceChildPropertiesToBeConverted($source)
     public function getTypeOfChildProperty($targetType, $propertyName, PropertyMappingConfigurationInterface $configuration)
     {
         $parsedTargetType = TypeHandling::parseType($targetType);
+        if ($parsedTargetType['elementType'] === null) {
+            throw new InvalidDataTypeException(sprintf('The collection type property "%s" is missing an element type (got "%s")', $propertyName, $targetType), 1600530641);
```
sorenmalling, line 90: I'm not entirely sure what this "collecction type property" exception tells me. What is the "collection type property" in this example? Is it the inside of the Collection annotation (`Collection<$elementType>`)?

Can we rephrase to 

"The annotated Collection property "%s" was not given a element type. Should be given in the format Collection<$elementType> and it ususally a class name of a different object"

albe:
```suggestion
            throw new InvalidDataTypeException(sprintf('The annotated collection property "%s" was not given an element type. Should be given in the format Collection<$elementType> and is ususally a class name of a different object.', $propertyName, $targetType), 1600530641);
```

Neos.Flow/Tests/Unit/Property/TypeConverter/CollectionConverterTest.php
```diff
@@ -50,9 +50,10 @@ public function getTypeOfChildPropertyReturnsElementTypeFromTargetTypeIfGiven()
 
     /**
      * @test
+     * @expectedException \Neos\Flow\Property\Exception\InvalidDataTypeException
      */
     public function getTypeOfChildPropertyReturnsEmptyStringForElementTypeIfNotGivenInTargetType()
     {
-        $this->assertEquals('', $this->converter->getTypeOfChildProperty('array', '', $this->createMock(PropertyMappingConfigurationInterface::class)));
+        $this->converter->getTypeOfChildProperty('array', 'collection', $this->createMock(PropertyMappingConfigurationInterface::class));
```
bwaidelich, line 57: The test method name is not correct any more, and this changed expectation makes me wonder: Is just `Collection` not a valid annotation any longer?

bwaidelich (changes requested):

Neos.Flow/Tests/Unit/Property/TypeConverter/CollectionConverterTest.php
albe: It is generally a valid annotation, but it's invalid in terms of property mapping, because without the element type the mapping can't continue into the elements (and will throw with an unspecific exception then). TBH I'm not sure why there was a test for this to return an empty string as the element type, other than assuming the test was just fixating the straightforward implementation.

Neos.Flow/Tests/Unit/Property/TypeConverter/CollectionConverterTest.php
```diff
@@ -50,9 +50,10 @@ public function getTypeOfChildPropertyReturnsElementTypeFromTargetTypeIfGiven()
 
     /**
      * @test
+     * @expectedException \Neos\Flow\Property\Exception\InvalidDataTypeException
      */
     public function getTypeOfChildPropertyReturnsEmptyStringForElementTypeIfNotGivenInTargetType()
```
albe, line 55:
```suggestion
    public function getTypeOfChildPropertyThrowsExceptionForMissingElementType()
```

Neos.Flow/Tests/Unit/Property/TypeConverter/CollectionConverterTest.php
bwaidelich: Of course, that makes sense!

bwaidelich (approved):
Thanks for clarifying

+1 by reading

albe:
Apparently, the exception will not be thrown when an empty collection is submitted. Which is a bit... shaggy.
See https://github.com/neos/flow-development-collection/pull/2218

Adding a check to the ObjectConverter is probably more sane, but also not as easy (we have three versions of where to check the type - constructor arguments, setter argument, property)

PR #2144, opened 2020-09-22
FEATURE: Easy injection of the active ServerRequest

bwaidelich:
This introduces a new singleton `ActiveServerRequestProvider` that
makes it really easy to get hold of the currently active HTTP request:

```
public function __construct(\Psr\Http\Message\ServerRequestInterface $request) {
    // $request will be the active ServerRequest in HTTP context, otherwise throws a \Neos\Flow\Http\Exception
}
```

Related: #658


sorenmalling:
Brilliant idea - will review tomorrow 👍

Neos.Flow/Classes/Http/ActiveServerRequestProvider.php
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @return ServerRequestInterface
+     */
+    public function getActiveServerRequest(): ServerRequestInterface
+    {
+        $requestHandler = $this->bootstrap->getActiveRequestHandler();
+        if ($requestHandler instanceof HttpRequestHandlerInterface) {
+            return $requestHandler->getHttpRequest();
+        }
+        return $this->serverRequestFactory->createServerRequest('GET', 'http://localhost');
```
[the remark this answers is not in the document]
bwaidelich: Good point, I'll give it a try!

Neos.Flow/Classes/Http/ActiveServerRequestProvider.php
```diff
@@ -0,0 +1,44 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Central authority to get hold of the active HTTP request.
+ *
+ * When no active HTTP request can be determined (for example in CLI context) a new instance is built using a ServerRequestFactoryInterface implementation
+ *
+ * @Flow\Scope("singleton")
+ */
+final class ActiveServerRequestProvider
```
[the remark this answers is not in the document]
bwaidelich: > Not fully sure about the "active"

The word came from `Bootstrap::getActiveRequestHandler()` and I like it slightly more than "current".

> Also unsure about ServerRequest

OK, what about `getActiveHttpRequest()` instead?

Neos.Flow/Classes/Http/ActiveServerRequestProvider.php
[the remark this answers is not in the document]
bwaidelich: Done with bddc83d18891798504c8dd93e18e00ec3208e89e

Neos.Flow/Classes/Http/ActiveServerRequestProvider.php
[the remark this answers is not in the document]
bwaidelich: I changed it to "active HTTP request" now with bddc83d18891798504c8dd93e18e00ec3208e89e and added a comment that the provider is usually not being used in userland code (this the naming is maybe not _thaat_ important in this case)

Neos.Flow/Classes/Http/ActiveHttpRequestProvider.php
```diff
@@ -33,12 +38,17 @@ final class ActiveServerRequestProvider
...
-    public function getActiveServerRequest(): ServerRequestInterface
+    public function getActiveHttpRequest(): ServerRequestInterface
     {
         $requestHandler = $this->bootstrap->getActiveRequestHandler();
         if ($requestHandler instanceof HttpRequestHandlerInterface) {
             return $requestHandler->getHttpRequest();
         }
-        return $this->serverRequestFactory->createServerRequest('GET', 'http://localhost');
+        try {
+            $baseUri = $this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest();
```
albe, line 50: The `OrFallbackToCurrentRequest` is probably not needed, because we just checked that the active request handler is not a `HttpRequestHandler` and hence no `HttpRequest` is "current" (also notice we called it "current request" there).

bwaidelich: It's the only public method in the [BaseUriProvider](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Http/BaseUriProvider.php) unfortunately..
We could just inject the `baseUri` setting instead and fallback to that.. And ultimately deprecate the `BaseUriProvider` because you could then just do a `RequestInformationHelper::generateBaseUri($injectedHttpRequest)` instead. WDYT?

albe: The `getConfiguredBaseUri()` method isn't public? Hmm. I would have expected that. TBH I wouldn't deprecate it, having a designated way to get the "current baseUri" is something that's surely needed in userland. But yes, let's just inject the setting here then.

kitsunet: I don't undesrtand why we would fallback anyways? If there is no ActiveHttpRequest then there is none and we shouldn't provide one?

bwaidelich: > having a designated way to get the "current baseUri" is something that's surely needed in userland

I always struggled a bit with the notion of a "base URI" and IMO it led to quite some confusion. But, granted, it is needed. I just wonder if we need a dedicated service for it since instead of

```php
/**
  * @Flow\Inject
  * @var BaseUriProvider
  */
protected $baseUriProvider;

private function someMethod() {
  $baseUri = $this->baseUriProvider-> getConfiguredBaseUriOrFallbackToCurrentRequest();
}
```

now you could just write:

```php
/**
  * @Flow\Inject
  * @var ServerRequest
  */
protected $activeHttpRequest;

private function someMethod() {
  $baseUri = RequestInformationHelper::generateBaseUri($activeHttpRequest);
}
```
Anyways,  I'm absolutely fine with not deprecating it.

> I don't undesrtand why we would fallback anyways?

@kitsunet Did you read the discussion from #658?

We are looking for a way to get hold of the active HTTP request because sometimes you don't get away with a custom HTTP component/middleware.
The current way of injecting the bootstrap, doing the `instanceof` check and referring to the `ComponentContext` is suboptimal since
* It's hard to test
* The `ComponentContext` will be removed
* It's very imperative (i.e. hard for us to tweak without breaking changes)

Also, you'll have to implement fallbacks whenever you need an instance for CLI context (for example when using the Fluid `StandaloneView`)

kitsunet: I have nothing against this provider in particular, in fact I welcome it, but I wouldn't fallback to some made up request because that most certainly is not the current or active request. IMHO that should be left to the userland code to decide.

bwaidelich: > I wouldn't fallback to some made up request

Mh, so you would suggest to return `null` if the active request can't be determined?

With that you would have to change sth like
```php
final class EmailService {
  public function __construct(ServerRequest $httpRequest) {
    //...
  }
}
```
to something like
```php
final class EmailService {
  public function __construct(ServerRequestFactoryInterface $serverRequestFactory, string $configuredBaseUri, ?ServerRequest $httpRequest) {
    if ($httpRequest === null) {
      $httpRequest = $serverRequestFactory->createServerRequest('GET', $configuredBaseUri);
    }
  }
}
```
which is quite cumbersome..

Granted, if you want to make the `EmailService` more explicit, you'd create a dedicated factory (or some `Objects.yaml` config) in order to have more control over the injected request.. But you can do this anyways.

Neos.Flow/Classes/Http/ActiveServerRequestProvider.php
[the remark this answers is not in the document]
bwaidelich: I simplified the baseUri fallback with 0abcac2d90c658d4bdd978846a66f05f4716c50a

Neos.Flow/Classes/Http/ActiveHttpRequestProvider.php
bwaidelich: To follow up on this: I agree that it is a little bit "magic" to just refer to a HTTP Request and that defaults to the currently active request or a new instance with the configured base URI.

But in this very case I think it's a pragmatic solution since it is probably the right behavior for a great majority of cases – and for the other ones nothing really changed since you can adjust the behavior for every injection

albe: > To follow up on this: I agree that it is a little bit "magic" to just refer to a HTTP Request and that defaults to the currently active request or a new instance with the configured base URI.
> 
> But in this very case I think it's a pragmatic solution since it is probably the right behavior for a great majority of cases – and for the other ones nothing really changed since you can adjust the behavior for every injection

💯 Agree with the above

> now you could just write:
>
> ```php
>  /**
>   * @Flow\Inject
>   * @var ServerRequest
>   */
> protected $activeHttpRequest;
> 
> private function someMethod() {
>   $baseUri = RequestInformationHelper::generateBaseUri($activeHttpRequest);
> }
> ```

Yeah, I guess that's the alternative (at least with this change). So rather a question of documenting this (note the BaseUriProvider isn't really documented either yet).

albe: PS: Coming back around to the  "`StandaloneView` in CLI" use case: Being able to inject a ServerRequest is nice, but only 50% of what is needed (and I'd probably even like to hide the need to have a HTTP request from the developer in this case - he only wants to render a template). So I'd follow this up with an improvement that makes a `StandaloneView` injectable or instanciable without manually creating an `ActionRequest` that needs a HTTP request below. Just mentioning here, so this can be considered.

kitsunet: Sure if you are both convinced that's the best way. I became vary of names that promise one thing and deliver (sometimes a different thing), in this case an active http request that is not active at all.

I find the Objects.yaml entry for this really to magic though. If you want it injected at least be conscious of what you do and do a factory config. I can totally see otherwise that this is used to "create" requests by injection.

albe: > promise one thing and deliver (sometimes) a different thing

Yeah. A more honest name would probably be "ActiveOrPseudoRequestProvider". But then again, in the cases where the `Pseudo` is returned, the developer likely doesn't really care about that - he just wants a Request for some other reason (e.g. `StandaloneView` see above).
Now the question is: Are there really any cases where he actually wants to get the "active HTTP request", but can't/won't and instead receives a dummy request that has no real meaning (apart from wrapping the configured baseUri).

bwaidelich: I think we're losing track a little (partly because of me introducing topics into the discussion).

So lets get back to the point:
We want to

bwaidelich: Personally I think it would be a good default behavior for the common case (HTTP context) since in PHP everything is centered around a single incoming request anyways.

As for the CLI, I share Christians concerns that it might be too magic but I tend to think that it won't be a problem in _practice_ and that it makes things easier.

I would vote against making the return type of the provider nullable since that doesn't make it any more explicit. I also wouldn't want to get rid of the default behavior (e.g. `Objects.yaml`) because that would kind of defeat the main purpose of this undertaking.

What I could imagine is something like this:

```php
final class ActiveHttpRequestProvider
{

    public function getActiveHttpRequest(): ServerRequestInterface
    {
        $requestHandler = $this->bootstrap->getActiveRequestHandler();
        if (!$requestHandler instanceof HttpRequestHandlerInterface) {
            throw new \RuntimeException('No http request');
        }
        return $requestHandler->getHttpRequest();
    }
}
```

and some proper documentation on how to create a new instance for the CLI.

albe: > get hold of the current HTTP request

Now **you** called it "current" ;) j/k - yes, that's the real topic here, and my question above is probably the "cornerstone" of dispute. And to partially answer it: It will only **not work** in CLI context (or in a class that is instanciated before `RequestHandler::boot()` is called - which I don't see reasonably). So would anyone expect to receive an actual "HTTP request" inside a class used from CLI (maybe accidentially)?

kitsunet: I find the above idea much better just because due to ease of use I can see projects were you get transient injections (eg you inject a service, that then injects the active http request) or just don't think about the implications and then the service becomes unreliable on CLI and it's not immediately clear why. At least this way it would throw an exception and you can follow that up and hopefully understand what is going on. Sorry for throwing a wrench in the almost done PR. Again I like making this easier for userland. :)

bwaidelich: > Now you called it "current"

*dang* :)

> Now you called it "current"

No, don't be sorry. That's exactly why I _created_ the PR and I agree that the exception solution is probably the most explicit one..
I'll push a follow-up that removes the fallback logic

bwaidelich: OK, what do you think of the current implementation?
I'm happy with it even though (or maybe even more because) it's just three lines left in the provider :)

Happy we had that discussion, thanks boyz!

kitsunet (approved):

bwaidelich:
mh, anybody knows why the CI build stopped?

bwaidelich:
> mh, anybody knows why the CI build stopped?

probably just a hickup. Now everything seems fine and I would merge this one within the next 24h if nobody rejects

albe (approved):

bwaidelich:
After talking to @albe again I'm thinking about reverting this change before it lands in a release.
Background: The injected request won't contain the most current request, and once injected it can never be replaced..

Issue #2142, opened 2020-09-21
Replace routing variables in RoutePart handler options


bwaidelich:
### Description

variables in `Routes.yaml` allow custom placeholders to be defined in a central place.
For example a `Routes.yaml`

```yaml
-
  name: 'Some Route with suffix <suffix>'
  uriPattern: 'foo/bar<suffix>'
  defaults:
    'suffix': '<suffix>'
    # ...
```
and a corresponding `Settings.yaml` of
```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Some.Package':
          suffix: '.html'
```

will be merged to an effective route configuration of
```yaml
-
  name: 'Some Route with suffix .html'
  uriPattern: 'foo/bar.html'
  defaults:
    'suffix': '.html'
    # ...
```
(see https://flowframework.readthedocs.io/en/6.3/TheDefinitiveGuide/PartIII/Routing.html#nested-subroutes for more examples)

However, those placeholders are currently _not_ replaced in route part handler options. But this would be useful for example in order to solve https://github.com/neos/neos-development-collection/issues/3020

bwaidelich:
**Note:** With #2155 variables will be replaces in _all_ route configuration options so you could in theory do:

```yaml
-
  uriPattern: '{foo}'
  routeParts:
    'foo':
      handler: '<handler>'
```

PR #2145, opened 2020-09-23
FEATURE: Consider route variables in handler options

bwaidelich:
Allows variables to be used in route part handler options like:

```yaml
-
  name: 'Some route'
  uriPattern: 'some/{part}'
  defaults:
    # ...
  routeParts:
    'part':
      handler:   'Some\RoutePart\Handler'
      options:
        someOption: '<someVariable>'
```

Related: https://github.com/neos/neos-development-collection/issues/3020
Resolves: #2142

### DIFF
--- a/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
+++ b/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
@@ -197,10 +197,11 @@ class RouteConfigurationProcessor
                         $mergedSubRouteConfiguration['defaults'][$key] = $this->replacePlaceholders($defaultValue, $variables);
                     }
                 }
-                foreach ($mergedSubRouteConfiguration['routeParts'] ?? [] as $routePartKey => $routePartValue) {
-                    $mergedSubRouteConfiguration['routeParts'][$routePartKey] = $this->replacePlaceholders($routePartValue, $variables);
-                    foreach ($mergedSubRouteConfiguration['routeParts'][$routePartKey]['options'] ?? [] as $optionKey => $optionValue) {
-                        $mergedSubRouteConfiguration['routeParts'][$routePartKey]['options'][$optionKey] = $this->replacePlaceholders($optionValue, $variables);
+                if (isset($mergedSubRouteConfiguration['routeParts'])) {
+                    foreach ($mergedSubRouteConfiguration['routeParts'] as $routePartKey => $routePartValue) {
+                        foreach ($mergedSubRouteConfiguration['routeParts'][$routePartKey]['options'] ?? [] as $optionKey => $optionValue) {
+                            $mergedSubRouteConfiguration['routeParts'][$routePartKey]['options'][$optionKey] = $this->replacePlaceholders($optionValue, $variables);
+                        }
                     }
                 }
                 $mergedSubRouteConfiguration = Arrays::arrayMergeRecursiveOverrule($routeConfiguration, $mergedSubRouteConfiguration);
--- a/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
+++ b/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
@@ -176,12 +176,12 @@ class RouteConfigurationProcessor
      */
     protected function buildSubRouteConfigurations(array $routesConfiguration, array $subRoutesConfiguration, $subRouteKey, array $subRouteOptions)
     {
-        $variables = isset($subRouteOptions['variables']) ? $subRouteOptions['variables'] : [];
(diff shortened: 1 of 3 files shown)


Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
```diff
@@ -197,6 +197,12 @@ protected function buildSubRouteConfigurations(array $routesConfiguration, array
                         $mergedSubRouteConfiguration['defaults'][$key] = $this->replacePlaceholders($defaultValue, $variables);
                     }
                 }
+                foreach ($mergedSubRouteConfiguration['routeParts'] ?? [] as $routePartKey => $routePartValue) {
```
albe, line 200: Not sure I prefer this over a separate `isset()` condition, but it surely is concise. If we opt for this, I'd like to change the other occurences of `isset+foreach` accordingly (not necessarily in this PR)

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
```diff
@@ -197,6 +197,12 @@ protected function buildSubRouteConfigurations(array $routesConfiguration, array
...
+                    $mergedSubRouteConfiguration['routeParts'][$routePartKey] = $this->replacePlaceholders($routePartValue, $variables);
```
albe, line 201: Don't we need to check if `$routePartValue` is actually a string? In a lot (most?) cases it definitely is an array (see next line)

albe (approved):
Generally approve. One code style question and one potential bug

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
bwaidelich: Yeah right, I'll stick to the `isset+foreach` mode for now and push a follow-up

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
bwaidelich: Doh, good point.. I'll fix

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
bwaidelich: Done

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
bwaidelich: I just removed the line, it didn't do anything really.. Good catch!

albe: I thought it was meant for hardcoded route parts... 
```yml
-
  name: 'Some route'
  uriPattern: 'some/{part}'
  routeParts:
    'part': '<fixedPartVariable>'
```
which doesn't make much sense if I think about it again :D

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
```diff
@@ -197,6 +197,13 @@ protected function buildSubRouteConfigurations(array $routesConfiguration, array
                         $mergedSubRouteConfiguration['defaults'][$key] = $this->replacePlaceholders($defaultValue, $variables);
                     }
                 }
+                if (isset($mergedSubRouteConfiguration['routeParts'])) {
+                    foreach ($mergedSubRouteConfiguration['routeParts'] as $routePartKey => $routePartValue) {
+                        foreach ($mergedSubRouteConfiguration['routeParts'][$routePartKey]['options'] ?? [] as $optionKey => $optionValue) {
```
albe, line 202: I think I see now why you used the `?? []` in the first place. You already (subconciously?) were aware this would lead to an ugly "if { foreach { if { foreach {" cascade

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
bwaidelich: No, it was targeted at the `handler` itself:
```yaml
-
  name: 'Some route'
  uriPattern: 'some/{part}'
  routeParts:
    'part':
      handler: '<variable>'
      options: []
```
but was misspelled and that usecase seems too random anyways

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
bwaidelich: very conciously so :)

albe:
Somehow this broke neos builds:
https://travis-ci.org/github/neos/neos-development-collection/jobs/732940206#L685

I don't know why though

PR #1734, opened 2019-08-23
BUGFIX: Don't validate transient properties during persistence

albe:
See https://github.com/neos/flow-development-collection/pull/1538#issuecomment-524485959

Currently, this will also prevent the validators for transient properties from running in 'Default' only validation group.

PR #2148, opened 2020-09-25
BUGFIX: Don't validate transient properties during persistence

albe:
See https://github.com/neos/flow-development-collection/pull/1538#issuecomment-524485959

A better version of #1734 that will really only ignore transient properties during persistence validation and nowhere else.


albe:
Now the question is: is it potentially problematic to skip some validation during a specific validation step only?

- validation would become inconsistent between controller and persistence steps
- the validation of the property *could* potentially check for some generic rule of the entity, rather than just check if the transient property is a string/number etc. (the latter being of no issue)
- the persistence validation step is only a second safety-net, so something generally invalid shouldn't even come that far

Neos.Flow/Classes/Validation/ValidatorResolver.php
```diff
@@ -309,6 +309,9 @@ protected function buildBaseValidatorConjunction($indexKey, $targetClassName, ar
                 if ($this->reflectionService->isPropertyAnnotatedWith($targetClassName, $classPropertyName, Flow\IgnoreValidation::class)) {
                     continue;
                 }
+                if ($classSchema !== null && $classSchema->isPropertyTransient($classPropertyName) && $validationGroups === ['Persistence', 'Default']) {
```
albe, line 312:
```suggestion
                if ($classSchema !== null
                    && $classSchema->hasProperty($classPropertyName) 
                    && $classSchema->isPropertyTransient($classPropertyName)
                    && $validationGroups === ['Persistence', 'Default']) {
```

albe:
Probably the better solution is to add an exception during reflection when a getter method return type does not match the return type of the according property. At least when there is a validator for that property, but maybe also generally (would be breaking ofc).

kdambekalns (approved):
To me that sounds (and looks) reasonable.

Whether a check on the reeturn type of a getter makes (more) sense, I have my doubts. In the past I had a few cases, where even for a persistent property I implemented a getter with a different return type. The internal representation is something different than the public API.

For validation that means one should either rely on the internal types or the API types, never a mix.

albe:
You are probably right, as always ;) Will merge after tests pass again

PR #2150, opened 2020-09-26
BUGFIX: Correctly recognize NULL method arguments in JoinPoint

### LINKED ISSUE #970: AOP: JointPoint->hasMethodArgument does not recognize NULL
### Description

I created a aspect for tracking model changes (executing on every setter).
To assure that argument is passed, I check the jointPoint for the method-argument.
As the model attribute value has been changed from a value (e.g. object) to null, I would like to track that it changed. But: ```$joinPoint->isMethodArgument('propertyName')``` returns ```false```. Although ```getArguments``` returns ```[propertyName => NULL]```

The problem is here:
https://github.com/neos/flow-development-collection/blob/a2d086a2a38c369a455f2e7cb2449f7df4f09cb6/Neos.Flow/Classes/Aop/JoinPoint.php#L176

```isset``` does return false, if value is set to NULL. Therfore, ```array_key_exists``` should be used:

```
    public function isMethodArgument($argumentName)
    {
        return array_key_exists($argumentName, $this->methodArguments);
    }
``` 
quickwin ;)

#### Expected behavior
JointPoint with arguments: ['myAttribute' => NULL]
Method isMethodArgument('myAttribute') should return true

#### Actual behavior
returns false

[What actually happened]

### Affected Versions

Neos: latest (but code exists more than 2 years)
Flow: latest

albe:
Fixes #970

### DIFF
--- a/Neos.Flow/Classes/Aop/JoinPoint.php
+++ b/Neos.Flow/Classes/Aop/JoinPoint.php
@@ -172,7 +172,7 @@ class JoinPoint implements JoinPointInterface
      */
     public function isMethodArgument($argumentName): bool
     {
-        return isset($this->methodArguments[$argumentName]);
+        return array_key_exists($argumentName, $this->methodArguments);
     }
 
     /**


sorenmalling:
Looks good by reading 👍

sorenmalling (approved):
Looks good by reading

PR #2152, opened 2020-10-02
!!! FEATURE: Remove neos/fluid-adaptor as required package

### LINKED ISSUE #2151: Remove neos/fluid-adaptor as hard dependency in neos/flow composer.json
This is a discussion issue about removing fluid-adaptor as hard dependency of neos/flow.

neos/fluid-adaptor is added in the `require`  section of the composer.json in neos/flow

https://packagist.org/packages/neos/flow

image

I suggest that we remove the hard dependency form neos/flow.

From reading the Flow core, we have a SimpleTemplateView that implements our ViewInterface and is already registered as `defaultImplementation` in Settings path `Neos.Flow.mvc.view.defaultImplementation`

The neos/neos package can require it itself, leaving that at no chance of breaking.

**Pros**
 * Lighter Flow installation
 * Less dependency on outside packages

**Cons**
 * You have to update your package to require the fluid-adaptor package (can be done with proper documentation and upgrade notes)

Please chime in on this topic. If nothing technically stops it, what would you say to have this in for 7.0?

sorenmalling:
Removes references to Fluid and the dependency to the `neos/fluid-adaptor` composer package.

This is a breaking change if you relied on the fact the Flow installs all Fluid dependencies.
In that case you'll need to require them explicitly in your distribution:
```
composer require neos/fluid-adaptor
```

Resolves #2151


Neos.Flow/Classes/Mvc/Controller/MvcPropertyMappingConfigurationService.php
```diff
@@ -135,7 +135,7 @@ public function initializePropertyMappingConfigurationFromRequest(ActionRequest
 
     /**
      * Modify the passed $propertyMappingConfiguration according to the $propertyConfiguration which
-     * has been generated by Fluid. In detail, if the $propertyConfiguration contains
+     * has been generated by ex. Fluid. In detail, if the $propertyConfiguration contains
```
albe, line 138:
```suggestion
     * has been generated by e.g. Fluid. In detail, if the $propertyConfiguration contains
```

Neos.Flow/Configuration/Settings.DocTools.yaml
```diff
@@ -20,34 +18,11 @@ Neos:
         packageKeys:
           - 'Neos.Flow'
           - 'Neos.Party'
-          - 'Neos.FluidAdaptor'
           - 'Neos.Kickstart'
           - 'Neos.Welcome'
         savePathAndFilename: '%FLOW_PATH_PACKAGES%Framework/Neos.Flow/Documentation/TheDefinitiveGuide/PartV/CommandReference.rst'
 
     references:
-      'TYPO3Fluid:ViewHelpers':
```
albe, line 29: Not entirely sure about that part. We should maybe keep providing the ViewHelper reference

sorenmalling: Move it to the Settings.yaml of the fluid adaptor package?

sorenmalling: Added it here
https://github.com/neos/flow-development-collection/pull/2152/files#diff-0b2ac3d01faed426f2d878ae4c6eebc7R22

and checked the output of `./flow configuration:show --type Settings --path Neos.DocTools` 👍

kdambekalns:
Question: Would it make sense to move the Fluid Adaptor into it's own repository?

Otherwise we might miss "hidden" dependencies, since it'll always be there when working with the development collection. We've had such issues with the Party package in the past…

kitsunet:
Sounds like a good idea @kdambekalns, we then need to think about the package type though, otherwise it conflicts if you install the dev collection and it tries to install it into Packages/Framework.

sorenmalling:
@kdambekalns @kitsunet That sounds like a good idea - does it require a update to this PR or how is that done?

bwaidelich:
> Would it make sense to move the Fluid Adaptor into it's own repository?

Yes please.. We could move it to the [Neos development collection](https://github.com/neos/neos-development-collection) or turn [neos/fluidadaptor](https://github.com/neos/fluidadaptor) into a "real" repository - is that possible?

sorenmalling:
> > Would it make sense to move the Fluid Adaptor into it's own repository?
> 
> 
> 
> Yes please.. We could move it to the [Neos development collection](https://github.com/neos/neos-development-collection) or turn [neos/fluidadaptor](https://github.com/neos/fluidadaptor) into a "real" repository - is that possible?

I'll vote for a real repository and give it its own life independent of Neos CMS.

albe:
> We could move it to the Neos development collection

Nah, then rather keep it in the Flow collection. Fluid has nothing really to do with Neos, other than being the default template language (still, but for how long?). The Fluid adaptor depends on Flow, so we do need to keep it in sync in some way. That's actually a reason to keep it in the Flow collection. But our experience shows, that the dependency on the framework is so shallow/stable (https://github.com/neos/flow-development-collection/commits/master/Neos.FluidAdaptor), that it should in fact be not too much of a problem (the widget stuff has a deeper dependence on the HTTP kernel).

Overall I'm okay with getting it an own repository. How do we solve the rendering of the viewhelper references into the RTD then?

Anyway, we can do this in a follow-up step - for starters, moving it from required to suggest is what we want, right?

> As a developer, I don't want to have Fluid installed every time I install Flow.

vs.

> As a maintainer, I don't want to have Fluid installed accidentially, hiding issues I'd have as a delevoper.

sorenmalling:
> Nah, then rather keep it in the Flow collection

The latest tries with installing `dev-master` of neos/flow, being blocked by the fluid adaptor dependency is my number one reason for moving it to it's own repository and live it's own life.

Only if we change the ViewInterface or HTTP stack, we should adapt that. But let that rather be a note for when touching that subject, rather than have it as a dependency in the development collection.

albe:
> The latest tries with installing dev-master of neos/flow, being blocked by the fluid adaptor dependency

Huh? Are you sure? I don't see how that should be affected by the package being maintained inside the same development repository?

sorenmalling:
Create a new project and require neos/flow dev-master and the fluid adapter will block it because of versions requirements.

Wrote about it on slack and remember it being about the split package and having fluid adapter as a dependency.

Written from iPhone - not by computer atm

bwaidelich (approved):
Great work, +1 (by reading)

sorenmalling:
ping @kdambekalns @kitsunet @albe 
How do we continue the "extraction" of Fluid (wow, that sounds super-science 🔬) as a dependency

albe:
> Overall I'm okay with getting it an own repository. How do we solve the rendering of the viewhelper references into the RTD then?
> 
> Anyway, we can do this in a follow-up step - for starters, moving it from required to suggest is what we want, right?

Still standing on this position from my side. Let's "suggest" this thing as is for now, then discuss how and if we want to move it to an own repository.

sorenmalling:
> Still standing on this position from my side. Let's "suggest" this thing as is for now, then discuss how and if we want to move it to an own repository.

Which is done for now in this ticket :)

sorenmalling:
ping @kdambekalns @kitsunet @albe @bwaidelich 

Can this go in? Or is there any routines, for the subsplit package, build process parts etc, when we remove a package? Let me know, if I can do anything to help on the topic

kdambekalns (changes requested):
The "collection-level" composer manifest needs to be adjusted, too (use `ComposerManifestMerger.php`.)

kdambekalns:
Oh… and when making this a suggestion, we need to actively require it for the RTD-reference-updating Jenkins jobs, I'd think. Testing would be another issue, but we run all tests on Neos, and that still needs Fluid–so those are still run for now…

bwaidelich (approved):
Thanks a lot for taking care

albe:
> actively require it for the RTD-reference-updating Jenkins jobs

Can we just add the `composer require neos-fluid-adaptor` in https://github.com/neos/flow-development-distribution/blob/master/Build/update-references.sh?

kdambekalns:
> Can we just add the `composer require neos-fluid-adaptor` in https://github.com/neos/flow-development-distribution/blob/master/Build/update-references.sh?

That should work, yes. Just require it along with `neos/doctools`…

kdambekalns (approved):
Fine with me!

albe (approved):

bwaidelich:
uhm.. This removed the dependency to the adaptor, but not to fluid itself !? https://github.com/neos/flow-development-collection/blob/7.0/composer.json#L34

kdambekalns:
Something else depends on it?!

albe:
Or the composer.json merger just doesn't regard the "suggest" for neos-fluidadaptor and just merges it's dependency?

Issue #2019, opened 2020-05-29
Convert existing HTTP Components to PSR-15 middleware


albe:
This is a follow-up to #1889 that is meant to track the status of "phasing out" the existing HTTP Components. 

```yaml
        'preprocess':
          position: 'before process'
          chain:
            'trustedProxies':
              position: 'start'
              component: 'Neos\Flow\Http\Component\TrustedProxiesComponent'
            'getSessionCookieFromRequest':
              position: 'after trustedProxies'
              component: 'Neos\Flow\Session\Http\SessionRequestComponent'

        'process':
          chain:
            'routing':
              position: 'start'
              component: 'Neos\Flow\Mvc\Routing\RoutingComponent'
            'parseBody':
              position: 'before dispatching'
              component: 'Neos\Flow\Http\Component\RequestBodyParsingComponent'
            'prepareActionRequest':
              position: 'after parseBody'
              component: 'Neos\Flow\Mvc\PrepareMvcRequestComponent'
            'dispatching':
              component: 'Neos\Flow\Mvc\DispatchComponent'
            'replaceHttpResponse':
              position: 'after dispatching'
              component: 'Neos\Flow\Http\Component\ReplaceHttpResponseComponent'
            'setHeader':
              position: 'after dispatching'
              component: 'Neos\Flow\Http\Component\SetHeaderComponent'
            'securityEntryPoint':
              position: 'after setHeaders'
              component: 'Neos\Flow\Http\Component\SecurityEntryPointComponent'

        'postprocess':
          chain:
            'setSessionCookie':
              position: 'start'
              component: 'Neos\Flow\Session\Http\SessionResponseComponent'
            'flashMessages':
              position: 'after setSessionCookie'
              component: 'Neos\Flow\Mvc\FlashMessage\FlashMessageComponent'
            'poweredByHeader':
              component: 'Neos\Flow\Http\Component\PoweredByComponent'
            'standardsCompliance':
              position: 'end'
              component: 'Neos\Flow\Http\Component\StandardsComplianceComponent'
```

This is the current (as of Flow 6.2) list of existing configured HTTP components. The order to convert them should be starting from pre-process, since the PSR-15 middlewares invoke the component chain. Hence the first step is the `TrustedProxiesComponent`. Also, some Components can probably be merged together (SessionRequest+Response), some need to be moved into the MVC stack maybe. We'll need to play and adapt there.

[...]

PR #2153, opened 2020-10-02
FEATURE: Move TrustedProxiesComponent to Middleware

sorenmalling:
Adjust TrustedProxiesComponent to match PSR-15 middleware implementation

- [x] Adjust/remove component tests

Related #2019


Neos.Flow/Classes/Http/Middleware/TrustedProxiesMiddleware.php
```diff
@@ -12,15 +14,18 @@
...
 use Neos\Flow\Utility\Ip as IpUtility;
+use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
-use Neos\Flow\Configuration\Exception\InvalidConfigurationException;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
 
 /**
- * HTTP component that checks request headers against a configured list of trusted proxy IP addresses.
+ * Midd leware that checks request headers against a configured list of trusted proxy IP addresses.
```
albe, line 26:
```suggestion
 * Middleware that checks request headers against a configured list of trusted proxy IP addresses.
```

sorenmalling:
@albe How do we port the tests from TrustedProxiesComponent to test the middleware?

albe:
> @albe How do we port the tests from TrustedProxiesComponent to test the middleware?

Just move the test to `../Middleware/TrustedProxiesMiddlewareTest.php`, then adjust the `callWithRequest()` method to invoke the middleware (`$middleware->process($request, $captureRequestHandler)`), capture the `$request` sent with the `$handler->handle($trustedRequest)` invokation and return that. So `$captureRequestHandler` basically needs to be a spy mock, that records the argument that the `handle()` method is invoked with.

sorenmalling:
> capture the `$request` sent with the `$handler->handle($trustedRequest)` invokation and return that. So `$captureRequestHandler` basically needs to be a spy mock, that records the argument that the `handle()` method is invoked with.

You lost me on that :) how do I capture the request being handled inside the middleware? I can commit what I tried so far if of any help

albe:
I was thinking something like
```php
class SpyRequestHandler {
     protected $handledRequest;
     public function handle(ServerRequest $request) {
         $this->handledRequest = $request;
     }
     public function getHandledRequest() {
         return $this->handledRequest;
    }
}

... 
function callWithRequest($request) {
    ...
    $spyHandler = new SpyRequestHandler();
    $response = $middleware->process($request, $spyHandler);
    return $spyHandler->getHandledRequest();
}
```
Alternatively, a spy is return from the PHPUnit `$this->any()` constraint: https://lyte.id.au/2014/03/01/spying-with-phpunit/
But I guess, that's not less code and is likely a bit less straightforward to grasp?

albe:
Green! 🟢🎉

albe (approved):

albe:
@bwaidelich I would suggest to merge this now and then rebase the next PR that moves the components to middleware and repeat, wdyt?

bwaidelich:
Makes sense! Just wondering: Will this affect the current documentation?
In any case we should mention here: https://neos-project.slack.com/archives/C04Q0TC15/p1601626445022900

albe:
Yes, it will affect documentation, as that still mentions the component chain. I would suggest to create an own documentation PR for the whole move and link that to the parent issue #2019. I can prepare a first draft for the docs change sometime this week.

albe:
@bwaidelich I'll merge this for now to move forward - if we do find this components->middleware change to not get through completely or create problems, then we can revert before branching off.

PR #2154, opened 2020-10-02
FEATURE: Move Session*Component to Middleware

sorenmalling:
Combine SessionRequestComponent and SessionResponseComponent into a single PSR-15 middleware implementation

- [x] Adjust/remove component tests

Related #2019
Depends on #2203


Neos.Flow/Classes/Http/Middleware/SessionMiddleware.php
```diff
@@ -37,30 +40,36 @@ class SessionRequestComponent implements ComponentInterface
      */
     protected $sessionManager;
 
-    /**
-     * @param ComponentContext $componentContext
-     */
-    public function handle(ComponentContext $componentContext)
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
albe, line 43:
```suggestion
    public function process(ServerRequestInterface $request, RequestHandlerInterface $next): ResponseInterface
```
`$next` transports the information that this handler is the next in the chain and avoids the awkward `$handler->handle()` invocation

Neos.Flow/Classes/Http/Middleware/SessionMiddleware.php
```diff
@@ -37,30 +40,36 @@ class SessionRequestComponent implements ComponentInterface
...
+
+        $response = $handler->handle($request);
+
+        $currentSession = $this->sessionManager->getCurrentSession();
+        if (!$currentSession->isStarted()) {
+            return $response;
+        }
+
+        $response = $response->withAddedHeader('Set-Cookie', (string)$currentSession->getSessionCookie());
+        return $response;
```
albe, line 72:
```suggestion
        return $response->withAddedHeader('Set-Cookie', (string)$currentSession->getSessionCookie());
```

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -26,6 +26,9 @@ Neos:
 
       # Defines the PSR-15 middlewares that a request goes through in a flow
       middlewares:
+        'session':
```
bwaidelich, line 29: shouldn't that be placed much further back in the middleware chain?
It used to be in the `postprocess` chain before so it should probably be put after the `dispatch` middleware at least

bwaidelich: Ah, nevermind.. I assume it's the other way around now :)

sorenmalling: It "wraps" the request in the implementation. Maybe we need to split it to two parts again?

bwaidelich: I'm not sure about the best practices here but my feeling is that excessive nesting will make it harder to debug since it basically makes the stack trace of every call between one step longer

sorenmalling:
Come to think of it, I might have made a bad choice in git history. I "git mv" the original http component. We should remove/delete it and create the middleware as new file - right?

albe:
Isn't `git mv` just an alias for `git rm` + `git add`? AFAIK git does not store git commands in the history, but only the diffs and infers the operation from that when showing the history. I read some argument from Linus back when this was discussed for implementation, as to why this makes more sense for a source control system.

https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -26,6 +26,9 @@ Neos:
...
+          position: '10'
```
albe, line 30: Let's change this to `after trustedProxies` after rebasing onto #2153 - once we get e.g. FlashMessages done, that will need to be moved in between or even before trustedProxies.

Neos.Flow/Classes/Http/Middleware/SessionMiddleware.php
```diff
@@ -37,30 +40,36 @@ class SessionRequestComponent implements ComponentInterface
...
 
-    /**
-     * @param ComponentContext $componentContext
-     */
-    public function handle(ComponentContext $componentContext)
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
         if (!$this->sessionManager instanceof SessionManager) {
-            return;
+            return $handler->handle($request);
```
albe, line 46:
```suggestion
            return $next->handle($request);
```

Neos.Flow/Classes/Http/Middleware/SessionMiddleware.php
```diff
@@ -37,30 +40,36 @@ class SessionRequestComponent implements ComponentInterface
...
         $sessionCookieName = $this->sessionSettings['name'];
         /** @var ServerRequestInterface $request */
-        $request = $componentContext->getHttpRequest();
         $cookies = $request->getCookieParams();
 
         if (!isset($cookies[$sessionCookieName])) {
             $sessionCookie = $this->prepareCookie($sessionCookieName, Algorithms::generateRandomString(32));
             $this->sessionManager->createCurrentSessionFromCookie($sessionCookie);
-            return;
+            return $handler->handle($request);
```
albe, line 56:
```suggestion
            return $next->handle($request);
```

Neos.Flow/Classes/Http/Middleware/SessionMiddleware.php
```diff
@@ -37,30 +40,36 @@ class SessionRequestComponent implements ComponentInterface
...
         }
 
         $sessionIdentifier = $cookies[$sessionCookieName];
-        $sessionCookie = $this->prepareCookie($sessionCookieName, $sessionIdentifier);
+        $sessionCookie = $this->prepareCookie((string) $sessionCookieName, (string) $sessionIdentifier);
         $this->sessionManager->initializeCurrentSessionFromCookie($sessionCookie);
         $this->sessionManager->getCurrentSession()->resume();
+
+        $response = $handler->handle($request);
```
albe, line 64:
```suggestion
        $response = $next->handle($request);
```

Neos.Flow/Configuration/Settings.Http.yaml
albe: btw., that's the reason the tests currently fail:
`flashMessagesGetRenderedAfterRedirect` - we need to move the flashmessages component first

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -70,9 +70,6 @@ Neos:
 
         'postprocess':
           chain:
-            'setSessionCookie':
-              position: 'start'
-              component: 'Neos\Flow\Session\Http\SessionResponseComponent'
             'flashMessages':
               position: 'before setSessionCookie'
```
albe, line 74:
```suggestion
              position: 'start'
```

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -42,9 +45,6 @@ Neos:
         'preprocess':
           position: 'before process'
           chain:
```
albe, line 47:
```suggestion
          chain: []
```

Neos.Flow/Classes/Http/Middleware/SessionMiddleware.php
```diff
@@ -37,30 +40,35 @@ class SessionRequestComponent implements ComponentInterface
...
         $sessionCookieName = $this->sessionSettings['name'];
         /** @var ServerRequestInterface $request */
-        $request = $componentContext->getHttpRequest();
         $cookies = $request->getCookieParams();
 
         if (!isset($cookies[$sessionCookieName])) {
             $sessionCookie = $this->prepareCookie($sessionCookieName, Algorithms::generateRandomString(32));
             $this->sessionManager->createCurrentSessionFromCookie($sessionCookie);
-            return;
+            return $next->handle($request);
```
albe, line 56: We need to handle the setSession case here instead of return - that's the reason the test fails.

albe:
Okay, found a bug in 6.2+ - the InternalRequestEngine is not invoking the middleware chain 🙈 hence test still fails

Neos.Flow/Configuration/Settings.Http.yaml
albe: That's indeed a drawback of the middlewares/onion approach, so let's keep those merged.

sorenmalling (commented):
I can approve you added changes - but can not do it in a review (being the original author)

albe:
Okay, this needs a rebase and then hopefully the tests should become green

albe:
Hm... I'm... not sure why that still fails with the flash messages test 🤔 

> 1) Neos\Flow\Tests\Functional\Mvc\ActionControllerTest::flashMessagesGetRenderedAfterRedirect
Failed asserting that an array is not empty.

The session middleware should now resume a session before it passes down into the components, so the FlashMessageComponent should see the session. In turn if the FlashMessageComponent starts a session, the Session middleware should set the cookie accordingly. Also the internal requestEngine should now use the middlewares correctly.

sorenmalling:
@albe We are green 🟢

albe (approved):
Yay, no 2

PR #2155, opened 2020-10-07
BUGFIX: Don't convert route options to strings when merging subroutes

bwaidelich:
Adjusts the `RouteConfigurationProcessor` so that it doesn't implicitly
convert route configuration options to strings when merging routes.

Background:

This fixes a regression introduced with #2145 that lead to the Neos
route

```yaml
-
  name: 'Homepage'
  uriPattern: '{node}'
  routeParts:
    'node':
      options:
        onlyMatchSiteNodes: true
```

to be converted such that the `onlyMatchSiteNodes` option was casted
to a string of `"1"` leading to the homepage route resolving all node links.

Resolves: #2142

### DIFF
--- a/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
+++ b/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
@@ -212,7 +212,7 @@ class RouteConfigurationProcessor
      *
      * @param string|array $value
      * @param array $variables
-     * @return string
+     * @return mixed
      */
     protected function replacePlaceholders($value, array $variables)
     {
--- a/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
+++ b/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
@@ -217,7 +217,7 @@ class RouteConfigurationProcessor
     protected function replacePlaceholders($value, array $variables)
     {
         if (is_array($value)) {
-            foreach($value as $arrayKey => $arrayValue) {
+            foreach ($value as $arrayKey => $arrayValue) {
                 $value[$arrayKey] = $this->replacePlaceholders($arrayValue, $variables);
             }
         } elseif (is_string($value)) {
--- a/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
+++ b/Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
@@ -193,16 +193,10 @@ class RouteConfigurationProcessor
                     $mergedSubRouteConfiguration['uriPattern'] = rtrim($this->replacePlaceholders($routeConfiguration['uriPattern'], [$subRouteKey => '']), '/');
                 }
                 if (isset($mergedSubRouteConfiguration['defaults'])) {
-                    foreach ($mergedSubRouteConfiguration['defaults'] as $key => $defaultValue) {
-                        $mergedSubRouteConfiguration['defaults'][$key] = $this->replacePlaceholders($defaultValue, $variables);
-                    }
+                    $mergedSubRouteConfiguration['defaults'] = $this->replacePlaceholders($mergedSubRouteConfiguration['defaults'], $variables);
                 }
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
```diff
@@ -216,16 +210,21 @@ protected function buildSubRouteConfigurations(array $routesConfiguration, array
...
-    protected function replacePlaceholders($string, array $variables)
+    protected function replacePlaceholders($value, array $variables)
     {
-        foreach ($variables as $variableName => $variableValue) {
-            $string = str_replace('<' . $variableName . '>', $variableValue, $string);
+        if (is_array($value)) {
+            foreach($value as $arrayKey => $arrayValue) {
+                $value[$arrayKey] = $this->replacePlaceholders($arrayValue, $variables);
+            }
+        } elseif (is_string($value)) {
```
bwaidelich, line 223: The `is_string()` check is the actual fix, but it led to wrong results with (nested) array values. So I tweaked this method to work recursively which leads to less code in the lines above as a nice side effect

Neos.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
```diff
@@ -1441,7 +1441,58 @@ public function buildSubrouteConfigurationsMergesSubRoutesAndProcessesPlaceholde
         $routeConfigurationProcessor = $this->getAccessibleMock(RouteConfigurationProcessor::class, ['dummy'], [], '', false);
         $actualResult = $routeConfigurationProcessor->_call('buildSubrouteConfigurations', $routesConfiguration, $subRoutesConfiguration, 'WelcomeSubroutes', $subRouteOptions);
 
-        self::assertEquals($expectedResult, $actualResult);
+        self::assertSame($expectedResult, $actualResult);
```
bwaidelich, line 1444: Note: never use `assertEquals`..

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
bwaidelich: btw, did you know: https://3v4l.org/F49Dq o.O

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
```diff
@@ -193,16 +193,10 @@ protected function buildSubRouteConfigurations(array $routesConfiguration, array
...
-                    }
+                    $mergedSubRouteConfiguration['defaults'] = $this->replacePlaceholders($mergedSubRouteConfiguration['defaults'], $variables);
                 }
                 if (isset($mergedSubRouteConfiguration['routeParts'])) {
-                    foreach ($mergedSubRouteConfiguration['routeParts'] as $routePartKey => $routePartValue) {
-                        foreach ($mergedSubRouteConfiguration['routeParts'][$routePartKey]['options'] ?? [] as $optionKey => $optionValue) {
-                            $mergedSubRouteConfiguration['routeParts'][$routePartKey]['options'][$optionKey] = $this->replacePlaceholders($optionValue, $variables);
-                        }
-                    }
+                    $mergedSubRouteConfiguration['routeParts'] = $this->replacePlaceholders($mergedSubRouteConfiguration['routeParts'], $variables);
```
albe, line 199: This will replace variables in all subkeys of 'routeParts', not only in the options - could this be an(other) issue? It could e.g. parameterize the handler name

albe (approved):
Overall a very good solution to make `replacePlaceholders` recursive. Note my comment about this now replacing more than just option variables.

Neos.Flow/Classes/Configuration/RouteConfigurationProcessor.php
bwaidelich: I thought about allowing this in the original change but didn't consider it valuable enough to make the code more complex.
But I don't think that it is a problem. I updated the issue description and added a comment (https://github.com/neos/flow-development-collection/issues/2142#issuecomment-704769881)

Issue #2084, opened 2020-08-12
Unable to generate links / actions in CLI mode


SC-cmarquardt:
### Description
Currently I am unable to create a link or an URI with `f:link.action` (or corresponding URI method) on a Windows server in a CommandController. I set `Neos.Flow.http.baseUri` but for some reason the ViewHelper creates a local file-based path, which means s.th. like `E:\my\path\to\flow\index.php`.
Is this a know issue? Also setting the `absolute` parameter to false for the ViewHelper does not change anything.

### Used Versions

Flow: 6.1.4

sorenmalling:
Hi @SC-cmarquardt 

The issue and topic have been raised in this former ticket

https://github.com/neos/flow-development-collection/issues/1848

Will you try and see if the provided examples and referred links helps you? If not, please open a thread on discuss.neos.io so we can get more people involved in the debugging. If the issue remain, this issue ticket is absolutely valid :-)

SC-cmarquardt:
Hi @sorenmalling,

thanks for your reply.
Christian (kitsunet) writes here https://github.com/neos/flow-development-collection/issues/1848#issuecomment-554409888 that in Flow 6 things behave differently and it should be easier to create a "fake" request.
Do you have an example for this fake request?

sorenmalling:
@SC-cmarquardt 

I don't know of such thing - I personally create links in a more "static way" (typing the url and using a setting for root domain) to avoid the CLI issue with trying to create urls.

I suggest that you create a ticket on discuss.neos.io with this question - more will probably benefit from it and can help on solving this

sorenmalling:
Did you try the example Christian provides https://github.com/neos/flow-development-collection/issues/1848#issuecomment-554410993

albe:
The most "correct" way as of Flow 6.0 to get a fake `ActionRequest` which you can pipe into the StandaloneView would be like this:
```php

/**
 * @Flow\Inject
 * @var ActionRequestFactory
 */
protected $actionRequestFactory;

/**
 * @Flow\Inject
 * @var \Neos\Http\Factories\ServerRequestFactory
 */
protected $serverRequestFactory;

...

$fakeHttpRequest = $this->serverRequestFactory->createServerRequest('GET', 'https://acme.com/your/fake/uri', $optionalHTTParguments);
$fakeActionRequest = $this->actionRequestFactory->createActionRequest($fakeHttpRequest, []);
$view = new StandaloneView($fakeActionRequest);
```
[...]

Issue #2157, opened 2020-10-08
Get rid of configured baseUri


bwaidelich:
Since the early days, Flow comes with a setting for the [baseUri](https://github.com/neos/flow-development-collection/blob/bf990307957630654f380aed3242249fa775b0d5/Neos.Flow/Configuration/Settings.Http.yaml#L25)

But this whole concept has been source of confusion and bugs (see #2084 for example) and doesn't make sense in muliti-domain scenarios (like multi site Neos installations).

The `BaseUriProvider` that was introduced with #1755 is currently used in the following 7 places in the core packages:

* **Neos.Flow:**
** [UriBuilder](https://github.com/neos/flow-development-collection/blob/bf990307957630654f380aed3242249fa775b0d5/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php#L364)
** [FileSystemTarget](https://github.com/neos/flow-development-collection/blob/bf990307957630654f380aed3242249fa775b0d5/Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php#L393)
** [WebRedirect](https://github.com/neos/flow-development-collection/blob/bf990307957630654f380aed3242249fa775b0d5/Neos.Flow/Classes/Security/Authentication/EntryPoint/WebRedirect.php#L55)
* **Neos.Neos:**
** [LinkingService #1](https://github.com/neos/neos-development-collection/blob/0247643f1dde66fcc31d49f999aedba5bd2e58d2/Neos.Neos/Classes/Service/LinkingService.php#L325), [LinkingService #2](https://github.com/neos/neos-development-collection/blob/0247643f1dde66fcc31d49f999aedba5bd2e58d2/Neos.Neos/Classes/Service/LinkingService.php#L360)
* **Neos.Fusion:**
** [BaseUriHelper](https://github.com/neos/neos-development-collection/blob/0247643f1dde66fcc31d49f999aedba5bd2e58d2/Neos.Fusion/Classes/Eel/BaseUriHelper.php#L34)
* **Neos.Form:**
** [RedirectFinisher](https://github.com/neos/form/blob/85feb5d2a3f9d6e0fe638f2124257eb7647d67f9/Classes/Finishers/RedirectFinisher.php#L79)

In all of these cases it should be possible to just fall back to the current ServerRequest (which is either available from the ActionRequest or can be injected (see #2144)).

For cases where it's not possible to get hold of the active request (e.g. in CLI context) it can be created just like before.
[...]

PR #2158, opened 2020-10-08
BUGFIX: Allow URI-building from CLI context

bwaidelich:
This adds a parameter `$fallbackRequest` to `BaseUriProvider::getConfiguredBaseUriOrFallbackToCurrentRequest()`
that allows it to work in contextx where the active HTTP request
can't get hold of (i.g. in CLI mode).

This also adjusts `UriBuilder::build()` to make use of that new parameter
such that the following code works again from CLI:

```php
$httpRequest = ServerRequest::fromGlobals();
$actionRequest = ActionRequest::fromHttpRequest($httpRequest);
$uriBuilder = new UriBuilder();
$uriBuilder->setRequest($actionRequest);
$uriBuilder->uriFor(...);
```

Previously this would throw an HTTP exception
```
No base URI could be provided. This probably means a call was made outside of an HTTP
request and a base URI was neither configured nor set during runtime.
```
when executed in CLI context.

Fixes: #2084

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
+++ b/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
@@ -361,7 +361,7 @@ class UriBuilder
         $uriPathPrefix = RequestInformationHelper::getScriptRequestPath($httpRequest) . $uriPathPrefix;
         $uriPathPrefix = ltrim($uriPathPrefix, '/');
 
-        $resolveContext = new ResolveContext($this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest(), $arguments, $this->createAbsoluteUri, $uriPathPrefix);
+        $resolveContext = new ResolveContext($this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest($httpRequest), $arguments, $this->createAbsoluteUri, $uriPathPrefix);
         $resolvedUri = $this->router->resolve($resolveContext);
         if ($this->section !== '') {
             $resolvedUri = $resolvedUri->withFragment($this->section);
--- a/Neos.Flow/Classes/Http/BaseUriProvider.php
+++ b/Neos.Flow/Classes/Http/BaseUriProvider.php
@@ -5,6 +5,7 @@ use GuzzleHttp\Psr7\Uri;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Core\Bootstrap;
 use Neos\Flow\Http\Helper\RequestInformationHelper;
+use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\UriInterface;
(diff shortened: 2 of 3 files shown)


daniellienert (approved):
Looks good from reading.

Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
```diff
@@ -361,7 +361,7 @@ public function build(array $arguments = [])
         $uriPathPrefix = RequestInformationHelper::getScriptRequestPath($httpRequest) . $uriPathPrefix;
         $uriPathPrefix = ltrim($uriPathPrefix, '/');
 
-        $resolveContext = new ResolveContext($this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest(), $arguments, $this->createAbsoluteUri, $uriPathPrefix);
+        $resolveContext = new ResolveContext($this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest($httpRequest), $arguments, $this->createAbsoluteUri, $uriPathPrefix);
```
albe, line 364: Aren't we running in circles with this? As in the `$httpRequest` in CLI here is just the fake http request that got made up to create the ActionRequest to inject into the UriBuilder? So what is the URI of that fake http request? And can't we simply use that here directly?

albe (commented):
Code looks good and there are tests 🎉 but I'm not so sure about this approach to solve the CLI uri building requires a HTTP request issue underneath?

Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
bwaidelich: That's what I also suggested when I talked to @kitsunet. But this would change the behavior as currently the configured baseUri has precedence over the current request URI.

Or, to put it simple:
When you have a `baseUri` configured, links will have that URI prefixed if you come from a different domain.

bwaidelich:
@albe Thanks for your comment. IMO this is a fix with the least risk of breaking code. It's not a beauty but it is just a temporary solution anyways.

> I'm not so sure about this approach to solve the CLI uri building requires a HTTP request issue underneath?

I don't understand that question?

To use the `UriBuilder` you'll have to have/create an `ActionRequest` (and always had to). And that in turn always requires a HTTP request.
The issue was that it didn't even work with a specified HTTP request, because that was basically ignored

kdambekalns (approved):

albe:
> To use the UriBuilder you'll have to have/create an ActionRequest (and always had to). And that in turn always requires a HTTP request.

Yes, but that is the real issue of "URI-building from CLI context" (see my discussion points in #2157). But you are right, this is a bugfix for a regression, not a conceptual refactor 👍

albe (approved):

PR #2163, opened 2020-10-11
TASK: Property typing for policy classes

daniellienert:


sorenmalling (approved):
Looks good from reading on mobile

Neos.Flow/Classes/Security/Context.php
```diff
@@ -396,6 +396,7 @@ public function getAuthenticationTokensOfType($className)
      * @return Role[]
      * @throws Exception
      * @throws Exception\NoSuchRoleException
+     * @throws \Neos\Flow\Configuration\Exception\InvalidConfigurationTypeException
```
bwaidelich, line 399: nitpick: import namespace?

Neos.Flow/Classes/Security/Policy/PolicyService.php
```diff
@@ -279,11 +292,13 @@ public function getPrivilegeTargets()
      *
      * @param string $privilegeTargetIdentifier Identifier of a privilege target
      * @return PrivilegeTarget
```
bwaidelich, line 294:
```suggestion
     * @return ?PrivilegeTarget
```

Neos.Flow/Classes/Security/Policy/Role.php
```diff
@@ -19,7 +21,7 @@
  */
 class Role
 {
-    const ROLE_IDENTIFIER_PATTERN = '/^(\w+(?:\.\w+)*)\:(\w+)$/';   // Vendor(.Package)?:RoleName
+    protected const ROLE_IDENTIFIER_PATTERN = '/^(\w+(?:\.\w+)*)\:(\w+)$/';   // Vendor(.Package)?:RoleName
```
bwaidelich, line 24: I'd say. Let's make the class `final` and this constant `private`

bwaidelich (commented):
Nice one, thanks for taking care!
Just a few (mostly nitpicky) comments

Neos.Flow/Classes/Security/Policy/PolicyService.php
albe: Do we actually use the `?` optional type prefix in docblocks? IIRC our current reflection does not parse that correctly - the alternative is `|null`

Neos.Flow/Classes/Security/Policy/Role.php
albe: Devil's advocat: Making that class final would have to be considered breaking and be documented clearly. Let's maybe do this in #1939 instead?

Neos.Flow/Classes/Security/Policy/PolicyService.php
bwaidelich: yes, you're right. Updated the suggestion!

Neos.Flow/Classes/Security/Policy/Role.php
bwaidelich: Yep, good point.

daniellienert:
@bwaidelich, sorry had to remove the `final` from the policy class again, as we mock it in tests. Already this task consumed a lot of time and I would like to merge it as is.

Issue #2162, opened 2020-10-11
FEATURE: Add meta data to roles and privileges


daniellienert:
When lots of roles (targeting lots of privileges) are defined, managing them and assigning them to users in the Neos backend gets quite complicated and error prone. The major issue is, that we only have the role identifier (somehow truncated) to be displayed in the backend module.

Having two additional (optional) configuration options, label and description for both configuration types would make things much clearer.

Example:

```
...
    'Neos.Neos:Backend.Service.Workspaces.Index':
      matcher: 'method(Neos\Neos\Controller\Service\WorkspacesController->(index|error|show)Action())'
      label: 'Index Workspaces'
      description: 'List workspaces in the Neos backend module'
...

...
roles:
  'Neos.Neos:LivePublisher':
    label: Neos Live Publisher'
    description: Enables the user to publish changes to the live workspace
    privileges:
...
```

The result in Neos could be something like this, when clicking on the ℹ️ icon:

![image](https://user-images.githubusercontent.com/642226/95674409-57325c80-0bb0-11eb-907c-a1b37632ef21.png)

daniellienert:
@bwaidelich, would love to read your thoughts on this.

bwaidelich:
I think it's a great idea without top much risk involved!

albe:
I second that assessment by Bastian :) That makes so much sense and should be painless to implement!

sorenmalling:
If implanted I suggest them being used/displayed in a cli context/command.

Can we make that?

daniellienert:
Wasn't that hard. Currently struggling with the necessary css and javascript magic to display it kind of clear. (Should just have used a command controller .... 😂 ) 
![image](https://user-images.githubusercontent.com/642226/95691416-5ab0fc80-0c1f-11eb-8469-7dd3a09a8de8.png)

PR #2166, opened 2020-10-12
FEATURE: Add meta data to roles and privilegeTargets

daniellienert:
This adds the optional configuration values label and description
to role definitions and label to privilege targets. The meta data can
be used to document roles and privilegeTarget and to guide
administrators to assign the correct roles to users.

Example:

```
  'Neos.Neos:UserManager':
    label: Neos User Manager
    description: A user with this role is able to create, edit and delete users which has the same or a subset of his own roles.
    privileges:
    ...
```

Resolves: #2162

### DIFF
--- a/Neos.Flow/Classes/Security/Authorization/Privilege/AbstractPrivilege.php
+++ b/Neos.Flow/Classes/Security/Authorization/Privilege/AbstractPrivilege.php
@@ -190,7 +190,7 @@ abstract class AbstractPrivilege implements PrivilegeInterface
      *
      * @return string
      */
-    public function getParsedMatcher()
+    public function getParsedMatcher(): string
     {
         $parsedMatcher = $this->matcher;
         // TODO: handle parameters that are not strings
--- a/Neos.Flow/Classes/Security/Policy/PolicyService.php
+++ b/Neos.Flow/Classes/Security/Policy/PolicyService.php
@@ -114,12 +114,8 @@ class PolicyService
                 if ($roleIdentifier === 'Neos.Flow:Everybody') {
                     $role = $everybodyRole;
                 } else {
-                    $role = new Role($roleIdentifier);
+                    $role = new Role($roleIdentifier, [], (string)($roleConfiguration['label'] ?? ''), (string)($roleConfiguration['description'] ?? ''));
                     $role->setAbstract((bool)($roleConfiguration['abstract'] ?? false));
-                    if (isset($roleConfiguration['label'])) {
-                        $role->setLabel((string)$roleConfiguration['label']);
-                    }
-                    $role->setDescription((string)($roleConfiguration['description'] ?? ''));
                 }
 
                 if (isset($roleConfiguration['privileges'])) {
--- a/Neos.Flow/Classes/Security/Policy/PolicyService.php
+++ b/Neos.Flow/Classes/Security/Policy/PolicyService.php
@@ -115,9 +115,11 @@ class PolicyService
                     $role = $everybodyRole;
                 } else {
                     $role = new Role($roleIdentifier);
-                    if (isset($roleConfiguration['abstract'])) {
-                        $role->setAbstract((boolean)$roleConfiguration['abstract']);
+                    $role->setAbstract((bool)($roleConfiguration['abstract'] ?? false));
+                    if (isset($roleConfiguration['label'])) {
+                        $role->setLabel((string)$roleConfiguration['label']);
                     }
+                    $role->setDescription((string)($roleConfiguration['description'] ?? ''));
                 }
 
                 if (isset($roleConfiguration['privileges'])) {
@@ -183,6 +185,7 @@ class PolicyService
         if (!isset($this->policyConfiguration['privilegeTargets'])) {
             return;
         }
+
         foreach ($this->policyConfiguration['privilegeTargets'] as $privilegeClassName => $privilegeTargetsConfiguration) {
             foreach ($privilegeTargetsConfiguration as $privilegeTargetIdentifier => $privilegeTargetConfiguration) {
                 if (!isset($privilegeTargetConfiguration['matcher'])) {
@@ -199,7 +202,9 @@ class PolicyService
                     }
                     $parameterDefinitions[$parameterName] = new PrivilegeParameterDefinition($parameterName, $privilegeTargetConfiguration['parameters'][$parameterName]['className']);
                 }
-                $privilegeTarget = new PrivilegeTarget($privilegeTargetIdentifier, $privilegeClassName, $privilegeTargetConfiguration['matcher'], $parameterDefinitions);
+
+                $label = $privilegeTargetConfiguration['label'] ?? $privilegeTargetIdentifier;
+                $privilegeTarget = new PrivilegeTarget($privilegeTargetIdentifier, $privilegeClassName, $privilegeTargetConfiguration['matcher'], $parameterDefinitions, $label);
                 $privilegeTarget->injectObjectManager($this->objectManager);
                 $this->privilegeTargets[$privilegeTargetIdentifier] = $privilegeTarget;
             }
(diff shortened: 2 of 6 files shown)


sorenmalling:
Should label and description be nullable instead of empty strings?

And can we use this meta data in command controller so this is not a Neos CMS-only change?

bwaidelich (dismissed):
Nice direction!!
Apart from the one typo I have another concern though.
I think we should refrain from mixing behavioral changes with refactorings and CGL adjustments as it
1) makes it harder to review the actual change
2) increases the risk of breaking changes to be mixed into non-breaking changes (e.g. remember when we introduced type hints to Eel helpers and it broke for many people)

So personally I'd vote to extract the new type hints, exception messages, access modifieds, ... into a separate PR – As much sense as they make!

Regarding @sorenmalling s question:

I think it's just fine to make the new fields _not_ nullable – I try to avoid those as much as I can since it makes the using code more complex (you'll always have to check a 3rd case and null pointer exceptions can occur)

As for CLI commands – I think it would be nice to have those. But AFAIK we don't yet have a command that lists all the roles and their privileges for example.. It would be great to add this, but IMO also a separate issue & PR

albe (approved):
Looks good. Just an error message nitpick

sorenmalling:
> As for CLI commands – I think it would be nice to have those. But AFAIK we don't yet have a command that lists all the roles and their privileges for example.. It would be great to add this, but IMO also a separate issue & PR

Good point 👍

Neos.Flow/Classes/Security/Authorization/Privilege/AbstractPrivilege.php
```diff
@@ -72,14 +72,16 @@ public function injectObjectManager(ObjectManagerInterface $objectManager)
      * @param PrivilegeTarget $privilegeTarget
      * @param string $matcher
      * @param string $permission One of the constants GRANT, DENY or ABSTAIN
-     * @param PrivilegeParameterInterface[] $parameters
+     * @param array $parameters
```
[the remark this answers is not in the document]
albe: Guess because the `$parameters` may now also contain string values :)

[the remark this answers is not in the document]
bwaidelich: Is that changed with this pr?

daniellienert:
Dear @bwaidelich, @albe, @sorenmalling,
as getting reviews timely is sometimes the hardest task in adding a new feature: Thank you all for the reviews!!!

> So personally I'd vote to extract the new type hints, exception messages, access modifieds, ... into a separate PR – As much sense as they make!

I agree absolutely! It is this one: https://github.com/neos/flow-development-collection/pull/2163. Should be merged already, just need to convince Psalm. The PR should loose all the junk then.

I will take care of your suggestions ASAP.

bwaidelich:
> I agree absolutely! It is this one: #2163

you're the man!

Neos.Flow/Classes/Security/Authorization/Privilege/AbstractPrivilege.php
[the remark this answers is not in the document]
daniellienert: Nope, this change is just wrong

Neos.Flow/Classes/Security/Authorization/Privilege/AbstractPrivilege.php
```diff
@@ -72,14 +72,16 @@ public function injectObjectManager(ObjectManagerInterface $objectManager)
...
      */
     public function __construct(PrivilegeTarget $privilegeTarget, $matcher, $permission, array $parameters)
     {
         $this->privilegeTarget = $privilegeTarget;
         $this->matcher = $matcher;
         $this->permission = $permission;
         $this->parameters = $parameters;
+        $this->label = $parameters['label'] ?? '';
+        $this->description = $parameters['descritption'] ?? '';
```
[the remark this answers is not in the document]
daniellienert: They shouldn't, neither this assignments. The meta data is part of the target not the privilege. Was some leftover code.

daniellienert:
Hey, the PR now contains mostly functional changes.

Something. I want to point you to: I added `label` and `description` to the role as I think a role deserves a readable name and also sometimes needs a description. I only added `label` to the privilege target as I thought that should be enough. What do you think?

Neos.Flow/Classes/Security/Policy/PolicyService.php
```diff
@@ -115,9 +115,11 @@ protected function initialize(): void
                     $role = $everybodyRole;
                 } else {
                     $role = new Role($roleIdentifier);
-                    if (isset($roleConfiguration['abstract'])) {
-                        $role->setAbstract((boolean)$roleConfiguration['abstract']);
+                    $role->setAbstract((bool)($roleConfiguration['abstract'] ?? false));
+                    if (isset($roleConfiguration['label'])) {
+                        $role->setLabel((string)$roleConfiguration['label']);
                     }
```
albe, line 121:
```suggestion
                    $role->setLabel((string)($roleConfiguration['label'] ?? $roleIdentifier));
```
Maybe to have those lines in line?

Neos.Flow/Classes/Security/Policy/Role.php
```diff
@@ -75,6 +89,7 @@ public function __construct(string $identifier, array $parentRoles = [])
         $this->identifier = $identifier;
         $this->packageKey = $matches[1];
         $this->name = $matches[2];
+        $this->label = $matches[2];
```
albe, line 92:
```suggestion
        $this->label = $matches[2];
        $this->description = '';
```
Otherwise description can be null (type mismatch to getter typehint).

albe (commented):
Found one type mismatch and a codestyle nitpick.

Regarding the question about "only label for PrivilegeTarget": Not sure - I guess you could put more than a (short) label on a target? Also in your issue and screenshot you had a description on them 🤷‍♂️

Neos.Flow/Classes/Security/Policy/PolicyService.php
daniellienert: That would mean that the default set in https://github.com/neos/flow-development-collection/pull/2166/files#diff-a6e6a5cf33830d6eb63ae27a24b440dcR92 would be overwritten in any case. I would prefer the split part there.

Neos.Flow/Classes/Security/Policy/Role.php
daniellienert: Good catch. Wouln't it be better to set the default in the property?

daniellienert:
@albe 
> Regarding the question about "only label for PrivilegeTarget": Not sure - I guess you could put more than a (short) label on a target? Also in your issue and screenshot you had a description on them 🤷‍♂️
In https://user-images.githubusercontent.com/642226/95691416-5ab0fc80-0c1f-11eb-8469-7dd3a09a8de8.png only the policy has a description. The privilegeTarget is shown with its label. I don't mind to add the description. It's optional anyway. But then we need a place to display it otherwise its useless.

Neos.Flow/Classes/Security/Policy/Role.php
albe: Is effectively equivalent, but I kept the initializers in the constructor together in the suggestion. Whichever you prefer is fine :)

sorenmalling (approved):
Looks find and simple add on. I do not have a usecase for testing it deeper, but Daniels task on Neos CMS for displaying it in the user mangemenet is a good test case 👍

Neos.Flow/Classes/Security/Policy/Role.php
bwaidelich: Why not add `label` and `description` to the constructor aruguments?
I would certainly expect this and would even tend to make all setters deprecated at some point in favor of immutability

bwaidelich: @daniellienert @albe _ping_  :)

daniellienert: Changed as suggested by @bwaidelich

bwaidelich (approved):
+1 by reading, thank you!!

bwaidelich:
While adding documentation for this feature I realized that we seem to have forgotten to adjust the schema, so defining labels/descriptions will lead to a warning in the Neos backend.. I'll create a bugfix

bwaidelich:
=> https://github.com/neos/flow-development-collection/pull/2353

PR #2167, opened 2020-10-12
BUGFIX: FileSystemStorage::getObjects correctly returns a generator of StorageObject

albe:
Somehow this went unnoticed and the `getObjects()` method returned a generator generator. Also the element type docblock was wrong.


albe:
See also https://github.com/neos/flow-development-collection/pull/1931#discussion_r503143247

kdambekalns (approved):
👀

PR #2169, opened 2020-10-13
FEATURE: Introduce connection factory

sorenmalling:
Questions:
 - [x] Should it be registred in Objects.yaml that it is used when injection the Connection directly?
 -  ~Should it have a options argument, so you can override configuration?~ (nope - see comment below)
 - [x] When using `singleton` in Objects.yaml can a custom factory override it (write tests and create test package) (@sorenmalling) (https://github.com/neos/flow-development-collection/pull/2169#discussion_r503796927) (find result in https://github.com/neos/flow-development-collection/pull/2169#issuecomment-708324572)

Resolves: #2170

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/ConnectionFactory.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/ConnectionFactory.php
@@ -20,7 +20,7 @@ use Doctrine\ORM\EntityManagerInterface;
  *
  * @Flow\Scope("singleton")
  */
-class ConnectionFactory
+final class ConnectionFactory
 {
     /**
      * @Flow\Inject
--- a/Neos.Flow/Classes/Persistence/Doctrine/ConnectionFactory.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/ConnectionFactory.php
@@ -1,0 +1,44 @@
+<?php
+namespace Neos\Flow\Persistence\Doctrine;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Annotations as Flow;
+use Doctrine\DBAL\Connection;
+use Doctrine\ORM\EntityManagerInterface;
+
+/**
+ * Connection factory for Doctrine connection class
+ *
+ * @Flow\Scope("singleton")
+ */
+class ConnectionFactory
+{
+    /**
+     * @Flow\Inject
+     * @var EntityManagerInterface
+     */
+    protected $entityManager;
+
+    /**
+     * Factory method which creates an Connection from
+     * the injected EntityManager.
+     *
+     * This ensure, that configured database and similar settings
+     * from Neos.Flow.persistence.doctrine is registered as expected
+     *
+     * @return Connection
+     */
+    public function create(): Connection
+    {
+        return $this->entityManager->getConnection();
+    }
+}
--- a/Neos.Flow/Configuration/Objects.yaml
+++ b/Neos.Flow/Configuration/Objects.yaml
@@ -367,6 +367,10 @@ Doctrine\ORM\EntityManagerInterface:
   scope: singleton
   factoryObjectName: Neos\Flow\Persistence\Doctrine\EntityManagerFactory
 
+Doctrine\DBAL\Connection:
+  scope: singleton
+  factoryObjectName: Neos\Flow\Persistence\Doctrine\ConnectionFactory
+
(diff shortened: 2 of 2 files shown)


Neos.Flow/Configuration/Objects.yaml
```diff
@@ -367,6 +367,10 @@ Doctrine\ORM\EntityManagerInterface:
   scope: singleton
   factoryObjectName: Neos\Flow\Persistence\Doctrine\EntityManagerFactory
 
+Doctrine\DBAL\Connection:
+  scope: singleton
```
bwaidelich, line 371: might that pose a problem if you work with multiple different connections?

sorenmalling: The keyword being "if" :-D So, should the general injection be the Flow configured and then we have a "createWithOptions" static method (which can then be used in entityMangerFactory?) if you want a diffrent connection, following Doctrines configuraton format?

bwaidelich:
Great initiative,

we needed something similar for the `Neos.EventSourcing` package: https://github.com/neos/Neos.EventSourcing/blob/master/Classes/EventStore/Storage/Doctrine/Factory/ConnectionFactory.php

> Should it be registred in Objects.yaml that it is used when injection the Connection directly?

Absolutely, and you already did that it seems :)

> Should it have a options argument, so you can override configuration?

No, it would be a different connection in that case (i.e. require a custom factory)

BTW: I created a feature issue and linked your PR

Neos.Flow/Configuration/Objects.yaml
bwaidelich: No, that's not what I meant.
I just wonder if this will break if you use a custom factory for your own connection instances

sorenmalling: Arh, okay :) But then you would set it as a property in your own class, and register your own factory. so the question here would be, which one (this registration or your own factory) will take care of creatig the instance?

sorenmalling:
> we needed something similar for the `Neos.EventSourcing` package: https://github.com/neos/Neos.EventSourcing/blob/master/Classes/EventStore/Storage/Doctrine/Factory/ConnectionFactory.php

The idea is copied from exactly that package :D And I've used it myself a number of time and @DavidSporer showed a example in Slack the other day.

Neos.Flow/Configuration/Objects.yaml
bwaidelich: i.e. will this work:
```php
class SomeClass {

  /**
   * @Flow\Inject
   * @var Connection
   */
  protected $flowConnection;
 
  /**
   * @Flow\Inject
   * @var Connection
   */
  protected $customConnection;
}
```

..if you configure a custom factory for the 2nd connection like

```yaml
Your\Package\SomeClass:
  properties:
    customConnection:
      object:
        factoryObjectName: Your\Package\SomeCustomConnectionFactory
```

or will the two instances be the same since its a singleton now?

sorenmalling: I will test that 👍

sorenmalling:
@bwaidelich i created a test package to be found here

https://github.com/sorenmalling/Vendor.TestConnectionFactory

result from running `./flow test:test` is that my customConnection uses the CustomConnectionFactory. This is concluded from looking at the params from the injected Connections configuration (see screenshots)

Screenshot 2020-10-14 at 12 50 39

Screenshot 2020-10-14 at 12 50 51

Neos.Flow/Configuration/Objects.yaml
sorenmalling: Tested with positive result - see comment here

https://github.com/neos/flow-development-collection/pull/2169#issuecomment-708324572

bwaidelich:
Awsome thanks, for checking!

Neos.Flow/Classes/Persistence/Doctrine/ConnectionFactory.php
```diff
@@ -0,0 +1,44 @@
...
+use Neos\Flow\Annotations as Flow;
+use Doctrine\DBAL\Connection;
+use Doctrine\ORM\EntityManagerInterface;
+
+/**
+ * Connection factory for Doctrine connection class
+ *
+ * @Flow\Scope("singleton")
+ */
+class ConnectionFactory
```
bwaidelich, line 23:
```suggestion
final class ConnectionFactory
```

bwaidelich (approved):
Just a mini suggestion (making classes final that aren't meant to be extended).
Apart from that: +1 (by reading)!

albe (approved):
I like it. Let's make that class final and get this in :)

sorenmalling:
> I like it. Let's make that class final and get this in :)

pushed and awaiting test results

DavidSporer:
Looks good for me as well :-)
We use this quite often and it will make things cleaner.

Issue #2141, opened 2020-09-21
Allow RoutePart handlers to access Route Parameters when resolving routes


bwaidelich:
### Description

With #1120 (precisely #1126) RoutePart handlers can access Route Parameters that can be specified in HTTP components in order to make routing dependent on some user defined context.
This can for example be used in order to match routes based on the current request host.

Currently the Route Parameters are only available during *matching* of incoming requests, but for proper cross domain linking the behavior is also dependent on the parameters when *resolving* a route, i.e. to determine whether a relative or absolute URL has to be generated.

For this we should extend the [ParameterAwareRoutePartInterface)
](https://github.com/neos/flow-development-collection/blob/c805e3141691b1b2e2c8541648ec473ff5e7f6e2/Neos.Flow/Classes/Mvc/Routing/ParameterAwareRoutePartInterface.php) by a corresponding `resolveWithParameters()` method.

PR #2173, opened 2020-10-13
!!! FEATURE: Allow RoutePart handlers to access Route Parameters when resolving routes

bwaidelich:
This feature allows route part handlers to access any Route Parameters
that has been set for the current request.
This will make it possible to implement cross-domain linking for example
with relative/absolute URLs depending on the current host.

This is a potentially breaking change because it extends the `ParameterAwareRoutePartInterface`
by a new method `resolveWithParameters`.
This means that custom RoutePartHandlers that implement this interface directly have to be
adjusted. The easiest way to adjust an existing handler is to implement this method as follows:

```php
final public function resolveWithParameters(array &$routeValues, RouteParameters $_)
{
    return $this->resolve($routeValues);
}
```

...basically ignoring the parameters.

Route Part handlers extending `DynamicRoutePart` don't need to be adjusted!

This also changes the (non-api) `Route::resolves()` method that now expects an instance of
`ResolveContext` instead of an array with the "routeValues".

neos/neos-development-collection#3020
Resolves: #2141


skurfuerst (commented):
@bwaidelich looks good by reading :+1:

I am lacking some context for what specific feature etc. we are needing this; but from what I've read, this makes sense :-) I cannot assess all the implications, which is why I don't formally "approve" this change in GitHub sense. We can have a quick phone call (I am back tuesday) if you  like!

All the best,
Sebastian

bwaidelich:
@skurfuerst great, thanks for your feedback. I'm more than happy to discuss this with you when you're back. It is mainly the preparation to allow us to move the whole domain resolving logic from the LinkingService into the Routing handler

skurfuerst:
@bwaidelich ah I see - that totally makes sense :) :)

Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
```diff
@@ -50,18 +50,25 @@ final class ResolveContext
...
+
     /**
      * @param UriInterface $baseUri The base URI, retrieved from the current request URI or from configuration, if specified. Required to fill in parts of the result when resolving absolute URIs
      * @param array $routeValues Route values to build the URI, for example ['@action' => 'index', 'someArgument' => 'foo', ...]
      * @param bool $forceAbsoluteUri Whether or not an absolute URI is to be returned
      * @param string $uriPathPrefix A prefix to be prepended to any resolved URI. Not allowed to start with "/".
+     * @param RouteParameters $parameters
      */
-    public function __construct(UriInterface $baseUri, array $routeValues, bool $forceAbsoluteUri, string $uriPathPrefix = '')
+    public function __construct(UriInterface $baseUri, array $routeValues, bool $forceAbsoluteUri, string $uriPathPrefix, RouteParameters $parameters)
```
albe, line 65: I somehow think the last three arguments should be optional with sane defaults (absoluteUri false, uriPathPrefix '' and empty parameters)

albe (approved):
Generally looking good by reading.

Neos.Flow/Classes/Http/ServerRequestAttributes.php
```diff
@@ -18,7 +18,12 @@ final class ServerRequestAttributes
     public const TRUSTED_PROXY = 'fromTrustedProxy';
 
     /**
-     * PSR-7 Attribute containing routing results as array to be merged in any merged arguments.
+     * PSR-7 Attribute containing routing parameters, so that they can be accessed when building URIs
+     */
+    public const ROUTING_PARAMETERS = 'routingParameters';
```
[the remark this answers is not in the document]
bwaidelich: yep, let's get your PR in first and I'll rebase this one

Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php
bwaidelich: Mathias Vereas once criticised the many optional arguments in our code and I didn't really agree at first. But in the meantime I learned to hate them too. In this case the invoking side should make an informed decision about whether to force absolute URIs etc. In practice this code will probably never be invoked in user land code anyways

Neos.Flow/Classes/Http/ServerRequestAttributes.php
```diff
@@ -18,7 +18,7 @@ final class ServerRequestAttributes
     public const TRUSTED_PROXY = 'fromTrustedProxy';
 
     /**
-     * PSR-7 Attribute containing routing results as array to be merged in any merged arguments.
+     * PSR-7 Attribute containing routing results as array to be merged in any merged arguments
```
bwaidelich, line 21:
```suggestion
     * PSR-7 Attribute containing routing results as array to be merged in any merged arguments.
```

bwaidelich:
Rebased to master, will merge after CI is done

Neos.Flow/Tests/Unit/Mvc/Routing/RouteTest.php
```diff
@@ -16,16 +16,19 @@
 use Neos\Flow\Mvc\Exception\InvalidRoutePartValueException;
 use Neos\Flow\Mvc\Exception\InvalidRouteSetupException;
 use Neos\Flow\Mvc\Exception\InvalidUriPatternException;
+use Neos\Flow\Mvc\Routing;
+use Neos\Flow\Mvc\Routing\Dto\RouteContext;
+use Neos\Flow\Mvc\Routing\Dto\RouteParameters;
```
albe, line 21:
```suggestion
```
Those are already imported below.

albe:
Green 🎉

bwaidelich:
Ah, that was too quick (or I was too slow rather) I found some more nitpicks I wanted to fix. But no problem, I'll create a new bugfix :)

bwaidelich:
=> #2224

PR #2174, opened 2020-10-13
FEATURE: Add StaticResource EEL Helper

mficzel:
Add a helper to read the uri and content of static (package) resources as this
previously often tedious. The primary usecase is creating resource urls in afx.

StaticResource.uri (packageKey, pathAndFilename, localize)
- (string) packageKey
- (string) pathAndFilename
- (boolean, optional) localize = false

StaticResource.content (packageKey, pathAndFilename, localize)
- (string) packageKey
- (string) pathAndFilename
- (boolean, optional) localize = false

example use in afx:
```
  <!-- create static resource uri -->
  <link rel="stylesheet" href={StaticResource.uri('Neos.Demo', 'Public/Styles/Main.css')} media="all" />

  <!-- get static resource content -->
  <style>{StaticResource.content('Neos.Demo', 'Public/Styles/Main.css')}</style>
```
Resolves: #2175

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
+++ b/Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
@@ -1,0 +1,78 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Flow\ResourceManagement\EelHelper;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Annotations as Flow;
+use Neos\Eel\ProtectedContextAwareInterface;
+use Neos\Flow\ResourceManagement\ResourceManager;
+use Neos\Flow\I18n\Service as I18nService;
+
+class StaticResourceHelper implements ProtectedContextAwareInterface
+{
+    /**
+     * @Flow\Inject
+     * @var ResourceManager
+     */
+    protected $resourceManager;
+
+    /**
+     * @Flow\Inject
+     * @var I18nService
+     */
+    protected $i18nService;
+
+    /**
+     * Get the public uri of a package resource
+     *
+     * @param string $packageKey
+     * @param string $pathAndFilename
+     * @param bool $localize
+     * @return string
+     */
+    public function uri(string $packageKey, string $pathAndFilename, bool $localize = false): string
+    {
+        $resourcePath = sprintf('resource://%s/%s', $packageKey, $pathAndFilename);
+        if ($localize === true) {
+            $resourcePath = $this->i18nService->getLocalizedFilename($resourcePath);
+        }
+        return $this->resourceManager->getPublicPackageResourceUriByPath($resourcePath);
+    }
+
+    /**
+     * Get the content of a package resource
+     *
+     * @param string $packageKey
+     * @param string $pathAndFilename
+     * @param bool $localize
+     * @return string
+     */
+    public function content(string $packageKey, string $pathAndFilename, bool $localize = false): string
+    {
+        $resourcePath = sprintf('resource://%s/%s', $packageKey, $pathAndFilename);
+        if ($localize === true) {
+            $resourcePath = $this->i18nService->getLocalizedFilename($resourcePath);
+        }
+        $content = file_get_contents($resourcePath);
+        return $content ?: '';
+    }
+
+    /**
+     * @param string $methodName
+     * @return bool
+     */
+    public function allowsCallOfMethod($methodName)
+    {
+        return true;
+    }
+}
--- a/Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
+++ b/Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
@@ -35,9 +35,9 @@ class StaticResourceHelper implements ProtectedContextAwareInterface
     /**
      * Get the public uri of a package resource
      *
-     * @param string $packageKey
-     * @param string $pathAndFilename
-     * @param bool $localize
+     * @param string $packageKey Package key where the resource is from.
+     * @param string $pathAndFilename The path and filename of the resource. Has to start with "Public/..." as private resources do not have a uri.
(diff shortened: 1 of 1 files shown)


mficzel:
For the review, this is added to the flow package as it required the resource manager and the i18n service. This is quite similar to the translation helper.

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
```diff
@@ -0,0 +1,80 @@
...
+     * @param string $pathAndFilename
+     * @param bool $localize
+     * @return string
+     */
+    public function uri(string $packageKey, string $pathAndFilename, bool $localize = false): string
+    {
+        $resourcePath = sprintf('resource://%s/%s', $packageKey, $pathAndFilename);
+        if ($localize === true) {
+            $localizedResourcePathData = $this->i18nService->getLocalizedFilename($resourcePath);
+            $resourcePath = $localizedResourcePathData[0] ?? $resourcePath;
```
sorenmalling, line 48: Is this intended behavior, to fallback to original? If you ask for a localized version, what do you expect in return? Do we have similar use case of the $localize, where we do the same, so we follow the same pattern everywhere?

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
```diff
@@ -0,0 +1,80 @@
...
+     * Get the content of a package resource
+     *
+     * @param string $packageKey
+     * @param string $pathAndFilename
+     * @param bool $localize
+     * @return string
+     */
+    public function content(string $packageKey, string $pathAndFilename, bool $localize = false): string
+    {
+        $resourcePath = sprintf('resource://%s/%s', $packageKey, $pathAndFilename);
```
sorenmalling, line 63: Please extract the code (similar as in the uri method) for getting the resourcePath to a method, to avoid duplicate :)

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
mficzel: if a no better version is abvailable the original resource path is used. This is pretty much what the View Helpers and the ResourceUri Fusion Object are doing.

sorenmalling: Perfect 👍

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
mficzel: OK

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
mficzel: Use case is when you have a special logo for a locale or a special css for languages with right to left fonts

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
sorenmalling: Suppose to include
```
        if ($localize === true) {
            $localizedResourcePathData = $this->i18nService->getLocalizedFilename($resourcePath);
            $resourcePath = $localizedResourcePathData[0] ?? $resourcePath;
    }
```

sorenmalling:
@bwaidelich @albe A quick vote - should/could this go into the Neos.Eel package instead? Something in me, says that it's a better location for this helper. Neos.Eel already depends on Flow (composer.json require section)

mficzel:
The reasoning why this is added to flow was to avoid circular dependencies as much as possible. The resource manager and translation service are flow specific. Eel currently relies on flow but should become more standalone in future if possible.

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
mficzel: Ahhh sorry, now this is included aswell. Makes totally sense.

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    protected $i18nService;
+
+    /**
+     * Get the public uri of a package resource
+     *
+     * @param string $packageKey
+     * @param string $pathAndFilename
+     * @param bool $localize
+     * @return string
```
bwaidelich, line 41: Since this source is used to generate the [Eel helper reference](https://neos.readthedocs.io/en/stable/References/EelHelpersReference.html) we should add descriptions for the params. Especially the `$localize` part that is a bit .. special :)

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
```diff
@@ -0,0 +1,89 @@
...
+            $resourcePath = $localizedResourcePathData[0] ?? $resourcePath;
+        }
+        return $resourcePath;
+    }
+
+    /**
+     * @param string $methodName
+     * @return bool
+     */
+    public function allowsCallOfMethod($methodName)
```
bwaidelich, line 85:
```suggestion
    public function allowsCallOfMethod($methodName): bool
```

bwaidelich (changes requested):
Nice helper, thanks!
Just two little comments. Apart from that a little unit test would be rad of course..

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
mficzel: Added some documentation

Neos.Flow/Classes/ResourceManagement/EelHelper/StaticResourceHelper.php
mficzel: Adjusted

mficzel:
@bwaidelich i did not add a test since this only exposes existing and tested functionality to eel. The real code is in getLocalizedResourcePath but even that is basically trivial.

bwaidelich (approved):
Fair enough, makes sense! Thanks for adding the doc comments

+1 by reading

daniellienert (approved):

PR #2180, opened 2020-10-14
BUGFIX: ActionResponse contains negotiated media type as content-type

albe:
This sets the negotiated media type from the Controller in the `ActionResponse` if nothing was set in the action.

Follow-up to #2005


bwaidelich:
Thanks for taking care. Makes sense by reading, but IMO this needs a test that simulates the regression

bwaidelich:
..btw: I'll gladly help with the test if you feel like a little pair programming session!

albe:
Right. Also this is somehow missing the test verifying the new behaviour that I had in the previous PR :/ meh, must have messed up the rebase

bwaidelich:
I tripped upon this bug when I tried to demonstrate the "minify html middleware" the other day, because that only works on a proper Content-Type header

albe:
Yes, I noticed and that's (one of) the reason I came back to this now :)

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -177,6 +185,7 @@ public function addFlashMessage($messageBody, $messageTitle = '', $severity = Er
      * @param string $packageKey Key of the package containing the controller to forward to. May also contain the sub package, concatenated with backslash (Vendor.Foo\Bar\Baz). If not specified, the current package is assumed.
      * @param array $arguments Arguments to pass to the target action
      * @return void
+     * @psalm-return never-returns
```
albe, line 188:
```suggestion
     * @psalm-return never-returns
```

Neos.Flow/Tests/Unit/Mvc/Controller/ActionControllerTest.php
```diff
@@ -210,6 +210,8 @@ public function processRequestInjectsControllerContextToView()
         $this->mockRequest->expects(self::any())->method('getHttpRequest')->will(self::returnValue($mockHttpRequest));
 
         $mockResponse = new Mvc\ActionResponse;
+        $mockResponse->setContentType('text/html');
```
albe, line 213: Avoids that ' setContentType()' is called

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -196,7 +196,7 @@ public function getStatusCode(): int
      */
     public function getContentType(): string
     {
-        return $this->contentType;
+        return $this->contentType ?? '';
```
albe, line 199: That's just semantic correction, since the return type hint was already in place and would have led to an fatal error if contentType was unset.

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -212,6 +212,10 @@ public function processRequest(ActionRequest $request, ActionResponse $response)
         }
 
         $this->callActionMethod();
+
+        if ($this->response->getContentType() === '') {
```
albe, line 216: This now allows any view or render method to set the content type to fix https://github.com/neos/flow-development-collection/pull/2005#issuecomment-707625891

albe:
Should be good now

Neos.Flow/Classes/Mvc/ActionResponse.php
bwaidelich: In 7.0 the return type hint has been changed (https://github.com/neos/flow-development-collection/blob/32149adc310623c9e43e6849128d8248a6e536cc/Neos.Flow/Classes/Mvc/ActionResponse.php#L251) – is it a good idea to do "" vs NULL here?

bwaidelich: I would suggest to omit this unrelated(?) change from this PR unless we're 100% certain about it

Neos.Flow/Classes/Mvc/Controller/ActionController.php
bwaidelich, line 216: This should cater for the `null` case I suppose:
```suggestion
        if (empty($this->response->getContentType()) {
```
even nicer would be
```php
if (!$this->response->hasContentType()) {
```
so that we can move the empty-logic into the response

bwaidelich (commented):
Looks good and I tested this with Neos 7.
Just a little remark/question regarding the additional change of `ActionResponse::getContentType()`

Neos.Flow/Classes/Mvc/Controller/ActionController.php
albe: `empty()` will not suffice, because if `$this->contentType` is `null`, calling the method will throw a fatal TypeError, as it expects to return a `string` (and even though we don't have strict_types on this file yet, that doesn't help for return types).

sorenmalling:
@albe @bwaidelich ping :) After talking about debugbar middleware on Slack, this was linked. Where did we end up on this topic? In a middleware, there is no content-type in the response, which stops (not all, probably) the debugbar from rendering

albe:
I guess we need to agree on how to proceed regarding https://github.com/neos/flow-development-collection/pull/2180#discussion_r550798576 - I still think the `?? ''` null coalescing is the right thing at that place and we should probably change 7.0 again - but that might be something discuss-worthy.

Neos.Flow/Classes/Mvc/Controller/ActionController.php
albe, line 216:
```suggestion
        if (!$this->response->hasContentType()) {
```

Neos.Flow/Classes/Mvc/ActionResponse.php
albe: See https://github.com/neos/flow-development-collection/pull/2180/files#r550798576 - this is a semantic/typehint error and psalm complains about it. So we need to deal with this one way or another (adjust return type hint or use null coalescing). I still see no value in the nullability of this getter though. It would just be a technical necessity due to the missing property initialization. For all means and purposes '' is equivalent to `null`, but as of the addition of the `hasContentType()` check method that prevents calling `getContentType()` when it would be null.

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -191,12 +191,17 @@ public function getStatusCode(): int
         return $this->statusCode ?? 200;
     }
 
+    public function hasContentType(): bool
```
albe, line 194: We should start removing useless docblocks that only repeat the existing PHP typehints

Neos.Flow/Classes/Mvc/ActionResponse.php
albe:
```suggestion
        return $this->contentType;
```
Removing this for now as of the `hasContentType()` check and moving this to an own PR where we can discuss if this getter should be nullable or not.

Neos.Flow/Tests/Unit/Mvc/Controller/ActionControllerTest.php
```diff
@@ -210,6 +210,8 @@ public function processRequestInjectsControllerContextToView()
...
+        $this->inject($this->actionController, 'response', $mockResponse);
```
albe, line 214:
```suggestion
```
Should no longer be needed with the `hasContentType()` check

Neos.Flow/Tests/Unit/Mvc/Controller/ActionControllerTest.php
```diff
@@ -238,6 +240,8 @@ public function processRequestInjectsSettingsToView()
...
```
albe, line 244:
```suggestion
```
See above

Neos.Flow/Tests/Unit/Mvc/Controller/ActionControllerTest.php
```diff
@@ -246,6 +250,33 @@ public function processRequestInjectsSettingsToView()
...
+
+        $mockMvcPropertyMappingConfigurationService = $this->createMock(Mvc\Controller\MvcPropertyMappingConfigurationService::class);
+        $this->inject($this->actionController, 'mvcPropertyMappingConfigurationService', $mockMvcPropertyMappingConfigurationService);
+
+        $mockHttpRequest = $this->getMockBuilder(ServerRequestInterface::class)->disableOriginalConstructor()->getMock();
+        $this->mockRequest->expects(self::any())->method('getHttpRequest')->willReturn($mockHttpRequest);
+
+        $mockResponse = new Mvc\ActionResponse;
+        $this->inject($this->actionController, 'response', $mockResponse);
+        $this->inject($this->actionController, 'negotiatedMediaType', 'application/json');
```
albe, line 274: Meh, this is a really bad test... it should inject supportedMediaTypes into the controller and not test this implementation detail

Neos.Flow/Tests/Unit/Mvc/Controller/ActionControllerTest.php
```diff
@@ -246,6 +248,41 @@ public function processRequestInjectsSettingsToView()
...
+
+        $mockHttpRequest = $this->getMockBuilder(ServerRequestInterface::class)->disableOriginalConstructor()->getMock();
+        $mockHttpRequest->method('getHeaderLine')->with('Accept')->willReturn($acceptHeader);
+        $this->mockRequest->method('getHttpRequest')->willReturn($mockHttpRequest);
+
+        $mockResponse = new Mvc\ActionResponse;
+        $this->inject($this->actionController, 'supportedMediaTypes', $supportedMediaTypes);
+
+        $this->actionController->processRequest($this->mockRequest, $mockResponse);
+        self::assertSame($expected, $mockResponse->getContentType());
```
albe, line 283: FYI @bwaidelich - we can't add a "does not set" case (supported != accepted) here, because `getContentType()` would try to return null and blow up - let's move the further discussion about the broken getter to https://github.com/neos/flow-development-collection/pull/2458

albe:
Looks good again from my side now. Let's move the discussion about the getContentType return value to #2458

bwaidelich (approved):
Looks great (by reading)
Thanks for taking care and for extracting the new PR

Neos.Flow/Tests/Unit/Mvc/Controller/ActionControllerTest.php
```diff
@@ -246,6 +248,66 @@ public function processRequestInjectsSettingsToView()
...
+
+        $mockMvcPropertyMappingConfigurationService = $this->createMock(Mvc\Controller\MvcPropertyMappingConfigurationService::class);
+        $this->inject($this->actionController, 'mvcPropertyMappingConfigurationService', $mockMvcPropertyMappingConfigurationService);
+
+        $mockHttpRequest = $this->getMockBuilder(ServerRequestInterface::class)->disableOriginalConstructor()->getMock();
+        $mockHttpRequest->method('getHeaderLine')->with('Accept')->willReturn('application/xml');
+        $this->mockRequest->method('getHttpRequest')->willReturn($mockHttpRequest);
+
+        $mockResponse = new Mvc\ActionResponse;
+        $mockResponse->setContentType('application/json');
```
albe, line 304: This test should show the actual regression that got introduced with #2005 
It also shows that the action can completely ignore the client accepted media-types - maybe we could at least add a feature at some point that checks the content-type of the ActionResponse against the accept header and logs a warning when they mismatch.

sorenmalling (approved):

PR #2181, opened 2020-10-14
WIP: Remove deprecated code

### LINKED ISSUE #2172: Remove classes marked as deprecated and can be removed by 7.0
In a large number of cases (Generic persitence for ex.) it's marked with

 `deprecated since Flow 6.0`

And in some, we have to check history, if no [version number] is mentioned :) This can require that we go through git history for that single file when in doubt, so we do not remove wrong stuff by accident :) 

Use `grep -Rin "deprecated" Neos.Flow/Classes` from the Framework folder to find for the `Neos.Flow` package, and so on with others.

sorenmalling:
- [ ] `Headers` are used in Browser class - how to replace?
 - [ ] The HTTP Component stack is not completely replaced by psr-15 - this makes it difficult to remove them before the other is rebased (vice/versa - one have to be done first) [#2221]
 - [ ] With https://github.com/neos/flow-development-collection/pull/2024  we renamed and deprecated a number of methods and configuration name - are they being removed for 7.0? Can we write a code migration for it 🤩 

Method used to find deprecated content (from root of Flow)

`cd Packages/Framework; grep -Rin "deprecated" */Classes | less`

Resolves #2172


albe:
Not sure about the `Headers` class - we deprecated it in the PSR-7 change, but maybe we should use it internally still, as it's a clean abstraction of the HTTP headers concept and tested against the RFC spec. We could use it in a middleware that allows setting headers via annotations (thinking of `Cache-Control` primarily) or similar

bwaidelich:
Hi @sorenmalling,

sorry again for re-creating basically the same with #2262 – No idea how I could miss this one..
I compared the two and they are really mostly the same, except:
* "your" PR removes support for the old `ignoredTags` configuration scheme. That's fair since that's deprecated for a while now. But since it is no PHP code, the IDE wouldn't have complained about it. I would suggest that we keep that since it doesn't cost us and dropping support might lead to subtle bugs that are hard to debug. We could instead detect the old syntax and throw an exception, but I doubt that it's worth the effort
* "my" PR removes the unsafe `SaltedMd5HashingStrategy` and tweaks some `@deprecated` annotations.

Since #2262 is already based on the current master, is it OK for you if we close this one?

sorenmalling:
> Since #2262 is already based on the current master, is it OK for you if we close this one?

Sure 👍

bwaidelich:
Closing in favor of #2262 as discussed

Issue #2183, opened 2020-10-15
Flow Configuration is unprocessed during `shutdownObject()` lifecycle method


nlx-lars:
### Description

If a setting contains variables (either environment variables with `%env:NAME%` or constants with `%NAME%`), those won't be replaced with their value during the `shutdownObject()` lifecycle method. Instead, the setting will contain the unprocessed `%env:NAME%` string, possibly leading to broken cleanup behavior.

=> This seems to only happen in the first request after a `./flow flow:cache:flush && ./flow cache:warmup`. Probably because `$cacheNeedsUpdate` is false in subsequent requests?

### Steps to Reproduce

1. Define a setting with an environment variable
```yaml
Foo:
  Bar:
    baz: '%env:SHUTDOWN_TEST%'
```
2. Create a class that uses this setting in it's `shutdownObject()` method
```php
<?php

use Neos\Flow\Annotations as Flow;

/**
 * @Flow\Scope("singleton")
 */
class FooService
{

    /**
     * @Flow\InjectConfiguration(package="Foo.Bar", path="baz")
     * @var string
     */
    protected $baz;
    
    public function initializeObject(): void
    {
        // $this->baz will contain the value of the environment variable "SHUTDOWN_TEST"
    }
    
    public function shutdownObject(): void
    {
        // $this->baz will contain the unprocessed string "%env:SHUTDOWN_TEST%"
    }
    
}
```

#### Expected behavior

The property `$baz` should contain the value of the `SHUTDOWN_TEST` environment variable.

#### Actual behavior

The property `$baz` should contains the unprocessed value (`%env:SHUTDOWN_TEST%`) of the configuration string.

### Affected Versions

Neos: NA

Flow: 5.3

### Further information
This is caused by the order in which the ConfigurationManager and ObjectManager are connected to the `bootstrapShuttingDown` Signal, see [Flow's Package.php](https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Classes/Package.php#L107).
First, `ConfigurationManager::shutdown()` is called which then saves the current configuration with a call to [$this-> saveConfigurationCache()](https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Classes/Configuration/ConfigurationManager.php#L550). That method (or the call to 
[...]

PR #2184, opened 2020-10-16
BUGFIX: Register ObjectManager::shutdown() before ConfigurationManage…

paxuclus:
The `ObjectManager::shutdown()` method will trigger the `shutdownObject()` lifecycle method of
all registered objects. If the `ConfigurationManager` is shutdown beforehand, the configuration
may be in an uninitialized state.

This change swaps the order in which `ConfigurationManager::shutdown()` and `…::shutdown()`
are registered in the Flow's `Package.php` to make sure that the `ConfigurationManager` is still
initialized while the `ObjectManager` is being shut down.

Resolves: #2183

### DIFF
--- a/Neos.Flow/Classes/Package.php
+++ b/Neos.Flow/Classes/Package.php
@@ -104,8 +104,9 @@ class Package extends BasePackage
             $dispatcher->connect(Monitor\FileMonitor::class, 'filesHaveChanged', Cache\CacheManager::class, 'flushSystemCachesByChangedFiles');
         }
 
-        $dispatcher->connect(Core\Bootstrap::class, 'bootstrapShuttingDown', Configuration\ConfigurationManager::class, 'shutdown');
+        // The ObjectManager has to be shutdown before the ConfigurationManager, see https://github.com/neos/flow-development-collection/issues/2183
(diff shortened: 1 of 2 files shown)


sorenmalling:
> I am unsure how to test this. I guess by adding a functional test that confirms that the ObjectManager is called first?

Can you create a test that does that? It will make sense in terms of this change we make 👍

nlx-lars:
@sorenmalling sure! 🙂

sorenmalling (commented):
Looks good from reading - great catch! It's not a topic I know much about I'd have any conflicts on compiling we similar, so I rely on the Travis test results

albe:
Yes, some way to fixate this order would be good, since it is not obvious it needs to be like this. A test that just checks the order is okay. At the very minimum a docblock at that place that states that the ObjectManager needs to shutdown *before* the ConfigurationManager would be helpful

nlx-lars:
@sorenmalling & @albe I added a functional test to check the order in which these two slots are registered in the Dispatcher.

albe (approved):
Awesome! Thanks a lot :)

kdambekalns (approved):

Issue #1127, opened 2017-11-29
Default orphan removal value for "non aggregate root"-properties is broken


dwinkelbauer:
### Description

After fixing the issue #1079, orphan removal is per default false for properties from classes which are no aggregate roots. 
The problem is that the default value for orphan removal in the ManyToMany, OneToOne and OneToMany annotations is "false". (See [here](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Mapping/OneToMany.php#L55))
Now as [this](https://github.com/neos/flow/blob/17f9801c02e94844895e36e5d567947adef93eaa/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php#L633) condition does now check, if the value is null and not only if the value is true, orphan removal is always set to false per default. So the following [else if](https://github.com/neos/flow/blob/17f9801c02e94844895e36e5d567947adef93eaa/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php#L635) for checking if the property is no aggregate root is never evaluated.

This is a major problem, because this minor update can completely change the behavior of an existing application.
### Steps to Reproduce

1. Create a class A with a repository and a class B without a repository
2. Add a property to A from class B e.q. as an OneToMany relation
3. Before #1079 this association had orphanRemovel=true, now it uses orphanRemovel=false per default

#### Expected behavior

orphanRemovel=true 

#### Actual behavior

orphanRemovel=false 

### Affected Versions

Flow: since 4.2.4 and 3.3.20

There is of course the workaround to set orphanRemovel=true manually, but is this the intended approach?

albe:
Thanks a lot for bringing this up! It looks indeed like a regression and buggy behaviour.

The question is, what is the expected behaviour? Since orphanRemoval has a default value, there is no way to distinguish the case if this value was implicitly or explicitly set to `false`. I'd expect that if nothing is explicitly set, the fallback rule for aggregate roots is checked and otherwise, the value defaulted to `false`. If however the value is explicity set to `false`, then that should override the aggregate root fallback.

So I only see one solution: remove the default value from the annotation, then add another `else` clause that sets the orphanRemoval to `false`

domin1101:
I also thought of removing the default, but as you already said this is not possible.

I think there are only a few other solutions
[...]

PR #2185, opened 2020-10-16
!!! BUGFIX: Prevent collection setters from overwriting collections

albe:
Overwriting a collection will lead to all related entities being deleted as of Doctrine ORM 2.6+, if the relation has orphanRemoval, as do all *ToMany relations to non-aggregate roots in Flow.

This is marked breaking, because if you have setters for collections in your domain models that simply replace the collection property, this will now throw an exception (to avoid leading to possible data loss).
You should replace your collection setters with code like this:

```php
/**
 * @ORM\OneToMany(mappedBy="parent")
 * @var Collection<MyEntity>
 */
protected $myCollection;

public function __constructor()
{
    $this->myCollection = new ArrayCollection();
}

public function setMyCollection(Collection $newCollection)
{
    foreach ($this->myCollection as $child) {
        if (!$newCollection->contains($child)) {
            $this->myCollection->removeElement($child);
        }
    }
    foreach ($newCollection as $child) {
        if (!$this->myCollection->contains($child)) {
            $this->myCollection->add($child);
        }
    }
}
```

i.e. only initialize the collection in your constructor and never replace it in a method, only add/removeElement to it.

Related to #2159

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
@@ -222,8 +222,8 @@ class PersistentObjectConverter extends ObjectConverter
                 );
                 throw new InvalidTargetException($exceptionMessage, 1297935345);
             }
-            if ($isCollectionType && $originalValue !== ObjectAccess::getProperty($object, $propertyName)) {
-                throw new TypeConverterException(sprintf('You overwrote the collection property "%s" of type "%s". This will lead to data loss. Read more about it and how to prevent this here: ', $propertyName, $targetType), 1602324196);
+            if ($isCollectionType && $originalValue !== ObjectAccess::getProperty($object, $propertyName, true)) {
+                throw new TypeConverterException(sprintf('You overwrote the collection property "%s" of type "%s". This will lead to data loss. Read more about it and how to prevent this here: ADDLINKHERE', $propertyName, $targetType), 1602324196);
             }
         }
 
--- a/Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
@@ -208,9 +208,8 @@ class PersistentObjectConverter extends ObjectConverter
             }
             $propertyType = TypeHandling::getTypeForValue($propertyValue);
             $isCollectionType = TypeHandling::isCollectionType($propertyType);
-            //$isAggregateRelation = TODO: Check if element type is aggregate root
             if ($isCollectionType) {
-                $originalValue = ObjectAccess::getProperty($object, $propertyName);
+                $originalValue = ObjectAccess::getProperty($object, $propertyName, true);
             }
             $result = ObjectAccess::setProperty($object, $propertyName, $propertyValue);
             if ($result === false) {
--- a/Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
@@ -19,6 +19,7 @@ use Neos\Flow\Property\Exception\InvalidPropertyMappingConfigurationException;
 use Neos\Flow\Property\Exception\InvalidSourceException;
 use Neos\Flow\Property\Exception\InvalidTargetException;
 use Neos\Flow\Property\Exception\TargetNotFoundException;
+use Neos\Flow\Property\Exception\TypeConverterException;
 use Neos\Flow\Property\PropertyMappingConfigurationInterface;
 use Neos\Flow\Property\TypeConverter\Error\TargetNotFoundError;
 use Neos\Utility\ObjectAccess;
@@ -205,6 +206,12 @@ class PersistentObjectConverter extends ObjectConverter
                     throw new InvalidTargetException($exceptionMessage, 1421498771);
                 }
             }
+            $propertyType = TypeHandling::getTypeForValue($propertyValue);
+            $isCollectionType = TypeHandling::isCollectionType($propertyType);
+            //$isAggregateRelation = TODO: Check if element type is aggregate root
+            if ($isCollectionType) {
+                $originalValue = ObjectAccess::getProperty($object, $propertyName);
+            }
             $result = ObjectAccess::setProperty($object, $propertyName, $propertyValue);
             if ($result === false) {
                 $exceptionMessage = sprintf(
@@ -214,6 +221,9 @@ class PersistentObjectConverter extends ObjectConverter
                     $targetType
                 );
                 throw new InvalidTargetException($exceptionMessage, 1297935345);
+            }
+            if ($isCollectionType && $originalValue !== ObjectAccess::getProperty($object, $propertyName)) {
+                throw new TypeConverterException(sprintf('You overwrote the collection property "%s" of type "%s". This will lead to data loss. Read more about it and how to prevent this here: ', $propertyName, $targetType), 1602324196);
             }
         }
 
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
```diff
@@ -205,6 +206,12 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
                     throw new InvalidTargetException($exceptionMessage, 1421498771);
                 }
             }
+            $propertyType = TypeHandling::getTypeForValue($propertyValue);
+            $isCollectionType = TypeHandling::isCollectionType($propertyType);
+            //$isAggregateRelation = TODO: Check if element type is aggregate root
```
albe, line 211: We could try to detect AR-relations and avoid the exception here. But then again, if that relation has a manual `orphanRemoval`, the issue would be the same.
```suggestion
```

sorenmalling:
You changed the requirement of inflector to 1.3 due to php version - what version of Flow are we targeting with this change? Flow 5.3 and up says 7.2 as requirements (or suggestion, can't really grasp that) from this matrix https://docs.neos.io/cms/installation-development-setup/system-requirements#php-version-matrix

kdambekalns (approved):
I like that, better than losing data! But we should have a broad consensus here…

Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
```diff
@@ -215,6 +222,9 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
                 );
                 throw new InvalidTargetException($exceptionMessage, 1297935345);
             }
+            if ($isCollectionType && $originalValue !== ObjectAccess::getProperty($object, $propertyName)) {
+                throw new TypeConverterException(sprintf('You overwrote the collection property "%s" of type "%s". This will lead to data loss. Read more about it and how to prevent this here: ', $propertyName, $targetType), 1602324196);
```
albe, line 226: TODO before merge:
```suggestion
                throw new TypeConverterException(sprintf('You overwrote the collection property "%s" of type "%s". This will lead to data loss. Read more about it and how to prevent this here: ADDLINKHERE', $propertyName, $targetType), 1602324196);
```

kdambekalns:
Travis trigger…

kdambekalns:
@bwaidelich @albe This is the lkind of important breaking bugfix we should consider now - because if it's too bad for 5.3 it should definitely go into the next major, IMHO.

bwaidelich:
I can't really judge how invasive it is in practice and thus whether this should go into 5.3. I fully trust you guys here.
The CI doesn't though, functional tests fail atm :)

Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
```diff
@@ -205,6 +206,12 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
                     throw new InvalidTargetException($exceptionMessage, 1421498771);
                 }
             }
+            $propertyType = TypeHandling::getTypeForValue($propertyValue);
+            $isCollectionType = TypeHandling::isCollectionType($propertyType);
+            //$isAggregateRelation = TODO: Check if element type is aggregate root
+            if ($isCollectionType) {
+                $originalValue = ObjectAccess::getProperty($object, $propertyName);
```
albe, line 213:
```suggestion
                $originalValue = ObjectAccess::getProperty($object, $propertyName, true);
```

Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
```diff
@@ -215,6 +222,9 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
                 );
                 throw new InvalidTargetException($exceptionMessage, 1297935345);
             }
+            if ($isCollectionType && $originalValue !== ObjectAccess::getProperty($object, $propertyName)) {
```
albe, line 225:
```suggestion
            if ($isCollectionType && $originalValue !== ObjectAccess::getProperty($object, $propertyName, true)) {
```

albe:
Now the test exactly shows what will break and how. We do not follow those semantics of "don't replace a collection" in our own test entities.

Neos.Flow/Classes/Property/TypeConverter/PersistentObjectConverter.php
```diff
@@ -215,6 +221,9 @@ public function convertFrom($source, $targetType, array $convertedChildPropertie
                 );
                 throw new InvalidTargetException($exceptionMessage, 1297935345);
             }
+            if ($isCollectionType && $originalValue !== ObjectAccess::getProperty($object, $propertyName, true)) {
+                throw new TypeConverterException(sprintf('You overwrote the collection property "%s" of type "%s". This will lead to data loss. Read more about it and how to prevent this here: ADDLINKHERE', $propertyName, $targetType), 1602324196);
```
bwaidelich, line 225: Just wondering:
Would it make sense to only throw the exception depending on the actual ORM behavior? i.e. someone with an Flow 5.3/Neos 4.3 installation could safely update to new patchlevel versions without risking a changed behavior?

albe: What do you mean with "actual ORM behavior"? In any Version 2.6+, changing the instance of a collection (with orphanRemoval) will result in a `DELETE * FROM RelatedTable WHERE parentId = X` statement. So we kind of already do this.

bwaidelich: > In any Version 2.6+, changing the instance of a collection [...] will result in a DELETE

Yes, _that_ behavior. Or rather: skip the exception when ORM < 2.6 is used.

Doesn't that make sense?

bwaidelich: ah, we don't even _allow_  ORM < 2.6... sorry ignore me

bwaidelich, line 225:
```suggestion
                throw new TypeConverterException(sprintf('You overwrote the collection property "%s" of type "%s". This will lead to data loss. Read more about it and how to prevent this here: https://www.neos.io/go/flow-7-0-release-notes.html', $propertyName, $targetType), 1602324196);
```

bwaidelich: ^ This "short link" will redirect to https://flowframework.readthedocs.io/en/7.0/TheDefinitiveGuide/PartV/ReleaseNotes/index.html once the release is done

bwaidelich (commented):
I still keep changing my mind about this one..
It can be annoying for projects that apply a patch level release and already worked around the issue in a way that now leads to this exception..
OTOH it might prevent data loss for projects that upgraded from a 4.x version.

Without a question this should be fixed for `master`.. For 5.x I'm still not 100% sure

sorenmalling:
If in doubt, leave it out.
To be fair, it's a different release, made by a different team of RM, so we will be putting a responsibility on them (but still us as a whole) and more support for a version that they/we don't want. 
We can take full responsibility for it in 7.0 and future at least

albe:
So, should we retarget 7.0 and merge this in @bwaidelich ? I guess @sorenmalling is somewhat right. Plus, as I think I said - we could still backport this after the release iff we do find it sensible

bwaidelich:
> So, should we retarget 7.0 and merge this in @bwaidelich ?

I'd say: retarget for `master`. I hope that we are really close to tag the 7.0 release and I don't want to add any more potentially breaking changes to it

mficzel:
If we have to make this change at some point it better be now.

albe:
Yeah, shifting this onto 8.0 will be... well, a bit of a let-down if we do think we should help our developers not fall into this trap too easily. Technically, it's even our fault that they are effectively opt-into this behavior (via our `orphanRemoval` policy) - and there's no escape hatch (can't "unset" orphanRemoval on non-ARs... - see discussion in https://github.com/neos/flow-development-collection/issues/1127)

albe:
Okay, so after talking with @bwaidelich again and going through this, trying to add a test, we found that this would for example also break the new `Account` with the transient `Roles` property. So we'd need to check reflection here (😬) or somehow otherwise pass through non-entity collections. All in all this is too hot a change to rush now, so we agreed to proceed as follows:
- retarget onto 5.3 again
- add a test that clearly covers the ill-behaviour
- possibly put the exception throwing behind a feature flag and enable it by default in 7.1 only

Issue #2187, opened 2020-10-24
InstallerScripts is not compatible to composer 2.0


daniellienert:
Relates to #2025

ariesprodesign:
Same problem:
`> Neos\Flow\Composer\InstallerScripts::postPackageUpdateAndInstall
PHP Fatal error:  Uncaught Error: Call to undefined method Composer\DependencyResolver\Operation\UpdateOperation::getJobType() in D:\xampp\htdocs\honey-dev\Packages\Framework\Neos.Flow\Classes\Composer\InstallerScripts.php:74
`

bwaidelich:
This can be closed, no?

kdambekalns:
Fixed with latest release(s).

PR #2188, opened 2020-10-24
BUGFIX: Make InstallerScripts compatible to composer version 2.0+

daniellienert:
Instead of querying the removed method ::getJobType we now check
the class of the job instance like we do in the first lines
of the method.

Resolves: #2187

### DIFF
--- a/Neos.Flow/Classes/Composer/InstallerScripts.php
+++ b/Neos.Flow/Classes/Composer/InstallerScripts.php
@@ -71,7 +71,7 @@ class InstallerScripts
         if (!$operation instanceof InstallOperation && !$operation instanceof UpdateOperation) {
             throw new Exception\UnexpectedOperationException('Handling of operation with type "' . $operation->getJobType() . '" not supported', 1348750840);
         }
-        $package = ($operation->getJobType() === 'install') ? $operation->getPackage() : $operation->getTargetPackage();
+        $package = ($operation instanceof InstallOperation) ? $operation->getPackage() : $operation->getTargetPackage();
         $packageExtraConfig = $package->getExtra();
         $installPath = $event->getComposer()->getInstallationManager()->getInstallPath($package);
 
@@ -83,7 +83,7 @@ class InstallerScripts
             $evaluatedInstallerResources = true;
         }
 
-        if ($operation->getJobType() === 'install') {
+        if ($operation instanceof InstallOperation) {
             if (isset($packageExtraConfig['typo3/flow']['post-install'])) {
                 self::runPackageScripts($packageExtraConfig['typo3/flow']['post-install']);
             }
@@ -92,7 +92,7 @@ class InstallerScripts
             }
         }
 
-        if ($operation->getJobType() === 'update') {
+        if ($operation instanceof UpdateOperation) {
             if (isset($packageExtraConfig['typo3/flow']['post-update'])) {
                 self::runPackageScripts($packageExtraConfig['typo3/flow']['post-update']);
             }


daniellienert:
With the patch applied, `composer update` also works on 2.0
but the updateScript then fails with:
```
 [Neos\Flow\Package\Exception]
  The package with the name "symfony/polyfill-ctype" was found more than once, please make sure it exists only once. Paths "Libraries/composer/411865d6/symfony-polyfill-ctype-aed5969/" and "Libraries/symfony/polyfill-ctype/".
```
thrown in `PackageManager::scanAvailablePackages`

daniellienert:
Note: The tests will never pass as they fail in the install step before this patch will be applied.

kdambekalns (approved):

kdambekalns:
The follow-up error you see does not appear for me. What I did see: Composer asked me to use `--no-plugins` because `The "ocramius/package-versions" plugin was skipped because it requires a Plugin API version ("^1.0.0") that does not match your Composer installation ("2.0.0"). You may need to run composer update with the "--no-plugins" option.`

Once I did that, it skipped our plugin, and installed some of our packages into `Packages/Libraries/neos/`, which lead to `The package with the name "neos/…" was found more than once, please make sure it exists only once.`.

Your error looks more like a leftover from composer dying mid-run…

daniellienert:
@kdambekalns that may be. Let's see if the tests run through

kdambekalns:
On the issue with `ocramius/package-versions` – that is required by `composer/package-versions-deprecated`, which is in turn required by `doctrine/orm`:

```
$ composer why ocramius/package-versions
The "ocramius/package-versions" plugin was skipped because it requires a Plugin API version ("^1.0.0") that does not match your Composer installation ("2.0.0"). You may need to run composer update with the "--no-plugins" option.
composer/package-versions-deprecated  1.11.99  replaces  ocramius/package-versions (1.11.99)
ocramius/package-versions             2.2.0    replaces  composer/package-versions-deprecated (*)

$ composer why composer/package-versions-deprecated
The "ocramius/package-versions" plugin was skipped because it requires a Plugin API version ("^1.0.0") that does not match your Composer installation ("2.0.0"). You may need to run composer update with the "--no-plugins" option.
composer/package-versions-deprecated  1.11.99  replaces  ocramius/package-versions (1.11.99)
doctrine/orm                          2.7.4    requires  composer/package-versions-deprecated (^1.8)
```

See https://github.com/doctrine/orm/issues/8320

kdambekalns:
See https://github.com/composer/package-versions-deprecated/issues/15

kdambekalns:
As pointed out to me my problem was caused by a globally installed `ocramius/package-versions` and running `composer global update` solved this… Case closed. 😎

bwaidelich:
Two related twitter "discussions": https://twitter.com/Ocramius/status/1320055972841295873 https://twitter.com/seldaek/status/1321009579690496001

baschny:
I guess this won't be back ported to Flow 4.3, will it? Even if "someone" provides a pull request? ;) @daniellienert @kdambekalns @albe

baschny:
All right, I understand.. But lets try it: https://github.com/neos/flow-development-collection/pull/2266

Neos.Flow/Classes/Composer/InstallerScripts.php
```diff
@@ -71,7 +71,7 @@ public static function postPackageUpdateAndInstall(PackageEvent $event)
         if (!$operation instanceof InstallOperation && !$operation instanceof UpdateOperation) {
             throw new Exception\UnexpectedOperationException('Handling of operation with type "' . $operation->getJobType() . '" not supported', 1348750840);
```
kdambekalns, line 72: Looking at that again I wonder… what happens if that line is executed and `getJobType()` is called?

bwaidelich: good call.. this will probably break

PR #2191, opened 2020-10-29
TASK: Use composer 2 in travis build

albe:
Related to #2188

t.b.d. should this land in all supported branches to make builds potentially faster?


daniellienert (approved):

kdambekalns:
> t.b.d. should this land in all supported branches to make builds potentially faster?

No harm done, I guess…

albe:
Yeah, guess so. So I'll provide another quick PR targeting 5.3

PR #2192, opened 2020-10-29
BUGFIX: Ensure memcache value is a string before searching for chunk

dogawaf:
`Memcache->get()` can return `false` if the key is not set, 
so the type must be checked before checking if the value was chunked.

**How to verify it**

Configure some caches to use Memcached as backend.
Flush the cache.
Go to the Neos login page: an exception is shown.

Apply the fix.
Flush the cache.
Go to the Neos login page: the login form is displayed.

Fixes: #2194

### DIFF
--- a/Neos.Cache/Classes/Backend/MemcachedBackend.php
+++ b/Neos.Cache/Classes/Backend/MemcachedBackend.php
@@ -260,7 +260,7 @@ class MemcachedBackend extends IndependentAbstractBackend implements TaggableBac
     public function get(string $entryIdentifier)
     {
         $value = $this->memcache->get($this->identifierPrefix . $entryIdentifier);
-        if (substr($value, 0, 13) === 'Flow*chunked:') {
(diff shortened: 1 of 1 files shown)


dogawaf:
This fix deserves a unit test, but I'am not qualified for that...

sorenmalling:
@dogawaf Thans for the PR - do you have a corresponding issue that we can link it to? 

>  Go to the Neos login page: an exception is shown.

Just to this comment and the PR clear.

If the value is `false` the substr fails, becuase it's not a string? Or what exception is being thrown?

dogawaf:
@sorenmalling I've just created the issue: #2194 , I hope the description is clear enough.

sorenmalling (approved):

sorenmalling:
> @sorenmalling I've just created the issue: #2194 , I hope the description is clear enough.

It's perfect! 👍

albe (approved):
Thanks a lot!

PR #2193, opened 2020-10-29
BUGFIX: Skip private methods in SessionObjectMethodsPointcutFilter

kdambekalns:
The matches() method did not skip private methods, so Flow tried to build interceptors for them. That breaks as of the bugfix in https://github.com/neos/flow-development-collection/pull/2131, leading to https://github.com/neos/flow-development-collection/issues/2190

Additionally a potential bug in the regular expression used to skip certain methods is fixed. Without the grouping the first alternative was anchored to the start, the last to the end, but others could be anywhere in the string.

### DIFF
--- a/Neos.Flow/Classes/Session/Aspect/SessionObjectMethodsPointcutFilter.php
+++ b/Neos.Flow/Classes/Session/Aspect/SessionObjectMethodsPointcutFilter.php
@@ -16,6 +16,7 @@ use Neos\Flow\Aop\Pointcut\PointcutFilterInterface;
 use Neos\Flow\ObjectManagement\CompileTimeObjectManager;
 use Neos\Flow\ObjectManagement\Configuration\Configuration as ObjectConfiguration;
 use Neos\Flow\Annotations as Flow;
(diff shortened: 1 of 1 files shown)


kdambekalns:
Doess not work for real (yet)

kitsunet (approved):
Ha, was just writing exactly the things you fixed now, thank you :)

kdambekalns:
Can confirm it fixes the error with Neos! 🎉

bwaidelich (approved):
didn't test yet but high quality Karsten-code as usual. and a unit/functional test is not needed since this will break compilation if it doesn't work anyways :)

kdambekalns:
> functional test is not needed since this will break compilation

Yes, I can confirm that, too. 😜

simonschaufi:
I just got that exact same error as well. Will there be a hotfix release for this?

kdambekalns:
> Will there be a hotfix release for this?

Of course, it is building right now!

albe:
Looks good post-merge. Thanks Karsten!

PR #2195, opened 2020-10-30
TASK: Clarify signal/slot documentation

### LINKED ISSUE #2061: Add Notice to Signals and Slots Docs to use ::class and beware of common file mistake
### Description

New developers could interpret the documentation regarding Signals and Slots (https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/SignalsAndSlots.html) wrong and use the file path to the class instead of the Namespace. 

### Resolution
We should update the example to use `ClassName::class `instead of a String containting the namespace since this is considered best practice. This would also eliminate the misinterpretation of the docs.

kdambekalns:
- recommend use of `::class` constant
- explicitly point out signal name vs method name difference
- add example for slot being a closure

Resolves #2061


bwaidelich (approved):
You're the man!

PR #2197, opened 2020-10-30
BUGFIX: Reduce maximum line length to 80 chars

simonschaufi:
This reduces the maximum line length of output to 80 chars when running core migrations.

See https://stackoverflow.com/questions/4651012/why-is-the-default-terminal-width-80-characters for more information


sorenmalling:
@simonschaufi Thanks for the pull requets. Do you have a corresponding issue, that can relate to this PR?

Fllowup question: This is changed in the `migrate.php` line - so does this make any change when executing regular command controllers?

simonschaufi:
> Do you have a corresponding issue, that can relate to this PR?

No, do I need to create one?

> does this make any change when executing regular command controllers?

No, this is only about migrations.

You can reproduce it if you execute `./flow flow:core:migrate My.Package`.

kdambekalns (approved):
🤷‍♂️ why not…

PR #2200, opened 2020-10-31
TASK: Run installation and compilation of neos distribution in builds

albe:
Run composer installation and flow compilation of the neos-base-distribution with lowest supported PHP version to check if this breaks with the Flow change.
This is supposed to reduce introducing breaking changes for Neos into Flow. See https://twitter.com/bwaidelich/status/1322482916996382722

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,12 +49,17 @@ jobs:
       env: DB=mysql BEHAT=true
       addons:
         mariadb: '10.2'
+    - php: 7.1
+      env: DB=mysql INTEGRATION=true
+      addons:
+        mariadb: '10.2'
 cache:
   directories:
     - $HOME/.composer/cache
 before_install:
   - export NEOS_TARGET_REPOSITORY=neos/flow-development-collection
   - export NEOS_TARGET_VERSION=5.3
+  - export NEOS_CMS_TARGET_VERSION=4.3
   - if [ "$DB" = "pgsql" ]; then sudo mount -o remount,size=25% /var/ramfs ; fi
   - phpenv config-rm xdebug.ini || echo "xdebug not available"
   - env
@@ -80,6 +85,7 @@ before_script:
   - echo 'report_memleaks = 0' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - echo 'bin/phpunit --colors -c Build/BuildEssentials/PhpUnit/UnitTests.xml && exit 0' > unittest.sh
   - echo 'if [ $? -eq 139 ]; then exit 0 ; else exit $? ; fi' >> unittest.sh
+  - if [ "$INTEGRATION" = "true" ]; then cd .. && git clone https://github.com/neos/neos-base-distribution.git -b ${NEOS_CMS_TARGET_VERSION} && cd neos-base-distribution; fi
   - rm -f Configuration/Routes.yaml
   - cp Configuration/Settings.yaml.example Configuration/Settings.yaml
   - Build/BuildEssentials/TravisCi/SetupDatabase.sh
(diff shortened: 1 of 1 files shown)


albe:
Okay, this seems to mostly do what is expected now. However...
```
  - Upgrading neos/flow-development-collection (5.3.x-dev 98caa52 => dev-travis ca8559d): Checking out ca8559d015 from cache
> Neos\Flow\Composer\InstallerScripts::postPackageUpdateAndInstall
  - Upgrading neos/form (4.3.x-dev a7ce71b => 5.2.0): Checking out 85feb5d2a3 from cache
> Neos\Flow\Composer\InstallerScripts::postPackageUpdateAndInstall
  - Upgrading neos/setup (4.0.x-dev 0bed74b => dev-master 4f6cdfb): Checking out 4f6cdfbe8d from cache
```
https://travis-ci.com/github/neos/flow-development-collection/jobs/423284556#L920-L924

This leads to the `compile` failing with
```
The object "Neos\Flow\Mvc\FlashMessage\FlashMessageService" which was specified as a property in the object configuration of object "Neos\Form\Finishers\FlashMessageFinisher" (automatically registered class) does not exist.
Check for spelling mistakes and if that dependency is correctly
configured.
  Type: Neos\Flow\ObjectManagement\Exception\UnknownObjectException
  Code: 1265213849
  File: Packages/Framework/Neos.Flow/Classes/ObjectManagement/DependencyInjection/P
        roxyClassBuilder.php
  Line: 461
```

Why does it upgrade neos/form to 5.2.0 (and neos/setup to dev-master) after upgrading to the current branch of flow? Seems as if the flow branch is considered to be equal `dev-master`, while it should be `5.3.x-dev`. See https://travis-ci.com/github/neos/flow-development-collection/jobs/423284556#L860 and
https://github.com/neos/form/blob/master/composer.json#L8

cc @bwaidelich @kdambekalns maybe?

albe:
Also - if we would change `ComposerManifestUpdater` https://github.com/neos/BuildEssentials/blob/master/TravisCi/ComposerManifestUpdater.php#L40 to just add the `flow-development-collection` require if it doesn't exist, we could build/compile the `neos-base-distribution` instead of the `development-distribution`. I'm not entirely sure that's needed, though the end-user is gonna install the former and this check is supposed to verify that works.

sorenmalling:
I disagree with this - I get the idea and what it can help with. 

Still, Neos CMS is build on top of Neos Flow, so doesn't this belongs more to the Neos CMS build chain

albe:
> Still, Neos CMS is build on top of Neos Flow, so doesn't this belongs more to the Neos CMS build chain

Yes, but we need feedback inside Flow directly if it breaks our most important "consumer". What drawbacks do you see apart from the feeling that it doesn't belong here? I mean, we do run this already in Neos tests, so if we don't want it here, we don't it at all. 

What else could we do to avoid Flow breaking Neos with a change?

kdambekalns:
Wait, if the goal is to make sure Flow works with the lowest supported PHP version, that should simply be done by using the lowest supported PHP version for running FLow tests… no?

bwaidelich:
> Wait, if the goal is to make sure Flow works with the lowest supported PHP version

No, the goal is to make sure that the latest _Neos_ version runs with the latest Flow version in order to prevent (bugfix) releases that break Neos.

I didn't look at the implementation but I don't see any issue in running this additional CI check even though Flow does not depend on Neos

kdambekalns:
To be fair, the bug was discovered before the release was even announced… And TBH for such a rare case it is quite a bad trade-off to slow down every CI build even more. Could we have this run once a day instead? Travis offers "cron jobs", I saw…

bwaidelich:
> To be fair, the bug was discovered before the release was even announced
I think nobody blames anyone! But it would be still be nice (and prevent the additional work of creating a bugfix-bugfix release) if we could have easily detected it.

> And TBH for such a rare case it is quite a bad trade-off to slow down every CI build even more. Could we have this run once a day instead?

Good call.. In fact we would only need it prior to an actual release

albe:
> trade-off to slow down every CI build

That's a fair point. Maybe only run that on version branches? Or maybe we do this as GH action?

albe:
See #2273 - let's incorporate this there and get that working.

Issue #2201, opened 2020-11-01
Accept \IteratorAggregate as a collection type in validation


sorenmalling:
### Description

The validation of Collections (CollectionValidator) makes use of the value from the `getProperty` when getting values from a model (or any other things being validated.

In this case, we created a RoleIdentifiers collection object to be used in the Account object. This is, to return a collection of objects, instead of string. But when persisting the Account model, the object is not accepted as a valid collection.

To solve this, we add `\IteratorAggregate` to `TypeHandling::$collectionTypes` and advice to use this as a interface to implement, when creating such valueObject collection classes

PR #2202, opened 2020-11-01
FEATURE: Accept \Traversable as a collection type in validation

sorenmalling:
This adds \Traversable to the array of valid collectionTypes in the TypeHandling class.

Fixes: #2201

### DIFF
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -34,7 +34,7 @@ abstract class TypeHandling
     /**
      * @var array
      */
-    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class, \IteratorAggregate::class];
+    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class, \Traversable::class];
 
     /**
      * Returns an array with type information, including element type for
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -34,7 +34,7 @@ abstract class TypeHandling
     /**
      * @var array
      */
-    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class, \Traversable::class];
+    protected static $collectionTypes = ['array', \Traversable::class];
 
     /**
      * Returns an array with type information, including element type for
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -34,7 +34,7 @@ abstract class TypeHandling
     /**
      * @var array
      */
-    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class];
+    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class, \IteratorAggregate::class];
 
     /**
      * Returns an array with type information, including element type for
--- a/Neos.Utility.ObjectHandling/Tests/Unit/TypeHandlingTest.php
+++ b/Neos.Utility.ObjectHandling/Tests/Unit/TypeHandlingTest.php
@@ -203,7 +203,8 @@ class TypeHandlingTest extends \PHPUnit\Framework\TestCase
             ['SplObjectStorage', true],
             ['Doctrine\Common\Collections\Collection', true],
             ['Doctrine\Common\Collections\ArrayCollection', true],
-            ['IteratorAggregate', true]
+            ['IteratorAggregate', true],
+            ['Iterator', true]
         ];
     }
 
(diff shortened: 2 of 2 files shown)


Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -34,7 +34,7 @@ abstract class TypeHandling
     /**
      * @var array
      */
-    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class];
+    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class, \IteratorAggregate::class];
```
albe, line 37: Hm, would it maybe make sense to add `\Traversable` instead of `\IteratorAggregate`? What we try to decide with the collection types is if the value can be `foreach`'d through (right?), which `\Traversable` is providing. `\IteratorAggregate` only adds to this the possibility to externalize the iterator to achieve the traversal instead of implementing the iteration methods itself. Alternatively we should also **add** `\Iterator`, as that and `\IteratorAggregate` are the two primary implementation interfaces for `\Traversable` (https://www.php.net/manual/en/class.traversable.php).
We could btw then even remove the `\ArrayObject`, `\SplObjectStorage` and `Collection`, since they all implement either `\Iterator` or `\IteratorAggregate`, but keeping them is a slight optimization for those cases.
```suggestion
    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class, \Traversable::class];
```

Neos.Utility.ObjectHandling/Tests/Unit/TypeHandlingTest.php
```diff
@@ -202,7 +202,8 @@ public function collectionTypes()
             ['ArrayObject', true],
             ['SplObjectStorage', true],
             ['Doctrine\Common\Collections\Collection', true],
-            ['Doctrine\Common\Collections\ArrayCollection', true]
+            ['Doctrine\Common\Collections\ArrayCollection', true],
+            ['IteratorAggregate', true]
```
albe, line 206:
```suggestion
            ['IteratorAggregate', true],
            ['Iterator', true]
```
If we do the above, we can add a test for `\Iterator` too

albe (commented):
Good move, but maybe we can open it up even more then.
We just need to get sure we did not intend to (or unintendedly did) encode an additional expectation (`ArrayAccess` or `Countable`?) in the `isCollection()` check, because then this would be breaking. i.e. is no place that uses `isCollection()` expecting the result to be countable or array accessible? In Flow at least that's not the case

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
sorenmalling: But `\Traversable` is an abstract class - it does not contain any method, so how can it be defined code wise (for ex `RoleIdentifiers`) that it can be iterated?

albe: It can't, it's an internal class, but all classes that implement `\Iterator` or ´\IteratorAggregate` are also `\Traversable`.
> This is an internal engine interface which cannot be implemented in PHP scripts. Either IteratorAggregate or Iterator must be used instead. When implementing an interface which extends Traversable, make sure to list IteratorAggregate or Iterator before its name in the implements clause.

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -34,7 +34,7 @@ abstract class TypeHandling
     /**
      * @var array
      */
-    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class];
+    protected static $collectionTypes = ['array', \ArrayObject::class, \SplObjectStorage::class, Collection::class, \Traversable::class];
```
bwaidelich, line 37:
```suggestion
    protected static $collectionTypes = ['array', \Traversable::class];
```
`ArrayObject`, `SplObjectStorage` and `Collection` already implement/extend `Traversable`

sorenmalling:
@bwaidelich @albe Should we us `is_iterable()` in the validator as a general if condition code?

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
albe: > We could btw then even remove the \ArrayObject, \SplObjectStorage and Collection, since they all implement either \Iterator or \IteratorAggregate, but keeping them is a slight optimization for those cases.

bwaidelich: > We could btw then even remove the [...]

That's what I suggested :)
Or did you mean in the test case? I would keep it there to avoid future regressions.

But I think we agreed to fix this whole thing in the CollectionValidator instead, right?

albe: > but keeping them is a slight optimization for those cases.

But yes, let's fix it in the validator. No side-effects.

kdambekalns (approved):

kdambekalns:
From original PR description:

---


**How to verify it**
 * Create your own VO collection class, implementing \IteratorAggregate and use it in the `get*` method (used for validating)
 * Try to persist your Model - validation will fail
 * Apply patch, and it will be accepted

**Checklist**

- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html) (considered this a feature, so only targeted master - correct me if wrong)

albe (approved):

bwaidelich (approved):

PR #2203, opened 2020-11-01
BUGFIX: InternalRequestEngine uses middlewares

albe:
Since the introduction of the PSR-15 middlewares in #1928, the internal request engine should have invoked those but didn't until now.

Related to #2019


sorenmalling (commented):
I'm missing a test case to try it, but from reading and comparing I'm good with it

bwaidelich (approved):
Looks good by reading

albe:
> I'm missing a test case to try it

We'll have a lot of test cases with the middlewares changes, because the functional tests will then depend on the middlewares to function :)

PR #2214, opened 2020-11-03
!!! BUGFIX: Relative position to non-existing key in PositionalArraySorter throws exception

### LINKED ISSUE #2213: PositionalArraySorter forgets items that are positioned relative to non-existing key
See https://github.com/neos/flow-development-collection/pull/2207#discussion_r516978425

The problem is https://github.com/neos/flow-development-collection/blob/master/Neos.Utility.Arrays/Classes/PositionalArraySorter.php#L223-L228 and https://github.com/neos/flow-development-collection/blob/master/Neos.Utility.Arrays/Classes/PositionalArraySorter.php#L250-L254

It should be checked if the matched key `$matches[1]` exists inside the *original untouched* `$arrayKeysWithPosition`. If not, the item should be inserted into `middleKeys` or an exception be thrown.
If the original order within the settings should be kept, then `position` of the key needs to be changed, and `middleKeys` be collected **after** before/after keys.

albe:
Until now, an element positioned relative to a non-existing key would just be skipped silently. With this, it will throw an exception to bring awareness to the "inactive" array element.

This is a breaking change because previously referring to a non existing position would be ignored.
For example:
```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Some.Package':
          position: 'before Some.NonExistingPackage'
```
previously: The corresponding routes would be inserted according to the loading order (i.e. non-deterministic basically)
now: An `InvalidPositionException` exception is thrown:
```
The positional string "before Some.NonExistingPackage" (defined for key "Some.Package") references a non-existing key.
```

Fixes #2213


albe:
Well, that second failure has an unexpected outcome:

```
2) Neos\Utility\Arrays\Tests\Unit\PositionalArraySorterTest::getSortedKeysTests with data set #15 ('Position after non-existing key', array(array('after second'), array('after nonexisting'), array('start')), 'position', array('first', 'second', 'third'))
Position after non-existing key
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
     0 => 'first'
-    1 => 'second'
-    2 => 'third'
+    1 => 'third'
+    2 => 'second'
+    3 => 'third'
 )
```
So for the `after` case, it does not skip the item, but add it twice?!

bwaidelich:
I love that you started with a failing test!

bwaidelich:
PR should be against lowest supported branch, though

bwaidelich:
"good first issue" - I don't know. That code is bloody complex 😄
But after giving it a quick shot and thinking about it more I think we should throw an exception when referring to a non existing key.
Good news: The `master` branch would the the right one if we go for that solution 😁

albe:
> "good first issue" - I don't know. That code is bloody complex

yeah well, I thought if the exact place to fix is known and it is covered with a greenable test, that's a good starting point.

> Good news: The master branch would the the right one if we go for that solution 😁

Bad news: The current test case is the totally wrong place when we throw an exception 😄

bwaidelich:
https://github.com/neos/flow-development-collection/blob/master/Neos.Utility.Arrays/Tests/Unit/PositionalArraySorterTest.php#L104-L107 this will probably fail now

albe:
> 'message' => 'Position before adds after start if named element not present',
> message' => 'Position after adds before end if named element not present',

🤯 So we had tests that covered exactly this case? But why did it fail with the new test then?? This bug seems to be more subtle

bwaidelich:
Yes, that's what I was trying to say with
> Obviously a breaking change (there is even a test case that breaks with that change)

in the issue :)

albe:
This is good to go now IMO @bwaidelich

Neos.Utility.Arrays/Classes/PositionalArraySorter.php
```diff
@@ -113,25 +113,19 @@ public function toArray(): array
...
 
-        $sortedKeysMap = $this->generateSortedKeysMap();
-
-        $sortedKeys = [];
-        array_walk_recursive($sortedKeysMap, function ($value) use (&$sortedKeys) {
-            $sortedKeys[] = $value;
-        });
-
-        return $sortedKeys;
+        return $this->generateSortedKeysMap();
```
bwaidelich, line 128: Do you know why the previous code is not needed any longer?

bwaidelich: Ah, the `array_merge([], ...$sortedKeysMap)` below, right?

But now the result will be an associative array, right?
If so, would it make sense to do
```suggestion
        return array_values($this->generateSortedKeysMap());
```
instead?

bwaidelich: Nevermind. The result is actually a numeric array and a unit test even covers that case

bwaidelich (approved):
Great one!

PR #2216, opened 2020-11-04
FEATURE: Pass SignalInformation instance to slot if possible

### LINKED ISSUE #1003: Signal/Slot should use an object when calling a slot
Right now the arguments passed to a signal are inherently "unstable", in the sense that what is passed is not strictly defined. So just switching on the "pass signal information" flag in wiring could have unintended side-effects.

We (@kitsunet and I) concluded: it would be good to pass an object instead, which could carry any arguments like before but have a defined API to fetch those as well as further data about the signal.

This could be added in a b/c way by checking the type of the (sole) argument expected by the slot.

kdambekalns:
With the new `wire()` method signal/slot connections use an instance of
`SignalInformation` as parameter for the called slot method.

Slots connected using `connect()` continue to receive a string argument
`EmitterClassName::signalName` like before, if requested.

Resolves #1003


Neos.Flow/Classes/SignalSlot/Dispatcher.php
```diff
@@ -82,11 +87,25 @@ public function connect($signalClassName, $signalName, $slotClassNameOrObject, $
             $method = $slotMethodName;
         }
 
+        if (is_object($slotClassNameOrObject) || class_exists($slotClassNameOrObject)) {
+            try {
+                if ($slotClassNameOrObject instanceof \Closure) {
+                    $methodParameters = (new \ReflectionFunction($slotClassNameOrObject))->getParameters();
+                } else {
+                    $methodParameters = (new \ReflectionMethod($slotClassNameOrObject, trim($method, ':')))->getParameters();
+                }
```
bwaidelich, line 96: Nice trick to make the change backwards compatible! But is it OK to use reflection during runtime? Or – if not – could this be put into a `compile static` function?

kdambekalns: Turned out `class_exists()` broke everything during compile phase. So, bad idea either way. The ReflectionService cannot be used here, as it is not (always) available at this point.

kdambekalns:
Dang, forgot to mark this as ready for review a week ago… 🙈

Neos.Flow/Classes/SignalSlot/SignalInformation.php
```diff
@@ -0,0 +1,60 @@
...
+ * source code.
+ */
+
+/**
+ * A SignalInformation instance carries information about the signal that arrives
+ * at a slot during a signal/slot dispatch operation.
+ *
+ * @api
+ */
+class SignalInformation
```
albe, line 22: We can even make this final I think (and maybe also proxy=false?)
```suggestion
final class SignalInformation
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/SignalsAndSlots.rst
```diff
@@ -41,44 +41,68 @@ like in this example:
...
 
 Defining Slots
 ==============
 
-Basically any method of any class can be used as a slot, even if never written
-specifically for being a slot. The only requirement is a matching signature between signal
-and slot, so that the parameters passed to the signal can be handed over to the slot
-without problems. The following shows a slot, as you can see it differs in no way from any
-non-slot method.
+You can use any method as a slot, or write one specifically for use s a slot.
```
albe, line 59:
```suggestion
You can use any method as a slot, or write one specifically for use as slot.
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/SignalsAndSlots.rst
```diff
@@ -41,44 +41,68 @@ like in this example:
...
+
+Any method of any class can be used as a slot, even if never written specifically
+for being a slot. The only requirement is a matching signature between signal and
```
albe, line 62: This sentence somehow repeats the previous sentence, just with slightly different words, no?

Neos.Flow/Classes/SignalSlot/SignalInformation.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public function getSignalName(): string
+    {
+        return $this->signalName;
+    }
+
+    public function getSignalArguments(): array
+    {
+        return $this->signalArguments;
```
albe, line 58: Maybe we can add a convenience method `getSignalArgument($argumentName)` that checks existence of the argument and throws (or returns NULL). Would avoid weird `->getSignalArguments()[$argumentName] ??` syntax on the caller side

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/SignalsAndSlots.rst
kdambekalns: Yes, but… I added two more headlines now.

albe (approved):

kdambekalns:
@bwaidelich Can this be merged?

bwaidelich (approved):

bwaidelich:
> Can this be merged?

Yes, once CI is happy. Thanks for this!

bwaidelich:
waiting for travis...

kdambekalns:
Well, it was green before the merge of master. Your call if you want to have that tested explicitly. 😜

bwaidelich:
I fully trus you, but hopefully we have a CI again by tomorrow so that we can avoid abusing admin rights to merge this :)

albe:
Should become green now 👍

PR #2218, opened 2020-11-05
BUGFIX: Fix and tweaks in PropertyMapperTest

kdambekalns:
This tweaks some assertions and by this exposes a test that was "risky"
in fact did not expose failure of expected behavior.


albe (approved):

jonnitto (approved):
Fine (by reading)

PR #2219, opened 2020-11-06
FEATURE: Move DispatchComponent to middleware

albe:
This moves the SetHeader, ReplaceHttpResponse and DispatchComponent to a single DispatchMiddleware.

(copied from https://github.com/neos/flow-development-collection/issues/2019#issuecomment-720463971)

> There was a discussion about the fact that `ActionResponse::setComponentParameter()` won't work with the middleware approach since it is only possible to pass information (other than the HTTP *response*) _down_ into the middleware chain, but the `Dispatch` MW has to be the inner most.
> 
> A few options were discussed:
> 
> ### Re-introduce some kind of global context
> 
> ...similar to the `ComponentContext`...
> We don't like that solution because it defeats the idea of having a clean PSR-15 implementation
> 
> ### Introduce a separate MVC Middleware Sub-Chain
> 
> This would introduce a new level of complexity – and probably only for really edgy edge cases that can be solved otherwise (see below)
> 
> ### Replace `ActionResponse::setComponentParameter()`
> 
> Our favorite solution for now, even if more breaking (see comment regarding b/c)
> 
> * deprecate `ActionResponse::setComponentParameter()`
>   * ...and make it throw an exception if the `$componentClassName` parameter is not `ReplaceHttpResponseComponent` or `SetHeaderComponent`
> * Introduce `ActionResponse::setHeader()`
>   * ...and provide core migration to turns `ActionResponse::setComponentParameter(SetHeaderComponent::class, 'foo', 'bar')` into `ActionResponse::setHeader('foo', 'bar')`
> * Introduce `ActionResponse::replaceHttpResponse()`
>   * ...and provide core migration to turns `ActionResponse::setComponentParameter(ReplaceHttpResponseComponent::class, ReplaceHttpResponseComponent::PARAMETER_RESPONSE, $response)` into `ActionResponse::replaceHttpResponse($response)`
> * Introduce `DispatchMiddleware` (see below)
>   * Keep it as simple as possible so that it can be replaced in case that's required
> * Document breaking changes and alternatives to communicate with other middlewares (since setComponentParameter is no longer available)
>   * Use `ServerRequest.withAttribute()` to communicate with _inner_ middlewares
>   * Replace Dispatch Middleware with a custom implementation
>   * HACK: Pass information via Http Response (e.g. a custom header that is removed in an outer MW)

Related to #2019
Depends on #2223


albe:
Maybe the ActionResponse Headers is stretching things a bit... have removed this for now and will create another PR suggesting to change to that. I do think the `Headers` class handles some cases for abstracting headers very well (e.g. `Cache-Control` and date based headers like `Last-Modified`) and should be the underlying ActionResponse API

psalm-baseline.xml
```diff
@@ -960,13 +960,9 @@
...
-    <InvalidNullableReturnType occurrences="1">
-      <code>string</code>
-    </InvalidNullableReturnType>
-    <NullableReturnStatement occurrences="2">
-      <code>null</code>
-      <code>$cacheControl === '' ? null : $cacheControl</code>
-    </NullableReturnStatement>
+    <PossiblyNullReference occurrences="1">
+      <code>expire</code>
+    </PossiblyNullReference>
```
albe, line 965:
```suggestion
    <InvalidNullableReturnType occurrences="1">
      <code>string</code>
    </InvalidNullableReturnType>
    <NullableReturnStatement occurrences="2">
      <code>null</code>
      <code>$cacheControl === '' ? null : $cacheControl</code>
    </NullableReturnStatement>
```

bwaidelich (approved):
+ 1 by reading

Neos.Flow/Classes/Mvc/DispatchMiddleware.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    protected $actionRequestFactory;
+
+    /**
+     * Create an action request from stored route match values and dispatch to that
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $next): ResponseInterface
+    {
+        $routingMatchResults = $request->getAttribute(ServerRequestAttributes::ROUTING_RESULTS) ?? [];
+        $actionRequest = $this->actionRequestFactory->createActionRequest($request, $routingMatchResults);
```
albe, line 46:
```suggestion
        $actionRequest = $request->getAttribute(ServerRequestAttributes::ACTION_REQUEST) ?? $this->actionRequestFactory->createActionRequest($request, $routingMatchResults);
```
Since we already build the `ActionRequest` in `SecurityEntryPoint` for now, we can reuse it here.

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -153,32 +151,79 @@ public function deleteCookie(string $cookieName): void
...
+            $this->setHttpHeader($parameterName, $value);
+            return;
         }
-        $this->componentParameters[$componentClassName][$parameterName] = $value;
+
+        if ($componentClassName === ReplaceHttpResponseComponent::class && $parameterName === 'response') {
+            $this->replaceHttpResponse($value);
+            return;
+        }
+        throw new \InvalidArgumentException('todo');
```
albe, line 167:
```suggestion
        throw new \InvalidArgumentException('setComponentParameters is deprecated and will not work with anything other than SetHeaderComponent and ReplaceHttpResponseComponent argument. If you want to send data to your middleware from the action, use response headers or introduce a global context. Both solutions are to be considered bad practice though.');
```
Wording @bwaidelich @sorenmalling ?

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -153,32 +151,79 @@ public function deleteCookie(string $cookieName): void
...
+     * @param array|string|\DateTime $headerValue An array of values or a single value for the specified header field
+     * @return void
      */
-    public function getContent(): string
+    public function setHttpHeader(string $headerName, $headerValue): void
     {
-        $content = $this->content->getContents();
-        $this->content->rewind();
-        return $content;
+        if ($headerValue instanceof \DateTimeInterface) {
```
albe, line 179: This is taken from the `Headers` class, which should eventually replace this implementation and add more response API methods.

albe:
Much easier to review now. Left some comments

Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -153,32 +151,79 @@ public function deleteCookie(string $cookieName): void
      * @param string $parameterName
      * @param mixed $value
      * @return void
-     * @api
+     * @deprecated since Flow 7.0 use setHttpHeader or replaceHttpResponse instead. See also #linkToDocs
```
bwaidelich, line 154: `#linkToDocs`?  ;)
We should also mention here, that this only works with a `$componentClassName` of "SetHeaderComponent" or "ReplaceHttpResponseComponent"

Neos.Flow/Classes/Mvc/ActionResponse.php
bwaidelich: What about
```suggestion
        throw new \InvalidArgumentException(sprintf('The method %s is deprecated. It will only allow a $componentClassName parameter of "%s" or "%s". If you want to send data to your middleware from the action, use response headers or introduce a global context. Both solutions are to be considered bad practice though.', __METHOD__, SetHeaderComponent::class, ReplaceHttpResponseComponent::class), 1605088079);
```

Neos.Flow/Classes/Mvc/ActionResponse.php
bwaidelich: We should add this as comment IMO

Neos.Flow/Classes/Mvc/DispatchMiddleware.php
bwaidelich: right, but I would suggest to remove line 45, too and throw an exception if the request is not available at this point.
Silently creating one at this point (skipping security initialization) seems error prone to me

bwaidelich (changes requested):
Looks great, thanks!
Just a couple of nitpicks and one actual issue: IMO we should fail if the ActionRequest wasn't created before this middleware is invoked.. Or at least make that an option of the middleware so it can be activated explicitly

Neos.Flow/Classes/Mvc/ActionResponse.php
albe:
```suggestion
        // This is taken from the Headers class, which should eventually replace this implementation and add more response API methods.
        if ($headerValue instanceof \DateTimeInterface) {
```

Neos.Flow/Classes/Mvc/DispatchMiddleware.php
albe: Well, the question is if the dispatch strictly depends on the security entrypoint. As is this is probably correct, so... yeah.

albe, line 46:
```suggestion
        $actionRequest = $request->getAttribute(ServerRequestAttributes::ACTION_REQUEST);
        if ($actionRequest === null) {
            throw new Exception('No ActionRequest was created before the DispatchMiddleware. Make sure you have the SecurityEntryPointMiddleware configured before dispatch.', 1605091292);
        }
```

Neos.Flow/Classes/Mvc/ActionResponse.php
albe, line 154: Yeah, I would like to add a link to the rendered release notes with the explanation here, so I put the placeholder in place. Maybe we can remove it for now and just not forget it (and if we do, we're at least not stuck with a strange hashtag).
```suggestion
     * @deprecated since Flow 7.0 use setHttpHeader or replaceHttpResponse instead. For now this will still work with $componentClassName of "SetHeaderComponent" or "ReplaceHttpResponseComponent" only.
```

albe:
Applied the suggestions, thanks! :)

bwaidelich (approved):
Lovely! Thanks!

Neos.Flow/Tests/Unit/Mvc/DispatchMiddlewareTest.php
```diff
@@ -0,0 +1,287 @@
...
+        $this->mockHttpRequest->method('getUploadedFiles')->willReturn([]);
+
+        $this->mockDispatcher = $this->getMockBuilder(Dispatcher::class)->getMock();
+        $this->inject($this->dispatchMiddleware, 'dispatcher', $this->mockDispatcher);
+
+        $this->mockActionRequest = $this->getMockBuilder(ActionRequest::class)->disableOriginalConstructor()->getMock();
+        $this->mockActionRequestFactory = $this->getMockBuilder(ActionRequestFactory::class)->disableOriginalConstructor()->onlyMethods(['prepareActionRequest'])->getMock();
+        $this->mockActionRequestFactory->method('prepareActionRequest')->willReturn($this->mockActionRequest);
+
+        $this->inject($this->dispatchMiddleware, 'actionRequestFactory', $this->mockActionRequestFactory);
```
albe, line 109:
```suggestion
        $this->mockHttpRequest->method('getAttribute')->with(ServerRequestAttributes::ACTION_REQUEST)->willReturn($this->mockActionRequest);
```

Neos.Flow/Tests/Unit/Mvc/DispatchMiddlewareTest.php
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * @var ActionRequest|\PHPUnit\Framework\MockObject\MockObject
+     */
+    protected $mockActionRequest;
+
+    /**
+     * @var ActionRequestFactory|\PHPUnit\Framework\MockObject\MockObject
+     */
+    protected $mockActionRequestFactory;
```
albe, line 67:
```suggestion
```

Neos.Flow/Classes/Mvc/DispatchMiddleware.php
```diff
@@ -0,0 +1,54 @@
...
+     * @Flow\Inject
+     * @var Dispatcher
+     */
+    protected $dispatcher;
+
+    /**
+     * @Flow\Inject
+     * @var ActionRequestFactory
+     */
+    protected $actionRequestFactory;
```
albe, line 38:
```suggestion
```

Neos.Flow/Classes/Mvc/DispatchMiddleware.php
```diff
@@ -0,0 +1,54 @@
...
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * A dispatch middleware that runs the current HTTP request through the MVC stack
+ */
+class DispatchMiddleware implements MiddlewareInterface
+{
+    /**
+     * @Flow\Inject
```
albe, line 29:
```suggestion
     * @Flow\Inject(lazy=false)
```

albe:
Hopefully tests should now pass again. Had to move all the ActionRequest building tests to SecurityEntryPointMiddleware tests, which is meh, but well... I left a "NOTE" docblock

psalm-baseline.xml
```diff
@@ -1237,6 +1237,12 @@
       <code>null</code>
     </PossiblyNullPropertyAssignmentValue>
   </file>
+  <file src="Packages/Framework/Neos.Flow/Classes/Mvc/ActionResponse.php">
+    <UndefinedClass occurrences="2">
```
albe, line 1241:
```suggestion
    <UndefinedClass occurrences="4">
```

albe:
Merging now with your prior approval @bwaidelich - we can then focus on collecting feedback and removing the component left-overs via #2221

PR #2207, opened 2020-11-03
FEATURE: Move Routing, AjaxWidget and ParseRequestBodyComponent to Middleware

albe:
This moves the Routing-, AjaxWidget- and ParseRequestBodyComponent to PSR-15 middlewares.

Related to #2019
Depends on #2204

PR #2223, opened 2020-11-06
FEATURE: Move SecurityEntryPointComponent to middleware

albe:
This moves the SecurityEntryPointComponent to a PSR-15 middleware. As a side-effect, this also removes the PrepareMvcRequestComponent, as the functionality is now also done by the SecurityEntryPoint.
If you want to build a middleware that depends on the security framework, place it `after securityEntryPoint`.

Related to #2019


Neos.Flow/Classes/Security/Exception/AuthenticationRequiredException.php
```diff
@@ -22,4 +24,9 @@ class AuthenticationRequiredException extends \Neos\Flow\Security\Exception
      * @var integer
      */
     protected $statusCode = 401;
+
+    /**
+     * @var ActionRequest
+     */
+    public $interceptedRequest;
```
albe, line 31: I'm not 100% sure about this - but it is the best idea I have to transport the ActionRequest to the outside SecurityEntryPointMiddleware. Otherwise we need to do the work inside the dispatcher (see https://github.com/neos/flow-development-collection/pull/2223/files#diff-34a65e00b17dd586f553255a62368688485b195e8ef5cc1878222156ca1452bfR106-R108)

Neos.Flow/Classes/Mvc/Dispatcher.php
```diff
@@ -97,20 +96,18 @@ public function injectFirewall(FirewallInterface $firewall)
...
         } catch (AuthenticationRequiredException $exception) {
-            // This case does exist!
-            /** @var ActionResponse $response */
-            $response->setComponentParameter(
-                SecurityEntryPointComponent::class,
-                SecurityEntryPointComponent::AUTHENTICATION_EXCEPTION,
-                $exception
-            );
-            return;
+            // TODO: Set intercepted request for security context here (?)
```
albe, line 106: See https://github.com/neos/flow-development-collection/pull/2223/files#diff-0ac5e31517d0f8fe5938f6af908c97ecc4cf5cb3255306e4274f5ea7ff2fc433R31

Neos.Flow/Classes/Mvc/Dispatcher.php
```diff
@@ -97,20 +96,18 @@ public function injectFirewall(FirewallInterface $firewall)
      */
     public function dispatch(ActionRequest $request, ActionResponse $response)
     {
+        $this->securityContext->setRequest($request);
```
albe, line 99: See https://github.com/neos/flow-development-collection/pull/2207/files#r516980799

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
```diff
@@ -0,0 +1,86 @@
...
+            });
+
+            if ($tokensWithEntryPoint === []) {
+                $this->securityLogger->notice('No authentication entry point found for active tokens, therefore cannot authenticate or redirect to authentication automatically.');
+                throw $authenticationException;
+            }
+
+            $response = $this->buildHttpResponse();
+            foreach ($tokensWithEntryPoint as $token) {
+                if ($token->isAuthenticated()) {
```
albe, line 58: This check wasn't there before. I'm not 100% sure about this, but shouldn't tokens that are authenticated **not** lead to a authentication entry point invokation?

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
```diff
@@ -0,0 +1,86 @@
...
+            $tokensWithEntryPoint = array_filter($this->securityContext->getAuthenticationTokens(), static function (TokenInterface $token) {
+                return $token->getAuthenticationEntryPoint() !== null;
+            });
+
+            if ($tokensWithEntryPoint === []) {
+                $this->securityLogger->notice('No authentication entry point found for active tokens, therefore cannot authenticate or redirect to authentication automatically.');
+                throw $authenticationException;
+            }
+
+            $response = $this->buildHttpResponse();
```
albe, line 56: Creating a new response (through a method for test-friendlyness), because we caught an exception, so there is no valid response built in the inner layer.

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
```diff
@@ -0,0 +1,86 @@
...
+                    continue;
+                }
+                $entryPoint = $token->getAuthenticationEntryPoint();
+                $this->securityLogger->info(sprintf('Starting authentication with entry point of type "%s"', \get_class($entryPoint)), LogEnvironment::fromMethodName(__METHOD__));
+
+                // Only store the intercepted request if it is a GET request (otherwise it can't be resumed properly)
+                // We also don't store the request for "sessionless authentications" because that would implicitly start a session
+                // TODO: Adjust when a session-independent storing mechanism is possible (see https://github.com/neos/flow-development-collection/issues/1693)
+                if (!$token instanceof SessionlessTokenInterface && $request->getMethod() === 'GET') {
+                    $this->securityContext->setInterceptedRequest($authenticationException->interceptedRequest);
```
albe, line 68: See comments below regarding the storage of the actionRequest in the exception

Neos.Flow/Classes/Security/Exception/AuthenticationRequiredException.php
```diff
@@ -22,4 +24,37 @@ class AuthenticationRequiredException extends \Neos\Flow\Security\Exception
...
+
+    /**
+     * Attach the given action request as intercepted request and return self.
+     *
+     * @param ActionRequest $actionRequest
+     * @return AuthenticationRequiredException
+     */
+    public function withInterceptedRequest(ActionRequest $actionRequest): self
+    {
+        $this->interceptedRequest = $actionRequest;
```
albe, line 41: Why not clone $this and make it immutable? Because exceptions can not be cloned in PHP :)

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
```diff
@@ -0,0 +1,100 @@
...
+    protected $actionRequestFactory;
+
+    /**
+     * @inheritDoc
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $next): ResponseInterface
+    {
+        try {
+            // FIXME: Currently the security context needs an ActionRequest, therefore we need to build it here
+            $routingMatchResults = $request->getAttribute(ServerRequestAttributes::ROUTING_RESULTS) ?? [];
```
albe, line 53: As discussed with @bwaidelich - the SecurityContext right now has an unfortunate dependency on the `ActionRequest`. Since this middleware wraps the security concept of the framework, and the `ActionRequest` is not really supposed to exist within the PSR middleware, this is the place to create it. That way, we do not reintroduce a middleware that creates an ActionRequest that 3rd party could depend on. The `ServerRequestAttributes::ACTION_REQUEST` is marked internal for this matter.

That's also why the `PrepareMvcRequestComponent` is removed with this now.

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
```diff
@@ -0,0 +1,100 @@
...
+            $actionRequest = $this->actionRequestFactory->createActionRequest($request, $routingMatchResults);
+            $this->securityContext->setRequest($actionRequest);
+            return $next->handle($request->withAttribute(ServerRequestAttributes::ACTION_REQUEST, $actionRequest));
```
bwaidelich, line 56: only this line should go into the `try` part IMO

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
bwaidelich: that double negation was too much for my monday-night braing :)

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Build a base HTTP Response in case of a missing authentication exception
+     *
+     * @return ResponseInterface
+     */
+    protected function buildHttpResponse(): ResponseInterface
+    {
+        // 403 to be b/c with the previous implementation, but 401 would be more fitting
```
bwaidelich, line 97: We could make that an option of the middleware (possibly in a separate PR of course)

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
albe: Ah, I forgot to mention this one (three) line(s) in our call @bwaidelich - maybe you have an idea if I am missing something or this causes bad side-effects if we add them?

Neos.Flow/Classes/Mvc/DispatchComponent.php
```diff
@@ -34,7 +35,7 @@ class DispatchComponent implements ComponentInterface
      */
     public function handle(ComponentContext $componentContext)
     {
-        $actionRequest = $componentContext->getParameter(DispatchComponent::class, 'actionRequest');
+        $actionRequest = $componentContext->getHttpRequest()->getAttribute(ServerRequestAttributes::ACTION_REQUEST);
```
bwaidelich, line 38: A little
```
if ($actionRequest === null) {
  throw meaningfullexception..
}
```
would be nice here in case the MW chain is configured wrongly or something else went wrong

Neos.Flow/Classes/Security/Exception/AuthenticationRequiredException.php
bwaidelich: Good point, I wasn't aware of that.
Just a minor issue with this. The `with*` pattern is used for immutable objects in our code base (and commonly in PHP anyways). I think we should go for a good old setter in this case, if immutability is not an option

bwaidelich (commented):
Looks great, thanks a lot for your work!
Just some minor nitpicks and comments

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
bwaidelich: Sure, let's talk about it tomorrow!

Neos.Flow/Classes/Mvc/DispatchComponent.php
albe: That thing will be gone with the follow-up PR anyway :)

Neos.Flow/Tests/Unit/Http/Middleware/SecurityEntryPointMiddlewareTest.php
```diff
@@ -73,32 +73,28 @@ class SecurityEntryPointComponentTest extends UnitTestCase
...
-            }
-            return null;
-        });
 
-        $this->mockAuthenticationRequiredException = new AuthenticationRequiredException();
+        $mockActionRequestFactory = $this->getMockBuilder(ActionRequestFactory::class)->getMock();
+        $mockActionRequestFactory->method('createActionRequest')->willReturn($this->mockActionRequest);
+        $this->inject($this->securityEntryPointMiddleware, 'actionRequestFactory', $mockActionRequestFactory);
+
+        $this->mockAuthenticationRequiredException = (new AuthenticationRequiredException())->withInterceptedRequest($this->mockActionRequest);
```
albe, line 96:
```suggestion
        $this->mockAuthenticationRequiredException = (new AuthenticationRequiredException())->attachInterceptedRequest($this->mockActionRequest);
```

bwaidelich (approved):
Looking good by reading.
I assume that we'll have to fine-tune a few things in the aftermatch

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
```diff
@@ -0,0 +1,100 @@
...
+            if ($tokensWithEntryPoint === []) {
+                $this->securityLogger->notice('No authentication entry point found for active tokens, therefore cannot authenticate or redirect to authentication automatically.');
+                throw $authenticationException;
+            }
+
+            $response = $this->buildHttpResponse();
+            foreach ($tokensWithEntryPoint as $token) {
+                if ($token->isAuthenticated()) {
+                    continue;
+                }
```
albe, line 72:
```suggestion
```

albe:
Merging and rebasing the other PR now, so that becomes easier to review.

Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
albe: Just for the record: We decided to remove this again and look at it separately - see #2234

PR #2226, opened 2020-11-09
BUGFIX: Adjust functional RoutingTest

kdambekalns:


bwaidelich (approved):
Thanks!

Neos.Flow/Configuration/Testing/Settings.yaml
```diff
@@ -49,8 +49,9 @@ Neos:
                 logFileURL: '%FLOW_PATH_DATA%Logs/I18n_Testing.log'
                 severityThreshold: '%LOG_DEBUG%'
 
-
-
+    mvc:
+      routes:
+        'Neos.Flow': true
```
bwaidelich, line 54: This will break functional tests in the Neos distribution

bwaidelich (changes requested):
It would be good to include Flow routes by default in testing context, but we probably have to do that in the CI for now

Neos.Flow/Configuration/Testing/Settings.yaml
kdambekalns: Dropped the commit again…

bwaidelich (approved):

albe (approved):

PR #2229, opened 2020-11-09
TASK: Use count instead of find in repository deleting resource

### LINKED ISSUE #2228: deleteResource function of ResourceManager inefficient for large number of resources
### Description

When a resource is updated (or replaced in a model), Flow automatically calls the deleteResource function of the ResourceManager.
This is using the findBySha1AndCollectionName function to check if the file of the resource is still being used (https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/ResourceManagement/ResourceManager.php#L331).
In this function it is used purely to check if the file of the resource is used for other resources by checking the SHA1 value so a simple count would be enough.
However, findBySha1AndCollectionName is reading all resources with the same SHA1 value from the given collection, reads them into an array.
If you have lots of resources with the same SHA1 hash, this can be quite expensive and should be done differently IMHO.

### Steps to Reproduce

1. Upload a file as a PersistentResource
2. Measure speed
3. Upload the same file 30.000 times
4. Measure speed again and compare results.

#### Expected behavior

Do the check based on SHA1 value in an efficient fashion so performance degradation is not as extreme as it is right now.

#### Actual behavior

Reading all PersistentResources into an array can lead to memory exhausted issues.

### Affected Versions

As far as I know this behaviour has always been that way in Flow.

DavidSporer:
This introduces a new function in the resourceRepository that allows to count the number of times a file is referenced in a PersistentResource. This function is used in deleteResource of the ResourceManager to improve speed when a file is used in a lot of Persistent Resources

Fixes #2228

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/ResourceRepository.php
+++ b/Neos.Flow/Classes/ResourceManagement/ResourceRepository.php
@@ -263,7 +263,7 @@ class ResourceRepository extends Repository
      *
      * @return int
      */
-    public function countBySha1AndCollectionName($sha1Hash, $collectionName)
+    public function countBySha1AndCollectionName(string $sha1Hash, string $collectionName): int
     {
         $query = $this->createQuery();
         $query->matching(
--- a/Neos.Flow/Classes/ResourceManagement/ResourceRepository.php
+++ b/Neos.Flow/Classes/ResourceManagement/ResourceRepository.php
@@ -256,6 +256,33 @@ class ResourceRepository extends Repository
     }
 
     /**
+     * Counts all resources with the same SHA1 hash and collection
+     *
+     * @param string $sha1Hash
+     * @param string $collectionName
+     *
+     * @return int
+     */
+    public function countBySha1AndCollectionName($sha1Hash, $collectionName)
+    {
+        $query = $this->createQuery();
+        $query->matching(
+            $query->logicalAnd(
+                $query->equals('sha1', $sha1Hash),
+                $query->equals('collectionName', $collectionName)
+            )
+        );
+        $noOfResources = $query->count();
+        foreach ($this->addedResources as $importedResource) {
+            if ($importedResource->getSha1() === $sha1Hash && $importedResource->getCollectionName() === $collectionName) {
+                $noOfResources++;
+            }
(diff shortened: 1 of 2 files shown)


albe:
Makes a lot of sense! The build (psalm) error is a false-negative. Or rather hints again at our incomplete type hinting. We'd need to change the docblocks of the QueryInterface methods logicalAnd/logicalOr to `@param mixed ...$constraint1` as shown in https://github.com/vimeo/psalm/issues/148#issuecomment-296687779

albe (approved):

Neos.Flow/Classes/ResourceManagement/ResourceRepository.php
```diff
@@ -255,6 +255,33 @@ public function findBySha1AndCollectionName($sha1Hash, $collectionName)
...
 
+    /**
+     * Counts all resources with the same SHA1 hash and collection
+     *
+     * @param string $sha1Hash
+     * @param string $collectionName
+     *
+     * @return int
+     */
+    public function countBySha1AndCollectionName($sha1Hash, $collectionName)
```
albe, line 266:
```suggestion
    public function countBySha1AndCollectionName(string $sha1Hash, string $collectionName): int
```
Let's add typehints to newly added methods

kdambekalns (approved):

kdambekalns:
The only thing I wonder about: Does anything speak against adding this to 5.3 and up?

DavidSporer:
> The only thing I wonder about: Does anything speak against adding this to 5.3 and up?

No, would absolutely make sense. How can I do that? Creating a new PR or can this be changed?

kdambekalns:
> No, would absolutely make sense. How can I do that? Creating a new PR or can this be changed?

Changing the base of the PR would pull in a lot more, I fear. Easiest will be to cherry-pick the commit to a new branch based on 5.3 and create a new PR (or force-push to this one)…

DavidSporer:
> Changing the base of the PR would pull in a lot more, I fear. Easiest will be to cherry-pick the commit to a new branch based on 5.3 and create a new PR (or force-push to this one)…

Here's the new PR: #2241

PR #2231, opened 2020-11-10
!!! TASK: Add type declarations to persistence interfaces

kdambekalns:


Neos.Flow/Classes/Persistence/Doctrine/Query.php
```diff
@@ -69,7 +68,7 @@ class Query implements QueryInterface
     /**
      * @var array
      */
-    protected $orderings;
+    protected $orderings = [];
 
     /**
      * @var integer
```
albe, line 74:
```suggestion
     * @var integer|null
```
Same for the offset below, but GH doesn't allow suggestions/comments on lines outside the diff lines

albe (commented):
Not sure about the ```ERROR: ImplementedReturnTypeMismatch - Packages/Framework/Neos.Flow/Classes/Persistence/Doctrine/Repository.php:130:16 - The inherited return type 'array<array-key, object>' for Doctrine\Persistence\ObjectRepository::findAll is different to the implemented return type for Neos\Flow\Persistence\Doctrine\Repository::findall 'Neos\Flow\Persistence\QueryResultInterface' (see https://psalm.dev/123)
     * @return QueryResultInterface The query result```
But guess we need to baseline this.

kdambekalns:
> `The inherited return type 'array<array-key, object>' for Doctrine\Persistence\ObjectRepository::findAll is different to the implemented return type for Neos\Flow\Persistence\Doctrine\Repository::findall 'Neos\Flow\Persistence\QueryResultInterface'`
> But guess we need to baseline this.

Yeah. The `QueryResultInterface` extends `\Countable, \Iterator, \ArrayAccess`, so it fits, even though strrictly speaking is breaks the contract. But it's only a "docblock contract", no real return type is declared in Doctrine.

Should I baseline this within this PR?

albe:
Yeah, would be good I think.

albe:
green ✔️

albe (approved):

PR #2237, opened 2020-11-11
TASK: Raise minimum PHP version to 7.3

albe:
Require PHP 7.3 in composer.json

### DIFF
--- a/Neos.Cache/composer.json
+++ b/Neos.Cache/composer.json
@@ -5,7 +5,7 @@
     "homepage": "http://flow.neos.io",
     "license": ["MIT"],
     "require": {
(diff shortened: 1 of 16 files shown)


albe:
🤦‍♂️

kdambekalns (changes requested):
This is not enough, at least it **must** be raised in `Neos.Flow/composer.json`, too. But **should** be done in `Neos.*/composer.json`, for consistency - then we can rely on that version in all packages…

albe:
Needed to correct the commit message anyway, so rewrite history ftw

kdambekalns (approved):
👏

PR #2241, opened 2020-11-12
TASK: Use count instead of find in repository deleting resource

### LINKED ISSUE #2228: deleteResource function of ResourceManager inefficient for large number of resources

DavidSporer:
This introduces a new function in the resourceRepository that allows to count the number of times a file is referenced in a PersistentResource. This function is used in deleteResource of the ResourceManager to improve speed when a file is used in a lot of Persistent Resources.
Same change as with #2229 but should now be merged into Flow 5.3.

Fixes #2228

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/ResourceManager.php
+++ b/Neos.Flow/Classes/ResourceManagement/ResourceManager.php
@@ -327,9 +327,9 @@ class ResourceManager
 
         $collectionName = $resource->getCollectionName();
 
-        $result = $this->resourceRepository->findBySha1AndCollectionName($resource->getSha1(), $collectionName);
-        if (count($result) > 1) {
-            $this->logger->debug(sprintf('Not removing storage data of resource %s (%s) because it is still in use by %s other PersistentResource object(s).', $resource->getFilename(), $resource->getSha1(), count($result) - 1));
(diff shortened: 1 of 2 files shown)


kdambekalns (approved):

albe (approved):
Thanks! :)

Issue #2243, opened 2020-11-12
Improve psalm setup / integration


kdambekalns:
- [x] have `vimeo/psalm` in the dev dist manifest (not hidden in the dev collection) [neos/flow-development-distribution#51]
- [x] update it to v4 [#2246]

see #2205
see #1960

kdambekalns:
> Yeah, I have that on my radar, but I think we should then try to keep versions across our branches as close as possible (optimally in sync), because upmerging the baseline might otherwise become a hassle.

(@albe in https://github.com/neos/flow-development-collection/pull/1880#issuecomment-726582974)

kdambekalns:
I see two options:

- ignore the baseline when merging from 6.3 to master/7.0 (like we do with XLIFF files)
- move the psalm setup to a separate manifest (similar to the way we handle Behat)

Because "in sync" would mean PHP 7.3 as a minimum with Psalm 4, and that is something we cannot do for Flow < 7.

albe:
> would mean PHP 7.3 as a minimum with Psalm 4

Are you sure?
https://github.com/vimeo/psalm/blob/master/composer.json#L17

kdambekalns:
Yes, I was sure. Looked it up yesterday. But whatever I saw there… 🙈 is not true. Even better!

albe:
> move the psalm setup to a separate manifest (similar to the way we handle Behat)

I'd need to check how that works exactly. That would mean having the psalm-baseline in another package/repo, which would need to be updated whenever a PR with new/fixed psalm errors is merged.

> ignore the baseline when merging from 6.3 to master/7.0 (like we do with XLIFF files)

But that would also mean we'd need to manually update the baseline after upmerge for every branch, no? so that would be viable with #1960

kdambekalns:
> That would mean having the psalm-baseline in another package/repo, which would need to be updated whenever a PR with new/fixed psalm errors is merged.

No, the baseline file could stay where it is. And it could even be just a separate composer manifest in a directory.

But: No need to separate anything, at least for now. 😎

kdambekalns:
I'd say this can be closed.

PR #2245, opened 2020-11-13
TASK: Upgrade psalm to 4.1

albe:
This just upgrades the psalm version across the board for now.

Related to #2243


albe:
Closing in favor of #2246

PR #2256, opened 2020-11-16
BUGFIX: AjaxWidgetsMiddleware initializes SecurityContext

albe:
Otherwise, the security context is not initialized and security would not work but throw an exception (e.g. Neos.Setup)

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/Widget/AjaxWidgetMiddleware.php
+++ b/Neos.FluidAdaptor/Classes/Core/Widget/AjaxWidgetMiddleware.php
@@ -79,6 +79,7 @@ class AjaxWidgetMiddleware implements MiddlewareInterface
 
         $actionRequest = $this->actionRequestFactory->createActionRequest($httpRequest, ['__widgetContext' => $widgetContext]);
         $actionRequest->setControllerObjectName($widgetContext->getControllerObjectName());
+        $this->securityContext->setRequest($actionRequest);
 
         $actionResponse = new ActionResponse();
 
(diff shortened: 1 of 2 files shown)


albe:
Test failures, see https://github.com/neos/flow-development-collection/pull/2257

kdambekalns (approved):
Sounds reasonable.

albe:
Only thing is this is now efffectively a "before dispatch", so this does not shortcut anything like routing or so unless we move the securityEntryPoint way earlier

bwaidelich:
> unless we move the securityEntryPoint way earlier

But Security requires an `ActionRequest` which is only available after Routing. Or how was that solved before?

albe:
> But Security requires an ActionRequest which is only available after Routing. Or how was that solved before?

🤔 good question. Looking at it, it seemed to be included "before routing" originally https://github.com/neos/flow-development-collection/blob/6.3/Neos.FluidAdaptor/Configuration/Settings.yaml#L8

bwaidelich (changes requested):
As discussed: This worked before since the Widget Component created an `ActionRequest` and initialized the Security Context: https://github.com/neos/flow-development-collection/blob/6.3/Neos.FluidAdaptor/Classes/Core/Widget/AjaxWidgetComponent.php#L82-L84

And that makes sense, because the original route is not considered if the widget kicks in

albe:
👍 approve this rewrite of this change :)

albe:
`sudo merge` now - the fix is obvious and travis only complained about #2257

PR #2257, opened 2020-11-16
!!! BUGFIX: Adjust to TYPO3Fluid 2.5.11 and 2.6.10 signature changes

albe:
With this you need to update to TYPO3 Fluid 2.5.11+ or 2.6.10+

See https://github.com/TYPO3/Fluid/commit/f20db4e74cf9803c6cffca2ed2f03e1b0b89d0dc#r44244534

PR #2259, opened 2020-11-17
!!! TASK: Remove custom FluidAdaptor Exceptions on invalid ArgumentDefinition

albe:
This removes the `Neos\FluidAdaptor\Core\Exception`s when the ArgumentDefinition is invalid in favor of the native TYPO3 Fluid exceptions. With this we remove the boilerplate we have to keep in sync with upstream.

See https://github.com/TYPO3/Fluid/issues/529 and https://github.com/neos/flow-development-collection/pull/2257#issuecomment-728825319

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractViewHelper.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractViewHelper.php
@@ -79,56 +79,6 @@ abstract class AbstractViewHelper extends FluidAbstractViewHelper
     }
 
     /**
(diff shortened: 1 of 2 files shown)


albe:
TODO: Check if there are more side-effects of changing those exceptions

kitsunet (approved):
I don't think there are, I originally put them in the adaptor to not change error codes and exception types just in case. I would say it's fine.

kdambekalns (approved):
👏

bwaidelich (approved):
Ha, nice one, thanks! (+1 by reading)

Issue #2258, opened 2020-11-17
Smooth upgrade for Component- to Middleware chain


bwaidelich:
### Description

With #2019 core HTTP components will be converted to the middleware chain and with #2221 the HTTP component infrastructure is being removed from Flow.

We should

- [x] Document the new middleware implementation (#2308)
- [x] Document how to migrate existing packages/components (#2330)
- [x] Provide a Core Migration that can detect (and partly migrate?) HTTP components (#2315 & #2320)
- [x] ~Potentially detect `Neos.Flow.http.chain` configuration and throw a helpful exception in order to prevent subtle bugs to creep in (#2306)~ (we agreed that a proper detection/core migration would be better)
- [x] Provide CLI command that lists all configured middleware components in the correct order (#2307)

albe:
@bwaidelich regarding the "Document how to migrate existing packages/components" - where should we document that? Only in the release notes on neos.io? Or as part of the PR message for the generated release notes?

bwaidelich:
I think a good start would be examples of at least two different kinds of middleware.
Then we could link to those from the release notes and the PR saying sth like

```php
$componentContext->replaceHttpResponse($componentContext->getHttpResponse()->withAddedHeader('X-Foo', 'bar');
```
=> This is how you tweak the response now

```php
$componentContext->replaceHttpResponse($newResponse);
$componentContext->setParameter(ComponentChain::class, 'cancel', true);
```

=> This is how you replace the response now

```php
$componentContext->setParameter(...)
```

=> This is how you communicate with other mw components now

PR #2261, opened 2020-11-17
BUGFIX: Fix default order of middleware components

bwaidelich:
Adjusts the order of the Middleware components so that
the `SessionMiddleware` is executed before the `RoutingMiddleware`.

Otherwise session based authentication won't work until the
routing middleware was executed.

This also removes most of the explicit `position` configurations
in order to avoid too much interdependency.

If a 3rd party middleware needs to be executed before/after another
one, it can still use `position: before/after <name>`  of course.
Depending on the order of _multiple_ other components is considered
bad practice. But if that's really required one could still add a
`position` setting to the existing middleware configuration.

Related: #2019

### DIFF
--- a/Neos.Flow/Configuration/Settings.Http.yaml
+++ b/Neos.Flow/Configuration/Settings.Http.yaml
@@ -31,10 +31,10 @@ Neos:
           position: 'start 100'
           middleware: 'Neos\Flow\Http\Middleware\StandardsComplianceMiddleware'
         'trustedProxies':
-          position: 'start'
+          position: 'start 50'
           middleware: 'Neos\Flow\Http\Middleware\TrustedProxiesMiddleware'
         'session':
-          position: 'start'
+          position: 'start 10'
           middleware: 'Neos\Flow\Http\Middleware\SessionMiddleware'
 
         # The execution order of the following middleware components is deliberately not specified.
--- a/Neos.Flow/Configuration/Settings.Http.yaml
+++ b/Neos.Flow/Configuration/Settings.Http.yaml
@@ -26,32 +26,34 @@ Neos:
 
       # Defines the PSR-15 middlewares that a request goes through in a flow
       middlewares:
+        # The following middleware components are executed first since they influence the incoming request or outgoing response
+        'standardsCompliance':
+          position: 'start 100'
+          middleware: 'Neos\Flow\Http\Middleware\StandardsComplianceMiddleware'
         'trustedProxies':
           position: 'start'
           middleware: 'Neos\Flow\Http\Middleware\TrustedProxiesMiddleware'
+        'session':
+          position: 'start'
+          middleware: 'Neos\Flow\Http\Middleware\SessionMiddleware'
+
+        # The execution order of the following middleware components is deliberately not specified.
+        # To run a middleware before or after an existing one, you can use "position: before <name>" or "position: after <name>"
         'routing':
-          # needs to be "after trustedProxies" because AjaxWidget comes "before routing" but still needs to be after trustedProxies
-          position: 'after trustedProxies'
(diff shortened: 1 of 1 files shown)


mficzel:
I totally like the concept but why not remove the "position" at all from the flow middleware configurations since they are already ordered in the flow default configuration. 

That way only packages that add to the flow middleware chain would need "position" to locate the middlewares in the chain.

bwaidelich:
> why not remove the "position" at all from the flow middleware configurations

Because the few left usually need to be executed before/after all the others (unless explicitly specified).
Otherwise, depending on the loading order, a 3rd party middleware could be executed after the `dispatch` middleware for example (effectively ignoring it) or before `standardsCompliance` (allowing to create a non-compliant response)

mficzel:
It will always be possible to screw up by messing with middlewares. I do not think that those position arguments will prevent that. 

Luckily the flow-configurationis is loaded first and non positioned middlewares will end up at the end after dispatch where they will be ignored since dispatch never calls $next->handle. So adding a position argument would be the only way for packages to get their middlewares to be executed if we remove the position from the flow-midleware configuration. At least for me that makes sense.

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -26,32 +26,34 @@ Neos:
...
       middlewares:
+        # The following middleware components are executed first since they influence the incoming request or outgoing response
+        'standardsCompliance':
+          position: 'start 100'
+          middleware: 'Neos\Flow\Http\Middleware\StandardsComplianceMiddleware'
         'trustedProxies':
           position: 'start'
           middleware: 'Neos\Flow\Http\Middleware\TrustedProxiesMiddleware'
+        'session':
+          position: 'start'
```
mficzel, line 37: This leads to two middlewares sharing position "start" which is ambiguous. This makes it hard to reflect about the order if third party packages place middlewares relative to "session" as "trustedProxies" may or may not be executed before.

mficzel:
Jus to be clear i would be fine with position arguments if they add value and make things easier understandable. What i really would like to have is a clearly defined and easy to understand order of the core middlewares that allows to locate other middlewares relative to those.

bwaidelich:
@mficzel Thanks again for your feedback, but I'm not sure whether I got it right

> Luckily the flow-configurationis is loaded first

That's only true if the composer dependencies are set correctly. I've seen a lot of bugs due to the 
```yaml
privilegeTargets: []
```
default configuration in Flow that reset custom privileges. Of course the actual bug is the missing dependency, but why not make it a bit more fault-tolerant. Especially if it doesn't really cost us anything?

> non positioned middlewares will end up at the end after dispatch where they will be ignored

I think that's a misunderstanding:
Positions that can't be resolved (e.g. no position set at all or `before nonexisting` or `after nonexisting`) will always be sorted after positions `start` and before positions `end`.

If I don't specify any positions for the core middlewares and add a new one..
..`position before nonexisting` will put it at the *beginning* of the chain (not what we want)
..`position after nonexisting` will put it at the *end* of the chain (not what we want at all)

With this PR applied..
..`position before nonexisting` will put it after the last `position start` (i.e. `session`)
..`position after nonexisting` will put it before the first `position end` (i.e. `dispatch`)


> This leads to two middlewares sharing position "start" which is ambiguous

Yes, by intention. We should not care whether `trustedProxies` is run before `session` vice versa. But they should be executed fairly early since they affect the incoming request/outgoing response.

> i would be fine with position arguments if they add value and make things easier understandable

I think we can provide some easy best practices with #2258

albe (approved):

kitsunet (approved):

daniellienert (approved):

bwaidelich:
Thank you guys! <3

PR #2262, opened 2020-11-17
!!! TASK: Remove deprecated code

bwaidelich:
### Remove obsolete and deprecated PHP code:
* `Cli/Request::getMainRequest()` & `Cli/Request::isMainRequest()`
  * Those were deprecated with 6.0 (via #1552) and never really served a purpose since
    CLI requests can't be nested
* `Neos\Flow\Persistence\Generic\*`
  * Before we had doctrine, we had a custom persistence layer that was kept as "generic" persistence when we introduced doctrine ten years ago (via 90cb65827c1550e9144e9f83b9231b430c106660). Since 6.0 this custom persistence was deprecated in favor of the corresponding `Neos\Flow\Persistence\Doctrine\*` classes.
    generic persistence is dead, long live generic persistence! (and thanks to @kdambekalns for spending million of minutes and brain cells on this)
* `Neos\Flow\Security\Cryptography\SaltedMd5HashingStrategy`
  * md5 is unsafe and the hashing strategy was deprecated with 6.0.
* `ObjectAccess::instantiateClass()`
  * deprecated with 5.3.16 (via #1972). With PHP 5.6+ `new $className(...$arguments)` can be used instead
* `HttpRequestHandlerInterface`/`HttpRequestHandler::getHttpResponse()`
  * deprecated with 6.0 (via #1755) and now gone. If you need the current HTTP Response, use a middleware, as the Response does not exist earlier at all

Related: #2172


sorenmalling:
Related draft PR #2181

sorenmalling:
> Related draft PR #2181

Can be closed in favor of this - I don't mind :)

bwaidelich:
@sorenmalling I'm sorry, I wasn't aware of your pr. will check tomorrow!

sorenmalling:
I believe it's mostly the same :-)

albe (approved):

Neos.Flow/Classes/Http/RequestHandler.php
```diff
@@ -124,7 +124,7 @@ public function getHttpRequest()
      * Returns the HTTP response corresponding to the currently handled request
      *
      * @return ResponseInterface|null
-     * @deprecated Don't depend on this method. The HTTP response only exists after the innermost middleware (dispatch) is done. For that stage use a middleware instead.
+     * @deprecated since 6.0. Don't depend on this method. The HTTP response only exists after the innermost middleware (dispatch) is done. For that stage use a middleware instead.
```
sorenmalling, line 127: "Deprecated since 6.0" - should we add when it should be removed, if not in 7.0?

sorenmalling:
How did the verdict on removal of Component code go?

sorenmalling:
Compared to https://github.com/neos/flow-development-collection/pull/2181/files I can see we do not remove ex. PsrSystemLoggerInterface?

Neos.Flow/Classes/Http/RequestHandler.php
albe: Yes, technically we could already remove it. Given using it is pretty useless as of now, it's maybe even sensible?

kdambekalns (changes requested):
There is documentation that needs to be updated (on generic persistence, probably more…)

bwaidelich:
> How did the verdict on removal of Component code go?

How do you mean? That was removed with #2019 (and we still have to smoothen the bumps with #2258)

> Compared to #2181 I can see we do not remove ex. PsrSystemLoggerInterface?

Good point, I missed that one when I compared the two (https://github.com/neos/flow-development-collection/pull/2181#issuecomment-729627924).

> There is documentation that needs to be updated (on generic persistence, probably more…)

True that..


@sorenmalling @kdambekalns Thanks for your feedback, I'll prepare a follow-up

Neos.Flow/Classes/Http/RequestHandler.php
bwaidelich: mh.. What about a helpful exception instead?

sorenmalling:
> How do you mean? That was removed with #2019 (and we still have to smoothen the bumps with #2258)

Roger that - got lost in back/forth on keeping or not 👍

bwaidelich:
Update: The Psr Logger interfaces have been removed with #2134, that's why they don't appear in the diff :)

Neos.Flow/Classes/Http/RequestHandler.php
bwaidelich: I removed it from the interface with 3e6c754da4d1e3b054723f4fa41013042c54e7e4 and added a `throw new \BadMethodCallException()` here to make it a bit easier for devs to catch this case

bwaidelich:
Looking at the documentation I found a lot of outdated examples and code.
I tried to address those with a separate issue (see #2280) 

> There is documentation that needs to be updated (on generic persistence, probably more…)

It seems like (apart from the CGL tweaks mentioned above) it's only the generic persistence part that is affected.
I'm a bit puzzled currently. @kdambekalns do you think it would be enough to just remove the part from the docs: https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Persistence.html#generic-persistence 
?

bwaidelich:
@kdambekalns could you remove your -1 if you agree with the current version?

Neos.Flow/Classes/Http/RequestHandler.php
bwaidelich: Doh.. We rely on that part in `Scripts::initializeExceptionStorage()` and `FileStorage::__construct()` and I think @albe an me already figured that out before.. How should we fix that part?

albe: Argh, yes right. I think there is one thing we need to consider:
- an exception logged inside the MVC layer/dispatch will not log a meaningful Response at all, so it should probably not log one at all -> those two places need to handle a `null`
- for exceptions to be logged during the middleware-chain, it could log the Response up to that point, but that means we'd need to adjust the chain with another `onStep()` callback, that updates the `Response` in the `RequestHandler`. I would beg the question though: "Does that make sense/provide helpful information?"

Maybe the whole exception storage logging process should be deferred and be done after the process finishes (`register_shutdown_function`?)? At that point it's more likely that `Response` is actually the one that was returned to the browser

kdambekalns (approved):
Looks good now…

Neos.Flow/Classes/Http/RequestHandler.php
bwaidelich: I think it's fine to skip the response from the output actually.. I doubt that it contains any useful information if an exception is thrown. @kitsunet what do you think?

bwaidelich: I removed it with 446259ab76d8a4dab1198f9f4da8ad1919632690 for now.. now waiting for CI and approval/objection :)

albe:
D'oh, that was `nop` StyleCI PR :D

albe:
We could also strip out the old autoloader instead of #2288, but TBH I never set the `FLOW_ONLY_COMPOSER_LOADER=true` env var to check if nothing breaks and I guess most didn't. Keeping the flag will at least provide an escape-hatch for another major version in case it causes problems.

sorenmalling (approved):

bwaidelich:
> We could also strip out the old autoloader

I prefer the way you went with #2288 now since it leave us with the mentioned escape hatch, thanks for that!

PR #2265, opened 2020-11-19
!!! BUGFIX: Make FluidAdaptor compatible with TYPO3Fluid 2.5.11+ and 2.6.10+

### LINKED ISSUE #2260: Flow 5.3-6.3 is incompatible with TYPO3 Fluid 2.5.11+ and 2.6.10+
See https://github.com/neos/flow-development-collection/pull/2257 - fixing this would be a breaking change though, because we need to adjust the signature of an `@api` method. Alternatively, we disallow installing those versions - but that is shutting down our users from any further security fixes (like the one causing this breaking change).

albe:
This is breaking in case you created your own ViewHelper that overrides the `registerArgument()` or `overrideArgument()` method. In that case you need to add a new boolean optional argument `$escape = null` and forward that to the parent method.

This is a backport of #2257
Fixes #2260

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/Parser/SyntaxTree/ResourceUriNode.php
+++ b/Neos.FluidAdaptor/Classes/Core/Parser/SyntaxTree/ResourceUriNode.php
@@ -49,8 +49,6 @@ class ResourceUriNode extends ViewHelperNode
         $this->uninitializedViewHelper = $this->viewHelperResolver->createViewHelperInstanceFromClassName($this->viewHelperClassName);
         $this->uninitializedViewHelper->setViewHelperNode($this);
         $this->argumentDefinitions = $this->viewHelperResolver->getArgumentDefinitionsForViewHelper($this->uninitializedViewHelper);
-        $this->rewriteBooleanNodesInArgumentsObjectTree($this->argumentDefinitions, $this->arguments);
-        $this->validateArguments($this->argumentDefinitions, $this->arguments);
     }
 
     /**
--- a/Neos.FluidAdaptor/Classes/Core/Parser/SyntaxTree/ResourceUriNode.php
+++ b/Neos.FluidAdaptor/Classes/Core/Parser/SyntaxTree/ResourceUriNode.php
@@ -49,6 +49,8 @@ class ResourceUriNode extends ViewHelperNode
         $this->uninitializedViewHelper = $this->viewHelperResolver->createViewHelperInstanceFromClassName($this->viewHelperClassName);
         $this->uninitializedViewHelper->setViewHelperNode($this);
         $this->argumentDefinitions = $this->viewHelperResolver->getArgumentDefinitionsForViewHelper($this->uninitializedViewHelper);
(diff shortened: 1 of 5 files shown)


albe:
Still need to adjust composer dependency to `~2.5.11 || ^2.6.10`

kdambekalns:
This affects Flow 4.3 as well, do we consider that worth a fix? It somehow falls into the "security" category, which means until December 4.3 is still covered…

kdambekalns:
We should get this merged and released: When using `roave/security-advisories` you cannot pin to 2.6.9 any longer…

kdambekalns:
Question: why not "backport" https://github.com/neos/flow-development-collection/pull/2259 as well?

bwaidelich:
> Question: why not "backport" #2259 as well?

Yep, I had the same question. @albe any reason for this over #2259 for old versions? I think the "breakiness" is roughly equal

bwaidelich:
Doh, I just realized it's not as easy because we rely on the former exception in https://github.com/neos/flow-development-collection/blob/5.3/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractViewHelper.php#L123 for example (for 6+ this was removed with #1574)

albe:
Well, we could fix that catch block. But yeah, it is a bit more breaking than just adjusting the signature. Anyway, without https://github.com/neos/flow-development-collection/pull/2265/commits/90edda3a6f1369ab1a8865b9474526f0a86acc40 and the Fluid version constraint the argument validation will run twice - for latest Fluid versions. Should we accept that?

bwaidelich (approved):
I reverted my last commit since it is more breaking than I thought initially.
Tested the original fix again and it works just fine.
Will merge if CI is green

albe:
> Alexander Berl:  5 minutes ago
Don't forget to raise the Fluid version constraint!

Or we revert 90edda3 too

bwaidelich:
> Or we revert 90edda3 too

@albe I lost track now.. And since I made a lot of problematic decisions today (including an accidental direct push to the `5.3` branch) I don't dare to hit the merge button now.. I guess I'll have to look at this when my brain works again, or you guys take it from here

albe:
@kdambekalns what do you think? Restrict Fluid version (which I'd prefer) or run with potentially running the arguments validation twice (should not break something but is obviously unnecessary work)?

kdambekalns:
> @kdambekalns what do you think? Restrict Fluid version (which I'd prefer) or run with potentially running the arguments validation twice (should not break something but is obviously unnecessary work)?

Restricting is fine with me, the older versions are rejected by `roave/security-advisories` and should not be used anyway.

albe:
Talking about stupid little mistakes... 🧸

albe:
So this should be fine now - I'd merge once travis passes (again). Can also do upmerges and tag releases later this evening or maybe tomorrow

kaystrobach:
2.4.3 => 2.4.4 for flow 4.x is also breaking.

kdambekalns:
But anything below Flow 5.3 is EOL by now…

albe:
Right. Anything Flow < 5.3 should pin the Fluid dependency.

kaystrobach:
That’s what we do right now 👍

PR #2266, opened 2020-11-19
BUGFIX: Make InstallerScripts compatible to composer version 2.0+

baschny:
Instead of querying the removed method ::getJobType we now check the class of the job instance like we do in the first lines of the method.

Cherry-picked from: f10e2570b04ad03efe27b1e2821e8d66f40cab3b

Fixes: #2187


kdambekalns:
The same as #2188, so 👍 - but looking at this I came up with one question, see https://github.com/neos/flow-development-collection/pull/2188#pullrequestreview-534660935

albe (approved):
The thing Karsten mentioned is unrelated to this. It will error out when it would be about to throw an exception. That's unfortunate, but not a dealbreaker for this IMO.

Issue #2031, opened 2020-06-19
Session cookies should set sameSite attribute


kdambekalns:
### Description

To avoid problems in the future the session cookies should set the sameSite attribute.

### Steps to Reproduce

1. Open browser console
2. Use a Flow site using sessions, e.g. the Neos UI

#### Actual behavior

`Cookie “Neos_Session” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite`

https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

albe:
Related to https://github.com/neos/flow-development-collection/pull/1948

kdambekalns:
Doh! So we have this already? How could I not know? 😬

Can we supply a (sane) default then? Or at least add documentation around this?

albe:
👍 for better default, if the current default (not setting it) causes issues. If I read the message correctly, it states that since we don't set `sameSite` ("invalid value"?) we need to set `secure` - so maybe this line here needs to be changed: https://github.com/neos/flow-development-collection/pull/1948/files#diff-f54a6ca136b1699030b162872fe3be26R155 to include `|| $sameSite === null`?

And I really thought there was a documentation segment, but it seems there isn't :/ Need to be better with checking this before merging in the future

kdambekalns:
Well, `seecure` makes the cookie unavailable in a non-TLS context. In production this should not be an issue, but probably a lot of development setups miss encryption. Turning secure on based on `sameSite` seeems wrong, those are only very slightly related (even the alrady merged change is a bit too much magic IMHO). I'd rather turn secure on if encryption is detected, but even that would probably be too much.

A different default value OTOH is something that's fine with me.

sorenmalling:
The setting `secure` is "dependent" on the request object at hand, right?

albe:
So coming back to the "better default" : would that be "strict" or "lax"? "none" is out of the question, since it (currently) imposes "secure" as well.

kdambekalns:
Seeing that `Lax` is the default in browsers, we could use that, too.

PR #2275, opened 2020-11-24
!!! BUGFIX: Define default SAMESITE attribute to LAX

markusguenther:
The neos-ui complaining with warning in the modern browsers because our session cookie has no defined same site attribute and so the browser expect to have a same site with the lax value or none but with the secure attribute.

As the browsers use LAX as default we now also define that.
For mor information read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

Resolves: #2031

Screenshot 2020-11-24 at 10 31 02

**What I did**
Define `SameSite=Lax` when no sameSite is defined.

**How to verify it**
Just load the neos backend and check the dev console for warning. There should be no warning regarding session cookies.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Http/Cookie.php
+++ b/Neos.Flow/Classes/Http/Cookie.php
@@ -143,7 +143,7 @@ class Cookie
             throw new \InvalidArgumentException('The parameter "path" passed to the Cookie constructor must be a valid path as per RFC 6265, Section 4.1.1.', 1345123078);
         }
 
-        $sameSite = $sameSite === null ? self::SAMESITE_LAX : $sameSite;
+        $sameSite = $sameSite ?? self::SAMESITE_LAX;
         if (!\in_array($sameSite, [self::SAMESITE_LAX, self::SAMESITE_STRICT, self::SAMESITE_NONE], true)) {
             throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict" and "lax"', 1584955500);
         }
--- a/Neos.Flow/Classes/Http/Cookie.php
+++ b/Neos.Flow/Classes/Http/Cookie.php
@@ -143,7 +143,6 @@ class Cookie
             throw new \InvalidArgumentException('The parameter "path" passed to the Cookie constructor must be a valid path as per RFC 6265, Section 4.1.1.', 1345123078);
         }
 
-        $sameSite = $sameSite === null ? self::SAMESITE_LAX : $sameSite;
         if (!\in_array($sameSite, [self::SAMESITE_LAX, self::SAMESITE_STRICT, self::SAMESITE_NONE], true)) {
             throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict" and "lax"', 1606595726);
         }
--- a/Neos.Flow/Classes/Http/Cookie.php
+++ b/Neos.Flow/Classes/Http/Cookie.php
@@ -122,7 +122,7 @@ class Cookie
      * @api
      * @throws \InvalidArgumentException
      */
-    public function __construct($name, $value = null, $expires = 0, $maximumAge = null, $domain = null, $path = '/', $secure = false, $httpOnly = true, $sameSite = 'lax')
+    public function __construct($name, $value = null, $expires = 0, $maximumAge = null, $domain = null, $path = '/', $secure = false, $httpOnly = true, $sameSite = null)
     {
         if (preg_match(self::PATTERN_TOKEN, $name) !== 1) {
             throw new \InvalidArgumentException('The parameter "name" passed to the Cookie constructor must be a valid token as per RFC 2616, Section 2.2.', 1345101977);
@@ -143,8 +143,9 @@ class Cookie
             throw new \InvalidArgumentException('The parameter "path" passed to the Cookie constructor must be a valid path as per RFC 6265, Section 4.1.1.', 1345123078);
         }
 
+        $sameSite = $sameSite === null ? self::SAMESITE_LAX : $sameSite;
         if (!\in_array($sameSite, [self::SAMESITE_LAX, self::SAMESITE_STRICT, self::SAMESITE_NONE], true)) {
-            throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict" and "lax"', 1606595726);
+            throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict" and "lax"', 1584955500);
         }
 
         $this->name = $name;
--- a/Neos.Flow/Classes/Http/Cookie.php
+++ b/Neos.Flow/Classes/Http/Cookie.php
@@ -122,7 +122,7 @@ class Cookie
      * @api
      * @throws \InvalidArgumentException
      */
-    public function __construct($name, $value = null, $expires = 0, $maximumAge = null, $domain = null, $path = '/', $secure = false, $httpOnly = true, $sameSite = null)
+    public function __construct($name, $value = null, $expires = 0, $maximumAge = null, $domain = null, $path = '/', $secure = false, $httpOnly = true, $sameSite = 'lax')
     {
         if (preg_match(self::PATTERN_TOKEN, $name) !== 1) {
             throw new \InvalidArgumentException('The parameter "name" passed to the Cookie constructor must be a valid token as per RFC 2616, Section 2.2.', 1345101977);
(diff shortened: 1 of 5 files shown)


markusguenther:
Reopen to trigger travis again

markusguenther:
Ah of course need to adjust the unit test. Will take care.

jonnitto (approved):

albe (approved):
Guess we can indeed close #2031 with this :) thanks!

kdambekalns (dismissed):
There is documentation and default value in the settings file - that needs to be adjusted, since `null` is behaving differently now and lax should be the default in the settings.

I'd even go so far to disallow `null`: Do not acceept it in line 145, change the default for the `$sameSite` parameter from `null` to `self::SAMESITE_LAX`, fix the default in the settings.

markusguenther:
Ok will change that :)
Thanks Karsten 🙇

markusguenther:
@kdambekalns Hope did the right adjustments, but I did not find the sameSite stuff in the docs :/
https://flowframework.readthedocs.io/en/latest/TheDefinitiveGuide/PartIII/Http.html?highlight=cookie#cookies

Neos.Flow/Classes/Http/Cookie.php
```diff
@@ -142,8 +142,10 @@ public function __construct($name, $value = null, $expires = 0, $maximumAge = nu
         if ($path !== null && preg_match(self::PATTERN_PATH, $path) !== 1) {
             throw new \InvalidArgumentException('The parameter "path" passed to the Cookie constructor must be a valid path as per RFC 6265, Section 4.1.1.', 1345123078);
         }
-        if (!\in_array($sameSite, [self::SAMESITE_LAX, self::SAMESITE_STRICT, self::SAMESITE_NONE, null], true)) {
-            throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict", "lax" and null', 1584955500);
+
+        $sameSite = is_null($sameSite) ? self::SAMESITE_LAX : $sameSite;
```
kdambekalns, line 146: But why even allow null? It's optional with the default value now, and `null` doesn't need to be accepted, eh?
```suggestion
        $sameSite = $sameSite === null ? self::SAMESITE_LAX : $sameSite;
```

markusguenther: But if we change that the change will be breaking the behavior completely for this case ... now the will just get the new default cookie value but without null transformation the will always end in the exception.

Think that would be bit hard or not?

kdambekalns: But that would only happen if someone explicitly set it to `false`… because it's either set by the user or coming from the default settings. No?

markusguenther: Adjusted the tests now and removed the conversion. But when you have user land code that set sameSite to null in the construct it will break the behaviour with the exception.

But the code change is minimal but need to be done.

kdambekalns: True. What do the others think? @albe @robertlemke @jonnitto

Neos.Flow/Classes/Http/Cookie.php
```diff
@@ -122,7 +122,7 @@ class Cookie
      * @api
      * @throws \InvalidArgumentException
      */
-    public function __construct($name, $value = null, $expires = 0, $maximumAge = null, $domain = null, $path = '/', $secure = false, $httpOnly = true, $sameSite = null)
+    public function __construct($name, $value = null, $expires = 0, $maximumAge = null, $domain = null, $path = '/', $secure = false, $httpOnly = true, $sameSite = 'lax')
```
bwaidelich, line 125: Changing the method signature is a bit more breaking than it has to be and a "magic default value" it is considered bad practice anyways (see https://github.com/kalessil/phpinspectionsea/blob/master/docs/code-style.md#non-null-parameters-default-value)

bwaidelich: Instead of the change here, you could adjust line 158 to
```php
$this->sameSite = $sameSite ?? self::SAMESITE_LAX
```

Neos.Flow/Classes/Http/Cookie.php
```diff
@@ -142,8 +142,9 @@ public function __construct($name, $value = null, $expires = 0, $maximumAge = nu
         if ($path !== null && preg_match(self::PATTERN_PATH, $path) !== 1) {
             throw new \InvalidArgumentException('The parameter "path" passed to the Cookie constructor must be a valid path as per RFC 6265, Section 4.1.1.', 1345123078);
         }
-        if (!\in_array($sameSite, [self::SAMESITE_LAX, self::SAMESITE_STRICT, self::SAMESITE_NONE, null], true)) {
-            throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict", "lax" and null', 1584955500);
+
+        if (!\in_array($sameSite, [self::SAMESITE_LAX, self::SAMESITE_STRICT, self::SAMESITE_NONE], true)) {
+            throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict" and "lax"', 1606595726);
```
bwaidelich, line 147: The exception code should not be updated here
```suggestion
            throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict" and "lax"', 1584955500);
```

Neos.Flow/Tests/Unit/Http/CookieTest.php
```diff
@@ -305,7 +296,7 @@ public function SameSiteReturnsStrict()
      */
     public function SameSiteThrowsExceptionForInvalidValues()
     {
-        $this->expectExceptionCode(1584955500);
+        $this->expectExceptionCode(1606595726);
```
bwaidelich, line 299:
```suggestion
        $this->expectExceptionCode(1584955500);
```
see above

bwaidelich (commented):
Thank you for taking care!
This makes sense to me, I just left some nitpicky comments

Neos.Flow/Classes/Http/Cookie.php
markusguenther: Even If we change the text? Did that because it is not exactly the same message anymore.

Neos.Flow/Classes/Http/Cookie.php
albe: I'm not soooo sure about the "magic default value" argument.

At least I find
```php
    /* original code: logic is mixed in the signature, not possible to use nullable types */
    function calculateRadial($radius, $pi = M_PI, $multiplier = 1) {
        return $radius * $pi * $multiplier;
    }
```
easier to read an grasp on a single look than
```php
    /* code refactoring step 1 (triggered by the inspection) */
    function calculateRadial($radius, $pi = null, $multiplier = null) {
        return $radius * ($pi ?? M_PI) * ($multiplier ?? 1);
    }
 ```

The former IMO better separates the algorithm from the (default) values. Default values are part of the API, not of the implementation.

File: Neos.Flow/Classes/Http/Cookie.php (hunk 3 above)
Comment (bwaidelich) [reply Neos.Flow/Classes/Http/Cookie.php]: I'd say so, yes. It's still the same exception: You specified an invalid "sameSite" parameter value

File: Neos.Flow/Classes/Http/Cookie.php (hunk 2 above)
Comment (markusguenther) [reply Neos.Flow/Classes/Http/Cookie.php]: But in line 158 it makes no sense because you will end in the exception earlier ;)
I changed the exception code back and set `$sameSite` before the exception to `lax` if the `$sameSite` is `null`.

Comment (bwaidelich) [reply Neos.Flow/Classes/Http/Cookie.php]: I also find the reasoning & example a bit weird but I agree to the practice: We had quite a lot of regressions and needless breaking changes due to updates in method signatures that actually just reflected the _behavior_.

Anyways, feel free to ignore the remark in this case. We already use tons of optional arguments here and it's quite unlikely that someone extended the `Cookie` class

File: Neos.Flow/Classes/Http/Cookie.php [hunk 5]
```diff
@@ -142,8 +142,10 @@ public function __construct($name, $value = null, $expires = 0, $maximumAge = nu
         if ($path !== null && preg_match(self::PATTERN_PATH, $path) !== 1) {
             throw new \InvalidArgumentException('The parameter "path" passed to the Cookie constructor must be a valid path as per RFC 6265, Section 4.1.1.', 1345123078);
         }
-        if (!\in_array($sameSite, [self::SAMESITE_LAX, self::SAMESITE_STRICT, self::SAMESITE_NONE, null], true)) {
-            throw new \InvalidArgumentException('The parameter "sameSite" passed to the Cookie constructor must be a valid samesite value. Possible values are "none", "strict", "lax" and null', 1584955500);
+
+        $sameSite = $sameSite === null ? self::SAMESITE_LAX : $sameSite;
```
bwaidelich, line 146: Sorry, but one last nitpick:

```suggestion
        if ($sameSite === null) {
            $sameSite = self:: SAMESITE_LAX;
        }
```
or
```suggestion
        $sameSite = $sameSite ?? self::SAMESITE_LAX;
```
is much easier to comprehend IMO

bwaidelich (approved):
+1 by reading ❤️
just a last little nitpick, feel free to ignore that

kdambekalns:
Thanks everyone. Impressive how such a "small" change can lead to such a discussion! 🤣

Issue #2263, opened 2020-11-17
port UriConstraint is ignored when it's equal to the base port


bwaidelich:
### Description

The title is a bit cryptic, but this basically means that cross-domain linking is broken when source and target domain have the same custom port:

### Steps to Reproduce

```php
echo (string)UriConstraints::create()->withPort(8080)->applyTo(new Uri('http://localhost:8080'), true);
```

#### Expected behavior

```http://localhost:8080```

#### Actual behavior

```http://localhost/```

### Affected Versions

Flow: master

Related: #1126

PR #2276, opened 2020-11-26
BUGFIX: Fix UriConstraints::withPort() when port is equal to current port

bwaidelich:
This fixes the behavior of `UriConstraints` when using `withPort()` with
a (custom) port that is equal to the port of the applied URL:

```
UriConstraints::create()->withPort(8080)->applyTo(new Uri('http://localhost:8080'), true);
```

Now creates `http://localhost:8080` while it was `http://localhost` before.

Fixes: #2263

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
@@ -358,7 +358,7 @@ final class UriConstraints
                 $uri = $uri->withHost($host);
             }
         }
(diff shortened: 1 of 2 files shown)


albe (approved):

kdambekalns (commented):
Fine, but why only `master`? The issue says it affects:

> Flow: 4.3+ (supported: 5.3+)

bwaidelich:
> Fine, but why only master? The issue says it affects [Flow: 4.3+]

That was actually a mistake, only master seems to be affected.
I updated the ticket description accordingly, good catch!

Issue #2284, opened 2020-11-27
DoctrineMigrations directory is no longer auto-generated


bwaidelich:
### Description

Running a migration command (for example `./flow doctrine:migrationgenerate`) in a fresh installation leads to an exception.

### Steps to Reproduce

1. Install the Flow/Neos master
2. Run `./flow doctrine:migrationgenerate`

#### Expected behavior

The `Data/DoctrineMigrations` folder is created and the new migration is created in it

#### Actual behavior

Exception:
```
Migrations directory "[...]/Data/DoctrineMigrations" does not exist.
```

### Affected Versions

Flow: `master`

PR #2285, opened 2020-11-27
BUGFIX: Auto-generate DoctrineMigrations folder if needed

bwaidelich:
Fixes: #2284


kitsunet (approved):
Weird that htis is suddenly necessary, but why not...

kdambekalns (approved):

Issue #2233, opened 2020-11-10
Support for PHP 8


kdambekalns:
- Travis tests on PHP 8 since https://github.com/neos/flow-development-collection/pull/2015
- `doctrine/dbal` runs with PHP8 as of 2.12.0: https://github.com/doctrine/dbal/pull/4347
- `doctrine/orm` is on the way: https://github.com/doctrine/orm/pull/8303

kdambekalns:
`doctrine/commons` is raised to 2.13.1 in #1880 and supports PHP 8 as of that version.

kdambekalns:
`doctrine/dbal` is raised to 2.12 in #1880 and supports PHP 8 as of that version.

kdambekalns:
The Doctrine dependencies need to be raised in Neos, too, once #1880 is merged.

albe:
With #2287 existing code runs on PHP 8. However, to fully support PHP 8 we need to be able to deal with union types, constructor properties and annotations (at least for proper proxy building).

First step is maybe to add meta-info on union types to the reflection service.

albe:
PHP8 Attributes are now supported as of #2468 but #2404 is still WIP

bwaidelich:
@albe the related PRs have been merged, can this one be closed?

PR #2287, opened 2020-11-28
TASK: Add basic support for PHP8

albe:
This makes our code base work on PHP 8. It does not yet add support for PHP 8 features. See #2404 for this.

Related to #2233

### DIFF
--- a/Neos.Cache/Classes/Backend/FileBackend.php
+++ b/Neos.Cache/Classes/Backend/FileBackend.php
@@ -283,7 +283,9 @@ class FileBackend extends SimpleFileBackend implements PhpCapableBackendInterfac
         if ($this->frozen === true) {
             try {
                 @unlink($this->cacheDirectory . 'FrozenCache.data');
-            } catch (\Throwable $e) {}
+            } catch (\Throwable $e) {
+                // Double-shrug catching
+            }
             $this->frozen = false;
         }
     }
--- a/Neos.Cache/Classes/Backend/FileBackend.php
+++ b/Neos.Cache/Classes/Backend/FileBackend.php
@@ -281,7 +281,9 @@ class FileBackend extends SimpleFileBackend implements PhpCapableBackendInterfac
     {
         Files::emptyDirectoryRecursively($this->cacheDirectory);
         if ($this->frozen === true) {
-            @unlink($this->cacheDirectory . 'FrozenCache.data');
+            try {
+                @unlink($this->cacheDirectory . 'FrozenCache.data');
+            } catch (\Throwable $e) {}
             $this->frozen = false;
         }
     }
--- a/Neos.Cache/Classes/Backend/FileBackend.php
+++ b/Neos.Cache/Classes/Backend/FileBackend.php
@@ -284,7 +284,7 @@ class FileBackend extends SimpleFileBackend implements PhpCapableBackendInterfac
             try {
                 @unlink($this->cacheDirectory . 'FrozenCache.data');
             } catch (\Throwable $e) {
-                // Double-shrug catching
+                // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
             }
             $this->frozen = false;
         }
--- a/Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
+++ b/Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
@@ -572,7 +572,7 @@ class FileBackendTest extends BaseTestCase
         $backend->setCache($mockCache);
 
         $entryIdentifier = 'SomePhpEntryWithPhpWarning';
-        $backend->set($entryIdentifier, '<?php trigger_error("Warning!", E_WARNING); ?>');
+        $backend->set($entryIdentifier, '<?php trigger_error("Warning!", E_USER_WARNING); ?>');
         $backend->requireOnce($entryIdentifier);
     }
 
--- a/Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
+++ b/Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
@@ -589,7 +589,7 @@ class FileBackendTest extends BaseTestCase
         $backend->setCache($mockCache);
 
         $entryIdentifier = 'SomePhpEntryWithPhpNotice';
-        $backend->set($entryIdentifier, '<?php $undefined ++; ?>');
+        $backend->set($entryIdentifier, '<?php trigger_error("Notice!", E_USER_NOTICE); ?>');
         $backend->requireOnce($entryIdentifier);
     }
 
--- a/Neos.Cache/Tests/Unit/Backend/SimpleFileBackendTest.php
+++ b/Neos.Cache/Tests/Unit/Backend/SimpleFileBackendTest.php
@@ -405,7 +405,7 @@ class SimpleFileBackendTest extends BaseTestCase
         $entryIdentifier = 'SomePhpEntryWithPhpWarning';
 
         $simpleFileBackend = $this->getSimpleFileBackend();
-        $simpleFileBackend->set($entryIdentifier, '<?php trigger_error("Warning!", E_WARNING); ?>');
+        $simpleFileBackend->set($entryIdentifier, '<?php trigger_error("Warning!", E_USER_WARNING); ?>');
         $simpleFileBackend->requireOnce($entryIdentifier);
     }
 
--- a/Neos.Cache/Tests/Unit/Backend/SimpleFileBackendTest.php
+++ b/Neos.Cache/Tests/Unit/Backend/SimpleFileBackendTest.php
@@ -414,7 +414,11 @@ class SimpleFileBackendTest extends BaseTestCase
      */
     public function requireOnceDoesNotSwallowPhpNoticesOfTheIncludedFile()
     {
-        $this->expectNotice();
+        if (PHP_MAJOR_VERSION >= 8) {
+            $this->expectWarning();
+        } else {
+            $this->expectNotice();
+        }
         $entryIdentifier = 'SomePhpEntryWithPhpNotice';
 
         $simpleFileBackend = $this->getSimpleFileBackend();
--- a/Neos.Cache/Tests/Unit/Backend/SimpleFileBackendTest.php
+++ b/Neos.Cache/Tests/Unit/Backend/SimpleFileBackendTest.php
@@ -414,15 +414,11 @@ class SimpleFileBackendTest extends BaseTestCase
      */
     public function requireOnceDoesNotSwallowPhpNoticesOfTheIncludedFile()
     {
-        if (PHP_MAJOR_VERSION >= 8) {
-            $this->expectWarning();
-        } else {
-            $this->expectNotice();
-        }
+        $this->expectNotice();
         $entryIdentifier = 'SomePhpEntryWithPhpNotice';
 
         $simpleFileBackend = $this->getSimpleFileBackend();
-        $simpleFileBackend->set($entryIdentifier, '<?php $undefined ++; ?>');
+        $simpleFileBackend->set($entryIdentifier, '<?php trigger_error("Notice!", E_USER_NOTICE); ?>');
         $simpleFileBackend->requireOnce($entryIdentifier);
     }
 
--- a/Neos.Eel/Classes/Helper/MathHelper.php
+++ b/Neos.Eel/Classes/Helper/MathHelper.php
@@ -422,14 +422,14 @@ class MathHelper implements ProtectedContextAwareInterface
      */
     public function sign($x)
     {
-        if ($x < 0) {
+        if (!is_numeric($x)) {
+            return NAN;
+        } elseif ($x < 0) {
             return -1;
         } elseif ($x > 0) {
             return 1;
-        } elseif ($x === 0 || $x === 0.0) {
-            return 0;
         } else {
-            return NAN;
+            return 0;
         }
     }
 
--- a/Neos.Eel/Classes/Helper/MathHelper.php
+++ b/Neos.Eel/Classes/Helper/MathHelper.php
@@ -407,7 +407,7 @@ class MathHelper implements ProtectedContextAwareInterface
         if (!is_numeric($subject)) {
             return NAN;
         }
-        $subject = floatval($subject);
+        $subject = (float)$subject;
         if ($precision != null && !is_int($precision)) {
             return NAN;
         }
@@ -492,9 +492,9 @@ class MathHelper implements ProtectedContextAwareInterface
         $sign = $this->sign($x);
         switch ($sign) {
             case -1:
-                return (int)ceil($x);
+                return (int)ceil((float)$x);
             case 1:
-                return (int)floor($x);
+                return (int)floor((float)$x);
             default:
                 return $sign;
         }
--- a/Neos.Flow/Classes/Core/LockManager.php
+++ b/Neos.Flow/Classes/Core/LockManager.php
@@ -135,7 +135,9 @@ class LockManager
             fclose($this->lockResource);
             unlink($this->lockPathAndFilename);
         }
-        @unlink($this->lockFlagPathAndFilename);
+        if (file_exists($this->lockFlagPathAndFilename)) {
+            @unlink($this->lockFlagPathAndFilename);
+        }
     }
 
     /**
--- a/Neos.Flow/Classes/Core/LockManager.php
+++ b/Neos.Flow/Classes/Core/LockManager.php
@@ -79,8 +79,10 @@ class LockManager
         if (filemtime($this->lockFlagPathAndFilename) >= (time() - self::LOCKFILE_MAXIMUM_AGE)) {
             return;
         }
-        @unlink($this->lockFlagPathAndFilename);
-        @unlink($this->lockPathAndFilename);
+        try {
+            @unlink($this->lockFlagPathAndFilename);
+            @unlink($this->lockPathAndFilename);
+        } catch (\Throwable $e) {}
     }
 
     /**
@@ -135,8 +137,10 @@ class LockManager
             fclose($this->lockResource);
             unlink($this->lockPathAndFilename);
         }
-        if (file_exists($this->lockFlagPathAndFilename)) {
-            @unlink($this->lockFlagPathAndFilename);
+        if ($this->isSiteLocked()) {
+            try {
+                @unlink($this->lockFlagPathAndFilename);
+            } catch (\Throwable $e) {}
         }
     }
 
--- a/Neos.Flow/Classes/Core/LockManager.php
+++ b/Neos.Flow/Classes/Core/LockManager.php
@@ -82,7 +82,9 @@ class LockManager
         try {
             @unlink($this->lockFlagPathAndFilename);
             @unlink($this->lockPathAndFilename);
-        } catch (\Throwable $e) {}
+        } catch (\Throwable $e) {
+            // Double-shrug catching
+        }
     }
 
     /**
@@ -140,7 +142,9 @@ class LockManager
         if ($this->isSiteLocked()) {
             try {
                 @unlink($this->lockFlagPathAndFilename);
-            } catch (\Throwable $e) {}
+            } catch (\Throwable $e) {
+                // Double-shrug catching
+            }
         }
     }
 
--- a/Neos.Flow/Classes/Core/LockManager.php
+++ b/Neos.Flow/Classes/Core/LockManager.php
@@ -83,7 +83,7 @@ class LockManager
             @unlink($this->lockFlagPathAndFilename);
             @unlink($this->lockPathAndFilename);
         } catch (\Throwable $e) {
-            // Double-shrug catching
+            // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
         }
     }
 
@@ -143,7 +143,7 @@ class LockManager
             try {
                 @unlink($this->lockFlagPathAndFilename);
             } catch (\Throwable $e) {
-                // Double-shrug catching
+                // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
             }
         }
     }
--- a/Neos.Flow/Classes/Core/LockManager.php
+++ b/Neos.Flow/Classes/Core/LockManager.php
@@ -81,9 +81,12 @@ class LockManager
         }
         try {
             @unlink($this->lockFlagPathAndFilename);
-            @unlink($this->lockPathAndFilename);
         } catch (\Throwable $e) {
             // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
+        }
+        try {
+            @unlink($this->lockPathAndFilename);
+        } catch (\Throwable $e) {
         }
     }
 
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
@@ -510,7 +510,7 @@ class ConfigurationBuilder
             try {
                 $classMethodNames = get_class_methods($className);
             } catch (\TypeError $error) {
-                throw new UnknownClassException(sprintf('The class "%s" defined in the object configuration for object "%s", defined in package: %s, does not exist.', $className, $objectConfiguration->getObjectName(), $objectConfiguration->getPackageKey()), 1352371371);
+                throw new UnknownClassException(sprintf('The class "%s" defined in the object configuration for object "%s", defined in package: %s, does not exist.', $className, $objectConfiguration->getObjectName(), $objectConfiguration->getPackageKey()), 1352371372);
             }
             if (!is_array($classMethodNames)) {
                 if (!class_exists($className)) {
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
@@ -507,7 +507,11 @@ class ConfigurationBuilder
                 continue;
             }
 
-            $classMethodNames = get_class_methods($className);
+            try {
+                $classMethodNames = get_class_methods($className);
+            } catch (\TypeError $error) {
+                throw new UnknownClassException(sprintf('The class "%s" defined in the object configuration for object "%s", defined in package: %s, does not exist.', $className, $objectConfiguration->getObjectName(), $objectConfiguration->getPackageKey()), 1352371371);
+            }
             if (!is_array($classMethodNames)) {
                 if (!class_exists($className)) {
                     throw new UnknownClassException(sprintf('The class "%s" defined in the object configuration for object "%s", defined in package: %s, does not exist.', $className, $objectConfiguration->getObjectName(), $objectConfiguration->getPackageKey()), 1352371371);
--- a/Neos.Flow/Classes/Package/PackageManager.php
+++ b/Neos.Flow/Classes/Package/PackageManager.php
@@ -764,7 +764,9 @@ class PackageManager
         // Clean legacy file TODO: Remove at some point
         $legacyPackageStatesPath = FLOW_PATH_CONFIGURATION . 'PackageStates.php';
         if (is_file($legacyPackageStatesPath)) {
-            @unlink($legacyPackageStatesPath);
+            try {
+                @unlink($legacyPackageStatesPath);
+            } catch (\Throwable $e) {}
         }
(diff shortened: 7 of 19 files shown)


albe:
This depends on https://github.com/doctrine/migrations/pull/980 now

bwaidelich:
FYI: I converted this PR to a draft so that it doesn't get merged by accident (=> see doctrine/migrations#980)

albe:
The blocker has been solved by the doctrine team for christmas, even for 2.3.x versions :) see https://github.com/doctrine/migrations/pull/1102

But...
> Cloning into 'Packages/Framework'...                                         
  fatal: reference repository '/home/travis/.cache/composer/vcs/-home-travis-  
  build-neos-flow-development-collection/' is shallow   
 
 ❓ ❔ ❓ WTF Travis?

albe:
> 0.16s$ composer config --json repositories.flow.options '{"symlink": false}'
>                                                                               
>  [RuntimeException]                                                           
>  Failed to update composer.json with a valid format, reverting to the origin  
>  al content. Please report an issue to us with details (command you run and   
>  a copy of your composer.json).                                               
>                                                                               
>                                                             
>  [Composer\Json\JsonValidationException]                    
>  "./composer.json" does not match the expected JSON schema                                                             

Oh boi... 🤷‍♂️

albe:
Holy cow - there you go ✔️🎁

albe:
Oops, the PHP 8 tests actually fail - they just were allowed to fail because they were still "nightly" 🙈

albe:
Intermediary result: `Errors: 18, Failures: 16` -> `Errors: 7, Failures: 9`

composer.json
```diff
@@ -21,7 +21,7 @@
         "psr/http-server-middleware": "^1.0",
         "psr/http-server-handler": "^1.0",
         "ramsey/uuid": "^3.0 || ^4.0",
-        "doctrine/orm": "^2.7",
+        "doctrine/orm": "^2.7 || 2.8.x-dev",
```
albe, line 24:
```suggestion
        "doctrine/orm": "^2.7",
```
2.8 has been released and is now covered with `^2.7` - see https://semver.mwl.be/?package=doctrine%2Form&constraint=%5E2.7&stability=stable

albe:
PHP8 currently compiles the return typehint of `SessionDataContainer::getInterceptedRequest()` to `public function getInterceptedRequest() : ?\?Neos\Flow\Mvc\ActionRequest` which then fails with a syntax error.

albe:
Now only Behat fails...
> Created database schema.
> ...............---..........................................
> 
> --- Failed hooks:
> 
>     BeforeScenario @fixtures # FeatureContext::resetTestFixtures()
>       There is no active transaction (PDOException)

Likely related to https://github.com/doctrine/migrations/issues/1104

Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
```diff
@@ -589,7 +589,7 @@ public function requireOnceDoesNotSwallowPhpNoticesOfTheIncludedFile()
         $backend->setCache($mockCache);
 
         $entryIdentifier = 'SomePhpEntryWithPhpNotice';
-        $backend->set($entryIdentifier, '<?php $undefined ++; ?>');
```
albe, line 592: This is no longer a E_NOTICE in PHP8

Neos.Cache/Tests/Unit/Backend/SimpleFileBackendTest.php
```diff
@@ -405,7 +405,7 @@ public function requireOnceDoesNotSwallowPhpWarningsOfTheIncludedFile()
         $entryIdentifier = 'SomePhpEntryWithPhpWarning';
 
         $simpleFileBackend = $this->getSimpleFileBackend();
-        $simpleFileBackend->set($entryIdentifier, '<?php trigger_error("Warning!", E_WARNING); ?>');
+        $simpleFileBackend->set($entryIdentifier, '<?php trigger_error("Warning!", E_USER_WARNING); ?>');
```
albe, line 408: Only `E_USER_X` can be manually triggered. PHP 7 was lenient with this though.

Neos.Flow/Classes/Reflection/MethodReflection.php
```diff
@@ -95,15 +95,15 @@ public function getDescription()
...
+     * @return string|null The name of a type (e.g. string, \stdClass) if it was declared as a return type, null otherwise
      */
     public function getDeclaredReturnType()
     {
         if (!is_callable([$this, 'getReturnType'])) {
             return null;
         }
         $type = $this->getReturnType();
-        return $type !== null ? (string)$type : null;
+        return $type !== null ? ltrim((string)$type, '?') : null;
```
albe, line 106: PHP8 returns the nullable type operator

Neos.Flow/Classes/Reflection/ParameterReflection.php
```diff
@@ -52,14 +52,14 @@ public function getClass()
...
 
     /**
-     * @return string The name of a builtin type (e.g. string, int) if it was declared for the parameter (scalar type declaration), null otherwise
+     * @return string|null The name of a builtin type (e.g. string, int) if it was declared for the parameter (scalar type declaration), null otherwise
      */
     public function getBuiltinType()
     {
         $type = $this->getType();
-        if ($type === null || !$type->isBuiltin()) {
+        if (!$type instanceof \ReflectionNamedType) {
```
albe, line 60: Could be a `ReflectionUnionType` in PHP 8

Neos.Flow/Classes/SignalSlot/Dispatcher.php
```diff
@@ -149,7 +149,7 @@ public function dispatch(string $signalClassName, string $signalName, array $sig
         }
 
         foreach ($this->slots[$signalClassName][$signalName] as $slotInformation) {
-            $finalSignalArguments = $signalArguments;
+            $finalSignalArguments = array_values($signalArguments);
```
albe, line 152: PHP8 named parameters...

Neos.Flow/Classes/Utility/PhpAnalyzer.php
```diff
@@ -82,8 +82,14 @@ public function extractNamespace(): ?string
                     break;
                 }
                 list($type, $value) = $token;
+                if (defined('T_NAME_QUALIFIED') && $type === T_NAME_QUALIFIED) {
```
albe, line 85: Tokenization of namespaces changed in PHP8 and now provides `T_NAME_(FULLY_)QUALIFIED` instead of a couple of `T_STRING` tokens.

Neos.Flow/Tests/Functional/Mvc/ActionControllerTest.php
```diff
@@ -429,7 +429,6 @@ public function argumentTestsDataProvider()
             'required date string'              => ['requiredDate', '1980-12-13T14:22:12+02:00', '1980-12-13'],
             'required date - missing value'     => ['requiredDate', null, $requiredArgumentExceptionText],
             'required date - mapping error'     => ['requiredDate', 'no date', 'Validation failed while trying to call Neos\Flow\Tests\Functional\Mvc\Fixtures\Controller\ActionControllerTestBController->requiredDateAction().'],
-            'required date - empty value'       => ['requiredDate', '', 'Uncaught Exception in Flow Argument 1 passed to Neos\Flow\Tests\Functional\Mvc\Fixtures\Controller\ActionControllerTestBController_Original::requiredDateAction() must be an instance of DateTime, null given'],
```
albe, line 432: The exception text changed in PHP8, so needs to be tested depending on PHP version => see below

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1788,9 +1788,6 @@ protected function convertParameterReflectionToArray(ParameterReflection $parame
         if ($parameter->isPassedByReference()) {
             $parameterInformation[self::DATA_PARAMETER_BY_REFERENCE] = true;
         }
-        if ($parameter->isArray()) {
```
albe, line 1791: `isArray()` is deprecated in PHP 8 - we should just handle array like any other builtin type

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1800,11 +1797,20 @@ protected function convertParameterReflectionToArray(ParameterReflection $parame
 
         $parameterType = $parameter->getType();
         if ($parameterType !== null) {
+            // TODO: This needs to handle ReflectionUnionType
             $parameterType = ($parameterType instanceof \ReflectionNamedType) ? $parameterType->getName() : $parameterType->__toString();
         }
-        if ($parameter->getClass() !== null) {
+        if ($parameterType !== null && !TypeHandling::isSimpleType($parameterType)) {
```
albe, line 1803: `isSimpleType()` is `true` for `array`

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1800,11 +1797,20 @@ protected function convertParameterReflectionToArray(ParameterReflection $parame
...
             // We use parameter type here to make class_alias usage work and return the hinted class name instead of the alias
             $parameterInformation[self::DATA_PARAMETER_CLASS] = $parameterType;
+        } elseif ($parameterType === 'array') {
+            $parameterInformation[self::DATA_PARAMETER_ARRAY] = true;
```
albe, line 1807:
```suggestion
            $parameterInformation[self::DATA_PARAMETER_ARRAY] = true;
            $parameterInformation[self::DATA_PARAMETER_TYPE] = $parameterType;
```
Maybe this is missing to stay fully b/c - was previously set below with "$parameter->isArray() ? 'array' : 'mixed'"

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1800,11 +1797,20 @@ protected function convertParameterReflectionToArray(ParameterReflection $parame
...
         $parameterType = $parameter->getType();
         if ($parameterType !== null) {
+            // TODO: This needs to handle ReflectionUnionType
             $parameterType = ($parameterType instanceof \ReflectionNamedType) ? $parameterType->getName() : $parameterType->__toString();
         }
-        if ($parameter->getClass() !== null) {
+        if ($parameterType !== null && !TypeHandling::isSimpleType($parameterType)) {
             // We use parameter type here to make class_alias usage work and return the hinted class name instead of the alias
             $parameterInformation[self::DATA_PARAMETER_CLASS] = $parameterType;
+        } elseif ($parameterType === 'array') {
```
albe, line 1806: Else here makes sure that `array` is not marked as `DATA_PARAMETER_SCALAR_DECLARATION`

Neos.Flow/Classes/Utility/PhpAnalyzer.php
```diff
@@ -82,8 +82,14 @@ public function extractNamespace(): ?string
...
+                    return $value;
+                }
+                if (defined('T_NAME_FULLY_QUALIFIED') && $type === T_NAME_FULLY_QUALIFIED) {
+                    return ltrim($value, '\\');
+                }
                 if ($type === T_STRING) {
-                    $namespaceParts[] = $value;
+                    $namespaceParts[] = ltrim($value, '\\');
```
albe, line 92: Not 100% sure any more this is required, because in PHP8 namespace is parsed with above tokens. But we always expect namespaces to not start with `\` and this guarantees it.

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
```diff
@@ -144,7 +144,7 @@ public function dispatchPassesTheSignalArgumentsToTheSlotMethod(): void
         $dispatcher->connect('Foo', 'bar', $mockSlot, '', false);
         $dispatcher->injectObjectManager($mockObjectManager);
 
-        $dispatcher->dispatch('Foo', 'bar', ['foo' => 'bar', 'baz' => 'quux']);
+        $dispatcher->dispatch('Foo', 'bar', ['bar', 'quux']);
```
albe, line 147: PHP 8 named parameters...

Neos.Eel/Classes/Helper/MathHelper.php
```diff
@@ -422,14 +422,14 @@ public function round($subject, $precision = 0)
      */
     public function sign($x)
     {
-        if ($x < 0) {
+        if (!is_numeric($x)) {
```
albe, line 425: In PHP8 'foo' compares differently to a number: see https://3v4l.org/Zhdsh
It is more sane to first check the non-numeric case and return NAN anyway, as it is done similarly in other math helper functions.

Neos.Cache/Classes/Backend/FileBackend.php
```diff
@@ -281,7 +281,11 @@ public function flush(): void
     {
         Files::emptyDirectoryRecursively($this->cacheDirectory);
         if ($this->frozen === true) {
-            @unlink($this->cacheDirectory . 'FrozenCache.data');
+            try {
+                @unlink($this->cacheDirectory . 'FrozenCache.data');
+            } catch (\Throwable $e) {
+                // Double-shrug catching
```
albe, line 287: Yeah, you might wonder about this bad practice. PHP 8 apparently now throws for unlink even with shutup operator, but we really don't care at those places. It's also the only way to handle this race-condition free.

albe (commented):
Added some comments for easier review

albe:
> Now only Behat fails...
> 
> > Created database schema.
> > ...............---..........................................
> > --- Failed hooks:
> > ```
> > BeforeScenario @fixtures # FeatureContext::resetTestFixtures()
> >   There is no active transaction (PDOException)
> > ```
> 
> Likely related to [doctrine/migrations#1104](https://github.com/doctrine/migrations/issues/1104)
> 
> Edit: Fixed with [neos/behat@f68dea2](https://github.com/neos/behat/commit/f68dea268d5e42cac57a9dd713859cb0d17558a6) and all green ✔️ - the exception basically just states that the transaction around the migration was already committed. _BUT_ that is only a broken fix, because effectively migrations will stop at the first migration containing DDL. The right solution would be to catch and ignore the exception around every single migration, which we can not do and needs to be done by doctrine/migrations. The other alternative is to change all existing mysql Migrations to include `isTransactional() { return false; }`
> See also [doctrine/migrations#1104 (comment)](https://github.com/doctrine/migrations/issues/1104#issuecomment-786311346)

Will be resolved once https://github.com/doctrine/migrations/pull/1131 is released. It's merged into 2.3.x-dev already, so won't take too long I guess. I would still keep the odd semi-fix of neos/behat@f68dea2, just in case.

Neos.Cache/Classes/Backend/FileBackend.php
bwaidelich: Is "double-shrug" a catchphrase? I've never heard of it and maybe we should just make your Github comment an inline comment instead :)

bwaidelich (approved):
+1 (by reading). Just a question/suggestion re "double-shrug" comments :)

Neos.Cache/Classes/Backend/FileBackend.php
albe: Not really, just read it in a blog post about PHP 8 here https://php.watch/versions/8.0#fatal-error-suppression and found it somewhat amusing
image

But 👍 for making this more helpful

albe:
```suggestion
                // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
```

Neos.Flow/Classes/Core/LockManager.php
```diff
@@ -79,8 +79,12 @@ protected function removeExpiredLock()
         if (filemtime($this->lockFlagPathAndFilename) >= (time() - self::LOCKFILE_MAXIMUM_AGE)) {
             return;
         }
-        @unlink($this->lockFlagPathAndFilename);
-        @unlink($this->lockPathAndFilename);
+        try {
+            @unlink($this->lockFlagPathAndFilename);
+            @unlink($this->lockPathAndFilename);
+        } catch (\Throwable $e) {
+            // Double-shrug catching
```
albe, line 86:
```suggestion
            // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
```

Neos.Flow/Classes/Core/LockManager.php
```diff
@@ -135,7 +139,13 @@ public function unlockSite()
             fclose($this->lockResource);
             unlink($this->lockPathAndFilename);
         }
-        @unlink($this->lockFlagPathAndFilename);
+        if ($this->isSiteLocked()) {
+            try {
+                @unlink($this->lockFlagPathAndFilename);
+            } catch (\Throwable $e) {
+                // Double-shrug catching
```
albe, line 146:
```suggestion
                // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
```

Neos.Flow/Classes/Package/PackageManager.php
```diff
@@ -764,7 +764,11 @@ protected function savePackageStates(array $orderedPackageStates): void
         // Clean legacy file TODO: Remove at some point
         $legacyPackageStatesPath = FLOW_PATH_CONFIGURATION . 'PackageStates.php';
         if (is_file($legacyPackageStatesPath)) {
-            @unlink($legacyPackageStatesPath);
+            try {
+                @unlink($legacyPackageStatesPath);
+            } catch (\Throwable $e) {
+                // Double-shrug catching
```
albe, line 770:
```suggestion
                // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
```

Neos.Utility.Files/Classes/Files.php
```diff
@@ -190,7 +190,11 @@ public static function removeEmptyDirectoriesOnPath(string $path, string $basePa
                 break;
             }
             if (file_exists($path . '/.DS_Store')) {
-                @unlink($path . '/.DS_Store');
+                try {
+                    @unlink($path . '/.DS_Store');
+                } catch (\Throwable $e) {
+                    // Double-shrug catching
```
albe, line 196:
```suggestion
                    // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
```

Neos.Cache/Classes/Backend/FileBackend.php
bwaidelich, line 287: Thanks for the explanation, so what about:
```suggestion
                // 🤷🏻‍♀️🤷🏼‍♂️
```
haha. Just kidding, I like your suggestion!

Neos.Flow/Classes/Core/LockManager.php
```diff
@@ -79,8 +79,12 @@ protected function removeExpiredLock()
...
             return;
         }
-        @unlink($this->lockFlagPathAndFilename);
-        @unlink($this->lockPathAndFilename);
+        try {
+            @unlink($this->lockFlagPathAndFilename);
+            @unlink($this->lockPathAndFilename);
+        } catch (\Throwable $e) {
+            // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
+        }
```
albe, line 87:
```suggestion
        try {
            @unlink($this->lockFlagPathAndFilename);
        } catch (\Throwable $e) {
            // PHP 8 apparently throws for unlink even with shutup operator, but we really don't care at this place. It's also the only way to handle this race-condition free.
        }
        try {
            @unlink($this->lockPathAndFilename);
        } catch (\Throwable $e) {
        }
```

Just noticed we need two trys because otherwise the second file might not get deleted.

albe:
https://twitter.com/paxuclus/status/1373351604930412547 - still need a second approval :)

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -507,7 +507,11 @@ protected function autowireProperties(array &$objectConfigurations)
                 continue;
             }
 
-            $classMethodNames = get_class_methods($className);
+            try {
+                $classMethodNames = get_class_methods($className);
+            } catch (\TypeError $error) {
+                throw new UnknownClassException(sprintf('The class "%s" defined in the object configuration for object "%s", defined in package: %s, does not exist.', $className, $objectConfiguration->getObjectName(), $objectConfiguration->getPackageKey()), 1352371371);
```
kdambekalns, line 513: This should get a new exceptipn code…

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
kdambekalns: This reflects the added `array_values()` in `Dispatcher`? But why, that call would do exactly that. This test change shows a change in the API, which did not really happen…

Neos.Flow/Classes/Reflection/ReflectionService.php
kdambekalns: Yes, please.

kdambekalns (changes requested):
👏 Some minor tweaks, please.

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
albe: Hm, it is the same cause, just in PHP8-case - but maybe it's not breaking to have a slightly different exception code? Not 100% sure
```suggestion
                throw new UnknownClassException(sprintf('The class "%s" defined in the object configuration for object "%s", defined in package: %s, does not exist.', $className, $objectConfiguration->getObjectName(), $objectConfiguration->getPackageKey()), 1352371372);
```

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
albe: Hm, indeed - I guess that was a change from before the `array_values()` was added?
```suggestion
        $dispatcher->dispatch('Foo', 'bar', ['foo' => 'bar', 'baz' => 'quux']);
```

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
```diff
@@ -160,7 +160,7 @@ public function dispatchPassesTheSignalArgumentsToTheStaticSlotMethod(): void
         $dispatcher->connect('Foo', 'bar', get_class($this), '::staticSlot', false);
         $dispatcher->injectObjectManager($mockObjectManager);
 
-        $dispatcher->dispatch('Foo', 'bar', ['foo' => 'bar', 'baz' => 'quux']);
+        $dispatcher->dispatch('Foo', 'bar', ['bar', 'quux']);
```
albe, line 163:
```suggestion
        $dispatcher->dispatch('Foo', 'bar', ['foo' => 'bar', 'baz' => 'quux']);
```

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
```diff
@@ -210,7 +210,7 @@ public function dispatchRetrievesSlotInstanceFromTheObjectManagerIfOnlyAClassNam
         $dispatcher->injectObjectManager($mockObjectManager);
         $dispatcher->connect('Foo', 'bar', $slotClassName, 'slot', false);
 
-        $dispatcher->dispatch('Foo', 'bar', ['foo' => 'bar', 'baz' => 'quux']);
+        $dispatcher->dispatch('Foo', 'bar', ['bar', 'quux']);
```
albe, line 213:
```suggestion
        $dispatcher->dispatch('Foo', 'bar', ['foo' => 'bar', 'baz' => 'quux']);
```

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
```diff
@@ -247,7 +247,7 @@ public function dispatchThrowsAnExceptionIfTheSpecifiedSlotMethodDoesNotExist():
         $dispatcher->injectObjectManager($mockObjectManager);
         $dispatcher->connect('Foo', 'bar', $slotClassName, 'unknownMethodName', true);
 
-        $dispatcher->dispatch('Foo', 'bar', ['foo' => 'bar', 'baz' => 'quux']);
+        $dispatcher->dispatch('Foo', 'bar', ['bar', 'quux']);
```
albe, line 250:
```suggestion
        $dispatcher->dispatch('Foo', 'bar', ['foo' => 'bar', 'baz' => 'quux']);
```

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
```diff
@@ -266,7 +266,7 @@ public function dispatchPassesArgumentContainingSlotInformationLastIfTheConnecti
         $dispatcher->connect('SignalClassName', 'methodName', $mockSlot, '', true);
         $dispatcher->injectObjectManager($mockObjectManager);
 
-        $dispatcher->dispatch('SignalClassName', 'methodName', ['foo' => 'bar', 'baz' => 'quux']);
+        $dispatcher->dispatch('SignalClassName', 'methodName', ['bar', 'quux']);
```
albe, line 269:
```suggestion
        $dispatcher->dispatch('SignalClassName', 'methodName', ['foo' => 'bar', 'baz' => 'quux']);
```

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
```diff
@@ -172,8 +172,8 @@ public function dispatchPassesTheSignalArgumentsToTheStaticSlotMethodIfNoObjectm
         $dispatcher = new Dispatcher();
         $dispatcher->connect('Foo', 'bar', get_class($this), '::staticSlot', false);
 
-        $dispatcher->dispatch('Foo', 'bar', ['no' => 'object', 'manager' => 'exists']);
-        self::assertSame(['object', 'exists'], self::$arguments);
+        $dispatcher->dispatch('Foo', 'bar', ['no', 'object', 'manager', 'exists']);
+        self::assertSame(['no', 'object', 'manager', 'exists'], self::$arguments);
```
albe, line 176:
```suggestion
        $dispatcher->dispatch('Foo', 'bar', ['no' => 'object', 'manager' => 'exists']);
        self::assertSame(['object', 'exists'], self::$arguments);
```

Neos.Flow/Tests/Unit/SignalSlot/DispatcherTest.php
albe: I instead added a new test to cover that unnamed parameters also work.

albe:
Thanks @kdambekalns ! All valid feedback and applied :)

kdambekalns (approved):
❤️

albe:
👿 were are those unrelated test failures coming from now? 😫

kdambekalns:
> 👿 were are those unrelated test failures coming from now? 😫

Maybe they are not enitrely unrelated? Could be from https://github.com/neos/flow-development-collection/pull/2287/commits/23f4bc33802b55dd84461eff4b36afe414ed1e87#diff-7d0a1ee9681327fd82562aa2d7fcd5ed3b083a2283085f8c760eb4fab0e8c73aR1808, no?

albe:
> > 👿 were are those unrelated test failures coming from now? 😫
> 
> Maybe they are not enitrely unrelated? Could be from [23f4bc3#diff-7d0a1ee9681327fd82562aa2d7fcd5ed3b083a2283085f8c760eb4fab0e8c73aR1808](https://github.com/neos/flow-development-collection/commit/23f4bc33802b55dd84461eff4b36afe414ed1e87#diff-7d0a1ee9681327fd82562aa2d7fcd5ed3b083a2283085f8c760eb4fab0e8c73aR1808), no?

🤔 ...

Neos.Flow/Classes/Reflection/ReflectionService.php
albe: Okay, I get why this breaks now: We set $parameterType here, where it is just 'array'. But we need to actually use the expanded parameter type information from the annotation below to keep the annotated element type:
`$parameterType = $this->expandType($method->getDeclaringClass(), $explodedParameters[0]);`

The DATA_PARAMETER_TYPE will then be set with the `if (!isset($parameterInformation[self::DATA_PARAMETER_TYPE]) && $parameterType !== null) {` condition

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1800,11 +1797,21 @@ protected function convertParameterReflectionToArray(ParameterReflection $parame
...
+            $parameterInformation[self::DATA_PARAMETER_ARRAY] = true;
+            $parameterInformation[self::DATA_PARAMETER_TYPE] = $parameterType;
```
albe, line 1808:
```suggestion
```
Needs to be set after extracting @var annotations with element type hints

albe:
There you go :)

PR #2280, opened 2020-11-27
TASK: Update PHP CGL & documentation examples to current practice

bwaidelich:
Resolves: #2279

PR #2288, opened 2020-11-29
!!! TASK: Make composer autoloader the default

albe:
The old behaviour can now still be achieved by setting `FLOW_ONLY_COMPOSER_LOADER=false`, but is (still) deprecated and will be gone at some point.

Related to #2262


Neos.Flow/Classes/Core/Bootstrap.php
```diff
@@ -540,11 +540,11 @@ protected function defineConstants()
             define('FLOW_PATH_TEMPORARY', $temporaryDirectoryPath);
         }
 
-        // Not using this flag and loading classes via our class loader is deprecated
-        // and we will remove class loading from our loader in the next major of Flow (5.0).
-        $onlyUseComposerAutoLoaderForPackageClasses = false;
-        if (in_array(self::getEnvironmentConfigurationSetting('FLOW_ONLY_COMPOSER_LOADER'), [true, 'true', 1, '1'])) {
-            $onlyUseComposerAutoLoaderForPackageClasses = true;
+        // Using this flag and loading classes via our class loader is deprecated
+        // and we will remove class loading from our loader in the next major of Flow (8.0).
```
bwaidelich, line 544: Good catch! I missed this one because I was searching for `@deprecated` annotations. I think we should add that.
Also in most places we now specify in which version the code became deprecated instead of trying to foresee when it will be removed.
I think this makes sense. Maybe something like:

```suggestion
        // Setting this flag to false will enable the custom Class Loader on top of the default autoloading provided by composer
        // @deprecated sind Version 4.3. Packages should use the default composer autoloading mechanism
```

bwaidelich (commented):

I can't really judge if our custom ClassLoader can be disabled without further side effects. Probably @kitsunet can tell that. Or we just test it :)

bwaidelich:
If we agree to this, we should also update the docs (see https://github.com/neos/flow-development-collection/pull/2280#discussion_r532221092)

Neos.Flow/Classes/Core/Bootstrap.php
```diff
@@ -540,11 +540,11 @@ protected function defineConstants()
             define('FLOW_PATH_TEMPORARY', $temporaryDirectoryPath);
         }
 
-        // Not using this flag and loading classes via our class loader is deprecated
-        // and we will remove class loading from our loader in the next major of Flow (5.0).
-        $onlyUseComposerAutoLoaderForPackageClasses = false;
-        if (in_array(self::getEnvironmentConfigurationSetting('FLOW_ONLY_COMPOSER_LOADER'), [true, 'true', 1, '1'])) {
-            $onlyUseComposerAutoLoaderForPackageClasses = true;
+        // Setting this flag to false will enable the custom Class Loader on top of the default autoloading provided by composer
+        // @deprecated sind Version 4.3. Packages should use the default composer autoloading mechanism
```
albe, line 544:
```suggestion
        // @deprecated since Version 4.3. Packages should use the default composer autoloading mechanism
```

bwaidelich (approved):
+1 by reading (and Neos demo seems to work with it) but I would like another opinion on that since I never really understood the differences of our own ClassLoader tbh

bwaidelich:
Note: I just realized that `./flow kickstart:site` (and probably also `./flow kickstart:package`) doesn't actually add the new package to the composer manifest. I could imagine that this will break without our own ClassLoader, right?

albe:
Good point. That's definitely something to check.

kdambekalns:
> I could imagine that this will break without our own ClassLoader, right?

That is fixed by now (see https://github.com/neos/neos-development-distribution/pull/59)

kdambekalns (approved):
Fine with me.

kitsunet (approved):

albe:
🤞

PR #2297, opened 2020-12-02
BUGFIX: Correct printed path for created doctrine migrations

### LINKED ISSUE #2296: Packages prefix is not printed as part of moved migration path
Once you create a migration and move it to one of your own packages, the following path is being printed

`The migration was moved to: Application/Project.Data/Migrations/Mysql/Version20201202082829.php`

You can not simply double click and copy/paste, since the `Packages` folder is missing

sorenmalling:
Tweaks the output of the `doctrine:migrationgenerate` command so that it renders the path of the new migration relative to the root directory.

Before:

    The migration was moved to: Application/<Package.Key>/Migrations/<DB-Type>/Version<Version>.php

Now:

    The migration was moved to: Packages/Application/<Package.Key>/Migrations/<DB-Type>/Version<Version>.php

Fixes #2296

### DIFF
--- a/Neos.Flow/Classes/Command/DoctrineCommandController.php
+++ b/Neos.Flow/Classes/Command/DoctrineCommandController.php
@@ -496,7 +496,7 @@ class DoctrineCommandController extends CommandController
             if ($selectedPackage !== $choices[0]) {
                 /** @var Package $selectedPackage */
                 $selectedPackage = $packages[$selectedPackage];
-                $targetPathAndFilename = Files::concatenatePaths([$selectedPackage->getPackagePath(), 'Migrations', $this->doctrineService->getDatabasePlatformName(), basename($migrationClassPathAndFilename)]);
+                $targetPathAndFilename = Files::concatenatePaths(['Packages', $selectedPackage->getPackagePath(), 'Migrations', $this->doctrineService->getDatabasePlatformName(), basename($migrationClassPathAndFilename)]);
                 Files::createDirectoryRecursively(dirname($targetPathAndFilename));
                 rename($migrationClassPathAndFilename, $targetPathAndFilename);
                 $this->outputLine('The migration was moved to: <comment>%s</comment>', [substr($targetPathAndFilename, strlen(FLOW_PATH_PACKAGES))]);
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Command/DoctrineCommandController.php
```diff
@@ -496,7 +496,7 @@ public function migrationGenerateCommand(bool $diffAgainstCurrent = true, string
             if ($selectedPackage !== $choices[0]) {
                 /** @var Package $selectedPackage */
                 $selectedPackage = $packages[$selectedPackage];
-                $targetPathAndFilename = Files::concatenatePaths([$selectedPackage->getPackagePath(), 'Migrations', $this->doctrineService->getDatabasePlatformName(), basename($migrationClassPathAndFilename)]);
+                $targetPathAndFilename = Files::concatenatePaths(['Packages', $selectedPackage->getPackagePath(), 'Migrations', $this->doctrineService->getDatabasePlatformName(), basename($migrationClassPathAndFilename)]);
```
bwaidelich, line 499: I don't think that the path was wrong here. Otherwise the next line would always create a wrong directory..
I guess that the bug is in line 502

bwaidelich (changes requested):
@sorenmalling Thanks for this!
I took the freedom to mark this as "BUGFIX" and I think it should be created against the lowest supported branch (5.3).

I assume that that the actual bug is in the output string not in the path itself (see comments)

sorenmalling:
> it should be created against the lowest supported branch (5.3).

I create it against the 5.3 branch and it will automatically be applied to all never branches?

bwaidelich:
> I create it against the 5.3 branch and it will automatically be applied to all never branches?

Not automatically, but we do upmerges from time to time, so you won't have to take care of that

Neos.Flow/Classes/Command/DoctrineCommandController.php
kdambekalns: What @bwaidelich says!

kdambekalns (changes requested):

bwaidelich:
I just tested this patch and it leads to an output of 
```
The migration was moved to: ackages/Sites/My.Package/Migrations/Mysql/Version20201207083419.php
```
and it created a new folder `<absolute-path>/Packages/Sites/My.Package/Migrations/Mysql` underneath the `Package` directory.. I'll push a follow-up

bwaidelich (approved):
This works for me. @kdambekalns could you remove the -1 please?

kdambekalns (approved):

PR #2306, opened 2020-12-02
TASK: Throw exception if HTTP chain is configured

bwaidelich:
Adjusts the default HTTP `RequestHandler` so that it throws
an exception if the (no longer supported) `Neos.Flow.http.chain`
is configured:

    Exception #1606912674 in line 104 of Packages/Framework/Neos.Flow/Classes/Http/RequestHandler.php:
    The settings still contain configuration for the no longer supported HTTP component chain.
    The component chain has been replaced in favor of PSR-15 middlewares with Flow version 7.0,
    please make sure to migrate any components accordingly.
    To see which components are still configured, run ./flow configuration:show --path Neos.Flow.http.chain

This will hopefully prevent some subtle bugs due to some components
that no longer work as expected.

Related: #2258

### DIFF
--- a/Neos.Flow/Classes/Http/RequestHandler.php
+++ b/Neos.Flow/Classes/Http/RequestHandler.php
@@ -13,7 +13,9 @@ namespace Neos\Flow\Http;
 
 use GuzzleHttp\Psr7\ServerRequest;
 use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Configuration\ConfigurationManager;
 use Neos\Flow\Core\Bootstrap;
+use Neos\Flow\Exception as FlowException;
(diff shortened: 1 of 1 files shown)


bwaidelich:
tbh: I'm not sure if this is really worth it since with the current state compilation will fail for any existing class implementing the `ComponentInterface`..

bwaidelich:
I just realized: This might still be useful if
* you configured a 3rd party component
* disabled one of the core components

albe (approved):

bwaidelich:
As discussed with @albe:
I think we should rather detect this case via core migration (or maybe even during compilation).
I'll give it a try tomorrow

PR #2307, opened 2020-12-02
FEATURE: Middleware CLI command

bwaidelich:
Adds a CLI command `middleware:list` that can be used to list
all configured middleware components in the order they will be
executed.

Related: #2258


markusguenther (approved):
Screenshot 2020-12-03 at 09 19 06

Looks pretty good 👍

albe (approved):
Awesome! Super straightforward and helpful

PR #2311, opened 2020-12-04
TASK: Show a helpful message when a class still implements ComponentInterface

albe:
Related to #2258

### DIFF
--- a/Neos.Flow/Classes/Reflection/Exception/ClassLoadingForReflectionFailedException.php
+++ b/Neos.Flow/Classes/Reflection/Exception/ClassLoadingForReflectionFailedException.php
@@ -20,13 +20,15 @@ class ClassLoadingForReflectionFailedException extends \Neos\Flow\Reflection\Exc
 {
     protected $className;
 
-    public static function forClassName($className): self {
+    public static function forClassName(string $className): self
+    {
         $exception = new self(sprintf('Required class "%s" could not be loaded properly for reflection.%2$s%2$sPossible reasons are:%2$s%2$s * Requiring non-existent classes%2$s * Using non-supported annotations%2$s * Class-/filename missmatch.%2$s%2$sThe "Neos.Flow.object.excludeClasses" setting can be used to skip classes from being reflected.', $className, chr(10)));
         $exception->className = $className;
         return $exception;
     }
 
-    public function getClassName(): ?string {
+    public function getClassName(): ?string
+    {
         return $this->className;
     }
 }
--- a/Neos.Flow/Classes/Reflection/Exception/ClassLoadingForReflectionFailedException.php
+++ b/Neos.Flow/Classes/Reflection/Exception/ClassLoadingForReflectionFailedException.php
@@ -18,17 +18,14 @@ namespace Neos\Flow\Reflection\Exception;
  */
 class ClassLoadingForReflectionFailedException extends \Neos\Flow\Reflection\Exception
 {
-    protected $className;
-
-    public static function forClassName(string $className): self
+    public static function forClassName(string $className, string $reflectedClass): self
     {
-        $exception = new self(sprintf('Required class "%s" could not be loaded properly for reflection.%2$s%2$sPossible reasons are:%2$s%2$s * Requiring non-existent classes%2$s * Using non-supported annotations%2$s * Class-/filename missmatch.%2$s%2$sThe "Neos.Flow.object.excludeClasses" setting can be used to skip classes from being reflected.', $className, chr(10)));
-        $exception->className = $className;
-        return $exception;
-    }
-
-    public function getClassName(): ?string
-    {
-        return $this->className;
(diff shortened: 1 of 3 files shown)


Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1224,7 +1224,15 @@ protected function reflectClass($className)
             throw new Exception\InvalidClassException('The class with name "' . $className . '" is a Doctrine proxy. It is not supported to reflect doctrine proxy classes.', 1314944681);
         }
 
-        $class = new ClassReflection($className);
+        // FIXME: This try catch was added with Flow 7.0 to improve the developer experience for upgrading components and can be removed later
```
sorenmalling, line 1227: How about we rather build a separate flowpack for upgrade checks?

albe:
image

That's the closest/earliest point where we could catch the exception (not neccessarily the best) and rephrase it to help the developer. Should obviously only be kept for the transition phase

Neos.Flow/Classes/Reflection/ReflectionService.php
albe: The issue is that an existing class that still implements ComponentInterface will cause an exception during the compilation step (even during booting of the compilation step) already, where we don't have any hook-into mechanism in place. Otherwise I'd highly favor that of course

sorenmalling: Damn

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1224,7 +1224,15 @@ protected function reflectClass($className)
...
+        try {
+            $class = new ClassReflection($className);
+        } catch (\Exception $e) {
+            if (stripos($e->getMessage(), '"Neos\Flow\Http\Component\ComponentInterface"') !== false) {
```
bwaidelich, line 1231: I wonder why we need to instantiate the ClassReflection to trigger the exception.. can't we add the check at the place where it was created before?
The check for the exception message is quite quirky, maybe we could instead add the failing class name to the (custom) exception somehow.

Also we should add a `@deprecated` annotation so that we don't forget to remove the hack in a future release

albe: > I wonder why we need to instantiate the ClassReflection to trigger the exception.. can't we add the check at the place where it was created before?

Well, the exception happens when the autoloader tries to load the ComponentInterface of course. Since we don't know in advance, which classes might have an `implements ComponentInterface`, we need to somehow include/autoload the current class and catch exceptions. Maybe we could also hook into the autoloading though? That could be a good idea.

albe: Okay, I took a look at the seemingly nice `autoloader` approach:
- we need to hardcode the registering of the autoloader somewhere (only for compile time though), so that's not much better than this try catch unless we go the full route to create a new package just for that and do it in the package boot... meh :/
- more importantly: We will not know the class which caused the `ComponentInterface` loading attempt there, so we can not tell that in the error message, which makes it a **lot less useful**

Hence I still think this is a good place to do this check. I adjusted the try catch and the exception so that we don't compare the exception message, but instead can retrieve the class name that caused the reflection loading error from the exception (via static constructor method).

Neos.Flow/Classes/Reflection/Exception/ClassLoadingForReflectionFailedException.php
```diff
@@ -18,4 +18,14 @@
  */
 class ClassLoadingForReflectionFailedException extends \Neos\Flow\Reflection\Exception
 {
+    public static function forClassName(string $className, string $reflectedClass): self
+    {
+        // @deprecated This check was added with Flow 7.0 to improve the developer experience for upgrading components and can be removed later
+        if ($className === \Neos\Flow\Http\Component\ComponentInterface::class) {
+            $message = sprintf('The class "%s" still implements the ComponentInterface. The component chain was replaced with a middleware chain in Flow 7. Please make sure you have read the upgrade instructions and converted your components to middlewares.', $reflectedClass);
+        } else {
+            $message = sprintf('Required class "%s" could not be loaded properly for reflection.%2$s%2$sPossible reasons are:%2$s%2$s * Requiring non-existent classes%2$s * Using non-supported annotations%2$s * Class-/filename missmatch.%2$s%2$sThe "Neos.Flow.object.excludeClasses" setting can be used to skip classes from being reflected.', $className, chr(10));
```
albe, line 27: Note: We could now also add an additional hint in this message which class caused the failing reflection class loading.

bwaidelich (approved):
I love it. It's a minimal-invasive fix with a maximal helpful behavior.
And you even improved the message for the good old "...could not be loaded for reflection" exception

PR #2315, opened 2020-12-05
TASK: Add core migration that warns about component chain usage

bwaidelich:
Provides a minimal core migration that will show a warning
when the migrated package contains PHP code that refers to
the no longer existing `ComponentInterface`.

Usage:

    ./flow flow:core:migrate <Package-Key> --version 20201205172733

Example output:

```
Warnings
--------------------------------------------------------------------------------
  * Following files refer to the ComponentInterface that has been removed
    with Flow 7.0:

    * Classes/Http/SomeComponent.php
    * Classes/Http/SomeOtherComponent.php

    The component chain was replaced with a middleware chain in Flow 7. Please make
    sure you have read the upgrade instructions and converted your components to
    middlewares.
--------------------------------------------------------------------------------
```

Related: #2258

### DIFF
--- a/Neos.Flow/Migrations/Code/Version20201205172733.php
+++ b/Neos.Flow/Migrations/Code/Version20201205172733.php
@@ -1,0 +1,41 @@
+<?php
+namespace Neos\Flow\Core\Migrations;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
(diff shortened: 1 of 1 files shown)


sorenmalling:
An this be executed before the reflection/compiling kicks in and throw an exception?

albe:
> An this be executed before the reflection/compiling kicks in and throw an exception?

If you follow the instructions and do run migrations when upgrading, yes :) They will completely bypass the Flow bootstrap, exactly because they might need to run a "broken" installation and fix it.

albe (approved):

bwaidelich:
Yes, core migrations are not executed through the Flow bootstrap so they can actually fix code

PR #2316, opened 2020-12-05
TASK: Some psalm corrections

albe:
This makes use of psalm specific docblock attributes for class-string types.
That tells Psalm to make sure that the value is always given a ::class constant

https://psalm.dev/docs/annotating_code/type_syntax/scalar_types/#class-string


albe:
https://travis-ci.com/github/neos/flow-development-collection/jobs/454999690 psalm is happy again

kdambekalns (approved):
Why did you remove the `assert(…)` calls?

albe:
Made tests fail, because they apparently inject non-classes failing the assertions 🤷‍♂️ the `/** @var */` blocks do it

kdambekalns:
I got fooled by the diff view. You removed those first, then added them back, so they are not even removed… 😎

bwaidelich:
Yay, thanks!

PR #2318, opened 2020-12-07
Update MINIMUM_PHP_VERSION and FLOW_VERSION_BRANCH constants

bwaidelich:
Updates `MINIMUM_PHP_VERSION` and `FLOW_VERSION_BRANCH` PHP constants

### DIFF
--- a/Neos.Flow/Classes/Core/Bootstrap.php
+++ b/Neos.Flow/Classes/Core/Bootstrap.php
@@ -34,7 +34,7 @@ class Bootstrap
     /**
      * Required PHP version
      */
-    const MINIMUM_PHP_VERSION = '7.1.0';
+    const MINIMUM_PHP_VERSION = '7.3.0';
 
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Core/Bootstrap.php
```diff
@@ -34,7 +34,7 @@ class Bootstrap
     /**
      * Required PHP version
      */
-    const MINIMUM_PHP_VERSION = '7.1.0';
+    const MINIMUM_PHP_VERSION = '7.3.0';
```
albe, line 37: Didn't even know we had that `const` 😅

albe (approved):

Neos.Flow/Classes/Core/Bootstrap.php
bwaidelich: I just found it myself :)

PR #2322, opened 2020-12-07
TASK: Adjust version numbers in tooling configuration

bwaidelich:
Updates for the 7.0 branch:

* conf.py of neos/flow
* .travis.yml
* apigen.yml


Neos.Flow/Documentation/conf.py
```diff
@@ -59,9 +59,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = 'dev-master'
+version = '7.0'
```
albe, line 62:
```suggestion
version = 'dev-7.0'
```
Was that `dev-7.0` intended? Because it should be just 7.0, no?

albe (approved):

Neos.Flow/Documentation/conf.py
bwaidelich: Well, I just copied what @kdambekalns did in 37396baf4725379f7d24146738fe17f1b8a30d4f

kdambekalns: Rightly so.

kdambekalns: Ah, that targets the next changed line, actually…

Neos.Flow/Documentation/conf.py
```diff
@@ -59,9 +59,9 @@
...
 # The full version, including alpha/beta/rc tags.
-release = 'dev-master'
+release = 'dev-7.0.x'
```
kdambekalns, line 64:
```suggestion
release = '7.0.0'
```

Yes, that never get's updated during the lifecycle (yet), but still…

kdambekalns: Hm… actually that's what I did in https://github.com/neos/flow-development-collection/commit/37396baf4725379f7d24146738fe17f1b8a30d4f. 🙈

Probably _because_ it is not updated automagically for releases and rendered from the branch tip anyway. All is well!

kdambekalns (approved):

PR #2324, opened 2020-12-07
TASK: Fix composer dependencies

albe:
typo3fluid was accidentially included in the collection require, vimeo/psalm was duplicated and doctrine dependencies were duplicated in require-dev.

### DIFF
--- a/Neos.FluidAdaptor/composer.json
+++ b/Neos.FluidAdaptor/composer.json
@@ -11,7 +11,7 @@
     "neos/cache": "*",
     "neos/utility-files": "*",
     "neos/utility-objecthandling": "*",
-    "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
+    "typo3fluid/fluid": "~2.6.10",
     "psr/log": "^1.0"
   },
   "autoload": {
--- a/Neos.FluidAdaptor/composer.json
+++ b/Neos.FluidAdaptor/composer.json
@@ -11,7 +11,7 @@
     "neos/cache": "*",
     "neos/utility-files": "*",
     "neos/utility-objecthandling": "*",
-    "typo3fluid/fluid": "~2.6.10",
+    "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
     "psr/log": "^1.0"
   },
   "autoload": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -31,7 +31,6 @@
         "neos/composer-plugin": "^2.0",
         "composer/composer": "^1.9 || ^2.0",
         "egulias/email-validator": "^2.1",
-        "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
```
albe, line 34: See #2152

Neos.FluidAdaptor/composer.json
```diff
@@ -11,7 +11,7 @@
     "neos/cache": "*",
     "neos/utility-files": "*",
     "neos/utility-objecthandling": "*",
-    "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
+    "typo3fluid/fluid": "~2.6.10",
```
albe, line 14:
```suggestion
    "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
```

kdambekalns:
Looking at this again, the fixes are nice,  but in parts futile, as the `ComposerManifestMerger` will undo them:

- duplication needs to be fixed in `ComposerManifestMerger`
- `neos/fluid-adaptor` is required in `Neos.Kickstarter`, so it will always come back here…

albe:
> Looking at this again, the fixes are nice, but in parts futile, as the `ComposerManifestMerger` will undo them:
> 
> * duplication needs to be fixed in `ComposerManifestMerger`

Correct, see https://github.com/neos/BuildEssentials/pull/45 for a fix

> * `neos/fluid-adaptor` is required in `Neos.Kickstarter`, so it will always come back here…

I guess you mean `typo3fluid/fluid` is required in `Neos.FluidAdaptor` so it will come back, which are both correct. We will need to split off the FluidAdaptor first, then make Neos.Kickstarter only optionally depend on Fluid as well.

albe:
PS: Just for reference/cross-linking related things, we only have neos/fluid-adaptor in Neos.Kickstarter since https://github.com/neos/flow-development-collection/pull/2323

PR #2325, opened 2020-12-07
Revert "TASK: Fix composer dependencies"

albe:
Reverts neos/flow-development-collection#2324

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "neos/composer-plugin": "^2.0",
         "composer/composer": "^1.9 || ^2.0",
         "egulias/email-validator": "^2.1",
+        "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
         "guzzlehttp/psr7": "^1.4 || ~2.0",
         "ext-mbstring": "*"
     },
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -31,6 +31,7 @@
         "neos/composer-plugin": "^2.0",
         "composer/composer": "^1.9 || ^2.0",
         "egulias/email-validator": "^2.1",
+        "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
```
albe, line 34: Removing that will make tests fail, because fluid will indeed not get installed. Fluid-Adaptor is in the collection and auto-exists, and hence all tests for that are also run, but as it's not composer required, it's own dependencies will not be installed.

composer.json
```diff
@@ -114,6 +115,9 @@
     "require-dev": {
         "mikey179/vfsstream": "~1.6",
         "phpunit/phpunit": "~9.0",
+        "vimeo/psalm": "~3.10.0",
+        "doctrine/orm": "^2.7",
+        "doctrine/common": "^2.13.1 || ^3.0",
```
albe, line 120:
```suggestion
```

PR #2326, opened 2020-12-08
TASK: Improve exception when a class can not be loaded during AOP building

albe:
This will now throw an exception that suggests to flush caches or check the existence and namespacing of the aspect class.

Until now such a case would lead to an unhelpful exception of:
```
Warning: Invalid argument supplied for foreach() in /…/Packages/Framework/Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php line 319
```
because the following `get_class_methods()` call would return `null` which is not checked for.

### DIFF
--- a/Neos.Flow/Classes/Aop/AspectContainer.php
+++ b/Neos.Flow/Classes/Aop/AspectContainer.php
@@ -84,9 +84,6 @@ class AspectContainer
      */
     public function __construct(string $className)
     {
-        if (!class_exists($className)) {
-            throw new InvalidTargetClassException(sprintf('The class "%s" is not loadable for AOP proxy building. This is most likely an inconsistency with the caches. Try running `./flow flow:cache:flush` and if that does not help, check the class exists and is correctly namespaced.', $className), 1607422151);
-        }
         $this->className = $className;
     }
 
--- a/Neos.Flow/Classes/Aop/AspectContainer.php
+++ b/Neos.Flow/Classes/Aop/AspectContainer.php
@@ -84,6 +84,9 @@ class AspectContainer
      */
     public function __construct(string $className)
     {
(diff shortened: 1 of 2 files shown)


albe:
Alternatively we only check for `null` in the AopProxyBuilder, but that could maybe have other reasons than the class not existing.

kdambekalns (approved):

albe:
Thinking about it, we should definitely still check the `null`...

Neos.Flow/Classes/Aop/AspectContainer.php
```diff
@@ -84,6 +84,9 @@ class AspectContainer
      */
     public function __construct(string $className)
     {
+        if (!class_exists($className)) {
+            throw new InvalidTargetClassException(sprintf('The class "%s" is not loadable for AOP proxy building. This is most likely an inconsistency with the caches. Try running `./flow flow:cache:flush` and if that does not help, check the class exists and is correctly namespaced.', $className), 1607422151);
+        }
```
albe, line 89:
```suggestion
```

Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -315,6 +316,9 @@ protected function buildAspectContainer(string $aspectClassName): AspectContaine
     {
         $aspectContainer = new AspectContainer($aspectClassName);
         $methodNames = get_class_methods($aspectClassName);
+        if ($methodNames === null) {
```
albe, line 319: I guess that's the better place to check this after all. Avoids another `class_exists()` check and also covers potentially other cases.

bwaidelich (approved):
Nice!

PR #2331, opened 2020-12-08
TASK: Replace symfony polyfills due to min PHP version

albe:
See https://github.com/symfony/polyfill#design

This prevents installing the symfony polyfills for php <= 7.1 and mbstring, since those are requirements of this Flow version.


bwaidelich (approved):

albe:
Will need to add symfony/polyfill-php72 and symfony/polyfill-php73 in 6.0 respectively 7.0 branches after upmerges.

Issue #2244, opened 2020-11-12
Finalize update to doctrine/migrations 3.0


kdambekalns:
- [x] adjust the DB migrations in Neos (the dev dist)
- [x] adjust the DB migrations in our other packages
  - [x] [neos/redirecthandler-databasestorage](https://github.com/neos/redirecthandler-databasestorage/tree/master/Migrations) – https://github.com/neos/redirecthandler-databasestorage/pull/26
  - [x] [neos/neos-googleanalytics](https://github.com/neos/neos-googleanalytics/tree/master/Migrations) - https://github.com/neos/neos-googleanalytics/pull/74
- [x] check everything works fine in more (real) projects
- [x] still different from before: output during migrations no longer shows what's happening. Need to figure out where the output goes (probably logger in the dependency factory needs to be populated and then read from)… #2328
- [x] go over CLI output in the documentation and update as needed… #2323
- [x] ~when a property has been removed from a model, `cache:flush` needs to be run to avoid an error (see below)~ -> #2336

_Originally posted by @kdambekalns in https://github.com/neos/flow-development-collection/issues/1880#issuecomment-726137105_

kdambekalns:
https://github.com/neos/flow-development-collection/issues/2244

kdambekalns:
@albe With "in our other packages" I meant  things _not_ pulled in with the dev distribution… Things like the redirect handler and similar packages…

bwaidelich:
When I executed `./flow doctrine:migrate` in a local test project of mine that already had migrations, I got an error
```
An exception occurred while executing 'RENAME TABLE typo3_flow_mvc_routing_objectpathmapping TO neos_flow_mvc_routing_objectpathmapping':
SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'neos_flow_mvc_routing_objectpathmapping' already exists
```
And the `flow_doctrine_migrationstatus` now has both, the old `<Version>` entries and the new ones with the FQN.

My bad because I didn't read the instructions and apparently I should have executed `./flow doctrine:migrationstatus` first. But I'm sure I won't be the only one so maybe
* We can tweak the `doctrine:migrate` command such that it checks existing migrations and/or
* Provide another command that fixes the issue (reverting the structural changes to `flow_doctrine_migrationstatus`, removing the new entries and fixing it properly)

bwaidelich:
[...]

PR #2333, opened 2020-12-09
BUGFIX: Persist doctrine MetadataStorage changes immediately

bwaidelich:
This basically adds a `EntityManager::flush()` call after `MetadataStorage::ensureInitialized()`
invocations in order to apply any metadata migrations immediately.

Without this a `./flow doctrine:migrate` call might leave the metadata storage in a half-migrated
state if it leads to an exception.

Related: #2244
Related: #2357

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Service.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Service.php
@@ -250,7 +250,7 @@ class Service
      */
     public function getFormattedMigrationStatus($showMigrations = false): string
     {
-        $this->getDependencyFactory()->getMetadataStorage()->ensureInitialized();
+        $this->initializeMetadataStorage();
 
         $infosHelper = $this->getDependencyFactory()->getMigrationStatusInfosHelper();
         $infosHelper->showMigrationsInfo($this->logMessages);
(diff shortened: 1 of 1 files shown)


kitsunet (approved):
Sooo weird that it's necessary but well, if that's what is needed....

kdambekalns (approved):
If it helps… 🤔

kdambekalns:
I am really at a loss here… Why? I tried PHP errors in migrations, invalid SQL in migrations, re-applying already executed migrations – and did not need that (again).

But, whatever, if it solves if, do it. 🤷‍♂️

bwaidelich:
@kdambekalns Mh, you're right (as usual).. And I can't reproduce this any longer..
What I tried:
```sql
UPDATE flow_doctrine_migrationstatus SET version = RIGHT(version, 14);
ALTER TABLE flow_doctrine_migrationstatus DROP executed_at, DROP execution_time;
```
(to revert the metadata table to its old state)

```sql
DELETE FROM flow_doctrine_migrationstatus WHERE version = '20200908155620';
```

to remove the migration that was missing in my case.

Previously that led to a swallowed error and the half-migrated table.
Now (without this PR applied) I get an exception
```
An exception occurred while executing 'ALTER TABLE neos_flow_resourcemanagement_persistentresource DROP md5':

SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP COLUMN `md5`; check that it exists
```
(as expected)

and the table is fully migrated..

Probably I _was_ in some weird state when I tried it last.

I'll close this one for now and hope that this heisenbug never returns :)

kdambekalns:
> I'll close this one for now and hope that this heisenbug never returns :)

🤞

kdambekalns:
Wait, we never merged this…

albe:
I thought that was because the issue that this *seemed* to solve was no longer reproducible without this change?

> Now (without this PR applied) I get an exception
> (as expected)
> and the table is fully migrated..

kdambekalns:
@albe Yes, but see #2357 …

albe:
🤔 well, yeah. If this fixes those cases...

PR #2335, opened 2020-12-10
BUGFIX: Don't access static properties in annotation classes

albe:
`get_class_vars()` will also return static properties. Hence we also try to access static properties for rendering the annotation and access them non-statically. This will lead to errors.

See https://github.com/zircote/swagger-php/issues/359#issuecomment-303058035 / https://github.com/zircote/swagger-php/issues/359#issuecomment-303227013

This solves that by using `get_object_vars()`, which will only return non-static properties with their current values.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -286,10 +286,10 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
     {
         $annotationAsString = '@\\' . get_class($annotation);
 
-        $optionNames = get_object_vars($annotation);
+        $optionDefaults = get_class_vars(get_class($annotation));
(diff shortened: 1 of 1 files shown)


kdambekalns (approved):
Looks good! 🎉 But how did you unearth this issue!? 😳

albe:
I suggested to use this package (after quickly googling "php swagger") yesterday on slack for generating swagger docu from existing Flow controllers. The person came back today stating that there were issues and linking to the above mentioned issue on the package. After reading through it, I saw that the assessment that there was a bug in our code was correct :) Well, only took the better of 3 years to reach us, but at least it's fixed now :tada:

albe:
I'm a bit lost on those behat failures on 5.3 still...

> require(Behat\MinkExtension)

That looks like it tries to load a namespace as a class, which is not working of course. But why? And all of a sudden at that?

albe:
Okay, I think I get it now. See https://github.com/neos/behat/issues/23

PR #2339, opened 2020-12-10
TASK: Only invoke authentication entry point for unauthenticated tokens

### LINKED ISSUE #2234: SecurityEntryPoint is invoked for already authenticated tokens
In the SecurityEntryPoint, when an `AuthenticationRequiredException` is handled and multiple tokens are "active" and have an entry point, the `startAuthentication()` will be invoked for all of them, even if the token is already authenticated.
This seems unnecessary.

This likely was not discovered yet because
- for single factor authentication this case does not exist (only the single unauthenticated token raised the exception)
- the entry points themself might check an existing authentication and return early

See also https://github.com/neos/flow-development-collection/pull/2223/files#r519190892

The solution would be to add a simple `if ($token->isAuthenticated()) continue;` check to the token iteration.

albe:
In the SecurityEntryPoint, when an AuthenticationRequiredException is handled and multiple tokens are "active" and have an entry point, the startAuthentication() will be invoked for all of them, even if the token is already authenticated. This could lead to wrong entry points being invoked in a multi-factor authentication setup.
This changes that by only invoking the entry point for unauthenticated tokens.

Resolves #2234


Neos.Flow/Classes/Http/Middleware/SecurityEntryPointMiddleware.php
```diff
@@ -67,6 +67,8 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
 
             $response = $this->buildHttpResponse();
             foreach ($tokensWithEntryPoint as $token) {
+                if ($token->isAuthenticated()) continue;
```
albe, line 70:
```suggestion
                if ($token->isAuthenticated()) {
                    continue;
                }
```

albe:
Thinking again, let's get this into next release and see how it behaves. If issues come up in lower versions, we can backport it then. If it breaks things, we can hotfix more easily and if needed postpone to next major.

kdambekalns (approved):
Makes sense.

kitsunet (approved):
Looks sound, BUT I have to pose the question: could this behavior have been used for something like oauth refreshes?

albe:
Hmmm... but shouldn't an OAuth token that needs refresh not lead to `AUTHENTICATION_REQUIRED`, i.e. not authenticated status?

kitsunet:
> Hmmm... but shouldn't an OAuth token that needs refresh not lead to `AUTHENTICATION_REQUIRED`, i.e. not authenticated status?

I guess? Depends how well managed it it. I rather meant if you were to rely on that being called to check if you need to refresh before the token lifetime runs out. But I guess a rather obscure usecase that shouldn't happen anywys. I'll merge 😆

PR #2342, opened 2020-12-11
TASK: Redact the 7.0 changelog

albe:
Order by !!!FEATURE, FEATURE, !!!BUG, BUG, !!!TASK, TASK and remove unhelpful "Apply fixes from StyleCI".
Also removed the Easy injection of request feature and it's revert.

### DIFF
--- a/Neos.Flow/Documentation/TheDefinitiveGuide/PartV/ChangeLogs/700.rst
+++ b/Neos.Flow/Documentation/TheDefinitiveGuide/PartV/ChangeLogs/700.rst
@@ -4,6 +4,598 @@
 Overview of merged pull requests
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


kdambekalns (approved):
Did not read it, actually… but I'm sure it's fine. :)

Note: Ordering this is not something our script does, so unless you want this to be a one-off, you hereby volunteer to teach our script to do that or be the "changelog maintainer" from now on. :stuck_out_tongue_winking_eye:

jonnitto (approved):

albe:
> Note: Ordering this is not something our script does, so unless you want this to be a one-off, you hereby volunteer to teach our script to do that or be the "changelog maintainer" from now on. 😜

See https://github.com/neos/flow-development-distribution/issues/48 :) If I were a bit better with bash scripting I'd probably already have done it. Still have it on my radar

PR #2343, opened 2020-12-11
BUGFIX: Fix creation of translations with targetLanguageKeys defined

Benjamin-K:
**What I did**

Previously, when trying to create translation files with targetLanguageKeys defined (`flow kickstart:translation --target-language-keys="de"` for example), an error was thrown. 
The error said, that translationUnits is not defined in Neos\Kickstarter\Service\GeneratorService. This is, because Neos\Flow\I18n\Xliff\V12\XliffParser::getParsedData will return an array (for each <file>-Entry in the Xliff-File, typically only one entry). So we need to use the first item of that data to get the translationUnits from it. 

**How I did it**

I only made a little change and added `[0]` to `$parsedXliffArray` in `Neos\Kickstarter\Service\GeneratorService::generateTranslation()`.

**How to verify it**

1. Try to kickstart a translation with `--target-language-keys` defined. An error will be thrown.
2. Change `$parsedXliffArray['translationUnits']` to `$parsedXliffArray[0]['translationUnits']` in the foreach loop of `Neos\Kickstarter\Service\GeneratorService::generateTranslation()`.
3. Retry to kickstart the translations.

**Checklist**

  - Lowest maintained branch is 4.3, so the PR is created against that version.

### DIFF
--- a/Neos.Kickstarter/Classes/Service/GeneratorService.php
+++ b/Neos.Kickstarter/Classes/Service/GeneratorService.php
@@ -411,7 +411,7 @@ class GeneratorService
             $parsedXliffArray = $xliffParser->getParsedData($sourceLanguageFile);
             foreach ($targetLanguageKeys as $targetLanguageKey) {
                 $contextVariables['targetLanguageKey'] = $targetLanguageKey;
-                $contextVariables['translationUnits'] = $parsedXliffArray['translationUnits'];
+                $contextVariables['translationUnits'] = $parsedXliffArray[0]['translationUnits'];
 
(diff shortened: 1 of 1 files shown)


albe:
Welcome, thanks a lot for the clear description and the fixing PR! This looks good, but it should actually target 5.3, as this is the lowest maintained Flow version - where the Kickstarter is living in. Gladly that sort of confusion will be gone once 7.x is the lowest maintained, but that's still a couple years away :)

albe (approved):

Benjamin-K:
Sry, i mixed Neos version with Flow version. Should be 5.3 of course.

PR #2348, opened 2020-12-12
BUGFIX: Don't persist Resources that failed validation

### LINKED ISSUE #769: PersistentResource gets persisted if a attached validation fails

albe:
Resources that failed validation will again be removed from the repository, by registering a callback for object validation and checking the result. So unless `persistAll()` was already called before validation, a validation failed `PersistentResource` will not even end up in the database. It will be temporarily imported into the storage and deleted again though.

This is an alternative implementation to #1931 which should be less far-reaching, but effectively introduces a new feature.

Fixes #769


Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -171,8 +187,7 @@ public function importResource($source, $collectionName = ResourceManager::DEFAU
             throw new Exception(sprintf('Importing a file into the resource collection "%s" failed: %s', $collectionName, $exception->getMessage()), 1375197120, $exception);
         }
 
-        $this->resourceRepository->add($resource);
-        $this->logger->debug(sprintf('Successfully imported file "%s" into the resource collection "%s" (storage: %s, a %s. SHA1: %s)', $source, $collectionName, $collection->getStorage()->getName(), get_class($collection), $resource->getSha1()));
+        $this->persistResource($resource, $collection, is_resource($source) ? get_resource_type($source) : sprintf('file "%s"', $source));
```
albe, line 190: Not sure it's okay to change this slightly, but before importing a PHP `resource` would log something like `Successfully imported file "Resource id #5" into the resource collection ...`. With this, it will become `Successfully imported stream into the resource collection ...`. We can change it back though.

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -218,8 +233,7 @@ public function importResourceFromContent($content, $filename, $collectionName =
             throw new Exception(sprintf('Importing content into the resource collection "%s" failed: %s', $collectionName, $exception->getMessage()), 1381156155, $exception);
         }
 
-        $this->resourceRepository->add($resource);
-        $this->logger->debug(sprintf('Successfully imported content into the resource collection "%s" (storage: %s, a %s. SHA1: %s)', $collectionName, $collection->getStorage()->getName(), get_class($collection->getStorage()), $resource->getSha1()));
+        $this->persistResource($resource, $collection, 'content');
```
albe, line 236: We could improve the source hint here to the `$filename`.

Neos.Flow/Classes/Validation/Validator/GenericObjectValidator.php
```diff
@@ -43,6 +48,15 @@ public function setValidatedInstancesContainer(\SplObjectStorage $validatedInsta
         $this->validatedInstancesContainer = $validatedInstancesContainer;
     }
 
+    /**
+     * @param object $object The object instance to register a validation callback for
+     * @param \Closure $callback The callback to invoke once this object has been validated
+     */
+    public static function onValidated(object $object, \Closure $callback)
```
albe, line 55: Static method callback registration. It's a special case, so signal slot didn't seem like the right approach. We'd need to register a single slot every time and invoke it for every object validation, not only `iff` a callback for the specific object instance is registered.

So why on the `GenericObjectValidator`? Because it's the root validator for every object type, so it will be invoked for every object that gets validated (see https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Classes/Validation/ValidatorResolver.php#L291-L296). The `AggregateBoundaryValidator` extends this and will just early out for unloaded proxies then delegate to this.

Neos.Flow/Classes/Validation/Validator/GenericObjectValidator.php
```diff
@@ -43,6 +48,15 @@ public function setValidatedInstancesContainer(\SplObjectStorage $validatedInsta
...
+    {
+        self::$validationCallbacks[spl_object_hash($object)][] = $callback;
```
albe, line 57: With 6.0 (PHP 7.2+) we could replace this with `spl_object_id()`, which is a tiny bit faster and has same semantics.

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -134,6 +136,20 @@ protected function initialize()
         $this->initialized = true;
     }
 
+    protected function persistResource(PersistentResource $resource, CollectionInterface $collection, string $sourceHint = ''): void
+    {
+        $this->resourceRepository->add($resource);
+        GenericObjectValidator::onValidated($resource, function (Result $result) use ($resource) {
```
albe, line 142: register a validation callback that will remove the object from persistence if it had errors.

Neos.Flow/Classes/Validation/Validator/GenericObjectValidator.php
```diff
@@ -43,6 +48,15 @@ public function setValidatedInstancesContainer(\SplObjectStorage $validatedInsta
         $this->validatedInstancesContainer = $validatedInstancesContainer;
     }
 
+    /**
+     * @param object $object The object instance to register a validation callback for
+     * @param \Closure $callback The callback to invoke once this object has been validated
+     */
+    public static function onValidated($object, \Closure $callback): void
```
albe, line 55: This should become `object $object` in 6.0/PHP 7.2

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -134,6 +136,20 @@ protected function initialize()
...
+            if ($result->hasErrors()) {
+                $this->deleteResource($resource, false);
```
albe, line 144: Just a `resourceRepository->remove($resource)` does not suffice, because the file has already been stored on the storage, which should not have happened in the first place...

albe:
Note: This implementation will lead to the resource being imported into the storage, be marked for persistence, then on validation error be removed from persistence UoW and deleted from the storage again. So this is still suboptimal from the expected behaviour, but the implementation is much more straightforward.

albe:
Apparently, the GenericObjectValidator *isn't* the right spot for this, because that is possibly combined with another custom model validator inside a ConjunctionValidator, so the `onValidated()` callback could be given a wrong `Result`. Also, the `GenericObjectValidator` can be removed again, if it contains no sub-property validators - so objects that just pass validation without checks will not trigger the `onValidated()` callback.

I have now prepared a set of tests that replicate the errorenous behaviour in two versions: direct `PersistentResource` upload and an object containing a `PersistentResource` inside. In the latter case the resource also shouldn't be persisted, even if only the validation for the wrapping object fails.

DavidSporer:
> Note: This implementation will lead to the resource being imported into the storage, be marked for persistence, then on validation error be removed from persistence UoW and deleted from the storage again. So this is still suboptimal from the expected behaviour, but the implementation is much more straightforward.

This is basically what we're doing in a custom validator. Basically the object is removed by calling ResourceManager's deleteResource inside the validator if validation fails.
I read your next comment and agree that it's not ideal but this issue also came up during security tests which is the reason why I went with it. Not sure if I can really add something to the discussion but in our case this was seen as a very critical issue not only because it allows users to spam servers with files.
In our case we're checking files for malicious content and it means that if files aren't removed properly (or not prevented from being persisted in the first place) a user could upload malicious files and they are saved even if validation fails.

Let me know if I can test something but as far as I understood the change you propose this should fix the issue.

albe:
Thanks @DavidSporer for the confirmation of the seriousness of the issue. I'm juggling back and forth on the two implementations I have to fix this behaviour (#1931 and this) and as mentioned above started to prepare tests that will be the base for both implementations to verify correct behaviour. Both have gone through some refactoring locally following my last comment and I'm now again leaning more towards the first solution (taking some ideas from this solution), but it's not finished yet. I'll try to find some time to work on it again the next weeks.

lorenzulrich:
Thanks @albe for this PR. Do I see correctly that the Core's validators (`FileTypeValidator` in Flow and `ImageTypeValidator` in Neos.Media) would be adapted in separate PRs?

lorenzulrich:
On a related note: In general, every public Neos form without proper protection (e.g. Captcha) can be used to upload any number of files of allowed types (currently any number of any file not intercepted earlier such as `.php`). That means that every form allowing to upload HTML will enable someone to plant a phising page and create a link to it (because they know their file's SHA1) - without the maintainer to necessarily take notice.

I think the default configuration of `Neos.Form` and `Neos.Form.Builder` needs to be adapted to set a sensible default that is not prone to such "attacks".

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
kdambekalns: Fine with me…

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
kdambekalns: That'd be nice.

Neos.Flow/Classes/Validation/Validator/GenericObjectValidator.php
kdambekalns: That means _now_, by now. 😎

Neos.Flow/Classes/Validation/Validator/GenericObjectValidator.php
kdambekalns: Go for it!

kdambekalns (approved):
I like that solution, way more than #1931.

fcool (approved):
Maybe we should rebase this to 7.3? But from reading this would be fine for me.

Is everyone aware, that together with this BUGFIX we introduce a new Feature?
If we think the idea is generally "fine", we should probably somewhere document that `GenericObjectValidator::onValidated` feature.

Neos.Flow/Classes/Validation/Validator/GenericObjectValidator.php
kdambekalns:
```suggestion
        self::$validationCallbacks[spl_object_id($object)][] = $callback;
```

Neos.Flow/Classes/Validation/Validator/GenericObjectValidator.php
kdambekalns:
```suggestion
    public static function onValidated(object $object, \Closure $callback): void
```

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
kdambekalns:
```suggestion
        $this->persistResource($resource, $collection, sprintf('content as "%s"', $filename));
```

kdambekalns (commented):
Rebased and added three suggestions for open review comments…

kdambekalns:
Trying to re-trigger CI…

kdambekalns:
Trying to re-trigger CI…

PR #2349, opened 2020-12-13
BUGFIX: Correctly handle missing configuration with identity creation

albe:
The code did not handle a missing `$configuration` when converting an hash and data array containing an identifier.

See https://github.com/neos/flow-development-collection/pull/1931/files#r503144893


albe:
Merging with previous approval by Karsten in the mentioned comment.

PR #2351, opened 2020-12-13
TASK: Fix some docblock typehints

albe:
This only fixes a couple of docblock typehints primarily in the ResourceManagement and TypeConverters.


albe:
This should also help reduce the noisy changes in #1931

albe:
Does not contain any functional changes, so would merge liberally after travis agrees.

PR #2359, opened 2020-12-17
FEATURE: Make rendering of request in exception log configurable

albe:
With this it is possible to configure if the request data should be rendered into exception stack traces with the newly added configuration setting `Neos.Flow.log.throwables.renderRequestInformation`.

Related #2360


Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.log.schema.yaml
```diff
@@ -3,6 +3,7 @@ properties:
   throwables:
     type: dictionary
     properties:
+      renderRequestInformation: { type: boolean }
       fileStorage:
```
albe, line 7: I noticed that this schema seems way off from the actual settings above?

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -268,17 +268,16 @@ public static function initializeSystemLogger(Bootstrap $bootstrap)
...
      */
-    protected static function initializeExceptionStorage(Bootstrap $bootstrap): ThrowableStorageInterface
+    protected static function initializeExceptionStorage(Bootstrap $bootstrap, array $settings): ThrowableStorageInterface
     {
-        $configurationManager = $bootstrap->getEarlyInstance(ConfigurationManager::class);
-        $settings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'Neos.Flow');
-
         $storageClassName = $settings['log']['throwables']['storageClass'] ?? FileStorage::class;
         $storageOptions = $settings['log']['throwables']['optionsByImplementation'][$storageClassName] ?? [];
+        $renderRequest = $settings['log']['throwables']['renderRequestInformation'] ?? true;
```
sorenmalling, line 280: If this way, lets name the variable "renderRequestInformation" to honour the full Settings key

sorenmalling:
@albe Very good idea - should we do solution discussion in a issue ticket as reference for coming release? 

I would like to look into a way to filter out parameters/header/body-data from log data, so it's not "yes or no" but more "yes, but only these fields"

albe:
Yes, that would be preferable. This is basically the low-hanging-fruit (read: simple) solution to the problem :) +1 for creating an issue that describes the background

sorenmalling:
> Yes, that would be preferable. This is basically the low-hanging-fruit (read: simple) solution to the problem :) +1 for discussing fine details in the issue

+1 for low hanging fruit - let's keep the issue open afterwards, if this change is merged in 👍

albe:
> > Yes, that would be preferable. This is basically the low-hanging-fruit (read: simple) solution to the problem :) +1 for discussing fine details in the issue
> 
> +1 for low hanging fruit - let's keep the issue open afterwards, if this change is merged in 👍

That's why I chose "Related" instead of "Resolves" in the PR message :)

Neos.Flow/Classes/Core/Booting/Scripts.php
albe, line 280:
```suggestion
        $renderRequestInformation = $settings['log']['throwables']['renderRequestInformation'] ?? true;
```

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -295,7 +294,7 @@ protected static function initializeExceptionStorage(Bootstrap $bootstrap): Thro
             return Debugger::getBacktraceCode($backtrace, false, true);
         });
 
-        $throwableStorage->setRequestInformationRenderer(function () {
+        $throwableStorage->setRequestInformationRenderer(function () use ($renderRequest) {
```
albe, line 297:
```suggestion
        $throwableStorage->setRequestInformationRenderer(function () use ($renderRequestInformation) {
```

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -310,7 +309,9 @@ protected static function initializeExceptionStorage(Bootstrap $bootstrap): Thro
 
             $request = $requestHandler->getHttpRequest();
             $response = $requestHandler->getHttpResponse();
-            $output .= PHP_EOL . 'HTTP REQUEST:' . PHP_EOL . ($request == '' ? '[request was empty]' : $request) . PHP_EOL;
+            if ($renderRequest) {
```
albe, line 312:
```suggestion
            if ($renderRequestInformation) {
```

bwaidelich (commented):
Makes sense and looks good (by reading) but IMO the new configuration option should be documented at https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Logging.html

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
```diff
@@ -107,16 +107,18 @@ You might already have stumbled across such an exception (though hopefully not!)
   HTTP REQUEST:
   127.0.0.1:8081keep-aliveno-cacheno-cacheimageMozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.132 Safari/537.361image/webp,image/apng,image/*,*/*;q=0.8same-originno-corshttp://127.0.0.1:8081/flow/welcomegzip, deflate, brde-DE,de;q=0.9,en-US;q=0.8,en;q=0.7
 
-  HTTP RESPONSE:
```
albe, line 110: The response is no longer logged as of Flow 7 (and was totally useless before anyway, as it was just a dummy response).

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Logging.rst
```diff
@@ -107,16 +107,18 @@ You might already have stumbled across such an exception (though hopefully not!)
...
   PHP PROCESS:
   Inode: 
   PID: 2296
   UID: 1
   GID: 1
   User: 
 
+.. note::
+
+  As of Flow 7.1 you can configure if the HTTP Request should be logged in the default exception handler via the setting ``Neos.Flow.log.throwables.renderRequestInformation``.
```
albe, line 119:
```suggestion
  As of Flow 7.1 you can configure if the HTTP Request should be logged in the default ThrowableStorage via the boolean setting ``Neos.Flow.log.throwables.renderRequestInformation``.
```

Issue #2360, opened 2020-12-18
Credentials or other sensitive data should be scrambled from log


sorenmalling:
Christian Krause mentioned on Slack, that login credentials was added to the logs

![image](https://user-images.githubusercontent.com/531918/102604106-fd0bc580-4123-11eb-917a-1e26d89d6da9.png)

This was due to the setInformationRenderer of the ThrowableStorageInterface. 

It should be able to scramble or exclude certain data from a request log in a somehow extendable way

albe:
First step would be that we don't just log down the whole request body. I guess that's the main point of information leakage. A more sophisticated approach could maybe somehow extract which fields in the request body are used for Token credentials and just anonymize those. Note though that the Tokens act on `ActionRequest` rather than `HttpRequest`.

Also, we should strip `'Authorization'` from the headers printed in `RequestInformationHelper::renderRequestHeaders()`

albe:
I guess the most important issue of the `Authorization` header is solved. For custom headers or parameters that include credentials, this is not as easy to solve. The alternative to not render the request at all into the exception log is an easy-way out and will likely land in 7.1.

PR #2361, opened 2020-12-18
BUGFIX: Anonymize authorization header in rendered request headers

albe:
This prevents credential information leakage into logs and generally when using RequestInformationHelper to render request headers. Using `Headers` class directly will still disclose `Authorization` header values though, so take care if you use that.

Related to #2360


Neos.Flow/Classes/Http/Headers.php
```diff
@@ -530,7 +530,7 @@ private function prepareValues($headerName, array $values)
     {
         $preparedValues = [];
         foreach ($values as $value) {
-            $preparedValues[] = sprintf("%s: %s", $headerName, $value);
+            $preparedValues[] = sprintf("%s: %s", $headerName, $headerName === 'Authorization' ? '****' : $value);
```
sorenmalling, line 533: Is it safe to scramble it here - where and how is this Headers being used?

albe: Right now (IIRC 6.0+) it's not really used anywhere (and marked deprecated). I do plan to reintroduce it for the ActionRequest/Response classes as an abstraction though, since the class generally gives good convenience constructs around some edgy headers like `cache-control`.

bwaidelich: Mh, I would suggest to omit this part for now (or create a dedicated PR for a newer branch) since it might break code that relies on `Headers::getPreparedValues()` to return the actual header values

albe, line 533:
```suggestion
            $preparedValues[] = sprintf("%s: %s", $headerName, $value);
```

albe: Convinced. I'll add that change back into 7.x though I'm not sure if that should really be scrambled. See example of the Browser, where we might at some point use the Headers class to render them and scrambling would be an issue.

Neos.Flow/Classes/Http/AbstractMessage.php
```diff
@@ -434,7 +434,7 @@ public function withHeader($name, $value)
     public function withAddedHeader($name, $value)
     {
         $newMessage = clone $this;
-        $newMessage->setHeader($name, $value);
+        $newMessage->setHeader($name, $value, false);
```
albe, line 437: D'uh, there was also another Bug regarding PSR7 compatibility... `withAddedHeader()` behaved the same as `withHeader()` and overwrote existing ones

Neos.Flow/Classes/Http/Helper/RequestInformationHelper.php
```diff
@@ -100,10 +100,14 @@ public static function renderRequestHeaders(RequestInterface $request): string
         $renderedHeaders = '';
         $headers = $request->getHeaders();
         if ($headers instanceof Headers) {
-            $renderedHeaders .= $headers->__toString();
+            $renderedHeaders .= preg_replace('/Authorization:[^\r\n]+/', 'Authorization: ****', $headers->__toString());
```
albe, line 103: Note: This if will be gone with later versions, hence why I put the preg_replace here only instead of doing it after the if for both cases.

albe:
Looks good for me now - @sorenmalling @bwaidelich

Neos.Flow/Classes/Http/AbstractMessage.php
sorenmalling: Is this the right PR to change this - seems more of a standalone PR also in terms of changelog for next releases? And what about tests?

albe: Yep. And yep. I'm laaaaaazy 😂

PR #2362, opened 2020-12-19
BUGFIX: Avoid passing signal information depending on other signals to same slot

albe:
This fixes a regression introduced with #2216

Fixes neos/content-repository-search#42


kdambekalns (approved):
Thanks for the fix, sorry for causing this! So much for our review process… 😬

albe:
Humans are better are finding conceptual errors rather than functional ones :) That's why we have a test suite (which is incomplete, yes)

daniellienert:
Thanks guys!

Issue #2367, opened 2020-12-31
DX: DebugExceptionHandler: Make backtrace code toggable with HTML5 element <details>


sorenmalling:
The current DebugExceptionHandler can be quiet a mouthful, when the full backtrace code is printed

![image](https://user-images.githubusercontent.com/531918/103406010-6eeb0280-4b59-11eb-9da7-aeca029bb3b4.png)

For a improved developer experiment, i suggest that we make us of the HTML5 element `<details>` for rendering the backtrace code. `<details>` has the great benefit of being collapsed by default (see image)
![image](https://user-images.githubusercontent.com/531918/103405976-4ebb4380-4b59-11eb-9f98-f9b95f19a5c3.png)

And have native browser support for expanding by clicking the `<summary>` content (in this example "Backtrace code") - see image:

![image](https://user-images.githubusercontent.com/531918/103406081-b4a7cb00-4b59-11eb-90fa-08f6b95ec544.png)

PR #2368, opened 2020-12-31
FEATURE: Make backtrace code expandable with HTML5 <summary> element

sorenmalling:
The backtrace code is printed when the DebugExceptionHandler is active (in "Development" context).

This can be queit a mouthful for new users and developers. The backtrace go all the way back to the Flow Bootstrapping - things that are interesting when you get more into Flow.

This is changed, by using the HTML5 element `<summary>` element that support collapsing without any extra libraries or code

Resolves: #2367


jonnitto (approved):
Small change, but a huge impact on the developer experience! Thanks a lot, @sorenmalling; great idea! ❤️

daniellienert (changes requested):
I am really unsure about this change as I think it does not improve the experience. 

-  As an experienced developer I now have an additional klick every time a get an error message to see the stack trace
- New developers are already lost with the information above the stack trace.
- The toggle for the backtrace is rather small compared to other visual elements of the error message and can be easily overlooked

sorenmalling:
> -  As an experienced developer I now have an additional klick every time I get an error message to see the stack trace

Is that because the exception that we bring with Flow is not usefully "low-quality", that this extra click is a issue? 
If so and we can agree upon, we ca  create a umbrella issue for working on that - that will bring better development experience for sure 😍

> - New developers are already lost with the information above the stack trace.

Where does this conclusion come from? Because of low-quality exception messages? Experience with new developers at work? And is that comment a pro or cons of the proposed change?


> - The toggle for the backtrace is rather small compared to other visual elements of the error message and can be easily overlooked

We talk about the visual of the "toggle" - so that means that you like the idea, despite above comments? 🤩

sorenmalling:
ping @daniellienert

sorenmalling:
ping @daniellienert

sorenmalling:
ping @albe care to consider this for 7.1 or if it should be rejected?

albe:
I do think this makes the first look on the exception page quite a bit less overwhelming, so 👍 for that. But I also do see the points that Daniel makes - especially in a relatively tight debug cycles with "edit code, reload, check exception stack trace, repeat" having to click again and again can become a nuisance. Not sure if that is more important than not overwhelming in the first place. And yes, let's make the backtrace toggle slightly larger and/or use the white font on dark blue color as for each stack trace method headline.

I'd not veto this and maybe we can somehow convince Daniel that having to click the toggle is not quite as bad?

sorenmalling:
@mficzel sounds superb 👍

sorenmalling:
@daniellienert @mficzel @albe 

Idea came up today:

 - Per default, the stack trace is "hidden".
 - If you expand it, we set a localStorage item like "Flow_Exception_Backtrace_Expanded" = true
 - If this is set, it's expanded by default and can be toggled.

All possible with vanilla js and no requirement from outside libraries

mhsdesign:
as im relatively new to neos i remember that these messages where a long time so overwehlming, that i didnt even really read the exception message. So i would love a more structured less bloated design which is more beginner friendly. (and maybe even more appealing ^^ )

(i do kind of agree with daniel on 'New developers are already lost with the information above the stack trace.' ;) - own experience ^^)

jonnitto:
> @daniellienert @mficzel @albe
> 
> Idea came up today:
> 
> * Per default, the stack trace is "hidden".
> * If you expand it, we set a localStorage item like "Flow_Exception_Backtrace_Expanded" = true
> * If this is set, it's expanded by default and can be toggled.
> 
> All possible with vanilla js and no requirement from outside libraries

@sorenmalling will you finish this feature before we branch for 7.3? Because I would also like to have it as you described

sorenmalling:
> > @daniellienert @mficzel @albe
> > Idea came up today:
> > 
> > * Per default, the stack trace is "hidden".
> > * If you expand it, we set a localStorage item like "Flow_Exception_Backtrace_Expanded" = true
> > * If this is set, it's expanded by default and can be toggled.
> > 
> > All possible with vanilla js and no requirement from outside libraries
> 
> @sorenmalling will you finish this feature before we branch for 7.3? Because I would also like to have it as you described

Yes, I will give it a go Monday 👍

sorenmalling:
ping all involved :)
Videos show the implementation of localStorage

By default it's closed
If you open, it will open on next request
If you close again, it will be closed on next request

https://user-images.githubusercontent.com/531918/143827853-fad892b7-cbad-4555-ba98-f20909cde4a5.mov

sorenmalling:
> I do not strongly disagree - it really seems valuable to others

@daniellienert I hope this solution can be useful for your workflow as well 💙

Neos.Flow/Resources/Public/Error/Exception.js
```diff
@@ -0,0 +1,14 @@
+let backtrace = document.querySelector(".Flow-Debug-Exception-Backtrace-Code");
+let localStorageItemIdentifier = 'Flow-Debug-Exception-Backtrace-Code-Open';
```
markusguenther, line 2: This could be a `const`, as we don't modify the values

markusguenther (approved):
I can understand the concerns of @daniellienert, but in general, I like the change.
Just had a tiny comment in the JS and the Psalm issue seems to be unrelated :)

Thanks Søren 💙

Neos.Flow/Resources/Public/Error/Exception.css
```diff
@@ -80,6 +80,16 @@ pre {
...
+.Flow-Debug-Exception-Backtrace-Code {
+    background: #fff;
+    color: #656a71;
+    padding: 5px;
+}
+.Flow-Debug-Exception-Backtrace-Code summary{
+    cursor: pointer;
+
+}
+
```
jonnitto, line 92:
```suggestion
.Flow-Debug-Exception-Backtrace-Code {
    background: #fff;
    color: #656a71;
    padding: 5px;
}

.Flow-Debug-Exception-Backtrace-Code summary{
    cursor: pointer;
}

```

jonnitto: Just a bit code cleanup ;-)

jonnitto (approved):
Just a small code cleanup…

PR #2372, opened 2021-01-05
TASK: Fix incomplete docblock typehints

tvt:
Fixed a couple of places where null was a possible return value, but wasn't added to the docblock typehint.


albe (approved):
Hey there, welcome and thanks a lot for your contribution! That makes a lot of sense and was a thorn in my side for some time!

daniellienert (approved):
Looks good. Thanks for your contribution!

tvt:
Thanks guys.

Issue #2375, opened 2021-01-12
Introduce ability to add custom default Fluid namespaces via configuration


martinlipp:
### Description

It would be nice to have the ability to add custom Fluid namespaces (beside the 'f' namespace) via configuration. Custom view helpers inside one's own package are often needed in all/many templates, partials and layouts, and it can be cumbersome to have to add the namespace to all of this files.
I can provide a pull request were this feature is added with minimal code updates inside the FluidAdaptor\Core\ViewHelper\ViewHelperResolver class.

sorenmalling:
Hi @martinlipp 
Did you know that you can call your viewhelper by the syntax

`<vendor.package:viewhelper arguments.... />`

Let me give you an example

```
<dafis.application:navigation.menu>
<li class="{f:if(condition: menuItem.active, then: 'menu-item--active')}">
					<a href="{f:uri.action(package: menuItem.targetPackageKey, controller: menuItem.targetControllerName, action: menuItem.targetActionName, arguments: {organizationIdentifier: organizationIdentifier})}">
						{menuItem.label}
					</a>
				</li>
</dafis.application:navigation.menu>
```

This renders my viewhelper located in the follow namespace/location

```
<?php

namespace Dafis\Application\ViewHelpers\Navigation;

use Neos\Flow\Annotations as Flow;
use Neos\FluidAdaptor\Core\ViewHelper\AbstractViewHelper;
use Dafis\Application\UserInterface\Navigation\Menu;

class MenuViewHelper extends AbstractViewHelper
{
// render method and friends goes here
}
```

so no need for importing the namespace, only to get a shorter syntax.

martinlipp:
Thank you, @sorenmalling.
Yes, I'm aware of that. But I think it would still be a nice feature. Having to write the whole namespace for every view helper, could easily become cumbersome too. I'd prefer the abbreviated syntax, especially when configureable with one line.

sorenmalling:
@martinlipp 
I'm all in favor for that - push a PR and I will review help :)

martinlipp:
Here's a pull request for this feature: https://github.com/neos/flow-development-collection/pull/2378

PR #2378, opened 2021-01-22
FEATURE: Introduce ability to add Fluid namespaces via configuration

martinlipp:
**What I did**
Added the ability to add Fluid namespaces via configuration.

**How I did it**
By adding a new configuration option:
```
Neos:
    FluidAdaptor: 
      defaultNamespaces
```
This configuration is used, to add namespaces inside the ViewHelperResolver.

**How to verify it**
Add a default namespace in Settings.yaml and try to use its view helpers in any template, without including the namespace there.

**Checklist**

Resolves: #2375

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
@@ -54,6 +54,12 @@ class ViewHelperResolver extends \TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperRes
     protected $namespaces = [];
 
     /**
+     * @Flow\InjectConfiguration(path="defaultNamespaces")
+     * @var array
+     */
+    protected $defaultNamespacesFromConfiguration;
+
+    /**
      * ViewHelperResolver constructor.
      */
     public function __construct()
@@ -77,6 +83,10 @@ class ViewHelperResolver extends \TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperRes
                 $viewHelperNamespace .= 'ViewHelpers';
                 $this->addNamespace(strtolower($package->getPackageKey()), $viewHelperNamespace);
             }
+        }
+
+        foreach ($this->defaultNamespacesFromConfiguration as $identifier => $namespace) {
+            $this->addNamespace($identifier, $namespace);
         }
     }
 
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
@@ -59,14 +59,6 @@ class ViewHelperResolver extends \TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperRes
      */
     protected $namespacesFromConfiguration;
 
-    /**
-     * ViewHelperResolver constructor.
-     */
-    public function __construct()
-    {
-        $this->setNamespaces($this->getDefaultNamespaces());
-    }
-
     public function initializeObject($reason)
     {
         if ($reason === ObjectManagerInterface::INITIALIZATIONCAUSE_RECREATED) {
@@ -88,19 +80,6 @@ class ViewHelperResolver extends \TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperRes
         foreach ($this->namespacesFromConfiguration as $identifier => $namespace) {
             $this->addNamespace($identifier, $namespace);
         }
-    }
-
-    /**
-     *
-     */
-    public function getDefaultNamespaces()
-    {
-        return [
-            'f' => [
-                'TYPO3Fluid\\Fluid\\ViewHelpers',
-                'Neos\\FluidAdaptor\\ViewHelpers'
-            ]
-        ];
     }
 
     /**
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
@@ -54,10 +54,10 @@ class ViewHelperResolver extends \TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperRes
     protected $namespaces = [];
 
     /**
-     * @Flow\InjectConfiguration(path="defaultNamespaces")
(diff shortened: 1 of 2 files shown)


Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
```diff
@@ -53,6 +53,12 @@ class ViewHelperResolver extends \TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperRes
      */
     protected $namespaces = [];
 
+    /**
+     * @Flow\InjectConfiguration(path="defaultNamespaces")
+     * @var array
+     */
+    protected $defaultNamespacesFromConfiguration;
```
sorenmalling, line 60: @martinlipp The `defaultNamespacesFromConfiguration` could/should be renamed to `configuredNamespaces` or similar - they are not default but are added from configuration.

Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
```diff
@@ -78,6 +84,10 @@ public function initializeObject($reason)
                 $this->addNamespace(strtolower($package->getPackageKey()), $viewHelperNamespace);
             }
         }
+
+        foreach ($this->defaultNamespacesFromConfiguration as $identifier => $namespace) {
```
sorenmalling, line 88: Does this support the syntax we see in `getDefaultNamespaces`, when we get the values form Settings.yaml? From `getDefaultNamespaces` we se one identifier (`f`) and multiple namespaces.

Neos.FluidAdaptor/Configuration/Settings.yaml
```diff
@@ -56,3 +56,9 @@ Neos:
           options:
             namespaces:
               f: 'Neos\FluidAdaptor\ViewHelpers'
+
```
sorenmalling, line 59: Have you given any thought to move the content of the `getDefaultNamespaces` method into this syntax?

sorenmalling (changes requested):
Added comments about how we resolve current default namespaces and a naming

Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
martinlipp: Yes, it does. The configuration does get injected as an array (if defined as an array inside Settings.yaml) and addNamespace() does allow arrays as second parameter too. I tried this syntax:
```
Neos:
  FluidAdaptor:
     defaultNamespaces:
       f:
         - 'TYPO3Fluid\Fluid\ViewHelpers'
         - 'Neos\FluidAdaptor\ViewHelpers'
```
This worked fine.

Neos.FluidAdaptor/Configuration/Settings.yaml
martinlipp: Yes, in my opinion getDefaultNamespaces() could then be fully replaced by setting the f namespaces via configuration. I tried this and removed the constructor and the getDefaultNamespaces method in the ViewHelperResolver class and it worked for me.
I also tried to simply inject the configuration into the existing $namespaces variable, which also worked. This way the new variable $defaultNamespacesFromConfiguration and the loop for adding them wouldn't be needed . But I'm not sure if this would be a good practice? Also the sanity checks performed by addNamespace(), wouldn't be executed this way.

Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
martinlipp: I'd prefer `namespacesFromConfiguration`. It's less ambiguous than `configuredNamespaces`.

Neos.FluidAdaptor/Configuration/Settings.yaml
sorenmalling: It's important that we keep the sanity check, so let's inject the configuration to a property and loop that 👍 

And replace `getDefaultNamepsaces` and move the configuration to Settings.yaml. I would love to hear from @albe or @bwaidelich if they see anything against doing that?

Neos.FluidAdaptor/Classes/Core/ViewHelper/ViewHelperResolver.php
sorenmalling: Agree - Please update the PR :)

bwaidelich:
disclaimer: I just skimmed over this one and since I don't really use Fluid myself anymore I'm not into it that much.
When we created the namespace support originally, one important feature was that a Fluid file is self-contained. I.e. that you know what it does by reading the source.
By introducing a "spooky action at a distance" we'd lose that ability and tbh I don't see the real benefit.
OTOH I think this feature has been added to TYPO3Fluid anyways so there's probably no reason not to use it

martinlipp:
@bwaidelich I can confirm that TYPO3 CMS (not the standalone TYPO3Fluid package) also has such a feature:
https://github.com/TYPO3/TYPO3.CMS/blob/cc8532473167975b51f2b6c30e684eafc4e09e81/typo3/sysext/fluid/Classes/Core/ViewHelper/ViewHelperResolver.php#L65

The benefit would be for developers, who often use non-default (non f) ViewHelpers. They would not need to set the namespace in every file any more.
Also, one could argue, that the automatic f namespace integration is already a "spooky action at a distance". So why not extend this, so that everyone can use it?

albe (approved):
I think this makes sense and the argument you brought up @martinlipp is very valid. Also, we already automagically add namespaces for all installed packages (see constructor / https://github.com/neos/flow-development-collection/pull/882). So this just makes this part more flexible, by allowing to specify a global alias.

albe:
@sorenmalling please re-review :)

sorenmalling (approved):

albe:
@bwaidelich do you still see this as problematic or are you okay with adding it?

albe:
Triggered a rebuild with gh actions by reopening

bwaidelich:
> @bwaidelich do you still see this as problematic or are you okay with adding it?

Sure! As mentioned above:
> I think this feature has been added to TYPO3Fluid anyways so there's probably no reason not to use it

sorenmalling:
Lovely, let's merge 🚀

albe:
Thanks a lot @martinlipp !

PR #2381, opened 2021-01-27
FEATURE: Increase PHP 8 compatibility

bwaidelich:
Related: #2233

### DIFF
--- a/Neos.Flow/Classes/Core/LockManager.php
+++ b/Neos.Flow/Classes/Core/LockManager.php
@@ -135,7 +135,9 @@ class LockManager
             fclose($this->lockResource);
             unlink($this->lockPathAndFilename);
         }
-        @unlink($this->lockFlagPathAndFilename);
+        if (file_exists($this->lockFlagPathAndFilename)) {
+            @unlink($this->lockFlagPathAndFilename);
+        }
     }
 
     /**
--- a/Neos.Flow/Classes/Reflection/ParameterReflection.php
+++ b/Neos.Flow/Classes/Reflection/ParameterReflection.php
@@ -57,9 +57,9 @@ class ParameterReflection extends \ReflectionParameter
     public function getBuiltinType()
     {
         $type = $this->getType();
-        if ($type === null || !$type->isBuiltin()) {
+        if (!$type instanceof \ReflectionNamedType) {
             return null;
         }
-        return $type instanceof \ReflectionNamedType ? $type->getName() : (string)$type;
+        return $type->isBuiltin() ? $type->getName() : null;
     }
 }
--- a/Neos.Flow/Classes/SignalSlot/Dispatcher.php
+++ b/Neos.Flow/Classes/SignalSlot/Dispatcher.php
@@ -149,7 +149,7 @@ class Dispatcher
         }
 
         foreach ($this->slots[$signalClassName][$signalName] as $slotInformation) {
-            $finalSignalArguments = $signalArguments;
+            $finalSignalArguments = array_values($signalArguments);
             if (isset($slotInformation['object'])) {
                 $object = $slotInformation['object'];
             } elseif (strpos($slotInformation['method'], '::') === 0) {
--- a/Neos.Flow/Classes/Utility/PhpAnalyzer.php
+++ b/Neos.Flow/Classes/Utility/PhpAnalyzer.php
@@ -82,8 +82,8 @@ class PhpAnalyzer
                     break;
                 }
                 list($type, $value) = $token;
-                if ($type === T_STRING) {
-                    $namespaceParts[] = $value;
+                if ($type === T_STRING || $type === T_NAME_FULLY_QUALIFIED || $type === T_NAME_QUALIFIED) {
+                    $namespaceParts[] = ltrim($value, '\\');
                     continue;
                 }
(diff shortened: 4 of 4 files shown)


Neos.Flow/Classes/Core/LockManager.php
```diff
@@ -135,7 +135,9 @@ public function unlockSite()
             fclose($this->lockResource);
             unlink($this->lockPathAndFilename);
         }
-        @unlink($this->lockFlagPathAndFilename);
+        if (file_exists($this->lockFlagPathAndFilename)) {
+            @unlink($this->lockFlagPathAndFilename);
+        }
```
bwaidelich, line 140: The shutup operator seems to be ignored for PHP 8 - at least I got a Warning that the file didn't exist

Neos.Flow/Classes/Reflection/ParameterReflection.php
```diff
@@ -57,9 +57,9 @@ public function getClass()
     public function getBuiltinType()
     {
         $type = $this->getType();
-        if ($type === null || !$type->isBuiltin()) {
+        if (!$type instanceof \ReflectionNamedType) {
             return null;
         }
-        return $type instanceof \ReflectionNamedType ? $type->getName() : (string)$type;
+        return $type->isBuiltin() ? $type->getName() : null;
```
bwaidelich, line 63: According to https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.reflection
> ReflectionType::isBuiltin() method has been moved to ReflectionNamedType. ReflectionUnionType does not have it.

Not sure if the change addresses this correctly!

Neos.Flow/Classes/SignalSlot/Dispatcher.php
```diff
@@ -149,7 +149,7 @@ public function dispatch(string $signalClassName, string $signalName, array $sig
         }
 
         foreach ($this->slots[$signalClassName][$signalName] as $slotInformation) {
-            $finalSignalArguments = $signalArguments;
+            $finalSignalArguments = array_values($signalArguments);
```
bwaidelich, line 152: Not sure if that's the proper fix, but I got an error
> named argument "$response" doesn't exist

(or similar) because PHP 8 tries to map this to a named argument if the array contains associative keys

Neos.Flow/Classes/Utility/PhpAnalyzer.php
```diff
@@ -82,8 +82,8 @@ public function extractNamespace(): ?string
                     break;
                 }
                 list($type, $value) = $token;
-                if ($type === T_STRING) {
-                    $namespaceParts[] = $value;
+                if ($type === T_STRING || $type === T_NAME_FULLY_QUALIFIED || $type === T_NAME_QUALIFIED) {
+                    $namespaceParts[] = ltrim($value, '\\');
```
bwaidelich, line 86: apparently the behavior of `token_get_all()` has changed..

Neos.Flow/Classes/Reflection/ParameterReflection.php
albe: I think we still need to do the instanceof check, because else we lose compatibility with PHP 7.x. Also, we did not invoke `isBuiltin()` here at all?

albe:
See #2287 - you could add your changes there and would immediately get feedback from the running PHP 8 builds

Neos.Flow/Classes/Reflection/ParameterReflection.php
bwaidelich: the instanceof check is in line 60 now but I'm not sure if 61 should be `return (string)$type` instead

Neos.Flow/Classes/SignalSlot/Dispatcher.php
bwaidelich: FYI: WIth PHP 8.1 this won't be required any longer (https://wiki.php.net/rfc/array_unpacking_string_keys)

Neos.Flow/Classes/Reflection/ParameterReflection.php
albe: In PHP8 there are only `ReflectionNamedType` and `ReflectionUnionType`. Hence the question becomes: How should we handle a union of a builtin type with other types here?
For PHP >= 7.1 only `ReflectionNamedType` exists, so returning `null` should be okay. If we want to be super sure, it should be `PHP < 8 ? (string)$type : null`

Neos.Flow/Classes/SignalSlot/Dispatcher.php
albe: Yeah, but will be some time until we have PHP >= 8.1 minimum :)

Neos.Flow/Classes/Core/LockManager.php
albe: Yep, see https://github.com/neos/flow-development-collection/pull/2287/checks?check_run_id=1974309752#step:12:28
PHP 8 will behave differently and I guess we need to try/catch here (our errorHandler will convert the error to an Exception) or introduce the "double shrug 🤷‍♀️🤷‍♂️()" function.
https://php.watch/versions/8.0/fatal-error-suppression

albe:
FYI: I merged these changes into #2287 prefering yours - so we can close this and continue on the other PR with getting us PHP8 ready

PR #2382, opened 2021-01-31
BUGFIX: Fix ResourceTypeConverter for "data" import

gerdemann:
Fix the ResourceTypeConverter so that you 
can also import resources via base64.

**What I did**
If one wanted to convert a base64 file via "data", this did not work.

**How I did it**
I changed the default value from "resource" to "null" so that the condition fits again.

**How to verify it**
Convert a resource based on an array with "data" and "filename".

**Checklist**


sorenmalling:
Hi @gerdemann 😃 
Thanks for the pull request 👍 

Will you be so kind and translate it to english? And can you provide a test case?

gerdemann:
Sorry for the german text, that was not intentional. 🙈

I can see tomorrow if I can create the missing tests.

gerdemann:
Hi @sorenmalling,

I have tried to add a test. Can you please look at it again? 😄

Neos.Flow/Tests/Unit/ResourceManagement/ResourceTypeConverterTest.php
```diff
@@ -92,7 +92,22 @@ public function convertFromReturnsNullIfSourceArrayIsEmpty()
     public function convertFromReturnsNullIfNoFileWasUploaded()
     {
         $source = ['error' => \UPLOAD_ERR_NO_FILE];
-        $this->assertNull($this->resourceTypeConverter->convertFrom($source, PersistentResource::class));
+        $this->assertTrue($this->resourceTypeConverter->convertFrom($source, PersistentResource::class));
```
sorenmalling, line 95: Why was this test result changed? Did your change break it?

sorenmalling (changes requested):

Neos.Flow/Tests/Unit/ResourceManagement/ResourceTypeConverterTest.php
gerdemann: Sorry, that should not be committed with, is changed now.

kdambekalns:
When did this break, I wonder?

gerdemann:
> When did this break, I wonder?

This has been changed from version 2.2 to 2.3 and has been broken since then.

albe (approved):
I was unsure about the new requirement to specify a filename, but it kind of makes sense, because else the media type also can not be inferred. Looking good from my side, apart from the fact that 4.3 is security-only by now and hence this should target 5.3

sorenmalling:
@gerdemann Care to change it to target 5.3 - then we can get it in :-)

gerdemann:
@sorenmalling Sorry, I had lost sight of that again. I have now adjusted the branch.

albe:
@sorenmalling can you remove your veto? Then I would merge this

sorenmalling (approved):

PR #2389, opened 2021-02-18
BUGFIX: ObjectAccess with direct access on private properties of proxied classes

### LINKED ISSUE #2388: ObjectAccess with direct access fails on private properties of proxied classes
### Description

Classes with private properties do not react as epxected to `ObjectAccess::setProperty()` with `$forceDirectAccess` enabled.

### Steps to Reproduce

Have a class wirh a private property that is proxied for DI/AOP purposes. You end up with a structure like this:

```
class SomeClass_Original {
    private $property = 'initial';
    public function showProperty() {
        echo $this->property;
    }
}
class SomeClass extends SomeClass_Original {
    // DI and AOP code built by Flow
}

$subject = new SomeClass();
```

1. `ObjectAccess::setProperty($subject, 'changed', true)`
2. `$subject->showProperty`

#### Expected behavior

Property is set in "original" class and output is: `changed`

#### Actual behavior

Property is set in proxy class and output is: `initial`

### Affected Versions

Flow: effectively all, the code in question is unchanged for the last 6 years.

### Possible solution

Change https://github.com/neos/flow-development-collection/blob/1a4e6ba72d6e20821a8c7e97a9fe109e6013eccb/Neos.Utility.ObjectHandling/Classes/ObjectAccess.php#L250-L257 to

```php
$className = TypeHandling::getTypeForValue($subject);
if (property_exists($className, $propertyName)) {
    $propertyReflection = new \ReflectionProperty($className, $propertyName);
    $propertyReflection->setAccessible(true);
    $propertyReflection->setValue($subject, $propertyValue);
} elseif (substr_compare(get_parent_class($className), '_Original', -9) === 0 && property_exists(get_parent_class($className), $propertyName)) {
    $propertyReflection = new \ReflectionProperty(get_parent_class($className), $propertyName);
    $propertyReflection->setAccessible(true);
    $propertyReflection->setValue($subject, $propertyValue);
} else {
    $subject->$propertyName = $propertyValue;
}
```

**Question** – do we want this?

kdambekalns:
With this classes with private properties do react as expected to
ObjectAccess::getProperty() and ObjectAccess::setProperty() with
$forceDirectAccess enabled, even when they have been subclassed by the
proxy building of Flow.

Fixes #2388


Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
```diff
@@ -252,6 +258,10 @@ public static function setProperty(&$subject, $propertyName, $propertyValue, boo
                 $propertyReflection = new \ReflectionProperty($className, $propertyName);
                 $propertyReflection->setAccessible(true);
                 $propertyReflection->setValue($subject, $propertyValue);
+            } elseif ($subject instanceof ProxyInterface && property_exists(get_parent_class($className), $propertyName)) {
```
albe, line 261: See above.

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
```diff
@@ -119,6 +120,11 @@ protected static function getPropertyInternal($subject, string $propertyName, bo
                 $propertyReflection->setAccessible(true);
                 return $propertyReflection->getValue($subject);
             }
+            if ($subject instanceof ProxyInterface && property_exists(get_parent_class($className), $propertyName)) {
```
albe, line 123: Would it maybe be more straight-forward if the check for ProxyInterface is done and only overwrites `$className` (inside the forceDirectAccess condition)? The access code is identical otherwise.

albe (approved):
Generally approve. I could imagine a slightly different code-path to the same behaviour, but that's debatable.

albe:
> 1) Neos\Flow\Tests\Unit\Persistence\Generic\SessionTest::getIdentifierByObjectReturnsUuidForObjectBeingAOPProxy
First parameter must either be an object or the name of an existing class

🤔

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
```diff
@@ -114,6 +115,9 @@ protected static function getPropertyInternal($subject, string $propertyName, bo
         $className = TypeHandling::getTypeForValue($subject);
 
         if ($forceDirectAccess === true) {
+            if ($subject instanceof ProxyInterface) {
```
fcool, line 118: Just a question: Isn't this something Neos::TypeHandling should do as well? It does exactly the same thing for Doctrine-Proxies already, which is no "Flow" class, but it does not for our own Proxies? Feels weird.

fcool (commented):
One question inline ;)

Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
bwaidelich: I agree, this should probably be fixed here https://github.com/neos/flow-development-collection/blob/475c71bc569f3900e15eb0a917ce49da819ac8ed/Neos.Utility.ObjectHandling/Classes/TypeHandling.php#L172-L184

albe: Valid point

fcool (approved):
Nice idea to use Class aliases for fine story telling!

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -172,7 +173,7 @@ public static function stripNullableType($type)
     public static function getTypeForValue($value): string
     {
         if (is_object($value)) {
-            if ($value instanceof Proxy) {
+            if ($value instanceof DoctrineProxy || $value insanceof FlowProxy) {
```
bwaidelich, line 176:
```suggestion
            if ($value instanceof DoctrineProxy || $value instanceof FlowProxy) {
```

do you even use an IDE? :)
I saw this on my mobile phone and had to start up my computer to complain. But luckily you had added a test that failed anyways

bwaidelich (dismissed):

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
fcool: Oooops. And proven again: Reading alone is never sufficient - I concentrated on the name and took the key words for granted. :/ Will pay more attention in future.

albe: > do you even use an IDE? :)

Haha, rarely for such supposedly simple changes, but maybe I should. But then again that would likely prevent me from actually doing many of those "supposedly simple changes" at all 🤷‍♂️

kdambekalns:
The main point of this PR is now gone again:
https://github.com/neos/flow-development-collection/pull/2389/commits/7a31ac0c47c94a68427701504138812f6b42cbe0#diff-557a2386bf7a2892685a9f361f85ca740178f53a6d836546fec6a01f8e22fd06R124-R126
That's probably why all the tests started to fail.

albe:
> The main point of this PR is now gone again:
> [7a31ac0#diff-557a2386bf7a2892685a9f361f85ca740178f53a6d836546fec6a01f8e22fd06R124-R126](https://github.com/neos/flow-development-collection/commit/7a31ac0c47c94a68427701504138812f6b42cbe0#diff-557a2386bf7a2892685a9f361f85ca740178f53a6d836546fec6a01f8e22fd06R124-R126)
> That's probably why all the tests started to fail.

But the class is now resolved to `get_parent_class()` for ProxyInterface in `getTypeForValue()` and the remainder of the check is the same:

```
        $className = TypeHandling::getTypeForValue($subject);
        if ($forceDirectAccess === true) {
            if (property_exists($className, $propertyName)) {
                $propertyReflection = new \ReflectionProperty($className, $propertyName);
                $propertyReflection->setAccessible(true);
                return $propertyReflection->getValue($subject);
            }
            if ($subject instanceof ProxyInterface && property_exists(get_parent_class($className), $propertyName)) {
                $propertyReflection = new \ReflectionProperty(get_parent_class($className), $propertyName);
                $propertyReflection->setAccessible(true);
                return $propertyReflection->getValue($subject);
            }
```
should be equivalent to
```
        $className = TypeHandling::getTypeForValue($subject);
        if ($subject instanceof ProxyInterface) {
            $className = get_parent_class($className);
        }
        if ($forceDirectAccess === true) {
            if (property_exists($className, $propertyName)) {
                $propertyReflection = new \ReflectionProperty($className, $propertyName);
                $propertyReflection->setAccessible(true);
                return $propertyReflection->getValue($subject);
            }
```
Or does the `$className` need to be the proxy in some cases?

kdambekalns:
> But the class is now resolved to `get_parent_class()` for ProxyInterface in `getTypeForValue()` and the remainder of the check is the same:

True…

> Or does the `$className` need to be the proxy in some cases?

Nope, properties on the actual proxy should not matter, ever.

I guess the tests just assume things they should not assume about the inner workings here.

kdambekalns:
> > But the class is now resolved to `get_parent_class()` for ProxyInterface in `getTypeForValue()` and the remainder of the check is the same:
> 
> True…

Not true, just checked again. TypeHandling only checks against `Doctrine\ORM\Proxy\Proxy` but my change extends this to the `Neos\Flow\ObjectManagement\Proxy\ProxyInterface` - and private properties are not accessible from the child class…

> > Or does the `$className` need to be the proxy in some cases?
> 
> Nope, properties on the actual proxy should not matter, ever.

And that is why even for the Flow proxies this should be resolved to the parent, IMHO.

kdambekalns:
At least @bwaidelich needs to re-review for this to "legally pass". But maybe the others can have a second look, too.

albe (approved):
Re-approve, especially in light of #2701 this will need to be extended to DependencyProxy as well

bwaidelich (approved):
Makes sense and the solution looks nicer than the one proposed in the related issue. +1 by reading!

BTW: accessing (and even changing) private fields is possible without reflection. Maybe that's something to consider in the future. See: https://3v4l.org/7jUpE#v8.1.4

kdambekalns:
> BTW: accessing (and even changing) private fields is possible without reflection

Ooooh! Nifty!

kdambekalns:
See https://github.com/neos/flow-development-collection/pull/2769 for the next step ;)

kdambekalns:
TBH: I don't get the test failures… will check.

kdambekalns:
Ok, here's why the tests fail:

- I had added a second block to check for the property on the parent class, if the property did not exist on the given class itself.
- @albe moved the check into the `TypeHandling`, which seemed very reasonable.
- The problem now is, that some properties, notably `$Persistence_Object_Identifier` exist _only_ in the proxy, but still need to be read. 🤪

PR #2273, opened 2020-11-21
TASK: Use GH actions for builds

albe:
Still WIP, needs to be backported to lowest maintained branch once finished, but I want to get psalm running too and that's not in <6.0

PR #2390, opened 2021-02-19
TASK: Use GH actions for builds (5.3)

albe:
This is a backport of #2273 with outcommented static analysis and adjusted PHP version matrix.


albe:
This should do now.

.github/workflows/build.yml
```diff
@@ -0,0 +1,205 @@
...
+  build:
+    if: "!contains(github.event.head_commit.message, '[skip ci]') && !contains(github.event.head_commit.message, '[skip travis]')"
+    name: "PHP ${{ matrix.php-versions }} Test ${{ matrix.static-analysis != 'no' && matrix.static-analysis || '' }} (deps: ${{ matrix.dependencies }})"
+
+    continue-on-error: ${{ matrix.experimental }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.1', '7.2', '7.3', '7.4']
```
daniellienert, line 19: 7.1 was not tested by travis. Do we need it here?

daniellienert (approved):
Thanks @albe! 
AFAIS the previous travis matrix seems completely covered (+ PHP 7.1).
PHP test code changes look good by reading

Neos.Cache/composer.json
```diff
@@ -13,7 +13,7 @@
         "neos/utility-opcodecache": "*"
     },
     "require-dev": {
-        "mikey179/vfsstream": "~1.6",
+        "mikey179/vfsstream": "^1.6.1",
```
kdambekalns, line 16: Is this a conscious change (here and elsewhere?) If so, why?

kdambekalns (commented):
Could we maybe extract the memcached changes into their own PR? If we ever need to find them again, nobody will look for them in a CI-related change… 😬

Other than that, awesome!

.github/workflows/build.yml
kdambekalns: Good point. According to https://docs.neos.io/cms/installation-development-setup/system-requirements#php-version-matrix we promise only 72+

Neos.Cache/composer.json
albe: Yes, due to the lowest-dependency test runs. Flow tests won't work with 1.6.0 due to the missing bugfix for `"uninitialized string offset"` (which you reported to vfsstreams btw :))
Lowest dependency tests still don't pass completely, but one issue less at least

.github/workflows/build.yml
albe: According to https://github.com/neos/flow-development-collection/pull/2390/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L10 we did run 7.1 builds on 5.3 and in composer it's also marked as ~7.1 (https://github.com/neos/flow-development-collection/blob/5.3/composer.json#L9), but we can surely strip the build.

Neos.Cache/composer.json
kdambekalns: Got it, thanks!

.github/workflows/build.yml
kdambekalns: No, then let's leave it in!

albe:
> Could we maybe extract the memcached changes into their own PR? If we ever need to find them again, nobody will look for them in a CI-related change… 😬

There you go: https://github.com/neos/flow-development-collection/pull/2393

kdambekalns (approved):

PR #2393, opened 2021-02-22
BUGFIX: Correctly extract port from memcached servers setting

albe:
This fixes the extraction of the port from the `servers` option(s) if they are not bound to the `tcp://` protocol and checks for missing `servers` option in constructor.


albe:
This is just an extraction of the three commits from #2390 - that they do make tests pass can be seen there.

kdambekalns (approved):
Looks good to me…

daniellienert (approved):

PR #2395, opened 2021-02-24
TASK: Correctly extract target branch for pull requests

albe:
For PRs the `GITHUB_REF` points to `refs/pull/$PRNUMBER/merge`, which does not contain the base branch at all.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,7 @@ jobs:
 
     steps:
       - name: Set Flow target branch name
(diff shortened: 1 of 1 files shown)


albe:
Correct behaviour can be checked here: https://github.com/neos/flow-development-collection/pull/2395/checks#step:8:10
The `FLOW_TARGET_VERSION` should always point to the **base branch name** in the flow collection, i.e. `5.3`, `6.0`, ... or `master`

daniellienert (approved):
Everything green, I assume it works :)
Thanks @albe

Issue #2363, opened 2020-12-21
Migration Version20201109224100 changes unrelated code


theilm:
### Description

The migration Version20201109224100 adds the correct return type to the getDescription method of DB migrations to be compatible with the method signature of the abstract class. It searches only for the method name, though, and changes all methods that have the same name. That could break userland code.

#### Expected behavior

Only DB migration files are changed.

#### Actual behavior

All methods with the same name are changed.

sorenmalling:
Hi @theilm - thanks for the report :) Do you think you can provide a fix to the code migration, so it does as expected?

theilm:
Hi @sorenmalling, I created a pull request - I am not 100% sure about the implementation and the correct target branch, so feedback is welcome! :)

PR #2397, opened 2021-02-25
BUGFIX: Prevent wrong code replacements in migration

theilm:
**What I did**
I added the Flow migration namespace to the regex in code migration Version20201109224100. This prevents getDescription methods that are not part of a doctrine migration class to be changed (see #2363).

**How I did it**
I added the namespace to the regex (which makes the regex a bit ugly). I am not sure if it was perhaps better to collect all relevant files (e.g. based on directory structure) instead and apply all searchAndReplace operations only to those files.

**How to verify it**
Create a regular class and a doctrine migration and add a getDescription method without return type signature to both classes. When you execute the code migration, only the method signature of the doctrine migration class should be changed.

### DIFF
--- a/Neos.Flow/Migrations/Code/Version20201109224100.php
+++ b/Neos.Flow/Migrations/Code/Version20201109224100.php
@@ -33,7 +33,7 @@ class Version20201109224100 extends AbstractMigration
     public function up()
     {
         $this->searchAndReplace('Doctrine\DBAL\Migrations\AbstractMigration', 'Doctrine\Migrations\AbstractMigration', ['php']);
-        $this->searchAndReplaceRegex('/public function getDescription\(\)(\s*\{)?$/m', 'public function getDescription(): string $1', ['php']);
+        $this->searchAndReplaceRegex('/(Neos\\\Flow\\\Persistence\\\Doctrine\\\Migrations.+)public function getDescription\(\)(\s*\{)?$/sm', '$1public function getDescription(): string $2', ['php']);
         $this->searchAndReplaceRegex('/public function (up|down|preUp|postUp|preDown|postDown)\(Schema \$schema\)(\s*\{)?$/m', 'public function $1(Schema \$schema): void $2', ['php']);
     }
 }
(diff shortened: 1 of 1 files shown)


Neos.Flow/Migrations/Code/Version20201109224100.php
```diff
@@ -33,7 +33,7 @@ public function getIdentifier()
     public function up()
     {
         $this->searchAndReplace('Doctrine\DBAL\Migrations\AbstractMigration', 'Doctrine\Migrations\AbstractMigration', ['php']);
-        $this->searchAndReplaceRegex('/public function getDescription\(\)(\s*\{)?$/m', 'public function getDescription(): string $1', ['php']);
+        $this->searchAndReplaceRegex('/(Neos\\\Flow\\\Persistence\\\Doctrine\\\Migrations.+)public function getDescription\(\)(\s*\{)?$/sm', '$1public function getDescription(): string $2', ['php']);
```
albe, line 36: Thanks for this! Two questions: Aren't there migrations that do not live in this namespace? Also this will capture cases where the Migrations namespace is imported, but the actual class itself is not a Migration.

Still an improvement over the current loose migration regex

albe (approved):

Neos.Flow/Migrations/Code/Version20201109224100.php
sorenmalling: > Aren't there migrations that do not live in this namespace

In our Doctrine Service class `Neos\Flow\Persistence\Doctrine\Migrations` is defined as the migraiton namespace. I consider this to be a safe approach

> Also this will capture cases where the Migrations namespace

That is a good point. Add a `namespace ` to the start of the line?

theilm: Thanks for the comments! I added `namespace ` to the regex.

albe:
Ah wait, this needs to target 7.0 since that is the version where this migration was added. So this needs a rebase onto 7.0 branch - can you do that @theilm ? Then I would merge this fix

theilm:
I created a [new PR that targets 7.0](https://github.com/neos/flow-development-collection/pull/2409).

PR #2404, opened 2021-03-01
TASK: Support PHP8 types and class property promotion

albe:
This change makes sure that PHP8 union types, nullable types and promoted properties can be used with proxied classes.

Depends on #2287
Related to #2468
Related to #2233


albe:
Apparently StyleCI does not yet support PHP 8 syntax 🤷‍♂️

.github/workflows/build.yml
```diff
@@ -16,19 +16,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.3', '7.4']
+        php-versions: ['8.0']
```
albe, line 19: This is just until we get the tests to correctly ignore the PHP8 implementations in 7.x

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/CompilerTest.php
```diff
@@ -69,27 +69,27 @@ public function annotationsAndStrings()
                 '@\Neos\Flow\Annotations\Session'
             ],
             [
-                new Validate(['value' => 'foo1', 'type' => 'bar1']),
+                new Validate('foo1', 'bar1'),
                 '@\Neos\Flow\Annotations\Validate(type="bar1", argumentName="foo1")'
             ],
             [
-                new Validate(['type' => 'bar1', 'options' => ['minimum' => 2]]),
+                new Validate(null, 'bar1', ['minimum' => 2]),
```
albe, line 76: This shows that the choice to have `argumentName` as the anonymous argument was probably suboptimal. I would rather consider the validate type to be anonmyous - maybe something to consider for next major?

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -51,9 +51,9 @@ public static function parseType(string $type): array
             throw new InvalidTypeException('Found an invalid element type declaration. A type "' . var_export($type, true) . '" does not exist.', 1264093630);
         }
 
-        $typeWithoutNull = self::stripNullableType($matches['type']);
-        $isNullable = $typeWithoutNull !== $matches['type'];
-        $type = self::normalizeType($typeWithoutNull);
+        $typeWithoutNull = self::stripNullableType($type);
```
albe, line 54: $typeWithoutNull/$isNullable never worked as expected - the nullable hint is not included in `$matches['type']`

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -51,9 +51,9 @@ public static function parseType(string $type): array
...
+        $isNullable = $typeWithoutNull !== $type || $type === 'null';
```
albe, line 55: The `$type === 'null'` is a borderline edge-case tbh - no one should annotate something with just the type `null` and in PHP types that is even invalid. Anyway, if you do have `@param null $foo` and check if it is "nullable" I guess the expectation is it to be true?

.github/workflows/build.yml
albe:
```suggestion
        php-versions: ['7.3', '7.4', '8.0']
```

Neos.Utility.ObjectHandling/Tests/Unit/TypeHandlingTest.php
```diff
@@ -225,22 +225,26 @@ public function nullableTypes()
...
             ['null|SplObjectStorage', 'SplObjectStorage'],
             ['Doctrine\Common\Collections\Collection|null', 'Doctrine\Common\Collections\Collection'],
             ['Doctrine\Common\Collections\ArrayCollection|null', 'Doctrine\Common\Collections\ArrayCollection'],
             ['array<\Some\Other\Class>|null', 'array<\Some\Other\Class>'],
             ['ArrayObject<int>|null', 'ArrayObject<int>'],
+            ['?ArrayObject<int>', 'ArrayObject<int>'],
             ['SplObjectStorage<\object>|null', 'SplObjectStorage<\object>'],
             ['Doctrine\Common\Collections\Collection<ElementType>|null', 'Doctrine\Common\Collections\Collection<ElementType>'],
-            ['Doctrine\Common\Collections\ArrayCollection<>|null', 'Doctrine\Common\Collections\ArrayCollection<>'],
+            ['Doctrine\Common\Collections\ArrayCollection<string>|null', 'Doctrine\Common\Collections\ArrayCollection<string>'],
```
albe, line 240: Those were not valid types that can be parsed by `parseType()`

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -24,7 +24,7 @@ abstract class TypeHandling
     /**
      * A property type parse pattern.
      */
-    const PARSE_TYPE_PATTERN = '/^(?:null\|)?\\\\?(?P<type>[a-zA-Z0-9\\\\_]+)(?:<\\\\?(?P<elementType>[a-zA-Z0-9\\\\_]+)>)?(?:\|null)?(?:\s|$)/';
+    const PARSE_TYPE_PATTERN = '/^\??(?:null\|)?\\\\?(?P<type>[a-zA-Z0-9\\\\_]+)(?:<\\\\?(?P<elementType>[a-zA-Z0-9\\\\_]+)>)?(?:\|null)?(?:\s|$)/';
```
albe, line 27: This effectively allows to annotate properties/arguments with `?string` - so in case we do parse PHP return types this would be covered then.

Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
```diff
@@ -598,6 +598,7 @@ protected function evaluatePropertyAnnotations(ORM\ClassMetadataInfo $metadata)
             $mapping['fieldName'] = $property->getName();
             $mapping['columnName'] = strtolower($property->getName());
             $mapping['targetEntity'] = $propertyMetaData['type'];
+            $mapping['nullable'] = $propertyMetaData['nullable'];
```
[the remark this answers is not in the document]
sorenmalling: Is this related to the other changes or is it a separate PR?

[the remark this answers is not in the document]
albe: It was the logical consequence of exposing the `nulllable` information in the ClassSchema, since this is the place we have it for primarily. But it could (and maybe should) go into an own PR that highlights this as an actual feature with a description on how it works/behaves.

Neos.Flow/Classes/I18n/Cldr/CldrParser.php
```diff
@@ -87,9 +87,10 @@ protected function parseNode(\SimpleXMLElement $node)
 
             $parsedChild = $this->parseNode($child);
 
-            if (count($child->attributes()) > 0) {
+            $childAttributes = $child->attributes();
```
albe, line 90: `attributes()` can be `null`

mficzel:
I generally like this but consider it a bit breaky especially since the annotation api changed. Also it would be great to be able to convert the doctrine annotations at the same time.

Not sure about how fast Doctrine will be to support annotations but imho this would be great for Flow8.

albe:
As of doctrine/orm 2.9.x PHP attributes are supported, BUT their changes are breaking for us - see https://github.com/doctrine/orm/issues/8753

albe:
Update: As of doctrine/orm 2.9.3+ - which is allowed now in 7.0+ as of #2496 - PHP 8 attributes work now

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
```diff
@@ -247,18 +250,15 @@ public function render()
...
     {
-        $classDocumentation = "/**\n";
-
         $classReflection = new ClassReflection($this->fullOriginalClassName);
-        $classDescription = $classReflection->getDescription();
-        $classDocumentation .= ' * ' . str_replace("\n", "\n * ", $classDescription) . "\n";
 
-        foreach ($this->reflectionService->getClassAnnotations($this->fullOriginalClassName) as $annotation) {
-            $classDocumentation .= ' * ' . Compiler::renderAnnotation($annotation) . "\n";
+        $classDocumentation = str_replace("*/", "* @codeCoverageIgnore\n */", $classReflection->getDocComment()) . "\n";
```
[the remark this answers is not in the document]
albe: Probably best to extract this part into a separate PR... it's in a single commit already.

albe:
Rebased cleanly now and also extracted the docblock-preserving commit into #2533

albe:
Given the psalm tests fail with the PHP8 methods - maybe it makes more sense to run psalm with latest PHP version, rather than lowest?

kitsunet (approved):
Looks fine to me. Wondering a bit if any of the new features will have impact on eg AOP and if we need to do something about it. but I guess it should be fine for now.

sorenmalling (approved):
From reading it looks fine to me.

I am a bit/very much in doubt how to properly test it deeper, to give valuable feedback

sorenmalling:
@albe With the new `NamedConstructorAnnotation` annotation you mention in #2535 - doesn't it mean that you need to have one merge before the other, and we can't do a release with only one of these merged?

albe:
> @albe With the new `NamedConstructorAnnotation` annotation you mention in #2535 - doesn't it mean that you need to have one merge before the other, and we can't do a release with only one of these merged?

We should have both in optimally, but it only is an issue if you can't install latest doctrine/annotations in your setup for some reason. In that case you will see an error like above (in DocParser) instead of composer rejecting to install things in the worst case.

> I am a bit/very much in doubt how to properly test it deeper, to give valuable feedback

Unless you have some code base using PHP8 types (union / nullable) and/or classes with property promotion, it's impossible to test deeper really :) So no worries. If you have some idea for test fixtures, i.e. exactly such classes (like https://github.com/neos/flow-development-collection/pull/2404/files#diff-d7acfbcf9e4274f911c5c5f5f23ae6f8c8cecf4d64ebcc52d40b04bc050779b3), then that would be a great addition to cover potential cases that break compilation

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
kdambekalns: I guess so…

kdambekalns (approved):

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -276,6 +276,29 @@ protected function stripOpeningPhpTag($classCode)
...
+     * @param \ReflectionAttribute $attribute
+     * @return string
+     */
+    public static function renderAttribute($attribute): string
+    {
+        $attributeAsString = '\\' . $attribute->getName();
+        if (count($attribute->getArguments()) > 0) {
+            $argumentsAsString = [];
+            foreach ($attribute->getArguments() as $argumentName => $argumentValue) {
+                $renderedArgumentValue = var_export($argumentValue, true);
```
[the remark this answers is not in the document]
kdambekalns: Which "other method" do you mean?

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/CompilerTest.php
kdambekalns: No idea on where validation is used more. But I have no issue with `@Flow\Validate("$myArg", type="NotEmpty")`, TBH. What is _not_ straighforward with that?

kdambekalns:
7.3 / lowest fails due to Guzzle vs. Behat…

PR #2406, opened 2021-03-05
BUGFIX: Ignore annotation properties that start with underscore when rendering

### LINKED ISSUE #2387: Undefined index with zircote/swagger-php annotations in class dockblock
### Description

I tried to use the package "zircote/swagger-php" (https://github.com/zircote/swagger-php) to generate a swagger documentation from Flow. Method annotations works well but if I use an annotation like `@OA\Info()` or `@OA\Schema()`, Flow throw me this error

```
Notice: Undefined index: filename in /var/www/gabwebapi.alloboissons.ch/Packages/Framework/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php line 310
Code: 1
File: Packages/Framework/Neos.Flow/Classes/Error/ErrorHandler.php
Line: 81
Open Data/Logs/Exceptions/20210209132528bbf790.txt for a full stack trace.
Exception Code	1355480641
Exception Type	Neos\Flow\Core\Booting\Exception\SubProcessException
Log Reference	202102091325270c9489
Thrown in File	Packages/Framework/Neos.Flow/Classes/Core/Booting/Scripts.php
Line	696
```
### More info

I already talk about that with @albe on Slack and there is his first impression

> How should be deal with annotations like https://github.com/zircote/swagger-php/blob/320ae56fa49f7a305c31e6e75786ce264711073e/src/Annotations/AbstractAnnotation.php#L31 that contain public properties that are automatically created and not supposed to be marked up in the @XX\YY() annotation docblock? I mean, I'm not even entirely sure why we manually rerender the annotation blocks from the instanciated class in the first place - do we ever add properties to annotations in the proxy? (see https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php#L285)

https://neos-project.slack.com/archives/C04V4C6B0/p1612877341056400

### Affected Versions

Flow: 7.0.2

Php: 7.4.3

Swagger-php: 3.1.0

albe:
This fixes an error when using annotations that use internal public properties starting with an underscore, like zircote/swagger-php. This is only a workaround for the more generic problem of rendering annotations that contain publicly accessible internal state. Optimally rendering annotations for proxies would only be based on the information provided by the original annotation, like PHP 8 Attributes do. Alternatively docblocks would not be altered in proxies at all and just copied, but this would potentially be a breaking change.

Resolves #2387

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -290,6 +290,7 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
         $optionValues = get_object_vars($annotation);
         $optionsAsStrings = [];
         foreach ($optionValues as $optionName => $optionValue) {
+            if ($optionName[0] === '_') continue;
             $optionValueAsString = '';
             if (is_object($optionValue)) {
                 $optionValueAsString = self::renderAnnotation($optionValue);
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -290,6 +290,7 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
         $optionValues = get_object_vars($annotation);
         $optionsAsStrings = [];
         foreach ($optionValues as $optionName => $optionValue) {
+            // FIXME: This is a workaround for https://github.com/neos/flow-development-collection/issues/2387
             if ($optionName[0] === '_') {
                 continue;
             }
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -290,7 +290,9 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
         $optionValues = get_object_vars($annotation);
         $optionsAsStrings = [];
         foreach ($optionValues as $optionName => $optionValue) {
-            if ($optionName[0] === '_') continue;
+            if ($optionName[0] === '_') {
+                continue;
+            }
             $optionValueAsString = '';
             if (is_object($optionValue)) {
                 $optionValueAsString = self::renderAnnotation($optionValue);
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,6 +135,7 @@ jobs:
       - name: Setup Flow configuration
         run: |
           rm -f Configuration/Routes.yaml
+          rm -f Configuration/Testing/Settings.yaml
           cat <<EOF >> Configuration/Testing/Settings.yaml
           Neos:
             Flow:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -170,7 +170,7 @@ jobs:
         if: matrix.static-analysis == 'no'
         #if: env.BEHAT == true
         run: |
-          FLOW_CONTEXT=Testing/Behat ./flow behat:setup && ./flow doctrine:create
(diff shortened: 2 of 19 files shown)


bwaidelich:
I fail to understand the issue to be honest. What about a little test case that demonstrates the bug and tests the fix at the same time? :)

albe:
Yeah, well... the thing is that it requires to write an annotation class that I would never suggest to write like that. It's a dirty workaround for dirty solution to a specific issue of a library.

```php
final class MyWeirdAnnotation
{
    public $_internalButPublicProperty;
    public $actualPublicProperty;

    public function __construct($actualPublicProperty)
    {
        $this->actualPublicProperty = $actualPublicProperty;
        $this->_internalButPublicProperty = $someCalculatedState;
    }
}

...
/**
 * @MyWeirdAnnotation(actualPublicProperty="SomeValue")
 */
public function someAnnotatedMethod()
{
}
```
When the proxy builder catches the annotated method, it will spit out the following:
```php
/**
 * @MyWeirdAnnotation(actualPublicProperty="SomeValue", _internalButPublicProperty="SomeCalculatedStateIfSerializableAtAll")
 */
public function someAnnotatedMethod()
{
}
```
In the given case, this will even blow up during proxy building, because the internal state is a class that can not be serialized into the rendered annotation.

I'm not sure we do want to have this behaviour covered by tests - as suggested in #2387 I would reconsider the rendering of annotations in master to **not alter** docblocks on methods instead, which would make this workaround obsolete.

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -290,6 +290,9 @@ public static function renderAnnotation($annotation)
         $optionValues = get_object_vars($annotation);
         $optionsAsStrings = [];
         foreach ($optionValues as $optionName => $optionValue) {
+            if ($optionName[0] === '_') {
```
albe, line 293:
```suggestion
            // FIXME: This is a workaround for https://github.com/neos/flow-development-collection/issues/2387
            if ($optionName[0] === '_') {
```

bwaidelich:
Thanks for the explanation.
So the bug is not really related to the undescore but to the missing matching constructor argument, right? Can't we detect _that_ instead?  Ignoring all options that start with an underscore seems a bit fragile to me.

> I'm not sure we do want to have this behaviour covered by tests

Fair point, I was just looking for a way to make this "hack" a little bit more explicit/easier to understand. A comment seems like an OK solution too.

albe:
> So the bug is not really related to the underscore but to the missing matching constructor argument, right?

Well, I'd rather say the bug is that we render annotations differently then they were declared in the original source (unless they are strict POPOs). We should maybe handle them like PHP8 does for attributes: The `ReflectionAttribute` class only contains the information that was in the original declaration and from that you can create new instances of the Attribute class. But that would be one hell of a change deep in the ReflectionService.

> Can't we detect that instead?

We could *maybe*, with some more reflection magic and comparing class properties against constructor arguments (but what about single `array $values` constructor arguments? So we need to have a special fragile case for that...). I really wouldn't want to do more than absolutely necessary to just get this library working now for old versions and find a proper solution for master ASAP.

bwaidelich:
I think I didn't really get the problem right. Feel free to get in touch sometimes next week if you need some more rubberducking!

albe:
Folks, after a lot of thinking back and forth, I'm going to merge this into 5.3 now and do upmerges for the release. If something unexpected totally blows up, I'll take care of fixing it and take the blame. I don't see this really happening, so I consider the risk far less than blocking the (broken) usage of zircote/swagger-php with 5.3-6.2 versions forever.

I'll prepare a PR that changes the proxy class and method docblocks rendering so that this hack becomes obsolete - we can then discuss if we do want to backport that or only have it in master.

bwaidelich:
Thank you for taking care! <3

PR #2409, opened 2021-03-06
BUGFIX: Prevent wrong code replacements in migration

theilm:
**What I did**
I added the Flow migration namespace to the regex in code migration Version20201109224100. This prevents getDescription methods that are not part of a doctrine migration class to be changed (see #2363).

**How I did it**
I added the namespace to the regex (which makes the regex a bit ugly). I am not sure if it was perhaps better to collect all relevant files (e.g. based on directory structure) instead and apply all searchAndReplace operations only to those files.

**How to verify it**
Create a regular class and a doctrine migration and add a getDescription method without return type signature to both classes. When you execute the code migration, only the method signature of the doctrine migration class should be changed.


albe (approved):

albe:
Tests currently only fail due to #2407 which needs an upmerge. Will merge as is. Thanks @theilm!

PR #2411, opened 2021-03-06
BUGFIX: Check for valid session identifier before accessing the cache

albe:
Currently it is possible to change the cookie value of `Neos_Flow_Session` (`TYPO3_Flow_Session`) to an invalid cache identifier. This leads to an `InvalidArgumentException` and an error code 500. 
This pull requests checks the validity of the session identifier before attempting to access the `metaDataCache`.

Retargeted follow-up to #1132
Thanks @Torsten85 and sorry for being late with accepting this


albe:
This is totally sensible to me. Merging with previous positive feedback in #1132

PR #2412, opened 2021-03-06
FEATURE: Configure MethodPrivilege policies from annotation

### LINKED ISSUE #2060: Allow to specify policies per action through annotations
Related to #2059

> @bwaidelich: As for Permissions I personally think that it is not a good idea to target controller actions at all since they should just delegate I/O..

> @radmiraal: the @Flow\Policy itself sounds like (next to the ordering of course) the easiest one I think.

> @albe:  I'm all for the route thing, but not so sure about policy (policy IMO doesn't follow a 1:1 scheme like route:action)

sorenmalling:
Annotate methods with `@Flow\Policy` to create policy configuration

The annotation format is as follows:

```
    /**
     * This annotation grants the role Neos.Flow:AnnotatedRole to a created privilege for the current method
     *
     * @Flow\Privilege(grantedRoles={"Neos.Flow:AnnotatedRole"})
     */
    public function singleRoleWithGrantedPermissionAction() {}

    /**
     * This annotation created a privilegeId that you can use on your own Policy.yaml configuration
     *
     * @Flow\Privilege(id="Vendor.Package:Edit.Customer")
     */
    public function actionWithPrivilegeIdAnnotationAction() {}
```

 * A single method can only contain one Privilege annotation
 * If both `grantedRoles` and `id` is given, the `id` can not be used in another Annotation
 * If only `id` is given, it can be used in a `Policy.yaml` file for granted several different roles

- [x] Test for roles being added to configuration
- [x] Test for a annotated rule is granted permission to annotated method (somehow get a identifier)
- [x] Test for method with multiple annotations
- [ ] Update documentation 

Resolves #2060


Neos.Flow/Classes/Security/Policy/PolicyService.php
```diff
@@ -314,6 +317,52 @@ public function reset(): void
         $this->roles = [];
     }
 
+    public function configurePolicyAnnotatedMethods(array &$policyConfiguration) {
```
albe, line 320:
```suggestion
    /**
     * Add policy configurations to the provided array based on methods annotated with Policy
     */
    public function configurePolicyAnnotatedMethods(array &$policyConfiguration): void {
```

Neos.Flow/Classes/Security/Policy/PolicyService.php
```diff
@@ -314,6 +317,52 @@ public function reset(): void
...
+            }
+        }
+    }
+
+    /**
+     * @param ObjectManagerInterface $objectManager
+     * @Flow\CompileStatic
+     * @return array
+     */
+    protected static function resolvePolicyAnnotatedMethods(ObjectManagerInterface $objectManager)
```
albe, line 339:
```suggestion
    protected static function resolvePolicyAnnotatedMethods(ObjectManagerInterface $objectManager): array
```

albe:
Needs a merge of master into this (or a rebase), then the tests should run

sorenmalling:
> Needs a merge of master into this (or a rebase), then the tests should run

I'll do that tomorrow along with return type.

I'll try and add more tests (from description) as well

sorenmalling:
Updating docs is required before merging

Neos.Flow/Tests/Functional/Security/Fixtures/Controller/PolicyAnnotatedController.php
```diff
@@ -0,0 +1,42 @@
...
+     * This method gives GRANT permission to the role "Neos.Flow.AnnotatedRole"
+     *
+     * @Flow\Policy(role="Neos.Flow:DeniedRole", permission="deny")
+     * @Flow\Policy(role="Neos.Flow:GrantedRole", permission="grant")
+     * @Flow\Policy(role="Neos.Flow:AbstainedRole", permission="abstain")
+     */
+    public function multipleAnnotationsWithDifferentPermissionsAction()
+    {
+    }
+
```
albe, line 41:
```suggestion
```

albe (approved):
😍 great work, tests nicely cover the feature. Thanks!

bwaidelich:
Nice one! In the docs we should also mention how this plays together with policies from `Policy.yaml` (i.e. what about conflicting rules etc)

sorenmalling:
> what about conflicting rules

What about them? :) can you put some more thought on what about conflicting rules should be covered. From my implementation, a configured rules via Policy.yaml is no different from a annotation, so conflicts should "behave similar"

bwaidelich:
> can you put some more thought on what about conflicting rules should be covered

```yaml
privilegeTargets:
  'Neos\Flow\Security\Authorization\Privilege\Method\MethodPrivilege':
    'Foo':
      matcher: 'method(Some\Package\SomeClass->someMethod()'

roles:
  'Some.Package:SomeRole':
    privileges:
      - privilegeTarget: 'Foo'
        permission: GRANT
```

```php
<?php
namespace Some\Package;

class SomeClass {
  /**
   * @Flow\Policy(role="Some.Package:SomeRole", permission="deny")
   */
   public function someMethod() {
   }
}
```

what wins?

albe:
Good point! That should be documented - from my gut the `DENY` should "win", independent of the position it is defined, because it is a "stronger" policy than `GRANT`. But I'm also unsure if that's really the case.

bwaidelich:
> from my gut the DENY should "win"

I didn't think it through but my gut feeling would be to leave it as is but make sure that the merge-order is such that yaml-based policies take precedence (with the same reasoning than https://github.com/neos/flow-development-collection/pull/2421#discussion_r594295087)

albe:
Not sure if a technical reason is a good decision factor in this specific case, even if it might make sense. Think about what someone would expect if he tried to `deny` access to this action through an annotation, but a configuration level policy (which might even cover multiple methods, so has likely lower "specifity") sets a `grant` and now this action is accessible to people it shouldn't be to. This can become a major issue and even burn reputation of the Framework in the worst-case.

OTOH if the global setting specifies a `deny`, but the method level policy specifies a `grant`, it would still be surprising if the action can not be accessed, but at least it would be **safe by default**.

If that's not possible, I would still prefer an annotation to overrule a global setting, with the same reasoning given for the routes - specifity is higher and what is directly visible contradicting the observed behaviour would be a DX nightmare.

bwaidelich:
> Not sure if a technical reason is a good decision factor in this specific case

But it _wasn't_ a technical reason.
Flow always allowed to overrule policies, routes, configuration, etc. from 3rd party (Flow) packages from within the distribution.
One could argue that it's a different case here, but we have to be sure that it is what we want, the combination works as expected. And we should make it clear that `Policy` annotations can't be overruled.

> I would still prefer an annotation to overrule a global setting [...]
> specifity is higher

Why do you think that specifity is higher?

bwaidelich:
btw: a slightly different approach could be a `Privilege` annotation (instead of `Policy`) that just marks a method/class to be part of a privilege, like so:

```php
<?php
namespace Some\Package;

class SomeClass {

  /**
   * @Flow\Privilege("Some.Package:MyPrivilege")
   */
   public function someMethod() {
   }
}
```

and then leave the actual _policy_ configuration in a (now very simplified) `Polcy.yaml`:

```yaml
roles:
  'Some.Package:SomeRole':
    privileges:
      - privilegeTarget: 'Some.Package:MyPrivilege'
        permission: GRANT
```

That would make a lot of sense to me, since it leaves (application specific) roles out of the equation and allows for defining (implementation specific) method/class <-> privilege assignments in the code.

If you agree to this, I could give this approach a try

sorenmalling:
> btw: a slightly different approach could be a `Privilege` annotation (instead of `Policy`) that just marks a method/class to be part of a privilege, like so:
> 

That was never the intention of the feature/proposal I created, so please don't :-)

bwaidelich:
> That was never the intention of the feature/proposal I created, so please don't :-)

Aye

sorenmalling:
> Flow always allowed to overrule policies, routes, configuration, etc. from 3rd party (Flow) packages from within the distribution.
> 
> One could argue that it's a different case here, but we have to be sure that it is what we want and that the combination works as expected. And we should make it clear that `Policy` annotations can't be overruled.
> 

Good argument, but can we ask a broader audience about this? What have symfony learned from there annotation of routes and policies/roles? What Flow apps out there would use this feature and what do they expect? 

I only know of my own work and this feature is focused on solving the task of configuring access to methods the most efficient way possible

bwaidelich:
> can we ask a broader audience about this?

That's exactly what I tried to do :)

Don't get me wrong. I don't mean to step on anyone's toes and I won't block this if the majority thinks that it makes sense like this. I just want to make sure that we are aware of all the implications because it's much easier to implement a new API/feature than to remove/adjust it lateron.

PRGfx:
I generally agree that the `Policy.yaml` should take priority as this is the best point where we can interact with other packages.

> What Flow apps out there would use this feature and what do they expect?
> I only know of my own work and this feature is focused on solving the task of configuring access to methods the most efficient way possible

For this reasoning I would expect that this feature mainly targets distribution packages. However as soon as there is this possibility it should be expected that other packages use this as well.
From past experience I'd be careful to argue that a library author would always be right when deciding that "no one would ever need to overwrite xy".

> If that's not possible, I would still prefer an annotation to overrule a global setting, with the same reasoning given for the routes - specifity is higher and what is directly visible contradicting the observed behaviour would be a DX nightmare.

I see this point, but this is in some way true throughout the application for every setting (or even object configuration) which might be confusing i.e. when not being careful with package loading order.
On the other hand there might be expectations on how code annotations work in relation to settings:
the [docs on object-management state](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/ObjectManagement.html#configuring-objects-through-annotations)
> It’s up to you defining the scope in the class directly or doing it in a `Objects.yaml` configuration file – both have the same effect. We recommend using annotations in this case, as the scope usually is a design decision which is very unlikely to be changed.

To me it's not clear what would take precedence.


What would be the possibilities for overriding code annotations?
What happens if we extend the class?
Are annotations inherited or have they to be re-applied?

albe:
> Why do you think that specifity is higher?

Specifity in my understanding is "how concrete" a declaration is, i.e. how little room on where it applies. In CSS inline styles always have higher specifity than any amount of class/tag selector, because an inline style can only apply to the single element it is written on. Transferring this to annotations, this rule can only apply to this single method and no other, so it has higher specifity than any global EEL filter expression.

> Flow always allowed to overrule policies, routes, configuration, etc. from 3rd party (Flow) packages from within the distribution.

I consider that a technical reason, but let's not get lost in that line of argument :)

> I just want to make sure that we are aware of all the implications because it's much easier to implement a new API/feature than to remove/adjust it lateron.

👍 agree and I won't ever take your persistent arguing personally or feel bad about it, so don't worry :)

> From past experience I'd be careful to argue that a library author would always be right when deciding that "no one would ever need to overwrite xy".

Definitely true. I always take those `final` decisions in abstractions with a grain of salt. That's why my line of argument would be, that those annotations should not be something that a library/package author should use at all. I see this as a tool for the application developer solely. Again, IMO that would be like a CSS framework using all inline-styles (maybe even with !important - ugh!), but that doesn't mean that inline styles should be overridable by CSS selectors.

> I see this point, but this is in some way true throughout the application for every setting (or even object configuration) which might be confusing i.e. when not being careful with package loading order.

Yes, that's true. And I can totally see this point as a valid argument to "stay consistent with conventions". OTOH that's a "we always did it like this" argument and I would see this specific case (and the other routing case) as a valid point where we could question this approach and maybe conclude that it does not make sense everywhere to be able to override things globally.

I won't block this though if the general agreement is that global overriding is more expected/agreed upon. Just trying to make the API as sane as possible :) So thanks for every feedback!

> a slightly different approach could be a Privilege annotation (instead of Policy)

Hmmm, that could also be useful, but is probably orthogonal to this annotation, which I see one level higher in the abstraction tree - directly below the application layer. `@Flow\Policy(role="Neos.Flow:DeniedRole", permission="deny")` for me means: "I want this method to definitely not be accessible by the `DeniedRole`, no matter what". As you said, those roles are application specific, so this makes sense for application level code only.
In the same way a route annotation only makes sense within the application layer, where you know exactly that this specific action should be accessible under route X. If you want to override this globally, then something is fishy for me.

bwaidelich:
> Specifity in my understanding is "how concrete" a declaration is, i.e. how little room on where it applies.

That's a brilliant definition!
And, frankly, I was playing devils advocate a little: It would be rather confusing if the annotation at the code would be overruled by some "distant" setting.

> those annotations should not be something that a library/package author should use at all

True. It would mean that we should not use that feature in the core packages itself but I think, that's fine.


> `@Flow\Policy(role="Neos.Flow:DeniedRole", permission="deny")` for me means:
> "I want this method to definitely not be accessible by the DeniedRole, no matter what".

My issue with this: Would it mean that it _is accessible_ to anonymous users? At least that wouldn't be totally clear to me by reading the code. While if we keep privileges and permissions separated it is a bit more comprehensible that privileges are *abstained* by default.

>> a slightly different approach could be a Privilege annotation (instead of Policy)

> Hmmm, that could also be useful, but is probably orthogonal to this annotation,
> which I see one level higher in the abstraction tree

Fair enough, maybe my approach isn't pragmatic enough.

But just to avoid misunderstandings:
My line of thinking was to enable devs to easily apply privileges to a set of methods and/or classes without them having to specify new roles etc.

e.g.:
```php
<?php

#[Flow\Privilege("Some.Package:SomePrivilege")]
class SomeApplicationService {}

class SomeOtherClass {
  #[Flow\Privilege("Some.Package:SomePrivilege")]
  public function invokeApplicationLogic() {
    // ...
  }
}
```

which would evaluate to s.th. like
```yaml
privilegeTargets:
  'Neos\Flow\Security\Authorization\Privilege\Method\MethodPrivilege':
    'Some.Package:SomePrivilege':
      matcher: 'method(Some\Package\SomeApplicationService->.*()) || method(Some\Package\SomeOtherClass->invokeApplicationLogic())'
```

internally, protecting all those methods by default.

albe:
> My line of thinking was to enable devs to easily apply privileges to a set of methods and/or classes without them having to specify new roles

Yep, got that :) And I do like that, but as an additional annotation. It would not conflict with global settings, as those would only be additive (or would you somehow allow to `unset` such a privilege?). It could even be applied in a library/package without issues.

> My issue with this: Would it mean that it is accessible to anonymous users? At least that wouldn't be totally clear to me by reading the code.

It would mean that IFF anonymoususer has/inherits the `DeniedRole` - which is another application level knowledge that doesn't fit here really. So that's fine IMHO - you could always specify all roles explicitly to be unambiguous about this:
```
@Flow\Policy(role="Neos.Flow:DeniedRole", permission="deny")
@Flow\Policy(role="Neos.Flow:Anonymous", permission="deny")
```

> While if we keep privileges and permissions separated it is a bit more comprehensible that privileges are abstained by default.

Yeah, indeed the conventions kick in more that way. The Policy annotation completely bypasses those - but as said, I consider this an application-level convenience helper only for very explicit security rules at the point of impact.

sorenmalling:
Only on mobile, sorry for short brief messages..

Thought came up this morning, to keep things closed to what we know today: when we annotate a controller we do it in a package that has a requirement defined (and therefore a "order").

When we build policy configuration from yaml, we iterate the packages in a order defined, so what if we only iterate classes and methods from the current package we are reading the Yaml from? The annotation in the package will create policy.yaml configuration as if it was written in the Yaml file. And packages coming later (because of composer.json requirement) will then do the same and create a result. And can override earlier packages like toda?! 🥳

sorenmalling:
@albe  @bwaidelich @PRGfx 
If my above idea of "bringing it back to as we know it" good to the topics raised about overriding, I will love to create such patch tomorrow 👍

bwaidelich:
I feel bad for "stepping on the brake" all the time, but I'm afraid that it's not yet "bringing it back to as we know it":

Taking your example from above:
Is the method `singleRoleWithGrantPermissionAction()` now accessible to anonymous users or users with other roles? And, if not, is it abstained or denied to those? (I can look it up in the implementation, but this behavior is not explicit IMO).

Also, how would this behavior be _overridden_ in general. E.g. what effective Policy would

```php
class SomeClass {

    /**
     * @Flow\Policy(role="Neos.Flow:AnnotatedRole", permission="grant")
     */
    public function singleRoleWithGrantPermissionAction() {}
}
```

and `Policy.yaml` of
```yaml
privilegeTargets:
  'Neos\Flow\Security\Authorization\Privilege\Method\MethodPrivilege':
    'Some.Package:SomePrivilege':
      matcher: 'method(SomeClass->.*()))'

roles:
  'Neos.Flow:AnnotatedRole':
    privileges:
      - privilegeTarget: 'Some.Package:SomePrivilege'
        permission: ABSTAIN
```

lead to?

(and making that depending on the _location_ of the `Policy.yaml` doesn't make it more explicit IMO)

Just to be clear: I _want_ to be able to specify permissions (and routes btw) more easily and closer to the code and I really appreciate your efforts. I just don't think that we have found the best way yet.

Maybe a solution would be to work with `ABSTAIN` only, i.e. have s.th. like this:

```php
class SomeClass {

    /**
     * @Flow\Privilege(grantedRoles=["Neos.Flow:AnnotatedRole"])
     */
    public function singleRoleWithGrantPermissionAction() {}
}
```
which would work like my suggestion above, but would also immediately add a GRANT permission to the specified roles, effectively creating a Policy of:
```yaml
privilegeTargets:
  'Neos\Flow\Security\Authorization\Privilege\Method\MethodPrivilege':
    '<SomeAutoGeneratedPrivilegeName>':
      matcher: 'method(Some\Package\SomeClass->singleRoleWithGrantPermissionAction()'

roles:
  'Some.Flow:AnnotatedRole':
    privileges:
      - privilegeTarget: '<SomeAutoGeneratedPrivilegeName>'
        permission: GRANT
```

The `SomeAutoGeneratedPrivilegeName` could be created in a deterministic way, so that it is self-explanatory when listing all privileges (e.g. `Some.Package:Annotations:SomeClass_singleRoleWithGrantPermissionAction`).

@sorenmalling do you think that could be a way to tackle this?

albe:
> Taking your example from above:
> Is the method `singleRoleWithGrantPermissionAction()` now accessible to anonymous users or users with other roles? And, if not, is it abstained or denied to those? (I can look it up in the implementation, but this behavior is not explicit IMO).

Just to repeat myself: without knowing the inheritance of the roles, you can not know this ofc. But you can always explicitly specify all (major) roles and if they're granted/denied for this specific method. It's an application level issue. I wouldn't even know how we could solve this?

> Also, how would this behavior be _overridden_ in general. E.g. what effective Policy would
> 
> ```
> class SomeClass {
> 
>     /**
>      * @Flow\Policy(role="Neos.Flow:AnnotatedRole", permission="grant")
>      */
>     public function singleRoleWithGrantPermissionAction() {}
> }
> ```
> 
> and `Policy.yaml` of
> 
> ```yaml
> privilegeTargets:
>   'Neos\Flow\Security\Authorization\Privilege\Method\MethodPrivilege':
>     'Some.Package:SomePrivilege':
>       matcher: 'method(SomeClass->.*()))'
> 
> roles:
>   'Neos.Flow:AnnotatedRole':
>     privileges:
>       - privilegeTarget: 'Some.Package:SomePrivilege'
>         permission: ABSTAIN
> ```
> 
> lead to?

:) I guess you know what my answer would still be on a conceptual level - on the technical level the answer comes down to what would the permissions be for this:
```
     /**
      * @Flow\Policy(role="Neos.Flow:AnnotatedRole", permission="grant")
      * @Flow\Policy(role="Neos.Flow:AnnotatedRole", permission="abstain")
      */
     public function singleRoleWithGrantPermissionAction() {}
```
resp.
```
roles:
  'Neos.Flow:AnnotatedRole':
    privileges:
      - privilegeTarget: 'Some.Package:SomePrivilege'
        permission: GRANT
      - privilegeTarget: 'Some.Package:SomePrivilege'
        permission: ABSTAIN
```
(same question with the second permission being "deny")?

My answers to those two would be:
- "grant" ("abstain" is lower prio/specifity than "grant")
- "deny" ("deny" is higher prio/specifity than "grant")

> Maybe a solution would be to work with `ABSTAIN` only, i.e. have s.th. like this:
> 
> ```
> class SomeClass {
> 
>     /**
>      * @Flow\Privilege(grantedRoles=["Neos.Flow:AnnotatedRole"])
>      */
>     public function singleRoleWithGrantPermissionAction() {}
> }
> ```
> 
> which would work like my suggestion above, but would also immediately add a GRANT permission to the specified roles, effectively creating a Policy of:
> 
> ```yaml
> privilegeTargets:
>   'Neos\Flow\Security\Authorization\Privilege\Method\MethodPrivilege':
>     '<SomeAutoGeneratedPrivilegeName>':
>       matcher: 'method(Some\Package\SomeClass->singleRoleWithGrantPermissionAction()'
> 
> roles:
>   'Some.Flow:AnnotatedRole':
>     privileges:
>       - privilegeTarget: '<SomeAutoGeneratedPrivilegeName>'
>         permission: GRANT
> ```

What would be the difference to the `role="Some.Flow:AnnotatedRole", permission="grant"` annotation example above? Would there also be `deniedRoles`?

bwaidelich:
I struggle to express my point in written form. @sorenmalling @albe Can we find a timeslot to talk about this?

sorenmalling:
> I struggle to express my point in written form. @sorenmalling @albe Can we find a timeslot to talk about this?

Sure - ping me on slack and I'm ready

albe:
as discussed with @sorenmalling and @bwaidelich we agreed to:
- change the annotation to `@Flow\Privilege(grantedRoles=[...])`
- add an `id=` argument to the annotation to be able to specify the generated privilege name, but since this creates new edge cases to take care of (e.g. the id and grantedRoles are exclusive, so you can only use the one or the other) not have that block this for 7.1
- have annotations not be overridable by configuration (annotations should only be used in application level)

bwaidelich (commented):
I didn't test this case, but I assume that the policy caches don't yet react on changes to the annotations in dev context (similar issue with #2421)

Neos.Flow/Classes/Security/Policy/PolicyService.php
```diff
@@ -314,6 +317,61 @@ public function reset(): void
...
+
+                if ($grantedRoles !== []) {
+                    if (isset($privilegesWithGrantedRoles[$privilegeId])) {
+                        throw new \RuntimeException(sprintf('You can not reuse a privilege id for granting roles access on a different method. Privilege id used: "%s". Method annotation: "%s->%s"', $privilegeId, $className, $methodName));
+                    }
+                    $privilegesWithGrantedRoles[$privilegeId] = [
+                       'roles' => $grantedRoles,
+                       'matcher' => $matcher
+                   ];
+                }
```
albe, line 368:
```suggestion
                if ($grantedRoles === []) {
                    $grantedRoles = ['Neos.Flow:Everybody'];
                } elseif (isset($privilegesWithGrantedRoles[$privilegeId])) {
                    throw new \RuntimeException(sprintf('You can not reuse a privilege id for granting roles access on a different method. Privilege id used: "%s". Method annotation: "%s->%s"', $privilegeId, $className, $methodName));
                }
                $privilegesWithGrantedRoles[$privilegeId] = [
                    'roles' => $grantedRoles,
                    'matcher' => $matcher
                ];
```

Neos.Flow/Tests/Functional/Security/Policy/PolicyTest.php
```diff
@@ -47,4 +57,40 @@ public function nonAuthenticatedUsersHaveTheEverybodyAndAnonymousRole()
...
+        self::assertTrue($annotatedRole->getPrivilegeForTarget('Neos.Flow:Granted.Roles.Privilege')->isGranted());
+    }
+
+    /**
+     * @test
+     */
+    public function annotatedPrivilegeWithIdConfiguresPrivilege()
+    {
+        $privilegeTarget = $this->policyService->getPrivilegeTargetByIdentifier('Neos.Flow:Privilege.From.Annotation');
+        self::assertInstanceOf(PrivilegeTarget::class, get_class($privilegeTarget));
```
albe, line 90:
```suggestion
        self::assertInstanceOf(PrivilegeTarget::class, $privilegeTarget);
```

Neos.Flow/Classes/Security/Policy/PolicyService.php
```diff
@@ -314,6 +317,61 @@ public function reset(): void
...
+                    $packageKey = $objectManager->getPackageKeyByObjectName($className);
+                    $privilegeId = sprintf('%s:Privilege.%s', $packageKey, md5($className.'->'.$methodName));
+                }
+                $matcher = sprintf('method(%s->%s())', $className, $methodName);
+
+                if ($grantedRoles === []) {
+                    $grantedRoles = ['Neos.Flow:Everybody'];
+                } elseif (isset($privilegesWithGrantedRoles[$privilegeId])) {
+                    throw new \RuntimeException(sprintf('You can not reuse a privilege id for granting roles access on a different method. Privilege id used: "%s". Method annotation: "%s->%s"', $privilegeId, $className, $methodName));
+                }
```
albe, line 364:
```suggestion
                if (isset($privilegesWithGrantedRoles[$privilegeId])) {
                    if ($grantedRoles !== [] || $privilegesWithGrantedRoles[$privilegeId]['roles'] !== []) {
                        throw new \RuntimeException(sprintf('You can not reuse a privilege id for granting roles access on a different method. Privilege id used: "%s". Method annotation: "%s->%s"', $privilegeId, $className, $methodName));
                    }
                    $matcher = $privilegesWithGrantedRoles[$privilegeId]['matcher'] . ' || ' . $matcher;
                }
```

Neos.Flow/Classes/Security/Policy/PolicyService.php
```diff
@@ -314,6 +317,62 @@ public function reset(): void
...
+                    $packageKey = $objectManager->getPackageKeyByObjectName($className);
+                    $privilegeId = sprintf('%s:Privilege.%s', $packageKey, md5($className.'->'.$methodName));
+                }
+                $matcher = sprintf('method(%s->%s())', $className, $methodName);
+
+                if (isset($privilegesWithGrantedRoles[$privilegeId])) {
+                    if ($grantedRoles !== [] || $privilegesWithGrantedRoles[$privilegeId]['roles'] !== []) {
+                        throw new \RuntimeException(sprintf('You can not reuse a privilege id for granting roles access on a different method. Privilege id used: "%s". Method annotation: "%s->%s"', $privilegeId, $className, $methodName));
+                    }
+                    $matcher = $privilegesWithGrantedRoles[$privilegeId]['matcher'] . ' || ' . $matcher;
```
albe, line 364: This needs a test to verify that multiple methods with a privilegeId are all covered by the matcher.

mficzel:
@sorenmalling do you have plans to work on this for Neos 8.0. I would love to see this feature.

kdambekalns (commented):
After heaving read through the discussion it seems the points raised were solved.

@sorenmalling Could you update this to current master, so the tests pass?

I am reluctant to approve, but don't block this… I think this creates a lot ot potential for confusion and makes understanding the effective policy a lot harder. At the same time the benefit seems small. 🤷‍♂️

kdambekalns:
Added @robertlemke as reviewer, I think this might be important to him…

mficzel:
Still would love this as this avoids the definition of Policy.yaml and Routes yaml that are always a bit cumbersome and "you simply have to know it". 

Some documentation should be added and mention that this is not adviced for libraries.

robertlemke:
Thanks for the pointer and everyone for the nice work and discussion!

What comes to my mind is that old idea we had for Flow years ago: First write the documentation, open the PR and then start coding. Maybe that would have made the goals of this feature clearer from the beginning and explained how it fits into the current concept.

I also wished a few times that this kind of feature exists, so I guess I'll like it in practice. I suggest that we put some extra effort into explaining the relation between the annotations and the Policy.yaml approach. Are you clear on that already? Then let's put that into writing. Otherwise, let's have a discussion on how to explain this in a few easy sentences.

My initial though was, that I'd configure general rules with Policy.yaml and then configure the fine-grained, specific privileges with annotations. The advantage of rules in Policy.yaml are that they can cover even code which has not been written yet. And the advantage of the annotations is that you can easily see by looking at a method how it is protected.

I hope this helps!

PR #1960, opened 2020-03-29
TASK: Psalm failures create a new branch with baseline update

albe:
TODO: Maybe only execute when the last commit was a merge commit and then directly push to the branch?

PR #2413, opened 2021-03-06
TASK: Automatically update psalm baseline after merge

albe:
Followup for #1960 adjusted to our GH actions builds. Needs to check the pushing, as that probably doesn't work as expected yet.


albe:
Note to self, we should probably really move the psalm-baseline into the distribution

albe:
Hard to test without merging... hmmm... guess we need to try it out and fix quickly if things go wrong

PR #2414, opened 2021-03-06
BUGFIX: WithAddedHeader does not overwrite existing header

albe:
Previously the PSR-7 forward compatibility method `withAddedHeader()` would overwrite existing headers and was hence the same as `withHeader()`. This change fixes that by correctly setting the `$overwrite` argument in the delegate method.

See https://github.com/neos/flow-development-collection/pull/2361#discussion_r588867079


sorenmalling (approved):

mficzel (approved):
Fine by reading

PR #2415, opened 2021-03-06
TASK: Add memcache extension to pass psalm checks

albe:


albe:
Merging as trivial build step fix that does not change any functionality

PR #2416, opened 2021-03-06
BUGFIX: Authentication tests don't depend on test running within a second

albe:
This prevents the tests from failing occasionally because the test starts directly before the current second switches.


albe:
Just noticed this has already been (partially) fixed in 6.0 be87613d58a46163ec119a9e1cffa1080103bd27

Issue #2417, opened 2021-03-06
Make Composer autoloader the default


albe:
Since https://github.com/neos/flow-development-collection/pull/925 we have technically deprecated our own class loader. However, we never followed up with making the composer class loader the default, not to mention completely stripping our own after another phase.

I suggest starting with changing

```php
    protected static function useClassLoader(Bootstrap $bootstrap)
    {
        return (!FLOW_ONLY_COMPOSER_LOADER || $bootstrap->getContext()->isTesting());
    }
```

to only return true if `FLOW_ONLY_COMPOSER_LOADER` is actively set to `false` and not setting an environment variable is equal to setting it to `true`. We also need to investigate if testing context really needs our old class loader.

In next major, we can maybe completely strip our own class loader.

albe:
Next step is to completely remove our class loader for next major

PR #2418, opened 2021-03-06
TASK: Disable old class loader in testing context by default

albe:
This makes our own tests only use the old class loader if the `FLOW_ONLY_COMPOSER_LOADER` env var is set to `false`.

Related to #2417


albe:
If this passes all checks, the next step is to completely remove the old class loader for next major.

PR #2419, opened 2021-03-07
TASK: Support header title for table output in console

sorenmalling:
The [Symfony Console Table components](https://symfony.com/doc/current/components/console/helpers/table.html) supports settings a header title for a table.

So should we :-)


albe (approved):
Sure, why not :)

sorenmalling:
@albe This one for 7.1?

albe:
Fine by me - would be cool to get another review from someone. Just not sure who to tag - but the change is really trivial

bwaidelich (approved):

PR #2421, opened 2021-03-07
FEATURE: Configure Routes from annotation

### LINKED ISSUE #2059: Support defining routes on the actions via annotation
See slack thread: https://neos-project.slack.com/archives/C050KKBEB/p1593764548080600

Some excerpt quotes:

> @mficzel: I noticed that quite often people struggle with the configuration of routes and policies that are required to actually use a flow controller. Made even worse by the need to write a setting for the route to be actually included.
Symfony offers configuration via annotation for that and i think @Flow\Route()  or @Flow\Policy˜  could be really helpful and easy to get when placed directly at the controller action. Offcourse you loose some flexibility. Is there a reason why we do’nt offer this as an additional option?

> @bwaidelich: Originally I thought that it would be a good idea to separate routes completely from the controllers so that you can for example change the URLs of a 3rd party package.
And I still think that this is a valid point..
But for the 90+% applications this is probably not a requirement at all and comes with needless complexity..
So I’d be all in for a Route Annotation (maybe even a way to specify the URL directly, so that no Routes.yaml file is required!?
As for Permissions I personally think that it is not a good idea to target controller actions at all since they should just delegate I/O..
But what I would like to see was an easy way to make a Domain Service restricted by default and then whitelist public methods via annotation to certain roles

> @albe: dotnet framework has a similar approach to defining routes and it works relatively well.
Basically: @HttpPost("route/to/action/{with}/{parameters}")

> @mficzel: Configuration should override annotation and for generic packages Routes.yaml should still be used.

> @radmiraal if you hit conflicts having the same route routing to different controllers based on loading order has a certain smell :slightly_smiling_face:

> @albe: I'd rather go the other route: annotation > configuration
If you clearly specify the route an action is supposed to be reachable at, that should not be overridden through some configuration somewhere

sorenmalling:
WIP and not working (will throw a exception in runtime build)

 * Add a signal to the `initializeRoutesConfiguration` method of the router
 * Introduces a resolve method to get controller class and action name
 * Adds it to the routerConfigruation values from the Routes.yaml file available
 * Passes it through the loop of `createRoutesFromConfiguration` to go through checks and validation with the other registered routes

## Question and tasks

- [x] Should the objectManager be injected into the router for the single purpose of using in the compileStatic method introduced?
- [x] Split a controllerClass and action name into the values needed for building configuration (@{package|subpackage|controller|action}) - what method can do that?
- [x] Write tests
- [ ] Update docs

Related #2059

### DIFF
--- a/Neos.Flow/Classes/Annotations/Route.php
+++ b/Neos.Flow/Classes/Annotations/Route.php
@@ -20,11 +20,18 @@ namespace Neos\Flow\Annotations;
 final class Route
 {
     /**
+     * Name
+     *
+     * @var string|null
+     */
+    public $name;
+
+    /**
      * HTTP Methods
      *
-     * @var array
+     * @var array|null
      */
-    public $httpMethods = [];
+    public $httpMethods;
 
     /**
      * URI Pattern
@@ -36,6 +43,22 @@ final class Route
     public $uriPattern;
 
     /**
+     * Format
+     *
+     * Example: html
+     *
+     * @var string|null
+     */
+    public $format;
+
+    /**
+     * Append Exceeding Arguments
+     *
+     * @var bool|null
+     */
+    public $appendExceedingArguments;
+
+    /**
      * @param array $values
      */
     public function __construct(array $values)
@@ -45,6 +68,9 @@ final class Route
         }
 
         $this->uriPattern = $values['uriPattern'];
-        $this->httpMethods = $values['httpMethods'] ?? [];
+        $this->name = $values['name'] ?? null;
+        $this->httpMethods = $values['httpMethods'] ?? null;
+        $this->format = $values['format'] ?? null;
+        $this->appendExceedingArguments = $values['appendExceedingArguments'] ?? null;
     }
 }
--- a/Neos.Flow/Classes/Annotations/Route.php
+++ b/Neos.Flow/Classes/Annotations/Route.php
@@ -34,7 +34,7 @@ final class Route
     public $httpMethods = ['GET'];
 
     /**
-	 * URI Pattern. (Can be given as anonymous argument.)
+     * URI Pattern. (Can be given as anonymous argument.)
      *
      * Example: 'path/to/action/{actionArgument}'
      *
@@ -71,13 +71,13 @@ final class Route
         $this->name = $values['name'] ?? null;
 
         if (isset($values['httpMethods'])) {
-			$this->httpMethods = $values['httpMethods'];
-		}
+            $this->httpMethods = $values['httpMethods'];
+        }
         if (isset($values['format'])) {
-			$this->format = $values['format'];
-		}
+            $this->format = $values['format'];
+        }
         if (isset($values['appendExceedingArguments'])) {
-			$this->appendExceedingArguments = $values['appendExceedingArguments'];
-		}
+            $this->appendExceedingArguments = $values['appendExceedingArguments'];
+        }
     }
 }
--- a/Neos.Flow/Classes/Annotations/Route.php
+++ b/Neos.Flow/Classes/Annotations/Route.php
@@ -27,16 +27,16 @@ final class Route
     public $name;
 
     /**
-     * HTTP Methods
+     * HTTP Methods. Default to GET
      *
-     * @var array|null
+     * @var array
      */
-    public $httpMethods;
+    public $httpMethods = ['GET'];
 
     /**
-     * URI Pattern
+	 * URI Pattern. (Can be given as anonymous argument.)
      *
-     * Example: GRANT
+     * Example: 'path/to/action/{actionArgument}'
      *
      * @var string
      */
@@ -49,28 +49,35 @@ final class Route
      *
      * @var string|null
      */
-    public $format;
+    public $format = 'html';
 
     /**
      * Append Exceeding Arguments
      *
-     * @var bool|null
+     * @var bool
      */
-    public $appendExceedingArguments;
+    public $appendExceedingArguments = false;
 
     /**
      * @param array $values
      */
     public function __construct(array $values)
     {
-        if (!isset($values['uriPattern'])) {
+        if (!isset($values['value']) && !isset($values['uriPattern'])) {
             throw new \InvalidArgumentException('uriPattern is not provided.', 1615113040);
         }
 
-        $this->uriPattern = $values['uriPattern'];
+        $this->uriPattern = $values['uriPattern'] ?? $values['value'];
         $this->name = $values['name'] ?? null;
-        $this->httpMethods = $values['httpMethods'] ?? null;
-        $this->format = $values['format'] ?? null;
-        $this->appendExceedingArguments = $values['appendExceedingArguments'] ?? null;
+
+        if (isset($values['httpMethods'])) {
+			$this->httpMethods = $values['httpMethods'];
+		}
+        if (isset($values['format'])) {
+			$this->format = $values['format'];
+		}
+        if (isset($values['appendExceedingArguments'])) {
+			$this->appendExceedingArguments = $values['appendExceedingArguments'];
+		}
     }
 }
--- a/Neos.Flow/Classes/Annotations/Route.php
+++ b/Neos.Flow/Classes/Annotations/Route.php
@@ -1,0 +1,50 @@
+<?php
+namespace Neos\Flow\Annotations;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+/**
+ * Used to configure Routes from a method
+ *
+ * @Annotation
+ * @Target("METHOD")
+ */
+final class Route
+{
+    /**
+     * HTTP Methods
+     *
+     * @var array
+     */
+    public $httpMethods = [];
+
+    /**
+     * URI Pattern
+     *
+     * Example: GRANT
+     *
+     * @var string
+     */
+    public $uriPattern;
+
+    /**
+     * @param array $values
+     */
+    public function __construct(array $values)
+    {
+        if (!isset($values['uriPattern'])) {
+            throw new \InvalidArgumentException('uriPattern is not provided.', 1615113040);
+        }
+
+        $this->uriPattern = $values['uriPattern'];
+        $this->httpMethods = $values['httpMethods'] ?? [];
+    }
+}
--- a/Neos.Flow/Classes/Mvc/Routing/Router.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Router.php
@@ -22,6 +22,8 @@ use Neos\Flow\Mvc\Exception\InvalidRouteSetupException;
 use Neos\Flow\Mvc\Exception\NoMatchingRouteException;
 use Neos\Flow\Mvc\Routing\Dto\ResolveContext;
 use Neos\Flow\Mvc\Routing\Dto\RouteContext;
+use Neos\Flow\Mvc\Routing\Exception\InvalidAnnotatedMethodException;
+use Neos\Flow\ObjectManagement\Exception\UnknownObjectException;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 use Neos\Flow\Reflection\ReflectionService;
 use Psr\Http\Message\UriInterface;
@@ -280,24 +282,9 @@ class Router implements RouterInterface
         $this->routesCreated = true;
     }
 
-    /**
-     * Emits a signal when the router configuration has been loaded
-     *
-     * This signal can be used to add routes during runtime. In the slot make sure to receive the
-     * $routerConfiguration array by reference so you can alter it.
-     *
-     * @param array $routerConfiguration The router configuration
-     * @return void
-     * @Flow\Signal
-     */
-    protected function emitConfigurationLoaded(array &$routerConfiguration): void
+    public function initializeAnnotatedRoutes(): array
     {
-    }
-
-    public function configureRouteAnnotatedMethods(array &$routerConfiguration): void
-    {
-        $annotatedRoutes = static::resolveRouteAnnotatedMethods($this->objectManager);
-        $routerConfiguration = array_merge(array_values($annotatedRoutes), $routerConfiguration);
+        return static::resolveRouteAnnotatedMethods($this->objectManager);
     }
 
 
@@ -305,6 +292,7 @@ class Router implements RouterInterface
      * @param ObjectManagerInterface $objectManager
      * @Flow\CompileStatic
      * @return array
+	 * @throws InvalidAnnotatedMethodException if a none-Action method is annotated
      */
     protected static function resolveRouteAnnotatedMethods(ObjectManagerInterface $objectManager): array
     {
@@ -319,12 +307,12 @@ class Router implements RouterInterface
             $methodNames = $reflectionService->getMethodsAnnotatedWith($controllerClassName, Flow\Route::class);
             foreach ($methodNames as $methodName) {
                 if (preg_match('/.*Action$/', $methodName) === 0 || $reflectionService->isMethodPublic($controllerClassName, $methodName) === false) {
-                    continue;
+                    throw new InvalidAnnotatedMethodException(sprintf('The method %s->%s is annotated with @Flow\Route. The Route annotation is only meant for *Action methods in your controller', $controllerClassName, $methodName));
                 }
                 $controllerObjectName = $objectManager->getCaseSensitiveObjectName($controllerClassName);
 
                 if ($controllerObjectName === null) {
-                    continue;
+                    throw new UnknownObjectException(sprintf('The object "%s" is not registered.', $controllerClassName), 1618384920);
                 }
                 $controllerPackageKey = $objectManager->getPackageKeyByObjectName($controllerObjectName);
 
@@ -343,42 +331,29 @@ class Router implements RouterInterface
                     $matches
                 );
 
-                $controllerSubpackageKey = $matches['subpackageKey'] ?? null;
+                $controllerSubpackageKey = ($matches['subpackageKey'] !== '') ? $matches['subpackageKey'] : null;
                 $controllerName = $matches['controllerName'];
 
                 $annotation = $reflectionService->getMethodAnnotation($controllerClassName, $methodName, Flow\Route::class);
 
-                $defaults = [];
-                $defaults['@package'] = $controllerPackageKey;
-                if (empty($controllerSubpackageKey) === false) {
-                    $defaults['@subpackage'] = $controllerSubpackageKey;
-                }
-                $defaults['@controller'] = $controllerName;
-                $defaults['@action'] = substr($methodName, 0, -6);
-                if ($annotation->format !== null) {
-                    $defaults['@format'] = $annotation->format;
-                }
+                $defaults = [
+                	'@package' => $controllerPackageKey,
+					'@subpackage' => $controllerSubpackageKey,
+					'@controller' => $controllerName,
+					'@action' => substr($methodName, 0, -6),
+					'@format' => $annotation->format
+				];
 
                 $routeConfiguration = [
-                    'name' => $annotation->name ?? 'Annotated Route',
+                    'name' => $annotation->name ?? sprintf('Annotated Route (%s->%s)', $controllerClassName, $methodName),
                     'uriPattern' => $annotation->uriPattern,
                     'defaults' => $defaults,
+					'httpMethods' => $annotation->httpMethods,
+					'appendExceedingArguments' => $annotation->appendExceedingArguments
                 ];
-
-                $httpMethods = $annotation->httpMethods ?? null;
-                if ($httpMethods !== null && is_array($httpMethods)) {
-                    $routeConfiguration['httpMethods'] = $httpMethods;
-                }
-
-                $appendExceedingArguments = $annotation->appendExceedingArguments ?? null;
-                if ($appendExceedingArguments !== null && is_bool($appendExceedingArguments)) {
-                    $routeConfiguration['appendExceedingArguments'] = $appendExceedingArguments;
-                }
-
                 $annotatedRoutes[] = $routeConfiguration;
             }
         }
-
         return $annotatedRoutes;
     }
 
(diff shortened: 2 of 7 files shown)


Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -267,9 +267,70 @@ protected function createRoutesFromConfiguration()
...
+    }
+
+    public function configureRouteAnnotatedMethods(array &$routerConfiguration): void
+    {
+        $annotatedRoutes = static::resolveRouteAnnotatedMethods($this->objectManager);
+        foreach ($annotatedRoutes as $annotatedRoute) {
+            /**
+             * Split controller class to values for the '@{package|subpackage|controller|action}' values
+             * Add to routes configuration
+             * createRoutesFromConfiguration will take care of validating the configuration (mismatch, httpmethods and uri used before, etc)
```
albe, line 295: See `ActionRequest::setControllerObjectName` for a start. Maybe something can be extracted, maybe something has to be c&p

sorenmalling:
Those added for review are originally commenting on the linked issue :-)

sorenmalling:
Tests fails due to ordering - the Neos.Flow default routes kicks in first

sorenmalling:
Maybe this whole thing should be moved into the RouteConfigurationProcessor or where we have the "position" available (from the Settings.yaml configuration)

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,76 @@
...
+     * @var bool|null
+     */
+    public $appendExceedingArguments;
+
+    /**
+     * @param array $values
+     */
+    public function __construct(array $values)
+    {
+        if (!isset($values['uriPattern'])) {
```
albe, line 66: Let's make `uriPattern` available as anonymous argument too?
```suggestion
        if (!isset($values['value']) && !isset($values['uriPattern'])) {
```

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,76 @@
...
+            throw new \InvalidArgumentException('uriPattern is not provided.', 1615113040);
+        }
+
+        $this->uriPattern = $values['uriPattern'];
```
albe, line 70:
```suggestion
        $this->uriPattern = $values['uriPattern'] ?? $values['value'];
```

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * HTTP Methods
+     *
+     * @var array|null
+     */
+    public $httpMethods;
+
+    /**
+     * URI Pattern
```
albe, line 37:
```suggestion
     * URI Pattern. (Can be given as anonymous argument.)
```

albe (commented):
Looking good so far!

Can we add

```php

/**
 * A route annotated controller fixture
 *
 * @Flow\Route("annotated/controller/{@action}")
 * @Flow\Scope("singleton")
 */
class RouteAnnotatedController extends ActionController
{
    public function someAction(): string
    {
        return 'Hello';
    }
}
```
As a test and make that work?

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -267,9 +276,112 @@ protected function createRoutesFromConfiguration()
...
+     * @Flow\Signal
+     */
+    protected function emitConfigurationLoaded(array &$routerConfiguration): void
+    {
+    }
+
+    public function configureRouteAnnotatedMethods(array &$routerConfiguration): void
+    {
+        $annotatedRoutes = static::resolveRouteAnnotatedMethods($this->objectManager);
+        $routerConfiguration = array_merge($routerConfiguration, array_values($annotatedRoutes));
```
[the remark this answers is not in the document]
bwaidelich: I need to look into this more thoroughly (hopefully later today or tomorrow). But IMO the `Routes.yaml` should have precedence so that the routing can be changed for 3rd party packages that define a default behavior via annotations.

And, related, it should be possible to override single routes (instead of adding more specific ones) I think (we could use the route name for that). Just a rough idea, but IMO this is a critical point in order to avoid non-canonical URL behavior (i.e. two URLs maching the same route)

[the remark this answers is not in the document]
sorenmalling: > And, related, it should be possible to override single routes (instead of adding more specific ones) I think (we could use the route name for that).

Sound legit. It will require us to make a "name" or "identifier" of a route required (or build from a hash of the uriPattern?)

> Just a rough idea, but IMO this is a critical point in order to avoid non-canonical URL behavior (i.e. two URLs matching the same route)

Router::createRoutesFromConfiguration makes sure the same uriPattern is not registered twice

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,76 @@
...
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+/**
+ * Used to configure Routes from a method
+ *
+ * @Annotation
+ * @Target("METHOD")
```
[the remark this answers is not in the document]
sorenmalling: Everything is possible, it's a matter of scope :)

Neos.Flow/Classes/Annotations/Route.php
sorenmalling: Do we have that in a existing annotation i can copy/paste from?

Neos.Flow/Classes/Annotations/Route.php
sorenmalling: What does a anonymous argument look like, when annotating a method?

albe:
```
     * @Flow\Route("path/to/action")
```
instead of
```
     * @Flow\Route(uriPattern="path/to/action")
```
would be possible (while the latter still being allowed).
Also, anonymous arguments can be combined with named arguments:
```
     * @Flow\Route("path/to/action", format="json")
```
It is the equivalent to PHP (8) positional arguments combined with named arguments:
```
function foo($first, $second, $third) { ... }

foo('first-value', second: 'second-value', third: 'third-value');
```
The positional arguments need to come first (though in PHP8 there can be more than one, while in Annotations only one argument can be "anonymous").

Neos.Flow/Classes/Annotations/Route.php
albe: Sure, just take a look at e.g. `IgnoreValidation` - or any other Annotation that takes arguments. Most of them have one argument defined as "anonymous". You can also just commit my suggestions

Neos.Flow/Classes/Mvc/Routing/Router.php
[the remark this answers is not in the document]
albe: > 3rd party packages that define a default behavior via annotations

IMO thats not the use-case for annotation routes and I would not recommend doing that. Defaults should be given in global settings and exceptions be defined inplace. A developer looking at a Controller with a `@Flow\Route("my/path")` annotation should always expect this method to be reachable exactly under that route.

> > Just a rough idea, but IMO this is a critical point in order to avoid non-canonical URL behavior (i.e. two URLs matching the same route)
>
> Router::createRoutesFromConfiguration makes sure the same uriPattern is not registered twice

Non-canonical means that two distinct uriPatterns would match a single action, which would lead to duplicate content (SEO issue). I think we could also check that to some degree, though we probably can't completely prevent that. It's already possible right now, so maybe adding such a check should be an own PR.

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,76 @@
...
+     * @var string|null
+     */
+    public $format;
+
+    /**
+     * Append Exceeding Arguments
+     *
+     * @var bool|null
+     */
+    public $appendExceedingArguments;
```
bwaidelich, line 59:
```suggestion
    public $appendExceedingArguments = false;
```

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -267,9 +276,112 @@ protected function createRoutesFromConfiguration()
...
+     * Emits a signal when the router configuration has been loaded
+     *
+     * This signal can be used to add routes during runtime. In the slot make sure to receive the
+     * $routerConfiguration array by reference so you can alter it.
+     *
+     * @param array $routerConfiguration The router configuration
+     * @return void
+     * @Flow\Signal
+     */
+    protected function emitConfigurationLoaded(array &$routerConfiguration): void
```
bwaidelich, line 293: The public API is `RouterInterface`.. I wonder if we could have the signal there somehow.. But I doubt that's possible :-/

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -267,9 +276,112 @@ protected function createRoutesFromConfiguration()
...
+    {
+    }
+
+    public function configureRouteAnnotatedMethods(array &$routerConfiguration): void
+    {
+        $annotatedRoutes = static::resolveRouteAnnotatedMethods($this->objectManager);
+        $routerConfiguration = array_merge(array_values($annotatedRoutes), $routerConfiguration);
```
bwaidelich, line 300: That's a critical part. I don't think it's enough to just prepend configured routes with the ones from the annotations.

Example:
Imagine some 3rd party package with the following code (and no other routes configured at all):

```php
class SomeController {

  /**
   * @Flow\Route("some/path")
   **/
  public function someAction() {}

}
```

and now I'd like to change the path (without using AOP). So I add some `Routes.yaml` à la:

```yaml
-
  uriPattern: 'some/modified/path'
  defaults:
    '@package': 'Some.Package'
    '@controller': 'SomeController'
    '@action': 'someAction'
```

In theory it should be possible to end up with only _one_ effective route (the one from `Routes.yaml`) in this case. But at least it should be evaluated _before_  the annotation based routes

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -267,9 +276,112 @@ protected function createRoutesFromConfiguration()
...
+        $controllerClassNames = $reflectionService->getAllSubClassNamesForClass(AbstractController::class);
+        foreach ($controllerClassNames as $controllerClassName) {
+            if ($reflectionService->isClassAbstract($controllerClassName)) {
+                continue;
+            }
+
+            $methodNames = $reflectionService->getMethodsAnnotatedWith($controllerClassName, Flow\Route::class);
+            foreach ($methodNames as $methodName) {
+                if (preg_match('/.*Action$/', $methodName) === 0 || $reflectionService->isMethodPublic($controllerClassName, $methodName) === false) {
+                    continue;
```
bwaidelich, line 322: Skipping other methods silently seems fragile to me

bwaidelich (commented):
Nice one! Just a few more thoughts

Neos.Flow/Classes/Annotations/Route.php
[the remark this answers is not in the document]
albe: There would be no default and it would need exist in the `uriPattern` as `{@action}` (or would be appended)

Neos.Flow/Classes/Mvc/Routing/Router.php
albe: Yes, let's throw an exception if a non-Action is annotated. It will only happen at compile-time in production anyway

Neos.Flow/Classes/Mvc/Routing/Router.php
albe: Again, I would not consider this a good use-case. You should not annotate an action in a base package but only in your final application, as that route can then not be overwritten. Specifying the route for an action directly onto the method should be like a `final Route`, or `!important` inline style. If you read this, you know exactly how it will behave, no matter what else may exist somewhere else.

Yes I know that goes somewhat against the implicit Flow convention of having things changeable in global settings, but it would create a developer experience hell in this case.

Neos.Flow/Classes/Mvc/Routing/Router.php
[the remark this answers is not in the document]
bwaidelich: you're right. It's probably not necessary to override single routes specifically. But I still think that it should be possible to add more specific routes via YAML (read: routes that are evaluated first)

Neos.Flow/Classes/Mvc/Routing/Router.php
bwaidelich: It's not a good use case I agree. But I can already see "base" packages that make use of this, e.g. Neos.Media – And IMO that's perfectly fine and we should not forbid that.
But I think that we should be able to _override_ the default behavior (e.g. change the action behind `/neos/media/some-service/foo`)

And this would be easily possible if we just evaluate the annotated routes _after_ the yaml-routes or am I missing something?

Neos.Flow/Classes/Mvc/Routing/Router.php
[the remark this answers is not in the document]
albe: Under which circumstances would you consider a YAML route more specific than one annotated directly on the method? The only case I could think of would be:
- settings route specifies URI constraints/http method and a single controller action (no `{@controller}` and `{@action}` parameter)
- the route on the method specifies no constraint/http method
- the incoming request exactly matches the constraint/http method

So with that we would need to assign specifity levels to package/controller/action/http method/uri constraint, then merge based on that level? Sounds like a lot of complexity, which I would rather add in a second step. In most cases, a Route annotation has higher specifity.

Neos.Flow/Classes/Mvc/Routing/Router.php
sorenmalling: Resolving this based on conclusion in last comment and after call

Neos.Flow/Classes/Mvc/Routing/Router.php
sorenmalling: @albe @bwaidelich Covered in next commit - looking for help on how to test a wrongly annotated method, without breaking all other tests (blocking compiling)

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -267,9 +276,112 @@ protected function createRoutesFromConfiguration()
...
+                }
+                $controllerObjectName = $objectManager->getCaseSensitiveObjectName($controllerClassName);
+
+                if ($controllerObjectName === null) {
+                    continue;
```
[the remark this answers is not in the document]
sorenmalling: Covered in next comit

Neos.Flow/Classes/Annotations/Route.php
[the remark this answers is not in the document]
sorenmalling: Resolving this based on conclusion from call (see comment)

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -267,9 +278,86 @@ protected function createRoutesFromConfiguration()
...
+                }
+                $controllerObjectName = $objectManager->getCaseSensitiveObjectName($controllerClassName);
+
+                if ($controllerObjectName === null) {
+                    throw new UnknownObjectException(sprintf('The object "%s" is not registered.', $controllerClassName), 1618384920);
+                }
+                $controllerPackageKey = $objectManager->getPackageKeyByObjectName($controllerObjectName);
+
+                $matches = [];
+                $subject = substr($controllerObjectName, strlen($controllerPackageKey) + 1);
```
bwaidelich, line 320: While trying to extract this to a factory (as discussed via Slack) I wonder: What is this line supposed to do?
It shortens the controller class name by the length of the package key, but that doesn't have to reflect the namespace

bwaidelich: ..and now I realized that this, too, is taken 1:1 from the `ActionRequest` 😬
Even more a reason to centralize (and fix) this piece of logic.. I'll give it a try

bwaidelich (changes requested):
I would suggest not to all the annotation resolution within the Router since that makes it harder to extend & test.
Instead this part could be done via some custom factory (PoC: https://github.com/sorenmalling/flow-development-collection/pull/2).

But I just realized something differently: The route cache won't be flushed when a route annotation is added/changed/removed. I think we need some additional file monitors for that to work...

albe:
> But I just realized something differently: The route cache won't be flushed when a route annotation is added/changed/removed. I think we need some additional file monitors for that to work...

Can't we just flush the cache when classes with Route annotation are reflected/proxied? Since the annotation routes are compiled statically, they will only change when the class is rewritten. They might change less often, but for making this work that should be fine and flushing route caches on deployment is okayish I think?

bwaidelich:
> Can't we just flush the cache when classes with Route annotation are reflected/proxied?

Yes, I think that's the only viable solution (since FileMonitors only work with file name patterns and can't look _into_ the files).

> what if we create a hook into the ProxyClassBuilder::build() method [...]

I guess that could work, but it might slow down the (already quite slow) proxy building. If we add such a hook, I'd suggest something more fine granular along the lines of:
```php
$someService->registerAnnotationListener(SomeAnnotation::class, $someClosure);
```
that way we wouldn't have to search through all annotations every time

albe:
> > what if we create a hook into the ProxyClassBuilder::build() method [...]
> 
> I guess that could work, but it might slow down the (already quite slow) proxy building. If we add such a hook, I'd suggest something more fine granular along the lines of:
> 
> ```
> $someService->registerAnnotationListener(SomeAnnotation::class, $someClosure);
> ```
> 
> that way we wouldn't have to search through all annotations every time

I just pushed a first PoC at https://github.com/neos/flow-development-collection/pull/2461 - I don't think the cost is so much worse, because the ReflectionService should already be initialized at that time, so it boils down to a (two) array lookups per class per configured annotation. Ofc should be measured/observed.

Neos.Flow/Tests/Unit/Mvc/Routing/RouterTest.php
```diff
@@ -405,6 +409,13 @@ public function routeLoadsRoutesConfigurationFromConfigurationManagerIfNotSetExp
     {
         /** @var Router|\PHPUnit\Framework\MockObject\MockObject $router */
         $router = $this->getAccessibleMock(Router::class, ['dummy']);
+        $mockReflectionService = $this->getMockBuilder(ReflectionService::class)->disableOriginalConstructor()->getMock();
+        $mockReflectionService->expects(self::once())->method('getAllSubClassNamesForClass')->with(AbstractController::class)->willReturn([RouteAnnotatedController::class]);
```
albe, line 413: This test fails because it is still missing a couple of methods to be mocked on the ReflectionService, like `getMethodsAnnotatedWith`, `isMethodPublic` and `getMethodAnnotation` - I'd suggest going with Bastians refactoring to move the configuration stuff out of the `Router`. Then test the configuration building from Reflection in isolation with a Functional test.

kitsunet:
I am not quite sure what the state is here, I guess this still needs some work?

sorenmalling:
> I am not quite sure what the state is here, I guess this still needs some work?

The state is, that I will focus on it for not this august release, but for the release after (where Tiga will be RM). same goes for #2412

kitsunet:
> > I am not quite sure what the state is here, I guess this still needs some work?
> 
> The state is, that I will focus on it for not this august release, but for the release after (where Tiga will be RM). same goes for #2412

Makes sense, thanks! I can certainly see to help because this would make a lot of lifes easier.

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,83 @@
...
+     * Append Exceeding Arguments
+     *
+     * @var bool
+     */
+    public $appendExceedingArguments = false;
+
+    /**
+     * @param array $values
+     */
+    public function __construct(array $values)
```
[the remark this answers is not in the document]
albe:
```suggestion
    public function __construct(string $uriPattern, string $name = null, array $httpMethods = ['GET'], string $format = 'html', bool $appendExceedingArguments = false)
```

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,83 @@
...
+
+        if (isset($values['httpMethods'])) {
+            $this->httpMethods = $values['httpMethods'];
+        }
+        if (isset($values['format'])) {
+            $this->format = $values['format'];
+        }
+        if (isset($values['appendExceedingArguments'])) {
+            $this->appendExceedingArguments = $values['appendExceedingArguments'];
+        }
```
albe, line 81:
```suggestion
        $this->uriPattern = $uriPattern;
        $this->name = $name;
        $this->httpMethods = $httpMethods;
        $this->format = $format;
        $this->appendExceedingArguments = $appendExceedingArguments;
```

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,83 @@
...
+ * source code.
+ */
+
+/**
+ * Used to configure Routes from a method
+ *
+ * @Annotation
+ * @Target("METHOD")
+ */
+final class Route
```
albe, line 20:
```suggestion
 * @NamedArgumentConstructor
 */
#[\Attribute(\Attribute::TARGET_METHOD)]
final class Route
```

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,83 @@
...
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
```
albe, line 13:
```suggestion

use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;

```

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,83 @@
...
+     * @var array
+     */
+    public $httpMethods = ['GET'];
+
+    /**
+     * URI Pattern. (Can be given as anonymous argument.)
+     *
+     * Example: 'path/to/action/{actionArgument}'
+     *
+     * @var string
```
albe, line 41:
```suggestion
     * @var string
     * @Required
```

albe (commented):
Added suggestions to adjust to PHP8 attributes compatibility

mficzel:
@sorenmalling do you have plans to work on this for Neos 8.0. I would love to see this feature.

sorenmalling:
To give this a somewhat life, it's now released as a separate package

https://github.com/wbsply/route-annotation

`composer require websupply/route-annotation`

kdambekalns:
Well, but it seems people were waiting for you on this PR? What was missing here, that could only be solved in a separate package?

sorenmalling:
@kdambekalns Refactoring the router default router implementation, to remove the usage of none-interface methods in the command line

https://github.com/neos/flow-development-collection/pull/2955

This route annotation concept will be updated afterwards, once the configuration is separated from the router

mhsdesign:
Routes via annotation have been implemented via https://github.com/neos/flow-development-collection/pull/3325 and released to 9.0.0

PR #2423, opened 2021-03-07
TASK: Avoid opening database connection if nothing to persist

### LINKED ISSUE #1893: PersistenceManager->persistAll() is called without reason, if no unpersisted changes are present
### Description

Setting up a new project, where there is no persistence credetnails in Settings.yaml i noticed exceptions eing registred, because the registered signal in Neos.Flow/Package.php for `persistAll` was called upon a site change (redirect to a login page).

There is no changes to persist at this time, since no objects are changed. So this is a proposal for checking there is any changes to persist after all

### Steps to Reproduce

1. setup a requestpattern that triggers a redirect
2. See a exception like this in your logs
`20-01-08 12:45:43 96381      ERROR     Neos.Flow            Exception in line 166 of /Users/soren/Projects/.../Packages/Libraries/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php: An exception occurred in driver: SQLSTATE[HY000] [1045] Access denied for user ''@'localhost' (using password: NO) - See also`

#### Expected behavior

Nothing is persisted, so no reason to try and persist anything

#### Actual behavior

The error above

### Affected Versions

Flow: 6

albe:
Also, this avoids calling calculateChangeset() twice for requests that only want allowed objects to be persisted.

Resolves #1893

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
@@ -14,6 +14,7 @@ namespace Neos\Flow\Persistence\Doctrine;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Exception\ConnectionException;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Event\OnFlushEventArgs;
 use Doctrine\ORM\Tools\SchemaTool;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Log\ThrowableStorageInterface;
@@ -65,6 +66,12 @@ class PersistenceManager extends AbstractPersistenceManager
     protected $reflectionService;
 
     /**
+     * A runtime flag to pass the argument of the persistAll() call to the onFlush() event listener
+     * @var boolean
+     */
+    protected $onlyAllowedObjects;
+
+    /**
      * Injects the (system) logger based on PSR-3.
      *
      * @param LoggerInterface $logger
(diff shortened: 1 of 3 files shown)


Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
```diff
@@ -86,23 +93,44 @@ public function injectLogger(LoggerInterface $logger)
...
             }
         }
 
-        if (!$this->entityManager->isOpen()) {
-            $this->logger->error('persistAll() skipped flushing data, the Doctrine EntityManager is closed. Check the logs for error message.', LogEnvironment::fromMethodName(__METHOD__));
-            return;
-        }
-
-        /** @var Connection $connection */
-        $connection = $this->entityManager->getConnection();
```
albe, line 105: This part was not guarded by a check that there are actual changes in the UnitOfWork. Simply calling `$this->hasUnpersistedChanges()` would have invoked the costly `computeChangesets()` twice, or even thrice for safe requests (onlyAllowedObjects = true). With this change, it will always only be invoked once, no matter what and the database reconnect only if there are changes.

robertlemke (commented):
Thanks for the change – it looks good! 

I tested it with an application which doesn't need persistence and no DBAL exception showed up in the logs anymore. I also tested the patch with a Neos site and applied some changes, which were successfully persisted.

I'm not sure if it was related to this change, but I had to configure the injection of Persistence Manager in a type converter of the Neos UI as lazy ([here](https://github.com/neos/neos-ui/blob/master/Classes/TypeConverter/ChangeCollectionConverter.php#L54)), because it is used later in the code ([here](https://github.com/neos/neos-ui/blob/master/Classes/TypeConverter/ChangeCollectionConverter.php#L137)). I wonder how that could have worked before, but at least it didn't when I applied thia patch.

kdambekalns:
> I'm not sure if it was related to this change, but I had to configure the injection of Persistence Manager in a type converter of the Neos UI as lazy […]

Huh, when this is made lazy, it cannot work any longer, right? Because the inject method has a type declaration using the `PersistenceManagerInterface`. Or did you make it non-lazy?

> I wonder how that could have worked before, but at least it didn't when I applied this patch.

Me too, either way…

robertlemke:
> Huh, when this is made lazy, it cannot work any longer, right? Because the inject method has a type declaration using the `PersistenceManagerInterface`. Or did you make it non-lazy?

I meant non-lazy 😎

robertlemke (approved):

albe:
Thanks a lot for the review!

simstern:
I tracked a bug after updating from 5.3.23 to 5.3.24 down to this problem as well. #2452 solves it 👍  Thanks!

Issue #2425, opened 2021-03-08
For a easier scalable setup, DoctrineMigrations folder in Data could be moved to Temporary


sorenmalling:
While creating a setup for the platform.sh hosting, I found my application crashing when spinning up due to the DoctrineMigrations folder was trying to be created on a mount not created.

This was due to `DoctrineMigrations`trying to be created in the root of the `Data` folder - but not before the version was deployed and the migration was about to be executed, and the `getDependencyFactotry` of the Doctrine Service class was called.

I suggest to move the `DoctrineMigrations` folder into the `Temporary` folder og use the `FLOW_PATH_TEMPORARY_BASE`. 

A mount for the temporary (cache for examples) files  is seen created, and if that is already created as a mount, this will lead to know issue with `DoctrineMigrations` is not writable (in a deployed read-only environment)

sorenmalling:
For a reference, these mounts are created for a Flow setup on platform.sh

```
mounts:
  '/Data/Logs':
    source: 'local'
    source_path: 'files/data-logs'
  '/Data/Temporary':
    source: 'local'
    source_path: 'files/data-temporary'
  '/Data/Persistent':
    source: 'local'
    source_path: 'files/data-persistent'
  '/Web/_Resources':
    source: 'local'
    source_path: 'files/web-resources'
  '/Data/DoctrineMigrations':
    source: 'local'
    source_path: 'data/doctrine-migrations'

```

albe:
Related to https://github.com/neos/flow-development-collection/issues/2284

PR #2426, opened 2021-03-08
TASK: Move DoctrineMigrations folder to Temporary Data folder

sorenmalling:
Move DoctrineMigrations folder to the Temporary folder. See related ticket #2425 for explanation on the topic

Resolves: #2425

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Service.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Service.php
@@ -217,7 +217,7 @@ class Service
      */
     protected function getDependencyFactory(): DependencyFactory
     {
-        $migrationsPath = Files::concatenatePaths([FLOW_PATH_DATA, 'Temporary', 'DoctrineMigrations']);
+        $migrationsPath = Files::concatenatePaths([FLOW_PATH_TEMPORARY, 'DoctrineMigrations']);
         if (!is_dir($migrationsPath)) {
             Files::createDirectoryRecursively($migrationsPath);
         }
--- a/Neos.Flow/Classes/Persistence/Doctrine/Service.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Service.php
@@ -217,7 +217,7 @@ class Service
      */
     protected function getDependencyFactory(): DependencyFactory
     {
-        $migrationsPath = Files::concatenatePaths([FLOW_PATH_DATA, 'DoctrineMigrations']);
+        $migrationsPath = Files::concatenatePaths([FLOW_PATH_DATA, 'Temporary', 'DoctrineMigrations']);
         if (!is_dir($migrationsPath)) {
             Files::createDirectoryRecursively($migrationsPath);
         }


akappler:
I’d suggest to make the complete Data  folder path configurable via env variable. This comes handy as more files get written to the Data folder. I did it already for a special Docker container and it works great. Can contribute the code if that's interesting.

bwaidelich:
From [Slack](https://neos-project.slack.com/archives/C050C8FEK/p1615304837007000):
> [...] I’d suggest to make the complete Data  folder path configurable via env variable. This comes handy as also Logs get written and some other stuff. So I made a little change to Neos so that I can change the path to e.g. \tmp\Neos and run it in a Docker container where only temp is writeable.

I second that but would suggest not to introduce a new option for the migrations specifically.
Instead we should use [Environment::getPathToTemporaryDirectory()](https://github.com/neos/flow-development-collection/blob/7d5f2dc9729e69bf93f1fed7ce9ec0a9f4c36aa2/Neos.Flow/Classes/Utility/Environment.php#L84) IMO

akappler:
This is what I've changed in the Neos.Flow Core Bootstrap.php:

```        
      if (!defined('FLOW_PATH_DATA_BASE')) {
            define('FLOW_PATH_DATA', self::getEnvironmentConfigurationSetting('FLOW_PATH_DATA_BASE') ?: FLOW_PATH_ROOT . 'Data/');
      }
``` 

Now you can set the base path of the Data folder via `FLOW_PATH_DATA_BASE`.

albe:
I don't think we need a new ENV var for this, we just need to check if `FLOW_PATH_DATA` is already set in Bootstrap:
```
if (!defined('FLOW_PATH_DATA')) { // <- this if needs to be added
    define('FLOW_PATH_DATA', FLOW_PATH_ROOT . 'Data/');
}
```
And I found a single place in the InstallerScript that still has `Data` hardcoded, so probably that needs adjustment too.

Neos.Flow/Classes/Persistence/Doctrine/Service.php
```diff
@@ -217,7 +217,7 @@ public function runDql(string $dql, int $hydrationMode = \Doctrine\ORM\Query::HY
      */
     protected function getDependencyFactory(): DependencyFactory
     {
-        $migrationsPath = Files::concatenatePaths([FLOW_PATH_DATA, 'DoctrineMigrations']);
+        $migrationsPath = Files::concatenatePaths([FLOW_PATH_DATA, 'Temporary', 'DoctrineMigrations']);
```
albe, line 220:
```suggestion
        $migrationsPath = Files::concatenatePaths([FLOW_PATH_TEMPORARY, 'DoctrineMigrations']);
```

bwaidelich: IIRC `FLOW_PATH_TEMPORARY` already takes `Environment::getPathToTemporaryDirectory()` into account. So +1 for changing the PR as suggested.

The "allow FLOW_PATH_DATA to come from env"-topic should be moved to a separate issue IMO

albe:
Just noticed: Could it be we have an actual bug with defining the FLOW_PATH_TEMPORARY env var?
```php
if (!defined('FLOW_PATH_TEMPORARY_BASE')) {
    define('FLOW_PATH_TEMPORARY_BASE', self::getEnvironmentConfigurationSetting('FLOW_PATH_TEMPORARY_BASE') ?: FLOW_PATH_DATA . '/Temporary');
    $temporaryDirectoryPath = Environment::composeTemporaryDirectoryName(FLOW_PATH_TEMPORARY_BASE, $this->context);
}
```
The define only happens if `FLOW_PATH_TEMPORARY_BASE` is not defined - but if it is, then `FLOW_PATH_TEMPORARY` will **not** be defined

derweba (approved):

derweba:
We have the same issue and i can confirm that moving the DoctrineMigrations folder into the Temporary folder will solve the issue with production environments and restricted Data directories.

sorenmalling:
@albe Dunno about the context of when `FLOW_PATH_TEMPORARY_BASE` should not be set, so can't elaborate on that.

But can we merge what this PR foucsses on? Should it target something else than master?

albe:
Sorry for the derailing of this! I think the question we need to decide on is:
_Should we move the DoctrineMigrations folder to temp or is it enough/better to make the whole data folder adjustable? Or maybe we should do both?_

I'm not sure what the best way forward is since I never had to deal with the issue behind this - so I'll stick to whatever you decide on. If we do decide for the `Data` way I only suggest we make the `FLOW_PATH_DATA` env var setable.

akappler:
Hi @albe,
for my use case, running Neos in AWS Lambda, all folders which needs write access, have to be placed in a special folder. Therefore I'd be very happy if it would be possible to define the `Data` folder path. In this case I don't have to adjust the Bootstrap.php file in my deployment manually. As this would have no effects on current installations, there wouldn't be also any compatibility issues or similar. 
If it helps, I'll adjust the code for that.

kdambekalns:
> > Flow tries to create the folder, if it doesn't exists, totally independent of creating migrations
> 
> OK, weirdo Flow.. :)

I veto that and say: _Weirdo Doctrine!_

Up to 6.3 we did this in `writeMigrationClassToFile()`:

https://github.com/neos/flow-development-collection/blob/6.3/Neos.Flow/Classes/Persistence/Doctrine/Service.php#L622

But now we do it in `getDependencyFactory()`

https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Persistence/Doctrine/Service.php#L222

Which, Sir, is your work: https://github.com/neos/flow-development-collection/pull/2285

😜

TIL: The old code creating the directory can be removed now, it seems…

bwaidelich:
> Which, Sir, is your work: #2285 😜

Weirdo me

PR #2434, opened 2021-03-22
FEATURE: Add rfc6750 bearer token

daniellienert:
Adds the bearer token class to extract a rfc6750 bearer token
from the HTTP header.

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
+++ b/Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
@@ -1,0 +1,79 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Flow\Security\Authentication\Token;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Mvc\ActionRequest;
+use Neos\Flow\Security\Authentication\TokenInterface;
+use Neos\Flow\Security\Exception\AuthenticationRequiredException;
+use Neos\Flow\Security\Exception\InvalidAuthenticationStatusException;
+
+/**
+ * rfc6750 Bearer token (https://tools.ietf.org/html/rfc6750)
+ *
+ * "A security token with the property that any party in possession of the token (a "bearer")
+ * can use the token in any way that any other party in possession of it can. Using a bearer
+ * token does not require a bearer to prove possession of cryptographic key material (proof-of-possession)."
+ */
+class BearerToken extends AbstractToken implements SessionlessTokenInterface
+{
+
+    /**
+     * The password credentials
+     * @var array
+     * @Flow\Transient
+     */
+    protected $credentials = ['bearer' => ''];
+
+    /**
+     * @param ActionRequest $actionRequest
+     * @throws AuthenticationRequiredException
+     * @throws InvalidAuthenticationStatusException
+     */
+    public function updateCredentials(ActionRequest $actionRequest)
+    {
+        $this->setAuthenticationStatus(self::AUTHENTICATION_NEEDED);
+        $httpRequest = $actionRequest->getHttpRequest();
+
+        if (!$httpRequest->hasHeader('Authorization')) {
+            return;
+        }
+
+        $authorizationHeader = $httpRequest->getHeader('Authorization')[0];
+
+        if (!str_starts_with($authorizationHeader, 'Bearer ')) {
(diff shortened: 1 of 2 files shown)


kdambekalns (dismissed):
Looks good to me…

Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        $authorizationHeader = $httpRequest->getHeader('Authorization')[0];
+
+        if (!str_starts_with($authorizationHeader, 'Bearer ')) {
+            $this->setAuthenticationStatus(TokenInterface::NO_CREDENTIALS_GIVEN);
+            throw new AuthenticationRequiredException('Could not extract access token from Authorization header: "Bearer" keyword is missing', 1616222749);
+        }
+
+        $this->bearer = substr($authorizationHeader, strlen('Bearer '));
```
albe, line 53: Q: Isn't the expectation of the `AbstractToken` that the `credentials` array property is set? So that `getCredenttials()` will return it later. Otherwise that method needs to be overridden. Also note that the `credentials` property is `@Flow\Transient` - though this token is probably never stored in the database anyway

kdambekalns: Doh, good point. That would always be empty with this implementation, which is unexpected.

daniellienert: Thanks @albe / @kdambekalns. Adjusted the code as suggested.

kdambekalns (approved):

Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
```diff
@@ -0,0 +1,80 @@
...
+        $this->setAuthenticationStatus(self::AUTHENTICATION_NEEDED);
+        $httpRequest = $actionRequest->getHttpRequest();
+
+        if (!$httpRequest->hasHeader('Authorization')) {
+            return;
+        }
+
+        $authorizationHeader = $httpRequest->getHeader('Authorization')[0];
+
+        if (!str_starts_with($authorizationHeader, 'Bearer ')) {
```
albe, line 55: https://www.php.net/manual/de/function.str-starts-with.php PHP8+ only - unfortunately not quite there yet :)
```suggestion
        if (substr($authorizationHeader, 0,  7) !== 'Bearer ') {
```

albe (commented):
One further issue - can't go minimum PHP 8 just yet ;)

Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
kdambekalns: Duh, that tells me: We need at least a basic linter test in CI, to catch such issues…

albe: Or a (unit) test for every new code ;)

daniellienert: Well, this is on purpose. The method is provided by symfony/polyfill-php80 which is required through symfony/console.

But yeah - it is not a direct dependency of Flow.

albe: Ah yes, the polyfill - well, feel free to make it a (direct) dependency of flow and we can keep this line as is I'd say :)

daniellienert: lol, committed your fix just in this moment ;) Thanks a lot @albe!!

daniellienert: Haha, there is the linting 😛 , Ill take care of the style issue

Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
```diff
@@ -0,0 +1,80 @@
...
+        $this->setAuthenticationStatus(self::AUTHENTICATION_NEEDED);
+        $httpRequest = $actionRequest->getHttpRequest();
+
+        if (!$httpRequest->hasHeader('Authorization')) {
+            return;
+        }
+
+        $authorizationHeader = $httpRequest->getHeader('Authorization')[0];
+
+        if (substr($authorizationHeader, 0,  7) !== 'Bearer ') {
```
albe, line 55:
```suggestion
        if (substr($authorizationHeader, 0, 7) !== 'Bearer ') {
```
Ugh... thanks StyleCI (?)

albe (approved):
Looks good now technically. Of course would be cool to have a docs section on this too, but would merge still

daniellienert:
@albe Thanks for being so relentless challenging my PRs 😂

albe:
I know the feeling when one of the others is so relentless with one of my PRs :) But it is for the better!

PR #2440, opened 2021-03-30
BUGFIX: Allow non-standard status codes in redirects

tvt:
Redirecting from controllers using redirectToUri allows the developer to
set the status code to be used. However the actual status code returned
to the client was always 303.

Changing the order so that statusCode is being handled after redirectUri fixes the issue.

**Checklist**


albe (approved):
Great catch and great PR even with a test :) Thanks a lot!

bwaidelich (approved):

Issue #1824, opened 2019-10-17
Allow registering configuration processing types 


sorenmalling:
### Description

As of now, the Configuration processing types are hardcoded into the `ConfigurationManager`. This is great as of now. This task/feature is about creating the possibility to register processing types via `Package.php` in the same was we can register a new Configuration type (ex. `Views` from Neos.Flow)

### What is possible today

Inside `Package.php` of a package, we can register a new configuration type

```
$configurationManager->registerConfigurationType('Policy');
```

The second parameter allows for setting a `$configurationProcessingType`. These are limited to pre-defined processing types inside the `loadConfiguration` method of ConfigurationManager.

### What we could do

To be able to introduce other configuration types and processingTypes or for separating ex. `Policy.yaml` handling to it's own Security package, we could introduce a API that would allow for

```
$configurationProcessingType = 'PolicyProcessing'; # To keep the current naming of the processing
$configurationManager->registerConfigurationProcessingType($configurationProcessingType, [a somehow implementation]);
```

The **[a somehow implementation]** would be based on a `ConfigurationProcessing` interface with a `process` method.

and then we can use the second parameter of `registerConfigurationType()` like so

```
$configurationManager->registerConfigurationType('Policy', 'PolicyProcessing');
```

#### Actual behavior

Today it's not possible to extend the configuration processing types to more that the current

### Affected Versions

Flow: All

mhsdesign:
Hi ;) Has this been added via: https://github.com/neos/flow-development-collection/pull/2449 ?

bwaidelich:
> Hi ;) Has this been added via: https://github.com/neos/flow-development-collection/pull/2449 ?

yes

PR #2443, opened 2021-04-02
WIP: FEATURE: Add callback configuration processing type

gjwnc:
**What I did**

Modify Flow's ConfigurationManager.php to allow a new configuration processing type called `CONFIGURATION_PROCESSING_TYPE_CALLBACK`. This processing type call's a callback to gather the final configuration used for a `configurationType`.

**What I tried to solve**

The goal was to get variable configuration places in the configurationManager. The dynamic configs I have in mind are:
* remote location, e.g. some cloud (yaml) configs
* dynamically created directories, e.g. for a package which allows the backend user to create new NodeTypes (and/or custom fusion) which are stored in a location that won't get overwritten by the deployment process
* load other than yaml configuration. The callback could preprocess some json and just return this configuration

**Possible Problems**

* Detecting file changes might be more complicated, but I think this is part of the responsibility of the user who needs this kind of feature
* Overwriting other configurationTypes. Again this would be the responsibility of the user to not use configurationTypes like `Policy` or `Objects` and/or to get the package order right and know what they are doing

**How I did it**

* I added the constant `CONFIGURATION_PROCESSING_TYPE_CALLBACK`
* Modified `registerConfigurationType` to allow an optional callback parameter
* Invoke the callback during `loadConfiguration()`

**How to verify it**

The goal was to solve https://github.com/neos/neos-development-collection/issues/3000 which tries to add a custom `NodeTypes` folder to load the configuration.

An example (working) implementation for `Neos.ContentRepository\Classes\Package.php` to load the nodeTypes could look like this:
```
        $dispatcher->connect(ConfigurationManager::class, 'configurationManagerReady', function (ConfigurationManager $configurationManager) {
            $configurationManager->registerConfigurationType('NodeTypes', ConfigurationManager::CONFIGURATION_PROCESSING_TYPE_CALLBACK, true,
                function(YamlSource $configurationSource, $packages, $orderedListOfContextNames, $configurationType, $allowSplitSource){
                    $configuration = [];

                    // NodeTypes Directory Configuration
                    $nodeTypesDirs = [];
                    foreach ($packages as $package) {
                        $nodeTypesDir = $package->getConfigurationPath().$configurationType;
                        if (!\is_dir($nodeTypesDir)) {
                            continue;
                        }
                        $nodeTypesDirs[] = $nodeTypesDir;
                    }
                    if(\is_dir(FLOW_PATH_CONFIGURATION.'/'.$configurationType)){
                        $nodeTypesDirs[] = FLOW_PATH_CONFIGURATION.'/'.$configurationType;
                    }
                    foreach ($orderedListOfContextNames as $contextName) {
                        if(\is_dir(FLOW_PATH_CONFIGURATION . $contextName . '/' . $configurationType)){
                            $nodeTypesDirs[] = FLOW_PATH_CONFIGURATION . $contextName . '/' . $configurationType;
                        }
                    }

                    foreach ($nodeTypesDirs as $nodeTypesDir) {
                        $iterator = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($nodeTypesDir));
                        $allFilesIterator = new \CallbackFilterIterator($iterator, function(\SplFileInfo $fileInfo) {
                            return $fileInfo->isFile() && $fileInfo->getExtension()=='yaml';
                        });
                        /** @var \SplFileInfo $fileInfo */
                        foreach ($allFilesIterator as $fileInfo) {
                            $path = $fileInfo->getPath().DIRECTORY_SEPARATOR.$fileInfo->getBasename('.' . $fileInfo->getExtension());
                            $configuration = Arrays::arrayMergeRecursiveOverrule($configuration, $configurationSource->load($path, $allowSplitSource));
                        }
                    }

                    // Package configuration
                    foreach ($packages as $package) {
                        $configuration = Arrays::arrayMergeRecursiveOverrule($configuration, $configurationSource->load($package->getConfigurationPath() . $configurationType, $allowSplitSource));
                    }
                    $configuration = Arrays::arrayMergeRecursiveOverrule($configuration, $configurationSource->load(FLOW_PATH_CONFIGURATION . $configurationType, $allowSplitSource));

                    // Context configuration
                    foreach ($orderedListOfContextNames as $contextName) {
                        foreach ($packages as $package) {
                            $configuration = Arrays::arrayMergeRecursiveOverrule($configuration, $configurationSource->load($package->getConfigurationPath() . $contextName . '/' . $configurationType, $allowSplitSource));
                        }
                        $configuration = Arrays::arrayMergeRecursiveOverrule($configuration, $configurationSource->load(FLOW_PATH_CONFIGURATION . $contextName . '/' . $configurationType, $allowSplitSource));
                    }

                    return $configuration;
                });
        });
```

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Configuration/ConfigurationManager.php
+++ b/Neos.Flow/Classes/Configuration/ConfigurationManager.php
@@ -112,6 +112,13 @@ class ConfigurationManager
     const CONFIGURATION_PROCESSING_TYPE_APPEND = 'AppendProcessing';
 
     /**
+     * Uses a callback to load the configuration
+     *
+     * @var string
+     */
+    const CONFIGURATION_PROCESSING_TYPE_CALLBACK = 'CallbackProcessing';
+
+    /**
      * Defines which Configuration Type is processed by which logic
      *
      * @var array
@@ -282,7 +289,7 @@ class ConfigurationManager
      * @throws \InvalidArgumentException on invalid configuration processing type
      * @return void
      */
-    public function registerConfigurationType(string $configurationType, string $configurationProcessingType = self::CONFIGURATION_PROCESSING_TYPE_DEFAULT, bool $allowSplitSource = true)
+    public function registerConfigurationType(string $configurationType, string $configurationProcessingType = self::CONFIGURATION_PROCESSING_TYPE_DEFAULT, bool $allowSplitSource = true, \Closure $callback=null)
     {
         $configurationProcessingTypes = [
             self::CONFIGURATION_PROCESSING_TYPE_DEFAULT,
@@ -290,12 +297,16 @@ class ConfigurationManager
             self::CONFIGURATION_PROCESSING_TYPE_POLICY,
             self::CONFIGURATION_PROCESSING_TYPE_ROUTES,
             self::CONFIGURATION_PROCESSING_TYPE_SETTINGS,
-            self::CONFIGURATION_PROCESSING_TYPE_APPEND
+            self::CONFIGURATION_PROCESSING_TYPE_APPEND,
+            self::CONFIGURATION_PROCESSING_TYPE_CALLBACK
         ];
         if (!in_array($configurationProcessingType, $configurationProcessingTypes)) {
             throw new \InvalidArgumentException(sprintf('Specified invalid configuration processing type "%s" while registering custom configuration type "%s"', $configurationProcessingType, $configurationType), 1365496111);
         }
-        $this->configurationTypes[$configurationType] = ['processingType' => $configurationProcessingType, 'allowSplitSource' => $allowSplitSource];
+        if ($configurationProcessingType == self::CONFIGURATION_PROCESSING_TYPE_CALLBACK && $callback === null) {
+            throw new \InvalidArgumentException(sprintf('No callback given for callback processing type'), 1617358429395);
+        }
+        $this->configurationTypes[$configurationType] = ['processingType' => $configurationProcessingType, 'allowSplitSource' => $allowSplitSource, 'callback' => $callback];
     }
 
     /**
@@ -490,6 +501,10 @@ class ConfigurationManager
                     }
                     $this->configurations[$configurationType] = array_merge($this->configurations[$configurationType], $this->configurationSource->load(FLOW_PATH_CONFIGURATION . $contextName . '/' . $configurationType, $allowSplitSource));
                 }
+            break;
+            case self::CONFIGURATION_PROCESSING_TYPE_CALLBACK:
+                $callback = $this->configurationTypes[$configurationType]['callback'];
+                $this->configurations[$configurationType] = $callback->__invoke($this->configurationSource, $packages, $this->orderedListOfContextNames, $configurationType, $allowSplitSource);
             break;
             default:
                 throw new Exception\InvalidConfigurationTypeException('Configuration type "' . $configurationType . '" cannot be loaded with loadConfiguration().', 1251450613);


bwaidelich:
Thanks for this!
I didn't have the chance to look into this one more deeply, but the closure approach makes me think:

What if we got rid of the nasty `configurationProcessingTypes` in favor of a more dynamic approach altogether?
So instead of 
```php
public function registerConfigurationType(string $configurationType, string $configurationProcessingType = self::CONFIGURATION_PROCESSING_TYPE_DEFAULT, bool $allowSplitSource = true, \Closure $callback=null)
```
something like
```php
public function registerConfigurationType(string $configurationType, \Closure $processor)
```

Alternatively, more OOP-style, with a new interface along the lines of
```php
interface ConfigurationType
{
    public function getName(): string;

    public function process(array $source): array;
}
```
(not the final interface!)

and

```php
public function registerConfigurationType(ConfigurationType $type): void
```

Reasoning: The `ConfigurationManager` has been growing in size and complexity and it violates many of the SOLID principles.
Maybe it's about time to extract some of its functionality into a more composable pattern!?

gjwnc:
I also thought that a larger rework of the processing types would be beneficial. Maybe I have a bit time to try to refactor the loadConfiguration() and make use of such a ConfigurationType interface.

If I understand it correct, the goal would be to have pretty much the same functionality, as the configuration manager has today, e.g. the API `loadConfiguration(string $configurationType, array $packages)` stays the same, also the cache handling and such.

The config processing is basically moved to other classes and the default configuration types, which are needed by the [bootstraping code](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Core/Booting/Scripts.php#L195) are moved to the constructor, something like this
```
public function __construct(ApplicationContext $context)
    {
        $this->context = $context;
        ....
        $this->configurationTypes = [
            self::CONFIGURATION_TYPE_CACHES => new ConfigurationType\Caches(),
            self::CONFIGURATION_TYPE_OBJECTS => new ConfigurationType\Objects(),
            self::CONFIGURATION_TYPE_ROUTES => new ConfigurationType\Routes(),
            self::CONFIGURATION_TYPE_POLICY => new ConfigurationType\Policy(),
            self::CONFIGURATION_TYPE_SETTINGS => new ConfigurationType\Settings()
        ];
    }
```

The configuration is then loaded like this
```
    protected function loadConfiguration(string $configurationType, array $packages)
    {
        $this->configurations[$configurationType] = [];
        $this->cacheNeedsUpdate = true;

        $configurationType = $this->resolveConfigurationType($configurationType);
        $this->configurations[$configurationType] = $configurationType->process($packages);
        $this->unprocessedConfiguration[$configurationType] = $this->configurations[$configurationType];
    }
```

sorenmalling:
@gjwnc This can be related to this issue/topic
https://github.com/neos/flow-development-collection/issues/1824 
With the goal to extend the API for configuration registration

gjwnc:
I'm reworking the implementation in favor of the OOP approach and noticed two minor bugs:  
1. Objects.yaml does not allow split source, which should be the case cause of PR #1825. The [configuration processing type](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Configuration/ConfigurationManager.php#L419) just ignore's `$allowSplitSource`.
2. [This](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php#L100) test trait is using the config processing type for the unset, which should be `ConfigurationManager::CONFIGURATION_TYPE_POLICY` instead.

I'll fix this with my rework, but I'm not sure if this might have side effects.

sorenmalling:
> I'll fix this with my rework, but I'm not sure if this might have side effects.

I'm here for review and support - super interesting topic you are pushing on!

sorenmalling:
@gjwnc Can this PR be closed in favor of #2449 ? 😃

gjwnc:
Yes, I'll close this.

PR #2444, opened 2021-04-04
BUGFIX: Support nested fluid variables by using the original fluid method

albe:
With this patch it's possible to use the fluid variable nesting feature as well as a specific accessor per element on the path.

**How to use it**

Create a fluid template with nested variables that access an array value:
```html
{config.{type}.value.{flavor}}
```
Assign an array and the two keys:
```php
$this->view->assign('type', 'thing');
$this->view->assign('flavor', 'yellow');
$this->view->assign('config', ['thing' => ['value' => ['yellow' => 'Okayish']]]);
```


albe:
Rebased follow-up to #2444 to get rid of missleading `Merge '5.3' into '5.0'` commit message

markusguenther (approved):
Looks good and seems to work :)

albe:
Merging as of my previous review on the original PR and above. Thanks @vertexvaar for the PR and @markusguenther for the review :) We need both!

PR #2445, opened 2021-04-04
TASK: Make JsonView datetime format configurable

albe:
This allows to override the badly chosen default format of DateTimeInterface::ISO8601 which is not really compatible to ISO8601 in the JsonView options `datetimeFormat` - see https://www.php.net/manual/en/class.datetimeinterface.php#datetime.constants.iso8601

### DIFF
--- a/Neos.Flow/Tests/Unit/Mvc/View/JsonViewTest.php
+++ b/Neos.Flow/Tests/Unit/Mvc/View/JsonViewTest.php
@@ -152,7 +152,7 @@ class JsonViewTest extends UnitTestCase
      */
     public function testTransformValue($object, $configuration, $expected, $description)
     {
-        $jsonView = $this->getAccessibleMock(Mvc\View\JsonView::class, ['dummy'], [], '', false);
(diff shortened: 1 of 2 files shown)


albe:
This was suggested in #2438

Neos.Flow/Tests/Unit/Mvc/View/JsonViewTest.php
```diff
@@ -152,7 +152,7 @@ public function jsonViewTestData()
      */
     public function testTransformValue($object, $configuration, $expected, $description)
     {
-        $jsonView = $this->getAccessibleMock(Mvc\View\JsonView::class, ['dummy'], [], '', false);
+        $jsonView = $this->getAccessibleMock(Mvc\View\JsonView::class, ['dummy'], [], '');
```
albe, line 155: Note: `false` prevents the constructor from being invoked, which means the default settings from `supportedOptions` are not applied. Therefore settings are empty and lead to accessing null.

albe:
ping @daniellienert - I'd like to decide if we should get this into 5.3 before it goes into security-only with the release

PR #2448, opened 2021-04-07
BUGFIX: Extract doctrine EventListener functionality from PersistenceManager

albe:
With the changes in #2423 the PersistenceManager was registered as a Doctrine EventListener.
Hence, when building the EntityManager, the PersistenceManager got instanciated with a completely new non-lazy EntityManager. This caused errors when trying to persist entities, as they were not known to that instance of the EntityManager.
This change extracts the EventListener behaviour from the PersistenceManager.


albe:
Okay, this should pass tests now. I'll also try to write another test-case that shows the previous wrong behaviour

albe:
So, this should be it - please check this out everyone!
/cc @rkrahnen @fcool @sorenmalling @lisakampert

Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * @Flow\Inject
+     * @var AllowedObjectsContainer
+     */
+    protected $allowedObjects;
+
+    /**
+     * @Flow\Inject(lazy=true)
+     * @var PersistenceManagerInterface
```
albe, line 56: Yes, this is odd and somehow goes against the "recursive dependency" explanation in the class header, but we need it for `getIdentifierByObject()` in the exception message and due to the lazy (made explicit) injection works.

Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
```diff
@@ -98,51 +90,11 @@ public function persistAll($onlyAllowedObjects = false)
             return;
         }
 
-        $this->onlyAllowedObjects = $onlyAllowedObjects;
+        $this->allowedObjects->checkNext($onlyAllowedObjects);
         $this->entityManager->flush();
-        $this->onlyAllowedObjects = false;
```
albe, line 103: This false assignment here was weird anyway. Having that "one-time" logic encapsulated in the `AllowedObjectsContainer` is an improvement.

Neos.Flow/Tests/Functional/Persistence/PersistenceTest.php
```diff
@@ -39,11 +39,17 @@ class PersistenceTest extends FunctionalTestCase
...
+     * @var EntityManagerInterface
+     */
+    protected $earlyEntityManager;
+
     /**
      * @return void
      */
     public function setUp()
     {
+        $this->earlyEntityManager = self::$bootstrap->getObjectManager()->get(EntityManagerInterface::class);
```
albe, line 52: The order of instanciation mattered for the bug.

Neos.Flow/Tests/Unit/Persistence/Doctrine/PersistenceManagerTest.php
```diff
@@ -78,7 +77,19 @@ public function setUp()
         $this->mockSystemLogger = $this->createMock(LoggerInterface::class);
         $this->persistenceManager->injectLogger($this->mockSystemLogger);
 
-        $this->inject($this->persistenceManager, 'throwableStorage', $this->getMockBuilder(ThrowableStorageInterface::class)->getMock());
+        $mockThrowableStorage = $this->getMockBuilder(ThrowableStorageInterface::class)->getMock();
+        $this->inject($this->persistenceManager, 'throwableStorage', $mockThrowableStorage);
+
+        $allowedObjectsContainer = new AllowedObjectsContainer();
+        $this->inject($this->persistenceManager, 'allowedObjects', $allowedObjectsContainer);
+        $allowedObjectsListener = new AllowedObjectsListener();
```
albe, line 85: Not a mock, because we actually depend on the code in the listener being executed.

Neos.Flow/Classes/Persistence/AbstractPersistenceManager.php
```diff
@@ -117,26 +110,6 @@ public function allowObject($object)
         $this->allowedObjects->attach($object);
     }
 
-    /**
```
albe, line 120: Moved to Listener

Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
```diff
@@ -0,0 +1,118 @@
...
+    protected $persistenceManager;
+
+    /**
+     * Doctrine onFlush listener that checks for only allowed objects and reconnects
+     * if the database connection was closed.
+     *
+     * @param OnFlushEventArgs $args
+     * @throws PersistenceException
+     */
+    public function onFlush(OnFlushEventArgs $args)
```
albe, line 67: 1:1 copy from the PersistenceManager apart from the `$this->allowedObjects->shouldCheck()` check

albe:
PS: The alternative (and much simpler) solution is to make the `PersistenceManager::$entityManager` injection lazy, but https://github.com/neos/flow-development-collection/commit/c160af4051a5118b1a5b5f65618b77d22e92f334

Neos.Flow/Classes/Persistence/AllowedObjectsContainer.php
```diff
@@ -0,0 +1,50 @@
+<?php
+namespace Neos\Flow\Persistence;
```
daniellienert, line 2:
```suggestion
declare(strict_types=1);

namespace Neos\Flow\Persistence;
```

Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
```diff
@@ -0,0 +1,118 @@
+<?php
+namespace Neos\Flow\Persistence\Doctrine;
```
daniellienert, line 2:
```suggestion
declare(strict_types=1);

namespace Neos\Flow\Persistence\Doctrine;
```

Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
daniellienert: Would it be better to get the PersistenceManager from the objectManager only in case of that exception rendering?

fcool: What do you think would be the difference? As far as I can see, it would be exactly the same - but with a not so obvious alternative dependency. (ObjectManager instead of PersistenceManager)

rkrahnen:
> PS: The alternative (and much simpler) solution is to make the `PersistenceManager::$entityManager` injection lazy, but [c160af4](https://github.com/neos/flow-development-collection/commit/c160af4051a5118b1a5b5f65618b77d22e92f334)

To be honest I do not like using non-lazy injection to cover the execution of a single test. It sounds a bit weird to me, that it is no problem when all functional tests are executed, so maybe we should take a closer look on that?

As @albe said, this solution would be much simpler, so I would prefer that, but I did not work with the tests until now and so I am not able to help investigating that. So if no one else has the time for that, let us go with the fix currently implemented to get it done as fast as possible. :-)

kdambekalns:
> To be honest I do not like using non-lazy injection to cover the execution of a single test.

The problem exposed by this test is very real, even though it rarely appears in the wild. It just cannot work – injecting a lazy proxy into a prpoerty with a type declarartion must fail on a PHP level.

rkrahnen:
> injecting a lazy proxy into a prpoerty with a type declarartion must fail on a PHP level

I absolutely agree on that point! But I think I am missing where the property has a type declaration, it is `protected $entityManager;` in the commit where the non-lazy was added and is still in the current master.
But let us not waist time here, if necessary at some point, I will find out or come back to you.

kdambekalns:
> I am missing where the property has a type declaration

Right, it doesn't. But it is passed on to a constructor having a type declaration (at least that's the case solved in https://github.com/neos/flow-development-collection/commit/c160af4051a5118b1a5b5f65618b77d22e92f334) eventually.

Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
kdambekalns: The OM could be used to fetch the dependency only when needed, not immediately when building the object instance.

robertlemke:
> > I am missing where the property has a type declaration
> 
> Right, it doesn't. But it is passed on to a constructor having a type declaration (at least that's the case solved in [c160af4](https://github.com/neos/flow-development-collection/commit/c160af4051a5118b1a5b5f65618b77d22e92f334)) eventually.

Actually, that could have been solved differently (in https://github.com/neos/flow-development-collection/blob/c160af4051a5118b1a5b5f65618b77d22e92f334/Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php#L271). Instead of disabling lazy injection altogether, the Persistence Manager could have done:

```php
        if ($this->entityManager instanceof DependencyProxy) {
            $this->entityManager->_activateDependency();
        }
       $schemaTool = new SchemaTool($this->entityManager);
```

rkrahnen:
Another thing @fcool and I are unsure about in #2423 is the setting:
```yaml
          'Neos\Flow\Persistence\Doctrine\PersistenceManager':
            events: ['onFlush']
            listener: 'Neos\Flow\Persistence\Doctrine\PersistenceManager'
```
Is it correct to use the specific class and not the interface, or could this lead to the same problem when someone uses an own implementation of the interface?

Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
daniellienert: That's what I meant

kdambekalns:
> Instead of disabling lazy injection altogether, the Persistence Manager could have done: […]

Yeah, but that seemed "too internal", somehow. 🤷‍♂️

Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
albe: Yeah, would basically be a manually coded lazy dependency. Would be fine for me too, though I always cringe a bit and ask "why was this necessary instead of injection?" when using the objectmanager manually.

albe:
> Another thing @fcool and I are unsure about in #2423 is the setting:
> 
> ```yaml
>           'Neos\Flow\Persistence\Doctrine\PersistenceManager':
>             events: ['onFlush']
>             listener: 'Neos\Flow\Persistence\Doctrine\PersistenceManager'
> ```
> 
> Is it correct to use the specific class and not the interface, or could this lead to the same problem when someone uses an own implementation of the interface?

Well, it's not an issue really, because the PersistenceManager there is used as an EventListener and the `onFlush()` method does not depend on the state of the PersistenceManager (only the logger injections, but even the EntityManager is directly used from the event args). So they could even be two distinct instances. But you are right that this is not so obvious and hence this change to extract the listener makes sense even without it fixing a bug.

> Actually, that could have been solved differently (in line 271). Instead of disabling lazy injection altogether, the Persistence Manager could have done:

Hmmm. That would be a good way to solve the issue. Yes, that's ugly code, but it makes the general issue we have with lazy dependencies and type hints visible (and easier to spot later on once we have a better solution).
So I would maybe even suggest going both ways:
- lazy injection of EntityManager + `instanceof DependencyProxy` check (separate PR)
- extract EventListener into own class

Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
albe: After thinking a bit more about it, I would still keep the annotation. The reason is this: using the static object manager to get the instance would create a very hard coded dependency and harder to test code. Injecting the ObjectManager would couple this class to our specific DI implementation. With the annotation all this is left as a transparent implementation of our proxy compilation - which I'd like to replace with ocramius/proxymanager at some point.

Neos.Flow/Classes/Persistence/AllowedObjectsContainer.php
```diff
@@ -0,0 +1,52 @@
...
+     *
+     * @param bool $checkNext
+     */
+    public function checkNext($checkNext = true)
+    {
+        $this->checkNext = $checkNext;
+    }
+
+    /**
+     * Returns true if allowed objects should be checked this time and resets the internal flag to false,
```
kdambekalns, line 41: The one question I have: Why the reset? That wasn't the case in the "original" implementation…

kdambekalns (commented):
Looks good, just one question…

Neos.Flow/Classes/Persistence/AllowedObjectsContainer.php
albe: See https://github.com/neos/flow-development-collection/pull/2448/files#r608769087 - it prevents that invoking `persistAll(true)` once will make the `true` leak through to subsequent `EntityManager->flush()` invocations

kdambekalns (approved):

fcool (approved):

albe:
❤️

kdambekalns:
Doh, the Psalm tests sometimes find real errors. We should listen. :)

`Method Neos\Flow\Persistence\Generic\PersistenceManager::throwExceptionIfObjectIsNotAllowed does not exist`

I'll fix that…

kdambekalns:
See PR #2544

PR #2449, opened 2021-04-07
!!! FEATURE: Extensible configuration loaders

gjwnc:
This PR allows to implement custom ConfigurationLoader classes which replace the currently rather hardcoded ConfigurationProcessor's in an OOP way.  

It is then possible to load custom configurations in the ConfigurationManager with custom locations and custom configuration formats. E.g. it is possible to load json files or NodeTypes from another directory than `Configuration`.

This is not a breaking change for most cases, but the signature of `ConfigurationManager::registerConfigurationType()` has been changed and the usage of "configurationProcessingTypes" deprecated in favor of `Neos\Flow\Configuration\Loader\LoaderInterface` instances:

```php
// legacy (and still supported)
$configurationManager->registerConfigurationType('SomeCustom', ConfigurationManager::CONFIGURATION_PROCESSING_TYPE_APPEND);

// now
$configurationManager->registerConfigurationType('SomeCustom', new AppendLoader($yamlSource, 'SomeCustom'));
```

Related: https://github.com/neos/neos-development-collection/issues/3000
Related: https://github.com/neos/flow-development-collection/issues/1824


bwaidelich:
As discussed with @gjwnc , I still have two issues with the current implementation:

1. The special handling of custom temporary policy files in `PolicyConfigurationSource` (but I'd be totally fine with cleaning that up in a separate PR!)
2. That registering a new configuration type doesn't reset any previously loaded configuration for that type

The latter issue makes the behavior pretty weird because:

```php
$c1 = $configurationManager->getConfiguration('SomeType');
$configurationManager->registerConfigurationType('SomeType', $newSource);
$c2 = $configurationManager->getConfiguration('SomeType');
```
The second call won't use the new source since that call is cached.

This problem is not new. The `ConfigurationManager` only works reliable if the configuration is loaded only _after_ all configuration types were registered.

But since we reworked big parts anyways, it might make sense to go that one step further and get rid of all the weird state and leaked internal logic.

A greatly simplified interface could look something like (dummy code):

```php
class ConfigurationManager {


    public function getConfiguration(string $configurationType, string $configurationPath = null)
    {
        // as before
    }

    public function registerConfigurationType(string $configurationType, $configurationSource): void
    {
        // ...
        unset($this->configuration[$configurationType]);
        // TODO flush file cache?
    }

    // TODO remove public function loadConfigurationCache()

    private function loadConfiguration(): array
    {
        if (fileCache) {
            return fileCache;
        }
        // TODO: load from config source
        // TODO:  save cache (& replace placeholders)
        // TODO: load cache => $this->configuration
        // TODO: return $this->configuration
    }
}
```

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -540,17 +376,28 @@ public function flushConfigurationCache()
...
     }
 
+    /**
+     * Generate configuration with environment variables replaced without modifying or loading the cache
+     * @param string $configurationType
+     */
+    protected function processConfiguration(string $configurationType)
+    {
+        $config = null;
+        eval('$config = '.$this->replaceVariablesInPhpString(var_export($this->unprocessedConfiguration[$configurationType], true)).';');
```
albe, line 386: My security sensors tingle with this eval, but I can't put my finger on a valid abuse-case, since the var_export somewhat limits what is possible on the right side of the assignment - if I see correctly, s.th. like `''; do_something_evil();` shouldn't be possible

Neos.Flow/Classes/Configuration/ConfigurationSource/DefaultConfigurationSource.php
```diff
@@ -0,0 +1,53 @@
...
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Configuration\Source\YamlSource;
+use Neos\Flow\Core\ApplicationContext;
+use Neos\Utility\Arrays;
+
+class DefaultConfigurationSource implements ConfigurationSourceInterface
```
albe, line 18: Maybe `MergeConfigurationSource`? Since that is what it *does* opposed to how it's used (as default)?

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -249,16 +259,17 @@ public static function initializeSystemLogger(Bootstrap $bootstrap): void
...
-     * @param array $settings The Neos.Flow settings
      * @return ThrowableStorageInterface
      * @throws FlowException
      * @throws InvalidConfigurationTypeException
      */
-    protected static function initializeExceptionStorage(Bootstrap $bootstrap, array $settings): ThrowableStorageInterface
+    protected static function initializeExceptionStorage(Bootstrap $bootstrap): ThrowableStorageInterface
     {
+        $configurationManager = $bootstrap->getEarlyInstance(ConfigurationManager::class);
+        $settings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'Neos.Flow');
```
albe, line 269: Why is this duplicated here from https://github.com/neos/flow-development-collection/pull/2449/files#diff-1da1ee0f3e18bd7b4ffbe3603640de236ca4dfd088f6414364536af0813d7438R244?

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -249,16 +259,17 @@ public static function initializeSystemLogger(Bootstrap $bootstrap): void
...
+
         $storageClassName = $settings['log']['throwables']['storageClass'] ?? FileStorage::class;
         $storageOptions = $settings['log']['throwables']['optionsByImplementation'][$storageClassName] ?? [];
-        $renderRequestInformation = $settings['log']['throwables']['renderRequestInformation'] ?? true;
```
albe, line 261: Is this related? Looks as if #2359 was reverted accidentially in this PR?

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
```diff
@@ -19,6 +18,7 @@
 use Neos\Utility\Arrays;
 use PHPUnit\Framework\Assert;
 use Psr\Http\Message\ServerRequestFactoryInterface;
+use Symfony\Component\Yaml\Yaml;
```
albe, line 21: Where is this new import used?

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
```diff
@@ -95,11 +95,6 @@ public function iHaveTheFollowingPolicies($string)
         self::$testingPolicyPathAndFilename = $this->environment->getPathToTemporaryDirectory() . 'Policy.yaml';
         file_put_contents(self::$testingPolicyPathAndFilename, $string->getRaw());
 
-        $configurationManager = $this->objectManager->get(ConfigurationManager::class);
-        $configurations = ObjectAccess::getProperty($configurationManager, 'configurations', true);
-        unset($configurations[ConfigurationManager::CONFIGURATION_PROCESSING_TYPE_POLICY]);
-        ObjectAccess::setProperty($configurationManager, 'configurations', $configurations, true);
```
albe, line 101: I don't understand why this was done until now, but this is really not needed any more?

Neos.Flow/Classes/Configuration/ConfigurationSource/AppendConfigurationSource.php
```diff
@@ -0,0 +1,53 @@
...
+     * @var YamlSource
+     */
+    private $yamlSource;
+
+    /**
+     * @var string
+     */
+    private $filePrefix;
+
+    public function __construct(YamlSource $yamlSource, string $filePrefix)
```
albe, line 30: Should we really bind those constructors to `YamlSource`? What about the possibility to load other formats?

albe (commented):
Totally like the direction this is going with the configuration source implementations that contain the logic on how configurations are combined and loaded.
What I'm not fully sure about is the namespace `Configuration\Source\YamlSource` vs `Configuration\ConfigurationSource\ConfigurationSourceInterface`. At first sight, it's unclear what the difference between `Source` and `ConfigurationSource` is. Maybe the latter is rather a `ConfigurationLoader`? Or is there something else to distinguish these two concepts?

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
gjwnc: In short, this code does nothing and is basically a typo which exists very long in the current master. In the current master (and in this PR) the key of `$configurations` is the configurationType, not the configurationProcessingType. Thus the `unset(...)` does nothing. For example, see https://github.com/neos/flow-development-collection/blob/7b800d8217b30a5e6b96f0c648aec034f82bc5ff/Neos.Flow/Classes/Configuration/ConfigurationManager.php#L379.

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
gjwnc: Ah, I'll remove this. This was introduced as we tried to avoid generating a temporary test policy file as it is currently the case (https://github.com/neos/flow-development-collection/commit/9a0e9d8a7d278a2790775ec48e370c6bc3184c4d#diff-43787f11e7a6f604003f41baab1eada8d5ee8399ac202e1f07157f5a9846da52).

Neos.Flow/Classes/Configuration/ConfigurationSource/DefaultConfigurationSource.php
sorenmalling: Good idea with the naming of the class 👍

Neos.Flow/Classes/Configuration/ConfigurationSource/AppendConfigurationSource.php
sorenmalling: Since Yaml is the only configuration format we support in Flow, I say that it's okay

albe: Was thinking of the "vision" of this PR in the PR message that states `"and custom configuration format's. E.g. it is possible to load json files or NodeTypes"`

sorenmalling: Got ya'! +1 for the vision 👍 
But these core configuration sources, don't they always follow the core, so it will not change unless the core changes it's configuration format?

Do you think about a interface instead that could be implemented or did you have any other concrete idea in mind?

albe: yes, technically it's not something that is gonna break, because we have those constructors in the core. But we also set an example with those implementations, so it's likely that custom implementations will copy this constructor signature. And last but not least, some people will possibly extend those implementations (unless we `final` them, which we could indeed do).

gjwnc: Sorry for the late response. The implemented ConfigurationSources already implement a `ConfigurationSourceInterface`. You only need to implement this if you want to provide a complete custom source without yaml.

Neos.Flow/Classes/Configuration/ConfigurationManager.php
gjwnc: I've refactored the `eval` and now use a temporary file to include. The tests are now failing again - but because of some other issue with the master I guess.

Neos.Flow/Classes/Configuration/ConfigurationSource/DefaultConfigurationSource.php
gjwnc: Good point, I renamed the class

Neos.Flow/Classes/Core/Booting/Scripts.php
gjwnc: Yeah, you're right. Sorry, I didn't notice. I tried to fix the revert. It seems fine now.

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
gjwnc: The `use ..Yaml` is removed.

gjwnc:
@albe I've renamed the `ConfigurationSource` to `Loader`, thus it is named according to `Source\YamlSource`.

kitsunet:
Force push? 😱 Makes reasoning about the changes and evolving of this PR really hard.

Neos.Flow/Classes/Configuration/Loader/AppendLoader.php
```diff
@@ -0,0 +1,53 @@
...
+     * @var YamlSource
+     */
+    private $yamlSource;
+
+    /**
+     * @var string
+     */
+    private $filePrefix;
+
+    public function __construct(YamlSource $yamlSource, string $filePrefix)
```
kitsunet, line 30: I still think it would make a lot of sense to provide an interface for sources here, I think replacing the whole loader is not great because that would force you to also reimplement the merging logic in invoke just to change the source. BUT I also think it doesn'T necessarily have to be this change doing it. We can probably refactor that later in a separate change.

kitsunet (commented):
Some comments, mostly nitpicks, I think this is a great change. Would advocate for marking it breaking to notify people of the deprecations. Also open question to others, do we want to deprecate registerConfigurationType until next major with a note that the signature will change (but the method will stay)

gjwnc:
@kitsunet It seemed that I reverted #2359 and thus tried to fix the history. Later on, also psalm changes were missing because of renaming the namespace and I thought, that rebasing from the master would be a bit cleaner - especially since this draft PR has a bunch of messy commit messages I wanted a be on top so I don't get confused by other commits in between.  
So, in the future I should just merge from the upstream, right?

Neos.Flow/Classes/Package.php
```diff
@@ -130,7 +132,7 @@ public function boot(Core\Bootstrap $bootstrap)
         $dispatcher->connect(Tests\FunctionalTestCase::class, 'functionalTestTearDown', Mvc\Routing\RouterCachingService::class, 'flushCaches');
 
         $dispatcher->connect(Configuration\ConfigurationManager::class, 'configurationManagerReady', function (Configuration\ConfigurationManager $configurationManager) {
-            $configurationManager->registerConfigurationType('Views', Configuration\ConfigurationManager::CONFIGURATION_PROCESSING_TYPE_APPEND);
+            $configurationManager->registerConfigurationType('Views', new AppendConfigurationSource(new YamlSource(), 'Views'));
```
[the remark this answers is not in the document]
gjwnc: In this special case you mentioned, it is not necessary. The `registerConfigurationType` converts the given CONFIGURATION_PROCESSING_TYPE_ to the corresponding loader.

But the third parameter `$allowSplitSource` is removed because it is hardcoded in the implemented loader's - like it is currently done by [`$configurationTypes`](https://github.com/neos/flow-development-collection/blob/5735c345d356bc8664e0ef43d91570d44e1a3c14/Neos.Flow/Classes/Configuration/ConfigurationManager.php#L119).  
So, yes there is a breaking change.

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -232,70 +197,44 @@ public function setPackages(array $packages)
...
-            self::CONFIGURATION_PROCESSING_TYPE_SETTINGS,
-            self::CONFIGURATION_PROCESSING_TYPE_APPEND
-        ];
-        if (!in_array($configurationProcessingType, $configurationProcessingTypes)) {
-            throw new \InvalidArgumentException(sprintf('Specified invalid configuration processing type "%s" while registering custom configuration type "%s"', $configurationProcessingType, $configurationType), 1365496111);
+        if ($configurationProcessingType === self::CONFIGURATION_PROCESSING_TYPE_DEFAULT) {
+            return new MergeLoader(new YamlSource(), $configurationType);
+        }
+        if ($configurationProcessingType === self::CONFIGURATION_PROCESSING_TYPE_APPEND) {
+            return new AppendLoader(new YamlSource(), $configurationType);
```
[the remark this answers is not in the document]
gjwnc: Good point, I've added all processing types here

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -317,31 +256,36 @@ protected function emitConfigurationManagerReady(ConfigurationManager $configura
      * If possible just use settings and have them injected.
      *
      * @param string $configurationType The kind of configuration to fetch - must be one of the CONFIGURATION_TYPE_* constants
-     * @param string $configurationPath The path inside the configuration to fetch
-     * @return array|null The configuration or NULL if the configuration doesn't exist
+     * @param string|null $configurationPath The path inside the configuration to fetch
+     * @return mixed|null The configuration or NULL if the configuration doesn't exist
```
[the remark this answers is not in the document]
gjwnc: I've removed the null from `mixed`, thanks.

What you mean by breaking here? I think `getConfiguration()` will return the same as before, that is `null` is returned if the given path is not found or an empty array if the given registered configurationType has no configuration. In the last case, the spaceship operator changes the null to an empty array

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -540,17 +384,32 @@ public function flushConfigurationCache()
...
 
+    /**
+     * Generate configuration with environment variables replaced without modifying or loading the cache
+     *
+     * @param string $configurationType The kind of configuration to fetch
+     */
+    protected function processConfigurationType(string $configurationType)
+    {
+        if ($this->temporaryDirectoryPath !== null) {
+            $cachePathAndFilename = $this->constructConfigurationCachePath().'.tmp';
```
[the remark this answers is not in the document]
gjwnc: Hmm, what about using something like
```
$cachePathAndFilename = $this->constructConfigurationCachePath() . '.' . uniqid($configurationType, true) . '.tmp';
```
But this also means, the `$configurationType` needs to be sanitized to allowed filesystem characters. Maybe `uniqid(crc32($configurationType))` could be enough?

Neos.Flow/Classes/Configuration/Loader/AppendLoader.php
albe: Agree

albe:
> do we want to deprecate registerConfigurationType until next major with a note that the signature will change (but the method will stay)

Not sure how we handled such cases before - but I don't see much value in this at all, since the user can not adjust the code early anyway. We'd need to introduce another method for that and then really switch that, but is that worth the additional work?

Neos.Flow/Classes/Configuration/ConfigurationSource/AppendConfigurationSource.php
albe: > But what now is a ConfigurationSource is really more of a ConfigurationProcessor, no? MergeConfigurationProecessor, AppendConfigurationProcessor etc. etc.

Yeah, you could call it a processor, but I also do think that Loader works well now. It does load config data from a source in a specific order and merge strategy.

> but I also feel that could be offloaded into an (what I would call) ConfigurationSourceInterface, that just wraps the loading and returns the loaded array so if you want to have something else than YAML you just need to implement that

By "wrap the loading" do you mean wrap `YamlSource`? Not sure this additional layer of abstraction would carry much benefit - right now, that's basically what `YamlSource` is doing with it's `load()` method. So if we just make this an interface, the same flexibility would be provided (and we can do this in a separate PR as you mentioned)

bwaidelich:
Thanks for your work on this. I think it would make a great feature for the upcoming 7.2 release (@daniellienert has just put it on the release board).

+1 for marking this breaking to make sure. I added the "!!!" prefix and extended the PR description accordingly.

btw: Daniel also made me aware of the fact that we miss the `declare(strict_types=1);` declaration in all files

gjwnc:
@bwaidelich I've added `declare(strict_types=1)` to the ConfigurationManager and the new Loader files

kitsunet:
> > do we want to deprecate registerConfigurationType until next major with a note that the signature will change (but the method will stay)
> 
> Not sure how we handled such cases before - but I don't see much value in this at all, since the user can not adjust the code early anyway. We'd need to introduce another method for that and then really switch that, but is that worth the additional work?

Na I don't think so. I just though as an additional method of drawing attention. But it's probably not a good idea. Lets just declare it breaking with mention of signature change in next major and hope people notice.

kitsunet:
> @kitsunet It seemed that I reverted #2359 and thus tried to fix the history. Later on, also psalm changes were missing because of renaming the namespace and I thought, that rebasing from the master would be a bit cleaner - especially since this draft PR has a bunch of messy commit messages I wanted a be on top so I don't get confused by other commits in between.
> So, in the future I should just merge from the upstream, right?

Yes please :) It's not a big deal but while reviewing it's nice to be able to follow up on additions to the PR from the point you reviewed to, especially with bigger changes like this.

Neos.Flow/Classes/Configuration/ConfigurationManager.php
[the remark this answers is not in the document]
kitsunet: sounds fine to me :)

Neos.Flow/Classes/Configuration/ConfigurationManager.php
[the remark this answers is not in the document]
kitsunet: You are right of course, it could never return null before. The change tricked me but it's really just a correction that makes sense. I thought it would return $configuration as NULL before but that couldn've never happeend because spaceship.

Neos.Flow/Classes/Configuration/ConfigurationManager.php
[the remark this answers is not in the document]
gjwnc: I've fixed concurrency using `uniqid` and `crc32`.

The test, that is now failing is because of the recent changes in the master I merged - I think.

bwaidelich:
I just turned this from a draft to a proper PR – or is there anything that still blocks this?

bwaidelich:
I'm biased because I worked on this one too, but I think that we should merge this one ASAP so that we can finetune/fix things before the release. Also this still needs the corresponding Neos change (https://github.com/neos/neos-development-collection/pull/3332) that makes use of the new config source to allow the NodeType definitions to be stored in subfolders

kitsunet (approved):

kitsunet:
@bwaidelich agreed, I gave it my thumbs up.

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -346,7 +356,7 @@ public static function initializeCacheManagement(Bootstrap $bootstrap)
         $environment = $bootstrap->getEarlyInstance(Environment::class);
 
         $cacheFactoryObjectConfiguration = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_OBJECTS, CacheFactoryInterface::class);
-        $cacheFactoryClass = isset($cacheFactoryObjectConfiguration['className']) ? $cacheFactoryObjectConfiguration['className'] : CacheFactory::class;
+        $cacheFactoryClass = $cacheFactoryObjectConfiguration && isset($cacheFactoryObjectConfiguration['className']) ? $cacheFactoryObjectConfiguration['className'] : CacheFactory::class;
```
albe, line 359: That additional condition is not necessary in PHP with `isset()` ;) But no blocker

Neos.Flow/Classes/Configuration/Loader/PolicyLoader.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        $this->yamlSource = $yamlSource;
+    }
+
+    /**
+     * Set an absolute file path to store configuration caches in. If null no cache will be active.
+     *
+     * @param string $temporaryDirectoryPath
+     */
+    public function setTemporaryDirectoryPath(string $temporaryDirectoryPath): void
```
albe, line 46: Since the loader will likely fail without having set this path, we could maybe make it a constructor dependency. Would also make instanciation a bit easier at https://github.com/neos/flow-development-collection/pull/2449/files#diff-83398c2914c2ce46b6e55c9821468814652a275d332692492bb6edbbb84607f9R245-R247

albe (approved):
Overall approve by reading, but haven't tested it.

Neos.Flow/Classes/Core/Booting/Scripts.php
gjwnc: Nice, didn't know that. I've removed the condition.

Neos.Flow/Classes/Configuration/Loader/PolicyLoader.php
gjwnc: The only reason, the temporary directory path is needed in the policy loader is for behavior testing. The path is not used in development or production. This is kind of nasty workaround - which also exists in the current ConfigurationManager.

The temporary Policy.yaml is written [here](https://github.com/neos/flow-development-collection/blob/c00e0daf1d1df64fa073d67b25753a504f1b9a95/Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php#L95). A few weeks ago I and @bwaidelich tried to fix this with something like a closure policy loader, but the ConfigurationManager just ignored them. This was related to how thows behaviour tests are creating an instance of the ConfigurationManager. Tbh, it was very complicated to understand why the test didn't work out and finally I couldn't fix this in a better way, than it is currently implemented.

So, this kinda nasty testing fix with a temporary policy file is still there and I would argue to keep the constructor for now, because I think someone with more insight in those behavior test might be able to fix this workaround - maybe with a custom yamlsource or such? And if the test doesn't need a temporary Policy.yaml we don't need to change the constructor.

daniellienert (commented):
Great feature, thanks to all involved!

We should have some description of that feature in https://github.com/neos/flow/blob/master/Documentation/TheDefinitiveGuide/PartIII/Configuration.rst

Sebobo:
Adjusted https://github.com/neos/neos-development-collection/pull/3332 and used it to test this feature with nodetypes in hierarchical folders and works perfectly fine 👍 
Can't say much about the inner parts of the change.

Sebobo (approved):

bwaidelich:
> We should have some description of that feature

I agree. But I would suggest that we don't block this PR until that's done since the docs are still correct (AFAIK) and this is just an *additional* feature.

One test still fails for some reason!? is this one good to go?

daniellienert:
The test uses lowest possible composer dependencies, which installs guzzle 1.6.1 which does not have the Utils library. We need to raise the minimal version to 1.7

daniellienert:
Should be fixed with: https://github.com/neos/flow-development-collection/pull/2525

daniellienert (approved):

Neos.Flow/Classes/Configuration/Loader/LoaderInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Core\ApplicationContext;
+
+/**
+ * The interface for a configuration source
```
kdambekalns, line 19:
```suggestion
 * The interface for a configuration loader
```

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -370,145 +315,44 @@ public function warmup()
...
+        if ($cachePathAndFilename === null) {
+            $cachePathAndFilename = $this->constructConfigurationCachePath();
+        }
         $configurations = @include $cachePathAndFilename;
         if ($configurations !== false) {
-            $this->configurations = $configurations;
+            if ($configurationType === null) {
+                $this->configurations = $configurations;
+            } else {
+                $this->configurations[$configurationType] = $configurations;
```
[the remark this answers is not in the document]
kdambekalns: So what is the code path and use case? Documenting it here (as a code comment) would be a first step to make it clearer… 😉

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -232,70 +197,44 @@ public function setPackages(array $packages)
...
-     * @param string $configurationType
-     * @return string
-     * @throws Exception\InvalidConfigurationTypeException on non-existing configurationType
+     * @param string $configurationType The type to register, may be anything
+     * @param LoaderInterface|callable|string $configurationLoader
+     * @return void
+     *@throws \InvalidArgumentException on invalid configuration processing type
      */
-    public function resolveConfigurationProcessingType(string $configurationType): string
+    public function registerConfigurationType(string $configurationType, $configurationLoader): void
```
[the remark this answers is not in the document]
kdambekalns: This means we need to target the next major, strictly speaking. I see two options:

- **Provide a code migration** that helps migrate or at least spot code that needs to be adjusted.
- **Make the `$configurationLoader` optional (as long as it may be a string for B/C reasons) and fill it depending on the type in `convertLegacyProcessingType()`

kdambekalns (changes requested):
I found some issues (see inline comments), sorry for being late with my review.

Another thing I asked myself, wondering why noone else asked… Why opt for an interface that requires `__invoke()`, instead of a method whose name clearly communicates intention? What is wrong about e.g. `load()`? Is it the "duplication" in `$loader->load()` you wanted to avoid? If you want to allow the use of anything callable–then why add an interface at all?

Neos.Flow/Classes/Core/ApplicationContext.php
```diff
@@ -131,4 +142,15 @@ public function getParent()
     {
         return $this->parentContext;
     }
+
+    /**
+     * Returns the names of this context and all parents in descending specificity
+     *
+     * @return string[] for example ['Production', 'Staging', 'Server1']
```
bwaidelich, line 149:
```suggestion
     * @return string[] for example ['Production', 'Production/Staging', 'Production/Staging/Server1']
```

Neos.Flow/Classes/Core/ApplicationContext.php
```diff
@@ -131,4 +142,15 @@ public function getParent()
     {
         return $this->parentContext;
     }
+
+    /**
+     * Returns the names of this context and all parents in descending specificity
```
bwaidelich, line 147:
```suggestion
     * Returns the names of this context and all parents in ascending specificity
```

Neos.Flow/Classes/Core/ApplicationContext.php
```diff
@@ -64,6 +69,13 @@ public function __construct(string $contextString)
     {
         $contextStringParts = explode('/', $contextString);
         $this->rootContextString = reset($contextStringParts);
+
+        $combinedContextStrings = '';
+        foreach ($contextStringParts as $contextStringPart) {
+            $this->hierarchy[] = $combinedContextStrings . $contextStringPart;
```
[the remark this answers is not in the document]
bwaidelich: Good point. It's just the doc comment that's wrong IMO, see suggestions below

Neos.Flow/Tests/Unit/Core/ApplicationContextTest.php
```diff
@@ -153,4 +153,25 @@ public function parentContextIsConnectedRecursively()
         $rootContext = $parentContext->getParent();
         self::assertSame('Production', (string) $rootContext);
     }
+
+    public function getHierarchyDataProvider(): array
+    {
+        return [
+            ['contextString' => 'Development', 'expectedResult' => ['Development']],
+            ['contextString' => 'Testing/Staging', 'expectedResult' => ['Testing', 'Testing/Staging']],
+            ['contextString' => 'Production/Staging/Stage1', 'expectedResult' => ['Production', 'Production/Staging', 'Production/Staging/Stage1']],
```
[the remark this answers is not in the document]
bwaidelich: Implementation and Tests are correct, the doc comment was wrong (see suggestions above)

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -232,70 +197,44 @@ public function setPackages(array $packages)
...
-     * $configurationType.
+     * Registers a new configuration type with the given source.
      *
-     * @param string $configurationType
-     * @return string
-     * @throws Exception\InvalidConfigurationTypeException on non-existing configurationType
+     * @param string $configurationType The type to register, may be anything
+     * @param LoaderInterface|callable|string $configurationLoader
+     * @return void
+     *@throws \InvalidArgumentException on invalid configuration processing type
```
bwaidelich, line 215:
```suggestion
     * @throws \InvalidArgumentException on invalid configuration processing type
```

Neos.Flow/Classes/Configuration/ConfigurationManager.php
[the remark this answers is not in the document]
bwaidelich: The signature has changed, but it should only break if someone actually *subclassed* the `ConfigurationManager` (that should have been `final` IMO).
The *implementation* is mostly backwards compatible, i.e.
```php
$configurationManager->registerConfigurationType('SomeType');
$configurationManager->registerConfigurationType('SomeType', ConfigurationManager::CONFIGURATION_PROCESSING_TYPE_OBJECTS);
$configurationManager->registerConfigurationType('SomeType', ConfigurationManager::CONFIGURATION_PROCESSING_TYPE_OBJECTS, true);
```
should all still work (thanks to line 215) with the exception that the last param (`allowSplitSource`) is ignored since that now just depends on the loader.

I think that both (overriding ConfigurationManager and relying on "allowSplitSource" to be false(!)) could be considered edge cases and don't require a migration nor major version jump as long as the "breaking change" is properly described (which it _is_ in this PR IMO)

bwaidelich:
@kdambekalns Thanks for your review and feedback!

> Why opt for an interface that requires __invoke(), instead of a method whose name clearly communicates intention?

I agree, that's somewhat odd. Here is my reasoning:
The interface is merely a convenience for devs implementing a custom loader. It is never actually checked or enforced (partly due to backwards compatibility).
Instead a loader must be a `callable` (e.g. you could replace it by a closure if you wanted to).

Using "invokable" classes is nothing that we really use a lot in Flow/Neos. I've seen it quite a lot in Symfony code though and kinda liked it.
Having said that, I'd be totally fine with a more explicit interface and replacing
```php
    public function registerConfigurationType(string $configurationType, $configurationSource): void
    {
        // ...
        if (!is_callable($configurationSource)) {
```
by sth like
```php
    public function registerConfigurationType(string $configurationType, $configurationSource): void
    {
        // ...
        if (!$configurationSource instanceof ConfigurationSourceInterface) {
```

(we still can't use the interface as type hint for B/C reasons)

Neos.Flow/Classes/Configuration/ConfigurationManager.php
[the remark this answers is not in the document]
kdambekalns: But `$configurationManager->registerConfigurationType('SomeType');` will lack the **required** second parameter!

kdambekalns:
> Instead a loader must be a callable (e.g. you could replace it by a closure if you wanted to).

Good point about the callable, that could make life easier. I admit the callable classes seem convenient, it's just the fact that having a "useless contract" seems weird. But it's the closest-to-the-code documentation, probably.

Neos.Flow/Classes/Configuration/ConfigurationManager.php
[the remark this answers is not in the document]
bwaidelich: *dang* of course.. That's why we need reviews :)
What about
```suggestion
    public function registerConfigurationType(string $configurationType, $configurationLoader = null): void
```
and sth along the lines of
```php
if ($configurationLoader === null) {
    $configurationLoader = new SettingsLoader(new YamlSource());
// B/C layer
} elseif (is_string($configurationLoader)) {
    $configurationLoader = $this->convertLegacyProcessingType($configurationType, $configurationLoader);
}
```

in the implementation?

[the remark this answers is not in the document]
bwaidelich:
```php
$configurationLoader = new MergeLoader(new YamlSource(), $configurationType);
```
not
```php
$configurationLoader = new SettingsLoader(new YamlSource());
```
of course

Neos.Flow/Classes/Configuration/ConfigurationManager.php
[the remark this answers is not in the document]
bwaidelich: True, that code path is pretty weird and should be split into two methods IMO. I'll take care

bwaidelich:
I added some more commits tackling
* making the `LoaderInterface` an actual contract (be5e60a6cdc768303e734bcd54f88bcf719cc252)
* make `$configurationLoader` param optional for better b/c (0547dd25024312000978d56061fbd7d4434f99d7)
* split up `ConfigurationManager::loadConfigurationCache()` into two methods for better readability (thanks to @gjwnc for helping with that) (2de0df1989f54d4f820b0ddeeb0461591265b8f2)

And the other minor tweaks mentioned in the comments.
The failing 7.3 test is not related as far as I can tell

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -232,70 +203,58 @@ public function setPackages(array $packages)
...
-    public function isSplitSourceAllowedForConfigurationType(string $configurationType): bool
-    {
-        if (!isset($this->configurationTypes[$configurationType])) {
-            throw new Exception\InvalidConfigurationTypeException('Configuration type "' . $configurationType . '" is not registered. You can Register it by calling $configurationManager->registerConfigurationType($configurationType).', 1359998400);
+        // B/C layer
+        } elseif (is_string($configurationLoader)) {
+            $configurationLoader = $this->convertLegacyProcessingType($configurationType, $configurationLoader);
+        }
+        if (!$configurationLoader instanceof LoaderInterface) {
+            throw new \InvalidArgumentException(sprintf('Specified invalid configuration load of type "%s" while registering custom configuration type "%s". This should be an instance of %s', is_object($configurationLoader) ? get_class($configurationLoader) : gettype($configurationLoader), $configurationType, LoaderInterface::class), 1617895964);
```
kdambekalns, line 227:
```suggestion
            throw new \InvalidArgumentException(sprintf('Specified invalid configuration loader of type "%s" while registering custom configuration type "%s". This should be an instance of %s', is_object($configurationLoader) ? get_class($configurationLoader) : gettype($configurationLoader), $configurationType, LoaderInterface::class), 1617895964);
```

kdambekalns (approved):
One last tiny exception message tweak, everything else seems fine by now. Thanks a lot!

kdambekalns:
PLEASE add documentation for this sometime soon… ❤️

PR #2451, opened 2021-04-07
TASK: Add test case for singleton EntityManager

albe:
This is a test that verifies the wrong-behaviour of #2423 and that #2448 fixes it.

Don't merge, it will be included in #2448 - just here to verify that the test fails without the fix.

### DIFF
--- a/Neos.Flow/Tests/Functional/Persistence/PersistenceTest.php
+++ b/Neos.Flow/Tests/Functional/Persistence/PersistenceTest.php
@@ -40,16 +40,31 @@ class PersistenceTest extends FunctionalTestCase
     protected $extendedTypesEntityRepository;
 
     /**
+     * @var EntityManagerInterface
+     */
+    protected $earlyEntityManager;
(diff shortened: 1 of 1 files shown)


albe:
image
👀 ❓

albe:
Okay, I'm dead serious - I retried running the test on my WSL ubuntu, with a mariadb from CLI to reflect the CI environment as closely as possible and the test there still fails. @kdambekalns maybe you have any idea? Does this test fail for you locally too?

kdambekalns:
Since this passes here and in #2448 - is that still an open question, @albe? And why is tis against 5.3 while #2448 is against 6.3?

albe:
The latter was rebased after 5.3 went security-only (https://github.com/neos/flow-development-collection/pull/2448#event-4860890542). Still totally clueless why this test failed for me locally, but passes here. Anyway, since it's included in 2448 I'll close this now with a few question marks left over in my head

PR #2452, opened 2021-04-08
BUGFIX: Use lazy EntityManager injection to prevent recursive instanciation

albe:
With the changes in #2423 the PersistenceManager was registered as a Doctrine EventListener.
Hence, when building the EntityManager, the PersistenceManager got instanciated with a completely new non-lazy EntityManager. This caused errors when trying to persist entities, as they were not known to that instance of the EntityManager.
This change makes the EntityManager instanciation lazy again and handles the type mismatch of the lazy DependencyProxy in the SchemaTool.

This is an alternative fix to #2448


albe:
See https://github.com/neos/flow-development-collection/pull/2448#issuecomment-815558971

kdambekalns (approved):

daniellienert (approved):

albe:
Given the positive reviews and the confirmation of a couple of people that the lazy injection solves the issue, I'll merge this as is and release it right away. Let's still consider getting #2448 in as well, as I do think the refactoring of the onFlush EventListener out of the PersistenceManager makes a lot of sense.

PR #2454, opened 2021-04-10
TASK: Add minimal dependencies build

albe:
This should make sure that our minimum dependency requirements actually lead to a working installation.
If this build fails, we need to raise some dependencies minimum version.


Neos.Cache/composer.json
```diff
@@ -14,7 +14,7 @@
     },
     "require-dev": {
         "mikey179/vfsstream": "^1.6.1",
-        "phpunit/phpunit": "~8.1"
+        "phpunit/phpunit": "~8.5"
```
albe, line 17: Required because of `expectExceptionMessageMatches()` in PersistenceManagerTest - that method was introduced in 8.4 actually, but 8.4 fails with `Cannot instantiate abstract class Neos\Eel\Tests\Unit\AbstractEvaluatorTest` due to the dataProviders there.

Neos.Flow/composer.json
```diff
@@ -41,6 +41,7 @@
         "doctrine/migrations": "^1.8",
         "doctrine/dbal": "^2.9",
         "doctrine/common": "^2.4",
+        "doctrine/persistence": "^1.3",
```
albe, line 44: Required due to `Doctrine\Persistence\Mapping\Driver\MappingDriver` in 40f35e0a3d0b12557621bf6498fb0b539c4d8494

Neos.Http.Factories/composer.json
```diff
@@ -9,7 +9,7 @@
     "require": {
         "php": "^7.1",
         "psr/http-factory": "^1.0",
-        "guzzlehttp/psr7": "^1.4, !=1.8.0 || ~2.0"
+        "guzzlehttp/psr7": "^1.4.1, !=1.8.0 || ~2.0"
```
albe, line 12: Required, because 1.4.0 throws an exception when `withPath()` is called without a leading slash - we need to fix this in Flow - see #2473

Neos.Flow/composer.json
```diff
@@ -41,6 +41,7 @@
         "doctrine/migrations": "^1.8",
```
albe, line 41:
```suggestion
        "doctrine/migrations": "^1.8.1",
```

composer.json
```diff
@@ -25,13 +25,14 @@
         "doctrine/migrations": "^1.8",
```
albe, line 25:
```suggestion
        "doctrine/migrations": "^1.8.1",
```

albe:
🎉 💚

PR #2455, opened 2021-04-10
TASK: Replace deprecated Guzzle functions

### LINKED ISSUE #2383: ActionResponse uses deprecated Guzzle stream_for method
The `setContent` method in the `ActionResponse` [class](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Mvc/ActionResponse.php#L77) uses, a deprecated (from guzzle/psr7) `stream_for` method

image

Guzzle already anotated the method with a alternative, that should be replaced

` * @deprecated stream_for will be removed in guzzlehttp/psr7:2.0. Use Utils::streamFor instead.`

albe:
Resolves #2383

### DIFF
--- a/Neos.Flow/Classes/Http/Client/CurlEngine.php
+++ b/Neos.Flow/Classes/Http/Client/CurlEngine.php
@@ -11,11 +11,11 @@ namespace Neos\Flow\Http\Client;
  * source code.
  */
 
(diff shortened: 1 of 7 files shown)


sorenmalling (approved):

albe:
Ignore the remaining psalm failure. It's an artefact of doctrine extending their psalm hints.

PR #2458, opened 2021-04-12
BUGFIX: Null coalesce content type in ActionResponse getter

albe:
The current PHP typehint of `: string` will cause this method to throw an error when `setContentType()` was not called before with a valid string. In Flow 7 we lifted the typehint to `?string`, but IMO that does only complicate the API unnecessarily, because `''` is not a valid content type any way and hence indistinguishable from "did not set content type" for any useful means and purposes.
Hence I suggest using null coalescing instead (and changing the 7+ typehint back to `string`, though that would be breaking).

See also discussion in https://github.com/neos/flow-development-collection/pull/2180#discussion_r550197400

### DIFF
--- a/Neos.Flow/Classes/Mvc/ActionResponse.php
+++ b/Neos.Flow/Classes/Mvc/ActionResponse.php
@@ -196,7 +196,7 @@ final class ActionResponse
      */
     public function getContentType(): string
     {
-        return $this->contentType;
+        return $this->contentType ?? '';
     }
 
     /**


Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -196,7 +196,7 @@ public function getStatusCode(): int
      */
     public function getContentType(): string
     {
-        return $this->contentType;
+        return $this->contentType ?? '';
```
bwaidelich, line 199: Just an alternative solution that makes the behavior more deterministic IMO.

```suggestion
        if ($this->contentType === null) {
            throw new \RuntimeException('No content type is set, use hasContentType() to determine that case', 1618240401);
        }
        return $this->contentType;
```

bwaidelich (commented):
Thanks for extracting this to a separate PR.
By silently converting the never set content type to an empty string, we assume that the consuming side caters for that case.
Throwing an exception would be a more explicit alternative IMO

Neos.Flow/Classes/Mvc/ActionResponse.php
albe: Not sure if throwing in that case is a good measure. It's pretty drastic, given the user just wanted to read the content-type set on the ActionResponse, no?
I see it like this: If you read the ContentType, which is a string, you should anyway handle an empty string (or we use `non-empty-string` psalm type and check for that in the setter). So at that point, what is the difference between having set an empty string and nothing having been set on the `ActionResponse`? We could also instead initialize `$contentType = '';`.

I would probably still prefer the exception to a nullable return type though

bwaidelich: >  you should anyway handle an empty string

if you _know_ that you need to handle an empty string, you can do the `hasContentType()` check as well.
The difference is: an exception _forces_ you to handle it in case you were doing something nasty like
```php
$filename = $someSlug . '.' . $this->response->getContentType();
```

> (or we [...] and check for that in the setter)

Yes, why not? :)


I'm fine without the exception obviously and I agree that this is an edge case of an "unexpected" behavior

albe:
ping @bwaidelich I think we should decide on this the next days before 6.0-6.2 run into security-only mode. I consider this method seriously broken right now and would like to find a good solution from the DX PoV. Maybe we can have a quick call monday or tuesday? 15-30 min max.

bwaidelich (approved):
@albe I'm happy to do a quick call with you today or tomorrow, but I'm also happy with the fix as is. I just wanted to point out some alternative solution but it has its drawbacks too

albe:
I'd love to talk about this quickly regarding throwing exception or not, but more about how to deal with 7 and it's nullable return type.

bwaidelich:
> I'd love to talk about this quickly regarding throwing exception or not

OK, sure, anytime!

albe:
As discussed, we'll keep it like this without throwing an exception in the getter. While that would be super safe to not forget handling the unset case (and it should maybe then even cover the empty case), it is pretty drastic if all you do is e.g. `if ($this->getContentType() !== '') { ... }` which would be totally sane. We'll document to use the `hasContentType()` method instead again, but won't dictate how not to write safe code.

The nullable return type will stay in 7.0.x, but I'll prepare another bugfix for 7.1 to revert that back to non-nullable. This is only breaking if you'd have code in place that did only cover the unset case by checking for `null` and ignores empty string.

PR #2459, opened 2021-04-13
!!! FEATURE: Enable URL Rewriting by default

bwaidelich:
This flips the default behavior for the `FLOW_REWRITEURLS` environment
variable: If it is _not_ specified (or contains a value other than `0`)
URL rewriting will be enabled.

Previously URL rewriting was enabled with a corresponding `SetEnv` configuration
for Apache.
For other servers and CLI a corresponding configuration (or a
`putenv('FLOW_REWRITEURLS=1')` call at runtime) was required in order to
activate URL rewriting.

This could be a breaking change in case you relied on the previous behavior.
For example: Using the `UriBuilder` in CLI previously created URLs in the format `/index.php/some/path`. Now it will lead to `/some/path` by default.
To re-establish the former behavior, the `FLOW_REWRITEURLS` can be set to 0 explicitly, for example via `putenv('FLOW_REWRITEURLS=1');`.

### DIFF
--- a/Neos.Flow/Classes/Utility/Environment.php
+++ b/Neos.Flow/Classes/Utility/Environment.php
@@ -109,7 +109,7 @@ class Environment
      */
     public function isRewriteEnabled(): bool
     {
-        return (bool)Bootstrap::getEnvironmentConfigurationSetting('FLOW_REWRITEURLS');
+        return (string)Bootstrap::getEnvironmentConfigurationSetting('FLOW_REWRITEURLS') !== '0';
     }
 
     /**
--- a/Neos.Flow/Resources/Private/Installer/Distribution/Defaults/Web/.htaccess
+++ b/Neos.Flow/Resources/Private/Installer/Distribution/Defaults/Web/.htaccess
@@ -17,8 +17,8 @@
 	# Enable URL rewriting
 	RewriteEngine On
 
-	# Set flag so we know URL rewriting is available
-	SetEnv FLOW_REWRITEURLS 1
+	# Set flag to "0" to disable URL rewriting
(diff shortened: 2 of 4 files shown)


bwaidelich:
@albe good point. I marked this a "breaking change" now and added some description. In practice I don't think that this is really breaking for anything in production though.

bwaidelich:
btw: the failing test is not related, is it?

albe (approved):

albe:
> btw: the failing test is not related, is it?

Nope. Doctrine just improved their psalm type hints

Neos.Flow/Classes/Utility/Environment.php
```diff
@@ -109,7 +109,7 @@ public function getMaximumPathLength(): int
      */
     public function isRewriteEnabled(): bool
     {
-        return (bool)Bootstrap::getEnvironmentConfigurationSetting('FLOW_REWRITEURLS');
+        return (string)Bootstrap::getEnvironmentConfigurationSetting('FLOW_REWRITEURLS') !== '0';
```
daniellienert, line 112: That should not work. Why casting the result to string when the function has a bool return type declaration?

daniellienert (changes requested):

Neos.Flow/Classes/Utility/Environment.php
bwaidelich: read it again :)

daniellienert: gnah, the binding ... 🙄

daniellienert (approved):

albe:
I guess let's do it then :)

PR #2461, opened 2021-04-16
FEATURE: Allow flushing caches by compiled classes

albe:
This adds a signal to the proxy compiler when classes have been compiled and a slot to the a new `AnnotationsCacheFlusher` class that checks the compiled classes for existing annotations and flushes matching configured caches.
This is required when annotations are used to change configuration, like the `@Flow\Route` or `@Flow\Privilege` annotations.

Related to https://github.com/neos/flow-development-collection/pull/2421
Related to https://github.com/neos/flow-development-collection/pull/2412

### DIFF
--- a/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
+++ b/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
@@ -1,0 +1,75 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Flow\Cache;
+
+use Neos\Flow\Configuration\ConfigurationManager;
+use Neos\Flow\Reflection\ReflectionService;
+use Neos\Flow\Log\Utility\LogEnvironment;
+use Psr\Log\LoggerInterface;
+
+final class AnnotationsCacheFlusher
+{
+    /**
+     * @Flow\Inject
+     * @var LoggerInterface
+     */
+    protected $logger;
+
+    /**
+     * @Flow\Inject
+     * @var ConfigurationManager
+     */
+    protected $configurationManager;
+
+    /**
+     * @Flow\Inject
+     * @var CacheManager
+     */
+    protected $cacheManager;
+
+    /**
+     * The ReflectionService is only needed during compile time for handling flushConfigurationCachesByCompiledClass
+     *
+     * @Flow\Inject
+     * @var ReflectionService
+     */
+    protected $reflectionService;
+
+    /**
+     * A slot that flushes caches as needed if classes with specific annotations have changed
+     *
+     * @param array<string> $classNames The full class names of the classes that got compiled
+     * @return void
+     */
+    public function flushConfigurationCachesByCompiledClass(array $classNames): void
+    {
+        $caches = [
+            Neos\Flow\Annotations\Route::class => ['Flow_Mvc_Routing_Route', 'Flow_Mvc_Routing_Resolve']
+        ];
+        $cachesToFlush = [];
+
+        foreach ($classNames as $className) {
+            foreach ($caches as $annotationClass => $cacheNames) {
+                if (!$this->reflectionService->isClassAnnotatedWith($className, $annotationClass)
+                    && count($this->reflectionService->getMethodsAnnotatedWith($className, $annotationClass)) === 0) {
+                    continue;
+                }
+                foreach ($caches[$annotationClass] as $cacheName) {
+                    $cachesToFlush[$cacheName] = $annotationClass;
+                }
+            }
+        }
+
+        foreach ($cachesToFlush as $cacheName => $annotationClass) {
+            $this->logger->info(sprintf('A class file containing the annotation "%s" has been changed, flushing related cache "%s"', $annotationClass, $cacheName), LogEnvironment::fromMethodName(__METHOD__));
+            $this->cacheManager->getCache($cacheName)->flush();
+        }
+
+        if (count($cachesToFlush) > 0) {
+            $this->logger->info('An annotated class file has been changed, refreshing compiled configuration cache', LogEnvironment::fromMethodName(__METHOD__));
+            $this->configurationManager->refreshConfiguration();
+        }
+    }
+
+}
--- a/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
+++ b/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
@@ -78,6 +78,9 @@ final class AnnotationsCacheFlusher
      */
     public function flushConfigurationCachesByCompiledClass(array $classNames): void
     {
+        if ($this->annotationToCachesMap === []) {
+            return;
+        }
         $cachesToFlush = [];
         foreach ($classNames as $className) {
             foreach ($this->annotationToCachesMap as $annotationClass => $cacheNames) {
--- a/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
+++ b/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
@@ -1,8 +1,19 @@
 <?php
 declare(strict_types=1);
 
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
 namespace Neos\Flow\Cache;
 
+use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Configuration\ConfigurationManager;
 use Neos\Flow\Reflection\ReflectionService;
 use Neos\Flow\Log\Utility\LogEnvironment;
@@ -71,5 +82,4 @@ final class AnnotationsCacheFlusher
             $this->configurationManager->refreshConfiguration();
         }
     }
-
 }
--- a/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
+++ b/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
@@ -13,12 +13,16 @@ declare(strict_types=1);
 
 namespace Neos\Flow\Cache;
 
+use Neos\Cache\Exception\NoSuchCacheException;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Configuration\ConfigurationManager;
 use Neos\Flow\Reflection\ReflectionService;
 use Neos\Flow\Log\Utility\LogEnvironment;
 use Psr\Log\LoggerInterface;
 
+/**
+ * @Flow\Scope("singleton")
+ */
 final class AnnotationsCacheFlusher
 {
     /**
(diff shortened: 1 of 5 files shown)


albe:
See discussion in https://github.com/neos/flow-development-collection/pull/2421#issuecomment-820572791

Neos.Flow/Classes/Cache/CacheManager.php
```diff
@@ -62,6 +63,12 @@ class CacheManager
      */
     protected $persistentCaches = [];
 
+    /**
+     * The ReflectionService is only needed during compile time for handling
```
albe, line 67:
```suggestion
     * The ReflectionService is only needed during compile time for handling flushConfigurationCachesByCompiledClass
```
Writibus interruptus...

Neos.Flow/Classes/Cache/CacheManager.php
```diff
@@ -402,6 +414,42 @@ protected function flushConfigurationCachesByChangedFiles(array $changedFiles):
...
+                }
+            }
+        }
+
+        foreach ($cachesToFlush as $cacheName => $annotationClass) {
+            $this->logger->info(sprintf('A class file containing the annotation "%s" has been changed, flushing related cache "%s"', $annotationClass, $cacheName), LogEnvironment::fromMethodName(__METHOD__));
+            $this->getCache($cacheName)->flush();
+        }
+
+        if (count($cachesToFlush) > 0) {
```
albe, line 447: Not sure, but I think this `count()` check should also happen in the other `flush*Caches...()` methods, no? But maybe I'm missing something

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -469,6 +469,8 @@ public static function initializeObjectManagerCompileTimeFinalize(Bootstrap $boo
         $objectManager->injectLogger($logger);
         $objectManager->initialize($packageManager->getAvailablePackages());
 
+        $cacheManager->injectReflectionService($reflectionService);
```
albe, line 472: A bit hackish to inject the reflection service into the CacheManager in the initialization of the ObjectManager, but that was the simplest place and right before the cache manager is made generally available. Otherwise, this would need another compiletime step that only does this.

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -199,7 +199,16 @@ public function compile()
                 }
             }
         }
-        return $classCount;
+        $this->emitCompiledClasses($compiledClasses);
```
albe, line 202: My first approach had a signal for each single class that would receive the `ProxyClass` as second argument. That might be useful for enhancing the proxies with hooked in functionality via signal/slot, but it would mean that the CacheManager slot would have to somehow receive another signal when all classes are finished or keep reflushing caches again and again.

bwaidelich (commented):
This is a clever approach with surprisingly little amount of code!
I struggle a little with the CacheManager having a dependency to the `ReflectionService` and some route annotations though.

What about a signal per compiled class and a separate

```php
$dispatcher->connect(Proxy\Compiler::class, 'compiledClass', RouteAnnotationCacheFlusher::class, 'registerCompiledClass');
```

That class would contain the reflection logic and flush caches in `shutdownObject()` when classes with Route annotations were built.

(reason for signal per class: don't loop over all classes multiple times)

As discussed: If we add the same functionality for several annotations (Privilege, ...) this might get very slow.

So maybe we could add one level of indirection and have sth like this (class names to be finalized):

```php
$cacheFlusher = $objectManager->get(AnnotationBasedCacheFlusher::class);
$cacheFlusher->registerAnnotationCache(Route::class, 'Flow_Mvc_Routing_Route');
$cacheFlusher->registerAnnotationCache(Route::class, 'Flow_Mvc_Routing_Resolve');

$dispatcher->connect(Compiler::class, 'compiledClass', $cacheFlusher, 'registerCompiledClass');
```

albe:
Extracted the Annotations based cache flushing into a separate class as sugested by Bastian.

albe:
Also have a commit ready that moves all this to a signal per compiled class. I'm still not sure that is more performant than just iterating the classes again instead of calling the SignalDispatcher, which calls the slot, especially given this is our dispatch method: https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/SignalSlot/Dispatcher.php#L145

Also the option to dynamically configure the annotation->caches involves some static method calling, or else we instanciate the AnnotationsCachesFlusher needlessly in most cases (only required during compiletime).
image

So while all that would be nice, it makes this PR more complicated than it needs to be and can be added on top any time. Hence I'd suggest keeping it at this for now.

kitsunet (approved):
Makes sense and the mentioned single signals can probably refactored to later.

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * A slot that flushes caches as needed if classes with specific annotations have changed
+     *
+     * @param array<string> $classNames The full class names of the classes that got compiled
+     * @return void
+     */
+    public function flushConfigurationCachesByCompiledClass(array $classNames): void
+    {
+        $caches = [
+            Neos\Flow\Annotations\Route::class => ['Flow_Mvc_Routing_Route', 'Flow_Mvc_Routing_Resolve']
```
albe, line 59:
```suggestion
            Neos\Flow\Annotations\Route::class => ['Flow_Mvc_Routing_Route', 'Flow_Mvc_Routing_Resolve'],
            Neos\Flow\Annotations\Privilege::class => ['Flow_Security_Policy']
```

bwaidelich: I didn't have the cance to test this (I will tomorrow) but I was wondering: Is flushing the routing/policy caches enough? since the annotations affect the _configuration_ too..?

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        foreach ($cachesToFlush as $cacheName => $annotationClass) {
+            $this->logger->info(sprintf('A class file containing the annotation "%s" has been changed, flushing related cache "%s"', $annotationClass, $cacheName), LogEnvironment::fromMethodName(__METHOD__));
+            $this->cacheManager->getCache($cacheName)->flush();
+        }
+
+        if (count($cachesToFlush) > 0) {
+            $this->logger->info('An annotated class file has been changed, refreshing compiled configuration cache', LogEnvironment::fromMethodName(__METHOD__));
+            $this->configurationManager->refreshConfiguration();
```
bwaidelich, line 82: Ah, I see you already flush the configuration cache.
As a consequence that means that e.g. all routing caches and all configuration caches are flushed whenever a controller with at least one route annotation is being changed. (not sure if that's a problem)

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
bwaidelich: The namespace `Neos\Flow\Annotations\Route` doesn't exist

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
```diff
@@ -0,0 +1,85 @@
...
+
+namespace Neos\Flow\Cache;
+
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Configuration\ConfigurationManager;
+use Neos\Flow\Reflection\ReflectionService;
+use Neos\Flow\Log\Utility\LogEnvironment;
+use Psr\Log\LoggerInterface;
+
+final class AnnotationsCacheFlusher
```
bwaidelich, line 22: This should be a singleton, right?

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
```diff
@@ -0,0 +1,85 @@
...
+        ];
+        $cachesToFlush = [];
+
+        foreach ($classNames as $className) {
+            foreach ($caches as $annotationClass => $cacheNames) {
+                if (!$this->reflectionService->isClassAnnotatedWith($className, $annotationClass)
+                    && count($this->reflectionService->getMethodsAnnotatedWith($className, $annotationClass)) === 0) {
+                    continue;
+                }
+                foreach ($caches[$annotationClass] as $cacheName) {
```
bwaidelich, line 69:
```suggestion
                foreach ($cacheNames as $cacheName) {
```

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
albe: > The namespace `Neos\Flow\Annotations\Route` doesn't exist

Not an issue effectively, because as you know, you can do `My\Madeup\Namespace\Class::class` and it will just resolve to the proper class string without checking if the class actually exists.
But as said, maybe we do want to add those mappings in the specific annotation PRs only.

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
albe: Yep. Though only necessary if you want to work with the insance state (see your PR). Right now the class is stateless.

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
bwaidelich: It was missing the trailing backslash, _that's_ what I meant :)

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
albe: Not sure either tbh :)

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * A slot that flushes caches as needed if classes with specific annotations have changed @see registerAnnotation()
+     *
+     * @param array<string> $classNames The full class names of the classes that got compiled
+     * @return void
+     * @throws NoSuchCacheException
+     */
+    public function flushConfigurationCachesByCompiledClass(array $classNames): void
+    {
+        $cachesToFlush = [];
```
albe, line 81:
```suggestion
        if ($this->annotationToCachesMap === []) {
            return;
        }
        $cachesToFlush = [];
```

Neos.Flow/Classes/Package.php
```diff
@@ -11,9 +11,12 @@
  * source code.
  */
 
+use Neos\Flow\Annotations\Route;
```
albe, line 14:
```suggestion
```

Neos.Flow/Classes/Package.php
```diff
@@ -146,5 +149,11 @@ public function boot(Core\Bootstrap $bootstrap)
 
         $dispatcher->connect(AuthenticationProviderManager::class, 'successfullyAuthenticated', Context::class, 'refreshRoles');
         $dispatcher->connect(AuthenticationProviderManager::class, 'loggedOut', Context::class, 'refreshTokens');
+
+        $dispatcher->connect(Proxy\Compiler::class, 'compiledClasses', function (array $classNames) use ($bootstrap) {
+            $annotationsCacheFlusher = $bootstrap->getObjectManager()->get(AnnotationsCacheFlusher::class);
+            $annotationsCacheFlusher->registerAnnotation(Route::class, ['Flow_Mvc_Routing_Route', 'Flow_Mvc_Routing_Resolve']);
```
albe, line 155:
```suggestion
```

albe:
This is now just the barebone feature without any registered annotations yet. Will merge as is and then let's rebase the annotation PRs and do further testing.

PR #2464, opened 2021-04-16
TASK: Remove generic persistence left-overs

albe:
This removes some left-over namespace imports and the signal/slot connection for `allObjectsPersisted` on the generic persistence.


kdambekalns (commented):
Well, the mocked interface is not to PhpUnit's delight…

PR #2465, opened 2021-04-16
TASK: Add test that covers signal arguments by reference passing

albe:
These tests verify that our current signal/slot dispatching works with byReference arguments.

Related to #2412


albe:
The UnitTests pass, but the FunctionalTest fails replicating the test failure in #2412. I suspected https://github.com/neos/flow-development-collection/pull/2381/files#diff-c2788d9d0354d1d92335269e81751bf64d89569966cb5e52395428aeb1098fcbR152 and/or https://github.com/neos/flow-development-collection/commit/2755619ca7620812706d89d64a51ef06b2d1d384 to be the cause, but apparently they aren't. When did this work? Did it ever?

albe:
Merging as this just covers more things that were not covered before with a test.

PR #2466, opened 2021-04-17
Suggestion: Make AnnotationCacheFlusher usable for 3rd parties

bwaidelich:
Adds a method `registerAnnotation()` that allows further annotations
to lead to the corresponding caches to be flushed.

This also tweaks `flushConfigurationCachesByCompiledClass()` and makes
the class a singleton.

### DIFF
--- a/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
+++ b/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
@@ -13,12 +13,16 @@ declare(strict_types=1);
 
 namespace Neos\Flow\Cache;
 
+use Neos\Cache\Exception\NoSuchCacheException;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Configuration\ConfigurationManager;
 use Neos\Flow\Reflection\ReflectionService;
 use Neos\Flow\Log\Utility\LogEnvironment;
 use Psr\Log\LoggerInterface;
 
+/**
+ * @Flow\Scope("singleton")
+ */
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
```diff
@@ -48,25 +52,40 @@ final class AnnotationsCacheFlusher
...
+     */
+    private $annotationToCachesMap = [];
+
+    /**
+     * Register an annotation that should trigger a cache flush
+     *
+     * @param string $annotationClassName fully qualified class name of the annotation
+     * @param string[] $cacheNames Cache names to flush if a class containing the given annotation is compiled (e.g. ["Flow_Mvc_Routing_Route", Flow_Mvc_Routing_Resolve"])
+     */
+    public function registerAnnotation(string $annotationClassName, array $cacheNames): void
```
albe, line 67: Having this as instance method means you have to create the insance to add a new mapping. Not sure that's much of an issue, but it should at least be considered. The $annotationToCachesMap could as well be static.

bwaidelich: True. I don't think that this would be a real problem. And for the "core mappings" we only create the instance in the slot anyways

albe: Was mostly thinking 3rd-party now. Since otherwise the class would only be instanciated at compile time, user code would need to act "clever" to avoid instanciating it too eagerly.

PR #2468, opened 2021-04-18
FEATURE: Support PHP8 attributes

albe:
This allows to use all existing Annotations as PHP8 Attributes and makes the ReflectionService pick up attributes like annotations.
Hence all `is*AnnotatedWith()` and `*Annotation()` methods will return attribute classes as if they were annotations.

```
#[Flow\Scope("singleton")]
class MyClass {

    /**
     * @var LoggerInterface
     */
    #[Flow\Inject]
    protected $logger;
```

Note though, that this means a class that has both annotation and the equal attribute will behave as if all annotations were duplicated.

Also, in case you manually instanciated an Annotation class, you need to adjust to the changed constructor, which no longer takes a named array, but the list of actual properties.
In most cases instead of `new Flow\Inject($args)` you can probably do `new Flow\Inject(...$args)` with PHP8 and named parameters.

For Doctrine Annotations - see https://github.com/doctrine/orm/pull/8266 which will be available with 2.9

### DIFF
--- a/Neos.Flow/Classes/Annotations/After.php
+++ b/Neos.Flow/Classes/Annotations/After.php
@@ -11,30 +11,28 @@ namespace Neos\Flow\Annotations;
  * source code.
  */
 
+use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
+
 /**
  * Declares a method as an after advice to be triggered after any
  * pointcut matching the given expression.
  *
  * @Annotation
+ * @NamedArgumentConstructor
  * @Target("METHOD")
  */
+#[\Attribute(\Attribute::TARGET_METHOD)]
 final class After
 {
     /**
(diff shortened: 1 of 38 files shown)


albe:
This is a reduced subset of #2404 that only contains the attributes support without changing any functionality

kdambekalns (approved):
Looks good by reading, I like how the annotation implementations get simpler!

daniellienert:
@albe - I also really like it ( as far as I understand it :) - would you take care of the failing psalm tests?

skurfuerst (approved):
looks good by reading :) Great job! Instanciating annotations manually is a case I have never seen in the wild, so for me that's safe.

albe:
Regarding psalm, I guess we need to do a baseline update in master anyway before the release. I can take care on friday if it's not too late

PR #2469, opened 2021-04-19
TASK: Tweak polyfill replacements

kdambekalns:
The replacements added according to the symfony/polyfill README (and
tweaked recently to actually appear in the split manifest) lead to an
installation issue (composer/composer#9834)

Thus this removes the PHP 5 polyfills (those not being replaced should
not be an issue) to fix that. Also, since we require PHP 7.1, the 7.2
polyfill must not be replaced.

### DIFF
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -54,12 +54,8 @@
     },
     "replace": {
         "typo3/flow": "self.version",
-        "symfony/polyfill-php54": "*",
-        "symfony/polyfill-php55": "*",
-        "symfony/polyfill-php56": "*",
         "symfony/polyfill-php70": "*",
         "symfony/polyfill-php71": "*",
-        "symfony/polyfill-php72": "*",
         "symfony/polyfill-mbstring": "*"
     },
     "suggest": {
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -55,8 +55,7 @@
     "replace": {
         "typo3/flow": "self.version",
         "symfony/polyfill-php70": "*",
-        "symfony/polyfill-php71": "*",
-        "symfony/polyfill-mbstring": "*"
+        "symfony/polyfill-php71": "*"
     },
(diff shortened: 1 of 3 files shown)


bwaidelich:
Thanks!
For reference: https://github.com/composer/composer/issues/9834#issuecomment-822066383

bwaidelich (approved):
+1 by reading

Neos.Flow/composer.json
```diff
@@ -54,12 +54,8 @@
     },
     "replace": {
         "typo3/flow": "self.version",
-        "symfony/polyfill-php54": "*",
-        "symfony/polyfill-php55": "*",
-        "symfony/polyfill-php56": "*",
         "symfony/polyfill-php70": "*",
         "symfony/polyfill-php71": "*",
-        "symfony/polyfill-php72": "*",
         "symfony/polyfill-mbstring": "*"
```
stof, line 59: replacing `symfony/polyfill-mbstring` without having a requirement for `ext-mbstring` looks also wrong to me

kdambekalns: Right… it should be replaced in Utility.Unicode, done.

composer.json
```diff
@@ -33,12 +33,8 @@
     "replace": {
         "typo3/eel": "self.version",
         "typo3/flow": "self.version",
-        "symfony/polyfill-php54": "*",
-        "symfony/polyfill-php55": "*",
-        "symfony/polyfill-php56": "*",
         "symfony/polyfill-php70": "*",
         "symfony/polyfill-php71": "*",
-        "symfony/polyfill-php72": "*",
         "symfony/polyfill-mbstring": "*",
```
stof, line 38: this one does not require `ext-mbstring` either

kdambekalns: It does in line 31…

stof: indeed. I missed it as it is not sorted with other extensions

albe (approved):

PR #2475, opened 2021-05-02
BUGFIX: Fix arguments to redis::watch() command

albe:
According to https://github.com/phpredis/phpredis#example-114 `watch()` takes a list of keys as array, but in https://github.com/phpredis/phpredis/blob/develop/redis.c#L1360 / https://github.com/phpredis/phpredis/blob/develop/redis_commands.c#L3391 the arguments are passed like for e.g. sdiff, which takes a variable amount of string keys.

Related to #2052

### DIFF
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -153,7 +153,7 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
 
         do {
             // Watch the given keys for changes so if the TTL is changed during the transaction it can be retried
-            $this->redis->watch($keysToWatch);
+            $this->redis->watch(...$keysToWatch);
 
             foreach ($redisTags as $i => $tag) {
                 $expire = $this->calculateRemainingLifetimeForKey($tag['key'], $lifetime);
--- a/Neos.Cache/Tests/Unit/Backend/RedisBackendTest.php
+++ b/Neos.Cache/Tests/Unit/Backend/RedisBackendTest.php
@@ -245,11 +245,9 @@ class RedisBackendTest extends BaseTestCase
         $this->redis->expects($this->once())
             ->method('watch')
             ->with(
-                [
-                    'Foo_Cache:tag:baz',
-                    'Foo_Cache:tags:foo',
-                    'Foo_Cache:entry:foo'
-                ]
+                'Foo_Cache:tag:baz',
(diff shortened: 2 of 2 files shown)


albe:
This was noticed after upmerging, when psalm complained about mismatching argument types. Another reason why psalm is really a life-safer and worth the added maintenance cost.

nlx-lars:
@albe if anything, it just doesn't watch the right keys. I just ran the functional tests against a v6.2.2 instance of redis to make sure.

I'll try to hack a test together to make sure the right keys are watched.

nlx-lars:
@albe it does not matter if a single array or multiple strings are passed to `watch()`, both work:

```
$clientA = $getRedis();
$clientA->watch(['foo', 'bar']);
// or $clientA->watch('foo', 'bar');
$clientA->multi();

$clientB = $getRedis();
$clientB->set('bar', 'baz');

$clientA->set('bar', 'foo');
$result = $clientA->exec();

self::assertFalse($result);
```

As soon as I change the key that `$clientB` sets, the result is `true` and the test fails as expected.

Now, writing a test for this that uses `RedisBackend::set()` seems hard, as I have no way of calling another redis client between the calls to `watch()` and `execute()`.

albe:
Thanks for checking so quickly! That's relieving to know! So I'll just change the invocation to `...$tagsToWatch` to make psalm happy in 6.0+ and leave it as is in 5.3 which goes into security-only today with the release.

kdambekalns:
Late to the party, but…

> … one complaint is that `redis->watch()` takes a string, but we are providing an array.
> See https://github.com/phpredis/phpredis#watch-unwatch

That documentation says (edited for clarity):

```php
$redis->watch('x');
// or for a list of keys:
$redis->watch(['x', 'another key']);
```

That being said, all these run error-free:

```php
$redis->watch('x');
$redis->watch(['y']);
$redis->watch('x', 'y');
$kyes = ['a', 'b', 'c'];
$redis->watch(...$keys);
```

So all variants seem to work.

kdambekalns:
> as I have no way of calling another redis client between the calls to `watch()` and `execute()`.

Only way I could think of is a call to a PHP sub-process changing it… 🙈

nlx-lars:
@kdambekalns but I wouldn't be able to spawn that subprocess in between `watch()` and `execute()`. Maybe one way to work around this would be to rapidly change the ket in the subprocess. 🤔

PR #2476, opened 2021-05-08
FEATURE:  I18n.translate() now accept $source to contain dots instead of only a path to the translation file

etlam:
**What I did**
`translateByExplicitlyPassedOrderedArguments()` and `I18n.translate()` now accept `$source` argument to contain dots instead of only a path to the translation file.

When we use translations we use for example the shorthand
```php
{I18n.translate('Muensmedia.DistributionPackage:NodeTypes.Content.Todo.Container:ui.label')}
```
when we want to pass arguments we had to use
```php
${I18n.translate('progress', null, {solved: this.checkedElementsCount, total: this.checkboxCount}, 'NodeTypes/Content/Todo/Container', 'Muensmedia.DistributionPackage')}
```
As you can see, you have to pass the **path** to the translation file instead of the well known dot-notation.

This commit enables you to use also the well known dot-notation for the source argument:
```php
${I18n.translate('progress', null, {solved: this.checkedElementsCount, total: this.checkboxCount}, 'NodeTypes.Content.Todo.Container', 'Muensmedia.DistributionPackage')}
```

**How I did it**
In the method `translateByShortHandString()` we already replace dots with slashes, so I just copied this behavior to the method `translateByExplicitlyPassedOrderedArguments()`
https://github.com/neos/flow-development-collection/blob/5b7b57523ab1a3b05105227e0a5266ece2777038/Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php#L140

**How to verify it**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
+++ b/Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
@@ -112,7 +112,7 @@ class TranslationHelper implements ProtectedContextAwareInterface
         $translationParameterToken
             ->value($originalLabel)
             ->arguments($arguments)
-            ->source($source)
+            ->source(str_replace('.', '/', $source))
             ->package($package)
             ->quantity($quantity);
 


etlam:
@markusguenther How do we get this pull request reviewed?

albe (approved):
Since it's just a port of the same behaviour from a sibling method

markusguenther:
Thanks Malte for this PR.

As this is a feature this will not be merged to version 5.3. The only branch for features is master. So this could be, if there are no concerns, go into version 7.2 

If you need that for version 5.3 you could of course use composer patches or so.

But for Neid in general we need to rebase that to master. And then you just wait ;)

Sometimes it takes some time.

albe:
> As this is a feature this will not be merged to version 5.3. The only branch for features is master. So this could be, if there are no concerns, go into version 7.2

Given the translate helper method behaves different on the `$source` when the `$id` is a label pattern, one could argue this to be a BUGFIX.
Anyway, it needs to target 6.3, as 5.3 is security-only by now, but just changing the target branch should work out (always much easier to retarget upwards).

albe:
Can someone take care of the rebase/retarget?

etlam:
I will never user the web editor again, now it is a bit strange as my branch is still named `5.3` but contains my commit rebased on `6.3`. But as far as I know I can not change the name of the source branch during the pull request.  
Is it fine this way or should I start a new merge request and give my branch a nice name?

albe:
> But as far as I know I can not change the name of the source branch during the pull request.

Unfortunately not. I do totally love the Web Editor for all simple fixes, but for anything other than editing a single file quickly it's really unfit, yes :| It's okay to merge as this, but you might still want to reset your fork's 5.3 branch back to origin in case you do want to use it again. So you could rename this branch, create a new PR and reset your 5.3. If you don't care that's fine also though :)

etlam:
I will reset my fork after this MR is merged. What a mess :stuck_out_tongue_closed_eyes:

albe:
ping - can we get another approval on this FEATURE-BUGFIX-Backport? :)

kitsunet (approved):

jonnitto:
This PR leads to an error if no name is given. Before it was possible to set the filename to `null`:

```
I18n.translate('foo', null, [], null, 'Foo.Bar')
```

to make it work you now have to set the file to `Main`:


```
I18n.translate('foo', null, [], 'Main', 'Foo.Bar')
```

I would consider replacing

```
->source(str_replace('.', '/', $source))
```

with something like  

```
->source(isset($source) ? str_replace('.', '/', $source) : null)
```

Wdyt, @albe / @kitsunet ?

albe:
Interesting, I would have thought it needs to be the same as in `translateByShortHandString`, but there the source is effectively required. Given the source parameter is annotated `@param string $source Name of file with translations` - i.e. it being nullable is not mentioned anywhere and it also being required as per documentation, that's a hidden feature really. I do see how in some cases it might be easier to pass in a `null` rather than do the null-coalescing in the script.

If we do change it, let's also update the parameter docblock

jonnitto:
If you read https://neos.readthedocs.io/en/stable/References/EelHelpersReference.html#translation-translate-id-originallabel-arguments-source-package-quantity-locale there is cleary written that the value is optional

albe:
Hm, I guess that's because it is an optional argument with a default value, i.e. you can call the PHP method with just `->translateByExplicitlyPassedOrderedArguments('foo.bar')`, but it's a bit misleading, because that does not mean you can just pass `null`. In JS for this case there is `undefined` that can be passed, which is indistinguishable from a missing argument and makes the language fall back to the method declared default:
image

Anyway, thanks for the PR! :)

PR #2478, opened 2021-05-17
BUGFIX: Apply Content-Type header from PSR-7 Responses

Sebobo:
Before this change a content type set as header in a
rendered view would be ignored.
As example this resulted in an incorrect header for FusionViews
which rendered a PSR-7 Response with a non html Content-Type.

This broke due the middleware change in Flow 6, as the behaviour of [`applyToHttpResponse`](https://github.com/neos/flow-development-collection/blob/701072a0c240a32442d5b80ad51464ce9571d57a/Neos.Flow/Classes/Mvc/ActionResponse.php#L291) was different than the now used `replaceHttpResponse` / [`buildHttpResponse`](https://github.com/neos/flow-development-collection/blob/9d4111765b729d59a3bec4085f1d8288f7cf74b0/Neos.Flow/Classes/Mvc/ActionResponse.php#L312).

Fixes: https://github.com/neos/neos-development-collection/issues/3346

**What I did**

Set response content type based on content type header optionally provided by a rendered view.

**How I did it**

If the view returns a PSR-7 Response and contains a `Content-Type` header, it is applied to the final response.

**How to verify it**

Run tests.

**Checklist**


Neos.Flow/Tests/Unit/Mvc/Controller/ActionControllerTest.php
```diff
@@ -314,6 +316,38 @@ public function processRequestUsesContentTypeFromActionResponse($supportedMediaT
...
+        $this->actionController->method('theActionAction')->willReturn(null);
+
+        $this->inject($this->actionController, 'objectManager', $this->mockObjectManager);
+
+        $mockMvcPropertyMappingConfigurationService = $this->createMock(Mvc\Controller\MvcPropertyMappingConfigurationService::class);
+        $this->inject($this->actionController, 'mvcPropertyMappingConfigurationService', $mockMvcPropertyMappingConfigurationService);
+
+        $mockHttpRequest = $this->getMockBuilder(ServerRequestInterface::class)->disableOriginalConstructor()->getMock();
+        $mockHttpRequest->method('getHeaderLine')->with('Accept')->willReturn('application/xml');
+        $mockHttpRequest->method('getHeaderLine')->with('Accept')->willReturn('application/xml');
```
albe, line 336:
```suggestion
        $mockHttpRequest->method('getHeaderLine')->with('Accept')->willReturn('application/xml');
```

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -789,6 +789,9 @@ protected function renderView()
 
         if ($result instanceof ResponseInterface) {
             $this->response->replaceHttpResponse($result);
+            if ($result->hasHeader('Content-Type')) {
+                $this->response->setContentType($result->getHeaderLine('Content-Type'));
```
albe, line 793: Hmmm... this is a bit confusing to me, because IMO the `replaceHttpResponse()` should already make sure that the ContentType (and any other headers really) that is potentially set inside the HttpResponse `$result` should be applied. But apparently it isn't - sooo... what is happening then?
Only thing I could imagine is this:
- the `ActionResponse` (`$this->response`) already has a ContentType set (e.g. from content negotiation)
- after replacing the `HttpResponse`, the `ActionResponse` gets merged into the `HttpResponse` (inside the dispatch middleware - see https://github.com/neos/flow-development-collection/blob/7.0/Neos.Flow/Classes/Mvc/DispatchMiddleware.php#L46 resp. https://github.com/neos/flow-development-collection/blob/7.0/Neos.Flow/Classes/Mvc/ActionResponse.php#L314)
- at that place, the ContentType from the view returned `HttpResponse` is overwritten again

So given that, I think even though this fixes this single symptom, there is a bigger general logic issue below that probably needs to be fixed - i.e. a `replaceHttpResponse()` not preventing the `ActionResponse` from overwriting properties in that HttpResponse again. Maybe that is even intended for some properties of the `ActionResponse`? Dunno. It basically boils down to what is expectation for the response of an action like this:

```php
public function weirdResponseAction()
{
    $this->response->setContentType('application/xml');
    $this->response->setRedirectUri(new Uri('somewhereelse'), 304);
    $this->response->setContent('Hello World!');
    $this->response->setCookie(new Cookie(...));
    $this->response->setHttpHeader('X-My-Header', 'My-Header-Value');
    $this->response->replaceHttpResponse(new Response(200, ['Content-Type' => 'application/json'])); // which is equivalent to letting the view render a PSR Response
}
```
- should the response status be 200 or 304?
- should the response header `Location` be set to indicate a redirect? (should probably be in sync with the above)
- should the response content be "Hello World!" or empty?
- should the response content type be "application/json" or "application/xml"?
- should the response header `X-My-Header` be set?
- should the response contain a `Set-Cookie` header?
- should any of the above change if the `replaceHttpResponse()` call happened earlier?

So is `replaceHttpResponse()` a full/partial/conditional escape hatch from the `ActionResponse` abstraction?

/cc @bwaidelich @kitsunet

albe (approved):
I'm inclined to merge this as is for now to just get the Neos bug fixed, then discuss and decide on how the `replaceHttpResponse()` should actually behave in all those edge-cases in another PR.

Sebobo:
Thx. And I agree :)

PR #2480, opened 2021-05-17
BUGFIX: Flow CacheAdapter extends doctrine CacheAdapter

### LINKED ISSUE #2479: Incompatible with "doctrine/persistence" 2.2
### Description

On May 15th "doctrine/persistence 2.2 was released, unfortunately Flow is not compatible with it. 
I added "doctrine/persistence": "2.1.*" to my composer.json as workaround.

### Steps to Reproduce

1. composer update

#### Expected behavior

Should work

#### Error

Exception:
`Cannot convert cache to PSR-6 cache

  Type: BadMethodCallException
  File: Packages/Libraries/doctrine/persistence/lib/Doctrine/Persistence/Mapping/Ab
        stractClassMetadataFactory.php
  Line: 91`

Neos: 7.1

albe:
This fixes the exception for the change in doctrine/persistence 2.2, that was an intentional documented bc-break for moving towards PSR-6 

See https://github.com/doctrine/persistence/pull/144/files#diff-9a566def484817e003ce815d52fae09a8c2b459c443fe015047ae043e170ef1a

Resolves #2479
Related to https://github.com/doctrine/persistence/issues/179

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/CacheAdapter.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/CacheAdapter.php
@@ -11,7 +11,7 @@ namespace Neos\Flow\Persistence\Doctrine;
  * source code.
  */
 
-use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\CacheProvider;
 use Neos\Flow\Annotations as Flow;
 use Neos\Cache\Frontend\FrontendInterface;
 use Neos\Flow\Security\Context;
@@ -19,7 +19,7 @@ use Neos\Flow\Security\Context;
 /**
  * Cache adapter to use Flow caches as Doctrine cache
  */
-class CacheAdapter implements Cache
+class CacheAdapter extends CacheProvider
 {
     /**
      * @var FrontendInterface


alcaeus:
Please note that this was a mistake on my part and fill be fixed in the upcoming 1.11.1 and 2.0.1 releases of doctrine/cache: doctrine/cache#370. Please let me know if that patch fixes the issue for you.

bwaidelich:
do we still need this fix if doctrine/cache#370 is merged and, if so, is it compatible with all doctrine versions we currently support?

alcaeus:
The fix is no longer necessary with doctrine/cache 1.11.1 or 2.0.1 and doctrine/persistence 2.2.1. Please let me know if you're still running into issues so I can look into them. Sorry for the mess, but turns out building compatibility layers is hard 😅

bwaidelich:
> The fix is no longer necessary with doctrine/cache 1.11.1 or 2.0.1 and doctrine/persistence 2.2.1

Thanks for the update!

> Please let me know if you're still running into issues so I can look into them

Yep, we'll do.

> Sorry for the mess, but turns out building compatibility layers is hard

No worries, you can't make an omelette without breaking eggs *g

albe:
Just checked and reran the build for one PR where the error occured. Unfortunately it seems like there is still something off, though I haven't had time to look closer:

https://github.com/neos/flow-development-collection/pull/2478/checks?check_run_id=2621383474
image

```
Return value of Doctrine\Common\Cache\Psr6\CacheAdapter::commit() must be of the type bool, null returned
Type: TypeError
  File:
Packages/Libraries/doctrine/cache/lib/Doctrine/Common/Cache/Psr6/CacheAdapt
        er.php
  Line: 210
```

alcaeus:
Hmm, L210 is returning the return value of `save` from the wrapped adapter. Can you please check your cache implementation to make sure it returns a bool as mandated by the interface? https://github.com/doctrine/cache/blob/163074496dc7c3c7b8ccbf3d4376c0187424ed81/lib/Doctrine/Common/Cache/Cache.php#L53

albe:
> L210 is returning the return value of save from the wrapped adapter.

That's very possibly the issue, so something on our side! Thanks for taking a look! I can check this evening for a solution

sorenmalling:
@albe 
Something along these lines in the `save` method

```
try {
            $this->cache->set($this->convertCacheIdentifier($id), $data, [], $lifeTime);
            return true;
            return false;
        }
```

since `set` does not return any value it self, but the implementation can throw a exception, than we can catch

sorenmalling:
@albe 

https://github.com/neos/flow-development-collection/pull/2481

albe:
Closing in favor of #2481

PR #2482, opened 2021-05-21
BUGFIX: Return boolean in doctrine CacheAdapter::save

sorenmalling:
Follow up to #2480 when the issue was found to live inside our CacheAdapter and missing return value

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/CacheAdapter.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/CacheAdapter.php
@@ -84,12 +84,8 @@ class CacheAdapter implements Cache
      */
     public function save($id, $data, $lifeTime = 0)
     {
-        try {
-            $this->cache->set($this->convertCacheIdentifier($id), $data, [], $lifeTime);
-            return true;
-        } catch (\Exception $exception) {
-            return false;
-        }
+        $this->cache->set($this->convertCacheIdentifier($id), $data, [], $lifeTime);
+        return true;
     }
 
     /**
--- a/Neos.Flow/Classes/Persistence/Doctrine/CacheAdapter.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/CacheAdapter.php
@@ -84,7 +84,12 @@ class CacheAdapter implements Cache
      */
     public function save($id, $data, $lifeTime = 0)
     {
-        $this->cache->set($this->convertCacheIdentifier($id), $data, [], $lifeTime);
+        try {
+            $this->cache->set($this->convertCacheIdentifier($id), $data, [], $lifeTime);
+            return true;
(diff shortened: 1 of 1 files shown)


albe (approved):

Neos.Flow/Classes/Persistence/Doctrine/CacheAdapter.php
```diff
@@ -84,7 +84,12 @@ public function contains($id)
      */
     public function save($id, $data, $lifeTime = 0)
     {
-        $this->cache->set($this->convertCacheIdentifier($id), $data, [], $lifeTime);
+        try {
+            $this->cache->set($this->convertCacheIdentifier($id), $data, [], $lifeTime);
+            return true;
+        } catch (\Exception $exception) {
+            return false;
+        }
```
bwaidelich, line 92: Mh, I have to say that I share the concerns @albe expressed in #2481 
Silently swallowing exceptions is mostly a bad idea.
Can't we just do

```suggestion
        $this->cache->set($this->convertCacheIdentifier($id), $data, [], $lifeTime);
        return true;
```
?

albe: That was my first version of the code I had in mind. But then I realized:

> since this in inside the doctrine adapter, it is only called from the doctrine side, which does expect this to return false on error so... this is probably fine

So in essence I'm okay either way

bwaidelich: But wouldn't it be helpful to know that a cache entry could not be created because the identifier is invalid for example?
..or if the backend could not be connected to. Those things are almost impossible to debug if they just silently don't work

bwaidelich (approved):
Thank you! :)

Neos.Flow/Classes/Persistence/Doctrine/CacheAdapter.php
kdambekalns: Didn't see this until now, but my vote would have gone to "catch, log, return false". 😎

bwaidelich: But in which scenario would it be helpful to avoid the exception?

kdambekalns: Well, if setting the cache entry fails, this is something that might degrade performance, but the result of whatever needed to be cached could very well be valid. Plus, if the calling code expects false in case the operation failed, that should be honoured. Throwing might break error handling further up the stack.

albe: Feel free to create a new PR that adds that change on top, I'm not opposed :) For now, the bug is solved and that was my main concern

PR #2484, opened 2021-05-27
PHP 8 Unit-Test fixes

gerdemann:
**Checklist**


Neos.Flow/Classes/Reflection/ParameterReflection.php
```diff
@@ -43,12 +43,12 @@ public function getDeclaringClass()
...
     {
         try {
-            $class = parent::getClass();
+            $class = parent::getType();
         } catch (\Exception $exception) {
             return null;
         }
 
-        return is_object($class) ? new ClassReflection($class->getName()) : null;
+        return is_object($class) && !$class->isBuiltin() ? new ClassReflection($class->getName()) : null;
```
albe, line 51: Are there actually any object-types that are builtin? `stdClass` for example isn't considered "builtin" (and even then, why shouldn't we get a class reflection?) What was the reason for this condition @gerdemann ?

See https://www.php.net/manual/de/reflectionnamedtype.isbuiltin.php#125927

gerdemann: Previously it was `getClass()`, which returned null for eg. "array". But with `getType()` you get back "array" for an "array" type hint. That's why I added the condition so that you don't get a ReflectionClass of "array".
Or have I made a thinking error, then I can take that out.

albe: Indeed, totally overlooked that subtle difference!

albe (approved):

PR #2488, opened 2021-05-31
TASK: Constrain doctrine/orm to < 2.9 due to currently breaking

albe:
See https://github.com/doctrine/orm/pull/8266#issuecomment-850421549


albe:
6.3 can't even install 2.9 anyway, because that requires doctrine/common 3.0 and we are still on 2.x here

PR #2491, opened 2021-06-07
BUGFIX: Avoid open_basedir restriction with realpath

mhsdesign:
## Bug

I encountered the following error in the setup (/setup/index?step=1):

(Plesk / PHP7.4 / Flow7.1)

image

### The error source:
```
$realPhpBinary = realpath(PHP_BINARY);
```

My web hoster doesn't allow me to change the open_basedir to include "/usr/local/php74/bin/php".

## Fix

But using:
`php -r "echo realpath(PHP_BINARY);"`
in `exec()` will work and bypass open_basedir.

### implemented:

```
exec(PHP_BINARY . ' -r "echo realpath(PHP_BINARY);"', $output);
$realPhpBinary = $output[0];
```
Tested with:
(Plesk / PHP7.4 / Flow7.1)

### similar exec use:
 `exec()` is also used in a similar manner on line 844:
```exec($phpBinaryPathAndFilename . ' -r "echo realpath(PHP_BINARY);"', $output, $result);```

### realpath?
... using realpath was introduced with #2032

## Recap

This change brings up the compatibility for some ISPs(web hosting)

By getting the realPhpBinary see #2032:
`$realPhpBinary = realpath(PHP_BINARY);`
a Neos\Flow\Error\Exception is thrown with the Code: 1355480641
`Warning: realpath(): open_basedir restriction in effect. File(/usr/local/php74/bin/php) is not within the allowed path(s)`
on the most(rather all) web hosting platforms(f.x. Plesk).

By using system commands to get the realpath inside exec() this behavior can be avoided.

### CODE REVIEW

**Comment by albe** [comment]
Hey, thanks a lot for this! I think this makes sense given the constraints some hosters put on the environment. We need to target 6.3, since 5.3 is in security-only mode by now (which means this "bug" will stay in versions 5.3-6.2 unfortunately)

**Review by albe — APPROVED** [review]

**Comment by albe** [comment]
I'll take care of the accidential 5.3 changelog when upmerging.

**Review by kdambekalns — COMMENTED** [review]
So `realpath()` is called in lines 831 and 837, and we call it again on the result in line 845? Why do we (need to) do this?

**Comment by sorenmalling** [comment]
> So `realpath()` is called in lines 831 and 837, and we call it again on the result in line 845? Why do we (need to) do this?

True - in the first try we use
```
static::ensureCLISubrequestsUseCurrentlyRunningPhpBinary($settings['core']['phpBinaryPathAndFilename']);
```
so from our Settings.

In the second (added with this patch) PHP_BINARY is used.

**Comment by albe** [comment]
> So realpath() is called in lines 831 and 837, and we call it again on the result in line 845?

Yes, but through potentially different binaries:
```
        // Try to resolve which binary file PHP is pointing to
        exec($phpBinaryPathAndFilename . ' -r "echo realpath(PHP_BINARY);"', $output, $result);
        if ($result === 0 && sizeof($output) === 1) {
            // Resolve any wrapper
            $configuredPhpBinaryPathAndFilename = $output[0];
        } else {
            // Resolve any symlinks that the configured php might be pointing to
            $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
        }
```

`exec($phpBinaryPathAndFilename . ' -r "echo realpath(PHP_BINARY);"', $output, $result);` will make the PHP binary that is configured evaluate `realpath(PHP_BINARY)` so we get the full path of that binary.
If that call fails, we instead fall back to `$configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);`, i.e. we resolve the full path that was configured (which btw. might also fail in the case this PR tries to fix, but there's no workaround).

So, what we actually try to accomplish is verify if `PHP_BINARY(currentProcess) === PHP_BINARY(configuredPhpPath)`. This change only avoids calling `realpath()` on a system path (the PHP_BINARY path) within the web context, which is apparently prohibited by the hoster.

> In the second (added with this patch) PHP_BINARY is used.

It is not added, just the way it is called is changed to not do the resolution of the PHP_BINARY path in the current process, but instead invoke `PHP_BINARY` and let itself resolve it's own path.

**Comment by mhsdesign** [comment]
https://github.com/neos/flow-development-collection/pull/2491#issuecomment-869525569 so much detail for such a small pull request 😂💪

**Comment by albe** [comment]
> [#2491 (comment)](https://github.com/neos/flow-development-collection/pull/2491#issuecomment-869525569) so much detail for such a small pull request 😂💪

The less lines of code, the more likely it is that it has some subtle depth involved ;)
Also your initial PR message is detail level over 9000, which is awesome! :)

@kdambekalns @sorenmalling any further questions or doubts?

**Review by kdambekalns — APPROVED** [review]
Fine, but the `5326.rst` should not go into 6.3…

**Review by daniellienert — APPROVED** [review]
Looks good apart from the already said changelog file

**Comment by mhsdesign** [comment]
i dont know where this file came from 😅 should i just delete it here?

**Comment by albe** [comment]
That's an artifact of changing the base-branch (and likely the 5.3 changelog not being upmerged yet). So don't worry :)
You will need to reset your fork's '5.3' branch though in case you might want to do something with that in the future, as it now contains these commits from 6.3

**Comment by mhsdesign** [comment]
This seems to cause a tiny minor performance regression.

Booting minimal flow (only configuration manager), and running `Scripts::buildPhpCommand` went from `0.025s` to `0.044s` on my super fast machine ... it might be a more measurable difference on slower systems.

@fcool as we just discussed things in similar direction, do you think its worth optimizing this?

either
- A: runtime cache the `buildPhpCommand` 
- B: remove the overhead of a second `exec` by reverting this and first using native `realpath` and only in case of an open base dir restriction use exec?
- ...


@kitsunet might also be interested, as i want to replace the neos/setup php checking logic by using `Scripts::buildPhpCommand` - so it should be fast ;) (our duplicate checking code takes also just `0.026s`)

**Comment by fcool** [comment]
A would not be hard and makes perfect sense, as we are going to call it again and again in a "cold" instance...

I don't get "B" though.

In general I would probably consider a setup to be broken, where the binary "in path" symlinks to a restriced location... One could argue, that this is a security bug of openbasedir restrictions, to be able to work at all...

**Comment by mhsdesign** [comment]
> I don't get "B" though.

```php
$realPhpBinary = @realpath(PHP_BINARY);
if ($realPhpBinary === false) {
    exec(PHP_BINARY . ' -r "echo realpath(PHP_BINARY);"', $output);
    $realPhpBinary = $output[0];
}
```


A would also mean, that in case you provide different $settings (which should never happen) you would't get the expected result. But i'd be fine with this?


A & B implemented in https://github.com/neos/flow-development-collection/pull/3119

PR #2493, opened 2021-06-10
BUGFIX: retrieve package by case insensitive packageKey

gjwnc:
**What I did**

The `PackageManager::getPackage($packageKey)` method should throw an exception or return the found package. There is a case, such that `getPackage` returns `null`. In recent php versions, this causes a php error because the return value of the api `public function getPackage($packageKey): PackageInterface` is not met:

```
Exception in line 514 of /var/www/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/Neos_Flow_ResourceManagement_Streams_ResourceStreamWrapper.php: Return value of Neos\Flow\Package\PackageManager::getPackage() must implement interface Neos\Flow\Package\PackageInterface, null returned - See also: 202106101712258f223a.txt
```

In older flow versions (4.0 and up), this might also be a problem, because the method actually can return null instead of throwing an exception.

**Problem analysis**

The problem occures, because the check, if an exception should be thrown by `isPackageAvailable()`, ignores the case during the check, whereas the actual return statement `return $this->packages[$packageKey];` needs the correct case.

**How I did it**

I'm using `$this->getCaseSensitivePackageKey($packageKey)` to retrieve the key in the correct case, such that `$this->packages` returns the correct package.

**How to verify it**

A call like `$packageManager->getPackage('Neos.some.package.with.wrong.case')` should throw a php error in recent versions.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Package/PackageManager.php
+++ b/Neos.Flow/Classes/Package/PackageManager.php
@@ -751,7 +751,8 @@ class PackageManager implements PackageManagerInterface
             $packageConfiguration = $this->preparePackageStateConfiguration($packageKey, $packagePath, $composerManifest, $state);
             if (isset($newPackageStatesConfiguration['packages'][$composerManifest['name']])) {
                 throw new PackageException(
-                    sprintf('The package with the name "%s" was found more than once, please make sure it exists only once. Paths "%s" and "%s".',
+                    sprintf(
+                        'The package with the name "%s" was found more than once, please make sure it exists only once. Paths "%s" and "%s".',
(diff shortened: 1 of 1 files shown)


jonnitto:
Hey @gjwnc 

Thank you for your PR. I'll try to get some flow experts for the needed review

bwaidelich:
Thanks for this contribution. At first sight the change makes sense to me, but IMO it should target branch `6.3` as lowest supported version (see https://www.neos.io/features/release-process.html)

albe (approved):
Makes sense and looks good, thanks!

PR #2494, opened 2021-06-14
BUGFIX: Allow for multiple authentication header

### LINKED ISSUE #2490: UsernamePasswordHttpBasic does not work if the app has also an provider with BearerToken
### Description

I planned to replace my own BearerToken with the new one from Flow 7.1, but had to find out that the application would then no longer allow any other type of Authorization header like "Basic".
My application has 2 authentication Provider, one JWT (bearer token) and one AppClient (Basic, custom Client id).

With Basic Authorization header:
Error message: Could not extract access token from Authorization header: "Bearer" keyword is missing

Flow: 7.1

daniellienert:
This allows to use the BearerToken in combination with other
authentication header bases authentications like UsernamePasswordHttpBasic

closes #2490

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
+++ b/Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
@@ -50,14 +50,14 @@ class BearerToken extends AbstractToken implements SessionlessTokenInterface
             return;
         }
 
-        $authorizationHeader = $httpRequest->getHeader('Authorization')[0];
+        $this->setAuthenticationStatus(TokenInterface::NO_CREDENTIALS_GIVEN);
 
-        if (substr($authorizationHeader, 0, 7) !== 'Bearer ') {
-            $this->setAuthenticationStatus(TokenInterface::NO_CREDENTIALS_GIVEN);
-            throw new AuthenticationRequiredException('Could not extract access token from Authorization header: "Bearer" keyword is missing', 1616222749);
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Security/Authentication/Token/BearerToken.php
```diff
@@ -50,14 +50,14 @@ public function updateCredentials(ActionRequest $actionRequest)
...
 
-        $authorizationHeader = $httpRequest->getHeader('Authorization')[0];
+        $this->setAuthenticationStatus(TokenInterface::NO_CREDENTIALS_GIVEN);
 
-        if (substr($authorizationHeader, 0, 7) !== 'Bearer ') {
-            $this->setAuthenticationStatus(TokenInterface::NO_CREDENTIALS_GIVEN);
-            throw new AuthenticationRequiredException('Could not extract access token from Authorization header: "Bearer" keyword is missing', 1616222749);
+        foreach ($httpRequest->getHeader('Authorization') as $authorizationHeader) {
+            if (strpos($authorizationHeader, 'Bearer ') === 0) {
+                $this->credentials['bearer'] = substr($authorizationHeader, strlen('Bearer '));
```
albe, line 57: Don't we need to `setAuthenticationStatus(TokenInterface::AUTHENTICATION_NEEDED)` here?

albe (approved):

PR #2496, opened 2021-06-14
TASK: Allow installing Doctrine 2.9

albe:
As of doctrine/orm 2.9.3 it is again compatible with Flow (see #2495), so we can allow installing it (again).


kdambekalns (approved):
💯

PR #2499, opened 2021-06-29
BUGFIX: Avoid bool return value in restoreFlashMessageContainerFromSession()

kdambekalns:
It can happen, that `getData(…)` returns a boolean, leading to an error
due to the return type declaration.


Neos.Flow/Classes/Mvc/FlashMessage/Storage/FlashMessageSessionStorage.php
```diff
@@ -83,7 +83,10 @@ private function restoreFlashMessageContainerFromSession(): ?FlashMessageContain
         }
         /** @var FlashMessageContainer $flashMessageContainer */
```
albe, line 84: Hm, that probably prevented psalm from making us aware of that type mismatch?

albe (approved):

daniellienert (approved):

PR #2500, opened 2021-06-30
BUGFIX: Ensure cache backends are prepared before usage

theilm:
If the flushByTag or findIdentifiersByTag methods of the TaggableMultiBackend are used before backend initialization by other methods, the backends have to be prepared. Otherwise, `$this->backends` is an empty array and no cache entries are flushed.

**What I did**
I added the `$this->prepareBackends()` calls in the two methods.

**How to verify it**
- Configure the TaggableMultiBackend for the Neos_Fusion_Content cache
- Change a node property in the Neos backend
- Reload the page

Before this change, the change of the node property was saved to the db, but the cache was not flushed. Thus, the incorrect property value was shown in the Neos backend after a page reload.


albe (approved):
Oh yeah, lazy initialization combined with inheritance :heart: :) Thanks a lot for finding and fixing this!

Issue #2473, opened 2021-05-02
UriConstraints should make sure the Uri path starts with a slash


albe:
See https://github.com/guzzle/psr7/blob/1.4.1/src/Uri.php#L704

albe:
Related to #2501 and #2502

arsors:
I also have a problem with "withPathPrefix()" in my package.

Since psr7 requires the path to start with a "/" [this query](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php#L261) should be exactly the opposite. 
Currently it checks if there is a "/" at the beginning and then gives an error. In my opinion it should rather check if there is no "/" at the beginning and then give an error.

In the core "withPathPrefix()" is currently only used in one place. I just don't know what effect it has on [this line](https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Mvc/Routing/Router.php#L200).

albe:
Linking in @bwaidelich since he is the most involved in that part of the code base. From my limited PoV that sounds reasonable, though I'm unsure what subtle consequences would come from turning around that method's expectations.

Also note, that we already enforce a `/` prefix on the final URI in https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php#L396
Plus we have this check here, that might also need to be changed to be consistent: https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Mvc/Routing/Dto/ResolveContext.php#L74 (and make `ResolveContext::getUriPathPrefix()` work with `UriConstraint::withPathPrefix()` in the Router as mentioned above).
Most notably, probably this here needs to change to accomodate joining prefixes with a leading `/` and a trailing `/` correctly: https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php#L258

Also should we still allow `''` as uriPathPrefix (for b/c) and translate it to `'/'`? IMO would make sense, but maybe there's an edge case where this would hide a subtle bug (e.g. a stringification of something that should not be empty).

bwaidelich:
Sorry for the late reaction, I lost track of this one.
IIRC we expect paths _not_ to start with `/` _internally_ for historic reasons (i.e. the routePath will always be without leading slash).
[...]

PR #2501, opened 2021-07-01
TASK: Allow guzzlehttp/psr7 2.0

albe:
This allows installing guzzlehttp/psr7 2.0


albe:
```
+Uncaught Exception in Flow The path of a URI with an authority must start with a slash "/" or be empty\n
+thrown in file Packages/Libraries/guzzlehttp/psr7/src/Uri.php\n
+in line 729\n
+\n
+45 GuzzleHttp\Psr7\Uri::validateState()\n
+44 GuzzleHttp\Psr7\Uri::withPath("test/mvc/uribuilder/target/my-path")\n
```
Seems there is another thing we need to adjust to correctly work with guzzle/psr7 2.0

kdambekalns (approved):

albe:
The open failing tests leaves the question if we should maybe also remove the ~2.0 in 7.1 (as in #2502) and keep guzzle/psr7 2.0 support for master in a dedicated PR

albe:
Note: I upmerged the changes in #2502, so unless we merge this change (or if we decide to remove the ~2.0 from this again), the next release will no longer allow installing guzzle/psr7 2.0

nlx-sascha:
FYI: guzzle/psr7 2.0 breaks neos/flow in version ^6.3 maybe also lower versions are affected, but didn't test them.
In my opinion there are 2 options. Replace the 9 deprecated functions class in flow core or disallow guzzle/psr7 2.0

albe:
> FYI: guzzle/psr7 2.0 breaks neos/flow in version ^6.3 maybe also lower versions are affected

Only 6.3+ as the "allow ~2.0" was added in 6.3 originally. It is already removed again in the branches, but needs a release to take effect. See https://github.com/neos/flow-development-collection/pull/2502

I'd like to have a decision on this before doing the releases though - should we keep 2.0 support in 7.1+ and fix it, or remove the support?

albe:
As discussed briefly with @daniellienert in slack, since getting 2.0 support in involves some more changes as the failing tests suggest, it's better to get all versions stable again and add this in 7.2 instead. I'll retarget this PR and then we should tag a release ASAP.

albe:
Note for the upcoming release manager: Need to decide if we want to keep guzzle 1.x support or just directly upgrade to ^2.0 as in #2508

kdambekalns:
To help me understand this:

- the lower branches are working again, not allowing `guzzle/psr7` 2.0?
- this PR allows `guzzle/psr7` 2.0 but it's still broken?

If making it work for both branches of `guzzle/psr7` is easy (enough), that option get's my vote.

albe:
> To help me understand this:
> 
> * the lower branches are working again, not allowing `guzzle/psr7` 2.0?

Correct.

> * this PR allows `guzzle/psr7` 2.0 but it's still broken?

Correct. See failing tests, but as far as I can see the primary issue to fix is #2473

albe:
If we could fix the error in https://github.com/neos/flow-development-collection/issues/2473 then we could merge this into the same branch as the fix. So b/c of this solely depends on the b/c of the fix for #2473

kdambekalns:
> If we could fix the error in #2473 then we could merge this into the same branch as the fix.

https://github.com/neos/flow-development-collection/pull/2813 fixes #2473 and targets 8.0 – so this could get into 8.0 as well, if the RMs agree…

albe:
@mficzel wdyt about

> Note for the upcoming release manager: Need to decide if we want to keep guzzle 1.x support or just directly upgrade to ^2.0 as in https://github.com/neos/flow-development-collection/pull/2508

mficzel:
If tests are green i am for doing it.

albe:
> > If we could fix the error in #2473 then we could merge this into the same branch as the fix.
> 
> #2813 fixes #2473 and targets 8.0 – so this could get into 8.0 as well, if the RMs agree…

Apparently #2813 doesn't suffice - or moreover doesn't really fix the issue. An URI in guzzle may indeed be fully empty, but if it's not empty it's path MUST start with a `/` - see https://github.com/guzzle/psr7/blob/2.1/src/Uri.php#L729-L730
However, the real culprit is indeed not the instanciation of the (empty) URIs, but when we call `withPath()` with a string not starting with a `/`, which fails to happen at e.g. https://github.com/neos/flow-development-collection/blob/8.0/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php#L383-L385 or https://github.com/neos/flow-development-collection/blob/8.0/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php#L367

So #2813 could even be reverted since it does not do anything of effect really.

bwaidelich:
> So https://github.com/neos/flow-development-collection/pull/2813 could even be reverted since it does not do anything of effect really.

+1

kdambekalns:
> So #2813 could even be reverted since it does not do anything of effect really.

See https://github.com/neos/flow-development-collection/pull/2814

kitsunet:
Revert merged, so what do we do with this? Do we want to try to fix the places that need fixing for 8.0? Sounds sensible to make guzzle 2.0 possible.

kdambekalns:
I'm giving it a try…

bwaidelich:
@kdambekalns did you see #2818 ?

mficzel (approved):
I like this and now tests are also green

bwaidelich (changes requested):
> No. Was it linked somewhere?

yes: #2473 – but I forgot to add it here, sorry

I suggest to merge #2818 (not because it's mine but because it separates the bugfix from the update and because it covers one potential case with `pathSuffix`)

albe:
> Afterwards this could be changed to a FEATURE IMO, only adjusting the composer dependencies

`FEATURE` is the benevolent interpretation of a `TASK` upgrading dependencies with b/c for old versions ;)

bwaidelich (approved):

PR #2502, opened 2021-07-01
TASK: Disallow installing guzzlehttp/psr7 2.0

albe:
It is incompatible with versions < 1.7 due to the replaced `stream_for` method. The ~2.0 dependency was added before the actual 2.0 release and this breaking change was added later, making it incompatible. If 2.0+ is needed, you need to upgrade to Flow 7.1

### DIFF
--- a/Neos.Http.Factories/composer.json
+++ b/Neos.Http.Factories/composer.json
@@ -9,7 +9,7 @@
     "require": {
         "php": "^7.1",
         "psr/http-factory": "^1.0",
-        "guzzlehttp/psr7": "^1.4, !=1.8.0 || ~2.0"
+        "guzzlehttp/psr7": "^1.4, !=1.8.0"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "neos/composer-plugin": "^2.0",
         "composer/composer": "^1.9 || ^2.0",
         "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
-        "guzzlehttp/psr7": "^1.4, !=1.8.0 || ~2.0",
+        "guzzlehttp/psr7": "^1.4, !=1.8.0",
         "ext-mbstring": "*"
     },
     "replace": {


albe:
See also #2501 which is the counterpart for Flow 7.1+

kdambekalns (approved):
I wonder why we allowed 2.0 if it wasn't even released… 🤔

sorenmalling (approved):

albe:
> I wonder why we allowed 2.0 if it wasn't even released… 🤔

Let's say I was a bit over-ambitious based on an early pre-release version that had no such breaking changes in it and was like "wooh, that's an easy upgrade to the new version"? :|

For reference the "corpus delicti" - https://github.com/neos/flow-development-collection/commit/200c8ec35f1082e4340ba7c2f4bc757eb67fe0c0#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34

hlubek:
This should be also fixed for the base distribution, it is currently broken after a creating a new project.

albe:
But the distribution does not require guzzle itself? Are you referring to a `composer create-project neos/flow-base-distribution ...`?

hlubek:
Well, it was `composer create-project neos/neos-base-distribution Neos-7.1` actually and I found a Stack Overflow discussion pointing to this issue. But yeah, the Neos distribution is broken per default since `neos/http-factories` and `neos/fusion-form` allow version 2.0 of `guzzlehttp/psr7`.

albe:
That should be resolved by latest versions of those packages too, though? https://packagist.org/packages/neos/http-factories#7.1.2
Or maybe composer just tries to install an earlier version in order to allow guzzle 2.0? Not sure 100% how the resolving strategy in composer handles that case. In that case, not sure how to proceed :| untag all "broken" versions? That would be quite a few. Or update the base distribution to specify allowing only fixed versions of flow and in turn http-factories/fusion-form (i.e. "~7.1.3" rather than e.g. "~7.1.0")?

PR #2503, opened 2021-07-01
Added all new types introduced since PHP7.1 to not being qualified

### LINKED ISSUE #2498: "iterable" gets qualified during proxy building
Since PHP 7.1 `iterable` is a new scalar pseudo-type, which is not allowed to be "qualified".

But having a method returning an `iterable` (and corresponding type-hint) will lead to a proxy method, annotated with `\iterable`, which will lead to a PHP Fatal error: `Type declaration 'iterable' must be unqualified in ...`

This could be solved in `\Neos\Utility\TypeHandling::isSimpleType`, or in `\Neos\Flow\Reflection\ReflectionService::reflectClassMethod` having an array of words, which should not be qualified `!in_array($returnType, ['self', 'null', 'callable', 'void'])`.

As I am unsure, where it really belongs: Ideas?

Flow: 5.3

fcool:
Since PHP 7.1 iterable, object and mixed are new allowed PHP types, which are not allowed to be "qualified".

But having a method returning one of those (and having a corresponding type-hint) did lead to a proxy method, annotated with \type, which will lead to a PHP Fatal error: Type declaration 'type' must be unqualified in ...

This resolves #2498

List of types taken from here:
https://www.php.net/manual/de/language.types.declarations.php

(and yes, this commit includes one type, which is part of PHP8.0 which is not (yet?) supported by Flow 6.3, but I guess it will not harm anyone and definitely makes live easier in upwards versions.

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1381,7 +1381,7 @@ class ReflectionService
         }
 
         $returnType = $method->getDeclaredReturnType();
-        if ($returnType !== null && !in_array($returnType, ['self', 'null', 'callable', 'void']) && !TypeHandling::isSimpleType($returnType)) {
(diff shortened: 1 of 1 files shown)


albe (approved):
Thanks!

kdambekalns (approved):

PR #2507, opened 2021-07-05
TASK: Run post-update scripts only once 

daniellienert:
Previously the scripts were run for every package at the end of the install process.
The scripts only need to be run once.

### DIFF
--- a/Neos.Flow/Classes/Composer/InstallerScripts.php
+++ b/Neos.Flow/Classes/Composer/InstallerScripts.php
@@ -1,4 +1,6 @@
 <?php
+declare(strict_types=1);
+
 namespace Neos\Flow\Composer;
 
 /*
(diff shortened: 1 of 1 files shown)


daniellienert:
See https://github.com/neos/flow-development-collection/pull/2507/commits/d79947442d7de8c59f0d6460e0affef30406148b for the pure code change without code cleanup.

albe:
😍 this was soooo long on my "this can't be right, need to take a look sometime" list!

Neos.Flow/Classes/Composer/InstallerScripts.php
```diff
@@ -56,17 +77,25 @@ public static function postUpdateAndInstall(Event $event)
...
      * @param PackageEvent $event
      * @return void
+     * @throws Exception\InvalidConfigurationException
      * @throws Exception\UnexpectedOperationException
+     * @throws \Neos\Utility\Exception\FilesException
      */
-    public static function postPackageUpdateAndInstall(PackageEvent $event)
+    public static function postPackageUpdateAndInstall(PackageEvent $event): void
     {
+        if (self::$postPackageUpdateAndInstallAlreadyRun) {
```
albe, line 95: Just one question, which I'm not sure about: Is this script run once after each installed/updated package, or once for each package after they are all installed/updated?

daniellienert: The script is triggered when all packages are installed updated for every of that packages

albe (approved):

fcool (approved):

Issue #2510, opened 2021-07-08
Resource:copy command fails for empty collections


albe:
When using `resource:copy` as described here https://github.com/flownative/flow-aws-s3#switching-the-storage-of-a-collection the copying fails with a `The target collection "tmpNewCollection" is not empty.` error.

The cause is the `!empty()` check here: https://github.com/neos/flow-development-collection/blob/7.0/Neos.Flow/Classes/Command/ResourceCommandController.php#L164-L167 because the `getObjects()` method returns a generator, which will always return `false`:
https://3v4l.org/LELMJ#v7.4.21

A good attempt would be to make `Collection` have a `isEmpty()` method, which could then be answered by iterating the storage, which again could be optimized for in the storage implementation, e.g. by returning an `EmptyIterator` that the Collection can check for, but if unchecked behaves as expected when iterating.
https://www.php.net/manual/de/class.emptyiterator.php

At minimum, the `!empty()` needs to be removed.

albe:
Has been solved by using `Generator::valid()`, which is a good option over a custom `Collection::isEmpty()` method.

PR #2512, opened 2021-07-09
BUGFIX: Remove empty check on target collection in resource:copy

paavo:
`$targetCollection->getObjects()` method returns a generator, which will always return `false`
This makes it impossible to use `resource:copy` as this always fails with a `The target collection "tmpNewCollection" is not empty.` error.

The problem is mentioned here: https://github.com/neos/flow-development-collection/issues/2510

**What I did**
Removed !empty()

**How to verify it**
Use `resource:copy` to copy assets to an empty Storage.

**Checklist**
- ~~[ ] Tests have been created, run and adjusted as needed~~ (not needed)


albe:
Hm, should target 6.3 but the base branch contains commits from 7.0 - @paavo can you rebase onto the 6.3 branch and force-push or create a new PR?

paavo:
> Hm, should target 6.3 but the base branch contains commits from 7.0 - @paavo can you rebase onto the 6.3 branch and force-push or create a new PR?

I created a new PR https://github.com/neos/flow-development-collection/pull/2513
Hope this makes a quick release easier.

PR #2513, opened 2021-07-15
BUGFIX: Change empty check on target collection to `valid()` in resource:copy

paavo:
`$targetCollection->getObjects()` method returns a generator, which will always return `false` in an `empty()` check.
This makes it impossible to use `resource:copy` as this always fails with a `The target collection "tmpNewCollection" is not empty.` error.

The problem is mentioned here: https://github.com/neos/flow-development-collection/issues/2510

**What I did**
Change `!empty()` against a `->valid()` check

**How to verify it**
Use `resource:copy` to copy assets to an empty Storage.

**Checklist**
- ~~[ ] Tests have been created, run and adjusted as needed~~ (not needed)

This replaced PR https://github.com/neos/flow-development-collection/pull/2512

### DIFF
--- a/Neos.Flow/Classes/Command/ResourceCommandController.php
+++ b/Neos.Flow/Classes/Command/ResourceCommandController.php
@@ -161,7 +161,7 @@ class ResourceCommandController extends CommandController
             $this->quit(1);
         }
 
-        if (!empty($targetCollection->getObjects())) {
+        if (!$targetCollection->getObjects()) {
             $this->outputLine('The target collection "%s" is not empty.', [$targetCollectionName]);
             $this->quit(1);
         }
--- a/Neos.Flow/Classes/Command/ResourceCommandController.php
+++ b/Neos.Flow/Classes/Command/ResourceCommandController.php
@@ -161,7 +161,7 @@ class ResourceCommandController extends CommandController
             $this->quit(1);
         }
 
-        if (!$targetCollection->getObjects()) {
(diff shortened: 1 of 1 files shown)


albe (approved):

Neos.Flow/Classes/Command/ResourceCommandController.php
```diff
@@ -161,7 +161,7 @@ public function copyCommand(string $sourceCollection, string $targetCollection,
             $this->quit(1);
         }
 
-        if (!empty($targetCollection->getObjects())) {
+        if (!$targetCollection->getObjects()) {
             $this->outputLine('The target collection "%s" is not empty.', [$targetCollectionName]);
             $this->quit(1);
         }
```
albe, line 167: Just checked again how generators behave with a simple truthy check:
https://3v4l.org/sk4P8#v7.4.21

So this whole check is effectively `if (false) {` now, which doesn't make much sense either. Looking again at the docs and checking behaviour of `Generator::valid()` this seems to be the correct way:
```suggestion
        if ($targetCollection->getObjects()->valid()) {
            $this->outputLine('The target collection "%s" is not empty.', [$targetCollectionName]);
            $this->quit(1);
        }
```
https://3v4l.org/UGZ2I#v7.4.21

albe:
See https://github.com/neos/flow-development-collection/pull/2513#discussion_r671659221 as for the reasoning why `valid()` is the right way

PR #2514, opened 2021-07-15
TASK: Fix documentation for firewall option "rejectAll"

dlubitz:
The ``rejectAll`` option needs to be set as boolean. 
See:  https://github.com/neos/flow-development-collection/blob/6.3/Neos.Flow/Classes/Security/Authorization/FilterFirewall.php#L53


daniellienert (approved):

albe (approved):

albe:
Thanks!

Issue #2483, opened 2021-05-26
neos/cache 5.3.25...5.3.26 Cannot add or modify cache entry because the affected keys are being modified by another process ("Flow_Session_Storage")


booooza:
### Description

Since neos/cache upgrade from 5.3.25 to 5.3.26 we (randomly?) got many of the following exception:

```
Neos\Cache\Exception: Cannot add or modify cache entry because the affected keys are being modified by another process ("Flow_Session_Storage") (1594725688)

#0 Packages/Libraries/neos/cache/Classes/Frontend/VariableFrontend.php(68): Neos\Cache\Backend\RedisBackend->set('...', '...', Array, 0)
#1 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Session_Session.php(479): Neos\Cache\Frontend\VariableFrontend->set('...', Array, Array, 0)
#2 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Session_Session.php(751): Neos\Flow\Session\Session_Original->putData('Neos_Flow_Secur...', Array)
#3 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Session_Session.php(689): Neos\Flow\Session\Session_Original->storeAuthenticatedAccountsInfo(Array)
#4 Packages/Framework/Neos.Flow/Classes/ObjectManagement/ObjectManager.php(561): Neos\Flow\Session\Session_Original->shutdownObject()
#5 Packages/Framework/Neos.Flow/Classes/ObjectManagement/ObjectManager.php(464): Neos\Flow\ObjectManagement\ObjectManager->callShutdownMethods(Object(SplObjectStorage))
#6 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Security_Context.php(219): Neos\Flow\ObjectManagement\ObjectManager->Neos\Flow\ObjectManagement\{closure}()
#7 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Security_Context.php(219): Closure->__invoke()
#8 Packages/Framework/Neos.Flow/Classes/ObjectManagement/ObjectManager.php(465): Neos\Flow\Security\Context_Original->withoutAuthorizationChecks(Object(Closure))
#9 [internal function]: Neos\Flow\ObjectManagement\ObjectManager->shutdown('Runtime', 'Neos\\Flow\\Core\\...')
[...]

PR #2516, opened 2021-07-17
TASK: Revert #2052 - Add TTL to tags in RedisBackend

albe:
This resolves the issues with the redis backend that came up after the changes in #2052, which unfortunately are worse than the benefits. We'd still like to get the TTL for tags in, but likely need to refrain from having them consistent/in sync.

See https://github.com/neos/flow-development-collection/issues/2483


albe:
FYI @kdambekalns @kitsunet - will merge as of the consensus of the unicorn weekly two weeks ago once the build passes. Will then do upmerges and tag releases latest tomorrow so bugfixes are out on monday.

PR #2517, opened 2021-07-17
BUGFIX: Do not remove related entities during persistence

albe:
Expected behavior by serialize is not to change the object to serialize but to provide
a textual representation. If your object contained persisted entities in a multivalued
property it had been silently removed.

Rebased follow-up to #2012


Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -131,7 +131,6 @@ private function Flow_searchForEntitiesAndStoreIdentifierArray($path, $propertyV
                 'identifier' => $identifier,
                 'entityPath' => $path
             ];
-            $this->$originalPropertyName = Arrays::setValueByPath($this->$originalPropertyName, $path, null);
```
albe, line 134: See https://github.com/neos/flow-development-collection/pull/2012/files#r429575763

kdambekalns: I _think_ the unset was in place to make sure persisting worked. The actual entities need to be encoded, but keeping them means they _get serialized as well_, doubling data and potentially having bad side-effects. I can imagine this even dates back to our own persistence layer before Doctrine 2… So, it might not be needed anymore.

That being said, the whole concept of `Flow_searchForEntitiesAndStoreIdentifierArray` might be useless by now?!

fcool: Not entirely, no. At least if you use Flowpack.JobQueue and the @Defer - Annotation, you will see, that just passing a Doctrine collection will lead to detached entities in the Job, while a DTO with a collection property will not. So it is still "of general use"

fcool (approved):

sorenmalling (approved):
Tested on a setup, where entities was removed, when added to a cache. This solves the issue!

albe:
So shall we merge this and see if it backfires? Can't see how and all tests are green, but still...

fcool:
I would go for it.

kitsunet:
Hold on a second, can someone enlighten me on the code path you are considering for this, because `during persistence`. Where does this happen during persistence just to make sure.

fcool:
Its "session persistence" unlucky chosen headline to be honest. Stupid author ;)

kitsunet:
Ah, I see, mmm but that happens somewhere at the end of the request doesn't it? Why is it a problem if the entity is gone then because I am still somewhat convinced we would otherwise serialize it and that might cause problems?
But if that's not the case, fine by me :)

fcool:
The problem is that the session persistence works using serializiation. But if you save some entities to some kind of cache, this uses the serialization too. 

And because serialization changes the state of the object to serialize, it is not repeatable. And as the session serialization usually comes last, it cannot serialize again, because the properties has been emptied.

And Sorens problem is, if I understand him correctly even more nasty. He saves some entity in a cache and after that updates it using the usual persistence. But as serialization changed the entity, it saves something which never was intended to be changed or worse, saved.

fcool:
And to finally address your concern:

You are absolutely right. Different from before the original property will now get serialized, too. 
But it will get overwritten during __wakeup, as there still the content from "Flow_Persistence_RelatedEntities" is used during reconstiution - so yeah - we serialize more than needed, this way.

The only "right" solution to this problem would be moving forward to new Serialization logic introduced with PHP 7.4 using `__serialize` and `__unserialize` - which is something for me on my wishlist - but a bit bigger and would change the minimal required interpret version.

albe:
So if @kitsunet says he has no more worries I will hit the button :)

kitsunet:
Sounds good, lets try this 😆

sorenmalling:
@albe @fcool Can we consider #2011 solved, with this bugfix merged into 6.3.14 and up

bwaidelich:
FYI: Not verified yet, but this seems to have introduced a major regression: https://github.com/neos/neos-development-collection/issues/3539

kdambekalns:
It seems this has some side-effects – should we revert this? Suggestions welcome…

sorenmalling:
> It seems this has some side-effects – should we revert this? Suggestions welcome…

Can we be concrete about why these side effects happens - as in, can we be sure that this change is the reason, and why? This knowledge should be used for improving the next version of this important change

bwaidelich:
I can verify that the fix by @dlubitz (https://github.com/neos/neos-development-collection/pull/3540) solves the symptoms for us.
I'm not into the persistence details enough to judge how sever the original issues were and if we can solve them otherwise. In general I would be in favor of reverting this one for now since I'm sure it has more side effects than the ones observed in the Media Browser

albe:
Agree to revert this change for now, there indeed seem to be more side effects: https://neos-project.slack.com/archives/C050C8FEK/p1637922313414100

PR #2519, opened 2021-07-19
FEATURE: CLI parse array type controller arguments

gjwnc:
**What I did**

With this PR, a CLI command can make use of array's as an argument type.

If an argument is given multiple times and the type is an array, the argument values are pushed at the end of the array.

As an example, this can be handy, if you need to process multiple language dimensions, nodes or urls, like
```
./flow my.pckg:index --dimension en --dimension de
```

Maybe this can also be considered for the 7.2 release?

**How I did it**

The CLI `RequestBuilder` now distinguishes between array and non-array arguments. If an argument is of type `array` the argumentValue is appended to the array.

**How to verify it**

Create a command controller command with an array signature like
```
public function indexCommand(array $dimension){
```
and then use the argument `--dimension` multiple times.

**Checklist**


kitsunet (approved):
Seems like a nice addition. I would probably look at making the code a bit nicer but I know this place isn't beautiful anyways, so that can happen as part of a bigger refactoring I would say.

albe (approved):
I think the code is pretty clear in this change. A refactoring wouldn't hurt in the bigger scheme of this class/method but definitely not in this PR :)

PR #2520, opened 2021-07-19
BUGFIX: Allow null for source in translation helper

jonnitto:
This fixes a bug that was introduced with PR #2476

Before it was possible to set the filename to `null`:

```
I18n.translate('foo', null, [], null, 'Foo.Bar')
```

but without this fix, you have to set the source to `Main`

```
I18n.translate('foo', null, [], 'Main', 'Foo.Bar')
```

### DIFF
--- a/Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
+++ b/Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
@@ -100,7 +100,7 @@ class TranslationHelper implements ProtectedContextAwareInterface
      * @param string $id Id to use for finding translation (trans-unit id in XLIFF)
      * @param string $originalLabel The original translation value (the untranslated source string).
      * @param array $arguments Numerically indexed array of values to be inserted into placeholders
-     * @param string $source Name of file with translations
+     * @param string|null $source Name of file with translations, defaults to 'Main'
      * @param string $package Target package key. If not set, the current package key will be used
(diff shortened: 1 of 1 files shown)


albe:
Please let's also adjust the argument docblocks to contain `|null` and mention the default value. Otherwise agree

albe (approved):

Neos.Flow/Classes/I18n/EelHelper/TranslationHelper.php
```diff
@@ -109,10 +109,11 @@ public function allowsCallOfMethod($methodName)
     protected function translateByExplicitlyPassedOrderedArguments($id, $originalLabel = null, array $arguments = [], $source = 'Main', $package = null, $quantity = null, $locale = null)
     {
         $translationParameterToken = $this->createTranslationParameterToken($id);
+        $source = isset($source) ? str_replace('.', '/', $source) : 'Main';
```
grebaldi, line 112:
```suggestion
        $source = $source === null ? 'Main' : str_replace('.', '/', $source);
```

Just I tiny nitpick: I'd go four using a null-reference check instead of `isset` here, because imho it's more readable. (`isset` has some tricky semantics. Those are completely irrelevant here, but whenever I see it, I have to think about the implications, while `=== null` tells the story immediatey).

grebaldi (approved):

Issue #2521, opened 2021-07-22
Authorization checks are re-enabled when exceptions are caught


bwaidelich:
### Description

Especially in CLI mode authorization checks should be disabled.
When code is wrapped in a `Security\Context::withoutAuthorizationChecks()` call and exceptions are caught, auth checks are no longer disabled in the outer closure leading to exceptions like
```
The security Context cannot be initialized yet
```

### Steps to Reproduce

```php
// this is the outer closure (e.g. triggered by CommandRequestHandler::handleRequest())
$this->securityContext->withoutAuthorizationChecks(function () {

  // some try/catch that swallows the exception, somewhere in the stack trace
  try {

    // some user code that should skip auth checks
    $this->securityContext->withoutAuthorizationChecks(function () {
      // ....and might lead to an exception
      throw new \Exception('Some exception');
    });
  } catch (\Exception $exception) {
  }

  // the bug (this should not be the case):
  $this->securityContext->areAuthorizationChecksDisabled() === false;
});
```

### Affected Versions

Flow: all (lowest supported branch: 6.3)

### Note:

Another reason not to silently swallow exceptions

sorenmalling:
@bwaidelich Not automatically closed after merge of #2522  - can I close this, or does it need anything more?

bwaidelich:
@sorenmalling yes, thanks for the heads up, I'll close the issue

PR #2522, opened 2021-07-22
BUGFIX: Keep authorization checks disabled if an exception is swallowed

bwaidelich:
When code is wrapped in a `Security\Context::withoutAuthorizationChecks()` call
and exceptions are caught, authorization checks are no longer disabled in the outer
closure leading to exceptions like
```
The security Context cannot be initialized yet
```

This change fixes this by resetting the `authorizationChecksDisabled` to the
previous value in a `finally` block.

Fixes: #2521


daniellienert (approved):
Looks good by reading

fcool (approved):

bwaidelich:
@skurfuerst gave a +1 via chat, I'll merge this with admin

PR #2524, opened 2021-07-23
TASK: Enable PHP nightly build

albe:
This enables the experimental php-nightly build. Right now this failing will make the whole build to fail, which is not intended.


sorenmalling:
@albe So we can enable this, without it blocking the possibility to merge changes? If yes, then I'm all for this to see the results

albe:
Unfortunately not - because of a missing feature on GH actions (https://github.com/actions/runner/issues/2347). I do like how https://github.com/ruby-concurrency/concurrent-ruby/pull/922/files#diff-d6f479696321bb36f55c90230909dad3bc2ed01da5ab7277192d6542c36fd256 solved it: they have a separate experimental build workflow that only runs on schedule and not per PR/Merge. That way PRs are not blocked, no confusing red dot in the checks and we still have feedback on PHP-next support.

albe:
So this now adds a new workflow that is scheduled to run every night at midnight and run builds agains PHP-nightly. A fail in those runs will not block PRs and not clobber the normal workflow, but if we are curious how our current codebase behaves against latest PHP version, we can just take a look at the workflow (https://github.com/neos/flow-development-collection/actions/workflows/experimental.yml - available once this is merged)

kdambekalns (approved):

.github/workflows/experimental.yml
```diff
@@ -0,0 +1,204 @@
+name: build
```
kdambekalns, line 1:
```suggestion
name: Experimental PHP test
```

PR #2310, opened 2020-12-02
TASK: Use PhpUnit 9

kdambekalns:
See https://github.com/neos/flow-development-collection/issues/2293

PR #2525, opened 2021-08-02
TASK: Raise minimal phpunit to 9.1

daniellienert:
Version 9.1 introduces the method `assertFileDoesNotExist()`
which is used since Flow 7.0 (https://github.com/neos/flow-development-collection/pull/2310)


daniellienert:
Mhm it still installs PHPUnit 9.0 ...

albe:
Needs to be added to the package composer.jsons as well

kdambekalns:
Let's see…

kdambekalns:
Oh well, that seems to be an unrelated new problem now… _sigh_

composer.json
```diff
@@ -32,7 +32,9 @@
         "composer/composer": "^1.9 || ^2.0",
         "egulias/email-validator": "^2.1 || ^3.0",
         "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
-        "guzzlehttp/psr7": "^1.4.1, !=1.8.0",
+        "guzzlehttp/psr7": "^1.7, !=1.8.0",
+        "neos/fusion-afx": "*",
```
albe, line 36: That comes from the kickstarter, where it should only be an optional dependency (but that has not been tackled yet). That new requirement leads to the Fusion tests being run which yield the test failure.

> 1) Neos\Fusion\Tests\Unit\Service\HtmlAugmenterTest::invalidAttributesTests with data set #0 ('', array(array()), null, null, '<root>array value ignored</root>')
PHPUnit\Framework\Exception: Class Neos\Neos\Exception does not exist

So seems like Fusion still depends a bit too much on the neos core - so either we fix that, or we remove the dependency to fusion-X from here

composer.json
```diff
@@ -32,7 +32,9 @@
...
+        "neos/fusion-form": "*",
```
albe, line 37:
```suggestion
```

composer.json
kdambekalns: - Imagine the collection manifest would be updated automagically–then such a change would be impossible
- If the kickstarter requires it, it must be installed, simply removing the dependency sounds foolish

But may I ask: Why does the kickstarter have that dependency? I can't seem to find it using Fusion…

albe: > Why does the kickstarter have that dependency? I can't seem to find it using Fusion…

You can generate Fusion templates with it if you want to, which would mean you'd have to have Fusion installed. So that dependency should rather be a "suggests"?

kdambekalns: No, in that case the generated package must have a dependency on Fusion. Right?

albe: That is... correct. I guess I was thinking in terms of "the new package should then directly work without another composer update" or s.th. like that

albe:
Thinking about it, since the guzzle functions were changed in 7.1 (https://github.com/neos/flow-development-collection/pull/2455/commits/37aafc67f3272720ba4e132105ffc7211839290f), shouldn't this land in 7.1?

Also https://github.com/neos/flow-development-collection/pull/2501 should be changed to "Allow guzzle/psr7 2.0" and not replicate this minimum version raising

composer.json
kdambekalns: Probably. But with the use of `DistributionPackages` an update is needed / done after kickstarting a package anyway, no?

kdambekalns:
> Thinking about it, since the guzzle functions were changed in 7.1 ([37aafc6](https://github.com/neos/flow-development-collection/commit/37aafc67f3272720ba4e132105ffc7211839290f)), shouldn't this land in 7.1?

Good point. But what about PhpUnit 9.1: Why is that needed and can that work with 7.1?

> Also #2501 should be changed to "Allow guzzle/psr7 2.0" and not replicate this minimum version raising

I changed the PR title :)

composer.json
albe: Right - I was not thinking this fully through I guess and just sticked to how the Fluid dependency was handled, though that is indeed a direct dependency of the kickstarter.

We'd need to adjust https://github.com/neos/flow-development-collection/blob/master/Neos.Kickstarter/Classes/Command/KickstartCommandController.php#L119 to include the dependency for fusion-afx/form (and also fluid btw) depending on if generate-fusion is selected.

albe:
> But what about PhpUnit 9.1: Why is that needed...

I'm not really sure... Daniel tried to solve an issue with `assertFileDoesNotExist()` - which was introduced by you :) https://github.com/neos/flow-development-collection/commit/4710d0203d53db6760dc3638c5116b7c46e10022

kdambekalns:
Ok, I probably didn't realize the need for PhpUnit 9.1, since it will usually be installed anyway with the dependency on ^9.0. Unless you test for "lowest deps"… 🙈

Yeah, agree on the split idea!

composer.json
albe: Let's revert this and merge #2526, then that part is bandaided at least

composer.json
```diff
@@ -32,7 +32,7 @@
         "composer/composer": "^1.9 || ^2.0",
         "egulias/email-validator": "^2.1 || ^3.0",
         "typo3fluid/fluid": "~2.5.11 || ^2.6.10",
-        "guzzlehttp/psr7": "^1.4.1, !=1.8.0",
+        "guzzlehttp/psr7": "^1.7, !=1.8.0",
```
albe, line 35:
```suggestion
        "guzzlehttp/psr7": "^1.4.1, !=1.8.0",
```

Neos.Http.Factories/composer.json
```diff
@@ -9,7 +9,7 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "psr/http-factory": "^1.0",
-        "guzzlehttp/psr7": "^1.4.1, !=1.8.0"
+        "guzzlehttp/psr7": "^1.7, !=1.8.0"
```
albe, line 12:
```suggestion
        "guzzlehttp/psr7": "^1.4.1, !=1.8.0"
```

albe:
Okay, so this now only resolves the PHPUnit version mismatch, which should not be critical

kdambekalns (changes requested):
This definitely is too many changes that seem completely unrelated to a PhpUnit dependency update…

albe:
Rebased and pushed to new branch with a new PR at #2530

PR #2526, opened 2021-08-10
TASK: Remove fusion dependencies

albe:
Only the created package has a dependency on Fusion, iff it uses generated Fusion templates

This needs to be followed up with a change that will add those dependencies to the created packages composer manifest via some way.


albe:
See also https://github.com/neos/flow-development-collection/pull/2525#discussion_r685915196

sorenmalling (approved):

PR #2530, opened 2021-08-15
TASK: Raise minimal phpunit to 9.1 

albe:
Version 9.1 introduces the method `assertFileDoesNotExist()`
which is used since Flow 7.0 (#2310)

### DIFF
--- a/Neos.Cache/composer.json
+++ b/Neos.Cache/composer.json
@@ -14,7 +14,7 @@
     },
     "require-dev": {
         "mikey179/vfsstream": "^1.6.1",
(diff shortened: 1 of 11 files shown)


albe:
Rebased replacement for #2525

albe:
> Installing phpunit/phpunit (9.0.0): Extracting archive

Hmmmm...

kdambekalns (approved):

kdambekalns:
I guess the dependency is just never seen. `require-dev` is root-manifest-only, so the dependency is taken from the distribution.

PR #2533, opened 2021-08-18
FEATURE: Preserve class and method doc blocks

albe:
This preserves the docblocks when generating proxies in order to not accidentially generate invalid annotations that break the classes.

See also #2387


kdambekalns (approved):

kdambekalns:
Funny, I could have sworn we specifically did not take the unchanged docblock over to the proxy. But I could not find any trace of anything backing that feeling, so… 🤷‍♂️

Less code is always good!

fcool (approved):

PR #2541, opened 2021-08-24
TASK: Fix Psalm issues

kdambekalns:


kdambekalns:
This leads to a weird situation: We accept a view parameter, that we no longer use. I fixed that, but I am unusre if putting that here is a great idea. I'll create a separate PR for 6.0 once this is merged, unless we think it should be fixed here right away.

kdambekalns:
The failing PhpUnit test is a mystery… it passes for me.

fcool (approved):
:+1:

PR #2545, opened 2021-08-24
BUGFIX: NullBackend ignores configured properties

gjwnc:
**What I did**

The `NullBackend` just return's `true` in `setProperty()` for all configured properties.

**What does this solve**

I'm using split sources Objects.local.yaml to disable caches in my local setup where I only change the `backend:` configuration to `Neos\Cache\Backend\NullBackend`.

The problem I have is, that the Objects.yaml merge strategy also includes the `backendOptions` and thus by just using
```
Neos_Fusion_Content:
  backend: Neos\Cache\Backend\NullBackend
```
in my `Objects.local.yaml` I get an Exception, thrown by AbstractBackend, that the `NullBackend` doesn't accept the backendOption `hostname` - from my default redis configuration.

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/NullBackend.php
+++ b/Neos.Cache/Classes/Backend/NullBackend.php
@@ -25,6 +25,18 @@ use Neos\Cache\Backend\TaggableBackendInterface;
 class NullBackend extends AbstractCacheBackend implements PhpCapableBackendInterface, TaggableBackendInterface
 {
     /**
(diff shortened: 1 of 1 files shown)


kdambekalns:
Strictly speaking this could be solved by adjusting your configuration, and since Neos is in security-fix-only mode for anything below 5.3, this should not be merged. But… 🤷‍♂️

Makes sense, though!

PR #2546, opened 2021-08-24
BUGFIX: Bring back RenderingContext view argument, pin Fluid version

kdambekalns:
With `typo3fluid/fluid` 2.7.0 the `$view` constructor argument to
`RenderingContext` is gone, but we support lower versions, thus things
break.

This brings back the argument in our code and pins Fluid to < 2.7.0
for Flow below 7.2.0.

See https://github.com/neos/flow-development-collection/issues/2541
See https://github.com/TYPO3/Fluid/pull/548


kdambekalns:
So, this is the "usual" 7.2 mystery test failure… The error this fixes is exposed by a "lowest deps" test, which runs only on 6.3+

robertlemke (approved):
Makes sense and I can't see how this would break.

Issue #2548, opened 2021-08-26
Exception logfile request information formatting


DrillSergeant:
Currently the request headers are rendered into the exception file just by concatenating the values. It's hard to read for humans and also hard to write a script to parse them.
I am missing the header names and also a seperator between the values.
**Is there a reason for this?** Otherwise I would come open a PR

Example (current state):
```
HTTP REQUEST:
_gcl_au=1.1.1316419741.1604049318; nQ_cookieId=8e8fbb8f-206d-5bc9-7915-c3b387e7d803; siga-user-context={%22suggestedCountry%22:%22global%22%2C%22suggestedLanguage%22:%22de%22}; _hjTLDTest=1; _hjid=0e5fcf3e-fb91-445e-8392-dc9cd9ec96f2; _fbp=fb.1.1604053281566.1843020996; phpMyAdmin=daaeca84a26317bc33e35c9266f8f1bb; _ga=GA1.2.2041892705.1608027526; Neos_Session=0zSWBYAysDggmG14AgHPqESH7lMInfRC; _hjIncludedInPageviewSample=1; _hjAbsoluteSessionInProgress=1; _hjIncludedInSessionSample=1de,en;q=0.9,fr;q=0.8,af;q=0.7,it;q=0.6,es;q=0.5,nb;q=0.4,ca;q=0.3,mt;q=0.2,nl;q=0.1,pl;q=0.1gzip, deflate, brhttps://siga.ddev.site/global_en/contentimageno-corssame-originimage/avif,image/webp,image/apng,image/*,*/*;q=0.8Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.66 Safari/537.36443onhttps172.18.0.1172.18.0.1closesiga.ddev.site
```

Example (what would be more readable by adding the headernames and put each header in a new line):
```
HTTP REQUEST:
[...]

PR #2552, opened 2021-08-30
TASK: improve request information in exception logfiles

kdambekalns:
According to https://github.com/neos/flow-development-collection/issues/2548
this is my suggestion for better request information in the exception
dump files.

The goal is better readability of the request information for humans
and the chance to parse them with scripts:

- Add the name of the request headers
- Separate each header (name: value) into an own line
- Add the request target (requested url)


kdambekalns:
This replaces #2550 and targets the 6.3 branch…

albe (approved):
Re-approve :) Thanks for the retarget!

sorenmalling (approved):

PR #2555, opened 2021-09-03
BUGFIX: Prevent proxies for classes with unsupported interfaces

bwaidelich:
Adjusts the proxy building for classes such that it throws a more
helpful exception when trying to generate proxies for classes
implementing an interface with constructor and parameters.

Fixes: #2554

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyConstructor.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyConstructor.php
@@ -11,6 +11,10 @@ namespace Neos\Flow\ObjectManagement\Proxy;
  * source code.
  */
 
+use Neos\Flow\ObjectManagement\Exception\CannotBuildObjectException;
+use Neos\Flow\Reflection\ClassReflection;
+use Neos\Flow\Reflection\Exception\ClassLoadingForReflectionFailedException;
+
 /**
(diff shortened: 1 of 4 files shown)


Neos.Flow/Classes/ObjectManagement/Proxy/ProxyConstructor.php
```diff
@@ -49,6 +54,11 @@ public function render()
 
         $code = '';
         if ($this->addedPreParentCallCode !== '' || $this->addedPostParentCallCode !== '') {
+            foreach ((new ClassReflection($this->fullOriginalClassName))->getInterfaceNames() as $interfaceName) {
+                if ($this->reflectionService->getMethodParameters($interfaceName, $this->methodName) !== []) {
+                    throw new CannotBuildObjectException(sprintf('Cannot build proxy for class "%s" because it implements interface "%s" that defines a constructor with arguments. Adjust the interface or avoid the proxy class from being created (for example by adjusting the Method privilege in your Policy)', $this->fullOriginalClassName, $interfaceName), 1630656424);
```
bwaidelich, line 59: I wasn't really creative with this one, feel free to improve the exception

kdambekalns (approved):
Looks good, thanks!

albe (approved):
Best solution as long as we can't just make it work.

PR #2559, opened 2021-09-09
BUGFIX: Handle required array arguments and exceeding arguments properly together

mficzel:
When a cli argument of type array was used it was not removed from the list of required arguments and thus
later when unnamed arguments where checked ended up as still missing.

This is solved by having an explicit list of required argumentNames instead of removing the configuration of a
required argument after it was set.

**How to verify it**

Create a cli controller that has array arguments and pass exceeding arguments.

**Checklist**


mficzel:
Will create new pr from my personal repo

PR #2562, opened 2021-09-09
BUGFIX: Handle required and exceeding arguments properly in combination with cli array arguments 

mficzel:
When a cli argument of type array was used it was not removed from the list of required arguments and thus
later when unnamed arguments where checked was regarded as still missing. That way exceeding unnamed arguments could not be used together with required array arguments.

This is solved by having an explicit list of required argumentNames instead of removing the configuration of a
required argument after it was set.

This fixes an issue with the array feature that was introduced with #2519

**Checklist**


kdambekalns (approved):
Looks good by reading…

albe (commented):
Makes sense and looks good by reading.

albe (approved):

PR #2567, opened 2021-09-14
BUGFIX: Correctly set up TestingProvider for "testable security"

### LINKED ISSUE #2386: Static instantiation breaks TestingProvider being a singleton as promised.
### Description

Static instantiation prohibits `TestingProvider` being a singleton as promised.

Since https://github.com/neos/flow-development-collection/pull/1213 a `create` factory method is part of `AuthenticationProviderInterface`. And it is used e.g. in `AuthenticationProviderManager` and `TokenAndProviderFactory`.

Now there is a `TestingProvider` annotated as a singleton, for use in functional tests using security – but singleton and `new static($name, $options)` don't play well together.

### Affected Versions

Flow: 5.0.0 and up

kdambekalns:
As of https://github.com/neos/flow-development-collection/pull/1213
the authentication providers are created using a factory method, this
breaks the assumption the `TestingProvider` is a singleton.

The result is that e.g. ´authenticateRoles([])` no longer works as
expected in functional tests.

This fixes the issue by fetching the provider from the factory instead
of the object manager.

Fixes #2386

### DIFF
--- a/Neos.Flow/Classes/Security/Authentication/Provider/TestingProvider.php
+++ b/Neos.Flow/Classes/Security/Authentication/Provider/TestingProvider.php
@@ -11,7 +11,6 @@ namespace Neos\Flow\Security\Authentication\Provider;
  * source code.
  */
 
-use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Security\Account;
 use Neos\Flow\Security\Authentication\Token\TestingToken;
 use Neos\Flow\Security\Authentication\TokenInterface;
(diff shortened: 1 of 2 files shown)


fcool (approved):
And I just found out, that I knew that already - and solved it only for me by overwriting the setup method with exact that fix. With this patch, it works without it. Did not realize that this is a general problem, not only for me. 

Did test it in the project which uses it, now it works without overriding the original testingProvider, which feels way better.

albe (approved):

PR #2568, opened 2021-09-14
!!! BUGFIX: Avoid broken proxy docblocks

### LINKED ISSUE #2564: Carrying over docblocks breaks (some) proxy classes
### Description

With PR #2533 docblocks are copied from the original class to the proxy class. This works fine for "standard code", but breaks under certain circumstances. One example is the `ImportedAsset` domain model.

### Steps to Reproduce

1. Check out Neos 7.2.x-dev as of *now*
2. Try to run `./flow`

#### Expected behavior

Output showing you some hints on how to use Flow.

#### Actual behavior

An error message is shown:

`[Semantical Error] The annotation "@Entity" in class Neos\Media\Domain\Model\ImportedAsset was never imported.`

### Affected Versions

Flow: 7.2.x-dev

### What happens

Compare this from after the change:

```php
<?php 
namespace Neos\Media\Domain\Model;

/*
 * This file is part of the Neos.Media package.
 *
 * (c) Contributors of the Neos Project - www.neos.io
  *
 * This package is Open Source Software. For the full copyright and license
 * information, please view the LICENSE file which was distributed with this
 * source code.
 */

use Neos\Flow\Annotations\Entity;
use Neos\Flow\Annotations\Identity;

/**
 * @Entity
 */
class ImportedAsset_Original
{
// original code here
}

#
# Start of Flow generated Proxy code
#
namespace Neos\Media\Domain\Model;

/**
 * @Entity
 * @codeCoverageIgnore
 */
class ImportedAsset extends ImportedAsset_Original implements \Neos\Flow\ObjectManagement\Proxy\ProxyInterface, \Neos\Flow\Persistence\Aspect\PersistenceMagicInterface {
// proxy code
}
```

with the code before the change:

```php
<?php
namespace Neos\Media\Domain\Model;

/*
 * This file is part of the Neos.Media package.
 *
 * (c) Contributors of the Neos Project - www.neos.io
  *
 * This package is Open Source Software. For the full copyright and license
 * information, please view the LICENSE file which was distributed with this
 * source code.
 */

use Neos\Flow\Annotations\Entity;
use Neos\Flow\Annotations\Identity;

/**
 * @Entity
 */
class ImportedAsset_Original
{
// original code
}

#
# Start of Flow generated Proxy code
#
namespace Neos\Media\Domain\Model;

/**
 *
 * @\Neos\Flow\Annotations\Entity
 */
class ImportedAsset extends ImportedAsset_Original implements \Neos\Flow\ObjectManagement\Proxy\ProxyInterface, \Neos\Flow\Persistence\Aspect\PersistenceMagicInterface {
// proxy coe
}
```

The previous way of rendering the annotations resulted in fully qualified annotation class names (in this case `@\Neos\Flow\Annotations\Entity`), thus the imports did not matter. Now the docblock is copied "as is" but the imports are not carried over.

kdambekalns:
With PR #2533 docblocks are copied from the original class to the proxy class. This breaks when using annotations without the "standard" imports `Flow` and `ORM`. One example is the ImportedAsset domain model.

This fixes that by some changes to the proxy building.

Note, if you use property introduction via AOP, those properties must from now on use fully-qualified classnames for annoatations on them!

Fixes #2564

### DIFF
--- a/Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
@@ -57,11 +57,6 @@ class ProxyClassBuilder
     protected $pointcutExpressionParser;
 
     /**
-     * @var ProxyClassBuilder
-     */
-    protected $proxyClassBuilder;
-
-    /**
      * @var VariableFrontend
      */
     protected $objectConfigurationCache;
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -243,7 +243,7 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
         $classCode = $this->stripOpeningPhpTag($classCode);
 
         $classNameSuffix = self::ORIGINAL_CLASSNAME_SUFFIX;
-        $classCode = preg_replace_callback('/^([a-z\s]*?)(final\s+)?(interface|class)\s+([a-zA-Z0-9_]+)/m', function ($matches) use ($pathAndFilename, $classNameSuffix, $proxyClassCode) {
+        $classCode = preg_replace_callback('/^([a-z\s]*?)(final\s+)?(interface|class)\s+([a-zA-Z0-9_]+)/m', static function ($matches) use ($pathAndFilename, $classNameSuffix, $proxyClassCode) {
             $classNameAccordingToFileName = basename($pathAndFilename, '.php');
             if ($matches[4] !== $classNameAccordingToFileName) {
                 throw new Exception('The name of the class "' . $matches[4] . '" is not the same as the filename which is "' . basename($pathAndFilename) . '". Path: ' . $pathAndFilename, 1398356897);
@@ -253,7 +253,7 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
 
         // comment out "final" keyword, if the method is final and if it is advised (= part of the $proxyClassCode)
         // Note: Method name regex according to http://php.net/manual/en/language.oop5.basic.php
-        $classCode = preg_replace_callback('/^(\s*)((public|protected)\s+)?final(\s+(public|protected))?(\s+function\s+)([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+\s*\()/m', function ($matches) use ($pathAndFilename, $classNameSuffix, $proxyClassCode) {
+        $classCode = preg_replace_callback('/^(\s*)((public|protected)\s+)?final(\s+(public|protected))?(\s+function\s+)([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+\s*\()/m', static function ($matches) use ($proxyClassCode) {
             // the method is not advised => don't remove the final keyword
             if (strpos($proxyClassCode, $matches[0]) === false) {
                 return $matches[0];
@@ -263,7 +263,7 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
 
         $classCode = preg_replace('/\\?>[\n\s\r]*$/', '', $classCode);
 
-        $proxyClassCode .= "\n" . '# PathAndFilename: ' . $pathAndFilename;
+        $proxyClassCode .= PHP_EOL . '# PathAndFilename: ' . $pathAndFilename;
 
         $separator =
             PHP_EOL . '#' .
@@ -329,7 +329,7 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
             $optionValueAsString = '';
             if (is_object($optionValue)) {
                 $optionValueAsString = self::renderAnnotation($optionValue);
-            } elseif (is_scalar($optionValue) && is_string($optionValue)) {
+            } elseif (is_string($optionValue)) {
                 $optionValueAsString = '"' . $optionValue . '"';
             } elseif (is_bool($optionValue)) {
                 $optionValueAsString = $optionValue ? 'true' : 'false';
@@ -370,7 +370,7 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
                 $value .= self::renderAnnotation($v);
             } elseif (is_array($v)) {
                 $value .= self::renderOptionArrayValueAsString($v);
-            } elseif (is_scalar($v) && is_string($v)) {
+            } elseif (is_string($v)) {
                 $value .= '"' . $v . '"';
             } elseif (is_bool($v)) {
                 $value .= $v ? 'true' : 'false';
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
@@ -229,11 +229,7 @@ class ProxyClass
         if ($methodsCode . $constantsCode === '') {
             return '';
         }
-        $classCode = ($namespace !== '' ? 'namespace ' . $namespace . ";\n\n" : '') .
-            "use Doctrine\\ORM\\Mapping as ORM;\n" .
-            "use Neos\\Flow\\Annotations as Flow;\n" .
-            "\n" .
-            $this->buildClassDocumentation() .
+        $classCode = $this->buildClassDocumentation() .
             $classModifier . 'class ' . $proxyClassName . ' extends ' . $originalClassName . ' implements ' . implode(', ', array_unique($this->interfaces)) . " {\n\n" .
             $traitsCode .
             $constantsCode .
--- a/Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
+++ b/Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
@@ -11,7 +11,6 @@ namespace Neos\Flow\Persistence\Aspect;
  * source code.
  */
 
-use Doctrine\ORM\Mapping as ORM;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Aop\JoinPointInterface;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
--- a/Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
+++ b/Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
@@ -70,8 +70,8 @@ class PersistenceMagicAspect
 
     /**
      * @var string
-     * @ORM\Id
-     * @ORM\Column(length=40)
+     * @Doctrine\ORM\Mapping\Id
+     * @Doctrine\ORM\Mapping\Column(length=40)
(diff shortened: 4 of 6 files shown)


Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -56,11 +56,6 @@ class ProxyClassBuilder
      */
     protected $pointcutExpressionParser;
 
-    /**
-     * @var ProxyClassBuilder
-     */
-    protected $proxyClassBuilder;
```
kdambekalns, line 62: This was simply unused… and supposed to be "itself". 🤷‍♂️

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -243,7 +243,7 @@ protected function cacheOriginalClassFileAndProxyCode($className, $pathAndFilena
         $classCode = $this->stripOpeningPhpTag($classCode);
 
         $classNameSuffix = self::ORIGINAL_CLASSNAME_SUFFIX;
-        $classCode = preg_replace_callback('/^([a-z\s]*?)(final\s+)?(interface|class)\s+([a-zA-Z0-9_]+)/m', function ($matches) use ($pathAndFilename, $classNameSuffix, $proxyClassCode) {
+        $classCode = preg_replace_callback('/^([a-z\s]*?)(final\s+)?(interface|class)\s+([a-zA-Z0-9_]+)/m', static function ($matches) use ($pathAndFilename, $classNameSuffix, $proxyClassCode) {
```
kdambekalns, line 246: Cleanup…

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -253,7 +253,7 @@ protected function cacheOriginalClassFileAndProxyCode($className, $pathAndFilena
 
         // comment out "final" keyword, if the method is final and if it is advised (= part of the $proxyClassCode)
         // Note: Method name regex according to http://php.net/manual/en/language.oop5.basic.php
-        $classCode = preg_replace_callback('/^(\s*)((public|protected)\s+)?final(\s+(public|protected))?(\s+function\s+)([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+\s*\()/m', function ($matches) use ($pathAndFilename, $classNameSuffix, $proxyClassCode) {
+        $classCode = preg_replace_callback('/^(\s*)((public|protected)\s+)?final(\s+(public|protected))?(\s+function\s+)([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+\s*\()/m', static function ($matches) use ($proxyClassCode) {
```
kdambekalns, line 256: Cleanup…

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -263,7 +263,7 @@ protected function cacheOriginalClassFileAndProxyCode($className, $pathAndFilena
 
         $classCode = preg_replace('/\\?>[\n\s\r]*$/', '', $classCode);
 
-        $proxyClassCode .= "\n" . '# PathAndFilename: ' . $pathAndFilename;
+        $proxyClassCode .= PHP_EOL . '# PathAndFilename: ' . $pathAndFilename;
```
kdambekalns, line 266: Cleanup…

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -329,7 +329,7 @@ public static function renderAnnotation($annotation)
             $optionValueAsString = '';
             if (is_object($optionValue)) {
                 $optionValueAsString = self::renderAnnotation($optionValue);
-            } elseif (is_scalar($optionValue) && is_string($optionValue)) {
+            } elseif (is_string($optionValue)) {
```
kdambekalns, line 332: Cleanup… a string is a scalar

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -370,7 +370,7 @@ protected static function renderOptionArrayValueAsString(array $optionValue)
                 $value .= self::renderAnnotation($v);
             } elseif (is_array($v)) {
                 $value .= self::renderOptionArrayValueAsString($v);
-            } elseif (is_scalar($v) && is_string($v)) {
+            } elseif (is_string($v)) {
```
kdambekalns, line 373: Cleanup… a string is a scalar

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
```diff
@@ -229,11 +229,7 @@ public function render()
         if ($methodsCode . $constantsCode === '') {
             return '';
         }
-        $classCode = ($namespace !== '' ? 'namespace ' . $namespace . ";\n\n" : '') .
-            "use Doctrine\\ORM\\Mapping as ORM;\n" .
-            "use Neos\\Flow\\Annotations as Flow;\n" .
-            "\n" .
-            $this->buildClassDocumentation() .
+        $classCode = $this->buildClassDocumentation() .
```
kdambekalns, line 232: That's the actual fix…

robertlemke (approved):
Looks good to me.

> Note, if you use property introduction via AOP, those properties must from now on use fully-qualified classnames for annoatations on them!

What happens if you don't use a fully-qualified class name? If you'll get an error during compile time and can make some sense of it to get it right, that would be fine.

kdambekalns:
> Looks good to me.
> 
> > Note, if you use property introduction via AOP, those properties must from now on use fully-qualified classnames for annoatations on them!
> 
> What happens if you don't use a fully-qualified class name? If you'll get an error during compile time and can make some sense of it to get it right, that would be fine.

You get an error like this one:

`Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@ORM\Id" in property Neos\Flow\Tests\Functional\Property\Fixtures\TestValueobject::$Persistence_Object_Identifier was never imported. Did you maybe forget to add a "use" statement for this annotation?`

That lacks the information about where that property actually comes from (as it's not really in `TestValueObject`), but the message allowed me to find the cause. YMMV.

Do you think that's sufficient?

robertlemke:
> Do you think that's sufficient?

Could be better, but since it's a Doctrine exception, there's not much we can do, I guess?

kdambekalns:
> > Do you think that's sufficient?
> 
> Could be better, but since it's a Doctrine exception, there's not much we can do, I guess?

Nope. We could invest a yet unknown amount of time into detecting and/or fixing this during compile time, but… 🤷‍♂️

Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
```diff
@@ -70,8 +70,8 @@ public function isEntityOrValueObject()
 
     /**
      * @var string
-     * @ORM\Id
-     * @ORM\Column(length=40)
+     * @Doctrine\ORM\Mapping\Id
```
albe, line 73: Curious, but doesn't the "manual" import in lines 14+15 suffice to have the short named annotation?

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/AspectOrientedProgramming.rst
```diff
@@ -1033,23 +1034,30 @@ The following example introduces a new property "subtitle" to the class
...
-	 * An aspect for demonstrating property introductions
-	 *
-	 * @Flow\Aspect
-	 */
-	class PropertyIntroductionAspect {
+    use Neos\Flow\Annotations as Flow;
 
-		/**
-		 * @var string
-		 * @Column(length=40)
```
albe, line 1047: Guess that never really worked anyway

albe (approved):
Generally approve, though I'm having a bit of a hard time with the "fully qualified namespace annotation required" break. Why is that needed if the original class imports the namespace? Is it really not possible to use a shortened namespace in the annotations?

Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
kdambekalns: No, because that docblock (of the aspect class) is not copied into the class the *property* is introduced into…

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/AspectOrientedProgramming.rst
kdambekalns: True. But even with `@ORM\Column` it won't work any longer…

Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
albe: Ah, right. Could we possibly monkey-patch https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Aop/PropertyIntroduction.php#L83 for a b/c fix to this? I guess it's still hard to resolve the string class name without parsing the `use` statements of the Aspect class... maybe at least resolve `@ORM\` and `@Flow\` there to somewhat lower breakyness?

kdambekalns: It did, because we did not copy the docblock, but re-rendered any annotations, making them fully qualified. As can be seen in the example with the `Entity` annotation in #2564.

kdambekalns: About the breakiness… TBH I have no clue how much property introduction is actually used and if those few places cannot be fixed faster, than it would take us to come up with a fix that doesn't come with additional (potential) bugs…

kdambekalns: https://github.com/search?l=PHP&p=8&q=%27%40Flow%5CIntroduce%27&ref=advsearch&type=Code is almost exclusively our (test) code.

albe: Okay - still, wdyt about making Flow and ORM work by hardcoding the FQCN resolution for those? Whenever you want to inject a property, it's most likely a DB field or some Flow injected property.

robertlemke: That would reduce the pain. But for the record, in random projects I randomly don't use the general `Flow\Annotations` import, but rather import `@Inject()` (for example) directly, because Phpstorm let's me do that very easily. These cases would still break.

robertlemke: > That would reduce the pain. But for the record, in random projects I randomly don't use the general `Flow\Annotations` import, but rather import `@Inject()` (for example) directly, because Phpstorm let's me do that very easily. These cases would still break.

uh, would _not_ break, right? Because I imported it and I'm not using the FQCN.

kdambekalns: **Would break**, because your import in the aspect class' docblock is never copied to the class you introduce a property to. This is _only about annotations on introduced properties_…

We can of course expand `@ORM\` and `@Flow`, feel free to improve this (I won't have time today)…

mficzel (approved):
Cannot say much about the code but can confirm that the issues are solved.

Issue #2569, opened 2021-09-15
XliffReader returns wrong offset value


MarcoPNS:
### Description

I have a Main.xlf with 3 file blocks. 2 blocks are fine. When I add a third one, Neos throw following exception:

````
Fatal error: Uncaught TypeError: Argument 2 passed to Neos\Utility\Arrays::arrayMergeRecursiveOverrule() must be of the type array, null given, called in /Users/marco/Sites/xlifftest/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/Neos_Flow_I18n_Xliff_Service_XliffFileProvider.php on line 164 and defined in /Users/marco/Sites/xlifftest/Packages/Libraries/neos/utility-arrays/Classes/Arrays.php on line 71
````

My Main.xlf looks a bit like this:

````
<?xml version="1.0"?>
<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
    <file original="" source-language="en" datatype="plaintext">
        <body>
            <trans-unit id="first">
                <source>First</source>
            </trans-unit>
        </body>
    </file>
    <file original="" source-language="en" datatype="plaintext">
        <body>
            <trans-unit id="second">
                <source>Second</source>
            </trans-unit>
        </body>
    </file>
    <file original="" source-language="en" datatype="plaintext">
        <body>
            <trans-unit id="third">
                <source>Third</source>
            </trans-unit>
        </body>
    </file>
</xliff>
````

I digged a little deeper into that problem and found out that the XliffReader->readFiles returns a wrong offset for the third block. Instead of `$offset = 2`, it returns `$offset = 3`. 

### Steps to Reproduce

1. Have a translation file with 3 file blocks
2. Try to output the third value.

#### Expected behavior

Translation get shown correctly

#### Actual behavior

````
Fatal error: Uncaught TypeError: Argument 2 passed to Neos\Utility\Arrays::arrayMergeRecursiveOverrule() must be of the type array, null given, called in /Users/marco/Sites/xlifftest/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/Neos_Flow_I18n_Xliff_Service_XliffFileProvider.php on line 164 and defined in /Users/marco/Sites/xlifftest/Packages/Libraries/neos/utility-arrays/Classes/Arrays.php on line 71
````

### My stupid solution that somehow work

I guess this is a really unsafe way to do it (that's why I only open a issue and not a PR) but currently that works with an endless amount of file blocks.

[...]

PR #2570, opened 2021-09-17
fixed $offset in XliffReader #2569

MarcoPNS:
**What I did**
Fixed problem with the $offset value described in #2569

**How I did it**
$offset must be incremented in the while loop before the itterator. The first element is already handled before the loop. 

**How to verify it**
Create a .xlf File with at least 3 file blocks. 

Tested with the latest Flow Version. 

4.2 is the first version where I could find the problem.


sorenmalling:
Hi @MarcoPNS 

Super finding and thanks for the PR! I've read the specification, and wasn't aware of the format allowing more than one <file> tags :)

Will you be able to write a test covering that this is now possible? I would go and look at XliffFileProviderTest.php or XliffReaderTest.php for inspiration

PR #2582, opened 2021-09-30
BUGFIX: Do not pass signal information to persistAll

kdambekalns:
With PR #2448 the `shouldCheck()` was added, having a return type
declaration of `bool`. The value returned is set only by the argument
given to `persistAll()`, so having Behat tests fail when the value was
a string seemed strange.

The actual value was even stranger:

    Neos\\Flow\\Cli\\SlaveRequestHandler::dispatchedCommandLineSlaveRequest

Reason: The `Package` class in `Neos.Flow` connects a signal to the
`persistAll()` method, and the "signal information" is passed as an
argument…

This change fixes that.

### DIFF
--- a/Neos.Flow/Classes/Package.php
+++ b/Neos.Flow/Classes/Package.php
@@ -71,7 +71,7 @@ class Package extends BasePackage
                 }
             }
         });
-        $dispatcher->connect(Cli\SlaveRequestHandler::class, 'dispatchedCommandLineSlaveRequest', Persistence\PersistenceManagerInterface::class, 'persistAll');
+        $dispatcher->connect(Cli\SlaveRequestHandler::class, 'dispatchedCommandLineSlaveRequest', Persistence\PersistenceManagerInterface::class, 'persistAll', false);
 
         if (!$context->isProduction()) {
             $dispatcher->connect(Core\Booting\Sequence::class, 'afterInvokeStep', function (Step $step) use ($bootstrap, $dispatcher) {


albe (approved):

kitsunet (approved):

robertlemke (approved):

kdambekalns:
Jobs are all past the MySQL Behat step already… 🎉

Neos.Flow/Classes/Package.php
```diff
@@ -71,7 +71,7 @@ public function boot(Core\Bootstrap $bootstrap)
                 }
             }
         });
-        $dispatcher->connect(Cli\SlaveRequestHandler::class, 'dispatchedCommandLineSlaveRequest', Persistence\PersistenceManagerInterface::class, 'persistAll');
+        $dispatcher->connect(Cli\SlaveRequestHandler::class, 'dispatchedCommandLineSlaveRequest', Persistence\PersistenceManagerInterface::class, 'persistAll', false);
```
albe, line 74:
```suggestion
        $dispatcher->connect(Cli\SlaveRequestHandler::class, 'dispatchedCommandLineSlaveRequest', function () use ($bootstrap) {
            $persistenceManager = $bootstrap->getObjectManager()->get(Persistence\PersistenceManagerInterface::class);
            $persistenceManager->persistAll(true);
        });
```
Alternative that keeps the `persistAll(trueish)` semantic from before. Not sure that is wanted?

kitsunet: I think I am fine with this for now

kdambekalns: No, I'm pretty sure this never was intended. And since a slave request does not have any notion of "safe requests" the expectation is to have things persist, IMHO.

PR #2584, opened 2021-09-30
!!!TASK: Add method argument and return type hints

albe:
Adds PHP typehints to some places.


sorenmalling:
@albe Can we use a tool to find places where we are missing type hints and make this a large commit of type hinting adding?

albe:
could use psalm or intellij, but it's a big undertaking to add typehints everywhere. Would like to add a few more places here though, yes.

jonnitto (approved):

PR #2585, opened 2021-09-30
!!! TASK: Add `persistAllowedObjects()` deprecate flag argument for `persistAll()` in PersistenceManager

albe:
If you ever used `persistAll(true)` you should replace those invocations with `persistAllowedObjects()`.

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
@@ -84,7 +84,7 @@ class PersistenceManager extends AbstractPersistenceManager
      * Commits new objects and changes to objects in the current persistence
      * session into the backend
      *
-     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject()). Deprecated: Use `persistOnlyAllowedObjects()` instead.
+     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject()). Deprecated: Use `persistAllowedObjects()` instead.
      * @return void
      * @throws PersistenceException
      * @api
@@ -92,7 +92,7 @@ class PersistenceManager extends AbstractPersistenceManager
     public function persistAll(bool $onlyAllowedObjects = false): void
     {
         if ($onlyAllowedObjects === true) {
-            $this->persistOnlyAllowedObjects();
+            $this->persistAllowedObjects();
             return;
         }
         if (!$this->entityManager->isOpen()) {
@@ -115,7 +115,7 @@ class PersistenceManager extends AbstractPersistenceManager
      * @throws PersistenceException
      * @api
      */
-    public function persistOnlyAllowedObjects(): void
+    public function persistAllowedObjects(): void
     {
         if (!$this->entityManager->isOpen()) {
             $this->logger->error('persistAll() skipped flushing data, the Doctrine EntityManager is closed. Check the logs for error message.', LogEnvironment::fromMethodName(__METHOD__));
--- a/Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
@@ -84,19 +84,44 @@ class PersistenceManager extends AbstractPersistenceManager
      * Commits new objects and changes to objects in the current persistence
      * session into the backend
      *
-     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject())
+     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject()). Deprecated: Use `persistOnlyAllowedObjects()` instead.
      * @return void
      * @throws PersistenceException
      * @api
      */
     public function persistAll(bool $onlyAllowedObjects = false): void
     {
+        if ($onlyAllowedObjects === true) {
+            return $this->persistOnlyAllowedObjects();
+        }
         if (!$this->entityManager->isOpen()) {
             $this->logger->error('persistAll() skipped flushing data, the Doctrine EntityManager is closed. Check the logs for error message.', LogEnvironment::fromMethodName(__METHOD__));
             return;
         }
 
-        $this->allowedObjects->checkNext($onlyAllowedObjects);
+        $this->allowedObjects->checkNext(false);
+        $this->entityManager->flush();
+        $this->emitAllObjectsPersisted();
+    }
+
+    /**
+     * Commits new objects and changes to objects in the current persistence
+     * session into the backend.
+     * An exception will be thrown if there are scheduled updates/deletes or
+     * insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject())
+     *
+     * @return void
+     * @throws PersistenceException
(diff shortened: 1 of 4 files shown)


albe:
Open question is, if we should add this method to the interface in Flow 8?

sorenmalling:
A comment to the making the method a part of the API: 
As a developer the term "allowed objects" is confusing, if I'm not fully aware of the underlying ORM used (in our case Doctrine). 

With that said, and that we are doing a good job of telling about "safe request method" and the consequences ([Link to documentation](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Persistence.html#safe-request-methods-are-read-only)), what issue are we tackling with bring a new method to the API?

albe:
> if I'm not fully aware of the underlying ORM used (in our case Doctrine).

In this case, it is even a term of our own and not a doctrine concept. It's just that it should only persist objects that have explicitly been allowed with `persistenceManager->allowObject($x)` beforehand. The core use-case is avoiding unnecessary persistence in safe requests.

> what issue are we tackling with bring a new method to the API?

Primarily to get rid of the error prone behaviour flag on the `persistAll()` method. It is a bad pattern and what should `persistAll(false)` really mean? We even had a very [subtle bug](https://github.com/neos/flow-development-collection/pull/2582) due to that, because we put the method as slot into the signal for `dispatchedCommandLineSlaveRequest` and forgot to disable sending signal information. This lead to a call of `persistAll('Neos\\Flow\\Cli\\SlaveRequestHandler::dispatchedCommandLineSlaveRequest')`, which in turn only worked because the argument wasn't strictly typed. But with my refactoring of the AllowedObjectsContainer, this blew up when I added typehints there. See this long slack thread for all the dirty details: https://neos-project.slack.com/archives/C04PYL8H3/p1632933052013300

That said, this does not mean we need to add the method to the API/interface of the PersistenceManager. It could be a primarily internal detail for our HTTP Stack. Not sure anyone would want to manually call `persistOnlyAllowedObjects()`?

sorenmalling:
> That said, this does not mean we need to add the method to the API/interface of the PersistenceManager. It could be a primarily internal detail for our HTTP Stack. Not sure anyone would want to manually call `persistOnlyAllowedObjects()`?

Good valid point - it will make great sense in our HTTP Stack 👍

Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
```diff
@@ -84,19 +84,45 @@ public function injectLogger(LoggerInterface $logger): void
...
      * @throws PersistenceException
      * @api
      */
     public function persistAll(bool $onlyAllowedObjects = false): void
+    {
+        if ($onlyAllowedObjects === true) {
+            $this->persistOnlyAllowedObjects();
+            return;
+        }
+        if (!$this->entityManager->isOpen()) {
```
sorenmalling, line 98: Should this "isOpen" check go into a separate method, to avoud duplicated code

sorenmalling (changes requested):

Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
albe: Not sure it makes sense here, since this is not duplicated anywhere and the log message is specific to the `persistAll()` call

albe:
@sorenmalling could you reconsider your veto?

mficzel:
I am for adding this still

Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
```diff
@@ -84,19 +84,45 @@ public function injectLogger(LoggerInterface $logger): void
...
+     * Commits new objects and changes to objects in the current persistence
+     * session into the backend.
+     * An exception will be thrown if there are scheduled updates/deletes or
+     * insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject())
+     *
+     * @return void
+     * @throws PersistenceException
+     * @api
+     */
+    public function persistOnlyAllowedObjects(): void
```
mficzel, line 118: Wording wise i would prefer `persistAllowedObjects` or `persistAllAllowedObjects`

mficzel (commented):
This makes sense but the calls to persistAll with a boolean argument should be removed by the same pr. I understand that the adjustment of the php chain to actually use the new method will happen in a later pr.

Also i wonder wether the new method should be public api.

kdambekalns (commented):
I like it.

- Naming should be adjusted as suggested by @mficzel – I'd vote for `persistAllowedObjects()`
- The adjustments to Flow / Neos to use that new method should be done timely, but that is not a blocker for this PR, IMHO

Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
```diff
@@ -84,19 +84,45 @@ public function injectLogger(LoggerInterface $logger): void
...
-     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject())
+     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject()). Deprecated: Use `persistOnlyAllowedObjects()` instead.
      * @return void
      * @throws PersistenceException
      * @api
      */
     public function persistAll(bool $onlyAllowedObjects = false): void
+    {
+        if ($onlyAllowedObjects === true) {
+            $this->persistOnlyAllowedObjects();
```
albe, line 95:
```suggestion
            $this->persistAllowedObjects();
```

Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
```diff
@@ -84,19 +84,45 @@ public function injectLogger(LoggerInterface $logger): void
      * Commits new objects and changes to objects in the current persistence
      * session into the backend
      *
-     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject())
+     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject()). Deprecated: Use `persistOnlyAllowedObjects()` instead.
```
albe, line 87:
```suggestion
     * @param boolean $onlyAllowedObjects If true an exception will be thrown if there are scheduled updates/deletes or insertions for objects that are not "allowed" (see AbstractPersistenceManager::allowObject()). Deprecated: Use `persistAllowedObjects()` instead.
```

Neos.Flow/Classes/Persistence/Doctrine/PersistenceManager.php
albe, line 118:
```suggestion
    public function persistAllowedObjects(): void
```

Neos.Flow/Classes/Persistence/PersistenceManagerInterface.php
```diff
@@ -162,8 +162,8 @@ public function remove($object): void;
     public function update($object): void;
 
     /**
-     * Adds the given object to a list of allowed objects which may be persisted when persistAll() is called with the
-     * $onlyAllowedObjects flag. This is the case if "safe" HTTP request methods are used.
+     * Adds the given object to a list of allowed objects which may be persisted when persistOnlyAllowedObjects() is called.
```
albe, line 165:
```suggestion
     * Adds the given object to a list of allowed objects which may be persisted when persistAllowedObjects() is called.
```

albe:
Adjusted the method name according to feedback and replaced the existing `persistAll(true)` occurences in Flow

mficzel (approved):
Fine by reading

kdambekalns (approved):

PR #2587, opened 2021-09-30
TASK: Add release notes for Flow 7.2

kdambekalns:


kdambekalns (commented):
The need for FQCN for annotations on introduced properties should be in there…

Neos.Flow/Documentation/TheDefinitiveGuide/PartV/ReleaseNotes/720.rst
```diff
@@ -0,0 +1,166 @@
...
+
+This release of Flow comes with some new features, bugfixes and a lot of modernisation of the existing code base.
+As usual, we worked hard to keep this release as backwards compatible as possible but some of the changes might require manual
+adjustments. So please make sure to carefully read the upgrade instructions below.
+
+************
+New Features
+************
+
+`!!! FEATURE: Extensible configuration loaders <https://github.com/neos/flow-development-collection/pull/2449>`_
```
albe, line 13:
```suggestion

`!!! BUGFIX: Avoid broken proxy docblocks <https://github.com/neos/flow-development-collection/pull/2568>`_
----------------------------------------------

With PR #2533 docblocks are copied from the original class to the proxy class. This breaks when using annotations without the "standard" imports Flow and ORM. One example is the ImportedAsset domain model.

This fixes that by some changes to the proxy building.

Note, if you use property introduction via AOP, those properties must from now on use fully-qualified classnames for annoatations on them!

`!!! FEATURE: Extensible configuration loaders <https://github.com/neos/flow-development-collection/pull/2449>`_
```

PR #2588, opened 2021-10-04
Throw exception if headers are set through curl options.

### LINKED ISSUE #1886: CurlEngine / CURLOPT_HTTPHEADER
At the moment it is possible to use the request engine to set some additional options
('Neos\Flow\Http\Client\CurlEngine'). 

```
$curlHttpHeader = [
    'Accept: application/json',
    'Authorization: Secret 56-78-89',
    'Cache-Control: no-cache'
];

$this->requestEngine->setOption(CURLOPT_HTTPHEADER, $curlHttpHeader);
```

This  additional options will be never part of the request. The options will be overwritten 
in the line 68 or 131.

`curl_setopt($curlHandle, CURLOPT_HTTPHEADER, ['Expect:']); // Line 68` 

`curl_setopt($curlHandle, CURLOPT_HTTPHEADER, $preparedHeaders); // Line 131` 

Perhaps it is better to throw an Exception if somebody use them.

dlubitz:
Fixes #1886

### DIFF
--- a/Neos.Flow/Classes/Http/Client/CurlEngine.php
+++ b/Neos.Flow/Classes/Http/Client/CurlEngine.php
@@ -41,7 +41,7 @@ class CurlEngine implements RequestEngineInterface
      *
      * @param integer $optionName One of the CURLOPT_* constants
      * @param mixed $value The value to set
-     * @throws Http\Exception
+     * @throws \InvalidArgumentException
      */
     public function setOption($optionName, $value)
     {
--- a/Neos.Flow/Classes/Http/Client/CurlEngine.php
+++ b/Neos.Flow/Classes/Http/Client/CurlEngine.php
@@ -46,7 +46,7 @@ class CurlEngine implements RequestEngineInterface
     public function setOption($optionName, $value)
     {
         if ($optionName === CURLOPT_HTTPHEADER) {
-            throw new Http\Exception("Just CurlEngine specific options are allowed here. Please use request object for setting up request headers.", 1633334307);
+            throw new \InvalidArgumentException("Setting CURL headers is only possible via the request object and not by using the setOption method.", 1633334307);
         }
         $this->options[$optionName] = $value;
     }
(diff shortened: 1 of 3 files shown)


dlubitz:
As it might be breaking, I created the PR against master.

Neos.Flow/Classes/Http/Client/CurlEngine.php
```diff
@@ -37,12 +37,17 @@ class CurlEngine implements RequestEngineInterface
...
+     * Setting headers is not allowed, please use request object.
      *
      * @param integer $optionName One of the CURLOPT_* constants
      * @param mixed $value The value to set
+     * @throws Http\Exception
      */
     public function setOption($optionName, $value)
     {
+        if ($optionName === CURLOPT_HTTPHEADER) {
+            throw new Http\Exception("Just CurlEngine specific options are allowed here. Please use request object for setting up request headers.", 1633334307);
```
sorenmalling, line 49: I suggest a more specific text such as
"Setting CURL headers is only possible via the request object, and not by using the setOption method"

sorenmalling (changes requested):
Suggestion for exception message.

Should the exception class be "InvalidArgument" instead for a Http Exception class?

Neos.Flow/Classes/Http/Client/CurlEngine.php
dlubitz: I'm also fine with that. But the generic `InvalidArgumentException` or a specifc one in the Flow/Http (`Neos\Flow\Http\InvalidArgumentException`) Package?

albe: I'd say the latter one.

sorenmalling (changes requested):
Can I suggest a Test like this

```
    /**
     * Check if setting Http Headers directly in Curl throws Exception
     *
     * @test
     */
    public function setRequestHeadersViaOptionThrowsException()
    {
        self::expectExceptionObject(InvalidArgumentException::class);
        self::expectExceptionCode(1633334307);
        $this->browser->getRequestEngine()->setOption(CURLOPT_HTTPHEADER, ['X-Custom-Header' => 'Hello world']);
    }
```

in the CurlEngineTest class?

daniellienert (approved):

PR #2590, opened 2021-10-04
BUGFIX: Don't rely on doctrine using spl_object_hash

albe:
Doctrine 2.10 switched spl_object_hash for spl_object_id which breaks the deduplication listener. Thanks @dlubitz.
We should use `scheduleForDelete` instead of manually unsetting the to be inserted entity via the object hash.

See https://github.com/doctrine/orm/commit/84ad007de39bc0947be838c8efcf1455513cbdca

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationAndDeDuplicationListener.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationAndDeDuplicationListener.php
@@ -104,8 +104,6 @@ class ObjectValidationAndDeDuplicationListener
                 $knownValueObjects[$className][$identifier] = true;
             }
         }
-
-        ObjectAccess::setProperty($unitOfWork, 'entityInsertions', $entityInsertions, true);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sorenmalling (approved):
Code change looks good from reading and from readnig your slack discussion. Note, I'm not deep in knowledge of the Doctrine ORM change

dlubitz:
Looks good. But shouldn't it go into 5.3 as it is a bugfix?

kdambekalns (approved):
Looks good.

My guess regarding the target branch: should be 7.2, earlier versions don't have the affected Doctrine version?!

dlubitz:
They do, AFAIK:
```
        "doctrine/orm": "^2.6",
```
https://github.com/neos/flow-development-collection/blob/5.3/composer.json#L21

albe:
> Looks good. But shouldn't it go into 5.3 as it is a bugfix?

Flow 5.3 is security-only by now

dlubitz:
🙈 You are right, I mixed it up with Neos. ... But luckily this confusion got removed.

PR #2589, opened 2021-10-04
BUGFIX: Fix missing initialization for tags and missing imports

tweis:

PR #2593, opened 2021-10-05
BUGFIX: Fix missing initialization for tags and missing imports

tweis:
New pull request with new base on 6.3 for pull request #2589.


albe (approved):

albe:
Thanks a lot!

PR #2596, opened 2021-10-08
BUGFIX: Fixed $offset in XliffReader

MarcoPNS:
_Same as #2570 with changed base branch and test_

**What I did**
Fixed problem with the $offset value described in #2569

**How I did it**
$offset must be incremented in the while loop before the itterator. The first element is already handled before the loop. 

**How to verify it**
Create a .xlf File with at least 3 file blocks.


MarcoPNS:
@sorenmalling  I stumbled across this problem because I want to overwrite various `main` translations from other packages in my package. This would be a case where you need more than one file item.

A realistic example could then look like this

```
<?xml version="1.0"?>
<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
  <file original="" source-language="en" target-language="de" datatype="plaintext">
    <body>
      <trans-unit id="register">
        <source>Register</source>
        <target>Registrieren</target>
      </trans-unit>
    </body>
  </file>
  <file original="" product-name="Foo.Bar.Login.Package" source-language="en" target-language="de" datatype="plaintext">
    <body>
      <trans-unit id="form.username" xml:space="preserve">
        <source>Email</source>
        <target>E-Mail</target>
      </trans-unit>
    </body>
  </file>
  <file original="" product-name="Foo.Bar.Login.Reset.Package" source-language="en" target-language="de" datatype="plaintext">
    <body>
      <trans-unit id="mail.noAccountSubject" xml:space="preserve">
        <source>Set your new password</source>
        <target>Neues Passwort festlegen</target>
      </trans-unit>
    </body>
  </file>
</xliff>

```

sorenmalling:
@MarcoPNS Thanks for the example :) @fcool uses it similar and he will be reviewing this so we can merge it :)

albe (approved):
Indeed a great catch and very good PR! Thanks a lot and congrats on your first contribution :)

daniellienert (approved):

PR #2604, opened 2021-10-15
TASK: Update editorconfig files

kdambekalns:


kdambekalns:
Since it has no impact on the way Flow works but makes things easier, I'd like to merge this in 6.3 even though it's marked as TASK.

jonnitto:
We could also add `charset = utf-8` to the `[*]` section, wdyt?

jonnitto (approved):

PR #2608, opened 2021-10-19
BUGFIX: Check if variable is an array before accessing it as such

Kleisli:
The current code lead to an error when an argument is of type `Neos\Http\Factories\FlowUploadedFile`, trying to access an array property on it. This fixes it by checking for `is_array`.

Fixes neos/neos-development-collection#3467

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/Argument.php
+++ b/Neos.Flow/Classes/Mvc/Controller/Argument.php
@@ -223,7 +223,7 @@ class Argument
         $configuredType = $configuration->getConfigurationValue(ObjectConverter::class, ObjectConverter::CONFIGURATION_TARGET_TYPE);
         if ($configuredType !== null) {
             $this->dataType = $configuredType;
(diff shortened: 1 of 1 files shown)


sorenmalling:
Hi @Kleisli 

Thanks for your PR!

Two things I would like you to do:

 1. Can you please create a issue describing what brought this bug to your attention
 2. Create the patch against the 6.3 branch, since it's a bug that should be fixed in lower supported versions aswell

Kleisli:
@sorenmalling Ok, I can do that. 
The problem didn't exist in version 7.1, it was added in 7.2, so I assumed, it's a new feature. But I will check 6.3 also and may be even try to write a test for it. :)
Is there a deadline for the next bugfix release?

sorenmalling:
> The problem didn't exist in version 7.1, it

Oh man, if so: I'm sorry for not checking that! :) We release new bugfix releases when we have gathered a number of issues and one does upmerges :-)

PR #2609, opened 2021-10-20
Check explicit if an array is given

dlubitz:
Fixes https://github.com/neos/neos-development-collection/issues/3467


albe:
Note: Targets 7.2 because the if condition doesn't exist earlier

PR #2611, opened 2021-10-20
BUGFIX: AroundAdvice emits adviceInvoked signal

gjwnc:
The [documentation](https://neos.readthedocs.io/en/stable/References/Signals/Flow.html#aroundadvice-neos-flow-aop-advice-aroundadvice) mentions, that the `AroundAdvice` emit's the `adviceInvoked` signal. This is not the case, because `AroundAdvice` overrides the `invoke(JoinPointInterface $joinPoint)` method from `AbstractAdvice`.

**What I did**

This bugfix, just add's the missing signal, thus every advice should be easier to log using a slot for `adviceInvoked`.


albe (approved):
Thanks for this!

daniellienert (approved):
Looks good by reading

daniellienert:
@gjwnc tried to rebase it, but haven't access right on your fork. Would you please rebase the PR.

gjwnc:
@daniellienert I've rebased from the 6.3 branch.

PR #2613, opened 2021-10-22
BUGFIX: Catch Warning for rmdir in PHP 8.0

tweis:
When a resource is deleted the method `removeEmptyDirectoriesOnPath` is used, which cause an E_WARNING with PHP 8.0. Even if the shutup operator is used. This bugfix catch the E_WARNING to break the foreach at the right time.


albe (approved):
Though I dislike the throw Exception within a try/catch control flow it's only a matter of taste. Thanks for finding the bug and the alledged fix!

daniellienert (approved):

PR #2616, opened 2021-11-05
TASK: Add configuration file for RTD

kdambekalns:
This fixes an error caused by recent docutils and old sphinx being
incompatible.

The location of the configuration file in the root of the development
collection allows to (finally!) build the documentation directly from
the collection repository instead of the neos/flow split repository.

That finally fixes the "edit on GH" links, too.


kdambekalns:
I'll merge this without reviews, as it can only be tested in merged state and when the RTD setup has been changed to point to this repository. So this PR is FYI, in a way…

kdambekalns:
🤞 – https://readthedocs.org/projects/flowframework/builds/15190890/

kdambekalns:
The styling of ToC lists is a bit different now, I'll to have that fixed.

Issue #2618, opened 2021-11-08
Redis Cache does not respect the application context


daniellienert:
### Description

The redis backend does not respect the application context in its `buildKey` method. Without that, cache keys are the same if Flow is executed in different contexts.

**Caution! If you use a type of deployment, where every release has its own directory path (like surf / deployer), you either need to set the `FLOW_PATH_ROOT` environment variable or adjust the `applicationIdentifier` in order to be able to flush the redis cache after the release switch step.**

#### Expected behavior

Like file and pdo backend, the redis backend should respect the application context.

### Affected Versions

Neos: * 
Flow: *

skurfuerst:
Hey @daniellienert , 

this can be closed as it is fixed, right? :-)

All the best,
Sebastian

daniellienert:
Right

PR #2619, opened 2021-11-08
TASK: Include the application context in cache keys

daniellienert:
resolves: #2618

- Refactors the applicationContextHash to be available for all backends
- Use the application context in pdo and redis backend

### DIFF
--- a/Neos.Cache/Classes/Backend/AbstractBackend.php
+++ b/Neos.Cache/Classes/Backend/AbstractBackend.php
@@ -35,6 +35,11 @@ abstract class AbstractBackend implements BackendInterface
     /**
      * @var string
      */
+    protected $applicationContextHash;
+
+    /**
+     * @var string
+     */
     protected $cacheIdentifier;
 
     /**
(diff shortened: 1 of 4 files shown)


daniellienert:
Happy to discuss, if this is considered a bugfix or a breaking change :)

kitsunet:
Sounds like a bugfix to me. we might consider thinking about this in a larger context, I think atm each backend decides for itself if and how to apply this information.

kdambekalns (approved):
Looks good and like a bugfix. Are there more backends to fix, potentially?

daniellienert:
@kdambekalns all other backends are fine
@kitsunet I am open for suggestions. Seemed to me, that other backends vary the cache key generation depending on their backend technology

gerhard-boden:
Thanks for the fix!
Since 90% of our projects run on the PdoBackend (SQLite) I'm very happy to untertake some real world testing
If I understand the problem correctly the backend would not differentiate between production and dev context for example?

daniellienert:
> Thanks for the fix! Since 90% of our projects run on the PdoBackend (SQLite)
@gerhard-boden The PDO backend already respected the context - you are safe :)

daniellienert:
Closed in favor of #2622

gerhard-boden:
> > Thanks for the fix! Since 90% of our projects run on the PdoBackend (SQLite)
> > @gerhard-boden The PDO backend already respected the context - you are safe :)

ah thanks for the info 😄 , thought this had something to do with it https://github.com/neos/flow-development-collection/pull/2619/files#diff-43cc6c26a0807aa1ab29bebc0bfcfcb50fafba53bcb2443830a160b6185162a5

PR #2621, opened 2021-11-09
Fix #2620

### LINKED ISSUE #2620: postPackageUpdateAndInstall not run on multiple package install
### Description

Upon installing a project either through create-project or when checking out a new project and running composer install the postPackageUpdateAndInstall is not run for all of the installed packages.

### Steps to Reproduce

1. check out a new neos package with multiple uses of installer-resource-folders
2. composer install

#### Expected behavior

I should see all of the resources defined by the installer-resource-folders available in the installed application

#### Actual behavior

Only one of the packages wins the race to postPackageUpdateAndInstall and will have resources installed, all others are ignored

### Affected Versions

Neos: 7.2.0

Flow: 7.2.1

marcrobertscamao:
This fix resolves the issue of postPackageUpdateAndInstall not installing resources from all installed packages

**What I did**

Remove the single run code for the postPackageUpdateAndInstall function only

**How to verify it**

1. Clone git@github.com:marcrobertscamao/resource-test.git
2. Run composer install
3. See resource1.txt in the root and resource2.txt in the root

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Composer/InstallerScripts.php
+++ b/Neos.Flow/Classes/Composer/InstallerScripts.php
@@ -29,11 +29,6 @@ class InstallerScripts
     /**
      * @var bool
      */
-    protected static $postPackageUpdateAndInstallAlreadyRun = false;
-
-    /**
-     * @var bool
-     */
     protected static $postUpdateAndInstallAlreadyRun = false;
 
     /**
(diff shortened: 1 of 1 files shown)


daniellienert:
@marcrobertscamao, thank you very much for the fix!
I did a proper revert commit to my original commit to keep the reference in #2623 and would like to close this one.

PR #2623, opened 2021-11-09
Revert "TASK: Run post-update scripts only once"

daniellienert:
This reverts commit d79947442d7de8c59f0d6460e0affef30406148b.

Resolves: #2621

Original by @marcrobertscamao:

This fix resolves the issue of postPackageUpdateAndInstall not installing resources from all installed packages

What I did

Remove the single run code for the postPackageUpdateAndInstall function only

How to verify it

Clone git@github.com:marcrobertscamao/resource-test.git
Run composer install
See resource1.txt in the root and resource2.txt in the root

### DIFF
--- a/Neos.Flow/Classes/Composer/InstallerScripts.php
+++ b/Neos.Flow/Classes/Composer/InstallerScripts.php
@@ -25,17 +25,6 @@ use Neos\Utility\Files;
  */
 class InstallerScripts
 {
-
-    /**
-     * @var bool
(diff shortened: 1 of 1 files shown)


albe (approved):
Too sad this wasn't side-effect free.

kdambekalns (approved):
Well, I was unmoved by the first change, so I'm fine with undoing it. 🤷‍♂️

PR #2624, opened 2021-11-10
BUGFIX: Fix proxy compilation for edge cases

### LINKED ISSUE #1835: Proxy compilation breaks on certain strings
### Description

If a class subject to proxy compilation contains certain strings, the proxy compilation will interpret them as the class / interface name and throw an exception saying, that the class name does not match the file name.

This happens for all strings (as in collection of characters, not actual PHP strings) in a file that follow the following rules:
* After the line break there must only be lower case characters or whitespace
* Must contain `class something` or `interface something`  

### Steps to Reproduce

1. Create class `TestClass` containing one of the examples below in a package
2. Trigger compilation by running `flow:cache:flush` and `flow:core:compile` commands
3. Observe the following exception: `The name of the class "bar" is not the same as the filename which is "TestClass.php"`

#### Example classes

**Breaks**
```
<?php declare(strict_types=1);

namespace Vendor\Foo\Bar;

class TestClass
{
    public function doSomething()
    {
        echo '
            only whitespace and lowercase class foo
        ';
    }
}

```

**Breaks**
```
<?php declare(strict_types=1);

namespace Vendor\Foo\Bar;

/*
class foo
*/
class TestClass
{
}
```

**Does not break: String prefixed by upper case character**
```
<?php declare(strict_types=1);

namespace Vendor\Foo\Bar;

class TestClass
{
    public function doSomething()
    {
        echo '
            Uppercase class foo
        ';
    }
}
```

**Does not break: String prefixed by special character `*`**
```
<?php declare(strict_types=1);

namespace Vendor\Foo\Bar;

/*
* class foo
*/
class TestClass
{
}
```

#### Expected behavior

Proxy should compile sucessfully, the `flow:core:compile` command should not have any output.

#### Actual behavior

Exception  `The name of the class "bar" is not the same as the filename which is "TestClass.php"` is thrown.

### Affected Versions

I have only tested this on my current setup, which uses the following versions:

Neos: 4.3.6
Flow: 5.3.7

### Additional information

So far I have tracked this issue down to the regex `/^([a-z\s]*?)(final\s+)?(interface|class)\s+([a-zA-Z0-9_]+)/m` used in `Neos\Flow\ObjectManagement\Proxy\Compiler::cacheOriginalClassFileAndProxyCode` which seems to be a bit to lax:

robertlemke:
This change fixes proxy compilation for certain cases where "class"
was used as a string in multi-line quoted strings or within comments.

Resolves: #1835

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -11,8 +11,8 @@ namespace Neos\Flow\ObjectManagement\Proxy;
  * source code.
  */
 
-use Neos\Flow\Annotations as Flow;
 use Neos\Cache\Frontend\PhpFrontend;
+use Neos\Flow\Annotations as Flow;
 use Neos\Flow\ObjectManagement\CompileTimeObjectManager;
 use Neos\Flow\Reflection\ReflectionService;
 use Neos\Flow\Tests\BaseTestCase;
@@ -245,16 +245,8 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
     protected function cacheOriginalClassFileAndProxyCode($className, $pathAndFilename, $proxyClassCode)
     {
         $classCode = file_get_contents($pathAndFilename);
+        $classCode = $this->replaceClassName($classCode, $pathAndFilename);
         $classCode = $this->stripOpeningPhpTag($classCode);
-
-        $classNameSuffix = self::ORIGINAL_CLASSNAME_SUFFIX;
-        $classCode = preg_replace_callback('/^([a-z\s]*?)(final\s+)?(interface|class)\s+([a-zA-Z0-9_]+)/m', static function ($matches) use ($pathAndFilename, $classNameSuffix, $proxyClassCode) {
-            $classNameAccordingToFileName = basename($pathAndFilename, '.php');
-            if ($matches[4] !== $classNameAccordingToFileName) {
-                throw new Exception('The name of the class "' . $matches[4] . '" is not the same as the filename which is "' . basename($pathAndFilename) . '". Path: ' . $pathAndFilename, 1398356897);
-            }
-            return $matches[1] . $matches[3] . ' ' . $matches[4] . $classNameSuffix;
-        }, $classCode);
 
         // comment out "final" keyword, if the method is final and if it is advised (= part of the $proxyClassCode)
         // Note: Method name regex according to http://php.net/manual/en/language.oop5.basic.php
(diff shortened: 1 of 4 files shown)


Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -349,4 +341,45 @@ protected static function renderOptionArrayValueAsString(array $optionValue)
...
+     * Appends ORIGINAL_CLASSNAME_SUFFIX to the original class name
+     *
+     * @param string $classCode
+     * @param string $pathAndFilename
+     * @return string
+     * @throws Exception
+     */
+    protected function replaceClassName(string $classCode, string $pathAndFilename): string
+    {
+        $tokens = token_get_all($classCode);
```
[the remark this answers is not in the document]
kdambekalns: 🤷‍♂️ Compilation is expensive anyway…

kdambekalns (approved):
Fine with me, by reading

kdambekalns:
Dang… seems as if the extra-weird testcase by @albe makes it fail… 🙈 What to do about that?

Neos.Flow/Tests/Functional/ObjectManagement/ProxyCompilerTest.php
```diff
@@ -134,6 +134,14 @@ public function proxiedFinalClassesAreStillFinal()
         self::assertTrue($reflectionClass->isFinal());
     }
 
+    /**
+     * @test
+     */
+    public function classKeywordIsIgnoredInsideClassBody()
+    {
+        $reflectionClass = new ClassReflection(Fixtures\ClassWithKeywordsInClassBody::class);
```
kitsunet, line 142: Does this maybe load the original class?

kitsunet (changes requested):
The test failure is correct, this does not work yet. If you look at the proxy class you find this:

```
/*
class foo
*/
/*
class bar_Original */class /* oddly placed comment for class */ ClassWithKeywordsInClassBody //class quux_Original
{
```

So it created "bar_Original" there, even though the proxy class file is named correctly AND the proxy is

`class ClassWithKeywordsInClassBody extends ClassWithKeywordsInClassBody_Original implements \Neos\Flow\ObjectManagement\Proxy\ProxyInterface`

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -349,4 +341,46 @@ protected static function renderOptionArrayValueAsString(array $optionValue)
...
+        $previousToken = [0 => null];
+        foreach ($tokens as $token) {
+            if (isset($classToken) && is_array($token) && $token[0] === T_STRING) {
+                $detectedClassName = $token[1];
+                $lines = explode(PHP_EOL, $classCode);
+
+                // We can't allow Unicode characters for class names even though PHP does, because such class names
+                // don't qualify as valid cache entry identifiers.
+                $classTokenLine = preg_replace('/([a-z\s]*?)(final\s+)?(interface|class)\s+([a-zA-Z0-9_]+)/', '$1$3 $4' . self::ORIGINAL_CLASSNAME_SUFFIX, $lines[$token[2] - 1]);
```
kitsunet, line 364: It becomes obvious the test must break this, we identify the line with the class token correctly, but then do preg replace without taking the position of the `class` token into account nor knowing if what follows the `class` keyword is actually the class name.

albe: Hm, so what if we just collect all tokens from the current line, then change only the one T_STRING with the class name and concatenate them back into source code for this one line? Should be ugly but fool-proof, no?

albe:
> Dang… seems as if the extra-weird testcase by @albe makes it fail… 🙈 What to do about that?

Sorry! 🙈 Maybe document as `DONT WRITE COMMENTS BETWEEN THE CLASS KEYWORD AND YOUR CLASSNAME THAT CONTAINS THE WORD "CLASS", STUPID!` cuts it? 😂

For real though, Christian is onto it - see the comment

robertlemke:
I took a new approach: Instead of using a regex replacing the class name, I regenerate the beginning of the source code until the class name appears by rendering the respective tokens. Usually, this amount of code until the class name appears, will be relatively short, so I guess that taking this approach is feasible in terms of speed.

I also added a new unit test cases which covers more edge cases than the ones we already had in mind.

I tested this patch with a real application and ran all tests, so this should hopefully work in all cases.

Is it okay to target this to 8.3 and 9.0 or do I need to port this to earlier versions, too?

kitsunet (approved):
> Is it okay to target this to 8.3 and 9.0 or do I need to port this to earlier versions, too?

Mmm, might be a bit annoying for upmerges? But then it's mostly about object management, so problaby not much will change anymore. Fine by me I guess. Also did a quick test with php 8.1 and neos 8.3 and it seems to work fine.

kdambekalns:
> Is it okay to target this to 8.3 and 9.0 or do I need to port this to earlier versions, too?

I think 8.3 is fine, but if you really like, it could even go into ealier releases. Not a must, though.

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -245,20 +245,12 @@ public function getStoredProxyClassMap()
...
-            if ($matches[4] !== $classNameAccordingToFileName) {
-                throw new Exception('The name of the class "' . $matches[4] . '" is not the same as the filename which is "' . basename($pathAndFilename) . '". Path: ' . $pathAndFilename, 1398356897);
-            }
-            return $matches[1] . $matches[3] . ' ' . $matches[4] . $classNameSuffix;
-        }, $classCode);
-
         // comment out "final" keyword, if the method is final and if it is advised (= part of the $proxyClassCode)
         // Note: Method name regex according to http://php.net/manual/en/language.oop5.basic.php
-        $classCode = preg_replace_callback('/^(\s*)((public|protected)\s+)?final(\s+(public|protected))?(\s+function\s+)([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+\s*\()/m', static function ($matches) use ($proxyClassCode) {
+        $classCode = preg_replace_callback('/^(\s*)((public|protected)\s+)?final(\s+(public|protected))?(\s+function\s+)([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+\s*\()/m', static function ($matches) use ($pathAndFilename, $proxyClassCode) {
```
kdambekalns:
```suggestion
        $classCode = preg_replace_callback('/^(\s*)((public|protected)\s+)?final(\s+(public|protected))?(\s+function\s+)([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+\s*\()/m', static function ($matches) use ($proxyClassCode) {
```
as `$pathAndFilename` is not used in the function body…

kdambekalns (approved):
A nitpick, but otherwise looks good to go.

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
robertlemke: ✅

robertlemke:
This fix is now targeted to 8.0.

PR #2626, opened 2021-11-15
BUGFIX: Correctly adjust Cache-Control in makeStandardsCompliant()

### LINKED ISSUE #2625: ResponseInformationHelper breaks Cache-Control header
### Description

When settings the `Cache-Control` header, the `ResponseInformationHelper` breaks it in `makeStandardsCompliant` under certain circumstances.

### Steps to Reproduce

1. `$this->response->setHttpHeader('Cache-Control', 'must-revalidate');`

#### Expected behavior

`Cache-Control: must-revalidate` in response

#### Actual behavior

`Cache-Control: e` in response

### Affected Versions

Flow: ^7.0

kdambekalns:
Fixes #2625


kitsunet (approved):

kdambekalns:
From the 🤦‍♂️ category…

robertlemke (commented):
Ideally we had a test which would cover this scenario …

sorenmalling:
> Ideally we had a test which would cover this scenario …

You are only one commit away from making that situation better :-)

sorenmalling:
My test writing skills are limited, but something along this in the `ResponseInformationHelperTest` class

```
    /**
     * @test
     */
    public function makeStandardCompliantEnsureCacheControlHeader()
    {
        $request = ServerRequest::fromGlobals();
        $response = new Response(200, ['Cache-Control' => 'must-revalidate']);
        $response = $response->withBody(Utils::streamFor(''));
        self::assertTrue($response->hasHeader('Cache-Control'));

        $compliantResponse = ResponseInformationHelper::makeStandardsCompliant($response, $request);
        self::assertTrue($compliantResponse->hasHeader('Cache-Control'));
        $cacheControlHeaderValue = $compliantResponse->getHeaderLine('Cache-Control');

        self::assertEquals('must-revalidate', $cacheControlHeaderValue);
    }
```

Inspired/copy-pasted from `makeStandardCompliantAddsContentLengthHeader` in same file

Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
```diff
@@ -198,11 +198,9 @@ public static function makeStandardsCompliant(ResponseInterface $response, Reque
             $response = $response->withBody(stream_for(''));
         }
 
-        $cacheControlHeaderLine = $response->getHeaderLine('Cache-Control');
-
-        if ((!empty($cacheControlHeaderLine) && strpos('no-cache', $cacheControlHeaderLine) !== false)
+        $cacheControlHeaderValue = $response->getHeaderLine('Cache-Control');
+        if (($cacheControlHeaderValue !== '' && strpos('no-cache', $cacheControlHeaderValue) !== false)
```
albe, line 202: Now that I look at this - the not empty check made me suspicious - `strpos` is "haystack, needle", so shouldn't that be `strpos($cacheControlHeaderValue, 'no-cache')` and then the empty check isn't needed any more? See https://3v4l.org/gOJHp#v7.4.25

kdambekalns:
Woah, it's actually worse. The code in question is supposed to remove the `max-age` directive, but it doesn't do that (correctly). This is the "old" version, before Flow 6.0: https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Classes/Http/Response.php#L462

albe:
> Woah, it's actually worse. The code in question is supposed to remove the `max-age` directive, but it doesn't do that (correctly). This is the "old" version, before Flow 6.0: https://github.com/neos/flow-development-collection/blob/5.3/Neos.Flow/Classes/Http/Response.php#L462

That's what https://github.com/neos/flow-development-collection/pull/2626/files#diff-02cb00b5cb546318cf5946806a395c950d6f1618328be8956e6532f6ec187f05L206 is supposed to do, right? It's missing the  value and the potential `,`, right?

Here's how it could work: https://3v4l.org/5BnpG#v7.4.25

sorenmalling:
Ping @albe @robertlemke Would you re-review latest changes?

Neos.Flow/Classes/Http/CacheControlDirectives.php
```diff
@@ -0,0 +1,195 @@
...
+     */
+    public function setDirective(string $name, ?string $value = null): void
+    {
+        switch ($name) {
+            case 'public':
+                $this->cacheDirectives['visibility'] = 'public';
+            break;
+            case 'private':
+            case 'no-cache':
+                $this->cacheDirectives['visibility'] = $name . (!empty($value) ? '="' . $value . '"' : '');
```
albe, line 84: TIL that *response* `no-cache` and `private` directives may specify a list of "fields" (headers, https://datatracker.ietf.org/doc/html/rfc2616#page-110) that may not be returned in the cached response without revalidation as opposed to the case without fields, where the whole response may not be returned without revalidation.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#no-cache is not comprehensive on this part

albe (approved):

PR #2628, opened 2021-11-16
BUGFIX: FunctionalTestCase doesn't reset PrivilegeManager on setup

tvt:
When extending the `FunctionalTestCase` without either `$testableSecurityEnabled` or `$testablePersistenceEnabled` set to `true`, the `privilegeManager` will get `overrideDecision` set to `true`.

All test classes executed later will now also have `overrideDecision` set to `true` instead of the expected `null`.

This fix ensures that `privilegeManager->reset()` is called no matter what.


sorenmalling:
@tvt Great catch 👍

Will you be so kind and create a corresponding issue, that this PR can be referred to with a "Fixes #.." line :)

tvt:
@sorenmalling #2629 should hopefully be enough? :)

sorenmalling:
> @sorenmalling #2629 should hopefully be enough? :)

It's perfect :)

kitsunet (approved):
Good point, thanks for digging in and taking care <3

fcool (approved):

PR #2632, opened 2021-11-19
BUGFIX: IgnoreValidation can be used without arguments on a property

albe:
It should be possible to annotate a property with just `@Flow\IgnoreValidation` without any arguments.

This was a regression from #2468.

Related to https://github.com/neos/neos-development-collection/issues/3515 and https://github.com/neos/neos-development-collection/pull/3411


albe:
Psalm failure see https://github.com/neos/flow-development-collection/pull/2633

bwaidelich (approved):
Thanks! <3

bwaidelich:
The failing Psalm check is unrelated and will be fixed with #2633

Issue #2634, opened 2021-11-26
Cache setup with PDOBackend can fail due to file monitoring calling flush() on a cache early on


kitsunet:
Due to file monitors (think Neos eg.) the cache setup can fail to be executed as a flush might be called on a cache that is configured as PDO without it being ever setup, leading to a cycle of death as you cannot call cache setup without the tables and you cannot get the tables without cache setup.

This is reproducible for me in a standard neos dev setup and configuring some of the Neos caches for PDO.

PR #2635, opened 2021-11-26
BUGFIX: PDOBackend ignores missing tables in flush

kitsunet:
This prevents errors for cache setup with PDOBackends for caches that are flushed early on eg. due to file monitors.

Fixes: #2634


sorenmalling (approved):

jonnitto (approved):
Make sense. Thank you for the fix

kdambekalns (commented):
👍

Issue #1108, opened 2017-11-08
Custom error views configuration is broken


kdambekalns:
### Description

When following https://flowframework.readthedocs.io/en/latest/TheDefinitiveGuide/PartIII/ErrorAndExceptionHandling.html#custom-error-views in a Flow version starting with 4.0 the results will not be as expected.

### Steps to Reproduce

1. Configure a custom error view as documented
2. Trigger an exception
3. Marvel at the expected template not being shown

#### Expected behavior

The custom template should be used.

#### Actual behavior

No template is used.

### Affected Versions

Flow: 4.0 (since the introduction of TYPO3Fluid, presumably)

kdambekalns:
The reason for this behaviour is as follows:

When rendering an exception using `echoExceptionWeb()` both the production and the debug exception handler check for `$this->renderingOptions['templatePathAndFilename'])` first, to see if a custom view is defined.

If that is the case, `buildView()` is used. That builds a view based on `$renderingOptions['viewOptions']` (with `viewOptions` not even mentioned in the documentation.) Then a method called `applyLegacyViewOptions()` is called, which uses `ObjectAccess::setProperty()` to make use of 

- `templatePathAndFilename`
- `layoutRootPath`
- `partialRootPath`
- `format`

if set directly in `$renderingOptions`.

The results is that those options are not set at all, since those properties are not existing in the view not do setters exist. Instead `setOption()` would need to be used.

In the end a user has to set the (no longer working) `templatePathAndFilename` (to anything non-empty, in fact) to enable custom error view processing _and_ the needed `viewOptions.templatePathAndFilename` to the requested template.

kdambekalns:
Another shortcoming: the `viewClassName` is only used if `templatePathAndFilename` is set, even if the view does not need a template. This could be fixed along the way…

kitsunet:
Note that IMHO the concept of custom error views is inherently broken because they need the resource wrapper which is not available early on and so you get an exception that the template could not be loaded, masking the original problem. (Could be solved with a fallback probably)

kdambekalns:
Well, you can use an absolute path in case you want to handle exceptions that occur "too early". But the concept works well for any "non-500" exceptions, IMHO.

mhsdesign:
I had the same problem yesterday - my duplicate issue: #2627

mhsdesign:
[...]

Issue #2636, opened 2021-11-26
Custom error views options are broken


mhsdesign:
Currently, custom error views https://flowframework.readthedocs.io/en/latest/TheDefinitiveGuide/PartIII/ErrorAndExceptionHandling.html#custom-error-views can be defined like this (as shown in the docs):

[link](https://github.com/neos/flow-development-collection/blob/0fd339c365397ed27430b1ce47963ec7761cb934/Neos.Flow/Configuration/Settings.Error.yaml#L33)
```yaml
Neos:
  Flow:
    error:
      exceptionHandler:
        renderingGroups:
          databaseConnectionExceptions:
            matchingExceptionClassNames: ['Neos\Flow\Persistence\Doctrine\Exception\DatabaseException']
            options:
              templatePathAndFilename: 'resource://Neos.Flow/Private/Templates/Error/Default.html'
              variables:
                errorDescription: 'Sorry, the database connection couldn''t be established.'
```

## Problem Nr 1.
the one problem is, that the template 'resource://Neos.Flow/Private/Templates/Error/Default.html' is fluid and flow is now independent of fluid and uses the Neos\Flow\Mvc\View\SimpleTemplateView [link](https://github.com/neos/flow-development-collection/blob/0fd339c365397ed27430b1ce47963ec7761cb934/Neos.Flow/Configuration/Settings.Error.yaml#L16) for template resolution unless the fluid adaptor is installed, which will override it and use: Neos\FluidAdaptor\View\StandaloneView [link](https://github.com/neos/flow-development-collection/blob/a8f81b4f3ddafe30f50dcd6092f98cb3c651004d/Neos.FluidAdaptor/Configuration/Settings.yaml#L19)

## Problem Nr 2.
and the second problem is that using `options.templatePathAndFilename` or really anything like `options.(templatePathAndFilename|layoutRootPath|partialRootPath|format)` are not set with `viewClassName->setOptions($options)` but are set with ObjectAccess::setProperty directly on the view instance. [link](https://github.com/neos/flow-development-collection/blob/0fd339c365397ed27430b1ce47963ec7761cb934/Neos.Flow/Classes/Error/AbstractExceptionHandler.php#L184)

That means that the view needs to work with object properties like `(templatePathAndFilename|layoutRootPath|partialRootPath|format)` but that's not the case for the default SimpleTemplateView and not the Neos way to do this. One is supposed to set the options with $view->setOption.
[...]

PR #2638, opened 2021-11-29
BUGFIX: Use custom error view for rendering group independent of the configuration 'templatePathAndFilename'

mhsdesign:
resolves: #1108

previously, a custom error view is for rendering groups only used if `templatePathAndFilename` is set. See #1108

**A custom view will now be used if ...**
-  ... there is a matching rendering group (no further checks. It *could* be also a rendering group with empty options (no viewClassName or viewOptions) ... what would then trow an error probably - depending on the view)
- ... `defaultRenderingOptions.templatePathAndFilename` passes isset() (to not change previous working behaviour - should get depreceated sometime)

### DIFF
--- a/Neos.Flow/Classes/Error/DebugExceptionHandler.php
+++ b/Neos.Flow/Classes/Error/DebugExceptionHandler.php
@@ -64,7 +64,7 @@ EOD;
             header(sprintf('HTTP/1.1 %s %s', $statusCode, $statusMessage));
         }
 
-        if (!isset($this->renderingOptions['templatePathAndFilename'])) {
+        if ($this->useCustomErrorView === false) {
             $this->renderStatically($statusCode, $exception);
             return;
         }
--- a/Neos.Flow/Classes/Error/DebugExceptionHandler.php
+++ b/Neos.Flow/Classes/Error/DebugExceptionHandler.php
@@ -64,8 +64,7 @@ EOD;
             header(sprintf('HTTP/1.1 %s %s', $statusCode, $statusMessage));
         }
 
-        $useCustomErrorView = isset($this->renderingOptions['templatePathAndFilename']) || isset($this->renderingOptions['renderingGroup']);
-        if ($useCustomErrorView === false) {
+        if ($this->useCustomErrorView() === false) {
             $this->renderStatically($statusCode, $exception);
             return;
         }
--- a/Neos.Flow/Classes/Error/DebugExceptionHandler.php
+++ b/Neos.Flow/Classes/Error/DebugExceptionHandler.php
@@ -64,7 +64,8 @@ EOD;
             header(sprintf('HTTP/1.1 %s %s', $statusCode, $statusMessage));
         }
 
-        if ($this->useCustomErrorView === false) {
+        $useCustomErrorView = isset($this->renderingOptions['templatePathAndFilename']) || isset($this->renderingOptions['renderingGroup']);
+        if ($useCustomErrorView === false) {
             $this->renderStatically($statusCode, $exception);
             return;
         }
--- a/Neos.Flow/Classes/Error/AbstractExceptionHandler.php
+++ b/Neos.Flow/Classes/Error/AbstractExceptionHandler.php
@@ -48,9 +48,18 @@ abstract class AbstractExceptionHandler implements ExceptionHandlerInterface
     protected $options = [];
 
     /**
+     * Merged custom error view options from defaultRenderingOptions and of the first matching renderingGroup
+     *
      * @var array
      */
     protected $renderingOptions;
+
+    /**
+     * Should a custom error view be used
+     *
+     * @var bool
+     */
+    protected $useCustomErrorView;
 
     /**
      * @param LoggerInterface $logger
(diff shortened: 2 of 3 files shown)


mhsdesign:
i think using `$this->useCustomErrorView` in resolveCustomRenderingOptions() like a side effect makes sense as this wont break any functionality (if someone builds themselves an own ExceptionHandler and used resolveCustomRenderingOptions())

mhsdesign:
hmm instead of using the property $useCustomErrorView one could also set an array key on the prop $renderingOptions - is this better?

mhsdesign:
or extract the minimal reused logic into a new method `useCustomErrorView(): bool;` ?

what do you say?

mhsdesign:
documentation should be updated too, will make a separate PR regarding that...

sorenmalling:
I feel like we don't address the real issue. How about we take this for a "bigger spin", something like: 

Since "templatePathAndFilename" is referring to FluidAdaptor or the SimpleTemplateView implementation, this could/should be removed from a requirement of configuration (sine it is view dependent) and instead be a "options" that a view can be initialized with.

Since Flows ViewInterface has a "createWithOptions" method, to initiliaze a certain view, how about we declare a view class name, and pass options to that view.

mhsdesign:
i do agree - this pr is more of a workaround.
But it is meant to be a non breaking bugfix...

In a breaking Neos update all this legacy fluid stuff should be removed. And only "options" should be used and not literal object props with ObjectAccess as you said.
I already explained it already here (but it does require some migration, and its not fully compatible yet as there is no supported option for 'format'):
https://github.com/neos/flow-development-collection/issues/2636#solution

sorenmalling:
We must dare to challenge the current to change what we have now

sorenmalling:
> We must dare to challenge the current to change what we have now

@mhsdesign asked what I meant by this :-) In short: Sometimes we have to break what we have now, to make it better

bwaidelich:
> Sometimes we have to break what we have now, to make it better

I agree, but my take on it would be: Once we have found a better interface/implementation, there is almost always a way to support the previous configuration (at least for a while).

mhsdesign:
current behavior:

https://github.com/neos/flow-development-collection/blob/97cfa9183f9a94bcbf3f7daf22542819bb833837/Neos.Flow/Classes/Error/AbstractExceptionHandler.php#L160-L161

As it seems to me @bwaidelich we are already in this step of having new and old ways to define the exception rendering view - but somehow it was forgotten about a way to use only the new way - thats what my PR provides as bugfix - independence of the old 'templatePathAndFilename' and check instead if a 'renderingGroup' was found. Also the old options like 'templatePathAndFilename' were never deprecated but co existed instead with the non documented better 'viewOptions' which is also used.

So saying that 'templatePathAndFilename' is 'old' is actually not justified, as it is not clear to the user yet as it was never deprecated nor does he know of the 'viewOptions.templatePathAndFilename' - and neos and flow still use it in core.

code wise its clear, that 'templatePathAndFilename' must be 'old' since its not set on the view via `AbstractView::setOption()` but directly on the object with `ObjectAccess::setProperty()` in a method named `applyLegacyViewOptions()` <- pointing to **Legacy**

---------

I did a little history check:
2012 was the feature added and had since the beginning the check for and template path as switch https://github.com/neos/flow-development-collection/commit/cece2b27a2a372538f31703b4ec1f2625d63e8ae#diff-cfac78b9057f388024a59d163f404725f1765650f5e30b3f037993f216b00d32R42

2016 with the Task to make Flow independent of Flow `viewOptions` and `viewClassName` were (silently) added and the legacy flow view options like `templatePathAndFilename` were moved to the method `applyLegacyViewOptions()`.
I think with that change, that `templatePathAndFilename` should have been deprecated and also the `viewOptions` and `viewClassName` should have been announced and documented.
Also i think, this change should have made the determination of using a view independent of `templatePathAndFilename` as my PR shows via: `renderingGroup`
https://github.com/neos/flow-development-collection/commit/4efc14074513338045d1b33312285ab28f63328f#diff-e7e989bb4b0cb3ac239d10d2d8b47a5488a88acc512bfd3253d2d9e4d354a52cR135-R199

Neos.Flow/Classes/Error/DebugExceptionHandler.php
```diff
@@ -64,7 +64,8 @@ protected function echoExceptionWeb($exception)
             header(sprintf('HTTP/1.1 %s %s', $statusCode, $statusMessage));
         }
 
-        if (!isset($this->renderingOptions['templatePathAndFilename'])) {
+        $useCustomErrorView = isset($this->renderingOptions['templatePathAndFilename']) || isset($this->renderingOptions['renderingGroup']);
```
bwaidelich, line 67: This could be extracted to a (protected) method of the `AbstractExceptionHandler` so that we could replace it below (and in the `ProductionExceptionHandler`) via:
```php
if (!$this->useCustomErrorView()) {
  ...
}
```

bwaidelich (commented):
To be honest, I didn't closely look into this one yet.
The
> there is almost always a way to support the previous configuration
was more of a general comment :)

I'm not sure how a v2 API for the exception handling could look like, so I just added a mini code style suggestion :)

Neos.Flow/Classes/Error/DebugExceptionHandler.php
mhsdesign: Done ;)

mhsdesign:
> I'm not sure how a v2 API for the exception handling could look like [...]

i do, but it would make sense to me to merge this one to fix the bug of the past - we can still chnage things with Neos 8.0 or later ;)

albe (approved):
Thanks for digging into this! I quickly looked at the issue and oh boy... good to have that fixed. A good follow-up would be to improve the documentation then (not sure exactly what would be missing there).

mhsdesign:
a good breaking change for 8.0 would be to remove the old legacy flow options

```diff
$options array{
    viewClassName: string,
    viewOptions: array,
    renderTechnicalDetails: bool,
    logException: bool,
    variables?: array,
-    templatePathAndFilename?: string,
-    layoutRootPath?: string,
-    partialRootPath?: string,
-    format?: string
}
```

one would to change a few things:

```diff
options:
-  templatePathAndFilename: 'file'
-  layoutRootPath: 'path'
-  partialRootPath: 'path'
-  format: 'html'
+  viewOptions:
+    templatePathAndFilename: 'file'
+    layoutRootPaths: ['path']
+    partialRootPaths: ['path']
```
(there is no equivalent for `format`)


(docs outdated https://flowframework.readthedocs.io/en/latest/TheDefinitiveGuide/PartIII/ErrorAndExceptionHandling.html#custom-error-views)

bwaidelich (approved):
makes sense (by reading) thanks for taking care!

mhsdesign:
next steps are to adjust the documentation and to propose a breaking removal of the legacy option in with flow 8

mhsdesign:
thanks for reviewing this ;)

could you please also take a look at the follow up docs pr #2731 ?

Issue #2639, opened 2021-12-01
Error after composer install/update with PHP 8 and Flow < 7


gerdemann:
### Description

I get the following error after `composer install` or `composer update` with PHP 8 and Flow < 7:
```
Fatal error: Uncaught Error: Class 'Package' not found in /data/Packages/Framework/Neos.Flow/Classes/Package/PackageFactory.php:53
Stack trace:
#0 /data/Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php(719): Neos\Flow\Package\PackageFactory->create('/data/Packages/', 'Framework/Neos....', 'Neos.Eel', 'neos/eel', Array, Array)
#1 /data/Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php(701): Neos\Flow\Package\PackageManager->registerPackageFromStateConfiguration('neos/eel', Array)
#2 /data/Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php(148): Neos\Flow\Package\PackageManager->registerPackagesFromConfiguration(Array)
#3 /data/Packages/Framework/Neos.Flow/Classes/Core/Booting/Scripts.php(183): Neos\Flow\Package\PackageManager->initialize(Object(Neos\Flow\Core\Bootstrap))
#4 /data/Packages/Framework/Neos.Flow/Classes/Core/Bootstrap.php(109): Neos\Flow\Core\Booting\Scripts::initializePackageManagement(Object(Neos\Flow\Core\Bootstrap))
#5 /data/Packages/Framework/Neos.Flow/Scripts/flo in /data/Packages/Framework/Neos.Flow/Classes/Package/PackageFactory.php on line 53
```

### Steps to Reproduce

1. run `composer install` with Flow < 7 and PHP 8 (older PHP Version set in composer.json under platform)
2. run `./flow help`

#### Expected behavior

Every thing works as expected. ;-)

#### Actual behavior

See error in description

### Affected Versions

Flow: < 7

PR #2640, opened 2021-12-01
BUGFIX: Fix PhpAnalyzer to support PHP 8

gerdemann:
When running composer with PHP 8, the PhpAnalyzer did not work properly.

Fixes: #2639

**Checklist**


sorenmalling:
@kitsunet You gave a comment about this in Slack, will you give some more details on it here, and comment on, whether or not this is a possible fix to the issue described?

sorenmalling:
@albe I've added you as a reviewer due to your work on PHP8 - can you perhaps say something about compatibility, this issue and corresponding PR? :)

albe:
This is basically a backport of the change here https://github.com/neos/flow-development-collection/pull/2287/files#diff-1cd00214dc173a34516d6f0aa372b178fc74317426c648d9487e9a28e15c9963 and it should be fully backwards compatible (with the `defined()` checks). The PhpAnalyzer afais is only used for package scanning and cache flushing. So this should not cause an issue and I don't see a problem with targeting it lower. However, is this targeting 5.3 on purpose, since that is security-only?

gerdemann:
I need it for a 5.3 project, but you're right, it's not a security fix 😅
I change the target tomorrow to 6.3.

albe (approved):
Yeah, okay for me :)

mficzel (approved):
Fine by reading

PR #2641, opened 2021-12-06
TASK: Update psalm baseline

markusguenther:


kdambekalns:
@markusguenther Was this created by you, locally, or via the GH action?

markusguenther:
Locally and it just works on my M1 machine ;)

kdambekalns:
> @markusguenther Was this created by you, locally, or via the GH action?

I am asking, since the memcache additions point to no memcache being available, which should not be the case on GH

markusguenther:
Yeah, it was more than expected :/

kdambekalns:
Ok, what you can do, is create the update and manually remove the unexpected changes. 😎

PR #2647, opened 2021-12-07
BUGFIX: Add correct changelog for 7.3.0

markusguenther:
The automatic changelog from the Jenkins job was just a GitHub comparison link for detailed information.
This adds the output from our changelog script.


markusguenther:
Thought that the Jenkins will trigger the same script and therefore has the same outcome.
That was my fault.

albe:
Just for the record: The original problem was just a wrong "previous version" when starting the job - it was given as "7.3.0" when it should be "7.2.0" so it created an empty changelog (7.3.0..7.3.0)

PR #2648, opened 2021-12-07
TASK: Fix step to automatically update psalm baseline on merge

albe:
🤞 still the force push will not work like that - see https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101/47 - we'd need a bot user with a personal access token and force-push rights that we make accessible to this workflow via GH secrets. That in turn means anyone with rights to edit this workflow can potentially force-push to protected branches via scripting.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -155,11 +155,13 @@ jobs:
           ./flow routing:list
 
       - name: Static analysis
+        id: psalm
         if: matrix.static-analysis == 'psalm'
+        continue-on-error: true
         run: composer test-static
 
       - name: Update psalm baseline
-        if: ${{ failure() && github.event_name == 'pull_request' && github.event.action == 'closed' && github.event.pull_request.merged == true }}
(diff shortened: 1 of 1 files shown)


markusguenther (approved):
Thanks for adjusting that :)
That will help in the future.

albe:
https://github.com/peter-evans/create-pull-request might be an alternative with a (questionably useful, given who reads/understands psalm baseline) review option

kdambekalns:
I am confused… will it work, or not?

albe:
> I am confused… will it work, or not?

Not as is. The push (`git push -f origin HEAD:${FLOW_TARGET_VERSION}`) will fail unless we provide a PAT with force push rights to this workflow. And then anyone updating this workflow can force push things. Still not sure if we do want that.

The goal of this is convenience in resolving psalm errors (semi-) automatically, so it might be okay. But even if there's a few things to consider:
- if it's fully automatic, this might hide actual coding issues by an eager merge of a PR (only admins can merge a PR with a failing step though)
- if it's semi automatic (with an automatic PR with the baseline update), this might create multiple such PRs from different feature PRs and then leads to a painful "merge - rebase/resolve - merge" chain of psalm baseline PRs that is even more work than doing a single update manually once

PR #2649, opened 2021-12-07
BUGFIX: Fix rendering of release notes

jonnitto:


jonnitto:
image

jonnitto:
What do we want to do with the older releases? Basically, we should adjust the older branches…

markusguenther (approved):
Thanks for taking care :)

sorenmalling (approved):

albe:
> What do we want to do with the older releases? Basically, we should adjust the older branches…

If someone feels like it, create PRs for those branches - changes don't need to be upmerged from there, only RTD needs to be rerendered.

PR #2654, opened 2021-12-21
BUGFIX: Wrong port assigned in applyTo()

kdambekalns:
This fixes the following:

- given an HTTPS connection to a proxy that passes the request handling to a server via HTTP
- given a shortcut node pointing to `http://www.acme.com`
- will result in `http://www.acme.com:443` leading to errors

This fixes it by using the (at this point already set!) scheme of the `$uri` to fill in the standard
port.

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
@@ -337,7 +337,7 @@ final class UriConstraints
                 $uri = $uri->withHost($baseUri->getHost());
             }
             if (empty($uri->getPort()) && !isset($this->constraints[self::CONSTRAINT_PORT])) {
-                $port = $baseUri->getPort() ?? UriHelper::getDefaultPortForScheme($baseUri->getScheme());
+                $port = $baseUri->getPort() ?? UriHelper::getDefaultPortForScheme($uri->getScheme());
                 $uri = $uri->withPort($port);
             }
         }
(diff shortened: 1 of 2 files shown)


kdambekalns:
Let's see if any test fails now… 😎

kdambekalns:
Ok, question… How on earth does that expectation make any sense?

```text

1) Neos\Flow\Tests\Unit\Mvc\Routing\Dto\UriConstraintsTest::applyToTests with data set #1 (array('https'), 'http://some-domain.tld', false, 'https://some-domain.tld:80/')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'https://some-domain.tld:80/'
+'https://some-domain.tld/'

/home/runner/work/flow-development-collection/flow-development-collection/flow-development-distribution/Packages/Framework/Neos.Flow/Tests/Unit/Mvc/Routing/Dto/UriConstraintsTest.php:101
```

I mean, if I really want to talk HTTPS to port 80, I better pass in the port explicitly. No?

kdambekalns:
For me that second dataset should be

```php
[
  'constraints' => [UriConstraints::CONSTRAINT_SCHEME => 'https'],
  'templateUri' => 'http://some-domain.tld',
  'forceAbsoluteUri' => false,
  'expectedUri' => 'https://some-domain.tld/'
]
```

albe (approved):
Hm, somehow makes sense...

sorenmalling (approved):

bwaidelich (approved):
Yes, that makes a lot of sense. Thanks for taking care!

PR #2656, opened 2022-01-06
!!!TASK: Make AbstractConditionViewHelper PHP 8.1 compatible

cvette:
Required parameters after optional parameters are deprecated since PHP 8.0

See https://php.watch/versions/8.0/deprecate-required-param-after-optional

Needs changes in the Neos project aswell.

**What I did**
Makes the AbstractConditionViewHelper compatible with PHP 8.1

**How I did it**
Changed order of parameters for the evaluateCondition method.

**How to verify it**
Run tests.

**Checklist**

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractConditionViewHelper.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/AbstractConditionViewHelper.php
@@ -71,7 +71,7 @@ abstract class AbstractConditionViewHelper extends AbstractViewHelper
      * @return boolean
      * @api
      */
-    protected static function evaluateCondition($arguments, RenderingContextInterface $renderingContext)
+    protected static function evaluateCondition(RenderingContextInterface $renderingContext, ?array $arguments)
     {
         return (boolean)$arguments['condition'];
     }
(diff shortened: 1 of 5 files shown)


mhsdesign:
It says `The deprecation does not apply if there is a type declared on the optional parameter, and the default value is null.`

so changing it to this would still be working and wouldnt break much right?

```protected static function evaluateCondition(array $arguments = null, RenderingContextInterface $renderingContext)```

as far as i know (at least on older php versions) one can still pass `null` to the above function...

**Comment by albe** [comment]
We should keep this code mostly in line with https://github.com/TYPO3/Fluid/blob/2.6/src/Core/ViewHelper/AbstractConditionViewHelper.php#L121
The typehint of the $arguments wasn't changed to avoid breaking 3rd party view helpers in Fluid:
https://github.com/TYPO3/Fluid/blob/2.6/src/Core/ViewHelper/AbstractConditionViewHelper.php#L108-L114

btw. the method did not specify the first argument as optional, so maybe I'm missing something?

Maybe we rather also add the `verdict()` method? /cc @kitsunet

**Review by mficzel — CHANGES_REQUESTED** [review]
I would rater define $arguments as non optional array
```
protected static function evaluateCondition(array $arguments, RenderingContextInterface $renderingContext)
```

can always be called like this:
```
static::evaluateCondition($this->arguments || [], $this->renderingContext)
```

mhsdesign:
not that i use fluid - so my review doesnt count too much ... 

but i would prefer @mficzel solution or a default value of `[]` (which is not deprecated!)

changing the Parameter order is not so nice for tge dev ^^

kdambekalns (changes requested):
This isn't even needed!?

This was fixed in https://github.com/neos/flow-development-collection/pull/2436 already and should work fine as of Flow 7.0.4, 7.1.0, 7.2.0 & 7.3.0 respectively.

Issue #2658, opened 2022-01-11
DynamicRoutePart only looks for a identifier from the persistencemanager


sorenmalling:
When passing a object to the UriBuilder, the DynamicRoutePart only looks for a value, via the PersistenceManager. So even if, the object had a `__toString` method, to represent ex. a DTO, this is never tried, if no identifier is found in the first place.

sorenmalling:
This is merged back in january 2022

PR #2659, opened 2022-01-11
TASK: Take __toString into account for resolving uri value

sorenmalling:
When resolving values for a route via DynamicRoutePart, object are only being looked up in the persistence manager.

With this change, we honour if a identifier is found, and if not, we look at the object to see, if it has a `__toString` method available, to give us a value.

This is backward compatible, since we respect the identifier, if given at first

Reolves #2658


bwaidelich:
IMO this should be a *BUGFIX* though (and thus target the `6.3` branch)

sorenmalling:
Let me update the PR :)

sorenmalling:
Closed in favor of https://github.com/neos/flow-development-collection/pull/2660 that target 6.3 as lowest

PR #2660, opened 2022-01-11
BUGFIX: Use __toString for resolving value in DynamicRoutePart

sorenmalling:
When resolving values for a route via DynamicRoutePart, object are only being looked up in the persistence manager.

With this change, we honour if a identifier is found, and if not, we look at the object to see, if it has a __toString method available, to give us a value.

This is backward compatible, since we respect the identifier, if given at first

Reolves #2658


bwaidelich (approved):
+1 by reading

albe (approved):
Looks good to me

PR #2666, opened 2022-01-16
BUGFIX: #2665 Eel arrow function, without arguments.

### LINKED ISSUE #2665: BUG: Eel arrow function, without arguments.
### Description
Eel arrow functions dont work with zero arguments like:
```
() => 'foo'
```
This line is failing: https://github.com/neos/flow-development-collection/blob/8a6154d3e11d7b229fa5c8b47d4650d1a7e7cd4f/Neos.Eel/Classes/CompilingEelParser.php#L323

### Steps to Reproduce
run this Fusion:
```
root = ${Array.map([''], () => "foo")[0]}
```
(idk for some reason it works in the fusionpen)

#### Expected behavior

output "foo"

#### Actual behavior

`Neos\Flow\Error\Exception: Notice: Undefined index: val in Data/Temporary/Testing/Cache/Code/Flow_Object_Classes/Neos_Eel_CompilingEelParser.php line 323`

### Affected Versions
Flow:
since at least 6.3

mhsdesign:
resolves #2665

Allows eel arrow functions to be declared with 0 arguments like in es6:

```
() => 'foo'
```

**What I did**
$args will be an empty array.

**Checklist**


jonnitto (approved):

albe (dismissed):
Looks too good to be true. Question though: what about ` public function ArrowFunction_arguments(&$result, $sub)`? Doesn't *that* need to be adjusted? tbh. this whole thing is a blackbox to me

mhsdesign:
@albe solved the mystery ^^ - i adjusted the - as i found out - poorly named 'MethodArguments' because it actually refers to the ArrowFunctionArguments - and is used in no other place.

(the `_arguments` suffix is autogenerated by peg)

```
ArrowFunction: arguments:MethodArguments < '=>' > exp:Expression
```

see orginial pr:

https://github.com/neos/flow-development-collection/pull/1276

albe:
Good dig! Didn't remember that one. But that brings up the next question:

```
    public function MethodArgumentsWithoutParens_Identifier(&$result, $sub)
    {
        $result['val'] = [$sub['text']];
    }
    public function MethodArgumentsWithParens_Identifier(&$result, $sub)
    {
        if (!isset($result['val'])) {
            $result['val'] = [];
        }
        $result['val'][] = $sub['text'];
    }

    public function MethodArguments_arguments(&$result, $sub)
    {
        $result['val'] = $sub['val'] ?? [];
    }

    public function ArrowFunction_arguments(&$result, $sub)
    {
        $result['args'] = $sub['val'];
    }
 ```
So if I understand it correctly now, depending on if the parser finds parens, it will invoke `MethodArgumentsWithoutParens_Identifier` or `MethodArgumentsWithParens_Identifier`, with the parsed identifier as `$sub`. Now if that isn't present in the WithParens case, `$sub['text']` would be null or unset, but apart from the generated NOTICE/WARNING, `$result['val']` would be an array with a single `null` element. Then that would be passed into `MethodArguments_arguments()`, which tries to null coalesce the 'val' but why does this work? An `[null]` is not coalesced anyway so the coalescing doesn't change the value for this case, does it?

But the error in the issue (`Neos\Flow\Error\Exception: Notice: Undefined index: val in Data/Temporary/Testing/Cache/Code/Flow_Object_Classes/Neos_Eel_CompilingEelParser.php line 323`) suggests something else is going on in between. So maybe then this is just fine as is 🤷‍♂️ I'm still staying with my approval

kitsunet (changes requested):
We can't just change the Parser, the Grammar file needs to be chnaged and the parser generted by the script otherwise it will be out of sync and stuff breaks on other changes.

PR #2669, opened 2022-01-19
BUGFIX: Avoid race condition on symlink publishing

### LINKED ISSUE #2667: Possible race condition in FileSystemSymlinkTarget
### Description

When `publishFile()` is called for the same target concurrently, the code may run into a race condition.

### Steps to Reproduce

1. clear thumbnails
2. open the media browser in mutiple tabs at the same time

#### Expected behavior

Nothing (bad) happens.

#### Actual behavior

`Could not publish "…" into resource publishing target "localWebDirectoryPersistentResourcesTarget" because the source file could not be symlinked at target location`

### Affected Versions

Flow: 6.3+

kdambekalns:
If the symlink could not be created but exists, check if it points to
the expected target and ignore the error in that case.

Fixes #2667

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
+++ b/Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
@@ -76,27 +76,9 @@ class FileSystemSymlinkTarget extends FileSystemTarget
             throw new TargetException(sprintf('Could not publish "%s" into resource publishing target "%s" because the source file is not accessible (file stat failed).', $sourcePathAndFilename, $this->name), 1415716366);
         }
 
-        if (!file_exists(dirname($targetPathAndFilename))) {
-            Files::createDirectoryRecursively(dirname($targetPathAndFilename));
-        }
-
-        try {
-            if (Files::is_link($targetPathAndFilename)) {
-                Files::unlink($targetPathAndFilename);
-            }
-
-            if ($this->relativeSymlinks) {
-                $result = Files::createRelativeSymlink($sourcePathAndFilename, $targetPathAndFilename);
-            } else {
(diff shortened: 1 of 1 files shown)


kdambekalns:
ping…

kdambekalns:
ping …

kdambekalns:
ping …

robertlemke (approved):
👍🏻

Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
```diff
@@ -183,4 +148,36 @@ protected function setOption($key, $value)
...
+            } else {
+                $temporaryTargetPathAndFilename = $targetPathAndFilename . '.' . Algorithms::generateRandomString(13) . '.tmp';
+                symlink($sourcePathAndFilename, $temporaryTargetPathAndFilename);
+                $result = rename($temporaryTargetPathAndFilename, $targetPathAndFilename);
+            }
+        } catch (\Exception $exception) {
+            $result = false;
+        }
+
+        if (Files::is_link($targetPathAndFilename) && realpath($targetPathAndFilename) === realpath($sourcePathAndFilename)) {
```
mhsdesign, line 176: can there be a situation where both `realpath` return `false`? (since `false === false`)
i only know that `realpath` doesnt like stream wrapper and returns false then ...

kdambekalns: Oh, well. Technically, yes, that could happen. But I think we can ignore that here.

mhsdesign: 👍

mhsdesign (approved):

mhsdesign:
btw i do noticed the odd exception handling and returning arrays and stuff, but fixing this is not a bugfix.

PR #2671, opened 2022-01-19
BUGFIX: Clear stat cache when checking for directory

### LINKED ISSUE #2670: Possible race condition with mkdir
### Description

Trying to make the same directory concurrently runs into a "notice turned logged exception"

### Steps to Reproduce

1. Use `createDirectoryRecursively()` on the same path concurrently

#### Expected behavior

Nothing (bad)

#### Actual behavior

An error is logged, even though everything is in fact fine:

`Exception #1: Warning: mkdir(): File exists in /application/Packages/Framework/Neos.Utility.Files/Classes/Files.php line 251`

### Affected Versions

Flow: 6.3+

kdambekalns:
Makes sure the check does not use potentially outdated information.

Fixes #2670

### DIFF
--- a/Neos.Utility.Files/Classes/Files.php
+++ b/Neos.Utility.Files/Classes/Files.php
@@ -243,10 +243,14 @@ abstract class Files
         if (substr($path, -2) === '/.') {
             $path = substr($path, 0, -1);
         }
+        if ($path === '') {
+            return;
+        }
+        clearstatcache(true, $path);
         if (is_file($path)) {
             throw new FilesException('Could not create directory "' . $path . '", because a file with that name exists!', 1349340620);
         }
-        if (!is_link($path) && !is_dir($path) && $path !== '') {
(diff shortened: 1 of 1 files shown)


jonnitto (approved):

Neos.Utility.Files/Classes/Files.php
```diff
@@ -246,7 +246,8 @@ public static function createDirectoryRecursively(string $path)
         if (is_file($path)) {
             throw new FilesException('Could not create directory "' . $path . '", because a file with that name exists!', 1349340620);
         }
-        if (!is_link($path) && !is_dir($path) && $path !== '') {
+        clearstatcache(true, $path);
```
albe, line 249: Shouldn't this be moved before the `is_file($path)` check above?

kdambekalns: Indeed, that would make sense.

Issue #2673, opened 2022-01-19
Add missing doctrine/annotations


markusguenther:
### Description

For some reason the doctrine annotations are missing but we are using the in the booting script.
Since doctrine common 2.4 it is not part of the package anymore.

### Steps to Reproduce

1. Install the flow-development-distribution with branch 7.0 for instance
2. composer install
3. run `./flow help`

#### Expected behavior

See the flow help

#### Actual behavior

<img width="1330" alt="Screenshot 2022-01-19 at 15 53 32" src="https://user-images.githubusercontent.com/1014126/150155724-23f401ab-dd1c-4d77-90a8-a7f792892c90.png">

### Affected Versions

Flow: 7.0

You can get this information by running `composer show`.

PR #2674, opened 2022-01-19
BUGFIX: Add doctrine/annotation as dependency

markusguenther:
Prevents exception in booting flow, as it is requiring the 'Doctrine\Common\Annotations\AnnotationRegistry'

fixes: #2673


kdambekalns (approved):

kdambekalns:
Merged, as the failing unit test has another cause.

PR #2676, opened 2022-01-20
BUGFIX: Fix UriConstraints.applyTo() port handling

kdambekalns:


bwaidelich (approved):
beautiful! :)

Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
```diff
@@ -358,7 +358,7 @@ public function applyTo(UriInterface $baseUri, bool $forceAbsoluteUri): UriInter
                 $uri = $uri->withHost($host);
             }
         }
-        if (isset($this->constraints[self::CONSTRAINT_PORT]) && ($forceAbsoluteUri || $this->constraints[self::CONSTRAINT_PORT] !== $baseUri->getPort()) && ($baseUri->getPort() !== null || $this->constraints[self::CONSTRAINT_PORT] !== UriHelper::getDefaultPortForScheme($baseUri->getScheme()))) {
+        if (isset($this->constraints[self::CONSTRAINT_PORT]) && ($forceAbsoluteUri || $this->constraints[self::CONSTRAINT_PORT] !== $baseUri->getPort()) && ($baseUri->getPort() !== null || $this->constraints[self::CONSTRAINT_PORT] !== UriHelper::getDefaultPortForScheme($this->constraints[self::CONSTRAINT_SCHEME] ?? $baseUri->getScheme()))) {
```
albe, line 361:
```suggestion
        if (isset($this->constraints[self::CONSTRAINT_PORT]) && 
        ($forceAbsoluteUri || $this->constraints[self::CONSTRAINT_PORT] !== $baseUri->getPort()) && 
        ($baseUri->getPort() !== null || $this->constraints[self::CONSTRAINT_PORT] !== UriHelper::getDefaultPortForScheme($uri->getScheme()))) {
```
Since the scheme constraint is already applied to the new URI this would be more straightforward maybe and be analogous to https://github.com/neos/flow-development-collection/pull/2676/files#diff-8df45416161efdbe2035f01361dc8245e81a126fc892f1497ba972acb07738bcR407?
I see it as "apply constraints to the new URI in a specific order and depending on previous applied constraints"

bwaidelich: > Since the scheme constraint is already applied to the new URI

Only if the scheme constraint does not _match_ the scheme of the base URI

kdambekalns: It's only applied if different, so this would not be working here. Tried…

albe (approved):
Tests pass again, thanks! Left a tiny suggestion but not urgent

markusguenther:
Thanks @kdambekalns, for dive deeper and fix that issue :)

PR #2678, opened 2022-01-20
BUGFIX: Correct check on cache file write success

### LINKED ISSUE #2677: Wrong success check in cache FileBackend.writeCacheFile()
### Description

When the same session is used concurrently, sometimes the session seemingly ends.

The log then shows an exception about `unserialize()` not being able to process the value retrieved from the cache backend in https://github.com/neos/flow-development-collection/blob/cb92ea7f5e979647277a8111aec4b3cfbccf144f/Neos.Cache/Classes/Frontend/VariableFrontend.php#L94

The reason in this case is an incomplete data file when using the `FileBackend` (other backends are probably not affected) which is weird, because our cache writes should be atomic.

### Steps to Reproduce

1. Have concurrent requests using a session
2. You will eventually observe #2672
3. You conclude this should never happen due to writes being atomic

#### Expected behavior

A cache file is either written or not.

#### Actual behavior

Incomplete cache files are written.

#### Reason

Maybe this line is wrong?

https://github.com/neos/flow-development-collection/blob/cb92ea7f5e979647277a8111aec4b3cfbccf144f/Neos.Cache/Classes/Backend/SimpleFileBackend.php#L541

`$result` is always `true` afterwards, which certainly is not intended… that has been changed in https://github.com/neos/flow-development-collection/commit/e80c5a958f2c5e436138cb130a4a9cb46153845b#diff-81c872d2429d48a1767632c0255f66f1281ec7a1a339b2f715b9af72322409f4L541-R541

### Affected Versions

Flow: 6.3+

kdambekalns:
Fixes #2677

### DIFF
--- a/Neos.Cache/Classes/Backend/SimpleFileBackend.php
+++ b/Neos.Cache/Classes/Backend/SimpleFileBackend.php
@@ -528,9 +528,11 @@ class SimpleFileBackend extends IndependentAbstractBackend implements PhpCapable
      */
     protected function writeCacheFile(string $cacheEntryPathAndFilename, string $data): bool
     {
+        // Once vfs is fixed for file_put_contents with LOCK_EX,
+        // see https://github.com/mikey179/vfsStream/wiki/Known-Issues
+        // this can be replaced by a simple:
+        // return !(file_put_contents($cacheEntryPathAndFilename, $data, LOCK_EX) === false);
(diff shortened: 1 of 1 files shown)


albe (approved):
Oh my, indeed totally bogus check there!

daniellienert (approved):

PR #2679, opened 2022-01-25
TASK: Drop DBAL reconnect code

kdambekalns:
Since DBAL 2.11.0 retry and reconnecting lost connections happens
automatically, ping() will be removed in DBAL 3, see the PR at
https://github.com/doctrine/dbal/pull/4119

Thus this code is dropped, as since 7.0.0 we require a newer DBAL
version.

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/AllowedObjectsListener.php
@@ -13,15 +13,11 @@ namespace Neos\Flow\Persistence\Doctrine;
  * source code.
  */
 
-use Doctrine\DBAL\Exception\ConnectionException;
 use Doctrine\ORM\Event\OnFlushEventArgs;
 use Neos\Flow\Annotations as Flow;
-use Neos\Flow\Log\ThrowableStorageInterface;
-use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\Persistence\AllowedObjectsContainer;
 use Neos\Flow\Persistence\Exception as PersistenceException;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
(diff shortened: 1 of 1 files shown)


robertlemke:
Makes sense. But I guess we also need to check further code if there are meaningful ways to handle `ConnectionLost` exceptions and trying to re-run queries – or?

kdambekalns:
> Makes sense. But I guess we also need to check further code if there are meaningful ways to handle `ConnectionLost` exceptions and trying to re-run queries – or?

If the reconnect is attempted automagically now, the behaviour is like before. We retried once, then failed. So… not sure we need to do anything.

albe:
> > Correct me if I'm wrong, but if I understand this change correctly, the difference is that the connection gets closed properly if the connection is lost. But it doesn't execute the query again, in such a case. So it only ensures that the next query succeeds again.
> >
> Sounds correct.

From what I read and understand so far I think the difference is, that until now we preemptively try to reestablish a connection before running the actual database queries, while with the new automatic reconnect it only means that if the first query fails due to lost connection, it tries to reestablish the connection and go on with the next query (or transaction, which in the worst case just means all persistence for this request is lost). Doctrine will not retry a failed query. Which makes sense from their PoV as they know absolutely nothing about the nature of the query and if retrying it is safe.

albe:
See https://github.com/neos/flow-development-collection/pull/2637 for an alternative

kdambekalns:
Indeed… I dug a bit deeper, and the deprecation message on `ping()` saying

    Retry and reconnecting lost connections now happens automatically, ping() will be removed in DBAL 3

is plain wrong, because `Connection` simply does this:

```php
    private function throw(Exception $e): void
    {
        if ($e instanceof ConnectionLost) {
            $this->close();
        }

        throw $e;
    }
```

and then (probably) tries to reconnect for the next statement. Doh.

PR #2683, opened 2022-01-27
TASK: Move test base classes into separate package that is required by flow as dev dependency

mficzel:
Extract the base classes for tests from flow into a separate package to be required as dev dependency. 

Background: This is needed to have custom test case base classes but not have PHPUnit as a dependency. The solution is to add the test case base classes in a separate package that is dev-dependency. 

Eventually, this should allow getting rid of the custom autoloader in testing here: https://github.com/neos/BuildEssentials/blob/master/PhpUnit/UnitTestBootstrap.php#L41-L65


mficzel:
Currently most unit tests work already but the following points should be discussed.

- Is this worth the hassle?
- How should the testing Package be named
- Should the new packaghe provide the Neos/Flow/Tests namespace or a different one

mhsdesign:
would it make using phpunit per cli easier? i mean we still need the bootstrap file right?

but if makes things for the beginner dev easier id like that ^^

mficzel:
I doubt we can get rid of the bootstrap entirely. So this is probably not worth the hassle. At least for now.

mhsdesign:
id like to reopen this - as this is the correct way i imagine and sometime we will choose it (just not jet atm) and then it will provide a good start ;)

mficzel:
I close this in favor of #3105

PR #2685, opened 2022-01-27
TASK: Return 400 response when HMAC validation fails

### LINKED ISSUE #2681: Failing HMAC validation leads to log flooding
### Description

When a request comes in and the HMAC validation fails, this causes a `CRITICAL` log message, as the resulting exception is not handled.

### Steps to Reproduce

1. have a website with some form to submit
2. wait for a spammer / security scan
3. marvel at the logs being flooded with `Exception #1320830018 in line 99 of /application/Packages/Framework/Neos.Flow/Classes/ObjectManagement/DependencyInjection/DependencyProxy.php: The given string was not appended with a valid HMAC.`

#### Expected behavior

Since this is hardly a critical issue, it should not cause `CRITICAL` log entries. `NOTICE` seems fine. And in terms of HTTP status this is clearly not a 500, but a 400.

#### Actual behavior

A lot of `CRITICAL` log messages and a response status of 500.

### Affected Versions

Neos: 5.3+
Flow: 6.3+

kdambekalns:
When HMAC validation fails during request processing, the controller
will return a response with a status code of 400, as that is caused
by a bad request. The exception is logged with a notice to the log,
to aid in debugging errors

Previously the uncaught exception would cause a status 500 response
and log a critical error.

Fixes #2681

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/ActionController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/ActionController.php
@@ -154,10 +154,9 @@ class ActionController extends AbstractController
     protected $logger;
 
     /**
-     * @Flow\Inject
      * @var ThrowableStorageInterface
      */
-    protected $throwableStorage;
+    private $throwableStorage;
 
     /**
      * @param array $settings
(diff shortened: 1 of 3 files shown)


albe (approved):
Sounds and looks absolutely reasonable

kdambekalns:
ping…

PR #2686, opened 2022-01-27
TASK: Return 400 response if required argument is missing

### LINKED ISSUE #2682: Missing required argument leads to log flooding
### Description

When a request comes in and a required controller argument is missing, this causes a `CRITICAL` log message, as the resulting exception is not handled.

### Steps to Reproduce

1. have a website with some form to submit
2. wait for a spammer / security scan
3. marvel at the logs being flooded with `Exception #1298012500 in line 202 of /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Mvc_Controller_ActionController.php: Required argument "something" is not set.`

#### Expected behavior

Since this is hardly a critical issue, it should not cause `CRITICAL` log entries. `NOTICE` seems fine. And in terms of HTTP status this is clearly not a 500, but a 400.

#### Actual behavior

A lot of `CRITICAL` log messages and a response status of 500.

### Affected Versions

Neos: 5.3+
Flow: 6.3+

kdambekalns:
When a required argument is missing request processing, the controller
will return a response with a status code of 400, as that is caused
by a bad request. The exception is logged with a notice to the log,
to aid in debugging errors.

Previously the uncaught exception would cause a status 500 response
and log a critical error.

Fixes #2682

### DIFF
--- a/Neos.Flow/Classes/Mvc/Controller/ActionController.php
+++ b/Neos.Flow/Classes/Mvc/Controller/ActionController.php
@@ -15,6 +15,7 @@ use Neos\Error\Messages\Result;
 use Neos\Flow\Annotations as Flow;
 use Neos\Error\Messages as Error;
 use Neos\Flow\Http\Component\ReplaceHttpResponseComponent;
+use Neos\Flow\Log\ThrowableStorageInterface;
 use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\Mvc\ActionRequest;
 use Neos\Flow\Mvc\ActionResponse;
@@ -22,6 +23,7 @@ use Neos\Flow\Mvc\Exception\ForwardException;
 use Neos\Flow\Mvc\Exception\InvalidActionVisibilityException;
 use Neos\Flow\Mvc\Exception\InvalidArgumentTypeException;
 use Neos\Flow\Mvc\Exception\NoSuchActionException;
+use Neos\Flow\Mvc\Exception\RequiredArgumentMissingException;
 use Neos\Flow\Mvc\Exception\UnsupportedRequestTypeException;
 use Neos\Flow\Mvc\Exception\ViewNotFoundException;
 use Neos\Flow\Mvc\View\ViewInterface;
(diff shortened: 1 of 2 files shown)


kdambekalns:
ping…

kdambekalns:
ping …

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -225,7 +226,13 @@ public function processRequest(ActionRequest $request, ActionResponse $response)
             $this->throwStatus(400, '400 Bad Request', 'Invalid HMAC submitted');
         }
 
-        $this->mapRequestArgumentsToControllerArguments();
+        try {
+            $this->mapRequestArgumentsToControllerArguments();
+        } catch (RequiredArgumentMissingException $e) {
+            $message = $this->throwableStorage->logThrowable($e);
+            $this->logger->notice('Request argument mapping failed due to a missing required argument. ' . $message, LogEnvironment::fromMethodName(__METHOD__));
+            $this->throwStatus(400, '400 Bad Request', 'Required argument is missing');
```
robertlemke, line 234: The second argument `400 Bad Request` won't be used by `throwStatus()` because you specified `$content` (the third argument). Rather skip the third argument and use `Required argument is missing` (sic!) as the second parameter.

robertlemke (changes requested):

Neos.Flow/Classes/Mvc/Controller/ActionController.php
kdambekalns: Doh. Then why do the tests pass?

kdambekalns: Ah, I see. The 2nd argument is superflous the way I use it. I find that method behaviour very weird, I have to admit.

kdambekalns: I'd rather skip the second argument, because the content being "400 …" seems weird.

robertlemke (approved):

mhsdesign (approved):
fine by reading ;)

PR #2688, opened 2022-02-01
!!! FEATURE: Implement Psr-18 HTTP\Client

mficzel:
The `Neos\Flow\Http\Client\Browser` now implements the `Psr\Http\Client\ClientInterface` and works with dependency injection by default.

The signature of the method sendRequest is relaxed slightly from ServerRequestInterface to RequestInterface and the curl requestEngine is injected by default. That way any browser is ready for use from the start.

Internally the request engines now also use the RequestInterface instead of ServerRequestInterface.
This should be almost non breaky because changes ServerRequestInterface is still allowed as it extends the RequestInterface. The InternalRequestEngide will upcast the RequestInterfaces to ServerRequestInterface when needed.

!!!The PR removes the @api annotation from the requestEngines to allow us to remove or alter those in the future.

Resolves: #2691 

**Checklist**

- [x] Docs are adjusted

### DIFF
--- a/Neos.Flow/Classes/Http/Client/InternalRequestEngine.php
+++ b/Neos.Flow/Classes/Http/Client/InternalRequestEngine.php
@@ -30,6 +30,8 @@ use Neos\Flow\Validation\ValidatorResolver;
 use Psr\Http\Message\ResponseFactoryInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ServerRequestFactoryInterface;
+use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\StreamFactoryInterface;
 
 /**
@@ -84,6 +86,12 @@ class InternalRequestEngine implements RequestEngineInterface
 
     /**
      * @Flow\Inject
+     * @var ServerRequestFactoryInterface
+     */
+    protected $serverRequestFactory;
+
+    /**
+     * @Flow\Inject
      * @var ResponseFactoryInterface
      */
     protected $responseFactory;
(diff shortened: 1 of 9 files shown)


mficzel:
It would be great to have a way to configure requestEngines and options as default so the Browser or even the ClientInterface could be simply injected.

sorenmalling:
> It would be great to have a way to configure requestEngines and options as default so the Browser or even the ClientInterface could be simply injected.

I would say, that is possible with Objects.yaml configuration as of today

mficzel:
@sorenmalling no because the request engine is not injected but is required

mficzel:
It may also make sense to require guzzle in the HttpFactories and add a ClientFactory there even thogh there is no ClientFactoryInterface yet in PSR

albe:
Good move, but I think there are still some deeper issues that couldn't be resolved easily in the last attempt, no?
https://github.com/neos/flow-development-collection/pull/1791#discussion_r327615908
and
https://github.com/neos/flow-development-collection/pull/1791#discussion_r327618228

Can you maybe have a look @mficzel ? Maybe I'm missing something. If you want we can schedule a call sometime this week

albe (approved):
Indeed the issues raised in the PR by Karsten do not really affect things here, because all methods can still receive a ServerRequestInterface and as long as you don't use an InternRequestEngine configured Browser to send external requests things still work. For the case when the InternalRequestEngine is used to send a (non-Server) `RequestInterface`, an upcast is now done.

Neos.Flow/Classes/Http/Client/InternalRequestEngine.php
```diff
@@ -115,6 +135,10 @@ public function sendRequest(ServerRequestInterface $httpRequest): ResponseInterf
         $this->validatorResolver->reset();
 
         $objectManager = $this->bootstrap->getObjectManager();
+
+        /**
+         * @var Http\Middleware\MiddlewaresChain $middlewaresChain
+         */
```
kdambekalns, line 141: Funny… I don't need this in PhpStorm, the type of `$middlewaresChain` is still known… 🤔

kdambekalns (approved):
Looks good by reading.

Doing the upcast as needed looks like a clever way to overcome the issue I had with my try back then…

mhsdesign:
btw there are now failing tests:

mficzel:
@mhsdesign fix is already on the way https://github.com/neos/flow-development-collection/pull/2749

PR #2689, opened 2022-02-01
FEATURE: Add methods getSimpleCache PSR-16 and getCacheItemPool PSR-6 to the cacheManager

mficzel:
The methods create PSR Frontends for the cache with the given identifier. This allows to configure psr caches via
objects and caches yaml. This is an improvement as the included factories are not that easy to use.

The following settings in Objects.yaml allow to inject PSR Caches into Flow Classes:
```
Vendor\Site\Service:

  properties:
    simpleCacheProperty:
      object:
        factoryObjectName: Neos\Flow\Cache\CacheManager
        factoryMethodName: getSimpleCache
        arguments:
          1:
            value: Cache_Identifier_1

    cachePoolProperty:
      object:
        factoryObjectName: Neos\Flow\Cache\CacheManager
        factoryMethodName: getCacheItemPool
        arguments:
          1:
            value: Cache_Identifier_2
```

!!! While possible it is not advisible to access the same cache with multiple Interfaces as the storage formats may differ. !!!

Resolves: #2690 

**Checklist**

- [x] Docs are adjusted

### DIFF
--- a/Neos.Flow/Classes/Cache/CacheManager.php
+++ b/Neos.Flow/Classes/Cache/CacheManager.php
@@ -22,7 +22,11 @@ use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\Utility\Environment;
 use Neos\Utility\Files;
 use Neos\Flow\Utility\PhpAnalyzer;
+use Neos\Cache\Psr\Cache\CachePool;
+use Neos\Cache\Psr\SimpleCache\SimpleCache;
 use Psr\Log\LoggerInterface;
+use Psr\SimpleCache\CacheInterface;
+use Psr\Cache\CacheItemPoolInterface;
 
 /**
  * The Cache Manager
@@ -53,9 +57,19 @@ class CacheManager
     protected $environment;
 
     /**
-     * @var array
+     * @var FrontendInterface[]
      */
     protected $caches = [];
+
+    /**
+     * @var CacheInterface[]
+     */
+    protected $simpleCaches = [];
+
+    /**
+     * @var CacheItemPoolInterface[]
+     */
+    protected $cacheItemPools = [];
 
     /**
      * @var array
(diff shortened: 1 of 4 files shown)


sorenmalling:
@mficzel Can you create a corresponding issue that this PR can link to?

sorenmalling:
How is this different from the possibilities today of injecting a configured cache?

mficzel:
@sorenmalling it allows to get instances of the psr Cache classes. Those are different from flowCache frontends but external libraries likely need those for caching.

mficzel:
@kitsunet I remember us discussing how the psr caches should be instantiated back then. I think back then you planned something similar but for some reason we decided to only provide the psr factories back then.

Had some use cases in the meantime where i would have liked the option to inject those cache-interfaces.

albe:
Just for reference, with the PSR factories of `Neos\Cache\Psr` one could inject a cache, but needs to define the backend class and options everytime. So the benefit is primarily that our existing cache configurations can be used to configure PSR caches, right?

> !!! While possible it is not advisible to access the same cache with multiple Interfaces. !!!

Yeah, that's really a pandora box waiting to be opened. I see no good way around that though

albe (approved):
Generally approve. It's nice to be able to configure PSR caches and inject them like Flow caches.

sorenmalling:
Can the documentation be extended with a explanation on why it's not a good thing to access the same vacate with multiple interfaces?

mficzel:
@sorenmalling changed
> !!! While possible it is not advisible to access the same cache with multiple Interfaces. !!!
to 
> !!! While possible it is not advisible to access the same cache with multiple Interfaces as the storage formats may differ. !!!
As already was in the docs

kdambekalns (approved):
Looks good by reading.

I guess people will shoot themselves in the foot by accessing the same cache from two sides, but… 🤷‍♂️

PR #2695, opened 2022-02-06
Make adjustments to support PHP8.1's stricter type system

nezaniel:
This adds additional type hints where required in PHP 8.1 while maintaining support for 7.4.
These are only the adjustments required fir the ESCR test suite, more are to follow

### DIFF
--- a/Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
+++ b/Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
@@ -104,7 +104,7 @@ trait SecurityOperationsTrait
      */
     public static function cleanUpSecurity()
     {
-        if (file_exists(self::$testingPolicyPathAndFilename)) {
+        if (self::$testingPolicyPathAndFilename && file_exists(self::$testingPolicyPathAndFilename)) {
             unlink(self::$testingPolicyPathAndFilename);
         }
     }
--- a/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
@@ -598,7 +598,7 @@ class FlowAnnotationDriver implements DoctrineMappingDriverInterface, PointcutFi
             $mapping['fieldName'] = $property->getName();
             $mapping['columnName'] = strtolower($property->getName());
             $mapping['targetEntity'] = $propertyMetaData['type'];
-            $mapping['nullable'] = $propertyMetaData['nullable'];
+            $mapping['nullable'] = $propertyMetaData['nullable'] ?? false;
 
             $joinColumns = $this->evaluateJoinColumnAnnotations($property);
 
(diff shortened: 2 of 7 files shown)


albe (approved):
Looks good by reading, but I agree having 8.1 builds run would be good

nezaniel:
8.1 builds are configured and run through smoothly, which leaves me a bit sceptical since that seems improbable. Is there more to be configured or is everything fine already?

albe:
> 8.1 builds are configured and run through smoothly, which leaves me a bit sceptical since that seems improbable. Is there more to be configured or is everything fine already?

> ==> Setup PHP
✓ PHP Found PHP 8.1.2

Looks good to me. Are there any specific reasons why you think running Flow on 8.1 should not work with this PR applied @nezaniel ? Are there any breaking changes in 8.1 that we should be subject to?
I could very well imagine Flow being okay and Neos having a few more edge cases from a gut feeling

mficzel:
i somehow miss the psalm test in the matrix now. Should offcourse not run on 7.3 but on 8.0 i think.

nezaniel:
@albe Well, I had to un-proxy literally every ESCR exception class for it to not throw a __wakeup() signature notice, so I expected this to happen all over the Flow code base... but maybe the ESCR test setup is a little more strict than the regular tests and PHP runtimes

As far as I could see, the psalm tests were configured for 7.3 only, so I disabled them

Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php
```diff
@@ -104,7 +104,7 @@ public function iHaveTheFollowingPolicies($string)
      */
     public static function cleanUpSecurity()
     {
-        if (file_exists(self::$testingPolicyPathAndFilename)) {
+        if (self::$testingPolicyPathAndFilename && file_exists(self::$testingPolicyPathAndFilename)) {
```
bwaidelich, line 107: nitpick: This is a "falsy" value but we prefer strict comparison according to the [CGL](https://neos.readthedocs.io/en/stable/References/CodingGuideLines/PHP.html#php-in-general)
```suggestion
        if (self::$testingPolicyPathAndFilename !== null && file_exists(self::$testingPolicyPathAndFilename)) {
```

.github/workflows/build.yml
```diff
@@ -16,16 +16,16 @@ jobs:
...
-        include:
-          - php-versions: '7.3'
-            static-analysis: 'psalm'
-            experimental: true
-            dependencies: 'highest'
+        #include:
+          #- php-versions: '7.3'
+          #  static-analysis: 'psalm'
+          #  experimental: true
+          #  dependencies: 'highest'
```
bwaidelich, line 28: I'd say: remove those lines.. If we want them back, we can always look them up in the git history

bwaidelich (approved):
Looking good in general, I just added to nitpicky comments

.github/workflows/build.yml
nezaniel: Since iirc the psalm checks failed before anyway, I'd like to leave them out for now, as I'd like this merged independently of us getting psalm back to life. Also, I never worked with psalm before, so I won't be able to help here

albe: Okay, we'll deal with that in #2704

PR #2697, opened 2022-02-06
BUGFIX: Web-Exception messages preserve whitespace (ASCII Art)

### LINKED ISSUE #2696: BUG: Web-Exception messages, preserve whitespace (ASCII Art)
before:
image

after:
image

Exceptions which have multiple spaces in their $message where previously not as expected displayed.
Exceptions outputted to the CLI or logged did preserve multiple whitespaces naturally, but since they are collapsed in HTML by default, they are not shown in the browser.

It was not possible to really work around this since the htmlspecialchars() would disarm an injected `<pre>` tag and using nonspeaking spaces as unicode "\u{00A0}" where logged unreadable and without monofont, the spaces don't take up as much space like expected in the CLI or log.

Displaying all exceptions in pre is something which could potentially harm readability.
so 'normal' messages without multiple following spaces should not be changed.

mhsdesign:
Closes #2696

Exceptions which have multiple spaces in their $message where previously not as expected displayed.
Exceptions outputted to the CLI or logged did preserve multiple whitespaces naturally, but since they are collapsed in HTML by default, they are not shown in the browser.

### DIFF
--- a/Neos.Flow/Classes/Error/DebugExceptionHandler.php
+++ b/Neos.Flow/Classes/Error/DebugExceptionHandler.php
@@ -96,11 +96,17 @@ EOD;
         while (true) {
             $filepaths = Debugger::findProxyAndShortFilePath($exception->getFile());
             $filePathAndName = $filepaths['proxy'] !== '' ? $filepaths['proxy'] : $filepaths['short'];
-            $exceptionMessageParts = $this->splitExceptionMessage($exception->getMessage());
 
-            $exceptionHeader .= '<h1 class="ExceptionSubject">' . htmlspecialchars($exceptionMessageParts['subject']) . '</h1>';
-            if ($exceptionMessageParts['body'] !== '') {
-                $exceptionHeader .= '<p class="ExceptionBody">' . nl2br(htmlspecialchars($exceptionMessageParts['body'])) . '</p>';
+            ['subject' => $exceptionMessageSubject, 'body' => $exceptionMessageBody] = $this->splitExceptionMessage($exception->getMessage());
+
+            $exceptionHeader .= '<h1 class="ExceptionSubject">' . htmlspecialchars($exceptionMessageSubject) . '</h1>';
+            if ($exceptionMessageBody !== '') {
+                if (str_contains($exceptionMessageBody, '  ')) {
+                    // contents with multiple spaces will be pre-served
+                    $exceptionHeader .= '<p class="ExceptionBodyPre">' . htmlspecialchars($exceptionMessageBody) . '</p>';
+                } else {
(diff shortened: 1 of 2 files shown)


albe (approved):
Generally approve, but should we maybe just always use `pre` formatting for the body?

mhsdesign:
hmm could potentially harm readability - what do you think? But is also easy to change later again when its not ideal.

Neos.Flow/Classes/Error/DebugExceptionHandler.php
```diff
@@ -90,11 +90,19 @@ protected function renderStatically(int $statusCode, \Throwable $exception)
...
-            if ($exceptionMessageParts['body'] !== '') {
-                $exceptionHeader .= '<p class="ExceptionBody">' . nl2br(htmlspecialchars($exceptionMessageParts['body'])) . '</p>';
+            [
+                'subject' => $exceptionMessagePartSubject,
+                'body' => $exceptionMessagePartBody,
+            ] = $this->splitExceptionMessage($exception->getMessage());
+
+            $exceptionHeader .= '<h1 class="ExceptionSubject">' . htmlspecialchars($exceptionMessagePartSubject) . '</h1>';
+            if ($exceptionMessagePartBody !== '') {
+                if (strpos($exceptionMessagePartBody, '  ') !== false) {
```
bwaidelich, line 101: ooh that's hacky ;)
I'd go for a unified rendering, too.

bwaidelich (dismissed):
I think it it is a bit flaky to change the formatting based on some characters in the exception message.. So +1 with what Alex wrote (given that we can keep it readable for both worlds)

bwaidelich:
If using monospace font and `white-space: pre` in general leads to problems with readability, we could maybe introduce another (optional) property to the Flow exception base class that allows to render formatted strings (similar to `statusCode`)

mhsdesign:
i thought about additional optional properties on the exception too - but we took already a way different way - the 'a let flow handle this all automatic way':
for example the exception message is splitt by some hacky regex into head and body:
here and this is not configureable:
https://github.com/neos/flow-development-collection/blob/0a19f923d959f6bc34ec61bc5ba7b673ed5478a3/Neos.Flow/Classes/Error/AbstractExceptionHandler.php#L338

i think then we should rather refactor the whole exception handling part and not introduce only a random `wrapInPre` option ...

kitsunet:
I wonder if we can't scoop a good solution eg. from symfony or whoops or so.

kdambekalns (approved):
I think that's fine, rather preserve formatting and risk readability than reverse. And we can change it, if it really becomes an issue.

mhsdesign:
@bwaidelich i just rebased this pr and its crazy that its already two years old ;D
I understand your preservations, but think it's okay to fix this as proposed.

PR #2701, opened 2022-02-21
FEATURE: Lazy injections for typed properties

### LINKED ISSUE #2114: Full support for PHP 7.4 property type declarations
### Description

Since version 7.4 PHP supports type declarations, see [RFC](https://wiki.php.net/rfc/typed_properties_v2) but we don't respect those in our ReflectionService, so the following won't work:

```php
class SomeClass
{
    /**
     * @Flow\Inject
     **/
    protected ?LoggerInterface $logger = null;
}
```

### LINKED ISSUE #2601: Full PHP8 typesafe injectable properties
### Description
At the moment this is not possible with Flow 7.2:

```php
class MonitoringCommandController extends CommandController
{
    #[Flow\Inject(lazy: true)]
    protected ?MyLazyService $myLazyService = null;

   #[Flow\Inject(lazy: false)]
    protected ?MyService $myService = null;

   /**
     * @var AnalyticsService|null
    */
   #[Flow\Inject(lazy: false)]
    protected ?MyAnotherService $myAnotherService = null;

  /*code */

}
```

- The first example does not work because Flow uses another type as a placeholder to load the type on demand
- The second (as well as the first) example does not work because of the missing type annotation.
- The third example does not work because of the "|null" in the type annotation

My suggestion would be make the property Injection more robust fit for PHP8+

robertlemke:
This change adjusts the proxy building and dependency injection process to properly handle PHP 7.4 class property type declarations.

It resolves #2114
Also resolve #2601 

**How to verify it**

What did *not* work before this change was using property type declarations for lazily injected properties. 

Consider the following code: injecting `$testSingleton` is what this fix makes possible.

```php
<?php
namespace Acme\LazyTest\Command;

use Acme\LazyTest\TestSingleton;
use Neos\Flow\Cache\CacheManager;
use Psr\Log\LoggerInterface;

/**
 * @Flow\Scope("singleton")
 */
class TestCommandController extends CommandController
{
    /**
     * LoggerInterface doesn't have to be nullable, because the proxy class
     * builder will not even try to use lazy injection, since it is an interface
     *
     * @Flow\Inject
     **/

    /**
     * CacheManager must be nullable, because it can be injected lazily
     *
     * @Flow\Inject
     */
    protected ?CacheManager $cacheManager;

    /**
     * @Flow\Inject
     */
    protected ?TestSingleton $testSingleton;

    /**
     * Test command
     *
     * @return void
     */
    public function testCommand(): void
    {
        # Will be the actual logger, because an interface is injected,
        # which results in an eager dependency:

        var_dump(get_class($this->logger)); # Neos\Flow\Log\Psr\Logger

        # Will be the actual Cache Manager, because as soon as the
        # Cache Manager is used (instantiated) elsewhere, it is also
        # injected eagerly into this controller:

        var_dump(get_class($this->cacheManager));                      # Neos\Flow\Cache\CacheManager

        # The TestSingleton is not injected into any other class and
        # therefore is a real Lazy Proxy before it is used for the
        # first time by calling hello():

        var_dump(get_class($this->testSingleton));                     # Acme\LazyTest\TestSingleton_LazyProxy
        var_dump($this->testSingleton instanceof TestSingleton); # bool(true)
        $this->testSingleton->hello('world');
        var_dump(get_class($this->testSingleton));                     # Acme\LazyTest\TestSingleton
    }
}
```

In order to verify this change, you can use the test package (see zip file) and play a bit with the injected properties. Things to try out (at least):

- verify the output in Development context after initial run of `./flow test:test` with an empty cache
- verify the output on a second rund
- verify that it works when you make a little change, so that proxy classes are built
- check that all this works in Production

[Acme.LazyTest.zip](https://github.com/neos/flow-development-collection/files/8118718/Acme.LazyTest.zip)


robertlemke:
The code works now and should solve the problem at hand.

What's still missing are the respective test adjustments.

robertlemke:
This is now ready for reviews.

robertlemke:
Hmm, that's weird – all functional and unit tests succeeded in Phpstorm using PHP 7.4.

image

robertlemke:
Sigh, PHP 7.3 fails. So I either need to scrub this more, so that it doesn't fail with PHP 7.3, or don't release this in Flow 7.3 but only 8.0.

albe (commented):
I'll take a look at the 7.3 failures this evening. Looks as if the generated code for the proxies is just slightly off, though not sure why only on 7.3? Maybe a different laminas version is installed there with slightly different behaviour?

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -217,7 +219,7 @@ public function get($objectName, ...$constructorArguments)
             return $this->instantiateClass($className, $constructorArguments);
         }
 
-        $this->objects[$objectName]['i'] = $this->instantiateClass($className, []);
+        $this->objects[$objectName]['i'] = $this->instantiateClass($className, $constructorArguments);
```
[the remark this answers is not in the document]
robertlemke: I am pretty sure that I added this due to a failing test which claimed that passing constructor arguments must be possible. But unfortunately I can neither reproduce this nor find the test, so I'll revert this change again.

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
```diff
@@ -11,9 +11,18 @@
  * source code.
  */
 
+use Laminas\Code\Generator\ClassGenerator;
```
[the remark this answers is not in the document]
robertlemke: I'll add it as an explicit requirement.

robertlemke:
This change is now targeted and rebased for master.

mhsdesign:
It would be really great to get this at least into Flow 8 ;) (followed your discussion: https://discuss.neos.io/t/rfc-drop-php-7-3-support-in-all-future-flow-versions/5844/41?u=marc)

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -561,7 +561,13 @@ protected function autowireProperties(array &$objectConfigurations)
...
                     /** @var Inject $injectAnnotation */
                     $injectAnnotation = $this->reflectionService->getPropertyAnnotation($className, $propertyName, Inject::class);
-                    $objectName = $injectAnnotation->name !== null ? $injectAnnotation->name : trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');
+                    $objectName = $injectAnnotation->name;
+                    if ($objectName === null) {
+                        $objectName = $this->reflectionService->getPropertyType($className, $propertyName);
+                    }
+                    if ($objectName === null) {
+                        $objectName = trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');
+                    }
```
mhsdesign, line 570: would that work or is this to fancy 😁

```suggestion
                    $objectName = $injectAnnotation->name
                        ?? $this->reflectionService->getPropertyType($className, $propertyName)
                        ?? trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');
```

or maybe this `??=`?

```suggestion
                    $objectName = $injectAnnotation->name
                    $objectName ??= $this->reflectionService->getPropertyType($className, $propertyName)
                    $objectName ??= trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');

```

Neos.Flow/Classes/ObjectManagement/DependencyInjection/DependencyProxyTrait.php
```diff
@@ -0,0 +1,90 @@
...
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+trait DependencyProxyTrait
+{
+    /**
+     * @var string
+     */
+    protected $_dependencyClassName;
```
mhsdesign, line 19: would it make sense to make the properties private?

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
robertlemke: Thanks for the suggestion!

If you wonder if this would work, I'd say it's too fancy.

Neos.Flow/Classes/ObjectManagement/DependencyInjection/DependencyProxyTrait.php
robertlemke: I haven't tried it. But we need to careful, because proxy classes can be extended (by Flow).

In my opinion it doesn't make a difference if this was `private`, because only Flow can extend proxy classes.

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
mhsdesign: true i have never seen a chained null coalescing - but i works ^^

https://3v4l.org/FtNJV

(and i love fancy 😁)

robertlemke: Alright, let's go fancy then 🕺

Neos.Flow/Classes/ObjectManagement/DependencyInjection/PropertyInjectionTrait.php
```diff
@@ -27,14 +27,18 @@ trait PropertyInjectionTrait
...
         if ($this->$propertyName === null) {
             $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->getLazyDependencyByHash($setterArgumentHash, $injection_reference);
             if ($this->$propertyName === null) {
                 $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->createLazyDependency($setterArgumentHash, $injection_reference, $propertyClassName, $lazyInjectionResolver);
+                if ($this->$propertyName === null) {
+                    // Fall back to eager injection, if lazy property injection was not possible
+                    $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->get($propertyObjectName);
+                }
             }
         }
```
mhsdesign, line 43: hmm it might add readability what do you say ^^

```suggestion
    $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->getInstance($propertyObjectName)
        ?? \Neos\Flow\Core\Bootstrap::$staticObjectManager->getLazyDependencyByHash($setterArgumentHash, $injection_reference)
        ?? \Neos\Flow\Core\Bootstrap::$staticObjectManager->createLazyDependency($setterArgumentHash, $injection_reference, $propertyClassName, $lazyInjectionResolver)
        // Fall back to eager injection, if lazy property injection was not possible
        ?? \Neos\Flow\Core\Bootstrap::$staticObjectManager->get($propertyObjectName);
```

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyClassTest.php
```diff
@@ -23,16 +24,20 @@ class ProxyClassTest extends UnitTestCase
...
+        require_once(__DIR__ . '/../Fixture/ClassWithoutNamespace.php');
+
         return [
             [
-                'originalClassName' => '\Acme\Namespace\ClassName',
+                'originalClassName' => ClassImplementingInterfaceWithConstructor::class,
                 'originalClassAnnotations' => [],
                 'originalClassDocumentation' => '',
                 'originalClassConstants' => [['name' => 'TEST_CONSTANT', 'value' => '1']],
                 'expectedProxyCode' =>
```
mhsdesign, line 35: maybe one could use heredoc but since this is working now fine already ...

  ```
                  <<<PHP
                    class ClassImplementingInterfaceWithConstructor extends ClassImplementingInterfaceWithConstructor$compilerOriginalClassnameSuffix implements \\Neos\\Flow\\ObjectManagement\\Proxy\\ProxyInterface
                    {
                        const TEST_CONSTANT = 1;
                    }
                    PHP
```

Comment (robertlemke) [reply Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyClassTest.php:35]: I don't want to touch too much code, as I think this PR is big enough already. To be honest, I'm also running out of time a bit before the feature freeze.

File: Neos.Flow/Classes/ObjectManagement/DependencyInjection/PropertyInjectionTrait.php (hunk 5 above)
Comment (robertlemke) [reply Neos.Flow/Classes/ObjectManagement/DependencyInjection/PropertyInjectionTrait.php:43]: Please, let's postpone such changes to a follow-up. There is so much code which can be style-improved, let's dedicate a new PR to those.

File: Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php [hunk 7]
```diff
@@ -466,7 +468,33 @@ public function buildPropertyInjectionCodeByString(Configuration $objectConfigur
...
+        if ($propertyClassName) {
+            foreach ((new ClassReflection($propertyClassName))->getInterfaceNames() as $interfaceName) {
+                if (method_exists($interfaceName, '__construct')) {
+                    $propertyClassHasInterfaceWithConstructor = true;
+                    break;
+                }
+            }
+            foreach (get_class_methods($propertyClassName) as $methodName) {
+                if ($this->reflectionService->isMethodFinal($propertyClassName, $methodName)) {
+                    $propertyClassHasFinalMethod = true;
```
albe, line 482:
```suggestion
                    $propertyClassHasFinalMethod = true;
                    break;
```
microoptimization

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -466,7 +468,33 @@ public function buildPropertyInjectionCodeByString(Configuration $objectConfigur
...
+                }
+            }
+        }
+
+        # There are several cases when providing lazy loading is not possible, due to the way
+        # lazy loading is implemented. If any of the check fails, we fall back to eager loading:
+        if (
+            $propertyConfiguration->isLazyLoading() &&
+            $this->objectConfigurations[$propertyObjectName]->getScope() !== Configuration::SCOPE_PROTOTYPE &&
+            !interface_exists($this->objectConfigurations[$propertyObjectName]->getClassName()) &&
```
albe, line 492: If the target class is an interface, but the interface is Objects.yaml configured to a specific implementation, why wouldn't that work lazily?

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -403,24 +405,37 @@ public function getLazyDependencyByHash($hash, &$propertyReferenceVariable)
...
-    public function createLazyDependency($hash, &$propertyReferenceVariable, $className, \Closure $builder): DependencyProxy
+    public function createLazyDependency(string $hash, &$propertyReferenceVariable, string $className, Closure $builder): ?DependencyProxy
     {
-        $this->dependencyProxies[$hash] = new DependencyProxy($className, $builder);
+        // Trigger auto-loading of the original class, because any generated lazy proxy will be
+        // contained in the same file and the auto-loader does not know how to find a lazy proxy:
+        class_exists($className) || interface_exists($className);
+        $lazyProxyClassName = $className . '_LazyProxy';
+
+        // If proxy classes were partially flushed in Development context, it might happen, that
```
albe, line 420:
```suggestion
        // If proxy classes were partially flushed in Development context, it might happen that
```
🧐

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
```diff
@@ -235,7 +246,27 @@ public function render()
...
+            if (method_exists($interfaceName, '__construct')) {
+                $hasInterfaceWithConstructor = true;
+                break;
+            }
+        }
+
+        $hasFinalMethod = false;
+        foreach (get_class_methods($this->fullOriginalClassName) as $methodName) {
+            if ($this->reflectionService->isMethodFinal($this->fullOriginalClassName, $methodName)) {
+                $hasFinalMethod = true;
```
albe, line 262:
```suggestion
                $hasFinalMethod = true;
                break;
```

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1306,6 +1317,10 @@ public function reflectClassProperty($className, PropertyReflection $property)
         $visibility = $property->isPublic() ? self::VISIBILITY_PUBLIC : ($property->isProtected() ? self::VISIBILITY_PROTECTED : self::VISIBILITY_PRIVATE);
         $this->classReflectionData[$className][self::DATA_CLASS_PROPERTIES][$propertyName][self::DATA_PROPERTY_VISIBILITY] = $visibility;
 
+        if ($property->hasType()) {
+            $this->classReflectionData[$className][self::DATA_CLASS_PROPERTIES][$propertyName][self::DATA_PROPERTY_TYPE] = trim((string)$property->getType(), '?');
```
albe, line 1321: Question: Should we check this preemptively against a potential value from the "TagValues" below, which contain the `@var` tag? Only valid edge-case I can think of is the tag containing a concrete class and the property being typed against an interface (or vice-versa if that makes any sense?), so we'd need to check `class_implements()` (and maybe `class_parents()`?). Maybe something for another PR

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP8/ClassWithLazyDependencies.php
```diff
@@ -1,5 +1,5 @@
-<?php
-namespace Neos\Flow\Tests\Functional\ObjectManagement\Fixtures;
+<?php /** @noinspection PhpMissingFieldTypeInspection */
+namespace Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PHP8;
```
albe, line 2: Should we maybe keep the original test for untyped lazy loading?

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
```diff
@@ -302,4 +333,68 @@ protected function renderTraitsCode()
...
+        }
+
+        $classGenerator = ClassGenerator::fromArray(['name' => $proxyClassName . '_LazyProxy'])
+            ->setExtendedClass($proxyClassName)
+            ->setImplementedInterfaces([DependencyProxy::class])
+            ->removeMethod('__call')
+            ->addMethods($methods)
+            ->addTrait('\\' . DependencyProxyTrait::class);
+        // This is just for making Psalm happy, since addTrait() returns TraitUsageInterface which doesn't provide generate()
+        assert($classGenerator instanceof ClassGenerator);
```
mhsdesign, line 371: isnt this `assert` odd? - does php doc suffice? (im just curious - shouldnt stop this pr ^^)
```suggestion
        /** @var $classGenerator ClassGenerator&TraitUsageInterface */
```

albe: Should work as well to make psalm understand, yes. The assert is more bullet-proof though (actual runtime check) and is optimized away in production anyway.

albe:
Left some comments. Overall looks very good and build is 💚

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
mhsdesign: we are basically rewriting js `Array.some()` - and i just looked it up - there is no php `array_some` lol

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP8/ClassWithLazyDependencies.php
mhsdesign: there is still `AnotherClassWithLazyDependencies` (which was a copy from the original `ClassWithLazyDependencies`)

https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/AnotherClassWithLazyDependencies.php

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -466,7 +468,33 @@ public function buildPropertyInjectionCodeByString(Configuration $objectConfigur
...
+            !$this->objectConfigurations[$propertyObjectName]->isCreatedByFactory() &&
```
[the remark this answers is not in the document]
robertlemke: We would need to analyze the Objects.yaml-configuration for the factory – recursively – in order to be sure what is going to be injected. That is, if the object which is created by factory needs some other singleton, which uses a factory for some object which in the end is a _prototype_, we must disable lazy injection. But I didn't dare to develop such an analysis in the scope of this PR. Therefore, better be safe and not use lazy injection in that case.

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -466,7 +468,33 @@ public function buildPropertyInjectionCodeByString(Configuration $objectConfigur
...
+            !$propertyClassHasInterfaceWithConstructor &&
+            !$propertyClassHasFinalMethod &&
```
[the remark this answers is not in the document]
robertlemke: Yes, that can be done. But it isn't trivial and justifies its own pull request.

robertlemke:
Thanks a lot for the accurate review! I tried to answer some questions and applied your code suggestions.

mhsdesign:
the failing tests will be gone after an upmerge ... see https://github.com/neos/flow-development-collection/pull/2756

mhsdesign:
i will test the functionality locally - i cannot say too much about all that proxybuilding ^^

mficzel:
Can confirm the functionality working and being great and the changes make sense ... however this proxy class handling is still a bit greek to me. Maybe count this as half a review. 👍

robertlemke:
@mficzel If it works for you and tests succeed, that's probably all we need right now. As for the code, just trust me – it's at least better than before ;-)

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
mhsdesign: ... this was not added ...

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -403,24 +405,37 @@ public function getLazyDependencyByHash($hash, &$propertyReferenceVariable)
...
+
+        // If proxy classes were partially flushed in Development context, it might happen that
+        // the original class $className was already loaded by the ConfigurationBuilder. In that
+        // case, the proxy class for $className cannot be loaded any more by the ProxyClassLoader,
+        // so we need to fall back to eager dependency injection.
+        if (!class_exists($lazyProxyClassName)) {
+            return null;
+        }
+
+        $this->dependencyProxies[$hash] = call_user_func_array([$lazyProxyClassName, '_createDependencyProxy'], [$className, $builder]);
```
mhsdesign, line 428:
```suggestion
        $this->dependencyProxies[$hash] = $lazyProxyClassName::_createDependencyProxy($className, $builder);
```

robertlemke: Makes sense – did you test this?

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
kdambekalns: > ... this was not added ...

I don't get it. What was not added where?

robertlemke: The suggestion was not committed to this PR. But now it was.

kdambekalns: Yes, got it. And now was wondering why I can't find the comment, but it was already resolved. 🤣

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyClassTest.php
```diff
@@ -52,7 +66,16 @@ public function proxyClassesDataProvider()
...
                     'class ClassWithoutNamespace extends ClassWithoutNamespace' . Compiler::ORIGINAL_CLASSNAME_SUFFIX . " implements \\Neos\\Flow\\ObjectManagement\\Proxy\\ProxyInterface {\n\n" .
                     "    const TEST_CONSTANT = 1;\n\n" .
-                    '}',
+                    "}\n\n" .
+                    "class ClassWithoutNamespace_LazyProxy extends ClassWithoutNamespace implements \Neos\Flow\ObjectManagement\DependencyInjection\DependencyProxy\n" .
+                    "{\n" .
+                    "    use \Neos\Flow\ObjectManagement\DependencyInjection\DependencyProxyTrait;\n\n" .
+                    "    public function __call(string \$methodName, array \$arguments)\n" .
+                    "    {\n" .
+                    "        [\$methodName, \$arguments] = func_get_args();\n" .
```
mhsdesign, line 75: why do we need this line? arent those parameter declared as method argument?

i i guess, if someone writes a `__call` method on their own class so its compatible?

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
mhsdesign: will ^^
but made something like this yesterday syntax wise and it worked for a static method on a classname variable

Neos.Flow/composer.json
```diff
@@ -35,6 +35,7 @@
         "psr/http-server-handler": "^1.0",
 
         "ramsey/uuid": "^3.0 || ^4.0",
+        "laminas/laminas-code": "~3.4.1|^4.0",
```
kdambekalns, line 41: Same result, but consistent with what we usually do. Also the official syntax according to https://getcomposer.org/doc/articles/versions.md#version-range

```suggestion
        "laminas/laminas-code": "~3.4.1 || ^4.0",
```

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
kdambekalns: At least the demo site works fine with this, and the tests pass, too.

kdambekalns (approved):
Tests pass, works with Neos.Demo site, looks good.

The two comments by @mhsdesign are worth being addressed, but no blockers IMHO.

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyClassTest.php
robertlemke: Yes, just to make sure that the method signature is compatible.

robertlemke:
Thanks, @kdambekalns!

At this point all pending questions are answered and code suggestions were committed.

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
```diff
@@ -302,4 +334,68 @@ protected function renderTraitsCode()
...
+                ParameterGenerator::fromArray(['name' => 'arguments', 'type' => 'array']),
+            ]);
+        }
+
+        $callMagicMethod->setBody(
+            <<< CODE
+                [\$methodName, \$arguments] = func_get_args();
+                return \$this->_activateDependency()->\$methodName(...\$arguments);
+            CODE
+        );
```
mhsdesign, line 397: isnt usign `func_get_args` always not so cool?
https://github.com/neos/flow-development-collection/pull/2701#discussion_r834056321

would be a minor performance boost in some cases: but should not block this pr!

```suggestion
        if (method_exists($this->fullOriginalClassName, '__call')) {
            $callMagicMethod = MethodGenerator::fromReflection(new \Laminas\Code\Reflection\MethodReflection($this->fullOriginalClassName, '__call'));
            $callMagicMethod->setBody(
                <<<'CODE'
                    [$methodName, $arguments] = func_get_args();
                    return $this->_activateDependency()->$methodName(...$arguments);
                CODE
             );
        } else {
            $callMagicMethod = MethodGenerator::fromArray(['name' => '__call']);
            $callMagicMethod->setParameters([
                ParameterGenerator::fromArray(['name' => 'methodName', 'type' => 'string']),
                ParameterGenerator::fromArray(['name' => 'arguments', 'type' => 'array']),
            ]);
            $callMagicMethod->setBody(
                <<<'CODE'
                    return $this->_activateDependency()->$methodName(...$arguments);
                CODE
            );
        }
```

mhsdesign: otherwise the intention about why we even care to do so: is not so clear to me ...

```
            $callMagicMethod->setParameters([
                ParameterGenerator::fromArray(['name' => 'methodName', 'type' => 'string']),
                ParameterGenerator::fromArray(['name' => 'arguments', 'type' => 'array']),
            ]); 
```

mhsdesign: thank you for your time and work already, and sorry for beeing maybe too hard on this pr - i just want to be a good reviewer. I will test this later locally and then martins 1/2 review and my half share are 1 so then we can maybe merge already ^^

mhsdesign:
@robertlemke i really want to approve this since the syntax is much much much much fancier ... and cooler ^^

but i found out in my tests that this really increases the time to get flow ready from 0 - 100.
eg. its common practice to do a `rm -rf Data/Temporary` or `./flow flow:cache:flush` every once and then if something isnt working that great and since caches often dont cover every scenario when they should be invalidated.

so i directly noticed a performance increase while doing functional Fusion tests (i was trying out your new functionality there)
from a flushed cache it took `flow@master` ~11.5 seconds to start the bootstrap and run one test.
with `flow@robert` it took ~ 18 seconds...

to make this more comparable i flushed the caches and manually timed how long `./flow flow:core:compile` took (made a few iterations, result is pie mal daumen):

| | `flow@master` | `flow@robert` |
| -- |  --- | --- | 
| `./flow flow:core:compile` | 2.6s | 5s |

mhsdesign (changes requested):
imagine i have this code: (DslFactory is a singelton)


```
    /**
     * @Flow\Inject
     * @var DslFactory
     */
    protected $dslFactory;
```

i now want to use typed properties, so i naively do so first:

```
    /**
     * @Flow\Inject
     */
    protected DslFactory $dslFactory;
```

and get some cryptic exception about: [code line](https://github.com/neos/flow-development-collection/pull/2701/files#diff-40ba3af0737074f90108680d4bc5dda2657ba595e31814904eac43f80f0af099R32):
```
Cannot access uninitialized non-nullable property Neos\Fusion\Core\Parser_Original::$dslFactory by reference
DependencyInjection/PropertyInjectionTrait.php:32
```

then i go back to your nice PR body and ahh i need to use `?` so i do so.

```
    /**
     * @Flow\Inject
     */
    protected ?DslFactory $dslFactory;
```

i think we should have a better exception message than the of php. -> thats easy.

but the problem is my code doesnt work now directly, i needed to flush the caches - what takes some time longer and then it works.

----

in short:

do this:

```
    /**
     * @Flow\Inject
     * @var DslFactory
     */
    protected $dslFactory;
```
run flow ...
```
    /**
     * @var DslFactory
     */
    protected $dslFactory;
```
run flow ... (doesnt work)
```
    /**
     * @Flow\Inject
     * @var DslFactory
     */
    protected $dslFactory;
```
run flow ... `flow@master` works here, but `flow@robert` doesnt... in my case the `DslFactory` had no working `@Flow\InjectConfiguration` and the $objectManger wasnt injected too (so no working DI) after a subtle change in the DslFactory File the injection was working again too (or after cache flush).

mhsdesign:
so at the moment this feels sometimes buggy - just go around fx. in the neos dev collection and try to change to typed properties here and there -> you will often find yourself in need to make a cache flush so the proxybuilding is properly working again.

robertlemke:
@mhsdesign very good catch!

Here are my timings for comparison:

| | `flow@master` | `flow@robert` |
| -- |  --- | --- | 
| neos/flow unit tests | 780ms | 782ms |
| neos/flow functional tests | 46s | 46s |
| neos/flow functional tests (only object management) | 115ms | 112ms |
| `rm -rf Data/Temporary && time ./flow` | 1.39s | 2.4s ⚠️|

I did a quick first profiling and apparently, the Reflection Service is the culprit. From a total wall time of 15.2 seconds (it takes longer because of the profiling engine), my branch is more than 6 seconds slower while reflecting classes.

image

I'll take a closer look and see where code can be optimized.

robertlemke:
As I feared … the slow part is the new Laminas Class Building code, which doesn't use our optimized Reflection Service. Let's see if I can work around this a bit.

image

robertlemke:
If someone is reviewing this right now – please postpone your review a little. I need to rework some of the internals due to the performance issues.

mhsdesign:
do i see this right that using `Laminas` is not crucial and just bewares us from manually string concatenating everything together?

robertlemke:
> do i see this right that using Laminas is not crucial and just bewares us from manually string concatenating everything together?

Mostly, yes. But it's also a handy API which allows you to pass an existing class reflection and then modify code through certain setters. The idea was to transition to a library (Laminas) instead of doing all that on our own (like we did before). But unfortunately, it doesn't work (fast enough) out of the box.

robertlemke:
PS: There are more problems than that, because the Laminas Code Builder tries to use our doc comment type hints in certain cases, which may be invalid, like in this example:

image

I'll work around that, but you see, it's a diligent but boring piece of work …

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
mhsdesign: that doesnt seem to be resolved yet - or am i wrong?
we say in the docs that this would work - and codewise, since that `interface_exists` was just added it must have worked ... https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/ObjectManagement.html?highlight=lazy#lazy-dependency-injection

can we still support this?

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -466,7 +468,34 @@ public function buildPropertyInjectionCodeByString(Configuration $objectConfigur
...
+                    break;
+                }
+            }
+        }
+
+        # There are several cases when providing lazy loading is not possible, due to the way
+        # lazy loading is implemented. If any of the check fails, we fall back to eager loading:
+        if (
+            $propertyConfiguration->isLazyLoading() &&
+            $this->objectConfigurations[$propertyObjectName]->getScope() !== Configuration::SCOPE_PROTOTYPE &&
```
mhsdesign, line 492: We dont allow prototypes lazly injected - but on my pc there are tousands of lazy proxies for prototypes generated ... isnt this a waste of performance?

Neos.Flow/Classes/ObjectManagement/DependencyInjection/PropertyInjectionTrait.php
```diff
@@ -27,14 +27,18 @@ trait PropertyInjectionTrait
...
-    private function Flow_Proxy_LazyPropertyInjection($propertyObjectName, $propertyClassName, $propertyName, $setterArgumentHash, callable $lazyInjectionResolver)
+    private function Flow_Proxy_LazyPropertyInjection(string $propertyObjectName, string $propertyClassName, string $propertyName, string $setterArgumentHash, callable $lazyInjectionResolver)
     {
         $injection_reference = &$this->$propertyName;
         $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->getInstance($propertyObjectName);
         if ($this->$propertyName === null) {
             $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->getLazyDependencyByHash($setterArgumentHash, $injection_reference);
             if ($this->$propertyName === null) {
                 $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->createLazyDependency($setterArgumentHash, $injection_reference, $propertyClassName, $lazyInjectionResolver);
+                if ($this->$propertyName === null) {
```
mhsdesign, line 38: do we allow with this change less lazy dependency injection that before?
related: interfaces fx. cant be injected lazily anymore? https://github.com/neos/flow-development-collection/pull/2701#discussion_r830984712

Neos.Flow/Classes/ObjectManagement/DependencyInjection/PropertyInjectionTrait.php
```diff
@@ -27,14 +27,18 @@ trait PropertyInjectionTrait
...
      */
-    private function Flow_Proxy_LazyPropertyInjection($propertyObjectName, $propertyClassName, $propertyName, $setterArgumentHash, callable $lazyInjectionResolver)
+    private function Flow_Proxy_LazyPropertyInjection(string $propertyObjectName, string $propertyClassName, string $propertyName, string $setterArgumentHash, callable $lazyInjectionResolver)
     {
         $injection_reference = &$this->$propertyName;
         $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->getInstance($propertyObjectName);
         if ($this->$propertyName === null) {
             $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->getLazyDependencyByHash($setterArgumentHash, $injection_reference);
             if ($this->$propertyName === null) {
                 $this->$propertyName = \Neos\Flow\Core\Bootstrap::$staticObjectManager->createLazyDependency($setterArgumentHash, $injection_reference, $propertyClassName, $lazyInjectionResolver);
```
mhsdesign, line 37: i finally kindof find the reson for this buggy behavior that on the fly di changes dont work correctly:
https://github.com/neos/flow-development-collection/pull/2701#pullrequestreview-921001408

it seems that this method `createLazyDependency` is returning `null` and the next call: $staticObjectManager->get($propertyObjectName);

doesnt return an object of the proxied class but rather an object based of the class in my working dir...

mhsdesign: so https://github.com/neos/flow-development-collection/pull/2701#pullrequestreview-921001408 even shorter: 
 go to a lazy dependency injection (in the old doc comment style - so you can compare this exact behaviour with `flow@master`)

- remove the `@Flow\Inject` and run a flow
- add the  `@Flow\Inject` again and run flow again -> now the property should be injected.

but at `flow@robert` as i seems this code:

https://github.com/neos/flow-development-collection/blob/aea746af48b02828cec7017aabeb278d5bf5a1f6/Neos.Flow/Classes/ObjectManagement/ObjectManager.php#L420

is triggered but `flow@master` is working in this scenario

mhsdesign:
@fcool thought me that it works already like this:

```php
#[Flow\Inject(name: BlaInterface::class, lazy: true)]
protected BlaInterface|DependencyProxy|null $bla;
```

it would be great of course if i could leave out the (redundant) `name: BlaInterface::class`
(and idk if `lazy: true` is needed i dont remember ^^)

mficzel:
I kindoff like this style. Makes the lazyness very explicit and opt in. For me it would be fine to document this.

mhsdesign:
> I kindoff like this style. Makes the lazyness very explicit and opt in. For me it would be fine to document this.

me too (especially cases, where youd need to pass the dependency it becomes clear why youd need to activate it first), but i would like to get rid of the extra arguments so we can write:

```php
#[Flow\Inject()]
protected BlaInterface|DependencyProxy|null $bla;
```

Issue #2698, opened 2022-02-07
Adjustments for PHP 8.1 


mficzel:
For projects that decide to use php 8.1 Flow should:

- Run flawlessly o php 8.1
- Allow to use features of php 8.1 especially the improved type hinting in projectcode 

Note: The framework itself should stay compatible wit php 7.4 for now.

PR #2702, opened 2022-02-21
BUGFIX: Don't create proxy classes for PHP enums

robertlemke:
This change introduces a check in the proxy class compiler which makes sure that PHP 8.1 enums are not considered as regular classes and are removed from the list of proxyable classes.

How to reproduce:

1. create an enum, for example in `Packages/Acme/Classes/ExampleEnum.php`:
```php
namespace Acme;
enum ExampleEnum
{
    case FOO;
    case BAR;
}
```

2. let Flow compile classes, the following error will appear:

> PHP Fatal error:  Cannot declare class Acme\ExampleEnum, because the name is already in use in /application/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/Acme_ExampleEnum.php on line 23

Addresses #2698

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
@@ -228,7 +228,10 @@ class CompileTimeObjectManager extends ObjectManager
                 }
                 if ($package instanceof FlowPackageInterface && $shouldRegisterFunctionalTestClasses) {
                     foreach ($package->getFunctionalTestsClassFiles() as $fullClassName => $path) {
-                        if (PHP_MAJOR_VERSION < 8 && strpos($fullClassName, '\\PHP8\\') !== false) {
+                        if (version_compare(PHP_VERSION, '8.0', '<=') && strpos($fullClassName, '\\PHP8\\') !== false) {
+                            continue;
+                        }
+                        if (version_compare(PHP_VERSION, '8.1', '<=') && strpos($fullClassName, '\\PHP81\\') !== false) {
                             continue;
                         }
                         if (substr($fullClassName, -9, 9) !== 'Exception') {
--- a/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP81/BackedEnumWithMethod.php
+++ b/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP81/BackedEnumWithMethod.php
@@ -1,5 +1,5 @@
 <?php
-namespace Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PHP8;
+namespace Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PHP81;
 
 /*
  * This file is part of the Neos.Flow package.
--- a/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP81/BackedEnumWithMethod.php
+++ b/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP81/BackedEnumWithMethod.php
@@ -11,8 +11,6 @@ namespace Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PHP81;
  * source code.
  */
 
-use Neos\Flow\Annotations\Proxy;
-
 /**
  * A PHP 8.1 value-backed enum with methods
  */
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,6 +1,6 @@
 preset: psr2
 
-version: 8
+version: 8.1
 
 finder:
   path:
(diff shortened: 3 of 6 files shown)


albe (commented):
Looks good, thanks! Can we do something like we did here https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php#L231 => https://github.com/neos/flow-development-collection/tree/master/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP8 and add such an enum to the tests conditionally for PHP81?

robertlemke:
> Looks good, thanks! Can we do something like we did here https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php#L231 => https://github.com/neos/flow-development-collection/tree/master/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP8 and add such an enum to the tests conditionally for PHP81?

Yes, more or less – we can't check if the enum was proxied, because it never will be, since the compiler would fail with a fatal error. But yes, we need to back this with a test, so I just pushed one.

robertlemke:
Hmm. Now we need to find a good way to sneak this test in without breaking the < PHP 8.1 runs.

robertlemke:
Okay, I think I found a solution an force-pushed the commit providing the functional test.

kdambekalns (approved):
👍

kdambekalns:
Sadly it fails for PHP 8.0 with

```
 syntax error, unexpected identifier "BackedEnumWithMethod"
Type: ParseError
  File:
Packages/Framework/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP
        8/BackedEnumWithMethod.php
  Line: 19
```

robertlemke:
Pretty weird, works on my machine™️ …
image

But I'll try to fix it.

robertlemke:
Oh and we should run tests for 8.1 I guess ;-)

kdambekalns (approved):

nezaniel:
@8.1 tests:
I've just merged https://github.com/neos/flow-development-collection/pull/2695, here you go :)

robertlemke:
StyleCI doesn't seem to like enums …

kdambekalns:
> StyleCI doesn't seem to like enums …

https://docs.styleci.io/standalone-php#version

kitsunet (approved):
Looks good, thanks! Good to have that fixed

Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
```diff
@@ -228,7 +228,10 @@ protected function registerClassFiles(array $packages)
                 }
                 if ($package instanceof FlowPackageInterface && $shouldRegisterFunctionalTestClasses) {
                     foreach ($package->getFunctionalTestsClassFiles() as $fullClassName => $path) {
-                        if (PHP_MAJOR_VERSION < 8 && strpos($fullClassName, '\\PHP8\\') !== false) {
+                        if (version_compare(PHP_VERSION, '8.0', '<=') && strpos($fullClassName, '\\PHP8\\') !== false) {
```
bwaidelich, line 231: Not sure if it's much better, but my PHPStorm suggests to change this line to
```suggestion
                        if (PHP_VERSION_ID <= 80000 && strpos($fullClassName, '\\PHP8\\') !== false) {
```

But apart from that: I'm wondering what these checks actually do?
Could we add some inline comment to explain that?

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP81/BackedEnumWithMethod.php
```diff
@@ -0,0 +1,38 @@
...
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Annotations\Proxy;
+
+/**
+ * A PHP 8.1 value-backed enum with methods
+ */
+enum BackedEnumWithMethod: string
```
bwaidelich, line 19: I assume that this fixture is not loaded with PHP < 8.1 and doesn't cause any harm that way!?

bwaidelich (commented):
I think this makes sense, but as written in Slack IMO we should strive towards creating proxies only if they are really needed. And, if a proxy was to be created for a class that is not supported (e.g. an enum) an exception should be thrown.
Otherwise: What happens if you add an AOP aspect for an enum with this change? Will it be ignored?

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP81/BackedEnumWithMethod.php
kdambekalns: Yes, exactly what you wondered about in [CompileTimeObjectManager.php](https://github.com/neos/flow-development-collection/pull/2702/files/6519b87dd9b6dbab2ee5b8c352ab108ffc53917d#diff-e4637b28695cb70185ef6ac4a5b949b4051de8e9844f2c5f797b472ea66884aa)

robertlemke:
What do we do with the failing psalm and styleci tests? They are false-positives, correct?

kdambekalns:
> What do we do with the failing psalm and styleci tests? They are false-positives, correct?

I'll take care…

albe:
Post-merge 👍 Thanks everyone!

PR #2703, opened 2022-02-21
Make the JsonArrayType able to handle ValueObjects

nezaniel:
This enables the JsonArrayType to de/serialize arbitrary value objects, as long as they implement \JsonSerializable and ::from$Type. Might be useful for a certain CMS ;)

Resolves: #2763

- [x] The PR is created against Flow 5.3

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
@@ -1,0 +1,182 @@
+<?php
+namespace Neos\Flow\Property\TypeConverter;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Property\Exception\TypeConverterException;
+use Neos\Flow\Property\PropertyMappingConfigurationInterface;
+use Neos\Flow\Property\TypeConverterInterface;
+
+final class DenormalizingObjectConverter implements TypeConverterInterface
+{
+    /**
+     * @return array<string>
+     */
+    public function getSupportedSourceTypes()
+    {
+        return ['array', 'string', 'bool', 'int', 'float'];
+    }
+
+    /**
+     * @return string
+     */
+    public function getSupportedTargetType()
+    {
+        return 'object';
+    }
+
+    /**
+     * @param mixed $source the source data
+     * @param string $originalTargetType the type we originally want to convert to
+     * @param PropertyMappingConfigurationInterface $configuration
+     * @return string
+     * @api
+     */
+    public function getTargetTypeForSource($source, $originalTargetType, PropertyMappingConfigurationInterface $configuration = null)
+    {
+        return $originalTargetType;
+    }
+
+    /**
+     * Return the priority of this TypeConverter. TypeConverters with a high priority are chosen before low priority.
+     *
+     * @return integer
+     * @api
+     */
+    public function getPriority()
+    {
+        return 100;
+    }
+
+    /**
+     * @param mixed $source the source data
+     * @param string $targetType the type to convert to.
+     * @return boolean true if this TypeConverter can convert from $source to $targetType, false otherwise.
+     * @api
+     */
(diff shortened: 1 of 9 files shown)


Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        if (class_exists($targetType)) {
+            switch ($sourceType) {
+                case 'array':
+                    return method_exists($targetType, 'fromArray');
+                case 'string':
+                    return method_exists($targetType, 'fromString');
+                case 'bool':
+                case 'boolean':
+                    return method_exists($targetType, 'fromBool');
```
bwaidelich, line 85: in the [Neos\EventSourcing\EventStore\Normalizer\ ValueObjectNormalizer(https://github.com/neos/Neos.EventSourcing/blob/ce024233297f06bf79334e94332c98208cac3564/Classes/EventStore/Normalizer/ValueObjectNormalizer.php#L55) we expect the named constructors to be called `fromBoolean` and `fromFloat`.
We also check if they are public static and expect the respective type as single argument.

At least the name should be in sync IMO

bwaidelich (changes requested):

mficzel:
Why 5.3? This sounds like a feature to me. And a great one.

albe (commented):
Generally approve apart from the comment by Bastian and the target version. 5.3 is EOL in april and security only by now

nezaniel:
The background is that I want to remove the otherwise necessary dark magic from "Go Go Value Object!" https://github.com/sitegeist/Sitegeist.InspectorGadget before its soonish release. It was created for 5.1 and is used in 5.x projects, so it would be nice to have it in 5.3 as well. But I totally forgot that 5.3 is security fixes only, so these projects need to be updated anyway. New target would be 7.3 then.

As to whether it is a feature: yes, it is, the argument for it being treated as a bugfix back then was that we this way could get it into older versions, although that way is of course a bit fishy.

I renamed the expected constructor methods in the meantime. Validation for it being a proper static constructor makes sense of course, but it also seems rather complex when I look at the implementation in the EventSourcing package...

sorenmalling:
PSR-12 teaches us to use short form of types

https://www.php-fig.org/psr/psr-12/#25-keywords-and-types

Questioning, which one of the forms we should follow, to keep the CGL in sync?!

nezaniel:
@sorenmalling to be more PSR consistent was the original implementation's intent iirc. If we move to PSR-12 (see https://discuss.neos.io/t/rfc-updated-default-code-style/5836/7), I'd say we rather go with ::fromBool than ::fromBoolean, which would affect the ES package of course.

To clean up the version mixup: InspectorGadget was implemented for **Neos** 5.x, so the PR should have gone to **Flow** 6.3 from the beginning on. My bad

bwaidelich:
PSR-12 doesn't say anything about the corresponding method names. I'm +1 on using the short version for type hints (obviously) but in Method names we explicitly want to [avoid abbreviations](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartV/CodingGuideLines/PHP.html#naming)

IMO there's absolutely nothing wrong with `fromBoolean(bool $foo)` and changing that in the ES package would be a mean breaking change

bwaidelich:
btw: I'm also +1 for targeting `master` but maybe we can add a mini package with a custom TypeConverter that can be installed with < 7.x

albe:
> To clean up the version mixup: InspectorGadget was implemented for **Neos** 5.x, so the PR should have gone to **Flow** 6.3 from the beginning on. My bad

I might get convinced to target 6.3 with this. It is a kind of feature, but it is also something I would expect to work (serializing VOs with JsonArrayType) out of the box, so with one eye closed it could be considered a bugfix.
So if no one has a strong veto to get it in as a bugfix, I'm okay with that.

mhsdesign:
or get this into 8.0 as feature?

mficzel:
I also consider this more like a feature. Adding this retrospectively is risky. Better to target 8.0 that is around the corner.

mficzel:
Rebased this on master as #2762 

If a package has to use dirty aspects now i suggest to release a version 1.0 with the aspect (for Neos 6-3-7.3) and a version 2.0 without the aspect for Neos 8+.

mficzel:
Closed as the changes from here are now in master via #2762 thank you for this great feature.

Issue #2714, opened 2022-02-28
UriConstraints don't apply default port if it differs from the current port


bwaidelich:
### Description

When using `UriConstraints::fromUri()` (introduced with Flow 7.0) to create an instance and apply that to an URL with a different host/scheme, the port constraints of the target URL are ignored if they match the default port for the given scheme:

### Steps to Reproduce

```php
(string)UriConstraints::fromUri(new Uri('https://some-other-domain.tld'))
    ->applyTo(new Uri('https://some-domain.tld:8080'), false);
```

#### Expected behavior

'https://some-other-domain.tld'

#### Actual behavior

'https://some-other-domain.tld:8080'

### Affected Versions

Flow: 7.0+

PR #2704, opened 2022-02-23
TASK: Reenable psalm static analysis

albe:
WIP to get psalm builds working again with increased minimum PHP version.

PR #2715, opened 2022-02-28
BUGFIX: Fix UriConstraints port constraints for default ports

bwaidelich:
Previously, if `UriConstraints` were applied to an URL with a non-default
port (e.g. "8080") this port constraint was applied to the target URL even
if no explicit port constraint was set.

Fixes: #2714

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
@@ -86,8 +86,8 @@ final class UriConstraints
         if ($uri->getPort() !== null) {
             $constraints[self::CONSTRAINT_PORT] = $uri->getPort();
         }
-        if ($uri->getPath() !== '') {
-            $constraints[self::CONSTRAINT_PATH] = $uri->getPath();
+        if ($uri->getPort() !== null || $uri->getScheme() !== '') {
+            $constraints[self::CONSTRAINT_PORT] = $uri->getPort() ?? UriHelper::getDefaultPortForScheme($uri->getScheme());
         }
         if ($uri->getQuery() !== '') {
             $constraints[self::CONSTRAINT_QUERY_STRING] = $uri->getQuery();
--- a/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
+++ b/Neos.Flow/Classes/Mvc/Routing/Dto/UriConstraints.php
@@ -83,11 +83,11 @@ final class UriConstraints
             $constraints[self::CONSTRAINT_SCHEME] = $uri->getScheme();
             $constraints[self::CONSTRAINT_HOST] = $uri->getHost();
         }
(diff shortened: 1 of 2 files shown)


albe:
How does this relate to https://github.com/neos/flow-development-collection/pull/2654 and https://github.com/neos/flow-development-collection/pull/2276? Sorry, haven't looked into this PR details yet, but had those other coming up in my head that seemed related. I can take a look this evening unless you have some clear separation already up your sleeve :)

bwaidelich:
@albe good question! They're related indeed, this one is complemental:

#2654 makes sure that

```php
UriConstraints::create()
    ->withScheme('https')
    ->applyTo(new Uri('http://some-domain.tld'), false);
```
results in `https://some-domain.tld` (rather than `https://some-domain.tld:80`)

#2276 makes sure that

```php
UriConstraints::create()
    ->withPort(8080)
    ->applyTo(new Uri('http://some-domain.tld:8080'), true);
```
results in `http://some-domain.tld:8080` (rather than `http://some-domain.tld`)

and this PR makes sure that

```php
UriConstraints::fromUri(new Uri('https://some-other-domain.tld'))
    ->applyTo(new Uri('https://some-domain.tld:8080'), false);
```
results in `https://some-other-domain.tld` (rather than `https://some-other-domain.tld:8080`).


This whole UriConstraints topic got rather confusing because there are much more cases than I had anticipated at first.
But I'm pretty sure that all three fixes make sense

bwaidelich:
I don't understand why the Psalm test fails

> ERROR: MissingFile - bin/psalm:112:9 - Cannot find file phpvfscomposer:/home/runner/work/flow-development-collection/flow-development-collection/flow-development-distribution/Packages/Libraries/vimeo/psalm/psalm to include (see https://psalm.dev/107)
>         include("phpvfscomposer://" . __DIR__ . '/..'.'/Packages/Libraries/vimeo/psalm/psalm');

But I don't think that it's related to this change

albe:
> But I don't think that it's related to this change

Yes, unrelated. This error is around still some time now - see #2704 which tries to solve the issue in master after psalm has been disabled for now.

Thanks for the clear explanation! The first two make a lot of sense. The third one, which this PR targets is not so straightforward.

Given `UriConstraints::fromUri(new Uri('https://some-other-domain.tld'))->applyTo(new Uri('https://some-domain.tld:8080'), false);` I wouldn't even know intuitively what I'd expect 🤔 The (new) logic is then "If the constraint URI has a scheme, it's default port is applied unless an explicit port is set in the constraint URI, but explicit ports in the target URI are ignored". The latter feels weird. If someone has somehow configured the URI to be port 8080 on https, that must have good reasons, no? I'm likely just missing the actual use-case, because I never came across such a scenario with weird ports (lucky me?!)

kdambekalns (approved):
Sounds reasonable.

albe (approved):
Thanks for the explanation! :)

PR #2719, opened 2022-03-02
BUGFIX: Use configured pdo cache tables during flush and garbage collection

Sebobo:
**What I did**

Without this patch using custom table names for the PDO cache backend would cause problems
as some queries didn’t respect the given table names but used the defaults during DELETE
operations. This has been adjusted.

**How I did it**

Use the configured table names as in all the other queries.

**How to verify it**

Use the backend with custom table names and try to flush them. Without the change an error would occur if the default tables have never been used and setup.

**Checklist**


kdambekalns (approved):
Ouch! Thanks for spotting and fixing!

jonnitto (approved):

PR #2720, opened 2022-03-02
FEATURE: Implement `flushByTags` for the PDO cache backend

Sebobo:
**What I did**

With this change the `flushByTags` is optimised to run in batches. 
The maximum batch size can and should be configured based on the used data source.

This change relies on the interface changes in #2718 

**How I did it**

Instead of querying identifiers for each tag separately, they are now queried in batches
for a number of tags and then deleted.

**How to verify it**

Configure PDO backend for the Neos content cache and force a flush by publishing something.

**Checklist**


Sebobo:
Still a draft, as I want to check out if we can combine the SELECT with the first DELETE query by using a JOIN.

bwaidelich:
```sql
DELETE c FROM cache c JOIN tags t ON t.identifier = c.identifier WHERE t.tag IN ('tag1', 'tag2');
```
should work

Sebobo:
@bwaidelich can't get it to work. Always get syntax error with the JOIN even though the same query works when directly executed in MySQL 😕

bwaidelich:
@Sebobo can you share your WIP?

Sebobo:
I tried various combinations of aliases etc., maybe its easier via screen share. I'll ping you.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Caching.rst
```diff
@@ -479,10 +479,22 @@ Options
...
+| username       | Username to use for the database connection.       | No        |        |         |
 +----------------+----------------------------------------------------+-----------+--------+---------+
 | password       | Password to use for the database connection.       | No        |        |         |
 +----------------+----------------------------------------------------+-----------+--------+---------+
+| cacheTableName | Table name to store cache entries                  | No        | string | cache   |
++----------------+----------------------------------------------------+-----------+--------+---------+
+| tagsTableName  | Table name to store cache tags                     | No        | string | tags    |
++----------------+----------------------------------------------------+-----------+--------+---------+
+| batchSize      | Maximum number of parameters per query for         | No        | int    | 999     |
+|                | for batch operations.                              |           |        |         |
```
kdambekalns, line 491:
```suggestion
|                | batch operations.                                  |           |        |         |
```

PR #2721, opened 2022-03-02
FEATURE: Optimised Redis cache backend

Sebobo:
**What I did**

With this change the `flushByTags` is optimised to run flush operations in batches. 
The maximum batch size can and should be configured based on the used data source.

This change relies on the interface changes in #2718

Additionally all optimisations from https://github.com/sandstorm/OptimizedRedisCacheBackend were added.

**How I did it**

Instead of uploading the flush Lua script and only flushing the entries for one tag.
The same is now done for batches of tags reducing the calls to Redis and the 
evaluation time Redis needs to parse the script.

**How to verify it**

#2718 includes updated tests for the Redis backend that will also check this method.

**Checklist**

- [x] Tests have been created, run and adjusted as needed -> see #2718 

### DIFF
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -631,8 +631,9 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
         if (!$this->entryIterator) {
             $prefix = $this->getPrefixedIdentifier('entry:');
             $prefixLength = strlen($prefix);
-            $keys = array_map(static fn (string $key) => substr($key, $prefixLength), $this->redis->keys($prefix . '*'));
-            $this->entryIterator = new \ArrayIterator($keys);
+            $keys = $this->redis->keys($prefix . '*') ?? [];
+            $entryIdentifiers = array_map(static fn (string $key) => substr($key, $prefixLength), $keys);
+            $this->entryIterator = new \ArrayIterator($entryIdentifiers);
         }
         return $this->entryIterator;
     }
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -254,11 +254,11 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
         // language=lua
         $script = "
         local keys = redis.call('KEYS', ARGV[1] .. '*')
-		for k1,key in ipairs(keys) do
-			redis.call('DEL', key)
-		end
-		";
-        $this->redis->eval($script, [$this->getPrefixedIdentifier('')], 0);
+        for k1,key in ipairs(keys) do
+            redis.call('DEL', key)
+        end
+        ";
+        $this->redis->eval($script, [$this->getPrefixedIdentifier('')], 2);
 
         $this->frozen = null;
     }
@@ -287,14 +287,14 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
         }
 
         $script = "
-		local entries = redis.call('SMEMBERS', KEYS[1])
-		for k1,entryIdentifier in ipairs(entries) do
-			redis.call('DEL', ARGV[1]..'entry:'..entryIdentifier)
-			redis.call('DEL', ARGV[1]..'tags:'..entryIdentifier)
-		end
-		redis.call('DEL', KEYS[1])
-		return #entries
-		";
+        local entries = redis.call('SMEMBERS', KEYS[1])
+        for k1,entryIdentifier in ipairs(entries) do
+            redis.call('DEL', ARGV[1]..'entry:'..entryIdentifier)
+            redis.call('DEL', ARGV[1]..'tags:'..entryIdentifier)
+        end
+        redis.call('DEL', KEYS[1])
+        return #entries
+        ";
         return $this->redis->eval($script, [$this->getPrefixedIdentifier('tag:' . $tag), $this->getPrefixedIdentifier('')], 1);
     }
 
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -290,6 +290,12 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
         local entries = redis.call('SMEMBERS', KEYS[1])
         for k1,entryIdentifier in ipairs(entries) do
             redis.call('DEL', ARGV[1]..'entry:'..entryIdentifier)
+
+            local tags = redis.call('SMEMBERS', ARGV[1]..'tags:'..entryIdentifier)
+            for k2,tagName in ipairs(tags) do
+                redis.call('SREM', ARGV[1]..'tag:'..tagName, entryIdentifier)
+            end
+
             redis.call('DEL', ARGV[1]..'tags:'..entryIdentifier)
         end
         redis.call('DEL', KEYS[1])
@@ -319,6 +325,12 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
             local entries = redis.call('SMEMBERS', KEYS[i])
             for k1,entryIdentifier in ipairs(entries) do
                 redis.call('UNLINK', ARGV[i]..'entry:'..entryIdentifier)
+
+                local tags = redis.call('SMEMBERS', ARGV[i]..'tags:'..entryIdentifier)
+                for k2,tagName in ipairs(tags) do
+                    redis.call('SREM', ARGV[i]..'tag:'..tagName, entryIdentifier)
+                end
+
                 redis.call('UNLINK', ARGV[i]..'tags:'..entryIdentifier)
             end
             redis.call('UNLINK', KEYS[i])
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -177,7 +177,7 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
      */
     private function calculateExpires(string $tag, int $lifetime): int
     {
-        $ttl = $this->redis->ttl($tag);
+        $ttl = (int)$this->redis->ttl($tag);
         if ($ttl < 0 || $lifetime === self::UNLIMITED_LIFETIME) {
             return -1;
         }
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -38,8 +38,8 @@ use Neos\Error\Messages\Result;
  *  - batchSize:       Maximum number of parameters per query for batch operations
  *
  * Requirements:
- *  - Redis 2.6.0+ (tested with 2.6.14 and 2.8.5)
- *  - phpredis with Redis 2.6 support, e.g. 2.2.4 (tested with 92782639b0329ff91658a0602a3d816446a3663d from 2014-01-06)
+ *  - Redis 6.0.0+
+ *  - phpredis with Redis 6.0 support
  *
  * Implementation based on ext:rediscache by Christopher Hlubek - networkteam GmbH
  *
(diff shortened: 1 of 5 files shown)


kitsunet:
Look neat! Thank you, does this superseed the [sandstorm opt](https://github.com/sandstorm/OptimizedRedisCacheBackend) or does that still have any _extras_ then?

Sebobo:
@kitsunet thanks. 

I only added the main new method in this PR, it doesn't cover all the other improvements they did. But the script is already based on the Sandstorm version of the single tag flush. I had a call yesterday with @skurfuerst and we also discussed to now bring over all improvements from the Sandstorm version.

I have a few more calls planned with people who modified the Redis backend to get their input too.

I would then put all those into a separate PR where everyone with improvements then can chime in.

Sebobo:
@kitsunet I now added in another commit the remaining optimisations from the sandstorm package. 
As my `flushByTags` is already based on their script and not on the `flushByTag` of the core version, I think it doesn't make sense to only optimise this one method and have it behave a bit different.

If the change should be "clean" I would need to make the `flushByTags` based on the old script.

Wdyt? Should I make a "clean" PR and then another for the other optimisations?
I mean the sandstorm version is running in a lot of projects afaik. 

If we go this way, I would of course adjust the PR title and description.

kdambekalns:
> Should I make a "clean" PR and then another for the other optimisations?

I remember faintly there were issues with that implementation when not used in a "strictly Neos" context, as it had cut some corners to optimize. Not sure if those are still existing, I'll try to research a bit more later today.

kdambekalns (commented):
Ok, I think the only potential issue with the `OptimizedRedisCacheBackend` left is the documented behaviour around leaving stale tags in place and the effect that could have on `findIdentifiersBytag()`. So if that is "copied"

Another thing I noticed: In the README of `OptimizedRedisCacheBackend` it says
> For implementing flush(), we just iterate over the relevant part of the keyspace and remove all elements. Before, this was done through a similar logic as explained in FlushByTag, because the Redis Backend was freezable (but we do not need this for Neos Content Cache).

So, if this is case, implementing `FreezableBackendInterface` should be removed from the backend, no?

jonnitto:
Trigger checks

kdambekalns:
I fixed the Psalm complaint, but the test failures look legit…

kdambekalns:
> > So, if this is case, implementing `FreezableBackendInterface` should be removed from the backend, no?
> 
> But the backend is not only for Neos? So I would leave it in there. Shouldn't hurt to keep it.

That's my point - if it's kept, it should behave like before, but I understand the README says it's differs… So either remove freezability or keep behaviour.

> What do you think about increasing the minimum Redis version to the 6.0.0 which is the oldest with EOL? Then we could also finally add some more advanced features in the nearer future than will also help with performance.

Fine with me…

Sebobo:
Will check

Sebobo:
@kdambekalns ok all tests run fine again locally and I implemented proper support for `freeze`, cleaning up all tags and it's iterable again.
I also added another test because the frozen state was not fully covered.

Now the implementation might be a tiny bit slower, but the main bottleneck is now anyway solved with `flushByTags`. In my tests with 500.000 tags, I couldn't get a difference.

I think if we want to have a specialised Neos Redis backend, that one should be in Neos.Neos and not in this package.

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -449,9 +512,15 @@ public function setCompressionLevel($compressionLevel): void
     }
 
     /**
-     * @param \Redis $redis
-     * @return void
+     * Sets the maximum number of items for batch operations
+     *
+     * @api
```
mhsdesign, line 517: hmm is this protected method supposed to be api?

```suggestion
```

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -265,33 +291,73 @@ public function flushByTag(string $tag): int
         }
 
         $script = "
```
mhsdesign, line 293: i would put those in `'LUA'` nowdocs, should have then ide language detection, but never mind ^^

kdambekalns, line 293:
```suggestion
        // language=lua
        $script = "
```

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -265,33 +291,73 @@ public function flushByTag(string $tag): int
...
     {
-        $flushed = 0;
-        foreach ($tags as $tag) {
-            $flushed += $this->flushByTag($tag);
+        if ($this->isFrozen()) {
+            throw new \RuntimeException(sprintf('Cannot add or modify cache entry because the backend of cache "%s" is frozen.', $this->cacheIdentifier), 1647642328);
         }
-        return $flushed;
+
+        $script = "
```
kdambekalns, line 325:
```suggestion
        // language=lua
        $script = "
```

kdambekalns (commented):
Two more tiny suggestions… and the class docblock still says:

```
 * Since Redis < 2.8.0 does not provide a mechanism for iterating over keys,
 * a separate list with all entries is populated
```

That is not true anymore?! At least the version mentioned is no longer supported, so _is there_ a better way?

kdambekalns (approved):
Notwithstanding my tiny remarks, this seems to work just fine… I only did limited testing (compared to what @Sebobo did), but… ✅

Neos.Cache/Classes/Backend/RedisBackend.php
Sebobo: You are right, mistake

Sebobo:
> Two more tiny suggestions… and the class docblock still says:
> 
> ```
>  * Since Redis < 2.8.0 does not provide a mechanism for iterating over keys,
>  * a separate list with all entries is populated
> ```
> 
> That is not true anymore?! At least the version mentioned is no longer supported, so _is there_ a better way?

You are right, that was my last change to clean that up after I had read it, but forgot to remove the text.
After working on this stuff for hours I'm half braindead 🧟‍♂️

mficzel:
After the updated docblocks i get "Typed property Neos\Cache\Backend\RedisBackend::$frozen must not be accessed before initialization" 

Does this rely on features from #2701

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -35,67 +35,53 @@
...
     /**
      * @var \Redis
      */
     protected $redis;
 
-    /**
-     * @var integer Cursor used for iterating over cache entries
-     */
-    protected $entryCursor = 0;
+    protected ?bool $frozen;
```
mhsdesign, line 64:
```suggestion
    protected ?bool $frozen = null;
```

mhsdesign: @mficzel nope but i assume we need a default type here:

Sebobo: Why does Psalm not report this in the checks?
For objects it was reported.

kdambekalns: Why is that nullable? I'd assume a cache is either frozen or not. So false sounds like a sane default…

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -35,67 +35,53 @@
...
     /**
      * @var \Redis
      */
     protected $redis;
 
-    /**
-     * @var integer Cursor used for iterating over cache entries
-     */
-    protected $entryCursor = 0;
+    protected ?bool $frozen = null;
```
kdambekalns, line 64:
```suggestion
    protected bool $frozen = false;
```
since the getter declares its return type as bool that would conflict anyway.

Sebobo: I assume it was done like that because when it's `null` the getter retrieves the value from Redis and "caches" it this way.

Neos.Cache/Classes/Backend/RedisBackend.php
Sebobo: See my comment below

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -139,28 +124,51 @@ public function set(string $entryIdentifier, string $data, array $tags = [], int
...
     /**
      * Loads data from the cache.
      *
      * @param string $entryIdentifier An identifier which describes the cache entry to load
-     * @return mixed The cache entry's content as a string or false if the cache entry could not be loaded
+     * @return bool|string The cache entry's content as a string or false if the cache entry could not be loaded
      * @api
      */
-    public function get(string $entryIdentifier)
+    public function get(string $entryIdentifier): string|bool
```
albe, line 171:
```suggestion
    public function get(string $entryIdentifier): string|false
```

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -310,28 +355,30 @@ public function findIdentifiersByTag(string $tag): array
     /**
      * {@inheritdoc}
      */
-    public function current()
+    public function current(): string|bool
```
albe, line 358:
```suggestion
    public function current(): string|false
```

albe (commented):
Don't know much about redis and I can't process the lua scripts in a way to decide if they are sane or not. So I just left three suggestions to use union `false` pseudo type instead of the more generic `bool`: https://www.php.net/manual/de/language.types.declarations.php#language.types.declarations.composite.union.false

mficzel (approved):
Looks good to me and works on my machine. Am no redis expert

skurfuerst:
awesome work everybody ❤️ ❤️ ❤️ 

I'll add docs to OptimizedRedisCacheBackend, marking it as obsolete starting with Neos 8.0 💯

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -217,33 +227,26 @@ public function remove(string $entryIdentifier): bool
...
-		redis.call('DEL', KEYS[1])
-		redis.call('DEL', KEYS[2])
-		";
-        $this->redis->eval($script, [$this->getPrefixedIdentifier('entries'), $this->getPrefixedIdentifier('frozen'), $this->getPrefixedIdentifier('')], 2);
+        local keys = redis.call('KEYS', ARGV[1] .. '*')
+        for k1,key in ipairs(keys) do
+            redis.call('DEL', key)
+        end
+        ";
+        $this->redis->eval($script, [$this->getPrefixedIdentifier('frozen'), $this->getPrefixedIdentifier('')], 1);
```
skurfuerst, line 241: `$this->getPrefixedIdentifier('frozen')` is not needed anymore, is it? At least KEYS[1] is never accessed.

Sebobo: It is still needed, but properly the script is incorrect now. Will check and adjust the test. Thx for noticing!

Sebobo: Ok, you are right, the key is already included in the args with wildcard.

Sebobo:
> awesome work everybody ❤️ ❤️ ❤️
> 
> I'll add docs to OptimizedRedisCacheBackend, marking it as obsolete starting with Neos 8.0 💯

Hi @skurfuerst, thx. 

There is still a difference to your package as I enabled the removal of related tags again, which you didn't do in the "optimised" version. 
It might not be a problem as long as `flushByTags` is used, as its so much faster, but there was a reason for the removal I guess. But you should only mark it obsolete if it behaves same or better in production related to the tag removal behaviour.

There might still be a need for a `Neos Content Cache` specific Redis backend implementation that does some things a bit differently than the general one.

bwaidelich:
This change made it into `neos/flow` [7.3.8](https://github.com/neos/flow-development-collection/blob/7.3.8/Neos.Cache/Classes/Backend/RedisBackend.php) but it seems to be missing from [7.3.9](https://github.com/neos/flow-development-collection/blob/7.3.9/Neos.Cache/Classes/Backend/RedisBackend.php)

kdambekalns:
Hm.

- The diff doesn't confirm: https://github.com/neos/flow-development-collection/compare/7.3.8...7.3.9
- Same in the split: https://github.com/neos/flow/compare/7.3.8...7.3.9

**The 7.3.8 version was tagged on the 8.2 branch** AFAICT.

kdambekalns:
And 7.3.7 was tagged in HEAD off of 7.3 and later merged back into 7.3 (which is confusing, but correct.)

Issue #2724, opened 2022-03-08
PDOException on running flow command database:setcharset with PHP8 / MariaDB


vixe76:
### Description

Running `./flow database:setcharset` with MariaDB / MySQL results in an PDOException

```
Exception in line 1853 of /data/Packages/Libraries/doctrine/dbal/lib/Doctrine/DBAL/Connection.php: There is no active transaction

13 PDO::rollBack()
12 Doctrine\DBAL\Connection::rollBack()
11 Neos\Flow\Command\DatabaseCommandController_Original::convertToCharacterSetAndCollation("utf8mb4", "utf8mb4_unicode_ci", NULL, false)
10 Neos\Flow\Command\DatabaseCommandController_Original::setCharsetCommand("utf8mb4", "utf8mb4_unicode_ci", NULL, false)
9 Neos\Flow\Cli\CommandController_Original::callCommandMethod()
8 Neos\Flow\Cli\CommandController_Original::processRequest(Neos\Flow\Cli\Request, Neos\Flow\Cli\Response)
7 Neos\Flow\Cli\Dispatcher_Original::dispatch(Neos\Flow\Cli\Request, Neos\Flow\Cli\Response)
6 Neos\Flow\Cli\CommandRequestHandler::Neos\Flow\Cli\{closure}()
5 Closure::__invoke()
4 Neos\Flow\Security\Context_Original::withoutAuthorizationChecks(Closure)
3 Neos\Flow\Cli\CommandRequestHandler::handleRequest()
2 Neos\Flow\Core\Bootstrap::run()
1 require("/data/Packages/Framework/Neos.Flow/Scripts/flow.php")
```

### Affected Versions

Flow: 7.3

sorenmalling:
@vixe76 Can you try and use the `$output` argument of the setcharset command, and execute the generated SQL directly to the databsae, to see if we get a error from maraidb?

vixe76:
The generated SQL runs without any error because there isn't a transaction contained. 
I think the cause is changed PDO transaction handling in PHP 8 for some statements (CREATE TABLE, ALTER TABLE, etc.).

https://www.doctrine-project.org/projects/doctrine-migrations/en/3.3/explanation/implicit-commits.html

bwaidelich:
This can be fixed in the command handler by removing the lines related to transactions (https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Command/DatabaseCommandController.php#L137)

[...]

PR #2725, opened 2022-03-08
BUGFIX: Don't use transactions to change db character set

bwaidelich:
Fixes: #2724


kdambekalns (approved):
Yup, makes sense. Obviously this code is older than my knowledge about MySQL not being transactional with DML statements… 🤷‍♂️

albe:
> Yup, makes sense. Obviously this code is older than my knowledge about MySQL not being transactional with DML statements… 🤷‍♂️

Yeah, that was something to chew on a year or so ago for me as well :D

albe (approved):

PR #2730, opened 2022-03-11
TASK: Fix some psalm complaints, update baseline

kdambekalns:


Neos.Flow/Classes/Mvc/ActionRequest.php
```diff
@@ -494,7 +494,7 @@ public function getControllerActionName(): string
         $controllerObjectName = $this->getControllerObjectName();
         if ($controllerObjectName !== '' && ($this->controllerActionName === strtolower($this->controllerActionName))) {
             $controllerClassName = $this->objectManager->getClassNameByObjectName($controllerObjectName);
-            $lowercaseActionMethodName = strtolower($this->controllerActionName) . 'action';
+            $lowercaseActionMethodName = $this->controllerActionName . 'action';
```
bwaidelich, line 497: removing the `strtolower` is related to psalm??

kdambekalns: Yes.

```
ERROR: RedundantFunctionCall - Packages/Framework/Neos.Flow/Classes/Mvc/ActionRequest.php:497:42 - The call to strtolower is unnecessary (see https://psalm.dev/280)
            $lowercaseActionMethodName = strtolower($this->controllerActionName) . 'action';
```

And it's true, because that block is only run if  `$this->controllerActionName === strtolower($controllerActionName)`…

bwaidelich: Wow. Psalm is cleverer than me o.O

bwaidelich (approved):
+1 by reading

mhsdesign (approved):
Fine by reading

bwaidelich:
For reference: This change broke PHP 8.1 compatibility:

> PHP Deprecated:  trim(): Passing null to parameter #1 ($string) of type string is deprecated in Packages/Framework/Neos.Flow/Scripts/flow.php on line 68

This is addressed with #2796 (but not yet released for 7.3)

PR #2733, opened 2022-03-13
BUGFIX: Eel Helper `Array.push()` auto cast string to array. #2710

### LINKED ISSUE #2710: Eel Helper `Array.push` auto cast string to array.
### Description

The eel helper `Array.concat` auto casts non array arguments to arrays, how about to do the same 'magic' for `Array.push` ?

### Steps to Reproduce

use `Array.push` on a string like `Array.push("foo", "bar")`

#### Expected behavior

expect some auto cast magic, so we get `["foo", "bar"]` as result.

#### Actual behavior

type php error, `iterable` is expected.

## Usecase in Neos.Fusion:

this way one could add via `@process` in fusion super easily an additional css class, even if the class was a string before:

currently: https://github.com/neos/neos-development-collection/blob/a48994472e5dd0fec137f41a51bca5a710879c86/Neos.Neos/Resources/Private/Fusion/Prototypes/ContentCollection.fusion#L32

**then, given any of these fusion configurations:**

**1.)**
`attributes.class = "class1 class2"`
**2.)**
`attributes.class = ${["class1", "class2"]}`
**3.)**
```
attributes.class = Neos.Fusion:DataStructure {
    class1 = "class1"
    class2 = "class2"
}
```

**the following process would be working for each config despite array or string**
`attributes.class.@process.addClass = ${Array.push(value, "class3")}`

after `@process`

**1.)**
`["class1 class2", "class3"]`
**2.)**
`["class1", "class2", "class3"]`
**2.)**
`["class1" => "class1", "class2" => "class2", 0 => "class3"]`

the rendered html would be for all the same:
`class="class1 class2 class3"`

mhsdesign:
fixes #2710

same lose behaviour exists already for `Array.concat()`

### DIFF
--- a/Neos.Eel/Classes/Helper/ArrayHelper.php
+++ b/Neos.Eel/Classes/Helper/ArrayHelper.php
@@ -342,14 +342,20 @@ class ArrayHelper implements ProtectedContextAwareInterface
      *
      *     Array.push(array, e1, e2)
      *
-     * @param iterable $array
+     * @param iterable|scalar $array
      * @param mixed $element
      * @return array The array with the inserted elements
      */
-    public function push(iterable $array, $element): array
+    public function push($array, $element): array
     {
         if ($array instanceof \Traversable) {
             $array = iterator_to_array($array);
+        }
+        if (is_scalar($array)) {
+            $array = [$array];
+        }
+        if (is_array($array) === false) {
+            throw new \InvalidArgumentException('$array must be of type iterable|scalar got: ' . gettype($array), 1647595715);
         }
         $elements = func_get_args();
         array_shift($elements);
--- a/Neos.Eel/Tests/Unit/Helper/ArrayHelperTest.php
+++ b/Neos.Eel/Tests/Unit/Helper/ArrayHelperTest.php
@@ -423,6 +423,9 @@ class ArrayHelperTest extends \Neos\Flow\Tests\UnitTestCase
             'string keys' => [['foo' => 'bar', 'baz' => 'foo', 'bar' => 'baz'], 42, 'foo', ['foo' => 'bar', 'baz' => 'foo', 'bar' => 'baz', 42, 'foo']],
             'mixed keys' => [['bar', '24' => 'foo', 'i' => 181.84, 'foo' => 'abc', '84216', 76, 'k' => 53], 42, 'foo', ['bar', '24' => 'foo', 'i' => 181.84, 'foo' => 'abc', '84216', 76, 'k' => 53, 42, 'foo']],
             'traversable' => [TestArrayIterator::fromArray(['a']), 'b', 'c', ['a', 'b', 'c']],
+            # expect cast scalar (as arg $array) to array
+            'string' => ['a', 'b', 'c', ['a', 'b', 'c']],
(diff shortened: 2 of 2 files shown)


mhsdesign:
im also fine when this should be merged as a feature in master, but its a nice to have utility

Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -342,15 +342,21 @@ public function pop(iterable $array): array
      *
      *     Array.push(array, e1, e2)
      *
-     * @param iterable $array
+     * @param iterable|scalar $array
      * @param mixed $element
      * @return array The array with the inserted elements
      */
-    public function push(iterable $array, $element): array
+    public function push($array, $element): array
```
mhsdesign, line 349: or should this better be target `master` with php 8 union types:

```suggestion
    public function push(iterable|string|int|float $array, $element): array
```

jonnitto: I would not use union types. It can also be possible that you push Arrays, Booleans, Objects (Like node), etc.

Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -342,15 +342,21 @@ public function pop(iterable $array): array
...
     {
         if ($array instanceof \Traversable) {
             $array = iterator_to_array($array);
         }
+        if (is_scalar($array)) {
```
jonnitto, line 354: With `is_scalar` it would not be possible to push objects like `node` or resources, right?

Neos.Eel/Classes/Helper/ArrayHelper.php
mhsdesign: sorry i might have been missleading, this pr is not about the things ($element) to push in, but the $array where to push onto ...

so this pr is to loosen the `iterable $array` and make it more flexible to call `Array.push()` on not only traversable see: https://github.com/neos/flow-development-collection/issues/2710

currently this works:

`Array.push(['foo'], 'bar')`  -  `['foo', 'bar']`
`Array.push(['foo'], true)`  -  `['foo', true]`
`Array.push(['foo'], node)`  -  `['foo', node]`

i also want to allow this:

`Array.push('foo', 'bar')`  -  `['foo', 'bar']`

Neos.Eel/Classes/Helper/ArrayHelper.php
mhsdesign: see comment above ^^ 
https://github.com/neos/flow-development-collection/pull/2733#discussion_r829944067

jonnitto (approved):

Neos.Eel/Classes/Helper/ArrayHelper.php
albe: We can check how it behaves with union types in Flow 8+ - we still don't properly support union types there when reflection is involved (which it isn't here IIRC).

albe (approved):

mhsdesign:
so this can be merged then as 'Bugfix' ?
oops before that i should add at least one test ^^

PR #2736, opened 2022-03-14
TASK: Cleanup PackageManager->findComposerPackagesInPath()

mhsdesign:
- use \DirectoryIterator for iterating over files.

- fix unit test: `self::assertEquals(sort($foo), sort($bar));` will always pass, since `sort()` returns not the sorted value but `bool`.

- cleanup unit tests to use `getMockBuilder`

- add test for package collections: `neos-package-collection`

- Solve TODO: `Maybe get rid of magic string "neos-package-collection" by fetching collection package types from outside.` by using AppendIterator and `->append` the packages of the collection.

### DIFF
--- a/Neos.Flow/Classes/Package/PackageManager.php
+++ b/Neos.Flow/Classes/Package/PackageManager.php
@@ -570,8 +570,20 @@ class PackageManager
     protected function scanAvailablePackages(): array
     {
         $newPackageStatesConfiguration = ['packages' => []];
-        foreach ($this->findComposerPackagesInPath($this->packagesBasePath) as $packagePath) {
+
+        $packages = new \AppendIterator();
+        $packages->append(new \NoRewindIterator(self::findComposerPackagesInPath($this->packagesBasePath)));
+
+        foreach ($packages as $packagePath) {
             $composerManifest = ComposerUtility::getComposerManifest($packagePath);
+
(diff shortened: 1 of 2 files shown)


albe:
From my experience it's next to always worth it getting smaller improvements in ASAP, even if a "better solution" is already scetched out. Most of the times, the better solution takes quite a while to get done done, so until then people (including us maintaining) would already benefit from slightly better code.

mhsdesign:
Nice 👍 i think the order of how packages will be scanned might be changed, as the `neos-package-collection` types will be traversed always last. Need to figure out if this matters ...

mhsdesign:
Would that be an issue that the package scan order changes slightly?

When one has those packages:

```
P1
P2 (Collection)
 - P3
P4 
```

i think currently they will be scanned as:

P1, P3, P4


but with this PR, due to the use of appending the iterator, i think the order changes to 

P1, P4, P3

mhsdesign:
as once discussed with @kitsunet we should reference all flow packages in the distributions composer manifest and let flow find them via this logic than looking inside of devcollections for package.jsons

see also https://github.com/neos/flow-development-collection/issues/2831

kitsunet:
> Would that be an issue that the package scan order changes slightly?

As we create a sorted version as well it shouldn't really matter I guess.

kitsunet (approved):

kdambekalns (approved):
Looks good.

PR #2744, opened 2022-03-16
!!! FEATURE: Introduce ActionUriBuilder

bwaidelich:
Introduce `ActionUriBuilder` as more solid building block to create relative
and absolute URLs for MVC actions.

This change should not be breaking, but it deprecates some classes that were
previously part of the public API:

* Deprecate `UriBuilder` (in favor of the new `ActionUriBuilder`)
* Deprecate `BaseUriProvider` and the use of the `Neos.Flow.http.baseUri` setting
* Deprecate `ControllerContext`
* Deprecate `Environment::isRewriteEnabled()`

### Obtain an instance:

In Controllers (extending `AbstractController`) the new `ActionUriBuilder` is already available via `$this->actionUriBuilder`.
For other usecases, the `ActionUriBuilderFactory` can be used to obtain an instance:

```php
// inject the factory via constructor injection

public function __construct(private ActionUriBuilderFactory $uriBuilderFactory) {}

// ...or property injection:

/**
 * @var ActionUriBuilderFactory
 * @Flow\Inject(lazy=false)
 */
protected ActionUriBuilderFactory $uriBuilderFactory;

// obtain an instance from an action request

$uriBuilder = $uriBuilderFactory->createFromActionRequest($actionRequest);

// ...from an HTTP request

$uriBuilder = $uriBuilderFactory->createFromHttpRequest($httpRequest);

// ...from a custom base URI (e.g. for CLI usage):

$uriBuilder = $uriBuilderFactory->createFromBaseUri(new Uri('http://localhost'));
```

### Build action URLs:

```php
// relative URL to a specified action
$this->actionUriBuilder->uriFor(Action::create($packageKey, $controllerName, $actionName));

// absolute URL to a different action of the current controller
$this->actionUriBuilder->absoluteUriFor(Action::fromActionRequest($this->request)->withActionName('anotherAction'));

// relative URL to the current action with some additional arguments
$this->actionUriBuilder->uriFor(Action::fromActionRequest($this->request)->withAdditionalArguments(['order' => true]));

// custom format
$this->actionUriBuilder->uriFor(Action::fromActionRequest($this->request)->withFormat('xml'));

// custom query parameters / fragment (not part of the route)
$this->actionUriBuilder->uriFor(Action::fromActionRequest($this->request))->withQuery('foo=bar')->withFragment('baz');
```

Related: #2157


mhsdesign:
Hi what is the state of this, ist this ready to be reviewed and ready for 8.2?

Neos.Flow/Classes/Mvc/Routing/Dto/Action.php
```diff
@@ -0,0 +1,94 @@
...
+final class Action
+{
+    private function __construct(
+        private string $actionName,
+        private string $controllerName,
+        private string $packageKey,
+        /** @deprecated with Flow 8.0 - The use of subpackage keys is discouraged and only supported for backwards compatibility */
+        private ?string $subpackageKey,
+        private string $format,
+        private array $additionalArguments,
```
mficzel, line 30: Why additionalArguments and not arguments

mficzel:
Generally I think the uriFor method should have the distinct parameters “routingArguments” and “queryParameters“. 

I know this can be done via -> withQuery but it could help to make the distinction clear.

Neos.Flow/Classes/Http/BaseUriProvider.php
```diff
@@ -12,6 +12,7 @@
  * Supports to get a baseUri from various possible sources.
  *
  * @Flow\Scope("singleton")
+ * @deprecated with Flow 8.0 - use the current HTTP request instead, e.g. via $actionRequest->getHttpRequest() or $bootstrap->getActiveRequestHandler()->getHttpRequest()
```
bwaidelich, line 15:
```suggestion
 * @deprecated with Flow 8.2 - use the current HTTP request instead, e.g. via $actionRequest->getHttpRequest() or $bootstrap->getActiveRequestHandler()->getHttpRequest()
```

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -38,16 +40,28 @@
 abstract class AbstractController implements ControllerInterface
 {
     /**
+     * @deprecated with Flow 8.0 - use $this->actionUriBuilder instead
```
bwaidelich, line 43:
```suggestion
     * @deprecated with Flow 8.2 - use $this->actionUriBuilder instead
```

Neos.Flow/Classes/Mvc/Routing/Dto/Action.php
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * @Flow\Proxy(false)
+ */
+final class Action
+{
+    private function __construct(
+        private string $actionName,
+        private string $controllerName,
+        private string $packageKey,
+        /** @deprecated with Flow 8.0 - The use of subpackage keys is discouraged and only supported for backwards compatibility */
```
bwaidelich, line 27:
```suggestion
        /** @deprecated with Flow 8.2 - The use of subpackage keys is discouraged and only supported for backwards compatibility */
```

Neos.Flow/Classes/Mvc/Routing/Dto/Action.php
```diff
@@ -0,0 +1,94 @@
...
+        return new self($actionName, $controllerName, $packageKey, null, '', []);
+    }
+
+    public static function fromActionRequest(ActionRequest $request): self
+    {
+        return new self($request->getControllerActionName(), $request->getControllerName(), $request->getControllerPackageKey(), $request->getControllerSubpackageKey(), $request->getFormat(), []);
+    }
+
+    /**
+     * @deprecated with Flow 8.0 - The use of subpackage keys is discouraged and only supported for backwards compatibility
```
bwaidelich, line 45:
```suggestion
     * @deprecated with Flow 8.2 - The use of subpackage keys is discouraged and only supported for backwards compatibility
```

Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
```diff
@@ -23,7 +23,7 @@
 /**
  * An URI Builder
  *
- * @api
+ * @deprecated with Flow 8.0 - use the new ActionUriBuilder instead
```
bwaidelich, line 26:
```suggestion
 * @deprecated with Flow 8.2 - use the new ActionUriBuilder instead
```

Neos.Flow/Classes/Utility/Environment.php
```diff
@@ -106,6 +106,7 @@ public function getMaximumPathLength(): int
      * Whether or not URL rewriting is enabled.
      *
      * @return bool
+     * @deprecated with Flow 8.0 - URL rewriting is now assumed to be supported
```
bwaidelich, line 109:
```suggestion
     * @deprecated with Flow 8.2 - URL rewriting is now assumed to be supported
```

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -16,13 +16,8 @@ Neos:
...
       applicationToken: 'MinorVersion'
 
-      # Defines an explicit base URI that should be used. This affects
-      # resource management, routing and all other parts accessing
-      # "Http\Request->getBaseUri()".
-      #
-      # If this setting is NULL, the base URI will be detected from the current
-      # request (default).
-      baseUri: NULL
+      # Deprecated with Flow 8.0
```
bwaidelich, line 19:
```suggestion
      # Deprecated with Flow 8.2
```

Neos.Flow/Classes/Mvc/Controller/ControllerContext.php
```diff
@@ -24,7 +24,7 @@
  * for conveniently passing the information to other classes who need it, usually
  * views being views or view helpers.
  *
- * @api
+ * @deprecated with Flow 8.0 - The controller context is an internal concept that should be avoided in userland code
```
bwaidelich, line 27:
```suggestion
 * @deprecated with Flow 8.2 - The controller context is an internal concept that should be avoided in userland code
```

Neos.Flow/Classes/Mvc/Routing/Dto/Action.php
bwaidelich: technically because `@action`, `@controller`, etc. are _all_ arguments. But from this DTO point of view, they are just arguments, I agree!

bwaidelich:
> Generally I think the uriFor method should have the distinct parameters “routingArguments” and “queryParameters“.

I agree. Query parameters are not part of the routing framework, but they are definitely a concern of the URI building. Also the `UriInterface::withQuery()` interface is cumbersome.

I'm just not sure whether it should be another argument to the `uriFor` method (I tried to introduce more single argument methods with this change) or a separate method on the builder, e.g.:
```php
$this->actionUriBuilder
  ->uriFor(
      Action::fromActionRequest($this->request)->withArguments(['order' => true]),
      ['foo' => 'bar']
  );
```
vs
```php
$this->actionUriBuilder
  ->uriFor(
    Action::fromActionRequest($this->request)->withArguments(['order' => true])
  )
  ->withQueryParameters(['foo' => 'bar']);
```

mficzel:
@bwaidelich I think it is important to make clear that there are routingParameters and that you can also have queryParameters at the same time.

Both routingParameters and queryParameter should be on the same api level so developers that see one also implicitly see the other. So i would think both belong to the `Action-DTO` which could have `withRouteParameters`,  `withQueryParameters` plus  `withAdditionalRouteParameters`,  `withAdditionalQueryParameters`. I personally would add both arguments also to the create method this is a complex task and justifies a more complex api.

bwaidelich:
> I think it is important to make clear that there are routingParameters and that you can also have queryParameters

Yes, but for the same reason I would suggest to stick to routing**Arguments**  vs query**Parameters** (even though we mixed up the two in the past in a couple of places).

> So i would think both belong to the `Action-DTO`

That was my first impression, too.. But the DTO in the current form describes a MVC action and that has nothing to do with query parameters.
But maybe it's just a naming issue. I wasn't 100% happy with the name "Action" anyways

mficzel:
> But maybe it's just a naming issue. I wasn't 100% happy with the name "Action" anyways

Agree, if it is called "Action" i would expect it to separate routingArguments and queryParameters internally which is the problem we are currently stuck in. Maybe "UriBuildingRequest" or "ActionUriSpecification"

albe:
> > But maybe it's just a naming issue. I wasn't 100% happy with the name "Action" anyways
> 
> Agree, if it is called "Action" i would expect it to separate routingArguments and queryParameters internally which is the problem we are currently stuck in. Maybe "UriBuildingRequest" or "ActionUriSpecification"

Agree that "Action" is not a good name here. What it actually *does* is take the description of an Action and turn it into routing values (or parameters?).
So what about `ActionRoute`?

mficzel:
One thing i am not totally sure about is how this would affect uriBuilding in Fluid/Fusion especially when subrequests are involved as is the case inside of backend modules or widgets.

bwaidelich:
My [comment from yesterday](https://github.com/neos/flow-development-collection/pull/2744#issuecomment-1246377260) wasn't really well-thought-out and the code example didn't make sense.

Thinking about it again and sticking to the current API (for example having `$uriBuilder->withAddedRouteParameters(...)` I would suggest to keep `Action` since it actually describes what MVC action to invoke and add an additional `withAddedQueryParameters()` and/or `withAddedQueryParameter()` to the builder:

```php
$this->actionUriBuilder
  ->withAddedQueryParameter('foo', 'bar')
  ->uriFor(Action::fromActionRequest($this->request)->withArguments(['order' => true]));
```

> One thing i am not totally sure about is how this would affect uriBuilding in Fluid/Fusion especially when subrequests are involved as is the case inside of backend modules or widgets.

Shouldn't that be possible with

```php
$uriBuilderFactory->createFromActionRequest($subRequest)->uriFor(...);
```
?

bwaidelich:
..and one additional thought: "Builder" is a bit quirky. especially with the combination of a "builder factory".
Maybe it makes more sense to just call this `ActionUri`:

```php
$actionUri = $uriFactory
    ->createFromActionRequest($actionRequest)
    ->for(Action::name('show'))
    ->withArguments(['id' => $entityId])
    ->withAddedQueryParameter('foo', 'bar');

// (string)$actionUri === '/entity/<id>?foo=bar'
// $actionUri->value === UriInterface of '/entity/<id>?foo=bar'
// $actionUri->absolute()->value === UriInterface of 'https://domain.tld/entity/<id>?foo=bar'
```

mficzel:
Would prefer that
```
$actionUri = $uriFactory
    ->createFromActionRequest($actionRequest)
    ->for(Action::name('show'))
    ->withRoutingArguments(['id' => $entityId])
    ->withQueryArguments('foo', 'bar');
```
We should use the term parameter or argument consistently.

bwaidelich:
> We should use the term parameter or argument consistently.

it is always query**parameter**  and route**argument** (or at least it should be IMO)

I'm fine with an array, but it should be clear that the query parameters are _added_ to any existing parameters from the routing. That's why I named it "withAdded*"

mficzel:
> I'm fine with an array, but it should be clear that the query parameters are added to any existing parameters from the routing. That's why I named it "withAdded*"

I know and it. I only would prefer the nicer solution that pretends we already removed "appendExceedingArguments". After that removek withQueryParamerters will be correct.

bwaidelich:
I pushed an update introducing the `ActionUriSpecification` DTO as discussed.

To be discussed is whether this object should also have a notion of `queryParameters` and `fragment` (and potentially other parts of the Uri). I left it out for now.

## Example:

For the Neos preview route this would mean, changing
```php
$uriBuilder
      ->reset()
      ->setSection($section)
      ->setFormat($format ?: $request->getFormat())
      ->setCreateAbsoluteUri($absolute)
      ->uriFor('preview', ['node' => $node], 'Frontend\Node', 'Neos.Neos');
```
to something like:
```php
$actionSpecification = ActionUriSpecification::for(Action::create('Neos.Neos', 'Frontend\Node', 'preview'))
  ->withAdditionalArguments(['node' => $node])
  ->withFormat($format ?: $request->getFormat())
$uri = ($absolute ? $actionUriBuilder->uriFor($actionSpecification) : $actionUriBuilder->absoluteUriFor($actionSpecification))
  ->withFragment($section)
```

or (if we properly use *query parameters*):

```php
$actionSpecification = ActionUriSpecification::for(Action::create('Neos.Neos', 'Frontend\Node', 'preview'))
  ->withFormat($format ?: $request->getFormat())
$uri = ($absolute ? $actionUriBuilder->uriFor($actionSpecification) : $actionUriBuilder->absoluteUriFor($actionSpecification))
  ->withQuery(Query::build(['node' => $node->getContextPath()])
  ->withFragment($section)
```

mficzel:
@bwaidelich  To me this looks still too complicated and i would be fine with an ActionUriBuilder that 

1. Drops stuff like "addQueryString", "argumentsToBeExcludedFromQueryString"
2. Returns PSR Uri Objects
3. Has a simple interface that has a clear distinction between routingArguments and queryParameters 

```php
class ActionUriBuilder {
  public static function fromActionRequest() { ... }
  public static function fromHttpRequest() { ... }
  public static function fromBaseUri() { ... }

  public function uriFor(Action $action, array $routingArguments = null, array $queryParameters = null, string $section = null): UriInterface  { ... }
  public function absoluteUriFor(Action $action, array $routingArguments = null, array $queryParameters = null, string $section = null): UriInterface  { ... }
}
```

with `Action` beeing a value object encapsulating action, controller, package and subpackage

bwaidelich:
A wise man once wrote
> I like simple. Especially since I am quite bad at dealing with complexity

:)

I like the suggestion, with one suggestion:
Instead of the many arguments what about just

```php
class ActionUriBuilder {
  // ...
  public function uriFor(Action $action, array $routingArguments = []): UriInterface  { ... }
  public function absoluteUriFor(Action $action, array $routingArguments =[]): UriInterface  { ... }
}
```

IMO readability is quite similar:

```php
$action = new Action(package: 'Neos.Neos', controller: 'Frontend\Node', action: 'preview');

// v1
$uriBuilder->uriFor($action, section: 'some-section'));
$uriBuilder->uriFor($action, queryParameters: ['foo' => 'bar']);

// v2
$uriBuilder->uriFor($action)->withSection('some-secton');
$uriBuilder->uriFor($action)->withQuery(http_build_query(['foo' => 'bar']));
```

mficzel:
I think the extra `queryParmeters` argument helps to understand the distinction between those and `routingArguments`. We even could document this and explain the differences between them.

Also since routes may still use "appendExceedingArguments" it is not really safe to use `withQuery` since one would actually have to use:

```php
parse_str($uri->getQuery(), $queryParametersFromRouting);
$mergedQueryParameters = Arrays::arrayMergeRecursiveOverrule($queryParametersFromRouting, $queryParameters);
$uri = $uri->withQuery(http_build_query($mergedQueryParameters, '', '&'));
```
Compared to that i like the separate `queryParmeters`  much batter.

This leaves the `section` argument which i have not much opinions about. Would be fine to remove this and use `$uri->withFragment()` instead or call the parameter simply `fragment`

bwaidelich:
@mficzel good point(s)!

bwaidelich:
What about

```php
final class ActionUriBuilder {
  public static function fromActionRequest(): self;
  public static function fromHttpRequest(): self;
  public static function fromBaseUri(): self;

  public function uriFor(Action $action, array $routingArguments = [], array $queryParameters = []): UriInterface;
  public function absoluteUriFor(Action $action, array $routingArguments = [], array $queryParameters =[]): UriInterface;
}
```

mficzel:
@bwaidelich i really like this. 

One thing to check while implementing would be wether those `fromActionRequest` ... methods best belong to the `ActionUriBuilder` or the `Action`. Beeing on both ends would be a little weird but also an option.

```php
$builder = ActionUriBuilder::fromActionRequest($request) 
```
vs: 
```php
$action = Action::fromActionRequest(request)
```

bwaidelich:
> One thing to check while implementing would be wether those fromActionRequest ... methods best belong to the ActionUriBuilder or the Action

Ah yes, right.
I kinda like the `Action` VO to be a simple DTO like:
```php
final class Action {
    private function __construct(
        public readonly string $actionName,
        public readonly string $controllerName,
        public readonly string $packageKey,
    ) {}

    public static function create(string $packageKey, string $controllerName, string $actionName): self
    {
        return new self($actionName, $controllerName, $packageKey);
    }

    public static function fromActionRequest(ActionRequest $request): self
    {
        return new self($request->getControllerActionName(), $request->getControllerName(), $request->getControllerPackageKey());
    }

    // ...
}
```

Alternatively we could make `$controllerName` and `$packageKey` _optional_ and store them in the ActionUriBuilder (if created from ActionRequest) but IMO that increases complexity and chances of runtime exceptions.


BTW: We'll need the factory IMO because the `ActionUriBuilder` needs access to the router and base uri.
So instead of
```php
$builder = ActionUriBuilder::from...(...)
```
it would be
```php
$builder = $this->actionUriBuilderFactory->createFrom...(...);
```
(according to the current implementation)

mficzel:
Maybe ActionUriGenerator and ActionUriGeneratorFactory would be better names in that case. A BuilderFactory seems a little weird to me but it may still be the correct name.

mhsdesign:
Hi ;) i remember we discussed a lot in november ^^ how is the status and do we want it into 8.3 or is it too breaky and needs to wait for 9.0?

bwaidelich:
I don't think, it's too breaky since we could introduce it in _addition_ to the existing (marked deprecated) `UriBuilder`and it would even smooth migration to 9.0.
But unfortunately I don't really find the time to work on this one right now

Neos.Flow/Classes/Mvc/Controller/ControllerContext.php
```diff
@@ -24,7 +24,7 @@
  * for conveniently passing the information to other classes who need it, usually
  * views being views or view helpers.
  *
- * @api
+ * @deprecated with Flow 8.2 - The controller context is an internal concept that should be avoided in userland code
```
mhsdesign: finally ;)

at some time we will get rid of it :D

Neos.Flow/Classes/Mvc/Routing/ActionUriBuilder.php
```diff
@@ -0,0 +1,63 @@
...
+        private RouteParameters $routeParameters,
+    ) {
+    }
+
+    public static function fromRouterAndBaseUriAndRouteParameters(RouterInterface $router, UriInterface $baseUri, RouteParameters $routeParameters): self
+    {
+        return new self($router, $baseUri, $routeParameters);
+    }
+
+    public function withAddedRouteParameters(RouteParameters $routeParameters): self
```
mhsdesign, line 39: idk, in case we dont like this name i have an alternative:
```suggestion
    public function unionRouteParameters(RouteParameters $routeParameters): self
```

Neos.Flow/Classes/Mvc/Routing/ActionUriBuilder.php
```diff
@@ -0,0 +1,63 @@
...
+final class ActionUriBuilder
+{
+    private function __construct(
+        private RouterInterface $router,
+        private UriInterface $baseUri,
+        private RouteParameters $routeParameters,
+    ) {
+    }
+
+    public static function fromRouterAndBaseUriAndRouteParameters(RouterInterface $router, UriInterface $baseUri, RouteParameters $routeParameters): self
```
mhsdesign, line 34: i guess this factory shall stay highly specific named and not be `::create()` as its kind of internal (never actually to be used in userland)?

mhsdesign:
maybe @mficzel and me can finish this pr on the sprint?

Neos.Flow/Classes/Mvc/Routing/Dto/Action.php
```diff
@@ -0,0 +1,94 @@
...
+ * source code.
+ */
+
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Mvc\ActionRequest;
+
+/**
+ * @Flow\Proxy(false)
+ */
+final class Action
```
bwaidelich: As discussed: We should rename this (maybe to `ActionUriSpecification`?  and add a property for QueryParameters and rename addtionalArguments to `routingArguments`

mhsdesign: done ;) (thanks for the note ^^)

mhsdesign:
The current uri builder does also allow non rewritten uris via:

```php
        $uriPathPrefix = $this->environment->isRewriteEnabled() ? '' : 'index.php/';
        $uriPathPrefix = RequestInformationHelper::getScriptRequestPath($httpRequest) . $uriPathPrefix;
        $uriPathPrefix = ltrim($uriPathPrefix, '/');
```

should (must?) we support this aswell?

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -16,13 +16,8 @@ Neos:
...
       applicationToken: 'MinorVersion'
 
-      # Defines an explicit base URI that should be used. This affects
-      # resource management, routing and all other parts accessing
-      # "Http\Request->getBaseUri()".
-      #
-      # If this setting is NULL, the base URI will be detected from the current
-      # request (default).
-      baseUri: NULL
+      # Deprecated with Flow 8.3
```
kitsunet: How will current use cases of this be handled afterwards, I see no suitable replacement for eg. resource uris or command line uri building (especially outside of a neos context).

bwaidelich:
That ENV Variable was deprecated with 8.0 IIRC and I think it's totally fine to drop support for it.
But we should adjust the docs (and the htaccess file) and communicate that in the release notes IMO

Neos.Flow/Configuration/Settings.Http.yaml
mhsdesign: related: https://github.com/neos/flow-development-collection/issues/2157

for cli you can just do:

```php
$actionUriBuilder = $this->actionUriBuilderFactory->createFromBaseUri(new Uri("https://mybaseuri.de")); // or inject a setting etc
```

which is imo much better than what we can do with the current uribuilder / baseuri (the base uri is just often useless in a multisite, or when im using another host so it should be given as cli param)

cant say much about the other use cases ...

bwaidelich: IIRC the Resource Manager uses the current HTTP Request. For CLI usage instead of configuring a global base URI, this can (and IMO should) be configured at the corresponding ResourceTarget (or Collection?) instead.
But we should document and test this.
In general I think we should never add a `@deprecated` annotation without describing/linking to alternatives and/or explaining why this concept is considered bad

kitsunet: Right, Resources use both in a fallback fahion (`\Neos\Flow\Http\BaseUriProvider::getConfiguredBaseUriOrFallbackToCurrentRequest`) 

I guess directly configurating the baseUri on the target(s) is fine, might become a bit cumbersome if you have multiple, but with ENV that should do.

I am still unsure how the CLI usecase for complex stuff (eg. rendering fusion with links or mails and such) is to be handled?

mhsdesign: rendering node uris is fine in fusion as you can set the request... but interestingly its pretty hard (because of getConfiguredBaseUriOrFallbackToCurrentRequest)  to render links of assets ... im doing this currently:


```php
     /**
     * @var BaseUriProvider
     * @Flow\Inject(lazy=false)
     */
    protected $baseUriProvider;
  
   public function myCommand(string $uri)
   {
         $restoreBaseUriProviderSingleton = $this->hackTheConfiguredBaseUriOfTheBaseUriProviderSingleton(new Uri($uri);

                 // somewhere in my code / fusion / convertUris
                $this->linkingService->resolveAssetUri("asset://3242234234234234")

       $restoreBaseUriProviderSingleton();

   }


    /**
     * @return callable restore the original state
     */
    private function hackTheConfiguredBaseUriOfTheBaseUriProviderSingleton(UriInterface $baseUri): callable
    {
        assert($this->baseUriProvider instanceof BaseUriProvider);

        static $originalConfiguredBaseUri;
        if (!isset($originalConfiguredBaseUri)) {
            $originalConfiguredBaseUri = ObjectAccess::getProperty($this->baseUriProvider, "configuredBaseUri", true);
        }

        ObjectAccess::setProperty($this->baseUriProvider, "configuredBaseUri", (string)$baseUri, true);

        return function () use($originalConfiguredBaseUri) {
            ObjectAccess::setProperty($this->baseUriProvider, "configuredBaseUri", $originalConfiguredBaseUri, true);
        };
    }
```

mhsdesign: `resolveAssetUri` could take in the baseUri or current request …

The linking service is generally due for a refactor after this pr 😄 … we can plan any breaking changes for 9.0 then …

mhsdesign: But if we require a second argument something like the `Neos.Link` helper becomes less sexy:

```
${Neos.Link.resolveAssetUri(uri)}
# new
${Neos.Link.resolveAssetUri(uri, request)}
```

on the other hand, the `Neos.Link.resolveNodeUri` also requires already the controllerContext ... so ...

mhsdesign:
And do we maybe want some validation while constructing the `ActionUriSpecification `? I can imagine checking if the controller exists?

Or we can make a factory by controller class name like:

```php

// option1
ActionUriSpecification::createForController("Foo\Bar\Controller\MyController::indexAction");
// option2
ActionUriSpecification::createForController(Foo\Bar\Controller\MyController::class, "indexAction");
// option3 (phpstorm ide support as this is infact a callable)
ActionUriSpecification::createForController([Foo\Bar\Controller\MyController::class, "indexAction"]);
# option 4
ActionUriSpecification::forController(Controller::class)->withActionName(‘index’)
```

what do you say? @bwaidelich

bwaidelich:
@mhsdesign I would definitely avoid some actual validation as that would introduce a dependency to the `ObjectManager` and it might break cases where the controller doesn't actually exists or a different naming scheme is used.
I think, "option 4" is a neat idea – but it also means that we duplicate the naming resolution logic from `\Neos\Flow\Mvc\ActionRequest::getControllerObjectName()` – I'd be OK with it, but we could also embrace the uncertainty of connecting two systems in this point

Neos.Flow/Configuration/Settings.Http.yaml
mhsdesign: well i discussed this dilemma with @sorenmalling and i get know why we can pass the baseUri or request that simple to `Neos.Link.resolveAssetUri` as there ResourceTargetInterface would need to be changed too and for non file system targets, its completely irrelevant what the base uri might be (cloud storage) so it might be odd to pass the base uri along.

I guess we are still stuck with some global singleton baseUri configuration which can be written to...

idea (pseudo code):


in the cli handler
```php
// set the baseuri via env or cli argument
$baseUri = getEnv("FLOW_BASE_URI") || $cliArguments["baseUri"];
$this->baseUriConfiguration->withBaseUri(new Uri($baseUri), fn () => $this->processCliRequest());
```

in the web handler
```php
// base uri inferred from request
$this->baseUriConfiguration->withBaseUri(
    RequestInformationHelper::baseUriFromRequest($httpRequest),
    fn () => $this->processRequest($httpRequest)
);
```


in my custom code, when i want to temporary change the base uri:
```php

$output = [];
foreach($siteUri as $uri) {
   $output[] = $this->baseUriConfiguration->withBaseUri($uri, fn () => $this->fusionRenderNode($node));
}

```

mhsdesign:
my end goal would be to allow a fully qualified class name in the controller option of the routes.yaml (as discussed with @sorenmalling)

Neos.Flow/Configuration/Settings.Http.yaml
bwaidelich: Shouldn't `${Neos.Link.resolveAssetUri(uri)}` just work as is?
It should hopefully use the configured `baseUri` of the resource target if it is set (do we already have an option for this in our implementations? otherwise I'd suggest to add this, since it's a fully backwards-compatible change).
If no base uri is configured for the target the default `\Neos\Flow\Http\BaseUriProvider::getConfiguredBaseUriOrFallbackToCurrentRequest()` is used IIRC and that should still work even if the `baseUri` configuration is deprecated (it will fall back to the current HTTP request).

So everything would work just like before only that configuring a base URI on the resource targets would be preferred to the (now deprecated but still supported) global base URI configuration

bwaidelich:
I would suggest to strive for some kind of "minimal viable PR" in order to get this into 8.3 still.
We can always extend the behavior afterwards, but IMO we should choose our fights carefully (i.e. avoid bike-shedding)

Neos.Flow/Configuration/Settings.Http.yaml
mhsdesign: It compose the baseuri of the the target options with the current host … and an absolute uri (with host) is required for emails fx…

bwaidelich:
@mhsdesign will this still make it into the upcoming release?

mhsdesign:
~yes i very much like so ;)~ yes

(i agree to delay the discussion of my above thoughts - now that they have been heard i can make a separate pr for this)

but i still feel the need to discuss https://github.com/neos/flow-development-collection/pull/2744#discussion_r1133546795
as we havent come to an conclusion yet. Are you @bwaidelich or @kitsunet or @mficzel  up for small discussion tomorrow?

Neos.Flow/Configuration/Settings.Http.yaml
bwaidelich: Just briefly discussed with @mhsdesign and after thinking about it a couple of minutes longer I think that we don't have to implement any further `baseUri` replacement _for now_.

In the `@deprecated` annotation and the upgrade guides we should explain how you are supposed to work around the globally configured base URI.

Some further thoughts and ideas:

### MVC URIs

To create absolute URIs for CLI or other "out of band rendering", the `ActionUriBuilderFactory::createFromBaseUri()` can be used.

But we have to think about solutions for Fusion (and maybe even Fluid?).
E.g. the `Neos.Fusion:ActionUri` Fusion prototype currently uses the action request to resolve default package, controller, ... and base URI.

A separate `baseUri` prop could work:

```
uri = Neos.Fusion:ActionUri {
  baseUri = ${Configuration.setting('Some.Package:someBaseUri')}
  package = 'Some.Package'
  controller = 'Standard'
  action = 'index'
  absolute = true
 }
```

But maybe it would be better to instantiate the runtime with some parameter.
Currently `\Neos\Fusion\Core\RuntimeFactory::createControllerContextFromEnvironment()` creates a HTTP request via `ServerRequest::fromGlobals()`. That means that absolute URLs will be generated with "http://localhost/...".

### Resources

resources will already work out of the box for regular HTTP requests even without configured baseUri because it will detect the base URI from the current server request.

For CLI requests one can use the already existing `baseUri` option of the resource targets to make that work out of the box:

So by replacing "_Resources/..." with "https://domain.tld/_Resources/..." in:

```yaml
Neos:
  Flow:
    resource:
      targets:
        localWebDirectoryStaticResourcesTarget:
          targetOptions:
            baseUri: '_Resources/Static/Packages/'

        localWebDirectoryPersistentResourcesTarget:
          targetOptions:
            baseUri: '_Resources/Persistent/'
```

For _Multi Site Installations_ that approach will work as well, but it has the (probably unexpected) side effect of resources being loaded from a different domain in some cases.

To work around that, we _could_ introduce some global concept (as Marc Henry suggested [above](https://github.com/neos/flow-development-collection/pull/2744#discussion_r1138872190)) but I would prefer not to go down that path (since it is basically the same as setting some superglobal variable, which is considered bad practice for obvious reasons).

An alternative could be some "context" parameter for the `ResourceManager::getPublicPersistentResourceUri()` method, i.e.:
```php
$context = ResourceTargetContext::fromRequest($httpRequest);
$uri = $this->resourceManager->getPublicPersistentResourceUri($resource, $context);
```

or – to prevent breaking changes – with some new `ContextAwareInterface` (or similar):
```php
$uri = $this->resourceManager
    ->withContext($context)
    ->getPublicPersistentResourceUri($resource);
```

mhsdesign: Thanks for the write down.

I agree that we don’t have to change anything further in this pr.

But after discussing the matter with @sorenmalling and @fcool we all came to the conclusion that the global concept still makes sense and is our best / safe bet.

i will prepare a draft pr for this were we can discuss further

mhsdesign:
I think conceptual this is ready, let’s have a look at the failing test 🙃

Neos.Flow/Configuration/Settings.Http.yaml
bwaidelich: > But after discussing the matter with @sorenmalling and @fcool we all came to the conclusion that the global concept still makes sense and is our best / safe bet.

I respectfully disagree and IMO this "global thinking" introduced a lot of issues. At least the "global global" approach as in Settings.

fcool:
Nobody really doubts that, I guess.

But as already discovered with Marc: The places, where we would need to decide on the BaseUri are places where currently is no request at hand. 
If this should fly, we would need to overwork the ActionController framework and all parts to make the request available - most of the time you would decide probably on something coming from Request or maybe Session...

And even harder it will get in CLI-Context. So you need to handle it "outside" of the place, where you generate the URL anyway. By a request attribute, by some global CLI configuration, etc.

So having it global (a lot like a our locale configuration) is probably the easiest (and at some places the only useful) way of controlling it, I think.

Yes, it has been a pain (and indeed we used a `BaseUriProvider`-Like setup we introduced using some Aspects) from Flow 3.x ongoing, migrated our stuff to the BaseUri-Provider (which made of course new trouble ;) )... 

But having all places, which need to generate URLs "BaseUriAware" is a huge task!

Neos.Flow/Configuration/Settings.Http.yaml
fcool: Nobody really doubts that, I guess.

But as already discovered with Marc: The places, where we would need to decide on the BaseUri are places where currently is no request at hand.
If this should fly, we would need to overwork the ActionController framework and all parts to make the request available - most of the time you would decide probably on something coming from Request or maybe Session...

And even harder it will get in CLI-Context. So you need to handle it "outside" of the place, where you generate the URL anyway. By a request attribute, by some global CLI configuration, etc.

So having it global (a lot like a our locale configuration) is probably the easiest (and at some places the only useful) way of controlling it, I think.

Yes, it has been a pain (and indeed we used a BaseUriProvider-Like setup we introduced using some Aspects) from Flow 3.x ongoing, migrated our stuff to the BaseUri-Provider (which made of course new trouble ;) )...

But having all places, which need to generate URLs "BaseUriAware" is a huge task!

Apart from that: for the majority of installations out there a simple configuration is probably sufficient. It's always the few really nice projects, that need a less rigid corset ;)

bwaidelich: I think it boils down to:

* do we agree, that we want to get rid of the global `baseUri` config in the long term? (if no, we should not deprecate it)
* if so, do we want to investigate towards a "parameter"-approach (i.e. replacing `ControllerContext` with something that is more reusable for CLI etc. and can hold the base URI)?
* or something like `Security\Context::withoutAuthorizationChecks()` for the `BaseUriProvider`, i.e.:

```php
functon renderWithBaseUri(UriInterface $baseUri) {
    BaseUriProvider::withBaseUri($baseUri, $this->render(...));
}
```

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -38,16 +40,28 @@
...
     /**
+     * @deprecated with Flow 8.3 - use $this->actionUriBuilder instead
      * @var UriBuilder
      */
     protected $uriBuilder;
 
+    /**
+     * @var ActionUriBuilder|null
+     */
+    protected ?ActionUriBuilder $actionUriBuilder;
```
mhsdesign, line 51: should probably not nullable so people dont have to check, and psalm or phpstan cant complain

Neos.Flow/Configuration/Settings.Http.yaml
mhsdesign: Hi, i just discussed this also with @mficzel and we found a minimal invasive solution: we will just introduce the enviroment setting `FLOW_BASE_URI`

so `BaseUriProvider::getConfiguredBaseUriOrFallbackToCurrentRequest` will just check for
1. `FLOW_BASE_URI`
2. deprecated setting `"Neos.Flow.http.baseUri"` 
3. base uri of $this->bootstrap->getActiveRequestHandler()


(the env can technically even be set in edge cases dynamically to cover my use case https://github.com/neos/flow-development-collection/pull/2744#discussion_r1138872190)

That way the BaseUriProvider could mostly just stay the same and we dont introduce any getter or setter (what we might regret in the future)

Also we have a nice replacement for the setting `"Neos.Flow.http.baseUri"` which will still be deprecated.

mhsdesign: Created as promised this follow up pr: https://github.com/neos/flow-development-collection/pull/3002

i hope youre all happy with this solution ...

Neos.Flow/Classes/Mvc/Routing/Dto/ActionUriSpecification.php
```diff
@@ -0,0 +1,169 @@
...
+            $request->getFormat(),
+            [],
+            []
+        );
+    }
+
+    /**
+     * @deprecated with Flow 8.3 - The use of subpackage keys is discouraged and only supported for backwards compatibility
+     */
+    public function withSubpackageKey(string $subpackageKey): self
```
mhsdesign, line 58: @bwaidelich is it intentional that we dont have `ActionUriSpecification::withPackageKey`

Neos.Flow/Configuration/Settings.Http.yaml
kitsunet: But how is this different than the setting, apart from making it harder for some folks to use the feature?

bwaidelich: I agree to @kitsunet .. let's discuss this in #3002

Neos.Flow/Classes/Mvc/Routing/Dto/ActionUriSpecification.php
bwaidelich: Not really IIRC. Do you think we need it?

bwaidelich: i.e. the `fromActionRequest` constructor is meant to easily create URLs to actions of the same package.
To link to _other_ packages you can always use `create()` because the current action request is irrelevant for that. I guess that was my thinking behind omitting it

mhsdesign: i thought rewriting redirect like:
```php
$targetActionUriSpecification = ActionUriSpecification::fromActionRequest(
    $this->request,
)
    ->withActionName($actionName)
    ->withRoutingArguments($arguments);


if ($packageKey !== null && str_contains($packageKey, '\\')) {
    [$packageKey, $subpackageKey] = explode('\\', $packageKey, 2);
} else {
    $subpackageKey = null;
}

if ($packageKey !== null) {
    $targetActionUriSpecification = $targetActionUriSpecification->withPackageKey($subpackageKey);
}
if ($subpackageKey !== null) {
    $targetActionUriSpecification = $targetActionUriSpecification->withSubpackageKey($subpackageKey);
}
```

kitsunet: you might as well do these decisions beforehand and use create?

kitsunet: I would encourage (by example) to avoid using the current request to build uris as much as possible.

mhsdesign: This is the way

Neos.Flow/Tests/Unit/Mvc/Routing/Dto/ActionUriSpecificationTest.php
```diff
@@ -0,0 +1,85 @@
...
+            ->withRoutingArguments(["neos" => 42])
+            ->withQueryParameters(["q" => "lol"]) // will not be encoded
+            ->withSubpackageKey("Application")
+            ->withFormat("json");
+
+        self::assertEquals(
+            [
+                'neos' => 42,
+                '@action' => 'index',
+                '@controller' => 'neos\flow\mvc\controller\actioncontroller',
```
mhsdesign, line 77: todo this is actually a conceptually wrong test value (i introduced it ^^)

Neos.Flow/Classes/Security/Authentication/EntryPoint/WebRedirect.php
```diff
@@ -78,18 +71,19 @@ public function startAuthentication(ServerRequestInterface $request, ResponseInt
...
      */
     protected function generateUriFromRouteValues(array $routeValues, ServerRequestInterface $request): string
     {
-        $actionRequest = ActionRequest::fromHttpRequest($request);
-        $this->uriBuilder->setRequest($actionRequest);
-
         $actionName = $this->extractRouteValue($routeValues, '@action');
         $controllerName = $this->extractRouteValue($routeValues, '@controller');
         $packageKey = $this->extractRouteValue($routeValues, '@package');
-        $subPackageKey = $this->extractRouteValue($routeValues, '@subpackage');
```
mhsdesign, line 91: do i see correct that we remove the `@subpackage` here silently in a minor?

Neos.Flow/Classes/Mvc/Routing/Dto/ActionUriSpecification.php
```diff
@@ -0,0 +1,169 @@
...
+ * @Flow\Proxy(false)
+ */
+final class ActionUriSpecification
+{
+    private function __construct(
+        private string $actionName,
+        private string $controllerName,
+        private string $packageKey,
+        /** @deprecated with Flow 8.3 - The use of subpackage keys is discouraged and only supported for backwards compatibility */
+        private ?string $subpackageKey,
```
mhsdesign, line 30: @fcool and me wondered what the alternative of subpackageKey would look like ...

We would be fine if we could do something like this:
https://github.com/neos/flow-development-collection/pull/2744#issuecomment-1469632526

but just deprecating without alternative is not fair imo.

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -253,33 +268,66 @@ protected function forwardToRequest(ActionRequest $request)
...
         } else {
             $subpackageKey = null;
         }
-        $this->uriBuilder->reset();
-        if ($format === null) {
-            $this->uriBuilder->setFormat($this->request->getFormat());
-        } else {
-            $this->uriBuilder->setFormat($format);
+
+        $targetActionUriSpecification = ActionUriSpecification::create(
```
mhsdesign, line 315: @fcool  and me discussed if it would make sense for the new ActionUriBuilder to support our legacy-but-still-used concept of namespaced parameters for subrequest.

If this is wanted we could introduce a `NameSpacedActionUriSpecification` which would take care of handling those specifics and serialize them correctly in `toRouteValues`

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
mhsdesign: ~yes but our ugly test would fail, as initializeController will be called with actionUriBuilderFactory as a non functioning mock currently ... so we assing null at some point~

edit: i was wrong it should work ...

Neos.Flow/Classes/Mvc/Routing/Dto/ActionUriSpecification.php
```diff
@@ -0,0 +1,169 @@
...
+        private array $queryParameters,
+    ) {
+    }
+
+    public static function create(string $packageKey, string $controllerName, string $actionName): self
+    {
+        return new self($actionName, $controllerName, $packageKey, null, '', [], []);
+    }
+
+    public static function fromActionRequest(ActionRequest $request): self
```
mhsdesign, line 42: we should probably annotate that we hate this method ^^

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
bwaidelich: It has to be nullable because it's not set at construction time unfortunately

mhsdesign: hmm this works?
https://onlinephp.io?s=dY-xDsIwDERnIuUfrCoDDF2ArdBu_AeEtIlUxVXiiAH13wlNqkZIeLw7v7Mv3aQnzjiT4917uCFy9uYM4kwOSUlST_DkjB1A9IhN8nIiPEYjoQ9WkkELXlEE7Iv4IQUzcSdIG1-30YBriZtL6n_-kPg_UKcoOAsbe4PO3884E6nQqhcs-80q1m0-uTqeT1XSldQIyVz7Fv0D&v=8.2.4

the only rule is, that it can only be initialized once and might not be accessed (except isset and ??)

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
mhsdesign: but i rather not just duplicate all the Uribuilder::mergeArgumentsWithRequestArguments into another place like: NameSpacedActionUriSpecification

we wouldnt gain anything.

what could be cool, if we can swap out all the internals of the uribuilder to use the ActionUriSpecification or NameSpacedActionUriSpecification so we internally use the new api. But i would wait for this refactor untill 9.0 as it could be breaky in edgecases

since our rewrite method in the abstract controller didnt really got better, i would be fine with continuing to use the old uribuilder here for now and wait for 9.0 to introduce the  NameSpacedActionUriSpecification.

mhsdesign:
After looking into the namespace dilemma yesterday with @fcool (thanks for your time it was really fun ^^) we found out that there are still some things left to discuss and we should think about how a roadmap of migrating to the new actionUribuilder could look like. (currently not everything is supported what the old one could do)

id very much like to sync with @bwaidelich our findings from yesterday and hope we can still make it into 8.3
(the safest bet would be: dont rewrite any internals but only offer the new actionUribuilder as new api)

mhsdesign:
@bwaidelich and me are thinking of delaying this for 9.0

#entwederganzodergarnicht

bwaidelich:
@mhsdesign @fcool @mficzel and everybody else who was involved: Thanks a lot for your efforts and the discussions.
I think we're almost there, but deprecating the `UriBuilder` while there is no 100% replacement (which means that we still have to use it ourselves in the core) does not feel right and pose the risk of a premature abstraction.

IMO the following issues should be addressed still:

- [ ] move `mergeQueryParametersIntoUri` and `toRouteValues` from the `ActionUriSpecification` to the `ActionUriBuilder` to turn it into a self-contained DTO (props can be `public readonly` with PHP 8.1)
- [ ] Maybe find a less convoluted name for `ActionUriSpecification`?
- [ ] Introduce some (static?) helper that implements our Sub- and PluginRequest merge magic that is currently in spread out to the `UriBuilder` and other parts (IMO it should _not_ be part of core MVC and maybe even deprecated – but good alternatives have to be found and documented)
- [ ] Make the `UriBuilder` use the `ActionUriBuilder` internally (possibly with the help of this new merge helper)

kitsunet:
Hey @bwaidelich I assume you are fine if I take this on and see to add the missing functionality?

bwaidelich:
> I assume you are fine if I take this on and see to add the missing functionality?

@kitsunet more than fine! And feel free to ping me for discussions!

kitsunet:
Some things already:

Alright so we remove the option to run without rewrites, fine by me I guess, but we might rephrase the PR a bit wider then so people understand the implications of this. 
What also seems to have gone is the ScriptRequestPath, while I have never run a flow/neos without rewrite, I am pretty sure I had to run it in a subdirectory (or reverse proxied into one), do we want to rely on those people needing it to handle it via  baseUri? Could be fine, I'll still have a look at those usecases and if it's feasible without globals somehow.

I would probably remove `ActionUriSpecification::queryParameters` as IMHO that is not integral nor necessary for the ActionUri, it's extra, and while convenient in that DTO not really fitting. So I will have a look at how separating it could look like. 

We _could_ interpret every sub/plugin request as ActionUriSpecification (because in essence that's what it is) that can be added to a (parent) specification together with the namespace it needs to appear in, that seems sensible and would allow a more programmatic approach to constructing a subrequest without introducing too much other stuff. WDYT?

The current code in here seems to now require a baseUri in the Specification, which is due to the RouteContext requiring it I assume, bu then I wonder why (I assume because the constraints can later on force an absoluteUri and then we need it), I guess fine by me, it feels weird requiring it, if all I want is a route path but I am also not seeing a nice solution otherwise that isn't muddy and requires exceptions. tl;dr; probably fine even if a bit weird API maybe, if you don't know the whys.

More to come

bwaidelich:
@kitsunet Thanks for your elaborate comment!

> Alright so we remove the option to run without rewrites

What do you mean? That we went for a new implementation instead of "fixing" the existing `UriBuilder`?

> What also seems to have gone is the ScriptRequestPath

Yes, that's true.. I can't really judge whether people rely on that widely but we should definitely consider it

> I would probably remove `ActionUriSpecification::queryParameters` as IMHO that is not integral nor necessary for the ActionUri

If I get you right, this is something we went back and forth with. I also thought that it makes sense to separate query parameters from the routing but @mficzel was in favor of solving it in the builder. And he has a point, setting and merging hierarchical query parameters is a pain with the `UriInterface` and we would need to add yet another utility class.

Or was your point just to remove it from the specification and leave it in the builder?

> We could interpret every sub/plugin request as ActionUriSpecification (because in essence that's what it is)

I have my concerns tbh. There is no such thing as a "sub request" and pretending there _was_, led to many issues. I hope that the whole concept goes away in the long run and I would suggest to keep the "work arounds" away from the core API, i.e. not adding something like `ActionUriSpecification::parentActionUriSpecification` because we might end up with a mess again – a shinier one, but still a mess :)

> The current code in here seems to now require a baseUri in the Specification

The `ActionUriSpecifcation` should not need it, but the builder needs it because: `$builder->uriFor(ActionUriSpecification::create(...))`  can return a different result depending on which host you're on currently (i.e. in Neos it will create an absolute URI to the target site if that has a different domain etc).

We should definitely document that – the current code almost contains no doc comments yet


One more thought:
IMO with the introduction of named parameters we don't have to focus _so much_ on methods with lesser arguments if the alternative is easier to read.

kitsunet:
> Alright so we remove the option to run without rewrites

Sorry that was imprecise, I meant FLOW_REWRITEURL basically, saying we can't run without path rewrites enabled. which as said, I think i sperfectly fine, I don#t know a single instance in which I had needed that.

And thank you for the quick replies <3

I might still make a suggestion then for splitting up the query params and the rest but know that was already considered.

kdambekalns:
> can return a different result depending on which host you're on currently (i.e. in Neos it will create an absolute URI to the target site if that has a different domain etc).

"Will" in the sense of "with this PR merged, in the future"?

Because right now it seems impossible… I needed to build an URI to a node recently, and even though I built a `ContentContext` with the correct `Site` and passed a node with the proper context and path, the `UriBuilder` always gave me an absolute URI pointing to the "currently requested host", not the host the `Site` was using… 😵‍💫

In the end I asked for a relative URI and glued it to the host data from the `Site` object. 🙈

bwaidelich:
@kdambekalns that _should_ work today as of Neos 7.0 IIRC. That logic was moved from the `LinkingService` to the route part handler with https://github.com/neos/neos-development-collection/commit/45207d79b388494bf619a45eb4f054b7954ea8f5

kdambekalns:
> @kdambekalns that _should_ work today as of Neos 7.0 IIRC. That logic was moved from the `LinkingService` to the route part handler with [neos/neos-development-collection@45207d7](https://github.com/neos/neos-development-collection/commit/45207d79b388494bf619a45eb4f054b7954ea8f5)

Well… maybe I just did somethign wrong. Then again, the commit message says

> The "current Site" is only determined once, not for every URI

and maybe exactly that is the problem? If "once" is per request, that is the requested site, but I need it to be the target site… Here is what I did: https://gist.github.com/kdambekalns/8d1628a355e6cf9ccdfd11fcfae23ddf

In the end the URI comes from https://github.com/neos/flow-development-collection/blob/7.3/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php#L367C75-L367C75 and all is lost, because the fallback is not used, since https://github.com/neos/flow-development-collection/blob/7.3/Neos.Flow/Classes/Http/BaseUriProvider.php#L85-L88 succeeds…

kdambekalns:
> > @kdambekalns that _should_ work today as of Neos 7.0 IIRC. That logic was moved from the `LinkingService` to the route part handler with [neos/neos-development-collection@45207d7](https://github.com/neos/neos-development-collection/commit/45207d79b388494bf619a45eb4f054b7954ea8f5)
> 
> Well… maybe I just did somethign wrong. Then again, the commit message says
> 
> > The "current Site" is only determined once, not for every URI
> 
> and maybe exactly that is the problem? If "once" is per request, that is the requested site, but I need it to be the target site… Here is what I did: https://gist.github.com/kdambekalns/8d1628a355e6cf9ccdfd11fcfae23ddf

In the end the URI comes from https://github.com/neos/flow-development-collection/blob/7.3/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php#L367C75-L367C75 and all is lost, because the fallback is not used, since https://github.com/neos/flow-development-collection/blob/7.3/Neos.Flow/Classes/Http/BaseUriProvider.php#L85-L88 succeeds…

bwaidelich:
> and maybe exactly that is the problem? If "once" is per request, that is the requested site,
> but I need it to be the target site

A middleware puts the current request host to the `requestUriHost` routing parameter (The routing should really have access to the host without this, but this is required for historic reasons).
Once that routing parameter is set to the current host (aka "source"), any URI for a node from the _target_ site should be absolute with the host of the corresponding domain).

If you're already on the target site (i.e. the `requestUriHost` matches the domain of the current site), a _relative_ URI should be generated

To make sure you can set the routing parameter explicitly to make sure, as stated in the commit message:

> The `FrontendNodeRoutePartHandler` expects the `requestUriHost` routing
  parameter to be set (usually done via HTTP Middleware). You can simulate that behavior via:
  `$httpRequest = $httpRequest->withAttribute(ServerRequestAttributes::ROUTING_PARAMETERS, RouteParameters::createEmpty()->withParameter('requestUriHost', 'some-host'));`

kdambekalns:
(Sorry for hijacking this PR!)

So this (from my gist) should be working?! But it doesn't?!

```php
        $baseUri = sprintf('%s://%s',
            $acmeSite->getFirstActiveDomain()->getScheme(),
            $acmeSite->getFirstActiveDomain()->getHostname()
        );
        $httpRequest = new ServerRequest('GET', $baseUri);
        $httpRequest = $httpRequest->withAttribute(
            ServerRequestAttributes::ROUTING_PARAMETERS,
            RouteParameters::createEmpty()->withParameter('requestUriHost', $httpRequest->getUri()->getHost())
        );
```

I guess to make it work, I'd need to replace the URI in the "actual HTTP request", to trick the `BaseUriProvider` into using it. But that would be so scary…

mhsdesign:
With the Neos part progressing nicely https://github.com/neos/neos-development-collection/pull/4892 we can apply what we learned and build the api more similar to how we want it for Neos, currently that would be:

```php
#[Flow\Inject]
protected ActionUriBuilderFactory $actionUriBuilderFactory;

$actionUriBuilder = $this->actionUriBuilderFactory->forRequest($actionRequest->getHttpRequest());

$actionUriBuilder->uriFor(
    Action::create(
        packageKey: 'Neos.Flow',
        controllerName: 'Apple',
        actionName: 'index',
        subpackageKey: 'NudelSuppe'
    ),
    Options::create(
        forceAbsolute: true,
        format: 'html',
        routingArguments: []
    )
);

$uri = UriHelper::uriWithQueryParameters($uri, ['query' => 'any-thing']);
```

A major todo would be still to know how to work with flows subrequests ...

bwaidelich:
Just wondering: Is this one still cooking?

mhsdesign:
jup but we have _real_ problems first :) Maybe 9.1 9.2 ... or 10.0

bwaidelich:
> jup but we have real problems first

I know, but only a couple of devs are working on these. No reason to stop the world on all other fronts, is it? :)
Anyways, I just wanted to know if someone still has that topic on their plate because it pops up in my "pending PRs"

PR #2749, opened 2022-03-18
BUGFIX: Adjust type annotations in browser to match the relaxing from ServerRequestInterface to RequestInterface

mficzel:
Cleanup an error introduced in https://github.com/neos/flow-development-collection/pull/2688


mficzel:
I will merge this without further review as follow up of https://github.com/neos/flow-development-collection/pull/2688

Issue #2717, opened 2022-03-02
!!! FEATURE: Introduce flushByTags to taggable cache backends


Sebobo:
### Description

Currently only `flushByTag` is supported. This is highly inefficient for large amount of tags to be flushed.
By introducing `flushByTags` each cache backend can have an optimised implementation on how to flush them with the least amount of calls to the underlying storage (Redis, PDO, files, etc.).

This change does not optimise anything yet and only internally calls `flushByTag` in a loop, as any improvements will be done in separate PRs for the individual cache backends.

PR #2750, opened 2022-03-18
FEATURE: `flushByTags` for the file based cache backend

Sebobo:
**What I did**

With this change the `flushByTags` is optimised by only reading the list of cache files once for all entries to be flushed instead of reading it for each entry.

This can yield a huge speedup in situations where a single file based cache contains many entries, especially on older (non-ssd) storages.

This change relies on the interface changes in #2717


mficzel (changes requested):
I generally like the idea but would avoid the seperate implementation of the plural  methods 
why not 
```
public function flushByTag(string $tag): int
{
   return $this->flushByTags([$tag]);
}
```
as is already done in findIdentifiersByTag.

Sebobo:
@mficzel true, I actually wanted to do that and then forgot. Will change it :)

Sebobo:
@mficzel adjusted the code and test.

mficzel (approved):
Fine by reading.

PR #2753, opened 2022-03-21
TASK: use Psalm template tag instead of PHPSTORM_META

mhsdesign:
\Neos\Flow\ObjectManagement\ObjectManagerInterface::get
\Neos\Flow\Core\Bootstrap::getEarlyInstance

make use of the `@template` tag:
 https://psalm.dev/docs/annotating_code/templated_annotations/

the current `.phpstorm.meta.php` is outdated too https://www.jetbrains.com/help/phpstorm/ide-advanced-metadata.html#legacy-metadata-format-deprecated


albe:
Test failures probably due to #2756, so this needs a merge

kdambekalns (changes requested):
Please, no!

`@return T` – no, this does not return `T`!

Why not just update `.phpstorm.meta.php` to the current syntax?

kdambekalns:
See https://github.com/neos/flow-development-collection/pull/2846

mhsdesign:
> Please, no!
> 
> `@return T` – no, this does not return `T`!
> 
> Why not just update `.phpstorm.meta.php` to the current syntax?

@kdambekalns I think the typescript like genetic way is better isn't it? Can you please elaborate what I got wrong with this approach?

mhsdesign:
got it, thanks! @kdambekalns

albe:
If we'd only want to help psalm understand things better, we could use `@psalm-param` and `@psalm-return` (https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-var-psalm-param-psalm-return-psalm-property-psalm-property-read-psalm-property-write-psalm-method), but the main use case for the meta file is having autocomplete support in the IDE.

> And again, read verbatim it's not true, the method actually expects a string (probably even by type declaration). Again, if that worked on a language level, yay!

It wouldn't even be true if there were a `classname` type, because the object manager also accepts "virtual" object names (https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/ObjectManagement.html#virtual-objects), which have no language level meaning anyway. So it's definitely a `string`.
Maybe it turns out that wasn't a good idea to add this feature? :joy:

mhsdesign:
Okay i will close this - i guess php is not typescript xD

bwaidelich:
For the record: Psalm, PHPStan and in the meantime even [PHPStorm](https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-release/#generics) support this generic type hints

kdambekalns:
It still doesn't "return T", but if it has arrived on a (meta-)language level by now, do as you like. 😇

PR #2755, opened 2022-03-21
TASK: Debug build failures

albe:
Just an attempt to find the cause of latest failing builds, which seems to be rooted in one of the dependencies.

- [ ] composer/composer 2.2.9
- [ ] composer/semver 3.3.1
- [x] guzzlehttp/psr7 1.8.5
- [ ] phpunit/phpunit 8.5.25

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "neos/composer-plugin": "^2.0",
         "composer/composer": "^1.10.22 || ^2.0.13",
         "typo3fluid/fluid": "~2.5.11 || ~2.6.10",
-        "guzzlehttp/psr7": "^1.4.1, !=1.8.0",
+        "guzzlehttp/psr7": "1.8.3",
         "ext-mbstring": "*"
     },
     "replace": {
--- a/composer.json
+++ b/composer.json
@@ -116,7 +116,7 @@
     },
     "require-dev": {
         "mikey179/vfsstream": "^1.6.1",
-        "phpunit/phpunit": "~8.5",
+        "phpunit/phpunit": "8.5.24",
         "vimeo/psalm": "~4.22.0"
     },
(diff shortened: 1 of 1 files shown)


mhsdesign:
nice idea ^^ i also came to the dependency conclusion

this lately working, now broken rerun: https://github.com/neos/flow-development-collection/actions/runs/1988665680

only is green for functional tests for `PHP 7.2 Test (deps: lowest)` run, as it seems because of the `--prefer-lowest` flag

albe:
Apparently guzzlehttp/psr7 is the culprit (once again :D)
The functional change from 1.8.3 is
https://github.com/guzzle/psr7/pull/491/files#diff-fb174524a7bba27ce140bc6ccd1c30811a6abeed9328e783b326189551ba7ed4R266 and https://github.com/guzzle/psr7/pull/486/files#diff-fb174524a7bba27ce140bc6ccd1c30811a6abeed9328e783b326189551ba7ed4R192 ff.

So it seems something with our headers we set in tests/internal request engine is incompatible with the new validation

albe:
Well, if we'd look closely, the failing tests already mention the issue:
```
"Access denied for method
Method: Neos\Flow\Tests\Functional\Security\Fixtures\Controller\RestrictedController::adminAction()

Evaluated following 1 privilege target(s):
"Neos.Flow:Tests.RestrictedController.adminAction": ABSTAIN
(0 granted, 0 denied, 1 abstained)

Authenticated roles: Neos.Flow:Everybody, Neos.Flow:Anonymous" is not valid header value
```
Well... obviously it isn't a valid header value 😄 So the question is, why does this error message end up in a header?

albe:
See https://github.com/neos/flow-development-collection/pull/2756 for a possible fix. Closing this.

PR #2760, opened 2022-03-23
BUGFIX: Make Array.push accept `null` for array type casting

mhsdesign:
related: https://github.com/neos/neos-development-collection/pull/3658
fixes: https://github.com/neos/neos-development-collection/issues/3657

### DIFF
--- a/Neos.Eel/Classes/Helper/ArrayHelper.php
+++ b/Neos.Eel/Classes/Helper/ArrayHelper.php
@@ -351,8 +351,11 @@ class ArrayHelper implements ProtectedContextAwareInterface
         if ($array instanceof \Traversable) {
             $array = iterator_to_array($array);
         }
-        if (is_scalar($array) || $array === null) {
+        if (is_scalar($array)) {
             $array = [$array];
+        }
+        if ($array === null) {
+            $array = [];
         }
         if (is_array($array) === false) {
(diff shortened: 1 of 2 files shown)


Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -351,11 +351,11 @@ public function push($array, $element): array
         if ($array instanceof \Traversable) {
             $array = iterator_to_array($array);
         }
-        if (is_scalar($array)) {
+        if (is_scalar($array) || $array === null) {
             $array = [$array];
         }
```
kdambekalns, line 356: I did this locally:

```suggestion
        if (is_scalar($array)) {
            $array = [$array];
        }
        if ($array === null) {
            $array = [];
        }
```

That way NULL does not end up in the array. This is more similar to the code replaced with #3654, where NULL would be skipped.

jonnitto: I would also prefer @kdambekalns solution

kdambekalns (approved):
I like it! 😎

mhsdesign:
lol i was on this the same second ^^

jonnitto (approved):

kdambekalns:
Thanks for fixing the test, @jonnitto!

PR #2761, opened 2022-03-23
TASK: Add unit test (thanks @sorenmalling)

kdambekalns:
Related #2626

### DIFF
--- a/Neos.Flow/Tests/Unit/Http/Helper/ResponseInformationHelperTest.php
+++ b/Neos.Flow/Tests/Unit/Http/Helper/ResponseInformationHelperTest.php
@@ -5,6 +5,7 @@ namespace Neos\Flow\Tests\Unit\Http\Helper;
 
 use GuzzleHttp\Psr7\Response;
 use GuzzleHttp\Psr7\ServerRequest;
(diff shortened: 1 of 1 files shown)


sorenmalling:
The test fail is true, in locked guzzlehttp/psr7 version 1.4.1 the Utils class did not exists - instead it was this method to be used insteadhttps://github.com/guzzle/psr7/blob/1.4.1/src/functions.php#L78

kdambekalns:
> The test fail is true, in locked guzzlehttp/psr7 version 1.4.1 the Utils class did not exists - instead it was this method to be used instead: https://github.com/guzzle/psr7/blob/1.4.1/src/functions.php#L78

Doh. Ok, lowest deps… Oh my, what a mess.

mficzel (approved):
Fine by reading

kdambekalns:
> Doh. Ok, lowest deps… Oh my, what a mess.

Note to self: When upmerging, use Utils::streamFor() again in 7.1 and up, see https://github.com/neos/flow-development-collection/commit/fe83e962353d3eb1af7f0b90c03c2bdbef762cee

PR #2762, opened 2022-03-24
FEATURE: Create ValueObjects from simple types by calling `from$TYPE` for objects and JsonArray properties

mficzel:
When converting from simple types the denormalizer now checks wether the target class has a method `from$TYPE` and uses this to create the object.  Supported method names are `fromArray`, `fromString`, `fromBool`, `fromInt` and `fromFloat`.

For example the following controller action would use the methods `BirthDate::fromString` or `BirthDate::fromArray` (depending on submitted data) to instantiate the $birthDate argument.
```
public function indexAction(BirthDate $birthDate) 
```

In addition the JsonArray Type is extended to support this aswell. The serialization as flow entities takes precedence over the serialization as value object flow objects that implement \JsonSerializable are still serialized as Flow Objects in JsonArray properties. While beeing a Flow feature this allows to store and restore ValueObjects in Neos Node properties aswell. 

Note: This is a rebased version of https://github.com/neos/flow-development-collection/pull/2703 from @nezaniel 

Resolves: #2763
Replaces:  #2703

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
@@ -1,0 +1,182 @@
+<?php
+namespace Neos\Flow\Property\TypeConverter;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Property\Exception\TypeConverterException;
+use Neos\Flow\Property\PropertyMappingConfigurationInterface;
+use Neos\Flow\Property\TypeConverterInterface;
+
+final class DenormalizingObjectConverter implements TypeConverterInterface
+{
+    /**
+     * @return array<string>
+     */
+    public function getSupportedSourceTypes()
+    {
+        return ['array', 'string', 'bool', 'int', 'float'];
+    }
+
+    /**
+     * @return string
+     */
+    public function getSupportedTargetType()
+    {
+        return 'object';
+    }
+
+    /**
+     * @param mixed $source the source data
+     * @param string $originalTargetType the type we originally want to convert to
+     * @param PropertyMappingConfigurationInterface $configuration
+     * @return string
+     * @api
+     */
+    public function getTargetTypeForSource($source, $originalTargetType, PropertyMappingConfigurationInterface $configuration = null)
+    {
+        return $originalTargetType;
+    }
+
+    /**
+     * Return the priority of this TypeConverter. TypeConverters with a high priority are chosen before low priority.
+     *
+     * @return integer
+     * @api
+     */
+    public function getPriority()
+    {
+        return 100;
+    }
+
+    /**
+     * @param mixed $source the source data
+     * @param string $targetType the type to convert to.
+     * @return boolean true if this TypeConverter can convert from $source to $targetType, false otherwise.
+     * @api
+     */
+    public function canConvertFrom($source, $targetType)
+    {
+        return self::canConvertFromSourceType(gettype($source), $targetType);
+    }
+
+    /**
+     * @param string $sourceType
+     * @param string $targetType
+     * @return boolean
+     */
+    public static function canConvertFromSourceType(string $sourceType, string $targetType): bool
+    {
+        if (class_exists($targetType)) {
+            switch ($sourceType) {
+                case 'array':
+                    return method_exists($targetType, 'fromArray');
+                case 'string':
+                    return method_exists($targetType, 'fromString');
+                case 'bool':
+                case 'boolean':
+                    return method_exists($targetType, 'fromBool');
+                case 'int':
+                case 'integer':
+                    return method_exists($targetType, 'fromInt');
+                case 'double':
+                case 'float':
+                    return method_exists($targetType, 'fromFloat');
+                default:
+                    break;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @param string $targetType
+     * @return boolean
+     */
+    public static function isDenormalizable(string $targetType): bool
+    {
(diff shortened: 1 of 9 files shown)


mficzel:
@albe @bwaidelich @mhsdesign as you commented on the original pr and already liked it just not as a bugfix.
It would be great to get this approved for 8.0

Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
```diff
@@ -0,0 +1,182 @@
...
+                case 'array':
+                    return method_exists($targetType, 'fromArray');
+                case 'string':
+                    return method_exists($targetType, 'fromString');
+                case 'bool':
+                case 'boolean':
+                    return method_exists($targetType, 'fromBoolean');
+                case 'int':
+                case 'integer':
+                    return method_exists($targetType, 'fromInteger');
```
kdambekalns, line 88: Since the types are "officially" `bool` and `int`, why not `fromBool` and `fromInt`?

Less typing, too (think `fn` vs `function`, people seem to like that…)

kdambekalns (commented):
Looks good, left a "naming question".

Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
mficzel: That was already a matter of discussion on the old PR with @bwaidelich mentioning that we do not like abbreviations in method names https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartV/CodingGuideLines/PHP.html#naming

kdambekalns: I know… but in this case the potential for confusion can be reduced IMHO, when using the short names.

PHP in itself is inconsistent (_again_), as a type cast can be done using  `(bool)` or `(boolean)` – but a type declaration **must** use `bool`. Same for integers…

This means that the only form that always works is the short form – thus I'd use that for the method names, too. 🤷‍♂️

mficzel: Ok beeing in line with type annotation is a very good argument will adjust the code

kdambekalns (approved):
### 😘

Approval by reading…

One last question: Do you think this would deserve some documentation? Or should we rely on generated type converter reference?

mficzel:
@kdambekalns docs would be great but i am not entirely sure where to put what. Can be done later imho.
I wonder where those psalm errors come from. The mentioned lines are not changed in the pr.

mficzel:
Found and fixed the issues psalm detected. All green now.

nezaniel (approved):
Still looks good to me :)

mficzel:
@kdambekalns @nezaniel pushed a final adjustment to handle the value object serialization in JasonArray after the check for flow Objects. That should avoid problems wit flow entities that implement JsonSerializable

mficzel:
After chat with @nezaniel we agreed that the last change makes sense and improves confidence. Will merge with the existing reviews as this is a minor cosmetic change.

bwaidelich:
Oh no, I was one hour too late..
I really like the change but I'm super sad that we now have `fromInt` vs `fromInteger` etc (see https://github.com/neos/flow-development-collection/pull/2703#issuecomment-1047787088)

mficzel:
@bwaidelich can you decide on this with @kdambekalns i am really not that opinionated on this and until we push release we can still adjust this.

To me the argument to use the same syntax as in typeAnnotations made more sense.

mhsdesign:
i guess there is no right or wrong way - but if we decided to took road `fromInt` in the event sourcing already, i think we keep that course - (or we end up like php namings ^^)

bwaidelich:
OK, I'll try to rephrase my arguments for the full version.
To recap, it is about `fromInt()` and `fromBool()` (this PR) vs `fromInteger()` and `fromBoolean()` (Neos.EventSourcing):

IMO...
* ...method names are less "technical" than type hints. While it makes sense to use short versions for the latter, we usually try to [avoid abbreviations](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartV/CodingGuideLines/PHP.html#naming) in class and method names (and I think for a good reason)
* ...we use the full versions in our [TypeHandling](https://github.com/neos/flow-development-collection/blob/c6f0b21e307554e1e10bd2d8468a31ac82830119/Neos.Utility.ObjectHandling/Classes/TypeHandling.php#L79)
* ...the `Neos.EventSourcing` package comes with a [ValueObjectNormalizer](https://github.com/neos/Neos.EventSourcing/blob/master/Classes/EventStore/Normalizer/ValueObjectNormalizer.php) that uses the full names

In the end it probably boils down to preference. If you guys agree to my suggestion, I'm happy to create a follow-up PR with adjustments (and provide some documentation on the way).
Otherwise I would at least vote for a version that supports both `fromInt()` and `fromInteger()` etc.

mficzel:
@bwaidelich you had me at "and provide some documentation on the way"

mficzel:
Thinking further i would prefer to add the additional methods as a last minute feature. If we do not get this intuitively noone else will.

kdambekalns:
> While it makes sense to use short versions for the latter

No, it does not… not anymore than for anything else. And I find it unfortunate that PHP now uses both ways, but not in all places. 🤷‍♂️

> we usually try to [avoid abbreviations](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartV/CodingGuideLines/PHP.html#naming) in class and method names (and I think for a good reason)

I repeat my argument about the short form being more "universal" in the PHP context. We do support `int` in docblocks, too, after all.

But I didn't know about the same methods used in the EventSourcing package already, and I do see value in both ways.

So… whatever. 🙃

albe:
I agree to what Martin said - having both ways supported is probably the most friendly way to handle this in the core. I do not see any drawbacks from that, since there are no subtle differences in using one or the other. It's just there to remove mental load from developers, which is good. Let's do this @bwaidelich

bwaidelich:
There you go: #2766

Issue #2763, opened 2022-03-24
FEATURE: Make the JsonArrayType able to handle ValueObjects


mficzel:
JsonArrayType should be able to de/serialize arbitrary value objects, as long as they implement \JsonSerializable and ::from$Type

PR #2766, opened 2022-03-25
FEATURE: Extend DenormalizingObjectConverter to support fromBoolean and fromInteger

bwaidelich:
...in addition to their short version (`fromBool()` and `fromInt()`).
This also adds a short documentation about Value Object property mapping.

Related: #2763

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/DenormalizingObjectConverter.php
@@ -149,7 +149,7 @@ final class DenormalizingObjectConverter implements TypeConverterInterface
 
     /**
      * @param mixed $source
-     * @param string $targetType
+     * @param class-string $targetType
      * @return mixed
      * @throws TypeConverterException thrown in case a developer error occurred
      */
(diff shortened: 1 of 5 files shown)


mficzel (approved):
Fine by reading, huge thanks for the documentation fragment

mficzel:
@bwaidelich the psalm issues should be addressed still

bwaidelich:
>  the psalm issues should be addressed still

Doh, I never really know what to do with those.. can anyone help here?

bwaidelich:
> Those messages always make sense in hindsight.

Well, it makes sense, because
```php
method_exists($targetType, 'someString');
```
can only work if `$targetType` is a `class-string` or an `object`..
But I'm surprised that it didn't trap over
```php
$targetType::fromBool($source);
```
before (since even PhpStorm complains about that part).

But, more importantly.. I don't know how to get rid of the message :)
I tried replacing `string` by `string-class` in the param annotation, but that bubbles up..

mficzel:
Checking for class_exists convinced psalm. It is a bit redundant but i also agree that relying on the order methods are called ist also not great.

albe (approved):

albe:
Thanks a lot!

kdambekalns:
❤️

PR #2769, opened 2022-03-25
TASK: forceDirectAccess in ObjectAccess without reflection

kdambekalns:
This removes the use of reflection for direct access in `getProperty`
and `setProperty` in `ObjectAccess`.

Kudos to @bwaidelich for the code!

### DIFF
--- a/Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
+++ b/Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
@@ -115,9 +115,7 @@ abstract class ObjectAccess
 
         if ($forceDirectAccess === true) {
             if (property_exists($className, $propertyName)) {
-                $propertyReflection = new \ReflectionProperty($className, $propertyName);
-                $propertyReflection->setAccessible(true);
-                return $propertyReflection->getValue($subject);
+                return \Closure::bind(static fn &($subject) => $subject->{$propertyName}, null, $subject)($subject);
             }
             if (property_exists($subject, $propertyName)) {
                 return $subject->$propertyName;
@@ -249,9 +247,8 @@ abstract class ObjectAccess
         if ($forceDirectAccess === true) {
             $className = TypeHandling::getTypeForValue($subject);
             if (property_exists($className, $propertyName)) {
-                $propertyReflection = new \ReflectionProperty($className, $propertyName);
-                $propertyReflection->setAccessible(true);
-                $propertyReflection->setValue($subject, $propertyValue);
+                $propertyReference = &\Closure::bind(static fn &($subject) => $subject->{$propertyName}, null, $subject)($subject);
+                $propertyReference = $propertyValue;
             } else {
                 $subject->$propertyName = $propertyValue;
             }
--- a/Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
+++ b/Neos.Utility.ObjectHandling/Classes/ObjectAccess.php
@@ -115,7 +115,11 @@ abstract class ObjectAccess
 
         if ($forceDirectAccess === true) {
             if (property_exists($className, $propertyName)) {
-                return \Closure::bind(static fn &($subject) => $subject->{$propertyName}, null, $subject)($subject);
+                $bound = \Closure::bind(static fn &($subject) => $subject->{$propertyName}, null, $subject);
+                if ($bound === false) {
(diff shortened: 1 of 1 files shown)


kdambekalns:
Fancy stuff!

sorenmalling:
Wow!

albe:
🤯 that's some JS-level trickery. Anyone has any kind of performance benchmark to showcase rough difference to reflection?

kdambekalns:
> Anyone has any kind of performance benchmark to showcase rough difference to reflection?

Came up with a simple benchmark (see https://3v4l.org/Xg8pr#v8.1.4 and below). Suprising result (for me):

```
/opt/homebrew/bin/php /Users/karsten/Sites/neosdev/bench.php
ClosureAccess took 0.42281818389893 seconds
ReflectionAccess took 0.36359786987305 seconds
```

Over multiple runs on 8.0.16 ends up with the same numbers (0.42 vs 0.36). So, reflection isn't slower!

```php
<?php

class ClosureAccess {
    static public function getProperty(object $instance, string $propertyName) {
        $bound = \Closure::bind(static fn &($instance) => $instance->{$propertyName}, null, $instance);
        if ($bound === false) {
            throw new \RuntimeException('Could not bind to the subject to set the property "' . $propertyName . '" on the subject.', 1648229960);
        }
        return $bound($instance);
    }

    static public function setProperty(object $instance, string $propertyName, $propertyValue): void {
        $bound = \Closure::bind(static fn &($instance) => $instance->{$propertyName}, null, $instance);
        if ($bound === false) {
            throw new \RuntimeException('Could not bind to the subject to set the property "' . $propertyName . '" on the subject.', 1648229960);
        }
        $propertyReference = &$bound($instance);
        $propertyReference = $propertyValue;
    }
}

class ReflectionAccess {
    static public function getProperty(object $instance, string $propertyName, string $className) {
        $propertyReflection = new \ReflectionProperty($className, $propertyName);
        $propertyReflection->setAccessible(true);
        return $propertyReflection->getValue($instance);
    }

    static public function setProperty(object $instance, string $propertyName, $propertyValue, string $className): void {
        $propertyReflection = new \ReflectionProperty($className, $propertyName);
        $propertyReflection->setAccessible(true);
        $propertyReflection->setValue($instance, $propertyValue);
    }
}

class SomeClass {
    public function __construct(
        private string $someProperty
    ) {}
    
    public function getValue(): string {
        return $this->someProperty;
    }
}

$instance = new SomeClass('initial');

$time_start = microtime(true);
for ($i = 0 ; $i < 1000000 ; $i++) {
    ClosureAccess::getProperty($instance, 'someProperty');
    ClosureAccess::setProperty($instance, 'someProperty', 'CHANGED' . $i);
}
$time_end = microtime(true);
$time = $time_end - $time_start;
echo "ClosureAccess took $time seconds\n";

$time_start = microtime(true);
for ($i = 0 ; $i < 1000000 ; $i++) {
    ReflectionAccess::getProperty($instance, 'someProperty', SomeClass::class);
    ReflectionAccess::setProperty($instance, 'someProperty', 'CHANGED' . $i, SomeClass::class);
}
$time_end = microtime(true);
$time = $time_end - $time_start;
echo "ReflectionAccess took $time seconds\n";
```

albe:
> ReflectionAccess took 0.54619288444519 seconds
ClosureAccess took 0.92316293716431 seconds

Interesting. I switched the order of calls as well to check if the difference wasn't just some side effect of the code path. But the result doesn't change, so indeed Reflection is faster. Bummer.

mhsdesign:
to bad i was really hyped for this new fancy stuff ✨

bwaidelich:
Wow, thanks for giving this a try and actually benchmarking it!
I stumbled upon this trick in https://leanpub.com/thinking-functionally-in-php recently, but never tried it.

Now I found an old blog post by Ocramius claiming that Reflection was slower (on PHP 5.4 that was): https://ocramius.github.io/blog/accessing-private-php-class-members-without-reflection/

This more recent blog post supports "your" numbers though: https://www.lambda-out-loud.com/posts/accessing-private-properties-php/

It also mentions another, yet faster, method. But IMO it's not worth the magic

albe:
So I guess let's close this, as much as I'd have loved to have this a bit faster :)

bwaidelich:
@albe apparently with PHP 8.1 performance is no longer an issue: https://twitter.com/marcelpociot/status/1557381241246212096 (didn't verify)

PR #2771, opened 2022-03-25
!!!TASK: Remove deprecated custom ClassLoader

### LINKED ISSUE #2417: Make Composer autoloader the default

albe:
Resolves #2417


albe:
So apparently the autoload-dev section only applies when inside the root composer.json (https://getcomposer.org/doc/04-schema.md#autoload-dev), which in turn means we need to sync autoload-dev sections of all packages into our dev-distribution(s).

mhsdesign:
> So apparently the autoload-dev section only applies when inside the root composer.json (https://getcomposer.org/doc/04-schema.md#autoload-dev), which in turn means we need to sync autoload-dev sections of all packages into our dev-distribution(s).

hmm that means they were all the time useless? ^^

albe:
> hmm that means they were all the time useless? ^^

Not if you cloned a project from a single package (e.g. Neos.Flow)... which doesn't make sense or even works... so, yes 😂

PR #2772, opened 2022-03-25
TASK: Avoid deprecated Doctrine ORM proxy

kdambekalns:
The `Doctrine\ORM\Proxy\Proxy` is deprecated and extends the
`Doctrine\Persistence\Proxy`.


jonnitto (approved):
by reading 📖

mficzel (approved):
Fine by reading, i am pretty sure this would fail in the tests if it was dangerous

PR #2773, opened 2022-03-25
TASK: Drop nullability of $propertyPath on getPropertyPath()

kdambekalns:
That has been deprecated for ages and was to be dropped with 5.0
already.


jonnitto (approved):

mficzel (changes requested):
Generally approve but the red tests should be addressed

mficzel (approved):

PR #2774, opened 2022-03-25
TASK: Drop deprecated $packagePath from getFilteredPackages()

kdambekalns:


jonnitto (approved):

mficzel (changes requested):
This is dangerous unless Neos is migrated at the same time. The SiteController uses the method and needs adjustments.

kdambekalns:
> This is dangerous unless Neos is migrated at the same time. The SiteController uses the method and needs adjustments.

Yep, as already linked above: https://github.com/neos/neos-development-collection/pull/3671

mficzel (approved):

PR #2775, opened 2022-03-25
TASK: Remove support for Flow's custom class loader

kdambekalns:
This drops support for the Flow ClassLoader – except for functional
testing. There it's still used to load test classes & fixtures.

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -81,23 +81,18 @@ class Scripts
                 'namespace' => 'Neos\\Flow\\',
                 'classPath' => FLOW_PATH_FLOW . 'Classes/',
                 'mappingType' => ClassLoader::MAPPING_TYPE_PSR4
-            ]
-        ];
-
-        if ($bootstrap->getContext()->isTesting()) {
-            $initialClassLoaderMappings[] = [
+            ],
+            [
                 'namespace' => 'Neos\\Flow\\Tests\\',
                 'classPath' => FLOW_PATH_FLOW . 'Tests/',
                 'mappingType' => ClassLoader::MAPPING_TYPE_PSR4
-            ];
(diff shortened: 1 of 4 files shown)


kdambekalns:
Getting the tests to run without the class loader would be awesome, too…

albe:
See also https://github.com/neos/flow-development-collection/pull/2771 - I think we can get rid of the testing case if we manage to add autoload-dev section to the root composer manifest in the dev-distribution

mficzel:
@albe that would work for flow alone but not for Neos … only solution I see is having test base classes in a separate package that is a dev dependency

mficzel:
@albe that is not so easy as this would be required in all projects that use our unit tests as base. The clean way would be to extract the testcase base classes into seperate package that is dev-dependency. That way the tests for each package could be executed without depending on other packages tests.

mficzel (changes requested):
This one causes an error in Neos that should be addressed. Do not get yet why that happens exactly.

```
Required class "Neos\CliSetup\Command\SetupCommandController" could not be loaded properly for reflection from "Neos\CliSetup\Command\SetupCommandController".
```

The error may be a good sign and even correct.

albe:
> The clean way would be to extract the testcase base classes into seperate package that is dev-dependency. That way the tests for each package could be executed without depending on other packages tests.

But how would that make autoloading for `Neos\Foo\Tests\X` classes work? Those namespaces then need to be listed in the `autoload` section of all packages that contain tests, no? Which means cluttering the production autoloader with testing namespaces?

mficzel:
Maybe this will solve this problem https://github.com/neos/neos-development-collection/pull/3680

mficzel (approved):
With this change things broke that should not have worked in the first place. Fine by testing but not fully understanding.

albe (approved):
Fine as is, let's tackle the bigger fish for next major

PR #2781, opened 2022-03-28
TASK: Remove unused parameter in Redis flush

Sebobo:
The `frozen` key is already included in the prefixed argument in combination with the wildcard in the Lua script.
Therefore it can be safely removed. The script anyway didn’t use it.

The matching functional test works correctly.


kdambekalns (approved):
Good catch by @skurfuerst, even less code now. 👍

Thanks @Sebobo for taking care.

mficzel (approved):
Sounds legit.

PR #2782, opened 2022-03-28
FEATURE: Type declaration support for property injection

robertlemke:
This change introduces basic support for PHP 7.4 class property types
for property injection.

It allows for using types as follows:

```php
/**
 * @Flow\Inject
 **/
```

In order to allow this syntax, Lazy Property Injection is disabled
when a type is declared. To still use Lazy Property Injection, don't
use a PHP type declaration but a `@var` annotation:

```php
/**
 * @Flow\Inject
 * @var LoggerInterface
 **/
protected $logger;
```

If using both (type declaration _and_ `@var`), it is handled non-lazily.

Lazy Property Injection on properties with type declarations may 
be re-implemented as part of a future release.

#2114

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
@@ -561,7 +561,13 @@ class ConfigurationBuilder
                 if (!array_key_exists($propertyName, $properties)) {
                     /** @var Inject $injectAnnotation */
                     $injectAnnotation = $this->reflectionService->getPropertyAnnotation($className, $propertyName, Inject::class);
-                    $objectName = $injectAnnotation->name !== null ? $injectAnnotation->name : trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');
+                    $objectName = $injectAnnotation->name;
+                    if ($objectName === null) {
+                        $objectName = $this->reflectionService->getPropertyType($className, $propertyName);
+                    }
+                    if ($objectName === null) {
+                        $objectName = trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');
+                    }
                     $configurationProperty =  new ConfigurationProperty($propertyName, $objectName, ConfigurationProperty::PROPERTY_TYPES_OBJECT, null, $injectAnnotation->lazy);
                     $properties[$propertyName] = $configurationProperty;
                 }
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
@@ -561,14 +561,18 @@ class ConfigurationBuilder
                 if (!array_key_exists($propertyName, $properties)) {
                     /** @var Inject $injectAnnotation */
                     $injectAnnotation = $this->reflectionService->getPropertyAnnotation($className, $propertyName, Inject::class);
+                    $enableLazyInjection = $injectAnnotation->lazy;
                     $objectName = $injectAnnotation->name;
                     if ($objectName === null) {
                         $objectName = $this->reflectionService->getPropertyType($className, $propertyName);
+                        if ($objectName !== null) {
+                            $enableLazyInjection = false; # See:  https://github.com/neos/flow-development-collection/issues/2114
+                        }
                     }
                     if ($objectName === null) {
                         $objectName = trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');
                     }
-                    $configurationProperty =  new ConfigurationProperty($propertyName, $objectName, ConfigurationProperty::PROPERTY_TYPES_OBJECT, null, $injectAnnotation->lazy);
+                    $configurationProperty =  new ConfigurationProperty($propertyName, $objectName, ConfigurationProperty::PROPERTY_TYPES_OBJECT, null, $enableLazyInjection);
                     $properties[$propertyName] = $configurationProperty;
                 }
             }
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -91,6 +91,7 @@ class ReflectionService
     const DATA_PROPERTY_TAGS_VALUES = 14;
     const DATA_PROPERTY_ANNOTATIONS = 15;
     const DATA_PROPERTY_VISIBILITY = 24;
+    const DATA_PROPERTY_TYPE = 26;
     const DATA_PARAMETER_POSITION = 16;
     const DATA_PARAMETER_OPTIONAL = 17;
     const DATA_PARAMETER_TYPE = 18;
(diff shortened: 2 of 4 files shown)


robertlemke:
This change is a shortcut / alternative to https://github.com/neos/flow-development-collection/pull/2701

bwaidelich:
Awesome!
I don't (yet) understand the full code, but I can confirm that this fixes injection with type hints (and without `@var` annotation).

I would suggest, not to communicate the nullable version though but:
```php
/**
 * @Flow\Inject
 */
```
to make code more resilient and clarify that this is currently not lazily injected.

This is great, and will bring me back to using inject annotations in application code. Thanks! <3

bwaidelich (approved):
+1 by manual testing (not so much by reading yet)

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
```diff
@@ -561,7 +561,13 @@ protected function autowireProperties(array &$objectConfigurations)
...
                     $injectAnnotation = $this->reflectionService->getPropertyAnnotation($className, $propertyName, Inject::class);
-                    $objectName = $injectAnnotation->name !== null ? $injectAnnotation->name : trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');
+                    $objectName = $injectAnnotation->name;
+                    if ($objectName === null) {
+                        $objectName = $this->reflectionService->getPropertyType($className, $propertyName);
+                    }
+                    if ($objectName === null) {
+                        $objectName = trim(implode('', $this->reflectionService->getPropertyTagValues($className, $propertyName, 'var')), ' \\');
+                    }
                     $configurationProperty =  new ConfigurationProperty($propertyName, $objectName, ConfigurationProperty::PROPERTY_TYPES_OBJECT, null, $injectAnnotation->lazy);
```
kitsunet, line 571: Can't we just decide here to ignore `$injectAnnotation->lazy` IF  we use `$this->reflectionService->getPropertyType` because in "legacy" code cases lazy should work fine?

kitsunet:
I am not sure why we would sacrifice one feature (lazy injections) for another (PHP 7.4 property type hint support).

bwaidelich:
> I am not sure why we would sacrifice one feature (lazy injections) for another (PHP 7.4 property type hint support).

Hang on. I expected this to work as before if I decide to omit the type hint (which is still allowed in all PHP versions)
If this disables lazy DI in _any_ case I would change my vote to -1

Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
robertlemke: Okay, I gave it a try, see latest commit.

kitsunet (approved):
Cool! This seems like a sensible tradeoff I guess, will be interesting to compare performance of a code base written with property type hints and one without.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -996,6 +997,18 @@ public function getPropertyTagValues($className, $propertyName, $tag)
...
     }
 
+    /**
+     * Returns the property type
+     *
+     * @param $className
+     * @param $propertyName
+     * @return string|null
+     */
+    public function getPropertyType($className, $propertyName)
```
albe, line 1007:
```suggestion
    public function getPropertyType(string $className, string $propertyName): ?string
```
Anything speaking against typehinting this new method?

albe (approved):
Looks good by reading. Only left a comment for type-hinting the new ReflectionService method

Neos.Flow/Classes/Reflection/ReflectionService.php
robertlemke: Well, only that all the other methods in the Reflection Service also should have proper type hints … so I just did what the surrounding code did, so that it fits. To me it doesn't matter if we add the type hints now or when we clean up the whole class properly in a dedicated change.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1315,7 +1331,7 @@ public function reflectClassProperty($className, PropertyReflection $property)
             $this->classReflectionData[$className][self::DATA_CLASS_PROPERTIES][$propertyName][self::DATA_PROPERTY_TAGS_VALUES][$tagName] = $tagValues;
         }
 
-        foreach ($this->annotationReader->getPropertyAnnotations($property, $propertyName) as $annotation) {
```
kdambekalns, line 1318: Nice one… that parameter was passed since 2013 but not used? Ever? 🤪

kdambekalns (approved):
Awesome step into the right direction!

albe:
Another question:

> It allows for using types as follows:
```
/**
 * @Flow\Inject
 **/
protected ?LoggerInterface $logger;
```
What was the exact reason the nullability was required? It wasn't just the lazyness, because that would mean we don't need it for typed properties since the injection is eager (for now).
`"Typed property {x} must not be accessed before initialization"` was the PHP error and it's happening when a typed property is accessed that wasn't initialized (in the constructor or declaration). So my question is: do we somewhere in the proxy code or somewhere else attempt to bypass the constructor, or not set the property in the constructor?
Does anyone have the full proxy code for such a class around right now?

kdambekalns:
I adjusted the PR description, is that fine for you, @bwaidelich and @robertlemke?

bwaidelich (approved):
Love it, and I can confirm that 
```php
    /**
     * @Flow\Inject
     * @var LoggerInterface
     */
    protected $logger;
```
injects the `DependencyProxy` and
```php
    /**
     * @Flow\Inject
     * @var LoggerInterface
     */
```
(or without the `@var` annotation) injects the actual logger instance!

Thank you so much! <3

bwaidelich:
@kdambekalns 

> To still use Lazy Property Injection, declare the type as nullable

That's not correct though. When the type hint is specified, we now always fall back to eager injection.

robertlemke:
> What was the exact reason the nullability was required? It wasn't just the lazyness, because that would mean we don't need it for typed properties since the injection is eager (for now).
> `"Typed property {x} must not be accessed before initialization"` was the PHP error and it's happening when a typed property is accessed that wasn't initialized (in the constructor or declaration). So my question is: do we somewhere in the proxy code or somewhere else attempt to bypass the constructor, or not set the property in the constructor?
> Does anyone have the full proxy code for such a class around right now?

The error should only occur with lazy loading, so nullability is not required at the moment. And when we have support for union types, it will probably not even be allowed or at least does not make sense.

robertlemke:
@kdambekalns @bwaidelich I updated the description to correct that.

kdambekalns:
@albe asked
> What was the exact reason the nullability was required?

It was needed for `protected ?LoggerInterface $logger = null;` – something not in the example. Obviously I thought about that but missed that this isn't needed anymore…

kdambekalns:
I added a hint on using type _and_ `@var` to the description. Just to be very clear.

robertlemke:
So, we're all set?

albe:
So who wants to have the honor of pushing the merge button on this PR we have been all longing for? :) @robertlemke maybe, as he has put in all the hard grunt work.
We (Martin and I) will then branch off 8.0 this afternoon

robertlemke:
> So who wants to have the honor of pushing the merge button on this PR we have been all longing for? :) @robertlemke maybe, as he has put in all the hard grunt work.
> We (Martin and I) will then branch off 8.0 this afternoon

LOL, this PR is just a couple of minute's work ;-) Anyway, I'll take the chance.

albe:
> LOL, this PR is just a couple of minute's work ;-) Anyway, I'll take the chance.

It sometimes requires a lifetime of work to finally come to the solution that only takes 5 minutes worth of work. That doesn't devalue the lifetime spent before :)

PR #2783, opened 2022-03-28
FEATURE: Add ArrayObjectConverter

### LINKED ISSUE #2041: ObjectAccess behaviour broken on ArrayObject
### Description

When using `getGettablePropertyNames()` on an `ArrayObject`, the returned result includes not only the "real" properties of the object, but also internal `ArrayObject` properties. This affects `getGettableProperties()`, too.

### Steps to Reproduce

```php
$object = new \ArrayObject([
    'foo' => 'bar'
]);
var_dump(ObjectAccess::getGettableProperties($object));
```

#### Expected behavior

```
array(1) {
  ["foo"]=>
  string(3) "bar"
}
```

#### Actual behavior

```
array(4) {
  ["arrayCopy"]=>
  array(1) {
    ["foo"]=>
    string(3) "bar"
  }
  ["flags"]=>
  int(0)
  ["iterator"]=>
  object(ArrayIterator)#10459 (1) {
    ["storage":"ArrayIterator":private]=>
    object(ArrayObject)#10460 (1) {
      ["storage":"ArrayObject":private]=>
      array(1) {
        ["foo"]=>
        string(3) "bar"
      }
    }
  }
  ["iteratorClass"]=>
  string(13) "ArrayIterator"
}
```

### Affected Versions

Flow: since ages ago…

kdambekalns:
This adds an ArrayObjectConverter that allows to convert to a plain
array. It uses `getArrayCopy()` to get the job done.

Fixes #2041

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/ArrayObjectConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/ArrayObjectConverter.php
@@ -1,0 +1,62 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Flow\Property\TypeConverter;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Property\Exception\InvalidSourceException;
+use Neos\Flow\Property\PropertyMappingConfigurationInterface;
+
+/**
+ * Converter which transforms ArrayObjects to arrays.
+ *
+ * @api
+ * @Flow\Scope("singleton")
+ */
+class ArrayObjectConverter extends AbstractTypeConverter
+{
+    /**
+     * @var array<string>
+     */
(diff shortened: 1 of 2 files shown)


kdambekalns:
Created this against 6.3 even though it's a `FEATURE`. The original issue is something I consider a bug, if you agree we could sneak it in as a bugfix. If not, I'll re-target master.

bwaidelich (approved):
+1 by reading

This seems like a clean and b/c way to solve the issues.
But I abstain from the target branch discussion and leave that to the RMs to decide

albe:
I'm personally fine with having this in 6.3 as a BUGFIX, since the original issue is buggy behaviour of the normal `ArrayConverter` with `ArrayObject`s. So I would mark this as `BUGFIX` and reference the illbehaviour of the old converter in the description unless @mficzel vetos

albe (approved):

mficzel:
Am not sure, i would slightly prefer to target master but if you are sure this is safe why not

albe:
Well, this PR basically just adds a new converter that will be autopicked for `ArrayObject` sources and doesn't touch a single line of old code. So the only effect to existing code is that such objects now get type converted (and serialized) properly without all the internal properties that no one really wants to have anywhere.

mficzel:
As i said i am not against it as a bugfix. Seems to be safe enough. Feel free to merge.

PR #2784, opened 2022-03-28
!!! TASK: Remove whitelistObject() / add allowObject() in PersistenceManagerInterface

### LINKED ISSUE #2398: allowObject($object) commented out in PersistenceManagerInterface
### Description

I updated from neos/flow version 6.1.3 to 6.1.13 and noticed that whitelistObject($object)
in the PersitenceManagerInterface is now deprecated and i should use allowObject instead. So i changed all occurences to allowObject.

However when i ran phpstan and psalm, it complained for `Call to an undefined method Neos\Flow\Persistence\PersistenceManagerInterface::allowObject()` 

So i looked and found that the method is commented out in the PersitenceManagerInterface but not in the AbstractPersistenceManager implementing the PersistenceManagerInterface.

### Steps to Reproduce

1. change $this->persistenceManager->whitelistObject($object) usage to $this->persistenceManager->allowObject()
2. Run phpstan or psalm

#### Expected behavior

No errors in psalm or phpstan when using the non-deprecated method.

#### Actual behavior

errors in phpstan or psalm that allowObject() is undefined

### Affected Versions

Flow: version 6.1.7 up to 7.0.2 and master.

kdambekalns:
Resolves #2398

### DIFF
--- a/Neos.Flow/Classes/Persistence/PersistenceManagerInterface.php
+++ b/Neos.Flow/Classes/Persistence/PersistenceManagerInterface.php
@@ -168,19 +168,8 @@ interface PersistenceManagerInterface
      * @param object $object The object
      * @return void
      * @api
-     * @deprecated Use allowObject() instead. See https://github.com/neos/flow-development-collection/pull/2024
      */
-    public function whitelistObject($object): void;
-
-    /**
-     * Adds the given object to a list of allowed objects which may be persisted when persistAll() is called with the
-     * $onlyAllowedObjects flag. This is the case if "safe" HTTP request methods are used.
-     *
-     * @param object $object The object
-     * @return void
-     * @api
-     */
-    //public function allowObject($object): void;
+    public function allowObject($object);
 
     /**
      * Returns true, if an active connection to the persistence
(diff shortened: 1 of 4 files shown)


kitsunet (approved):

kdambekalns:
@kitsunet Please check again – does this adding of type declarations exceed the scope? If so, I'll split that into two PRs…

kitsunet (approved):
Seems fine to me, major version so makes sense, and I think this is all low level enough to ensure this is probably only ever called with the right types. Maybe needs breaking because it's the final removal of the deprecated method? Just to be sure

albe (approved):
❤️ thanks for this!

Neos.Flow/Classes/Persistence/PersistenceManagerInterface.php
```diff
@@ -159,7 +159,7 @@ public function remove($object): void;
      * @throws UnknownObjectException
      * @api
      */
-    public function update($object): void;
+    public function update(object $object): void;
 
     /**
      * Adds the given object to a list of allowed objects which may be persisted when persistAll() is called with the
```
albe, line 165: Note (to self): Adjust this line in case https://github.com/neos/flow-development-collection/pull/2585 gets merged

kdambekalns:
The type declarations caused errors and have been reverted in #2787

I created https://github.com/neos/flow-development-collection/pull/2788 for this, to be thoroughly tested/fixed.

PR #2786, opened 2022-03-28
BUGFIX: Correctly reflect builtin types in method arguments

albe:
Before builtin types like `object` or `iterable` in method arguments would lead to being reflected as if they were class types, leading to the AOP proxy builder prefixing those types with `\` which results in a PHP error "Type declaration 'object' must be unqualified". Only "simple" types were handled correctly as they had a separate check.
This adds a test that verifies that builtin types like `object` and `iterable` are correctly reflected as non-class types in the ReflectionService.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethod.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethod.php
@@ -224,7 +224,7 @@ class ProxyMethod
             $methodTags = $this->reflectionService->getMethodTagsValues($className, $methodName);
             $allowedTags = ['param', 'return', 'throws'];
             foreach ($methodTags as $tag => $values) {
-                if (in_array($tag, $allowedTags)) {
+                if (in_array($tag, $allowedTags, true)) {
                     if (count($values) === 0) {
                         $methodDocumentation .= '     * @' . $tag . "\n";
                     } else {
@@ -285,7 +285,7 @@ class ProxyMethod
                     if ($methodParameterInfo['optional'] === true) {
                         $rawDefaultValue = $methodParameterInfo['defaultValue'] ?? null;
                         if ($rawDefaultValue === null) {
-                            $defaultValue = ' = NULL';
+                            $defaultValue = ' = null';
                         } elseif (is_bool($rawDefaultValue)) {
                             $defaultValue = ($rawDefaultValue ? ' = true' : ' = false');
                         } elseif (is_numeric($rawDefaultValue)) {
@@ -341,7 +341,7 @@ class ProxyMethod
             $code .= (is_string($key)) ? "'" . $key . "'" : $key;
             $code .= ' => ';
             if ($value === null) {
-                $code .= 'NULL';
+                $code .= 'null';
             } elseif (is_bool($value)) {
                 $code .= ($value ? 'true' : 'false');
             } elseif (is_numeric($value)) {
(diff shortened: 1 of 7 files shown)


kdambekalns:
Ok, the change subtly changes the way "simple types" are reflected on method parameters… A "simple type", e.g. a string was not declared "scalar" before. In combination with the AOP proxy building, this now leads to type declarations being stricter than in the original class.

The reason is we now enter this branch:

https://github.com/neos/flow-development-collection/blob/f8aec784fd283e1e9fce707ee5032b3f924f8b59/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethod.php#L275-L276

which leads to e.g. `?string` being added in this case from `Session`:

`Fatal error:  Declaration of Neos\Flow\Session\Session::destroy(?string $reason = NULL) must be compatible with Neos\Flow\Session\Session_Original::destroy($reason = NULL) in /…/Flow_Object_Classes/Neos_Flow_Session_Session.php on line 1064`

albe:
> A "simple type", e.g. a string was **not** declared "scalar" before.

🤯 I totally misread the code there

kdambekalns:
> > A "simple type", e.g. a string was **not** declared "scalar" before.
> 
> 🤯 I totally misread the code there

To be fair, that scalar types are not makred as such is… counter-intuitive, at least. 🙃

albe:
Should we maybe adapt https://github.com/neos/flow-development-collection/blob/c4df7ec2520fc8cf85ae15c3570e0483d8c3782b/Neos.Flow/Classes/Reflection/ReflectionService.php#L1384-L1386 as well in this PR to keep the two places more in line? Note though that return types have `self` and `static`, which are **NOT** "builtin" (for whatever obscure reason, `void` is builtin though): https://3v4l.org/gCd1H#v8.0.17

kdambekalns (approved):
Count as half of an approval, as I was involved in tweaking the code…

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1381,7 +1381,7 @@ protected function reflectClassMethod($className, MethodReflection $method)
         }
 
         $returnType = $method->getDeclaredReturnType();
-        if ($returnType !== null && !in_array($returnType, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed']) && !TypeHandling::isSimpleType($returnType)) {
+        if ($returnType !== null && !in_array($returnType, ['self', 'static', 'parent']) && !$method->getReturnType()->isBuiltin()) {
```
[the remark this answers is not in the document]
albe: See https://github.com/neos/flow-development-collection/pull/2819 for a fix for return types in 6.3 without the builtin check

albe, line 1384:
```suggestion
        if ($returnType !== null && !in_array($returnType, ['self', 'static', 'parent']) && $method->getReturnType() !== null && !$method->getReturnType()->isBuiltin()) {
```

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethod.php
```diff
@@ -368,7 +368,8 @@ protected function getMethodVisibilityString()
         }
         if ($this->reflectionService->isMethodProtected($this->fullOriginalClassName, $this->methodName)) {
             return 'protected';
-        } elseif ($this->reflectionService->isMethodPrivate($this->fullOriginalClassName, $this->methodName)) {
+        }
+        if ($this->reflectionService->isMethodPrivate($this->fullOriginalClassName, $this->methodName)) {
```
mhsdesign, line 372: hehe good one xD

mhsdesign:
I dont understand why we have this and #2819 xD

albe:
> I dont understand why we have this and #2819 xD

tbh. I don't 100% know any more, but AFAIS the 6.3 PR only cares about the 'parent'/'static' cases and was supposed to be a quick fix for this common case, while this handles the more generic "isBuiltinType()" case, which could only be fixed properly in 7.0+

mhsdesign:
I think we wont need this fix for Flow 9 and even 8.3 seems to work fine?

PR #2787, opened 2022-03-28
Revert "TASK: Add type declarations in PersistenceManagerInterface"

mficzel:
This reverts commit 18386878d77b5d93ff976769f967a3713ce2d7e6. 

The return type `object` is not handled properly yet in proxy class building which causes trouble on the neos side. We are reverting this now as hotfix. The proper handling of `object` types can be added as bugfix. The type annotations for the PersistenceManagerInterface will have to wait until next major.

**Checklist**


albe (approved):

mficzel:
Merging without second review because we need a green neos again

kdambekalns (commented):
👍 post-merge - thanks for taking care.

PR #2788, opened 2022-03-28
TASK: Add type declarations in PersistenceManagerInterface

kdambekalns:


kdambekalns:
Some background (from Slack):

`Type declaration 'object' must be unqualified in /tmp/neos/Testing/Cache/Code/Flow_Object_Classes/Neos_Flow_Persistence_AbstractPersistenceManager.php on line 227`
when running
`../../bin/phpunit --colors --stop-on-failure -c ../../Build/BuildEssentials/PhpUnit/FunctionalTests.xml Neos.EventSourcedNeosAdjustments/Tests/Functional`

It is declared in `AbstractPersistenceManager::convertObjectToIdentityArray(object $object): array`

This is the generated code in the AOP proxy:
```php
public function convertObjectToIdentityArray(\object $object) : array
```

The `\` must not be rendererd

albe:
The issue is tried to be solved in https://github.com/neos/flow-development-collection/pull/2786
The root issue is that in https://github.com/neos/flow-development-collection/pull/2786/files#diff-7d0a1ee9681327fd82562aa2d7fcd5ed3b083a2283085f8c760eb4fab0e8c73aL1804 we check for the type to be "simple" with our definition of it being `array` or a scalar. It does not include any other builtin types though, like `object` (or `iterable`, etc. - anything not a class, interface or trait). This in turn means those types are considered to be "class" types and end up being rendered into the proxy with a leading slash to fully qualify the classname - this is where PHP complains, rightfully.

As to the possible solution, we could start adding any other possible types into `isSimpleType()`, but the method is used at many places that would require inspection first. So the idea was to only fix it at this specific place, similar to how it's done for the return type already (! - https://github.com/neos/flow-development-collection/blob/4f8d6ea97974ff1c1b5920554f9b004e4721f57b/Neos.Flow/Classes/Reflection/ReflectionService.php#L1384). In this case we already check for builtin type via reflection so that sounds reasonable to handle all builtins instead of hardcoding another list.

kdambekalns:
This seems to work fine with https://github.com/neos/flow-development-collection/pull/2786 in place…

kitsunet (approved):
Makes sense

kdambekalns:
Makes sense, yes – but should be next major (breaking by definition) and needs another PR (#2786) to be merged.

mhsdesign (approved):
Now targeting 9.0

PR #2789, opened 2022-03-28
!!!TASK: Update PSR Cache and SimpleCache Interfaces to 3.0

mficzel:
Raise the versions of the cache interfaces and add type annotations.

- psr/simple-cache: "^1.0" > "^3.0",
- psr/cache: "^1.0" > "^2.0",


kdambekalns:
Oh, this leads to an incompatibility with `google/auth`… in turn breaking the dependency chain if you need that (for `google/apiclient`, which is needed for e.g. `flowpack/googleapiclient`).

PR #2792, opened 2022-03-28
!!!TASK: Update psr/log to 2.0

albe:
This updates the psr/log interface to version 2.0 with type hints.


mficzel (approved):
Fine by reading, and testing

PR #2793, opened 2022-03-28
FEATURE: Provide central `PsrHttpFactory` that implements all interfaces from PSR 17

mficzel:
Injecting multiple factory objects to create PSR Http objects has been tedious in the past.

To overcome this the implementations of the PSR 17 methods are extracted into traits
and a central `PsrHttpFactory` is created. that implements all those interfaces.
That way a single object can be used to create Requests, Uris and Streams.

**Checklist**


mficzel:
I think the __construct method in the ServerRequestFactoryTrait needs some additional thinking.

Neos.Http.Factories/Classes/HttpFactory.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Neos\Http\Factories;
+
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ServerRequestFactoryInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Http\Message\UploadedFileFactoryInterface;
+use Psr\Http\Message\UriFactoryInterface;
+
+class HttpFactory implements ServerRequestFactoryInterface, RequestFactoryInterface, ResponseFactoryInterface, UriFactoryInterface, StreamFactoryInterface, UploadedFileFactoryInterface
```
kdambekalns, line 13: I think the name needs some more thinking. This factory doesn't "create HTTP"…

kdambekalns (commented):
I get the idea. Not sure it's that much of a burden, but having a single factory for all those objects would be nice. 👍

Just the name `HttpFactory` seems off…

Neos.Http.Factories/Classes/HttpFactory.php
mficzel: Agree ... suggestions are welcome aswell as an idea how to deal with the __constructor of the ServerRequestFactory

mficzel: Actually telling someone that the way to do a single http request is to inject 3 factories and plus Browser/Client (at least the engine is now injected by default) is an interesting experience.

kitsunet: yeha, I thought about that too, you can always sacrifice portability by directly instantiating a specific implementation that you then obvs need to require in your package.

kdambekalns: How is "supposed to work" according to "the spec"? Is there anything in the PSR or in Guzzle?

albe: > [](https://www.php-fig.org/psr/psr-17/#2-interfaces)2. Interfaces
The following interfaces MAY be implemented together within a single class or in separate classes.

https://www.php-fig.org/psr/psr-17/#2-interfaces

mficzel: The par looks to me weird but HttpObjectFactory sounds just fine

kitsunet: Either seem fine to me :) Although HttpObjectFactory sounds a bit smoother

kdambekalns: Indeed. That factory doesn't produce "HTTP", after all. ;)

mficzel: @kitsunet @kdambekalns renamed to HttpObjectFactory

kdambekalns (approved):

Neos.Http.Factories/Classes/HttpObjectFactory.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Neos\Http\Factories;
+
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ServerRequestFactoryInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Http\Message\UploadedFileFactoryInterface;
+use Psr\Http\Message\UriFactoryInterface;
+
+class HttpObjectFactory implements ServerRequestFactoryInterface, RequestFactoryInterface, ResponseFactoryInterface, UriFactoryInterface, StreamFactoryInterface, UploadedFileFactoryInterface
```
mhsdesign: One possible alternative name for this class could be PsrHttpFactory, as it clearly indicates that it implements the PSR HTTP message factories. Another option could be HttpMessageFactory, as it emphasizes the fact that this class is responsible for creating HTTP messages.

mhsdesign (changes requested):
...

mhsdesign (approved):

ahaeslich (approved):

Neos.Http.Factories/Classes/ServerRequestFactoryTrait.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * ServerRequestFactory constructor.
+     *
+     * @param UriFactoryInterface $uriFactory
+     * @param string $defaultUserAgent
+     * @param string $scriptPath
+     * @param string $defaultHttpVersion
+     */
+    public function __construct(
```
mhsdesign, line 46: i just noticed that this trait comes with a constructor which is barely usable in real word ... and an odd piece of code to look at :D Maybe we should fix this somehow?

PR #2796, opened 2022-03-29
BUGFIX: Address deprecation warnings from PHP 8.1

mficzel:
The trim method did not like beeing called with a null argument when the FLOW_CONTEXT environment was not set.

```
PHP Deprecated:  trim(): Passing null to parameter #1 ($string) of type string is deprecated in /Users/ficzel/PhpStormProjects/neos-development-distribution-8-0/Packages/Framework/Neos.Flow/Scripts/flow.php on line 68
```

### DIFF
--- a/Neos.Flow/Classes/Core/Bootstrap.php
+++ b/Neos.Flow/Classes/Core/Bootstrap.php
@@ -620,7 +620,7 @@ class Bootstrap
      * - $_SERVER[REDIRECT_ . $variableName] (again for php cgi environments)
      *
      * @param string $variableName
-     * @return string or NULL if this variable was not set at all.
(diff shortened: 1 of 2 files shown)


kdambekalns (approved):
A `(string)` cast would suffice, too…

mficzel:
@kdambekalns adjusted to `(string)`

albe (approved):

bwaidelich:
Are you aware that this change broke PHP 8.1 compatibility?

> PHP Deprecated:  trim(): Passing null to parameter #1 ($string) of type string is deprecated in Packages/Framework/Neos.Flow/Scripts/flow.php on line 68

PR #2805, opened 2022-03-30
TASK: Relax psr cache constraints

### LINKED ISSUE #2804: Dependency on psr/cache ^3.0 is too "eager"
It leads to an incompatibility with `google/auth`… in turn breaking the dependency chain if you need that (for `google/apiclient`, which is needed for e.g. `flowpack/googleapiclient`).

_Originally posted by @kdambekalns in https://github.com/neos/flow-development-collection/issues/2789#issuecomment-1083319800_

kdambekalns:
Fixes #2804

### DIFF
--- a/Neos.Cache/composer.json
+++ b/Neos.Cache/composer.json
@@ -6,8 +6,8 @@
     "license": ["MIT"],
     "require": {
         "php": "^8.0",
-        "psr/simple-cache": "^3.0",
-        "psr/cache": "~3.0",
+        "psr/simple-cache": "^2.0 || ^3.0",
+        "psr/cache": "^2.0 || ^3.0",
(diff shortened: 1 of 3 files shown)


kdambekalns:
I checked, and yes, one can `provide` multiple versions of something…

mficzel (approved):
Fine by reading. If the tests get green (especially with deps lowest) everything is fine.

albe (approved):

mficzel:
@albe wonder aswell, there has been one if remember correctly

albe:
See https://github.com/neos/flow-development-collection/pull/2806

kdambekalns:
Ok, this time with a lowest-dependencies check. 🤞

PR #2806, opened 2022-03-30
TASK: Reenable lowest dependencies build

albe:
This reenables the lowest dependencies build that was deactivated with https://github.com/neos/flow-development-collection/commit/cf8d28d87af78e71cc7cb6656172dfe6904f37db#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R38-R41

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -823,7 +823,7 @@ class ReflectionService
             $methodAnnotations = $this->annotationReader->getMethodAnnotations($method);
             if (PHP_MAJOR_VERSION >= 8) {
                 foreach ($method->getAttributes() as $attribute) {
-                    if ($this->isAttributeIgnored($attribute->getName()) {
+                    if ($this->isAttributeIgnored($attribute->getName())) {
                         continue;
                     }
                     $methodAnnotations[] = $attribute->newInstance();
@@ -1300,7 +1300,7 @@ class ReflectionService
         if (PHP_MAJOR_VERSION >= 8) {
             foreach ($class->getAttributes() as $attribute) {
                 $annotationClassName = $attribute->getName();
-                if ($this->isAttributeIgnored($annotationClassName) {
+                if ($this->isAttributeIgnored($annotationClassName)) {
                     continue;
                 }
                 $this->annotatedClasses[$annotationClassName][$className] = true;
@@ -1353,7 +1353,7 @@ class ReflectionService
         }
         if (PHP_MAJOR_VERSION >= 8) {
             foreach ($property->getAttributes() as $attribute) {
-                if ($this->isAttributeIgnored($attribute->getName()) {
+                if ($this->isAttributeIgnored($attribute->getName())) {
                     continue;
                 }
                 $this->classReflectionData[$className][self::DATA_CLASS_PROPERTIES][$propertyName][self::DATA_PROPERTY_ANNOTATIONS][$attribute->getName()][] = $attribute->newInstance();
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -774,6 +774,17 @@ class ReflectionService
     }
 
     /**
+     * Tells if a specific PHP attribute is to be ignored for reflection
+     */
+    public function isAttributeIgnored(string $attributeName): bool
+    {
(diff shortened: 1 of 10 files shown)


kdambekalns:
Hm… is it required for this to pass? In fact it shows the change is needed. 🤓

kdambekalns (approved):

albe:
Well, it means we have a problem with at least one dependency on the lowest version. Though I'm curious which dependency has a `ReturnTypeWillChange` annotation and states it works on PHP 8.0

kdambekalns:
I don't get why we need it. We can just as well fix the return type declaration!?

One cause is https://github.com/neos/flow-development-collection/commit/c19897130eb336e4725a309682687038f244a61e

kdambekalns:
> But it would still blow up if a third party dependency uses this attribute and claims PHP 8.0 support. Which is partly legit.

True. Fun fact: it seems from that change linked above only one annotation is left… 🤷‍♂️

kdambekalns:
We have a code block dealing with "PHP 8+" doing this in `ReflectionService`. The best solution would be to ignore any attribute class only available in 8.1. Where to get such a list? I'd not ingore all unknown classes…

kdambekalns:
Then again, it seems this is the only new attribute in 8.1

albe:
> The best solution would be to ignore any attribute class only available in 8.1. Where to get such a list? I'd not ingore all unknown classes…

> This is the first built-in attribute added to PHP.

Also https://php.watch/versions/8.1 mentions no other attribute change, so I guess it's not just the first, but also the only so far.

kdambekalns:
So we do this?

```php
            $method = new MethodReflection($className, $methodName);
            $methodAnnotations = $this->annotationReader->getMethodAnnotations($method);
            if (PHP_MAJOR_VERSION >= 8) {
                foreach ($method->getAttributes() as $attribute) {
                    if (PHP_MINOR_VERSION > 0 || $attribute->getName() !== 'ReturnTypeWillChange') {
                        $methodAnnotations[] = $attribute->newInstance();
                    }
                }
            }
```

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1286,6 +1300,9 @@ protected function reflectClass($className)
         if (PHP_MAJOR_VERSION >= 8) {
             foreach ($class->getAttributes() as $attribute) {
                 $annotationClassName = $attribute->getName();
+                if ($this->isAttributeIgnored($annotationClassName) {
```
kdambekalns, line 1303: I'll have another `)`, please!

kdambekalns, line 1303:
```suggestion
                if ($this->isAttributeIgnored($annotationClassName)) {
```

kdambekalns:
Nice! Almost! ❤️

albe:
> Nice! Almost! ❤️

I need to stop doing this in the Github editor 🤦‍♂️

kdambekalns:
> I need to stop doing this in the Github editor 🤦‍♂️

You did? I wouldn't have seen the other places there… 🙈

kdambekalns:
WUT?

`PHP Fatal error:  Array and string offset access syntax with curly braces is no longer supported in /home/runner/work/flow-development-collection/flow-development-collection/flow-development-distribution/Packages/Libraries/mikey179/vfsstream/src/main/php/org/bovigo/vfs/vfsStream.php on line 349`

kdambekalns:
Ah, we need to raise the minimum version of `mikey179/vfsstream` to … 1.6.7?

kdambekalns:
Wait, there are multiple places we require that… :/

albe:
> Wait, there are multiple places we require that… :/

Just noticed as well 🙈 okay... need to do this in the IDE

kdambekalns:
And I think we can use 1.6.10 (the latest) right away as minimum – fixes 8.1 deprecations as well. :)

kdambekalns:
😬 https://github.com/neos/flow-development-distribution/commit/1e386c4daba79abf8868830684b31ba247aa51b4

kdambekalns:
Now we are talking…

albe:
Still baffled how I missed that one occasion... I used intellij global search for `vfsstream":` and changed all occurences... 🤷‍♂️

kdambekalns:
Well, different repository. And I should have know right away, that dev-dependencies are only taken from the root manifest. 

LGTM

PR #2808, opened 2022-03-31
TASK: Handle ReflectionUnionType & ReflectionIntersectionType

kdambekalns:
This adds explicit handling of `ReflectionUnionType` & `ReflectionIntersectionType` instead of falling back to the deprecated `__toString()` method on `ReflectionType`.

Since we do no longer suppport PHP < 7.1 only the two mentioned types as well as `ReflectionNamedType` are supported, a "pure" `ReflectionType` can no longer be present at this point.

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1798,8 +1798,18 @@ class ReflectionService
 
         $parameterType = $parameter->getType();
         if ($parameterType !== null) {
-            // TODO: This needs to handle ReflectionUnionType
-            $parameterType = ($parameterType instanceof \ReflectionNamedType) ? $parameterType->getName() : $parameterType->__toString();
+            if ($parameterType instanceof \ReflectionUnionType || $parameterType instanceof \ReflectionIntersectionType) {
+                // ReflectionUnionType as of PHP 8, ReflectionIntersectionType as of PHP 8.1
+                $parameterType = implode('|', array_map(
+                    static function (\ReflectionNamedType $type) {
+                        return $type->getName();
+                    },
+                    $parameterType->getTypes()
+                ));
+            } else {
+                // ReflectionNamedType as of PHP 7.1
+                $parameterType = $parameterType->getName();
+            }
         }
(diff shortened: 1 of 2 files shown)


albe (approved):
Ah yes, I was stumbling across that as well yesterday or the day before and wondered at which version the __toString() was actually deprecated and if we don't need to remove that part.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1798,8 +1798,18 @@ protected function convertParameterReflectionToArray(ParameterReflection $parame
 
         $parameterType = $parameter->getType();
         if ($parameterType !== null) {
-            // TODO: This needs to handle ReflectionUnionType
-            $parameterType = ($parameterType instanceof \ReflectionNamedType) ? $parameterType->getName() : $parameterType->__toString();
+            if ($parameterType instanceof \ReflectionUnionType || $parameterType instanceof \ReflectionIntersectionType) {
+                // ReflectionUnionType as of PHP 8, ReflectionIntersectionType as of PHP 8.1
+                $parameterType = implode('|', array_map(
```
albe, line 1803: What do we do further with the reflection information where union types could blow things up?

kdambekalns: Nothing that would break now - the resulting string is the same as with `__toString()`.

**Except I made a mistake** and the intersection must be imploded using `&`!

albe:
Tagging in @mficzel for deciding if we merge this before 8.0.0 - IMO looks good and given no test breaks and Karsten confidently states this doesn't change the outcome over `__toString()`, I guess it's fine to get in.

albe:
Tagging in @mficzel for deciding if we merge this before 8.0.0 - IMO looks good and given no test breaks and Karsten confidently states this doesn't change the outcome over `__toString()`, I guess it's fine to get in.

kitsunet:
Do we actually want to back merge this as bugfix? Otherwise I would move it to 8.3 and merge it.

kdambekalns:
Changed base to 7.3 - but to me it's actually rather a bugfix, so that seems fine.

kitsunet (approved):

PR #2810, opened 2022-04-01
TASK: Remove interactive shell and deprecate infrastructure

kitsunet:
This removes the interactive shell functionality and deeprecates
the support classes for this. Those are still used by the behat
tests and everything should be moved to a separate package at
which point we will reintroduce the interactive shell in that
new package. At that point renaming will likely happen, therefore
the respective classes are marked deprecated. None of it was public
API before so this is not marked breaking.


robertlemke (approved):
Ha, it was so much fun developing this! But it was probably useless 🙃

kitsunet:
> Ha, it was so much fun developing this! But it was probably useless 🙃

Nothing speaks against an interactive shell package maybe 😆

mficzel (approved):
Fine by reading

PR #2813, opened 2022-04-04
BUGFIX: Uri must not be constructed with empty string

### LINKED ISSUE #2473: UriConstraints should make sure the Uri path starts with a slash

kdambekalns:
Fixes #2473


mficzel (approved):
Fine by reading

albe (approved):
Reads good, almost too easy

albe:
That was indeed "too easy" and doesn't resolve #2473 - see explanation at https://github.com/neos/flow-development-collection/pull/2501#issuecomment-1087385623

kdambekalns:
Hm. Should this be reverted then, as it was completely useless?

PR #2814, opened 2022-04-04
Revert "BUGFIX: Uri must not be constructed with empty string"

kdambekalns:
Reverts neos/flow-development-collection#2813


bwaidelich (approved):
Thanks! <3

Issue #2815, opened 2022-04-04
Routing: Request base URI is not taken into consideration when storing resolve results


bwaidelich:
The `RouterCachingService` currently builds the "resolveCacheIdentifier" taking into account
* `ResolveContext::isForceAbsoluteUri()`
* `ResolveContext::getUriPathPrefix()`
* and the resolved route values

It does not, however, consider `ResolveContext::getParameters()`.

As a result, `Router::resolve()` might return invalid results.

See steps to reproduce: https://github.com/neos/neos-development-collection/issues/3692

Related: #1126

PR #2816, opened 2022-04-04
BUGFIX: Respect RouteParameters in ResolveCache

bwaidelich:
Adjusts `RouterCachingService` such that it includes the specified
`RouteParameters` into the generated `resolveCacheIdentifier`.

Fixes: #2815


kdambekalns (approved):
Sounds as if that could fix the issue with wrong domains we tried to track down a while ago…

albe (approved):
Sounds and looks absolutely reasonable

PR #2818, opened 2022-04-04
BUGFIX: Prevent UriConstraints to generate invalid URIs

bwaidelich:
Adjusts `UriConstraints::applyTo()` such that it never creates
an `Uri` instance with a path of `/` (which according to `guzzle/psr` was
deprecated since 1.x and invalid with 2.x)

Fixes: #2473


bwaidelich:
FYI: I didn't add new tests, because this is all internal and I found no way to mock the `Uri` instance in `applyTo()`. But IMO it should fix guzzle compatibility and all existing tests are still green

kdambekalns (approved):
Another fine way of solving this.

albe (approved):
Looks good by reading - can not spot any obvious gotcha and I hope this solves the guzzle issue finally :) Thanks a lot for taking care!

PR #2827, opened 2022-04-06
TASK: Allow installing psr/log 3.0 and provide implementation for it

albe:
The change in https://github.com/neos/flow-development-collection/pull/2792 did only allow for psr/log 2.0 while already adding the return type hints that are only part of 3.0. This change allows installing both 2.0 || 3.0 like we do for the other psr interfaces and provides implementations for both versions. Thanks @jonnitto for bringing this up
See also https://github.com/neos/flow-development-collection/pull/2805


mficzel (approved):
I like

PR #2830, opened 2022-04-14
BUGFIX: PDO cache backend expects binary data

stephanschuler:
When variable frontends detect igbinary, chances are the inserted data
contains binary data that cannot be stored in regular MEDIUMTEXT columns.


stephanschuler:
@kdambekalns Oh sorry, didn't mean to leave this one in its current state. I remember not being able to write a proper test that actually uses MySQL. Wanted to close this useless PR and create a proper issue with detailed explanation but it seems I didn't save the issue at all 🙈 .

PR #2833, opened 2022-04-22
TASK: Use type casting instead of intval/floatval

simonschaufi:
Use type casting instead of slower function calls.

### DIFF
--- a/Neos.Eel/Classes/Helper/MathHelper.php
+++ b/Neos.Eel/Classes/Helper/MathHelper.php
@@ -398,7 +398,7 @@ class MathHelper implements ProtectedContextAwareInterface
      * The precision defines the number of digits after the decimal point.
      * Negative values are also supported (-1 rounds to full 10ths).
      *
-     * @param float $subject The value to round
+     * @param mixed $subject The value to round
      * @param integer $precision The precision (digits after decimal point) to use, defaults to 0
      * @return float The rounded value
      */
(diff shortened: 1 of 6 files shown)


kdambekalns:
**Checklist**

kdambekalns (approved):
Fine with me

kdambekalns:
BTW: Flow 5.3 is no longer supported, this should go into 6.3 directly. See https://www.neos.io/features/release-process.html

simonschaufi:
Then the readme is outdated. It still says 5.3. Could you please adjust that?

kdambekalns:
> Then the readme is outdated. It still says 5.3. Could you please adjust that?

Done in 6.3 (https://github.com/neos/flow-development-collection/blob/6.3/Readme.rst), will be upmerged eventually…

Neos.Eel/Classes/Helper/MathHelper.php
```diff
@@ -398,7 +398,7 @@ public function randomInt($min, $max)
      * The precision defines the number of digits after the decimal point.
      * Negative values are also supported (-1 rounds to full 10ths).
      *
-     * @param float $subject The value to round
+     * @param mixed $subject The value to round
```
kdambekalns, line 401: This makes Psalm happy and is technically correct, as the method handles any type.

albe (approved):

PR #2836, opened 2022-05-03
TASK: Support promoted properties in validator

### LINKED ISSUE #2835: Validator fails if class properties of parameter are not using @var annotations
### Steps to Reproduce

Given, there is an action as part of an Action Controller which expects an object as its argument:

```php
    /**
     * @param MyValueObject $myValueObject
     */
    public function someAction(MyValueObject $myValueObject): string
    { … }
```

The object's constructor uses a promoted property (`$theValue`):

```php
final class MyValueObject
{
    protected function __construct(readonly public string $theValue)
    {
    }

    public static function fromString(string $theValue): MyValueObject
    {
        return new MyValueObject($theValue);
    }
}

```

#### Expected behavior

The validator checking the arguments for the given action method should accept the value object.

#### Actual behavior

The validator fails with the following exception:

> Exception 1363778104: There is no @var annotation for property "theValue" in class "…\MyValueObject"

### Affected Versions

Flow: All currently active versions which should support PHP 7.4 (Flow 5.3 and higher)

robertlemke:
This change adds logic to the ValidatorResolver to accept properties
which were promoted in a constructor (PHP 8.1 promoted properties)
and thus do not have a @var annoation in a doc comment.

At the same time, @var annotations are not necessary anymore if a
type declaration was provided.

Resolves #2835

### DIFF
--- a/Neos.Flow/Classes/Validation/ValidatorResolver.php
+++ b/Neos.Flow/Classes/Validation/ValidatorResolver.php
@@ -14,6 +14,7 @@ namespace Neos\Flow\Validation;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\ObjectManagement\Configuration\Configuration;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
+use Neos\Flow\Reflection\PropertyReflection;
 use Neos\Flow\Reflection\ReflectionService;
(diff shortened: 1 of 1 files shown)


daniellienert (approved):
Looks good to me

kdambekalns (approved):

PR #2838, opened 2022-05-03
BUGFIX: Store cache content in BLOB field with PdoBackend

kdambekalns:
The cache `PdoBackend` used a `MEDIUMTEXT` column for the content, with
some special handling for PostgreSQL to fix null bytes in the content.

When using igbinary, even with MariadDB problems can occur, since the
written content is garbled when fetched back and cannot be unserialized
anymore, leading to cache misses / read errors.

This change fixes it by using a `MEDIUMBLOB` (`BYTEA` on PostgreSQL)
field for the content.

See #2830

**Upgrade instructions**

To make use of the new field, drop the `cache` table and run the cache setup again.

Another option is to switch the column type manually using a DB management tool
of your choice.

If you keep the old DB structure, aside from not having the fixed bug fixed:

- MariaDB doesn't care at all, the new code with the old DB structure works just fine.
- PostgreSQL will "work" but need a cache flush, as previously cached content shows wrong.

**Review instructions**

Hard to say, but you definitely need igbinary to run into the issue. Probably also a
specific kind of data that cannot be stored in MariaDB `MEDIUMTEXT`.

For PostgreSQL some code that used to do bin2hex/hex2bin on the data has been
removed, so if things still work as before, this is good.

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/PdoBackend.php
+++ b/Neos.Cache/Classes/Backend/PdoBackend.php
@@ -160,18 +160,19 @@ class PdoBackend extends IndependentAbstractBackend implements TaggableBackendIn
 
         $lifetime = ($lifetime === null) ? $this->defaultLifetime : $lifetime;
 
-        // Convert binary data into hexadecimal representation,
-        // because it is not allowed to store null bytes in PostgreSQL.
-        if ($this->pdoDriver === 'pgsql') {
-            $data = bin2hex($data);
-        }
 
         $this->databaseHandle->beginTransaction();
         try {
             $this->removeWithoutTransaction($entryIdentifier);
 
             $statementHandle = $this->databaseHandle->prepare('INSERT INTO "' . $this->cacheTableName . '" ("identifier", "context", "cache", "created", "lifetime", "content") VALUES (?, ?, ?, ?, ?, ?)');
-            $result = $statementHandle->execute([$entryIdentifier, $this->context(), $this->cacheIdentifier, time(), $lifetime, $data]);
+            $statementHandle->bindValue(1, $entryIdentifier);
+            $statementHandle->bindValue(2, $this->context());
+            $statementHandle->bindValue(3, $this->cacheIdentifier);
+            $statementHandle->bindValue(4, time(), \PDO::PARAM_INT);
+            $statementHandle->bindValue(5, $lifetime, \PDO::PARAM_INT);
+            $statementHandle->bindValue(6, $data, \PDO::PARAM_LOB);
+            $result = $statementHandle->execute();
             if ($result === false) {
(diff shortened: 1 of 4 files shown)


kitsunet (approved):
Great, makes sense!

bwaidelich:
Nice one.
I'm wondering: I guess this breaks if the upgrade instructions were not followed, right? (i.e. it is not compatible with the _old_ table scheme).

If that's so I would suggest to mark this breaking to make sure

kdambekalns:
> Nice one. I'm wondering: I guess this breaks if the upgrade instructions were not followed, right? (i.e. it is not compatible with the _old_ table scheme).

I'd have to actually try that… 🙈

kdambekalns:
> > Nice one. I'm wondering: I guess this breaks if the upgrade instructions were not followed, right? (i.e. it is not compatible with the _old_ table scheme).
> 
> I'd have to actually try that… 🙈

- MariaDB doesn't care at all, the new code with the old DB structure works just fine…
- PostgreSQL will "work" but need a cache flush, as previously cached content shows wrong.

bwaidelich (approved):
Thanks for checking!
+1 by reading

kdambekalns:
Hm, interesting – this fails on PHP 8 now that it's upmerged. 🤔

Sebobo:
@kdambekalns does this still work for you?

I get `hex2bin() expects parameter 1 to be string, resource given` in [`rewind`](https://github.com/neos/flow-development-collection/blob/8.2/Neos.Cache/Classes/Backend/PdoBackend.php#L619).
When I remove these lines it works as expected 🤔

Neos 7.3 + Flow 7.3

kdambekalns:
Doh. TBH I am not using this in production and rely on the tests (and reviews). 🙈

kdambekalns:
Which PHP version are you using, @Sebobo?

Sebobo:
@kdambekalns PHP 7.4.30 with igbinary

kdambekalns:
Can be reproduced, yay! https://github.com/neos/flow-development-collection/actions/runs/3241354597/jobs/5313181390

Sebobo:
Just as a note, this change was sadly actually breaking for my customer and created quite some issues.
They used the PDOBackend with a Persistent Cache and the stored data turned unreadable. Mostly without throwing exceptions 😔
The explanation contained the note about the non persistent cache which just needs flushing, but the persistent ones cannot simply be flushed in all cases.
The implementation in my customers project might have been risky, but nevertheless we have to be more careful.
Thx @kdambekalns for fixing the other issue so quickly ❤️

Also upgrade instructions might need to be added to the release notes.

kdambekalns:
True, I didn't really consider persistent caches on PostgreSQL — there should have been a big fat warning for those. Sorry!

PR #2844, opened 2022-05-18
FEATURE: Support object serialization without var annotation

### LINKED ISSUE #2843: FEATURE: Support object serialization without @var annotation

### Description

Flow uses the `ObjectSerializationTrait` to serialize objects for persistence and sessions. Until now the type of properties of serializable objects need to be configured with `@var` annotations in a respective doc block. Using just PHP 7.4 class property types, without a `@var` annotation, results in the following error:

> Warning: Undefined variable $className in /application/Packages/Framework/Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php line 70

Such an error occurs, for example, when the Flowpack Job Queue serializes a command in order to _defer_ a method call.

See also #2114 

### Possible Solution

Fix the code in `Neos\Flow\ObjectManagement\Proxy\ObjectSerializationTrait` to support type declarations and annotations.

robertlemke:
This change introduces support for PHP 7.4 style class property
type declarations for objects to be serialized by Flow. Previously,
a `@var` annotation had to be specified for each class property, in
order to communicate the type to Flow's serialization mechanism.

Resolves #2843


robertlemke:
Somewhat related to #2835

robertlemke:
See also #2114

kdambekalns (approved):

PR #2848, opened 2022-05-18
Fix stacktrace message in exception CLI output

robertlemke:
When the "logException" option is turned off for an exception handler, the CLI output still contained a reference to the stacktrace file like so:

> Open Data/Logs/Exceptions/20220518164330b1120c.txt for a full stack trace.

That was confusing, because the mentioned files never exist.

With this change, the following notice is displayed instead, when "logException" was turned off:

> The exception stacktrace was not logged, because "logException" is turned off.


kitsunet (approved):
Code change makes a lot of sense.

kdambekalns (approved):
Another mystery solved, yay!

kdambekalns:
Just one more thing… why not against Flow 6.3?

daniellienert:
@robertlemke / @kdambekalns this breaks a lot of sentry implementations including the punkt.de and the Flownative one, as the exceptionCli method signature is overwritten which now changed. (https://github.com/flownative/flow-sentry/blob/master/Classes/Exception/ProductionExceptionHandler.php).

daniellienert:
How to proceed here? Reverting this one, updating the sentry packages?

robertlemke:
I think we should consider the `echoExceptionCli()` method as not public / no API and other packages need to adjust their code. Thinking about it, the `AbstractExceptionHandler` shouldn't be extended at all by third-party packages. I'll prepare a new release for the [Flownative Sentry package](https://github.com/flownative/flow-sentry).

robertlemke:
FYI, I released [a new version of flownative/sentry](https://github.com/flownative/flow-sentry/releases/tag/v1.1.4) which works with previous and current Flow versions.

Issue #2849, opened 2022-05-28
BUG: DatesReader uses non-unique cache key for formatType dateTime which leads to cache collision between different languages 


daniellienert:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The DatesReader (https://github.com/neos/flow/blob/master/Classes/I18n/Cldr/Reader/DatesReader.php) caches the parsed format (which is a multi-dimensional array) by its format string (Like "d.m.Y").

For a dateTime this format just contains a placeholder like "{0} {1}".

With this, it's the same cache key for all languages. Which means, that the first format rendered is cached and used for all languages .

### Expected Behavior

DateTime is rendered in the locale as requested.

### Environment

```markdown
- Flow: Since 2010 ...
- PHP: irrelevant
```

PR #2850, opened 2022-05-30
BUGFIX: Use expanded pattern as cache key for dateTime format

daniellienert:
Closes: #2849

The cache key for a dateTime format is now the expanded format pattern instead of `{0} {1}`.


kdambekalns (approved):
Looks good…

Would it be possible to have a test that makes sure the actual results are not only different, but also what is expected? 🙊

daniellienert:
@kdambekalns done.

PR #2851, opened 2022-06-01
BUGFIX: InternalRequestEngine should behave like HttpRequestEngine

fcool:
The RequestHandler holds a reference to the "current" http
request in each step of the middleware chain. The internal request
engine does not, making it hard to write tests, that predict system
behavior.

With this patch you can fetch the current request (and not only the
outer most) even in InternalRequestEngine.

See \Neos\Flow\Http\RequestHandler::handleRequest()

Maybe this bugfix needs to be created against 7.0 - if I understand the support matrix the right way, that would be probably the lowest branch

**Checklist**


fcool:
Will come rebased against 7.0

fcool:
Nope. But 6.3. But this "different behavior" has been introduced with Flow 7.0 - before that (the middleware introduction) the internal request engine and the http request handler still behave identical.

fcool:
See new PR for lowest maintained branch here:
https://github.com/neos/flow-development-collection/pull/2854

PR #2853, opened 2022-06-05
BUGFIX: IfAuthenticatedViewHelper does not manage AuthenticationStrategy

### LINKED ISSUE #2852: BUG: IfAuthenticatedViewHelper does not manage AuthenticationStrategy

### Current Behavior

If we configure the Authentication Strategy to allTokens, the viewhelper IfAuthenticated returns true if one token is authenticated without checking if all tokens are authenticated.

### Expected Behavior

The viewhelper will have to return true only if all tokens are authenticated

### Steps To Reproduce

- configure the authenticationStrategy to allTokens
- configure at least 2 providers
- try to authenticate in a case where only one token is authenticated

### Environment

```markdown
- FluidAdaptor : 7.3.6
- PHP: 7.4
```

### Anything else?

_No response_

b3nkai:
Resolves #2852

The patch manage the authenticationStrategy configuration so the viewhelper will check if all tokens are authenticated in case of "allTokens" strategy.

For others stragegies, the behavior is the same as before.

### DIFF
--- a/Neos.FluidAdaptor/Classes/ViewHelpers/Security/IfAuthenticatedViewHelper.php
+++ b/Neos.FluidAdaptor/Classes/ViewHelpers/Security/IfAuthenticatedViewHelper.php
@@ -11,7 +11,6 @@ namespace Neos\FluidAdaptor\ViewHelpers\Security;
  * source code.
  */
 
-use Neos\Flow\Security\Authentication\TokenInterface;
 use Neos\Flow\Security\Context;
 use Neos\FluidAdaptor\Core\ViewHelper\AbstractConditionViewHelper;
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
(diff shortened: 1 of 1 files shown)


Neos.FluidAdaptor/Classes/ViewHelpers/Security/IfAuthenticatedViewHelper.php
```diff
@@ -79,8 +78,17 @@ protected static function evaluateCondition($arguments, RenderingContextInterfac
             return false;
         }
 
-        return array_reduce($securityContext->getAuthenticationTokens(), function (bool $isAuthenticated, TokenInterface $token) {
-            return $isAuthenticated || $token->isAuthenticated();
-        }, false);
+        $isAuthenticated = false;
+        $isAllAuthenticated = true;
```
sorenmalling, line 82: Would love to avoid a default "allow" value, for something related to security

sorenmalling:
Hi @b3nkai - thank you for your pull request 🥳  Great finding!

b3nkai:
Here's an other approach with counters.

albe:
Hey @b3nkai thanks for this, very good find indeed! The counting approach is good I think, but I'm currently thinking a bit deeper:
The reason this ViewHelper didn't cover other strategies was that we effectively wrongly duplicated knowledge about how to do this "isAuthenticated" check from the `AuthenticationProviderManager` instead of making the code DRY.
So I'm thinking about what the unintended side-effects would be, if we just got hold of the `AuthenticationProviderManager` and returned the `isAuthenticated()` method return value here?
```
/** @var AuthenticationProviderManager $authenticationProviderManager */
$authenticationProviderManager = $objectManager->get(AuthenticationProviderManager::class);

return $authenticationProviderManager !== null && $authenticationProviderManager->isAuthenticated();
```

First thing I see: `isAuthenticated()` will do an `authenticate()` call if not yet done (and swallow any exception, meh). Is it really intended that we do not attempt an `authenticate()` step from the ViewHelper if it hasn't happened yet? If so we'd need to segregate the two things (authenticate all tokens, check authentication status from authenticated tokens)
Maybe @kitsunet @kdambekalns ?

kitsunet:
I would love to use the authenticationProviderManager, I see the problem though that we just want to know if someone is authenticated and not necessarily start an authentication if not, as the template could very well be intended for anonymous audiences just with some extra hidden information. I wouldn't mind implementing something appropriate on the security side though, OR do we try/catch that in the VH then? I am not sure what the better way is, but being able to ask if someone is authenticated without triggering authentication seems sensible?

PR #2857, opened 2022-06-22
TASK: Remove noise from session LoggingAspect

kdambekalns:
This cleans up the log messages of the `LoggingAspect` for
`SessionInterface`, to reduce log volume.

Before:

    22-06-21 14:35:46 70133      INFO                           Session: Started session with id ixadt08m586zKtcJmUTZend6ykWvP9fM.
        [array] =>
            packageKey:
            [string] => Neos.Flow
            className:
            [string] => Neos\Flow\Session\Session
            methodName:
            [string] => start

After:

    22-06-21 14:35:46 70133      INFO                           Session: Started session with id ixadt08m586zKtcJmUTZend6ykWvP9fM.

No information is lost: The `Session` prefix to the log message is the class
name and will already be fully qualified if not within `Neos\Flow\Session`,
the package key was hardcoded and the method name is always what the
advice targets.

**Upgrade instructions**

If you for some reason relied on the log messages having that exact
format, you will need to adjust whatever you need to adjust.

**Review instructions**

The session activity should be logged as before, just less noisy.

I did not create this against 5.3, because, … well. But we could, if you like.

**Checklist**


kdambekalns:
Oh, Psalm… 🙄

albe (approved):
> Oh, Psalm… 🙄

We should rename the build step "3rd party method signature change watcher" 😂

I guess it's fine technically, just not sure about the targeting. Either it's considered "buggy", then it should probably also just go into 6.3 unless there are reasons to not do so. Or it's considered a breakish change, then it should ofc go into main.

mhsdesign (approved):
silence :D

PR #2860, opened 2022-06-23
BUGFIX: Reimplement request method override via middleware

### LINKED ISSUE #2856: Accidental? Feature removal: <form> tunnel request methods via field `__method`
This documentation is not up to date in projects since Flow 7.0:
https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Http.html?highlight=X-HTTP-Method-Override#request-methods

@kitsunet said
>  interesting, that seems to have been lost with the PSR-7 adoption, I found it back in Flow 5.3. We could/should re-add it

so to tunnel request methods like this wont work anymore:
`<input type="hidden" name="__method" value="PUT" />`

(in the code base is not a single occurrence of `__method`)

albe:
The documented feature to override the request method with the POST argument `__method` or one of the HTTP headers X-Http-Method-Override or X-Http-Method was accidentally removed in the transition towards PSR-7. This adds it back in by adding a middleware that does the same logic that was removed with the previous Http\Request class.

Resolves #2856

### DIFF
--- a/Neos.Flow/Classes/Http/Middleware/MethodOverrideMiddleware.php
+++ b/Neos.Flow/Classes/Http/Middleware/MethodOverrideMiddleware.php
@@ -3,7 +3,6 @@ declare(strict_types=1);
 
 namespace Neos\Flow\Http\Middleware;
 
-use Neos\Flow\Annotations as Flow;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
@@ -20,13 +19,14 @@ class MethodOverrideMiddleware implements MiddlewareInterface
     /**
      * Process an incoming server request and overwrite the request method of a POST based on a __method argument or either
      * the X-Http-Method-Override or X-Http-Method header.
+     * @noinspection CallableParameterUseCaseInTypeContextInspection
      */
     public function process(ServerRequestInterface $request, RequestHandlerInterface $next): ResponseInterface
     {
         if ($request->getMethod() === 'POST') {
-            $arguments = $request->getParsedBody();
-            if (isset($arguments['__method'])) {
-                $request = $request->withMethod($arguments['__method']);
+            $parsedBody = $request->getParsedBody();
+            if (isset($parsedBody['__method'])) {
+                $request = $request->withMethod($parsedBody['__method']);
             } elseif ($request->hasHeader('X-Http-Method-Override')) {
                 $request = $request->withMethod($request->getHeaderLine('X-Http-Method-Override'));
             } elseif ($request->hasHeader('X-Http-Method')) {
--- a/Neos.Flow/Classes/Http/Middleware/MethodOverrideMiddleware.php
+++ b/Neos.Flow/Classes/Http/Middleware/MethodOverrideMiddleware.php
@@ -35,5 +35,4 @@ class MethodOverrideMiddleware implements MiddlewareInterface
         }
         return $next->handle($request);
     }
-
 }
--- a/Neos.Flow/Classes/Http/Middleware/MethodOverrideMiddleware.php
+++ b/Neos.Flow/Classes/Http/Middleware/MethodOverrideMiddleware.php
@@ -1,0 +1,39 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Flow\Http\Middleware;
+
+use Neos\Flow\Annotations as Flow;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * Allows to override the request HTTP Method via different overrides in this order:
+ * - a "__method" argument passes in via request body
+ * - X-Http-Method-Override header
+ * - X-Http-Method header
+ */
+class MethodOverrideMiddleware implements MiddlewareInterface
+{
+    /**
+     * Process an incoming server request and overwrite the request method of a POST based on a __method argument or either
+     * the X-Http-Method-Override or X-Http-Method header.
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $next): ResponseInterface
+    {
+        if ($request->getMethod() === 'POST') {
+            $arguments = $request->getParsedBody();
+            if (isset($arguments['__method'])) {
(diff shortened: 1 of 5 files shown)


kitsunet (approved):
Thanks for this! Looks good to me

Neos.Flow/Configuration/Settings.Http.yaml
```diff
@@ -33,6 +33,9 @@ Neos:
         'trustedProxies':
           position: 'start 50'
           middleware: 'Neos\Flow\Http\Middleware\TrustedProxiesMiddleware'
+        'methodOverride':
+          position: 'start 30'
```
albe, line 37: I put it here instead of after the `parseBody` because we do want the method override to affect the routing (right?) and it previously also only worked with what was available in `$_POST`, i.e. the PHP parsed form/multipart request body. So this is in line with that.

Neos.Flow/Classes/Http/Middleware/MethodOverrideMiddleware.php
```diff
@@ -0,0 +1,39 @@
...
+class MethodOverrideMiddleware implements MiddlewareInterface
+{
+    /**
+     * Process an incoming server request and overwrite the request method of a POST based on a __method argument or either
+     * the X-Http-Method-Override or X-Http-Method header.
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $next): ResponseInterface
+    {
+        if ($request->getMethod() === 'POST') {
+            $arguments = $request->getParsedBody();
```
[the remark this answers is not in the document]
kitsunet: Good catch, it's filled with $_POST but not with eg. parsed JSON, for now I would argue that is fine, because would you want to bury the __method in a JSON request for an API? Seems tricky.

bwaidelich (changes requested):
Thank you!
Looks great, but I'm not sure if we should maybe just provide the MW but not _enable_ it by default (and document how to do that in the corresponding part of the docs).

Also this should be covered by a test so that we don't remove it again by accident :)
(and also to get a definite response on the `Request::getParsedBody()` question).

I'm happy to help with that, @albe !

Neos.Flow/Classes/Http/Middleware/MethodOverrideMiddleware.php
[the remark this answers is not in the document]
albe: ☝️ `$_POST` (https://github.com/guzzle/psr7/blob/master/src/ServerRequest.php#L179) so for POST requests, which we accept only, it contains the parsed form/multipart parameters.
If we did want to parse any other content types, like for JSON APIs, we'd need to juggle the order of `parseBody` to come *before* routing and `overrideMethod`. Doesn't feel right and isn't how it worked before either. We could document this though or add it as a feature in an upcoming release if it's seen useful

albe:
> Thank you! Looks great, but I'm not sure if we should maybe just provide the MW but not _enable_ it by default (and document how to do that in the corresponding part of the docs).

Well, I consider it a BUGFIX, so I do think it should be enabled by default to restore the previous functionality fully. I'd be fine with uncommenting the middleware in a new release and adjust the documentation accordingly. So we basically back out of default features somewhat, which isn't a bad thing.

> Also this should be covered by a test so that we don't remove it again by accident :)

Yes, I agree. And it should be a functional test IMO ;)

> I'm happy to help with that, @albe !

Any help welcome. I'm low on time and focus budget lately

bwaidelich (approved):
> Well, I consider it a BUGFIX, so I do think it should be enabled by default to restore the previous functionality fully.

Good point!

I did some minor cosmetic tweaks to get rid of PhpStorm Warnings and added unit and functional tests.

Neos.Flow/Tests/Unit/Http/Middleware/MethodOverrideMiddlewareTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $this->mockRequest->expects(self::once())->method('withMethod')->with($expectedMethod)->willReturn($mockAlteredRequest);
+        $this->mockRequestHandler->expects(self::once())->method('handle')->willReturnCallback(function($request) use ($mockAlteredRequest) {
+            self::assertSame($request, $mockAlteredRequest);
+            return $this->mockResponse;
+        });
+
+        $this->middleware->process($this->mockRequest, $this->mockRequestHandler);
+    }
+
+    public function process_nonMatchingRequests_dataProvider2(): \Traversable
```
albe, line 88:
```suggestion
    public function nonMatchingRequests_dataProvider2(): \Traversable
```
Doesn't it make more sense to name this after *what* it provides, rather than *for whom*? I'd even be inclined to strip off the `_dataProvider` alltogether.

Neos.Flow/Tests/Unit/Http/Middleware/MethodOverrideMiddlewareTest.php
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * @test
+     * @dataProvider process_matchingRequests_dataProvider
+     */
+    public function process_matchingRequests(string $method, array $headers, $parsedBody, string $expectedMethod): void
+    {
+        $this->mockRequest->method('getMethod')->willReturn($method);
+        $this->mockRequest->method('getParsedBody')->willReturn($parsedBody);
+        $this->mockRequest->method('hasHeader')->willReturnCallback(fn($header) => isset($headers[$header]));
+        $this->mockRequest->method('getHeaderLine')->willReturnCallback(fn($header) => $headers[$header]);
```
albe: When we're at it we could as well extract this part into a `prepareMockRequest(string $method, array $headers, $parsedBody)` method to make the tests themself more readable. I always found our tests way too hard to decypher because we never cared to replace PHPUnit technicalities with meaningful patterns.

albe (commented):
Thanks for the help with the tests! Just two small comments

bwaidelich:
I tweaked the unit test as suggested.
What's with the breaking [psalm test](https://github.com/neos/flow-development-collection/runs/7041953918?check_suite_focus=true) ? do we ignore these?

albe:
> I tweaked the unit test as suggested. What's with the breaking [psalm test](https://github.com/neos/flow-development-collection/runs/7041953918?check_suite_focus=true) ? do we ignore these?

I'd say yes. It's merely a discrepancy between latest doctrine docblocks and ours

mhsdesign (approved):
Thx for reimplementing it :P

Issue #2864, opened 2022-07-01
BUG: Proxy classes change default argument types for strings that only contain numbers


Sebobo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Given the following method in a controller class:

```php
public function doSomeAction(
    string $name,
    string $someId = '1',
    string $someOtherId = '823578798'
): void
```

Loading the controller would yield a type error as the proxy class method now looks like this:

```php
public function doSomeAction(
    string $name,
    string $someId = 1,
    string $someOtherId = 823578798
): void
```

The type error occurs as an integer is now the default argument for a string value:

`Fatal error: Default value for parameters with a string type can only be string or NULL`

### Expected Behavior

The default arguments should not be modified when the proxy class is generated.

### Steps To Reproduce

Create a new controller or add the above mentioned method to an existing controller.
Load the controller in the frontend and encounter the error.

### Environment

```markdown
- Flow: 7.3 probably affects all versions
- PHP: 7.4
```

kitsunet:
Happens in `\Neos\Flow\ObjectManagement\Proxy\ProxyMethod::buildMethodParametersCode` 
```

                        $rawDefaultValue = $methodParameterInfo['defaultValue'] ?? null;
                        if ($rawDefaultValue === null) {
                            $defaultValue = ' = NULL';
                        } elseif (is_bool($rawDefaultValue)) {
                            $defaultValue = ($rawDefaultValue ? ' = true' : ' = false');
                        } elseif (is_numeric($rawDefaultValue)) {
                            $defaultValue = ' = ' . $rawDefaultValue;
                        } elseif (is_string($rawDefaultValue)) {
                            $defaultValue = " = '" . $rawDefaultValue . "'";
                        } elseif (is_array($rawDefaultValue)) {
                            $defaultValue = ' = ' . $this->buildArraySetupCode($rawDefaultValue);
                        }
```

We should probably check is_string before is_numeric here... I can create a PR I guess.

Sebobo:
Yes I can confirm that switching the two checks fixes the issue. I can also create the PR.

kitsunet:
If you already have the Code around I guess do it. I don't know if we should mark it breaking just to get attention in case it breaks for someone (because this is a behavior change and someone with "dirty" code might get differnet results?)

PR #2865, opened 2022-07-01
BUGFIX: Don't convert numeric strings to numbers in proxy method argument defaults

Sebobo:
The check `is_numeric` evaluates to true for strings like `"1"` therefore the generated code would be missing the quotes
necessary to keep the default argument a string. This causes type errors in the proxy class when the arguments type is `string`.
Therefore the check whether an argument value is a string should be done first.

This change could be breaking if no type is used and a method somehow relies on the proxy class turning the argument default into a number.

Resolves: #2864

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethod.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethod.php
@@ -288,10 +288,10 @@ class ProxyMethod
                             $defaultValue = ' = NULL';
                         } elseif (is_bool($rawDefaultValue)) {
                             $defaultValue = ($rawDefaultValue ? ' = true' : ' = false');
-                        } elseif (is_numeric($rawDefaultValue)) {
-                            $defaultValue = ' = ' . $rawDefaultValue;
                         } elseif (is_string($rawDefaultValue)) {
                             $defaultValue = " = '" . $rawDefaultValue . "'";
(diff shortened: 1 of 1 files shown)


kitsunet (approved):
Thanks for taking care :)

kitsunet:
I don't understand these test failures. I think I have seen that somewhere else in the last days, but it makes no sense, the collection repo and the split repos both have that installer folder...

kdambekalns:
> I don't understand these test failures. 

My guess is, they are caused by:

```
neos/composer-plugin contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe. See https://getcomposer.org/allow-plugins
You can run "composer config --no-plugins allow-plugins.neos/composer-plugin [true|false]" to enable it (true) or keep it disabled and suppress this warning (false)
```

See (the new) PR I created:

- https://github.com/neos/neos-base-distribution/pull/57
- https://github.com/neos/flow-base-distribution/pull/17

**BUT** since the flow-development-collection has that setting I wondered… and noticed that this PR is against the 5.3 branch. But **Flow 5.3** is no longer supported. Let's see what happens when we change the target branch to 6.3

kdambekalns (approved):
Funny… I wonder when/why this broke. 🤔

albe (approved):

Sebobo:
@kdambekalns Argh yes, I'm so used now to target the same Flow and Neos versions 🤦‍♂️ thx!

PR #2871, opened 2022-07-07
BUGFIX: Allow updating of entities related to value objects after persisting

### LINKED ISSUE #2870: BUG: Calling PersistenceManager->persistAll() with scheduled ValueObjects twice breaks Unit Of Work
I'm not yet sure what the "minimal" set of actions triggers the behvior. So the complete story for now is:
1. Create new ValueObjects as related objects in a Doctrine lifecycle event (prePersist)
2. Call persistAll() as usual
3. Create more objects, creating new instances of the theoretically same ValueObject as before
4. Call persistAll() as usual
5. System crashes with 
```
A managed+dirty entity Acme\Package\Domain\Model\ValueObject@OID can not be scheduled for insertion.
```

### Expected Behavior

No crash

### Steps To Reproduce

see above

### Environment

```markdown
- Flow: 7.3.0 
- PHP: 7.4.30
```

### Anything else?

Probably the error has been introduced with this two commits. Reverting to the behavior to "as before" but with `spl_object_id` instead of `spl_object_hash` solves the problem. So probably it exists since 6.3.13

https://github.com/neos/flow-development-collection/commit/2e914c06eb9b68335357d3f974a6f907deaba912
https://github.com/neos/flow-development-collection/commit/50d3ff3c47591125c17e6ee237c919acfbb0a5ac

fcool:
Fixes #2870

**Upgrade instructions**

**Review instructions**

**Checklist**


Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationAndDeDuplicationListener.php
```diff
@@ -89,20 +90,21 @@ private function deduplicateValueObjectInsertions()
...
                 if (isset($knownValueObjects[$className][$identifier]) || $unitOfWork->getEntityPersister($className)->exists($entity)) {
-                    $unitOfWork->scheduleForDelete($entity);
+                    unset($entityInsertions[$oid]);
                     continue;
                 }
 
                 $knownValueObjects[$className][$identifier] = true;
             }
         }
+        ObjectAccess::setProperty($unitOfWork, 'entityInsertions', $entityInsertions, true);
```
kitsunet, line 107: scary shit, but seems there is no better way :/

fcool: It would be great, if we could modify the result of a constructor. So that `new ValueObject()` returns a managed entity, if there already is a fitting ValueObject in the database. But I found no way for a stunt like that.

Probably one could try to adjust the ProxyGenerator to replace construction of ValueObjects with some service glue... but thats definitely no low hanging fruit.

The object deduplication is in its current way a scary hack... even `scheduleForDelete` does not really change that. It is public, but marked as internal - so we are definitely doing something "unsupported" here.

albe (approved):
Thanks for providing a test with this! Too bad we can't easily phase out our hard internal knowledge about doctrine entityInsertion maps, but `scheduleForDelete()` was a bad solution probably from the beginning :|

Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationAndDeDuplicationListener.php
kdambekalns: As long as the scary shit is not replaced by even scarier shit… 🤷‍♂️

kdambekalns (approved):

Issue #2872, opened 2022-07-13
FEATURE: typed env variable replacements in Settings


bwaidelich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Placeholders like `%env:<ENV_VARIABLE>%` can be used in all configurations and will be replaced with the value of their corresponding environment variable (see https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/Configuration.html#constants-and-environment)

Unfortunately this only works for *strings*.

It would be helpful to be able to define a format so that the replacement works for settings of other types (bool, int, float).

This could also be used to solve another issue with the env placeholders:
When referring to a _non-existing_ environment variable, the setting will be casted to `false` currently.
Although we can't change this behavior for backwards-compatibility reasons it can be made more explicit by specifying a type.

### Possible Solution

Suggested format (in addition to the already existing syntax):
```
%env(<type>):<ENV_VARIABLE>%
```
Where `<type>` can be one of `int`, `bool`, `float` or `string`.

Example:
```yaml
someSetting: %env(int):SOME_ENV_VARIABLE%
```

PR #2873, opened 2022-07-13
FEATURE: typed env variable replacements in Settings

bwaidelich:
With this change, the type of environment variables can be specified in Configuration files:

```yaml
someSetting: %env(int):SOME_ENVIRONMENT_VARIABLE%`
```

This will cast the corresponding value to the specified type (and to the base value of the type when referring to undefined environment variables).

Supported types and their base value are:

* ``int``: 0
* ``bool``: false
* ``float``: 0.0
* ``string``: "" (empty string)

Resolves: #2872


Neos.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
```diff
@@ -784,6 +784,58 @@ public function replaceVariablesInPhpStringReplacesEnvMarkersWithEnvironmentValu
...
+        yield 'concatenating format bool' => ['envVarName' => 'NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR', 'envVarValue' => 'true', 'setting' => 'prefix%env(bool):NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR%suffix', 'expectedResult' => 'prefix1suffix'];
+        yield 'format bool casts non-existing env variable to false' => ['envVarName' => '', 'envVarValue' => '', 'setting' => '%env(bool):NEOS_FLOW_TESTS_UNIT_CONFIGURATION_NON_EXISTING_ENVIRONMENT_VARIABLE%', 'expectedResult' => false];
+
+        yield 'format float casts string to 0.0' => ['envVarName' => 'NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR', 'envVarValue' => 'foo', 'setting' => '%env(float):NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR%', 'expectedResult' => 0.0];
+        yield 'format float casts "1,4" to 1.0' => ['envVarName' => 'NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR', 'envVarValue' => '1,4', 'setting' => '%env(float):NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR%', 'expectedResult' => 1.0];
```
bwaidelich, line 810: weird but.. PHP

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -512,6 +513,10 @@ protected function replaceVariablesInPhpString(string $phpString): string
             $replacement .= ($constantDoesNotStartAsBeginning ? $matchGroup['startString'] . "' . " : '=> ');
 
             if (isset($matchGroup['prefix']) && $matchGroup['prefix'] === 'env') {
+                $allowedTypes = ['int', 'bool', 'float', 'string'];
```
mhsdesign, line 516: hold on we dont use this line :P

Neos.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
```diff
@@ -784,6 +784,58 @@ public function replaceVariablesInPhpStringReplacesEnvMarkersWithEnvironmentValu
...
+        yield 'format int casts non-existing env variable to 0' => ['envVarName' => '', 'envVarValue' => '', 'setting' => '%env(int):NEOS_FLOW_TESTS_UNIT_CONFIGURATION_NON_EXISTING_ENVIRONMENT_VARIABLE%', 'expectedResult' => 0];
+
+        yield 'format bool casts "1" to true' => ['envVarName' => 'NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR', 'envVarValue' => '1', 'setting' => '%env(bool):NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR%', 'expectedResult' => true];
+        yield 'format bool casts "true" to true' => ['envVarName' => 'NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR', 'envVarValue' => 'true', 'setting' => '%env(bool):NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR%', 'expectedResult' => true];
+        yield 'format bool casts "false" to true' => ['envVarName' => 'NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR', 'envVarValue' => 'false', 'setting' => '%env(bool):NEOS_FLOW_TESTS_UNIT_CONFIGURATION_CONFIGURATIONMANAGERTEST_MOCKENVVAR%', 'expectedResult' => true];
```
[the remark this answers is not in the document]
mhsdesign: wait a sec this a usecase to even write `"false"` in an env variable? If people do that i think we should cast it to false - otherwise this would be confusing

Neos.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
mhsdesign: understandable - because the kommata ...

Neos.Flow/Classes/Configuration/ConfigurationManager.php
bwaidelich, line 516:
```suggestion
```

bwaidelich: you're right. My first version threw an exception for unknown types

Neos.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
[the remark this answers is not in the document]
bwaidelich: I used PHPs default casting behavior.. But you're right. Maybe we should treat "false", "FALSE", "no", ... differently for boolean

bwaidelich (commented):
thanks for the feedback, I'll tweak the PR tomorrow

Neos.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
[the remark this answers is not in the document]
mhsdesign: but as said if this is not good practice at all than we dont need it i think...

Neos.Flow/Classes/Configuration/ConfigurationManager.php
kitsunet: Haha, someone I mentally saw that exception while reviewing. At first reading I also wondered where this is used, but somehow then I thought I had seen the Exception

mhsdesign: yes me too :joy:

Neos.Flow/Tests/Unit/Configuration/ConfigurationManagerTest.php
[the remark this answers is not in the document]
bwaidelich: > but as said if this is not good practice at all than we dont need it i think...

In fact that was the trigger for this issue: We used env Flags like "ENABLE_TRACKING=false" and then had to check for the actual string `false` in all places.
I think there is no harm in evaluating the string "false" to `false` as well if the type is set to `bool`

mhsdesign (approved):
Looks good!

bwaidelich:
Strings like "false" (or "FALSE") are now evaluated to `false` when casting to boolean, thanks @mhsdesign for the heads-up. @kitsunet feel free to hit the merge button if you still agree :)

PR #2874, opened 2022-07-15
FEATURE: Add method `askQuestion` to console output that accepts a `Question` object

mficzel:
The other question methods take arrays or strings as arguments which is easy and convenient however the underlying symfony console offers more flexibility.
This pr adds the method `askQestion` that accepts a smfony console question object and thus allows the full flexibility at the price of complexity.

In addition the method `getQuestionHelper` is made public to make this as accessible as the `getProgressBar` method.

**Review instructions**

I chose not to allow Question objects in the `ask` method because this method has a default argument that only makes sense when the question is passed as array or string.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Cli/ConsoleOutput.php
+++ b/Neos.Flow/Classes/Cli/ConsoleOutput.php
@@ -191,6 +191,18 @@ class ConsoleOutput
     }
 
     /**
+     * Asks a question to the user
+     *
+     * @param Question $question The question to ask as an Object.
+     * @return mixed The user answer
+     * @throws \RuntimeException If there is no data to read in the input stream
+     */
+    public function askQuestion(Question $question)
+    {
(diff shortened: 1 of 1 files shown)


mhsdesign:
I do agree with the idea - but in the bigger picture it now doesnt make sense that we have two ways to archive the same thing - that is confusing ... either abstract smfony or use smfony directly ... i think we need to deprecate the current way then?

But this can obviously degenerate into a larger discussion ^^

mficzel:
The main problem is that the abstraction by the "select" method does not cover possible ways for asking questions.

kitsunet:
I can totally see dropping more abstractions in the future, we have lived in this in between of abstraction and direct access to symfony cli methods fro a while now, I don't mind adding more direct symfony, but it's good to discuss if we actually want to fully and directly embrace it in the future, but that discussion is for discuss I would say (or a feature ticket).

kitsunet (approved):

PR #2880, opened 2022-08-11
TASK: Raise guzzlehttp/psr7 to ^1.8.4 (security fix)

kdambekalns:
This makes sure `guzzlehttp/psr7` is will not be used in version with a known
security issue.

**Checklist**

### DIFF
--- a/Neos.Http.Factories/composer.json
+++ b/Neos.Http.Factories/composer.json
@@ -9,7 +9,7 @@
     "require": {
         "php": "^7.1",
         "psr/http-factory": "^1.0",
-        "guzzlehttp/psr7": "^1.4.1, !=1.8.0"
+        "guzzlehttp/psr7": "^1.8.4"
     },
     "autoload": {
         "psr-4": {


Neos.Http.Factories/composer.json
```diff
@@ -9,7 +9,7 @@
     "require": {
         "php": "^7.1",
         "psr/http-factory": "^1.0",
-        "guzzlehttp/psr7": "^1.4.1, !=1.8.0"
```
bwaidelich, line 12: Why did we _disallow_ 1.8.0?

bwaidelich (approved):
To give some context: Apparently 1.8.4 is the first version without known vulnerabilities (according to https://packagist.org/packages/guzzlehttp/psr7).
It's a shame that they didn't/couldn't fix this in a patch release, but I wonder if we should mark this as a breaking change?

Neos.Http.Factories/composer.json
kdambekalns: Because of https://github.com/neos/flow-development-collection/pull/2433

kdambekalns:
> It's a shame that they didn't/couldn't fix this in a patch release, but I wonder if we should mark this as a breaking change?

Semver says 1.8.4 is drop-in-compatible with anything 1.0, so they did fix it in a good-enough manner, I'd say. Conversely I don't see why this would be a breaking change?

kitsunet:
If we assume semver this shouldn't be breaking? And I guess we should assume that.

kdambekalns:
I checked and at least there are no additional restrictions composer-wise that the newer versions pull in that would break stuff. 🤷‍♂️

bwaidelich:
I meant breaking as in: you patch Flow and can no longer install library x because it relies on a different version of guzzle..
But discard that comment – in that case those packages have to be patched, too, anyways!

PR #2883, opened 2022-08-19
BUGFIX: Avoid errors when `./flow help` is shown for an action without doc-block

mficzel:
The `./flow help` command currently throws errors when rendering details for command that has no doc-block.
Since modern php and flow need those doc-blocks less and less this should be supported aswell.

This change will return the argument name as description when no @param annotation is found for a parameter.
This is not as helpful as a description but shows the parameter and avoids an unnecessary error.

**Review instructions**

Remove the doc comments from a cli-command and run `./flow help command:name` ... without this change the code will end throwing an exception.

Fixes: #2886

**Checklist**


robertlemke:
My gut feeling is that we may see more undocumented commands when we start accepting documentation-less command controllers.

What if we display a helpful message instead, which explains how to properly document a command? 

I fully agree that in modern PHP we don't need doc comments for code documentation, but in this particular case we need it for providing meaningful help texts. I currently don't imagine a situation where it would be okay to leave a command completely undocumented.

mficzel:
@nezaniel do you agree with @robert? I would be fine with that and adjust the pr but since I stumbled over this multiple times with controllers from you it may not be the obvious solution.

@robertlemke I’ve seen many controllers that hardly provide more informations than the types so one could argue that this documentation enforcement does not work anyways. That would speak in favor of the current pr.

robertlemke:
> I’ve seen many controllers that hardly provide more informations than the types so one could argue that this documentation enforcement does not work anyways.

Then we should add some AI process which checks if the doc comment is meaningful and conforms with The Chicago Manual of Style 😬

I don't want to block this change, so feel free to decide as you see fit. However, I see this as an opportunity to suggest developers to provide proper documentation.

mficzel:
@robertlemke the reason I am sceptic is that only devs that actually test flow help will notice and adjust but those would do that already.

On the other hand the users that use code from a lazy dev are currently left with a non helpful error and have to llook into the code for the supported arguments which is exactly the situation I found myself in.

robertlemke:
@mficzel okay, I'm convinced.

> we could ensure that commandcontroller require a docblock by throwing the same error when the command is executed 

I'd like that. We can try something like that in a separate PR.

robertlemke (approved):
LGTM

mhsdesign (approved):
I do understand both of your opinions, and it may be that through the occasional error one will remember to write the annotations (happened to me ^^) and otherwise this 'doc' knowledge will be lost over time.
On the other hand, it's not cool that "flow help" will throw when you need its help the most :P

so the error should be thrown, but rather at creation of a command controller that on execution.
and for those with php-doc-phobia, we should provide php attribute annotations ^^

bwaidelich:
Please no exceptions for commands without doc comments. I use commands a lot during dev time (or internal commands that I still would like to get basic help for if needed).

IMO an exception is not a good way to encourage folks to write documentation :)

bwaidelich (approved):

PR #2884, opened 2022-08-24
BUGFIX: Reset cache backend iterator state on update

### LINKED ISSUE #2828: PdoBackend: Iterator state drift compared to database content
### Description

The PdoBackend implements the IterableBackendInterface.
To do so, it keeps an internal object of type "ArrayAccess" named $cacheEntriesIterator holding all keys and values of the database once initialized.

When the cache state changes during runtime by e.g. flushing, the $cacheEntriesIterator holds on to the previous data.

This occurred to me in functional tests where (at least in my setup) the very same PdoBackend object is used for multiple tests.

As far as I can tell, the same holds true for Apcu and SimpleFile backends. Redis does this differently.

### Steps to Reproduce

```php
assert($cacheManager instanceof CacheManager);

$cache = $cacheManager
    ->getCache('CACHE_WITH_PDO_BACKEND');
$backend = $cache
    ->getBackend();
assert($backend instanceof IterableBackendInterface);

$cache->set('key_before', 'value_before');
$dataBefore = \iterator_to_array($backend);

self::assertEquals(['key_before' => 'value_before'], $dataBefore);

$cache->flush();
$cache->set('key_after', 'value_after');
$dataAfter = \iterator_to_array($backend);

self::assertEquals(['key_after' => 'value_after'], $dataAfter);

/*
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    'key_after' => 'value_after'
+    'key_before' => 'value_before'
 )
 */

```

#### Expected behavior

If there's any user interaction with a certain cache, every change should be reflected in the iteration.
I do realize taking lifetime into account might be a bit too much to ask, but at least when I manually flush or add new values, the iterator should be set to null.

#### Actual behavior

The first time of doing any iteration interaction (method call to key, current, rewind, valid) with an iterable cache backend constitutes the value of iterable until the PHP process terminates.

#### To be discussed

What's the reason behind having the iterable state cached inside of the PdoBackend object?

I could e.g. change the $this->cacheEntriesIterator from ArrayAccess to Generator. 
Rewind could aways create a new one and the internal state of all database rows would be removed from the PHP runtime.

I'd even think about making the IterableBackendInterface not extend Iterator but Traversable to allow exposing the Generator directly, reducing the clutter of having all methods forced by Iterator just proxy to $this->cacheEntriesIterator. But I realize that's a breaking change and maybe not a good idea to be changed for such a detail.

kdambekalns:
Some cache backends hold an internal state when used as an iterator.
This state needs to be reset whenever the source data is updated.

Fixes #2828 

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/ApcuBackend.php
+++ b/Neos.Cache/Classes/Backend/ApcuBackend.php
@@ -58,7 +58,7 @@ class ApcuBackend extends IndependentAbstractBackend implements TaggableBackendI
     protected $identifierPrefix;
 
     /**
-     * @var \APCUIterator
+     * @var \APCUIterator|null
      */
     protected $cacheEntriesIterator;
 
@@ -111,7 +111,7 @@ class ApcuBackend extends IndependentAbstractBackend implements TaggableBackendI
      * @param string $entryIdentifier An identifier for this specific cache entry
      * @param string $data The data to be stored
      * @param array $tags Tags to associate with this cache entry
-     * @param integer $lifetime Lifetime of this cache entry in seconds. If NULL is specified, the default lifetime is used. "0" means unlimited lifetime.
+     * @param int|null $lifetime Lifetime of this cache entry in seconds. If NULL is specified, the default lifetime is used. "0" means unlimited lifetime.
      * @return void
      * @throws Exception if no cache frontend has been set.
      * @throws \InvalidArgumentException if the identifier is not valid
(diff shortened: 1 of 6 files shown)


kdambekalns:
@stephanschuler Feel free to have a look.

kdambekalns:
Oh, Psalm… 😵‍💫

fcool (approved):
Fine for me!

But may I ask, what work load brought this behavior to your attention?

kitsunet (approved):

kdambekalns:
CHanged base to 7.3, as it's not a security fix.

PR #2891, opened 2022-09-06
FEATURE: Skip unhealthy cache backends in MultiBackend

### LINKED ISSUE #2890: FEATURE: Skip unhealthy cache backends within the same request

### Description

The `MultiBackend` cache backend catches errors by the respective concrete backends and tries to execute the operation on the next backend. For example, if there are two backends – a RedisBackend and a FileSystemBackend – and the Redis server does not reply within the timeout threshold (let's say, 5 seconds), Flow will use the FileSystemBackend instead.

That's all fine in theory, however what can happen is, that every API call dealing with the cache will first try Redis. So, if within one request there are 25 cache operations, the whole request will take more than two minutes.

### Possible Solution

The MultiBackend should introduce the concept of a backend health status. That can (probably should) be limited to the scope of one CLI / web request. If a backend throws an error during one operation, the MultiBackend will mark it as unhealthy and not try using it again for the remainder of the request. 

If the unhealthy backend in question is the last backend to try, Flow should not try to catch any further errors and let the Throwable pass further up.

robertlemke:
This introduces a new feature for the MultiBackend and the TaggableMultiBackend which automatically removes unhealthy sub backends from the list of backends for the remainder of a request.

If a sub backend throws any error, the error will be caught, logged and the backend removed from the list of sub backends. If the sub backend causing trouble is the last one configured for the MultiBackend, it will not be removed.

The feature can be disabled by setting the option `removeUnhealthyBackends` to `false`.

Resolves #2890

### DIFF
--- a/Neos.Cache/Classes/Backend/MultiBackend.php
+++ b/Neos.Cache/Classes/Backend/MultiBackend.php
@@ -81,7 +81,7 @@ class MultiBackend extends AbstractBackend
             $backend = $this->instantiateBackend($backendClassName, $backendOptions, $this->environmentConfiguration);
             $backend->setCache($this->cache);
         } catch (Throwable $throwable) {
-            $this->logger && $this->throwableStorage && $this->logger->error(sprintf('Failed creating sub backend %s for %s: %s', $backendClassName, get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
+            $this->logger?->error('Failed creating sub backend ' . $backendClassName . ' for ' . get_class($this) . ': ' . $this->throwableStorage?->logThrowable($throwable), LogEnvironment::fromMethodName(__METHOD__));
             $this->handleError($throwable);
             $backend = null;
         }
@@ -98,7 +98,7 @@ class MultiBackend extends AbstractBackend
             try {
                 $backend->set($entryIdentifier, $data, $tags, $lifetime);
             } catch (Throwable $throwable) {
-                $this->logger && $this->throwableStorage && $this->logger->error(sprintf('Failed setting cache entry using backend %s in %s: %s', get_class($backend), get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
+                $this->logger?->error('Failed setting cache entry using backend ' . get_class($backend) . ' in ' . get_class($this) . ': ' . $this->throwableStorage?->logThrowable($throwable), LogEnvironment::fromMethodName(__METHOD__));
                 $this->handleError($throwable);
                 $this->removeUnhealthyBackend($backend);
                 if (!$this->setInAllBackends) {
@@ -118,7 +118,7 @@ class MultiBackend extends AbstractBackend
             try {
                 return $backend->get($entryIdentifier);
             } catch (Throwable $throwable) {
-                $this->logger && $this->throwableStorage && $this->logger->error(sprintf('Failed retrieving cache entry using backend %s in %s: %s', get_class($backend), get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
+                $this->logger?->error('Failed retrieving cache entry using backend ' . get_class($backend) . ' in ' . get_class($this) . ': ' . $this->throwableStorage?->logThrowable($throwable), LogEnvironment::fromMethodName(__METHOD__));
                 $this->handleError($throwable);
                 $this->removeUnhealthyBackend($backend);
             }
@@ -136,7 +136,7 @@ class MultiBackend extends AbstractBackend
             try {
                 return $backend->has($entryIdentifier);
             } catch (Throwable $throwable) {
-                $this->logger && $this->throwableStorage && $this->logger->error(sprintf('Failed checking if cache entry exists using backend %s in %s: %s', get_class($backend), get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
+                $this->logger?->error('Failed checking if cache entry exists using backend ' . get_class($backend) . ' in ' . get_class($this) . ': ' . $this->throwableStorage?->logThrowable($throwable), LogEnvironment::fromMethodName(__METHOD__));
                 $this->handleError($throwable);
                 $this->removeUnhealthyBackend($backend);
             }
@@ -155,7 +155,7 @@ class MultiBackend extends AbstractBackend
             try {
                 $result = $result || $backend->remove($entryIdentifier);
             } catch (Throwable $throwable) {
-                $this->logger && $this->throwableStorage && $this->logger->error(sprintf('Failed removing cache entry using backend %s in %s: %s', get_class($backend), get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
+                $this->logger?->error('Failed removing cache entry using backend ' . get_class($backend) . ' in ' . get_class($this) . ': ' . $this->throwableStorage?->logThrowable($throwable), LogEnvironment::fromMethodName(__METHOD__));
                 $this->handleError($throwable);
                 $this->removeUnhealthyBackend($backend);
             }
@@ -173,7 +173,7 @@ class MultiBackend extends AbstractBackend
             try {
                 $backend->flush();
             } catch (Throwable $throwable) {
-                $this->logger && $this->throwableStorage && $this->logger->error(sprintf('Failed flushing cache using backend %s in %s: %s', get_class($backend), get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
+                $this->logger?->error('Failed flushing cache using backend ' . get_class($backend) . ' in ' . get_class($this) . ': ' . $this->throwableStorage?->logThrowable($throwable), LogEnvironment::fromMethodName(__METHOD__));
                 $this->handleError($throwable);
                 $this->removeUnhealthyBackend($backend);
             }
@@ -190,7 +190,7 @@ class MultiBackend extends AbstractBackend
             try {
                 $backend->collectGarbage();
             } catch (Throwable $throwable) {
-                $this->logger && $this->throwableStorage && $this->logger->error(sprintf('Failed collecting garbage using cache backend %s in %s: %s', get_class($backend), get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
+                $this->logger?->error('Failed collecting garbage using cache backend ' . get_class($backend) . ' in ' . get_class($this) . ': ' . $this->throwableStorage?->logThrowable($throwable), LogEnvironment::fromMethodName(__METHOD__));
                 $this->handleError($throwable);
                 $this->removeUnhealthyBackend($backend);
             }
@@ -255,7 +255,7 @@ class MultiBackend extends AbstractBackend
         $i = array_search($unhealthyBackend, $this->backends, true);
         if ($i !== false) {
             unset($this->backends[$i]);
-            $this->logger && $this->throwableStorage && $this->logger->warning(sprintf('Removing unhealthy cache backend %s from backends used by %s', get_class($unhealthyBackend), get_class($this)), LogEnvironment::fromMethodName(__METHOD__));
+            $this->logger?->warning(sprintf('Removing unhealthy cache backend %s from backends used by %s', get_class($unhealthyBackend), get_class($this)), LogEnvironment::fromMethodName(__METHOD__));
         }
     }
 }
--- a/Neos.Cache/Classes/Backend/MultiBackend.php
+++ b/Neos.Cache/Classes/Backend/MultiBackend.php
@@ -14,58 +14,24 @@ namespace Neos\Cache\Backend;
  */
 
 use Neos\Cache\BackendInstantiationTrait;
-use Neos\Cache\EnvironmentConfiguration;
+use Neos\Cache\Exception as CacheException;
+use Throwable;
 
 /**
  * A multi backend, falling back to multiple backends if errors occur.
  */
-class MultiBackend extends AbstractBackend implements BackendInterface
+class MultiBackend extends AbstractBackend
 {
     use BackendInstantiationTrait;
 
-    /**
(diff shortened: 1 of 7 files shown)


Neos.Cache/Classes/Backend/TaggableMultiBackend.php
```diff
@@ -59,19 +42,17 @@ public function flushByTag(string $tag): int
         foreach ($this->backends as $backend) {
             try {
                 $count |= $backend->flushByTag($tag);
-            } catch (\Throwable $t) {
-                $this->handleError($t);
+            } catch (Throwable $throwable) {
+                $this->logger && $this->logger->error(sprintf('Failed flushing cache by tag using backend %s in %s: %s', get_class($backend), get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
```
kitsunet, line 46: Logging flag as for the rest

Neos.Cache/Classes/Backend/TaggableMultiBackend.php
```diff
@@ -93,7 +74,10 @@ public function findIdentifiersByTag(string $tag): array
         foreach ($this->backends as $backend) {
             try {
                 $identifiers[] = $backend->findIdentifiersByTag($tag);
-            } catch (\Throwable $t) {
+            } catch (Throwable $throwable) {
+                $this->logger && $this->logger->error(sprintf('Failed finding identifiers by tag using backend %s in %s: %s', get_class($backend), get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
```
kitsunet, line 78: Logging flag as for the rest

kitsunet (commented):
Looks good, some nice cleanups. I would prefer if logging was a configuration flag, given that the idea is to cover short interruptions of primary cache backends it should not lead to a log flooding (and potentail performance hit due to it).

Neos.Cache/Classes/Backend/MultiBackend.php
```diff
@@ -125,190 +105,147 @@ public function set(string $entryIdentifier, string $data, array $tags = [], int
...
     {
-        if ($this->debug) {
-            throw $throwable;
+        if ($this->removeUnhealthyBackends === false || count($this->backends) <= 1) {
+            return;
+        }
+        foreach ($this->backends as $i => $backend) {
+            if ($backend === $unhealthyBackend) {
+                unset($this->backends[$i]);
+                $this->logger && $this->logger->warning(sprintf('Removing unhealthy cache backend %s from backends used by %s', get_class($unhealthyBackend), get_class($this)), LogEnvironment::fromMethodName(__METHOD__));
```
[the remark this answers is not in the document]
robertlemke: > Also what happens if no backend is left?

A backend is only removed, if more than 1 backend is left, so as far as I can see, it won't happen that no backend is left.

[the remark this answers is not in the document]
kitsunet: Right, didn't see the condition before the loop. Good! THanks!

Neos.Cache/Classes/Backend/MultiBackend.php
```diff
@@ -83,40 +69,34 @@ protected function prepareBackends(): void
...
     protected function buildSubBackend(string $backendClassName, array $backendOptions): ?BackendInterface
     {
-        $backend = null;
         try {
             $backend = $this->instantiateBackend($backendClassName, $backendOptions, $this->environmentConfiguration);
             $backend->setCache($this->cache);
-        } catch (\Throwable $t) {
-            $this->handleError($t);
+        } catch (Throwable $throwable) {
+            $this->logger && $this->logger->error(sprintf('Failed creating sub backend %s for %s: %s', $backendClassName, get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
```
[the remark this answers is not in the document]
robertlemke: > This should be configurable I think, as writing these logs might be a massive performance hit if a backend is down.

As far as I understand it, only wrong configuration or code errors could lead to an error here. When the sub backend is *created*, there shouldn't be any access to the actual infrastructure (such as MySQL or Redis server) – or?

robertlemke:
Thanks for the review! I'll prepare another commit which makes logging configurable.

Neos.Cache/Classes/Backend/MultiBackend.php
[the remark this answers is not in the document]
kitsunet: Mmm, I guess, maybe the rare case that short redis outage happens just as the backend is created but fine to keep it permanently on for the creation :)

robertlemke:
Error logging is now configurable. However, I turned it on by default – would that be acceptable?

My main scenario is one of a high-traffic website with occasional, but rare downtimes of a Redis Sentinel cluster. Certainly, there will be a lot of log messages when that happens. But the log pipeline should be able to cope with that, and I'd rather notice that there's a problem than having a clean log.

I'd say, it's fine to have this enabled in *most* cases.

kitsunet:
Perfectly fine by me, just wanted to have the possibility to disable it :)

Neos.Cache/Classes/Backend/MultiBackend.php
```diff
@@ -83,40 +74,34 @@ protected function prepareBackends(): void
...
     protected function buildSubBackend(string $backendClassName, array $backendOptions): ?BackendInterface
     {
-        $backend = null;
         try {
             $backend = $this->instantiateBackend($backendClassName, $backendOptions, $this->environmentConfiguration);
             $backend->setCache($this->cache);
-        } catch (\Throwable $t) {
-            $this->handleError($t);
+        } catch (Throwable $throwable) {
+            $this->logger && $this->throwableStorage && $this->logger->error(sprintf('Failed creating sub backend %s for %s: %s', $backendClassName, get_class($this), $this->throwableStorage->logThrowable($throwable)), LogEnvironment::fromMethodName(__METHOD__));
```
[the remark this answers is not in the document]
bwaidelich: Or, since this is used in many places, maybe extract it to a private method!?

Neos.Cache/Classes/Backend/MultiBackend.php
```diff
@@ -125,190 +110,155 @@ public function set(string $entryIdentifier, string $data, array $tags = [], int
...
+        if ($this->removeUnhealthyBackends === false || count($this->backends) <= 1) {
+            return;
+        }
+        foreach ($this->backends as $i => $backend) {
+            if ($backend === $unhealthyBackend) {
+                unset($this->backends[$i]);
+                $this->logger && $this->throwableStorage && $this->logger->warning(sprintf('Removing unhealthy cache backend %s from backends used by %s', get_class($unhealthyBackend), get_class($this)), LogEnvironment::fromMethodName(__METHOD__));
+                return;
+            }
         }
```
kdambekalns, line 262: Instead of a loop, `array_search(…)` could be used, it's even faster: https://3v4l.org/ubsXd

Neos.Cache/Classes/Backend/MultiBackend.php
```diff
@@ -15,57 +15,48 @@
 
 use Neos\Cache\BackendInstantiationTrait;
 use Neos\Cache\EnvironmentConfiguration;
+use Neos\Flow\Core\Bootstrap;
+use Neos\Flow\Log\ThrowableStorageInterface;
+use Neos\Flow\Log\Utility\LogEnvironment;
+use Neos\Flow\ObjectManagement\ObjectManagerInterface;
+use Psr\Log\LoggerInterface;
+use Throwable;
```
kdambekalns, line 23: IMHO importing from the global namespace is confusing, but whatever. 😇

bwaidelich:
I added some nitpick comments regarding code style, but apart from that I wonder if we should not just inject the LoggerBackend and allow that to be a `NullBackend` if needed.
This could be configured via `Objects.yaml` and with "virtual objects" you could even use different logger settings for different `MultiBackend` instances.

Not easier to use, but it would be a common pattern

Neos.Cache/Classes/Backend/MultiBackend.php
```diff
@@ -15,57 +15,48 @@
...
-     * @param EnvironmentConfiguration $environmentConfiguration
-     * @param array $options Configuration options - depends on the actual backend
-     */
-    public function __construct(EnvironmentConfiguration $environmentConfiguration, array $options)
+    public function __construct(EnvironmentConfiguration $environmentConfiguration = null, array $options = [])
     {
         parent::__construct($environmentConfiguration, $options);
+
+        /** @noinspection ClassConstantCanBeUsedInspection */
+        if ($this->logErrors && class_exists('Neos\Flow\Core\Bootstrap') && Bootstrap::$staticObjectManager instanceof ObjectManagerInterface) {
```
kdambekalns, line 48: Why not use the `::class` constant here? Any special reason?
```suggestion
        if ($this->logErrors && class_exists(Bootstrap::class) && Bootstrap::$staticObjectManager instanceof ObjectManagerInterface) {
```

Neos.Cache/Classes/Backend/TaggableMultiBackend.php
```diff
@@ -59,19 +42,19 @@ public function flushByTag(string $tag): int
         foreach ($this->backends as $backend) {
             try {
                 $count |= $backend->flushByTag($tag);
```
kdambekalns, line 44: Uh, this is wrong, I broke that back in the days… I'll fix it in this PR, ok?

kdambekalns:
> I added some nitpick comments regarding code style, but apart from that I wonder if we should not just inject the LoggerBackend and allow that to be a `NullBackend` if needed.

The current way is cheaper, if no logging is needed, as that would mean building the log message just to throw it away.

And we don't have a `NullThrowableStorage`, so you'd get traces without log messages.

kdambekalns:
> I added some nitpick comments regarding code style

I can't seem to find those!? 😳

Neos.Cache/Classes/Backend/TaggableMultiBackend.php
kdambekalns: No, needs to be fixed in an earlier branch…

robertlemke:
> I added some nitpick comments regarding code style, but apart from that I wonder if we should not just inject the LoggerBackend and allow that to be a `NullBackend` if needed. This could be configured via `Objects.yaml` and with "virtual objects" you could even use different logger settings for different `MultiBackend` instances.

I agree with @kdambekalns here – rather use `null` instead of a `NullBackend` in order to save rendering time for log messages which are never logged.

Neos.Cache/Classes/Backend/MultiBackend.php
robertlemke: Thanks, I fixed that.

Neos.Cache/Classes/Backend/TaggableMultiBackend.php
kdambekalns: https://github.com/neos/flow-development-collection/pull/2893

Neos.Cache/Classes/Backend/MultiBackend.php
robertlemke: I got used to it – otherwise PhpStorm subtly complains, so I obey …

Neos.Cache/Classes/Backend/MultiBackend.php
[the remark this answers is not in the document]
kdambekalns: Yep, what I asked earlier. :)

robertlemke: I always forget that `::class` is even allowed for classes which don't exist (which is somewhat confusing). Fixed that.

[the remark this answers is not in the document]
robertlemke: See comment for Karsten's nitpick.

Neos.Cache/Classes/Backend/MultiBackend.php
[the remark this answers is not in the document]
robertlemke: I used `$this->logger && $this->logger->info()` all the time in my code, I tend to like it.

In this case, I added `&& $this->throwableStorage` only because Psalm complained. Which is only theoretical, because there can't be a logger if there's no Throwable Storage as per the constructor code.

But yeah, I'll put it into a method.

Neos.Cache/Classes/Backend/MultiBackend.php
kdambekalns: Oh, I disabled that for the global namespace… 😉

Neos.Cache/Classes/Backend/MultiBackend.php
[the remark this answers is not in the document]
mhsdesign: just `$this->logger?->info()` with php 8 ^^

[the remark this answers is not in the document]
robertlemke: But thinking about it … if we put that into a private method, how would that look like? If that method logs, we loose the context information (LogEnvironment). And it's not always "error", it's also "warn". And sometimes it's `sprintf()`ed.

It's really just a Psalm thing.

robertlemke:
Unless you have concerns, I'd like to leave it as it is at this point.

kitsunet (approved):

Neos.Cache/Classes/Backend/MultiBackend.php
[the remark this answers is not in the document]
robertlemke: I'll try that.

robertlemke:
I pushed a more concise version regarding the logger calls. Let's see if Psalm is happy.

robertlemke:
thanks @kdambekalns !

kdambekalns:
> thanks @kdambekalns !

Can't let Psalm ruin a day… 🙃

mficzel:
Had some side effects while backporting this to Neos 7.3 running on PHP 8. I leave this comment her for anyone hitting this aswell.

The error "Search results for “Cannot build object Psr\Log\LoggerInterface because it is unknown to the compile time Object Manager.” Was thrown after flushing the caches until i added `class_exists(LoggerInterface::class)` to the condition in the constructor. 

For now i assume that this is Neos 7.3 specific.

kitsunet:
@mficzel it's not, there was a fix to this after the fact:
https://github.com/neos/flow-development-collection/commit/5d7fcd637b7fd29c050e1e05c934db6e6016b0df

mficzel:
@kitsunet thanks a lot

PR #2898, opened 2022-09-08
BUGFIX: Fix logging in multibackend during compile time

robertlemke:
Added an additional safeguard in MultiBackend to make sure that logging is disabled when no logger can be created yet.

During compile time, `LoggerInterface` might not be available yet. However, the MultiBackend may already be invoked at that point and try to log.

**Checklist**


kdambekalns (approved):
That should do the job.

Eventually adding a way to inject the logger manually (via `injectLogger(…)`) would be nice, now that I think about it.

kdambekalns:
> Eventually adding a way to inject the logger manually (via `injectLogger(…)`) would be nice, now that I think about it.

Then again, that would need at least yet another check for `LogEnvironment`. _sigh_

kitsunet (approved):
I have some ideas for future refactoring but for now this should do the job :)

PR #2911, opened 2022-10-01
FEATURE: Introduce `session:collectgarbage` command and move the garbage collection from session to sessionManager

mficzel:
The `session:collectgarbage` command will remove session-data and -metadata of outdated sessions identified by lastActivityTimestamp being older than inactivityTimeout. This is usually done automatically after shutdown for the percentage of requests specified in the setting `Neos.Flow.session.garbageCollection.probability`. Use this command only if you need more direct control over the cleanup intervals.

In addition the `collectGarbage` method is moved from the session to the sessionManager. As this was api before a backwards compatible implementation was left behind that can be removed with Flow 9.

**Review instructions**

The main purpose of the pr is moving the garbage collection to the session manager in preparation of further adjustments in session handling where it will likely would become problematic that the session is responsible for removing other outdated sessions.

**Checklist**


sorenmalling:
@mficzel will you be so awesome and add a reference to a issue?

mficzel:
@sorenmalling this is a prerequisite of finally going trying to solve https://github.com/neos/flow-development-collection/issues/525. There is no direct issue for the pr here.

kitsunet:
+1 by reading, need to test this still.

kitsunet (approved):

Neos.Flow/Classes/Session/Session.php
```diff
@@ -621,50 +616,16 @@ public function destroy($reason = null)
...
-            }
-            if ($sessionRemovalCount >= $this->garbageCollectionMaximumPerRun) {
-                break;
-            }
-        }
-
-        $this->metaDataCache->remove('_garbage-collection-running');
-        return $sessionRemovalCount;
+        $result = $this->sessionManager->collectGarbage();
+        return is_null($result) ? 0 : $result;
```
mhsdesign, line 628: i have the feeling we should return `false` here @mficzel ? Otherwise the return type seems to change?

Issue #2913, opened 2022-10-04
Modernize Reflection Service


robertlemke:
Modernize and adjust the Reflection Service based on PHP 8.1 with full support for analyzing PHP code based on 8.0, 8.1 or 8.2.

mhsdesign:
Can this be closed? The reflection service just got another overhaul via https://github.com/neos/flow-development-collection/pull/3443

PR #2914, opened 2022-10-04
!!! TASK: Modernized code style in ReflectionService

robertlemke:
Code in the reflection service was adjusted to the current code style best practices. 

The method arguments in the Reflection Service are now strictly typed. Therefore, third-party code which relied on loose types and passes invalid types, need to be adjusted. Tests in the Flow package were adjusted were necessary.

As part of the clean up, the setStatusCache() method in ReflectionService was fixed which used a wrong order of parameters in its is_callable() call.

Preparation for #2913

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 namespace Neos\Flow\Reflection;
 
 /*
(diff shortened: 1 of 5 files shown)


albe:
Can we just merge this as is (after fixing the builds) to keep the following functional changes to the ReflectionService away from the noise of these (useful! 👏) syntax changes? :)

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -13,27 +14,33 @@
 
 use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\PhpParser;
+use Doctrine\Common\Annotations\Reader;
 use Doctrine\ORM\Mapping as ORM;
 use Doctrine\Persistence\Proxy as DoctrineProxy;
-use Neos\Flow\Annotations as Flow;
-use Neos\Cache\Frontend\FrontendInterface;
+use InvalidArgumentException;
```
mhsdesign: i think you configured your ide like this...

i dont use it myself and it is an anomaly to the rest of neos/flow
just wanted to start the diskussion.

do we want this for new code?

mhsdesign:
i agree with @albe this syntax and type adjustment is already a lot to review ... if its non breaking it can go directly to 8.2

Neos.Flow/Classes/Reflection/ReflectionService.php
robertlemke: Hmm, I don't know … I like it and do that for a couple of years now in my own code. It's the easiest rule: Import if possible.

kdambekalns:
> if its non breaking it can go directly to 8.2

It could, but keeping the non-breaking tweaks seperated from the (potentially) breaking changes is adding a lot of overhead during development. And eventually something will break. So let's stick to the 9.0 branch for this!

Neos.Flow/Classes/Reflection/ReflectionService.php
kdambekalns: Good point. I dislike this and exclude the global namespace from being imported in PhpStorm.

For me imports like these do not add anything useful.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -627,10 +510,10 @@ public function isClassAbstract($className)
      * Tells if the specified class is final or not
      *
      * @param string $className Name of the class to analyze
-     * @return boolean true if the class is final, otherwise false
+     * @return bool true if the class is final, otherwise false
```
kdambekalns, line 512: This could be left out, as the comment does not add information that's not already in the name…

kdambekalns (commented):
Maybe take the "global namespace import" issue to the team slack for a quick vote?

Neos.Flow/Classes/Reflection/ReflectionService.php
robertlemke: > This could be left out, as the comment does not add information that's not already in the name…

well, yes … but I guess the ReflectionService will change a lot anyway during the next days …

Neos.Flow/Classes/Reflection/ReflectionService.php
kdambekalns: To me this usually adds a feeling of "oh, something custom here…" – as in, not built-in into PHP itself. 🤷‍♂️

But the "one rule to rule them all" is a point… I'd follow suit if we decide to do it.

robertlemke:
Sorry for the force-pushes … The state should now be fine, I removed the global import of `\InvalidArgumentException` as the general consensus in the team is to rather not import such classes.

Neos.Flow/Classes/Reflection/ReflectionService.php
mhsdesign: slack has your back @kdambekalns https://neos-project.slack.com/archives/C04Q0TC15/p1683558099679779 😎

kdambekalns (approved):

kdambekalns:
Regarding the Psalm failures, the one about Fluid might need serious checking, that could be true.

The others _might_ be false positives, but maybe when the refactoring has reached "the next level" we could tackle what is left…

robertlemke:
Thanks for your reviews!

PR #2917, opened 2022-10-07
TASK: use PHPs hash_pbkdf2

### LINKED ISSUE #2916: FEATURE: Use PHPs hash_pbkdf2

### Description

Currently the flow framework implements it's own calculation logic for generating `PBKDF2` hashes.

In `\Neos\Flow\Security\Cryptography\Algorithms` there is even a comment, that it should be replaced.

https://github.com/neos/flow-development-collection/blob/34ef4625b86f7888ef70dd71cfc78ba128f5ab0b/Neos.Flow/Classes/Security/Cryptography/Algorithms.php#L28-L29

### Possible Solution

As of PHP >= 5.5 [`hash_pbkdf2`](https://www.php.net/manual/de/function.hash-pbkdf2.php) was implemented. 
We can simply replace the current generation logic with this function.

IMO `\Neos\Flow\Security\Cryptography\Algorithms` can then be marked as `@deprecated`.

reflexxion:
Use `hash_pbkdf2` provided by PHP (since 5.5) instead of own generation logic.

Fixes #2916

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Security/Cryptography/Algorithms.php
+++ b/Neos.Flow/Classes/Security/Cryptography/Algorithms.php
@@ -16,17 +16,12 @@ namespace Neos\Flow\Security\Cryptography;
  *
  * Right now this class provides a PHP based PBKDF2 implementation.
  *
+ * @deprecated since 8.2, use PHPs `hash_pbkdf2`
  */
 class Algorithms
 {
     /**
      * Compute a derived key from a password based on PBKDF2
-     *
-     * See PKCS #5 v2.0 http://tools.ietf.org/html/rfc2898 for implementation details.
-     * The implementation is tested with test vectors from http://tools.ietf.org/html/rfc6070 .
-     *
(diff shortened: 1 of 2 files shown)


kdambekalns (approved):
Awesome trip down history lane, thanks!

Benjamin-K:
Shouldn't this be merged against 6.3?

reflexxion:
> Shouldn't this be merged against 6.3?

Though it isn't a bug I targeted 8.2.
I'm open to other decisions.

Benjamin-K:
You're right. Nevermind. If we target an older release, then it should be 7.3

reflexxion:
Who would decide that?

kdambekalns:
I'm fine with 8.2 - it doesn't fix a bug, so no older version, and it's not breaking, so next minor. All fine.

albe (approved):
Awesome!

PR #2919, opened 2022-10-09
FEATURE: Introduce `package://` stream wrapper to access files outside of resources

mficzel:
Packages often rely on files from other packages and in some cases those might not be resources. Typical use cases are NodeTypes, Tests, YAML or other data files. Similar to the resources it not advisible to build pathes using the ``FLOW_PATH_*`` constants or include files via relative pathes from your package. The package stream wrapper offers a more convenient and safe way to achieve this.

The following example reads the content of the file
``Acme.Demo/NodeTypes/Document.yaml`` into a variable:

*Example: Accessing package resources* ::

	$template = file_get_contents(
		'package://Acme.Demo/NodeTypes/Document.yaml'
	);

You are encouraged to use this stream wrapper wherever you need to access a a file from another
package that is not located in the package resources.
 
**Review instructions**

The feature set is pretty much identical to the resource:// stream wrapper with the following deviations:
- Traversal upwards via `..` leads to an invalid argument exception  (should be added to resources stream wrapper later aswell)
- The base path is now the package path and not the resource path allowing to access all files inside that package

**Checklist**


sorenmalling:
Will you be so kind and update the resource management documentation part and add a paragraph about the new stream wrapper: Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/ResourceManagement.rst

mficzel:
BTW: I tried to extract a common base class from the resource and package stream wrapper but this ended in a big mess because of many methods that are not returning the annotated type.

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
```diff
@@ -0,0 +1,124 @@
...
+        $openedPathAndFilename = '';
+        $this->packageStreamWrapper->open('invalid-scheme://foo/bar', 'r', 0, $openedPathAndFilename);
+    }
+
+    public function providePathesWithForbiddenUpwardsTraversel(): array
+    {
+        return [
+            ['package://Some.Package/../bar'],
+            ['package://Some.Package/bar/..'],
+            ['package://Some.Package/foo/../bar']
```
mhsdesign: this may be allowed no?

mhsdesign:
For the docs:

Since the goal of flow is to have only one way of doing things, we should prepare ourselves for an argument about this: 

`resource://My.Package/Private/Foo.html` vs `package://My.Package/Resource/Private/Foo.html`

Or do you see this less critical. For example, it could happen that new users start using `package` for everything - would that be okay?

(btw im not at all against the package feature ^^)

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
mficzel: Not sure, should it? Would probably make the adjustment in the fusionCache in the other pr obsolete

mficzel:
@mhsdesign i think it is no problem that resource:// slightly overlaps with package:// as both can do specific things the other cannot. Would still recommend resource for all use cases inside the Ressource folder

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
mficzel: Adjusted to be tolerant to .. at the end. Really helps to allow iteration over those such directories

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
```diff
@@ -0,0 +1,128 @@
...
+        $this->packageStreamWrapper->open('invalid-scheme://foo/bar', 'r', 0, $openedPathAndFilename);
+    }
+
+    public function providePathesWithForbiddenUpwardsTraversel(): array
+    {
+        return [
+            ['package://Some.Package/../bar', true],
+            ['package://Some.Package/foo/../bar', true],
+            ['package://Some.Package/bar/..', false],
+            ['package://Some.Package/..', false],
```
bwaidelich: Shouldn't that be an exceptional case?

bwaidelich (commented):
Awesome, even with tests ❤️

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
```diff
@@ -0,0 +1,128 @@
...
+     * @test
+     */
+    public function openThrowsExceptionForInvalidScheme()
+    {
+        $this->expectException(\InvalidArgumentException::class);
+        $openedPathAndFilename = '';
+        $this->packageStreamWrapper->open('invalid-scheme://foo/bar', 'r', 0, $openedPathAndFilename);
+    }
+
+    public function providePathesWithForbiddenUpwardsTraversel(): array
```
bwaidelich:
```suggestion
    public function providePathesWithForbiddenUpwardsTraversal(): array
```

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
mhsdesign: thanks ;)

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
mficzel: No because this happens while iterating over a folder. Since one cannot only traverse up but not down again to another file/folder this should be safe

bwaidelich: What about `package://Foo.Bar/../../../`?

mficzel: Is forbidden because of upwards traversal in the middle of the path ... added example to test to clarify that

Neos.Flow/Classes/ResourceManagement/Streams/PackageStreamWrapper.php
```diff
@@ -0,0 +1,526 @@
...
+        $requestPathParts = explode('://', $requestedPath, 2);
+        if ($requestPathParts[0] !== self::SCHEME) {
+            throw new \InvalidArgumentException('The ' . __CLASS__ . ' only supports the \'' . self::SCHEME . '\' scheme.', 1256052544);
+        }
+
+        if (!isset($requestPathParts[1])) {
+            return false;
+        }
+
+        if (preg_match('/(^|\/)\\.\\.\//us', $requestPathParts[1])) {
```
mhsdesign: What do you think about resolving the $resourceUri path with realpath and check if it is inside of the current package. That could save us this regex (which confuses me a bit ^^) ;)

On the other hand it is all tested so fine either way - thanks ;)

mhsdesign: but using realpath woud be an unnecessary i/o maybe we can use a fake version of realpath https://www.php.net/manual/de/function.realpath.php
that operates on the package:// path to see if it traverses outward

... or i start understanding the regex ;)

bwaidelich: At https://stackoverflow.com/questions/611520/preventing-directory-traversal-with-web-facing-application-are-regular-express someone wrote

>  using a Regex based protection for such a scenario is like wearing a T-shirt against a hurricane

While I wouldn't put it that harsh, I agree that using `realpath` is probably a good idea ;)

mficzel: How would you compare that? Realpath will just convert `..` without complaining

mhsdesign: i would check if the realpath path starts with $package->getPackagePath() if not - someone used dark magic to traverse outwards

-> given that $package->getPackagePath() returns a realpath itself

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
kdambekalns: But `['package://Some.Package/..', false]` still looks wrong to me. It traverses outside the package!

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/ResourceManagement.rst
```diff
@@ -449,6 +449,27 @@ this. To use this feature, just pass the resource hash:
...
+======================
+
+Packages often rely on files from other packages and in some cases those might not be resources.
+Typical use cases are NodeTypes, Tests, YAML or other data files.
+Similar to the resources it not advisible to build pathes using the ``FLOW_PATH_*`` constants or
+include files via relative pathes from your package. The package stream wrapper offers a more
+convenient and safe way to achieve this.
+
+The following example reads the content of the file
+``Acme.Demo/NodeTypes/Document.yaml`` into a variable:
```
kdambekalns, line 462: To not confuse anyone, that should refer to `Acme.Demo/Configuration/NodeTypes/Document.yaml`, I'd say.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/ResourceManagement.rst
```diff
@@ -449,6 +449,27 @@ this. To use this feature, just pass the resource hash:
...
+
+*Example: Accessing package resources* ::
+
+	$template = file_get_contents(
+		'package://Acme.Demo/NodeTypes/Document.yaml'
```
kdambekalns, line 467: … and here, too.

Neos.Flow/Tests/Unit/ResourceManagement/Streams/PackageStreamWrapperTest.php
```diff
@@ -0,0 +1,134 @@
...
+        $mockPackage = $this->createMock(FlowPackageInterface::class);
+        $mockPackage->expects(self::any())->method('getPackagePath')->will(self::returnValue('vfs://Foo'));
+        $this->mockPackageManager->expects(self::once())->method('getPackage')->with($packageKey)->will(self::returnValue($mockPackage));
+
+        $openedPathAndFilename = '';
+        self::assertTrue($this->packageStreamWrapper->open('package://' . $packageKey . '/Some/Path', 'r', 0, $openedPathAndFilename));
+        self::assertSame($openedPathAndFilename, 'vfs://Foo/Some/Path');
+    }
+
+    /**
```
kdambekalns, line 122:
```suggestion
    /**
     * This makes sure the code does not see a 40-charatcer package key as a resource hash.
```

Neos.Flow/Classes/ResourceManagement/Streams/PackageStreamWrapper.php
kdambekalns: +100 for checking with `realpath` here! Security over performance, definitely. If it makes a difference at all, given that regex handling isn't the fastest thing on earth anyway.

kdambekalns (changes requested):

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/ResourceManagement.rst
mhsdesign: Isnt `NodeTypes` a new top level folder?

Neos.Flow/Classes/ResourceManagement/Streams/PackageStreamWrapper.php
mhsdesign: we can still have a "fast" path when no `..` are included in the string ^^

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/ResourceManagement.rst
kdambekalns: Would be news to me. 🗞

bwaidelich: Yes, that's correct since 7.2: https://www.neos.io/blog/neos-flow-72-released.html#NodeTypes-in-Subfolders

mhsdesign: hmm i was wrong - not in flow context ...

flow doesnt care about NodeTypes

bwaidelich: @mhsdesign no, you were correct and so is the snippet

mhsdesign: yes but this is a flow documentation - and flow is not about neos...

kdambekalns: Uh, I knew about subfolders, but never realized this is package-top-level. TBH I think that's ugly AF. Suddenly something that is configuration (YAML, read using our configuration management, …) is no longer inside `Configuration`. *shudder* Well, too late…

But still, maybe the example could be less Neos-y. 🤷‍♂️

bwaidelich: > yes but this is a flow documentation

Ah right, good point :)

> I think that's ugly AF.

Not related to this change, but for the record: There are reasons for it. One of it being that it is not a (context specific) configuration and that we wanted to make NodeType and Integration Fusion co-locatable (i.e. put it in the same folder and don't require some `NodeType.*` prefix). See the full discussion at https://github.com/neos/neos-development-collection/issues/3000

kitsunet (changes requested):
See my comment on https://github.com/neos/neos-development-collection/pull/3903

mficzel:
@kitsunet we could define which folders are allowed for accessing via package:// ... on the other hand limiting the access to certain folders seems a bit random to me and encourage working around the framework. In the end folders other than Resources/Public are not that special and more like a convention. 

I actually played with a nodetypes:// streamWrapper and did not like it for it felt way to specific to be of any other use.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/ResourceManagement.rst
mficzel: Actually even if flow is not about Neos it is the most common use-case. That is why i used this as example.

mficzel:
@mhsdesign @kdambekalns problem with realpath is that it only works on existing pathes. Since in the stream wrapper $checkForExistence is an optional argument we imho cannot use this. Playing around implementing a solution myself in the Files-Utility but better ideas are welcome.

kdambekalns:
> Since in the stream wrapper $checkForExistence is an optional argument we imho cannot use this.

What is this thing? 🤯 I need to look at that flag, wasn't aware it exists and wonder what the use case is… 🤔

mhsdesign:
well done @mficzel if i read that correctly you implemented some kind of "fake" realpath? https://github.com/neos/flow-development-collection/pull/2919#discussion_r990831952

mficzel:
@mhsdesign @kdambekalns added a custom method `pathRespectsTraversalBoundaries` ... not sure yet.

mficzel:
@mhsdesign it is not actually a fake realpath as it does not care about symlinks ... it just detects traversal out of the scope

mhsdesign:
> 

it does make sense ... a new dir or file doesnt exist yet but the path needs to be still evaluated to create a file then...

kitsunet:
I am torn, I see the utility and the use case of course, but I also feel this is... not good. Gives to much random power to people to mess with package files.

As a core developer I want this for myself and for making my life easier, but forsupport I don't think this is a good idea, but I have to gather my thoughts, and provide some good arguments.

mficzel:
I do not consider stream wrappers a security feature. I mean who gets access to them has found a way to access the file system in the first place. It is merely an abstraction to access files without messing with package-pathes plus the Resources/Public magic.

Neos.Flow/Classes/ResourceManagement/Streams/PackageStreamWrapper.php
```diff
@@ -0,0 +1,547 @@
...
+        if ($checkForExistence === false || file_exists($resourceUri)) {
+            return $resourceUri;
+        }
+
+        return false;
+    }
+
+    private function pathRespectsTraversalBoundaries(string $path): bool
+    {
+        $path = Files::getUnixStylePath($path);
```
mhsdesign, line 529:
```suggestion
        $path = Files::getUnixStylePath($path);
        if (str_contains($path, '/../') === false) {
            return true;
        }
```

Shortcut -> as using `..` is more like the exception.

kitsunet:
> I do not consider stream wrappers a security feature. I mean who gets access to them has found a way to access the file system in the first place. It is merely an abstraction to access files without messing with package-pathes plus the Resources/Public magic.

My point is not in security, but in the system / us defining what is sensible to do and what is not. Currently it's obvious that accessing files in packages is not sensible, it's a lot of work and you should notice that you are working around the framework there. Adding this suddenly makes it an "approved" thing to do and I am not happy about that.

We are generalising this broad functionality to allow for a very specific usecase that even in documenting is refers to Neos. I think this is too much generalisation for the sake of it. Potentially giving us headaches in the long run.

Neos.Flow/Classes/ResourceManagement/Streams/PackageStreamWrapper.php
kdambekalns: Would a path ending with `/..` be rejected elsewhere?

mficzel:
I would agree that for things in Resources resource:// is preferrable but do not see a point in making access to other files harder than needed in the end it is a dependency between packages which we have in many other places aswell.

Also as long as there is the `NodeTypes` folder colocating the config in there with fusion makes sense and should be supported in a non hacky way. Will give a seperate `nodetypes://` stream wrapper another shot this evening.

kitsunet:
Thank you so much for trying that (again) I really think I would prefer it that way, sorry for jumping in so late. I would also disallow any write operations.

mhsdesign:
i would prefer: https://github.com/neos/neos-development-collection/pull/3903#issuecomment-1274597011

and im willing to implement it ^^

mficzel:
> i would prefer: https://github.com/neos/neos-development-collection/pull/3903#issuecomment-1274597011
> and im willing to implement it ^^

@mhsdesign ok ... then let us try that approach first ... at least that would not open another api. I will wait until this is tested before continuing or closing this pr here

mficzel:
Close in favor of a `nodetypes://` stream wrapper on the neos side

mhsdesign:
Thanks for your effort!

Issue #2922, opened 2022-10-12
BUG: CachePools don't support PSR-6 compliant identifier


bwaidelich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Creating an instance of `\Neos\Cache\Psr\Cache\CachePool` with a key containing a dot leads to an exception:

> "foo.bar" is not a valid cache identifier.

even though dots should be allowed according to https://www.php-fig.org/psr/psr-6/#definitions

### Expected Behavior

No exception should be thrown

### Steps To Reproduce

```php
new CachePool('foo.bar', new NullBackend());
```

### Environment

```markdown
- Flow: 5.0 - 8.1
```

PR #2923, opened 2022-10-12
BUGFIX: Relax CachePool key check

bwaidelich:
Adjusts the `CachePool` regex that checks the key (aka entry identifier) such that it allows "." as character.

*Note:* According to https://www.php-fig.org/psr/psr-6/#definitions the regex should be changed to `/^[a-zA-Z0-9_\.]{1,64}$/` (like done in the [SimpleCache implementation](https://github.com/neos/flow-development-collection/blob/d11ff78a9e419c5b215d6e951e4fb9eed63e4ffa/Neos.Cache/Classes/Psr/SimpleCache/SimpleCache.php#L30) – But this is out of scope of this bugfix since it would be a breaking change

Fixes: #2922

### DIFF
--- a/Neos.Cache/Classes/Psr/Cache/CachePool.php
+++ b/Neos.Cache/Classes/Psr/Cache/CachePool.php
@@ -27,7 +27,7 @@ class CachePool implements CacheItemPoolInterface
     /**
      * Pattern an entry identifier must match.
      */
-    const PATTERN_ENTRYIDENTIFIER = '/^[a-zA-Z0-9_%\-&]{1,250}$/';
+    const PATTERN_ENTRYIDENTIFIER = '/^[a-zA-Z0-9_%\-&\.]{1,250}$/';
 
     /**
      * @var BackendInterface
--- a/Neos.Cache/Tests/Unit/Psr/Cache/CachePoolTest.php
+++ b/Neos.Cache/Tests/Unit/Psr/Cache/CachePoolTest.php
@@ -12,6 +12,7 @@ namespace Neos\Cache\Tests\Unit\Psr\Cache;
  */
 
 use Neos\Cache\Backend\AbstractBackend;
+use Neos\Cache\Backend\BackendInterface;
 use Neos\Cache\Psr\Cache\CachePool;
 use Neos\Cache\Psr\Cache\CacheItem;
 use Neos\Cache\Psr\InvalidArgumentException;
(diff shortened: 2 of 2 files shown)


bwaidelich:
@kitsunet could you have a quick look into this one? It's just one removed character (and a lot of new tests) :)

kdambekalns (commented):
Wait, what? Why is `PATTERN_ENTRYIDENTIFIER` used to check a `CachePool` identifier? Seems weird to me! I'll look up what a `CachePool` is now…

bwaidelich:
@kdambekalns see https://github.com/php-fig/cache/blob/master/src/CacheItemPoolInterface.php#L25-L32

kdambekalns:
> Wait, what? Why is `PATTERN_ENTRYIDENTIFIER` used to check a `CachePool` identifier? Seems weird to me! 

To me it seems wrong, after having checked. The _cache pool identifier_ is not bound to the same rules as the _cache item identifier_.

Furthermore, it seems as if `CachePool.identifier` isn't even used _anywhere_!?

kdambekalns:
Ah, now… I somehow looked at the constructor of `CachePool`, which is using the same constant to check.

Yes, your change makes sense – but the check in the constructor doesn't. Right?

bwaidelich:
> but the check in the constructor doesn't. Right?

Good point. I think I confused the two for a minute.
The `PATTERN_ENTRYIDENTIFIER` is currently used to verify entry ids _and_ the cache id itself. I agree that this is weird, but OTOH there needs to be some validation on the cache name too and I don't really care whether it follows the same rules.

kdambekalns:
> OTOH there needs to be some validation on the cache name too

But it's not used anywhere? Or am I blind?

bwaidelich:
> But it's not used anywhere? Or am I blind?

I expected this to be the filename of the cache for a file backend for example? But I'm guessing at this point, probably @kitsunet can help us out :)

kitsunet:
I don't see a reason why this should be problematic. We use it to check the name of the cache and also the identifier for entries, BUT this is the given identiifer from userland code, backends will append context information (as in the application identifier) and potential other stuff and _usually_ then create an MD5 of that whole string. so the Identifier never appears verbatim eg in the filesystem.

bwaidelich:
> I don't see a reason why this should be problematic

@kitsunet what do you refer to? (sorry, I lost context of this one slightly).
Without this bugfix I can't use [firebase/php-jwt](https://packagist.org/packages/firebase/php-jwt) since it relies on the supported characters for the JWK keys cache

kdambekalns (approved):
I still think it's highly weird to check the cache pool identifier using the cache entry identifier rules, but… this does not harm.

kitsunet (approved):

bwaidelich:
> I still think it's highly weird to check the cache pool identifier using the cache entry identifier rules

Right, but this is unrelated to this PR, correct?

PR #2925, opened 2022-10-15
FEATURE: Add cache lifetime to routing MatchResult and ResolveResult DTOs in addition to cache tags

### LINKED ISSUE #2904: FEATURE: Make cache lifetime of routing caches configurable

### Description

The routing caches currently have no lifetime and thus all entries live forever unless someone configures a default lifetime. 
This is generally fine but some routes may still require different lifetimes than others. It would be great to configure the lifetime for the routing and resolve cache for each route individually.

### Possible Solution

Add the options `routeCacheLifetime` and `resolveCacheLifetime` to the routing configuration

```yaml
-
  name: 'Some route'
  ...
  cacheLifetime: null|int(default null)
```

This is a based on the issue #1426 but uses a different approach

mficzel:
This change allows routeParts to specify the maximum lifetime for all cacheItems associated with the current route. This makes sense as routeParts already are allowed to define tags for the cache. If multiple routeParts set a cacheLifetime the lowest value that is not null is used for the cache entries.

As an additional option the routing configuration now allows to configure "cache.lifetime" and "cache.tags" to specify the lifetime and tags that are merged with the lifetime and tags defined by the route parts.

Resolves: #2904 

**Checklist**


mficzel:
I am not entirely sure yet wether the configuration should be `cacheLifetime` or `cache.lifetime` as the second would allow further extensions in the future but would be cumbersome if those never occur. 

For now i chose the `cacheLifetime` but i will gladly adjust if reviewers prefer the nested form.

sorenmalling:
I dislike this solution. Seems like we are adding a setting to routing configuration, that should not exist.

Reading the original thread, it looks like we are solving a issue of bad data from a provider (if I get your issue correct) by changing the framework.

@bwaidelich seems to suggest a solution and a explanation to your issue. So why this change with a addon to the configuration?

mficzel:
> So why this change with a addon to the configuration?

Because regardless of the optimizations to the routing that we plan and should do i do not think that the current "all routing cache items use the same lifetime" approach is really valid for all cases so i still would like to have a way to adjust this on a per route basis.

mficzel:
> Seems like we are adding a setting to routing configuration, that should not exist.

This is only valid if we really can handle all cases well and are really sure about this. I personally doubt that this is really possible and would like to have ways to adjust those caches. Especially since those already have caused issues in big systems by cache flooding. Even with multiple core devs looking it took days to really understand what was going on.

A welcome side effect of adding a configuration also is that it makes the need to be aware of the caching transparent to developers.

sorenmalling:
> A welcome side effect of adding a configuration also is that it makes the need to be aware of the caching transparent to developers.

Totally disagree. It clutters the configuration and introduces a thing that I should/should not focus on.

Why is this not a routeparthandlers job to determine how to handle this?

And if takes multiple days for us to find out what happens, how are we making the situation better for future us, by adding this configuration? Would a focus on simplicity of the routing functionality, and then a focus on hooks/events, help us more?

mficzel:
There is no fix for appendExceedingArguments other then deprecating it and that means we have to still support it for a while. It may be technically possible to handle it but that would add a lot of complexity to the router and would be really risky.

I consider configuring caching the simplest simple way to mitigate the problems we have and still useful once appendExceeding arguments is removed. Probably only used for edge cases then.

Moving this to the route part is a good idea, however usually routes are configured so the lifetime should be configured aswell. It would make sense to allow the routePart to set a cachelLfetime and use the lowest or null. Would be breaky as it would add a method to the routePartInterface.

mficzel:
@sorenmalling i adjusted the implementation to primarily work via the MatchResult and ResolveResult DTO which already handle the cache-tags. Adding lifetime here is a logical extension and possible since the DTOs are final and thus are no extension point. 

The routing configuration that i still would like to add now configures values for tags and lifetime that are merged with the results from the route parts.

If we can agree on that approach i would like to add some more tests still.

mficzel:
@bwaidelich Do you have any opinions about the adjusted suggestion.

Neos.Flow/Classes/Mvc/Routing/Dto/MatchResult.php
```diff
@@ -31,14 +31,20 @@ final class MatchResult
...
+     * @var RouteLifetime|null
+     */
+    private $lifetime;
+
     /**
      * @param mixed $matchedValue
      * @param RouteTags $tags
      */
-    public function __construct($matchedValue, RouteTags $tags = null)
+    public function __construct($matchedValue, RouteTags $tags = null, RouteLifetime $lifetime = null)
```
bwaidelich, line 43: nitpick: AFAIK we target PHP 8.1+ with Neos 8.3 so we could replace this with
```suggestion
    public function __construct(
        $matchedValue,
        RouteTags $tags = null,
        private readonly ?RouteLifetime $lifetime = null
    )
```

Neos.Flow/Classes/Mvc/Routing/Dto/RouteLifetime.php
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @Flow\Proxy(false)
+ */
+final class RouteLifetime
+{
+
+    /**
+     * @var int|null
+     */
+    private $value;
```
bwaidelich, line 31: Why `int|null`
I assume because `null` has some special meaning (does not expire?).
IMO we should make that more explicit or at least document it and add an example.

Neos.Flow/Classes/Mvc/Routing/Route.php
```diff
@@ -72,6 +73,20 @@ class Route
      */
     protected $appendExceedingArguments = false;
 
+    /**
+     * Specifies the tags to be associated with cache items generated by this route
+     *
+     * @var RouteTags|null
+     */
+    protected $cacheTags = null;
```
bwaidelich, line 81: Is there a specific reason for not adding type hints like
```suggestion
    protected ?RouteTags $cacheTags = null;
```
?
To stay in sync with the rest of the file?

Neos.Flow/Classes/Mvc/Routing/Dto/RouteLifetime.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        $this->value = $value;
+    }
+
+    /**
+     * Creates an empty instance without a specific lifetime
+     *
+     * @return RouteLifetime
+     */
+    public static function createEmpty(): self
```
bwaidelich: An "empty" lifetime does not make sense to me. I think we should pull the nullable type up one level

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -265,6 +267,14 @@ protected function createRoutesFromConfiguration()
                 }
                 $routesWithHttpMethodConstraints[$uriPattern] = false;
             }
+            if (isset($routeConfiguration['cache'])) {
+                if (isset($routeConfiguration['cache']['lifetime']) && is_int($routeConfiguration['cache']['lifetime'])) {
```
bwaidelich: slightly more defensive would be something along the lines of
```suggestion
                if (isset($routeConfiguration['cache']['lifetime'])) {
                    assert(is_int($routeConfiguration['cache']['lifetime']), 'cache.lifetime must be of typ e integer');
```

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -265,6 +267,14 @@ protected function createRoutesFromConfiguration()
...
+                    $route->setCacheLifetime(RouteLifetime::createFromInt($routeConfiguration['cache']['lifetime']));
+                }
+                if (isset($routeConfiguration['cache']['tags']) && is_array($routeConfiguration['cache']['tags'])) {
```
bwaidelich: same here
```suggestion
                if (isset($routeConfiguration['cache']['tags'])) {
                    assert(is_array($routeConfiguration['cache']['tags']), 'cache.tags must be of type array');
```

bwaidelich (commented):
I'm still not super convinced that a lifetime on routes is the best way to fix issues. But it doesn't do any harm I guess. And the option to attach tags to routes is very useful indeed.
Apart form the (partly nitpicky) comments I wonder if the `RouterCachingService` should be extended by a method to flush caches of certain tags

Neos.Flow/Classes/Mvc/Routing/Dto/RouteLifetime.php
mficzel: true ... renamed to "createUndefined". Left the method in place as it matches the createEmpty of the RouteTags DTO. This allows to handle both very similar in the RouterCachingService.

mficzel:
@bwaidelich the method \Neos\Flow\Mvc\Routing\RouterCachingService::flushCachesByTags already exists :-)

Neos.Flow/Classes/Mvc/Routing/Route.php
mficzel: Yes it would look foreign in there. That is also why the new Route Lifetime DTO looks more like the RouteTags

Neos.Flow/Classes/Mvc/Routing/Router.php
mficzel: "Assertions should not be used for normal runtime operations like input parameter checks. As a rule of thumb your code should always be able to work correctly if assertion checking is not activated." 

https://www.php.net/manual/en/function.assert.php 

Maybe i get something wrong but it would consider settings kind of an input parameter.

Neos.Flow/Classes/Mvc/Routing/Router.php
mficzel: see below

Neos.Flow/Classes/Mvc/Routing/Dto/MatchResult.php
mficzel: To be in line with the rest of the file

Neos.Flow/Classes/Mvc/Routing/Dto/RouteLifetime.php
mficzel: Added a comment to explain null = undefined vs 0 = infinite

Neos.Flow/Classes/Mvc/Routing/Dto/MatchResult.php
bwaidelich: makes sense

Neos.Flow/Classes/Mvc/Routing/Router.php
bwaidelich: This was just meant as an example.
My point was not to silently ignore invalid configuration, you could replace the `assert` with an exception of course.

Neos.Flow/Tests/Unit/Mvc/Routing/Dto/RouteLifetimeTest.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public function mergeReturnsLowerLifetimeOfNonNullValuesDataProvider(): array
+    {
+        return [
+            [100, 200, 100],
+            [100, 100, 100],
+            [null, 200, 200],
+            [200, null, 200],
+            [null, null, null]
```
bwaidelich: This needs a case that contains `0` IMO since that seems to have a special meaning

Neos.Flow/Classes/Mvc/Routing/Dto/RouteLifetime.php
bwaidelich: I see, thanks.
I still think it's kind of weird to have a Value Object that can represent `undefined` and I would suggest to solve that in a higher level (i.e. either a route _has_ a lifetime or not).

Neos.Flow/Classes/Mvc/Routing/Dto/RouteLifetime.php
```diff
@@ -0,0 +1,88 @@
...
+     * Merges two instances of this class combining results ny using the lowest lifetime
+     *
+     * @param RouteLifetime $lifetime
+     * @return RouteLifetime
+     */
+    public function merge(RouteLifetime $lifetime): self
+    {
+        $values = array_values(array_filter([$this->getValue(), $lifetime->getValue()]));
+        if (!empty($values)) {
+            return self::createFromInt(min($values));
```
bwaidelich: A lifetime of 0 (infinite) would always win the race, which isn't correct I suppose.
I would even suggest to make `0` a more explicit case (e.g. by adding `createInfinite()` and `isInfinite()`)

bwaidelich (commented):
>  the method \Neos\Flow\Mvc\Routing\RouterCachingService::flushCachesByTags already exists :-)

Ah, of course :)

Neos.Flow/Tests/Unit/Mvc/Routing/Dto/RouteLifetimeTest.php
mficzel: You are right ... added more cases

Neos.Flow/Classes/Mvc/Routing/Dto/RouteLifetime.php
mficzel: Adjusted the merge logic and added the methods `createInfinite` `isInfinite` and `isUndefined`

While i understand that the undefined case is a bit weird in the dto. This is what allows to handle the conditions how two are merged in the dto. Otherwise the route would always have to check wether or not there is a previous lifetime to merge with the lifetime from the next route part.

mficzel: Also i now only ignore the configuration if it is null for lifetime or empty for tags. Other type issues would be raised by the php types in the dtos.

mficzel:
@bwaidelich afaik 8.3 will still support php 8.0. The 9.0 branch raises the min php version to 8.1

mhsdesign (approved):
Fine by explanation and reading and concept :D

ahaeslich (approved):
Fine by reading. Thanks @mficzel for explaining the changes.

mficzel:
Was merged after a mob review session on the sprint. Style arguments are correct but it was decided that an implementation parallel to existing tags was preferrable.

PR #2926, opened 2022-10-22
BUGFIX: disable invisible composer interaction in `./flow package:create`

### LINKED ISSUE #2757: BUG: `./flow package:create` freezes
### Description
`./flow package:create`

(and also `./flow kickstart:site`)

seem to freeze consistenly on (some?) setups on Flow + Neos 7.3

i experienced it myself and i have heard about that ...

i dont get any cli output and there is nothing in the logs ...

the freezing always happened after the new folder and subfolders and the composer.json of the new package where created in the `DistributionPackages` and before the new package would be required in the root compose.json

there doesnt seem to happen anything after waiting some time.

i could only quit the cli by pressing `CTRL + C` AND `Enter`

### Steps to Reproduce

run `./flow package:create`

#### Expected behavior

expect a new fancy package, and it being required in composer root

and expect message:
`Created new package "Foo.Bar" at Bar`

#### Actual behavior

cli freeze, and no require in composer root, and when using  `./flow kickstart:site` no further template files are created.

### Affected Versions

Only?
Neos: 7.3.2
Flow: 7.3.2

### Code location
this is the controller `./flow package:create`, which uses: `$this->packageManager->createPackage` some thing must have changed ...
https://github.com/neos/flow-development-collection/blob/f3358f46f4ff16df530913fa0ab587d1176d99a5/Neos.Flow/Classes/Command/PackageCommandController.php#L99

mhsdesign:
closes #2757

with #2571 was introduced, that the composer subcommand used by `./flow package:create` hides its output. But using the `new NullOutput()` doesn't prevent any interaction initiated by composer such as the infamous:

> neos/composer-plugin contains a Composer plugin which is currently not in your allow-plugins config. See https://getcomposer.org/allow-plugins
> Do you trust "neos/composer-plugin" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?]

the current problem is that composer asks this question but shows not output. So one has to blind input `y` without even knowing the question. This is perceived as bug but clearly a feature see #2757.

With this bugfix we use the official supported cli flag `--quiet` to archive the same as the NullOutput and furthermore disable any invisible requested interactivity.

in the above case instead of a question - an error will be thrown, which is understandable and fixable by the user:

> In PluginManager.php line 768:
> neos/composer-plugin contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe.
> ...
> The installation was not successful.
> Composer returned the error code: 1
>
>  Type: Neos\Flow\Package\Exception
>  Code: 1572187932
>  File: Packages/Framework/Neos.Flow/Classes/Package/PackageManager.php
>  Line: 391
>
> Open Data/Logs/Exceptions/202210221257240bfbde.txt for a full stack trace.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Package/PackageManager.php
+++ b/Neos.Flow/Classes/Package/PackageManager.php
@@ -24,7 +24,6 @@ use Neos\Utility\OpcodeCacheHelper;
 use Neos\Flow\Package\Exception as PackageException;
 use Composer\Console\Application as ComposerApplication;
 use Symfony\Component\Console\Input\ArrayInput;
-use Symfony\Component\Console\Output\NullOutput;
 
 /**
  * The default Flow Package Manager
@@ -401,12 +400,13 @@ class PackageManager
             $composerRequireArguments = new ArrayInput([
                 'command' => 'require',
                 'packages' => [$manifest['name'] . ' @dev'],
-                '--working-dir' => FLOW_PATH_ROOT
+                '--working-dir' => FLOW_PATH_ROOT,
+                '--quiet'
             ]);
 
             $composerApplication = new ComposerApplication();
             $composerApplication->setAutoExit(false);
-            $composerErrorCode = $composerApplication->run($composerRequireArguments, new NullOutput());
+            $composerErrorCode = $composerApplication->run($composerRequireArguments);
 
             if ($composerErrorCode !== 0) {
                 throw new Exception("The installation was not successful. Composer returned the error code: $composerErrorCode", 1572187932);


mhsdesign:
## How to test?

edit you root composer json to not include 
```
    "allow-plugins": {
      "neos/composer-plugin": true
    }
```

use `./flow package:create Foo.Bar`

expect an error.

dont expect an infinite empty shell.

mficzel (approved):
Makes sense, much better then hiding errors

mhsdesign:
test failing is unrelated

bwaidelich (approved):
+1 by reading

PR #2930, opened 2022-11-07
BUGFIX: Atomic reads for cache files

kitsunet:
Reading from the same cache file several times and relying on a previously determined file size for offsets is dangerous because the file could be overwritten or deleted at any point in between. It is safer to read the whole file content and process it based on what was read afterwards.

This introduces a `FileBackendDataDto` that encapsulates encoding and decoding of the complex data needed for the `FileBackend`, which include tags and lifetime.

This means reads and writes are now the same as for the `SimpleFileBackend` and atomic at all times.

**Checklist**


kitsunet:
UnitTests need still to be adjusted, I am aware, just wanted to put this out there because it's quite the change for a bugfix.

kitsunet:
Psalm failure seems unrelated. Good to go IMHO.

mhsdesign:
I did came across the error `fread(): Argument #2 ($length) must be greater than 0` while doing heavy e2e testing.

image

I will check if this fixes it ;)
Or will this also be fixed with: https://github.com/neos/flow-development-collection/pull/2934?

Edit: it seems the other fix will do too - but this seems like a more proof fix in general ...

kdambekalns:
> Edit: it seems the other fix will do too - but this seems like a more proof fix in general ...

The other PR probably fixes this, yes. But has also value on it's own, as there can be empty cache entries…

mficzel:
I really like the change but since the other pr addresses the same error i would prefer to introduce the DTO not in a bugfix but use the other pr for that. That would also give the option to introduce methods with strict types like readCacheDto and writeCacheDto on top of readCacheFile and writeCacheFile instead of creating the DTO on various places for backwards compatibility.

kdambekalns:
I just merged https://github.com/neos/flow-development-collection/pull/2934, so this can be continued in a relaxed manner to fix it properly.

mhsdesign:
so we should target main/8.3?

kitsunet:
How do we proceed with this? To me this is still a bugfix as the problem of atomicity of operations still exists regardless of the empty file problem.

mficzel:
I would like to see this in 8.3 as it imho is a little complex to be added to 7.3+ via bugfix. But i am not strictly against that either.

kitsunet:
Fine by me :)

kdambekalns:
@kitsunet Do those failures point to a problem with the cache? 😎

kitsunet:
I probably messed up the merge, I will have a look.

kitsunet:
Mmmm, I guess psalm is not wrong there, I will make the DTO final...

kitsunet:
Much better, only unrelated psalm issues left. IMHO raedy to merge. :)

kdambekalns (approved):

mhsdesign:
> I did came across the error `fread(): Argument #2 ($length) must be greater than 0` while doing heavy e2e testing.
> 
> image
> 
> I will check if this fixes it ;) Or will this also be fixed with: #2934?
> 
> Edit: it seems the other fix will do too - but this seems like a more proof fix in general ...

I just tried really hard to reproduce this (with reverted https://github.com/neos/flow-development-collection/pull/2934) by doing what i thought would cause trouble: opening the ui in parallel (automatically) a lot and doing changes, but i was out of luck, it worked flawless without #2934 and so i imagine also without your fix ... im a bit out of ideas how to reproduce this in real live so i guess im stuck with reading the code :D

Neos.Cache/Classes/Backend/FileBackendEntryDto.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    private $expiryTime;
+
+
+    public function __construct(string $data, array $tags, int $expiryTime)
+    {
+        $this->data = $data;
+        $this->tags = $tags;
+        $this->expiryTime = $expiryTime;
+    }
```
mhsdesign:
```suggestion
    public function __construct(
        private string $data,
        /** @var string[] */
        private array $tags,
        private int $expiryTime
    ) {
    }
```

mficzel:
@mhsdesign can confirm that the error occurred frequently in ddev on mac. May be hard to trigger manually but is was real.

mhsdesign:
@kitsunet with #2934 merged this is now a clean up of the hotfix right?

kitsunet:
Empty data was really only a side thing of this. The issue is that we read from the same file multiple times (for metadata and content) which is dangerous as the file could very well be removed or changed in between, that is compounded by relying on the metadata we read to read the content of the file, which might be completely inaccurate if the file has changed.

Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
```diff
@@ -531,7 +536,7 @@ public function requireOnceDoesNotCheckExpiryTimeIfBackendIsFrozen()
         $backend = $this->prepareDefaultBackend(['isCacheFileExpired']);
         $backend->setCache($mockCache);
 
-        $backend->expects(self::once())->method('isCacheFileExpired'); // Indirectly called by freeze() -> get()
+        $backend->expects(self::never())->method('isCacheFileExpired'); // Indirectly called by freeze() -> get()
```
mhsdesign:
```suggestion
        $backend->expects(self::never())->method('isCacheFileExpired');
```

Neos.Cache/Classes/Backend/FileBackend.php
```diff
@@ -251,35 +251,27 @@ public function findIdentifiersByTag(string $searchedTag): array
     public function findIdentifiersByTags(array $tags): array
     {
         $entryIdentifiers = [];
-        $now = $_SERVER['REQUEST_TIME'];
```
mhsdesign, line 254: is there a reason to why we used that to calculate the expiry? (now your dto uses time())

Neos.Cache/Classes/Backend/FileBackendEntryDto.php
kitsunet: Right, I guess now with targeting 8.x we can do this, thx!

Neos.Cache/Classes/Backend/FileBackend.php
kitsunet: Mmm, I guess it makes it consistent for this request, why not. I will change it in the DTO.

kitsunet:
Interesting test failure, I guess correct but how did it work before my changes then....

mhsdesign:
wtf is wrong with the tests xD

Neos.Cache/Classes/Backend/FileBackend.php
mhsdesign: I mean i dont really know whats "better" :D time() feels a bit cleaner though ...

_technically_ its not clean to do this at all in the dto since its a sideeffect ... but if we dont tell anyone :D

kitsunet:
I almost think it was never tested with these tests before LOL.... I will have a deeper look this afternoon.

kitsunet:
it gets worse, I currently can't replicate those failures locally.

kdambekalns:
I can confirm the test failure locally – but adjusting the test to use `time() - 2` fixes it. I guess if the tests run fast enough, the entry just isn't expired yet…

Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
```diff
@@ -292,35 +294,38 @@ public function getReturnsFalseForExpiredEntries()
...
+        $backend = $this->prepareDefaultBackend(['dummy'], [
             __DIR__ . '~Testing',
             'vfs://Foo/',
             255
         ]);
 
-        $backend->expects(self::once())->method('isCacheFileExpired')->with('vfs://Foo/Cache/Data/UnitTestCache/ExpiredEntry')->will(self::returnValue(true));
+        $entryIdentifier = 'ExpiredEntry';
+        $pathAndFilename = 'vfs://Foo/Cache/Data/UnitTestCache/' . $entryIdentifier;
+        $entryDto = new FileBackendEntryDto('data', [], time() - 1);
```
kdambekalns:
```suggestion
        $entryDto = new FileBackendEntryDto('data', [], time() - 5);
```

kdambekalns:
> it gets worse, I currently can't replicate those failures locally.

@kitsunet – it seems you need a faster machine. 🙈

kitsunet:
> > it gets worse, I currently can't replicate those failures locally.
> 
> @kitsunet – it seems you need a faster machine. 🙈

Ha, well I guess I wil have to try without docker running? IDK what I am doing wrong 🙈

kitsunet:
> > I did came across the error `fread(): Argument #2 ($length) must be greater than 0` while doing heavy e2e testing.
> > image
> > I will check if this fixes it ;) Or will this also be fixed with: #2934?
> > Edit: it seems the other fix will do too - but this seems like a more proof fix in general ...
> 
> I just tried really hard to reproduce this (with reverted #2934) by doing what i thought would cause trouble: opening the ui in parallel (automatically) a lot and doing changes, but i was out of luck, it worked flawless without #2934 and so i imagine also without your fix ... im a bit out of ideas how to reproduce this in real live so i guess im stuck with reading the code :D

You would have to really produce A LOT of requests and get lucky (unlucky rather)

PR #2932, opened 2022-11-10
TASK: Extract methods `set`, `get`, `getByTag`, `has`, and `remove` from `FrontendInterface` into `FlowCacheFrontendInterface`

mficzel:
The frontend interface now only describes method needed by the cache manager and the methods used for interaction with the cache are moved to the new `FlowCacheFrontendInterface`. This change will allow to implement frontends that provide a differrent api like PSR6, PSR16 directly.

About breakiness:
- Since code always use one of the concrete frontend implementations no changed are needed
- Cache Backends still get the same Frontend set and as long as the backends only use the getIdentifier
method which is common no changes are needed.

**Review instructions**

Stumbled upon this when i played with https://github.com/sitegeist/Sitegeist.StaleMate (Caches with Stale items and async udates) and was not able to implement those in an alternate CacheFrontend. Also this will allow to create PSR6, PSR16  compatible cache frontends that are handled like all the other frontends and 

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Frontend/AbstractFrontend.php
+++ b/Neos.Cache/Classes/Frontend/AbstractFrontend.php
@@ -21,7 +21,7 @@ use Neos\Cache\Backend\TaggableBackendInterface;
  *
  * @api
  */
-abstract class AbstractFrontend implements FrontendInterface
+abstract class AbstractFrontend implements FlowCacheFrontendInterface
 {
     /**
      * Identifies this cache
--- a/Neos.Cache/Classes/Frontend/CacheEntryIterator.php
+++ b/Neos.Cache/Classes/Frontend/CacheEntryIterator.php
@@ -23,7 +23,7 @@ use Neos\Cache\Backend\IterableBackendInterface;
 class CacheEntryIterator implements \Iterator
 {
     /**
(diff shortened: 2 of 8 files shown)


mficzel:
I wonder wether the constructors should be part of the cache interface since the cache manager relies on those for instantiating. Not sure wether this would be breaking as caches have implement them implicitly anyways.

kitsunet:
Definitely breaking as any existing Frontend would not be implementing the right interface anymore. Will look more into it later. But generally like the separation

mficzel:
@kitsunet I adjusted to extract a LowLevelCacheInterface and leave the FrontendInterface as it is ... not sure wether this is less breaky or a good idea.

mficzel:
I reverted the idea of a LowLevelCacheInterface since this would require changes in the backend implementations (other frontend is passed) and am back to the `FlowCacheFrontendInterface ` on top of `FrontendInterface`.

mficzel:
To discuss:
- Which methods should be extracted from the frontend interface
- Is there a better name then FlowCacheFrontendInterface

mficzel:
Close for now as this is not finished and will need a rework afterwards

PR #2933, opened 2022-11-10
BUGFIX: Fix truncation of cache file when lock not acquired

christianharrington:
Previously, SimpleFileBackend would truncate the contents of the cache file if it couldn't acquire the lock.

To fix this, the `c` flag is used for `fopen`, so the file is not truncated if it already exists. This necessitates that we manually truncate the file if the lock is acquired.

I've added a test that fails without these changes, and passes with them.

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/SimpleFileBackend.php
+++ b/Neos.Cache/Classes/Backend/SimpleFileBackend.php
@@ -535,11 +535,16 @@ class SimpleFileBackend extends IndependentAbstractBackend implements PhpCapable
         for ($i = 0; $i < 3; $i++) {
             $result = false;
             try {
-                $file = fopen($cacheEntryPathAndFilename, 'wb');
+                // It's important that we use the 'c' flag below, as `fopen` will otherwise truncate the file
+                // if it already exists. If we used 'w' and then didn't get the lock, the file would end up empty.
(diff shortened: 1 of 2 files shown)


kdambekalns (commented):
Looks good by reading. Good catch!

kitsunet (approved):

kitsunet:
Indeed great catch.

kitsunet:
waiting for ci would merge then.

christianharrington:
I think the failing CI check is unrelated. Let me know if I need to do anything.

christianharrington:
@kdambekalns @kitsunet I would love to get this merged in. Is there anything I can do to help?

christianharrington:
@kdambekalns 
With #2934 merged, this bug should no longer cause an exception directly (since `fread` will not be called with length 0), but will instead cause `get` to return an unexpected value (`false`, I believe, since `unserialize('') === false`).

This could introduce subtle bugs in code calling `get` but not checking the returned value, such as in [CldrModel](https://github.com/neos/flow-development-collection/blob/aeff374ffb973f7f597da3f106be3d252f9a2639/Neos.Flow/Classes/I18n/Cldr/CldrModel.php#L109-L110).

In our production application, this race-condition is triggered many times each day, so I would love to see this fix get merged.

kdambekalns:
@christianharrington Thanks for the reminder, I had completely forgotten about this one! Sorry!

PR #2936, opened 2022-11-27
FEATURE: Add support for secrets stored in a file for the configuration

### LINKED ISSUE #2935: FEATURE: Add support for secrets stored in a file for the configuration

### Description

The configration files already accepts environment variables but it would be helpful if we also can use sensitive data that was stored in a file in example in a [Docker Secret](https://docs.docker.com/engine/swarm/secrets/).

### Possible Solution

A solution could  be something like this:

```yaml
# Settings.yaml

Neos:
  Flow:
    persistence:
      backendOptions:
        driver: pdo_mysql
        dbname: '%file:/run/secrets/db_dbname%'
        user: '%file:/run/secrets/db_user%'
        password: '%file:/run/secrets/db_password%'
        host: '%file:/run/secrets/db_host%'
```

erkenes:
Adds the support of secrets stored in files for configurations

Resolves: #2935

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Configuration/ConfigurationManager.php
+++ b/Neos.Flow/Classes/Configuration/ConfigurationManager.php
@@ -494,7 +494,7 @@ class ConfigurationManager
     }
 
     /**
-     * Replaces variables (in the format %CONSTANT% or %env:ENVIRONMENT_VARIABLE%)
+     * Replaces variables (in the format %CONSTANT%, %env:ENVIRONMENT_VARIABLE% or %file:/path/to/file%)
      * in the given php exported configuration string.
      *
      * This is applied before caching to allow runtime evaluation of constants and environment variables.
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -482,7 +482,7 @@ public function refreshConfiguration(): void
     }
 
     /**
-     * Replaces variables (in the format %CONSTANT% or %env:ENVIRONMENT_VARIABLE%)
+     * Replaces variables (in the format %CONSTANT%, %env:ENVIRONMENT_VARIABLE% or %file:%PATH_TO_FILE%)
```
tantegerda1: documentation contains an extra `%` before `PATH_TO_FILE`.
I'd also suggest to write the example as `or %file:/path/to/file%`.

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -500,7 +500,7 @@ protected function replaceVariablesInPhpString(string $phpString): string
             |                                  # or
             (?:(?P<prefix>[a-z]+)\:)           # a prefix followed by : (like "env:")
             )?
-            (?P<name>[A-Z_0-9]+))              # the actual variable name in all upper
+            (?P<name>[A-Za-z_0-9\/]+))         # the actual variable name in all upper
```
tantegerda1: we're loosening the restrictions for all replacement types here, but the comment still says "variable name in all upper"

erkenes:
@tantegerda1 I updated the documentation :)

sorenmalling:
Can you add a test for this feature?

mficzel:
That is a feature and should not be introduced in a bugfix, i think the pr should target 8.3. Also there have been extensions to the env handling in https://github.com/neos/flow-development-collection/pull/2873 which will interfere with this change.

kitsunet:
I guess this needs an additional condition to check for env if the given string is uppercase.

erkenes:
@sorenmalling I've finally added a unit test.

@kitsunet Furthermore, env variables are checked to ensure the name is uppercase.

PR #2942, opened 2022-12-13
TASK: Add PHP 8.1 & 8.2 to CI build

kdambekalns:


kdambekalns:
I couldn't care less about those Psalm errors right now, and there are a ton of deprecation wanrings on PHP 8.2 (no more dynamic properties…) – but it passes.

fcool (approved):

albe (approved):
Did something change in DateTime parsing in 8.2? I couldn't find something about that. Anyway, yes, the psalm errors are likely unrelated and subject to some subtle changes in Doctrine.

kdambekalns:
> Did something change in DateTime parsing in 8.2? 

Why DateTime parsing?

PR #2944, opened 2022-12-13
BUGFIX: Support PHP8 union and intersection types as method return value in reflection service

### LINKED ISSUE #2941: BUG: Use of union types breaks proxy building, creates broken PHP code

### Current Behavior

When union types are used as return values in methods that are proxied types like `int|false` are interpreted as `\int|false` which leads to php errors.

### Expected Behavior

Union types should be allowed to be used as method arguments and return values also on class-methods that are proxied. At the very least an error should be raised and notify the user about the missing support for union types.

### Steps To Reproduce

Crete a class with a method that uses union types and somehow force flow to crate an proxy for that. An aspect on that method should do the trick. You should now get a php error.

### Environment

```markdown
- Flow: 7.3
- PHP: 8.0
```

### Anything else?

Union types are allowed since php 8 which is the minimal php version flow currently supports. Given this it can be expected that union types can be used in code based on flow.

mficzel:
The change allows to use union types and intersection types in classes which are proxied by the flow framework. 
Offcourse the php version Flow is running on has to support the used features aswell.

In addition: 
- `true`, `false` and `null` are added to the list of simple types which are not prefixed by the reflection service
- `parent`, `static`, `never` and `resource` are added to the list of additional types that need no prefix during reflection aswell

Resolves: #2941 

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1414,7 +1414,6 @@ class ReflectionService
         $applyLeadingSlashIfNeeded = function (string $type): string {
             if (!in_array($type, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed'])
                 && !TypeHandling::isSimpleType($type)
-                && !TypeHandling::isLiteralType($type)
             ) {
                 return '\\' . $type;
             }
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1410,12 +1410,27 @@ class ReflectionService
             $this->classesByMethodAnnotations[$annotationClassName][$className][] = $methodName;
         }
 
-        $returnType = $method->getDeclaredReturnType();
-        if ($returnType !== null && !in_array($returnType, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed']) && !TypeHandling::isSimpleType($returnType)) {
-            $returnType = '\\' . $returnType;
-        }
-        if ($method->isDeclaredReturnTypeNullable()) {
-            $returnType = '?' . $returnType;
+        $returnType= $method->getDeclaredReturnType();
+        $applyLeadingSlashIfNeeded = function (string $type) {
+            if (!in_array($type, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed'])
+                && !TypeHandling::isSimpleType($type)
+                && !TypeHandling::isLiteralType($type)
+            ) {
+                return '\\' . $type;
+            }
+            return $type;
+        };
+        if ($returnType !== null) {
+            if (TypeHandling::isUnionType($returnType)) {
+                $returnType = implode('|', array_map($applyLeadingSlashIfNeeded, explode('|', $returnType)));
+            } elseif (TypeHandling::isIntersectionType($returnType)) {
+                $returnType = implode('&', array_map($applyLeadingSlashIfNeeded, explode('&', $returnType)));
+            } else {
+                $returnType = $applyLeadingSlashIfNeeded($returnType);
+                if ($method->isDeclaredReturnTypeNullable()) {
+                    $returnType = '?' . $returnType;
+                }
+            }
         }
         $this->classReflectionData[$className][self::DATA_CLASS_METHODS][$methodName][self::DATA_METHOD_DECLARED_RETURN_TYPE] = $returnType;
 
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1412,7 +1412,7 @@ class ReflectionService
 
         $returnType= $method->getDeclaredReturnType();
         $applyLeadingSlashIfNeeded = function (string $type): string {
-            if (!in_array($type, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed'])
+            if (!in_array($type, ['self', 'parent', 'static', 'null', 'callable', 'void', 'never', 'iterable', 'object', 'resource', 'mixed'])
                 && !TypeHandling::isSimpleType($type)
             ) {
                 return '\\' . $type;
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1411,7 +1411,7 @@ class ReflectionService
         }
 
         $returnType= $method->getDeclaredReturnType();
-        $applyLeadingSlashIfNeeded = function (string $type) {
+        $applyLeadingSlashIfNeeded = function (string $type): string {
             if (!in_array($type, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed'])
                 && !TypeHandling::isSimpleType($type)
                 && !TypeHandling::isLiteralType($type)
--- a/Neos.Flow/Tests/Functional/Reflection/Fixtures/DummyClassWithTypeHints.php
+++ b/Neos.Flow/Tests/Functional/Reflection/Fixtures/DummyClassWithTypeHints.php
@@ -31,4 +31,8 @@ class DummyClassWithTypeHints
     public function methodWithArrayTypeHintAndAnnotation(array $array)
     {
     }
+
+    public function methodWithUnionReturnType(): string|false
+    {
+    }
 }
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -111,18 +111,7 @@ abstract class TypeHandling
      */
     public static function isSimpleType(string $type): bool
     {
-        return in_array(self::normalizeType($type), ['array', 'string', 'float', 'integer', 'boolean', 'null'], true);
-    }
-
-    /**
-     * Returns true if the $type is a literal type.
-     *
-     * @param string $type
-     * @return boolean
-     */
-    public static function isLiteralType(string $type): bool
-    {
-        return in_array($type, ['false', 'true'], true);
+        return in_array(self::normalizeType($type), ['array', 'string', 'float', 'integer', 'boolean', 'null', 'false', 'true'], true);
     }
 
     /**
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -122,7 +122,7 @@ abstract class TypeHandling
      */
     public static function isLiteralType(string $type): bool
     {
-        return in_array(self::normalizeType($type), ['false', 'true'], true);
+        return in_array($type, ['false', 'true'], true);
     }
 
     /**
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -111,7 +111,7 @@ abstract class TypeHandling
      */
     public static function isSimpleType(string $type): bool
     {
-        return in_array(self::normalizeType($type), ['array', 'string', 'float', 'integer', 'boolean'], true);
+        return in_array(self::normalizeType($type), ['array', 'string', 'float', 'integer', 'boolean', 'null'], true);
     }
 
     /**
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -115,6 +115,17 @@ abstract class TypeHandling
     }
 
     /**
+     * Returns true if the $type is a literal type.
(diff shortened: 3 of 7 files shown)


mficzel:
@albe do you agree with this solution in principle? What should we tests and what branch to target? Is this a bugfix? I personally think this should be a bugfix for 8.0.

Also do we need to handle literal, union and intersection types in method arguments and property types aswell?

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -114,6 +114,17 @@ public static function isSimpleType(string $type): bool
...
 
+    /**
+     * Returns true if the $type is a literal type.
+     *
+     * @param string $type
+     * @return boolean
+     */
+    public static function isLiteralType(string $type): bool
+    {
+        return in_array(self::normalizeType($type), ['false', 'true'], true);
```
albe:
```suggestion
        return in_array(self::normalizeType($type), ['false'], true);
```
I think `true` is not a valid literal type, only `false` is allowed. `null` would be another value in the same category (but we could discuss that in another PR). https://php.watch/versions/8.2/null-false-types

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1447,12 +1447,27 @@ protected function reflectClassMethod($className, MethodReflection $method)
...
+            if (!in_array($type, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed'])
+                && !TypeHandling::isSimpleType($type)
+                && !TypeHandling::isLiteralType($type)
+            ) {
+                return '\\' . $type;
+            }
+            return $type;
+        };
+        if ($returnType !== null) {
+            if (TypeHandling::isUnionType($returnType)) {
```
albe, line 1423: Basically this is okay as is for now.
I do think we should move all this type handling stuff back into PHP standard reflection API land though. So instead of our custom `getDeclaredReturnType()` that returns the stringified type, then parse it back if it was a union/intersection type, we should use the `getReturnType()` method as primary source of information directly
https://www.php.net/manual/en/class.reflectiontype.php
That can quickly become complex, because we might want to wrap that into a new method that falls back to docblock parsing info, which would mean we'd need to use own inherited `ReturnType` classes that we can manually instanciate (I think we don't do this yet, because we do the fallback way deeper in the code, but actually the ReflectionService should be the place to "normalize" PHP type declarations and docblock types into one souce of information). Or maybe there is a simpler intermediate step.

albe (approved):
I think this is okay for a bugfix solution to get union+intersection types to work. I outlined my prefered (midterm) solution in the comment.

> Also do we need to handle literal, union and intersection types in method arguments and property types aswell?

For completeness yes, but I'm not even sure where/how we do reflect on those in a specific way (we have https://github.com/mficzel/flow-development-collection/blob/bugfix/handleUnionTypes/Neos.Flow/Classes/Reflection/ParameterReflection.php but that does not do anything fancy besides the `isBuiltin`), that requires adjustment. A lot of adjustments will also be necessary in higher levels, where we do use the Reflection info, and that's a way bigger topic to tackle.

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
mficzel: true is valid since php 8.2 that is why i included it. See https://www.php.net/releases/8.2/en.php#null_false_true_types

mficzel: Added `null` to the simpleTypes as it is documented here https://www.php.net/manual/de/language.types.type-system.php

mficzel:
Bugfix for 8.0 or 7.3? Older versions without php 8 support are not affected imho.

fcool:
I would definitely vote for Bugfix in 7.3

Neos.Flow/Tests/Functional/Reflection/Fixtures/DummyClassWithTypeHints.php
```diff
@@ -31,4 +31,8 @@ public function methodWithArrayTypeHint(array $array)
     public function methodWithArrayTypeHintAndAnnotation(array $array)
     {
     }
+
+    public function methodWithUnionReturnType(): string|false
```
fcool: This in a test like that will probably break Testing for older PHP versions, wouldn't it? I guess we need some kind of language level detection to skip tests, which would break older PHP parsers - but I am not 100% sure.

The current Tasks only checked PHP >=8.0, if we were targetting 7.3 it would need to degrade gracefully in PHP 7.4

fcool (changes requested):

Neos.Flow/Tests/Functional/Reflection/Fixtures/DummyClassWithTypeHints.php
fcool: (If we stay with 8.x as target, that probably is not neccessary - and I'm fine with all other aspects.)

mficzel: That is why i would prefer to target 8.0. Support 7.3 would require some additional adjustments.

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
mficzel: Also: I wonder wether we should simply add `true` and `false` to the list of simple types and avoid the seperate "isLiteralType" method. Not sure.

Neos.Flow/Classes/Reflection/ReflectionService.php
mficzel: Maybe we can parse the docblocks to ReflectionTypes ...

mficzel:
@fcool @albe adjusted and rebased to 7.3 .... could actually be ready. 

The psalm errors are caused by changes in doctrine ... again.

mficzel:
I wonder wether the use of "str_contains" (php8) is ok. There is a polyfill from symfony that is loaded as dependency of the symfony components anyways. Also i tested on Neos 7.3 &  php 7.4 in addition to the test matrix here.

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -111,7 +111,18 @@ public static function isLiteral(string $type): bool
...
+
+    /**
+     * Returns true if the $type is a literal type.
+     *
+     * @param string $type
+     * @return boolean
+     */
+    public static function isLiteralType(string $type): bool
+    {
+        return in_array($type, ['false', 'true'], true);
```
grebaldi: I'm confused by the semantics here. The reason is basically that I don't know what the idea of "simple type" (which has not been introduced in this PR) refers to.

There are three methods `isLiteral`, `isSimpleType` and `isLiteralType`. Looking at:
https://github.com/neos/flow-development-collection/blob/7d02951763753b9ff80bb6b4a478b20631e76059/Neos.Utility.ObjectHandling/Classes/TypeHandling.php#L29-L32

I assume that `isLiteral` and `isSimpleType` address the same set of types (in that case, the `LITERAL_TYPE_PATTERN` would need to be adjusted to cover `null` as well).

But I'd like to suggest another way:

Maybe this would be a good opportunity to adjust the `TypeHandling` class to operate closer to PHP's type system definitions (as per https://www.php.net/manual/en/language.types.type-system.php). I'd suggest the following methods:

* `isScalarType`
  * covers `bool`, `int`, `float` and `string`
* `isBuiltInType`
  * covers `null`, `array`, `object`, `resource`, `never`, `void`, `self`, `parent` and `static`
  * falls back to `isScalarType`
* `isLiteralType`
  * Same as it is right now in this PR
* `isCallableType`
  * covers `callable`
* `isBuiltInTypeAlias`
  * covers `mixed` and `iterable`
* `isUserDefinedType`
  * calculated from: `!isBuiltInType && !isLiteralType && !isCallableType && !isBuiltInTypeAlias`
  * can be used in `ReflectionService` to determine whether the type name needs to be prefixed with `\` or not

For backwards compatibility `isLiteral` and `isSimpleType` need to continue to work, but can use the new methods. Given my suggestion, `isLiteral` and `isSimpleType` should (probably) also be deprecated.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1410,12 +1410,27 @@ protected function reflectClassMethod($className, MethodReflection $method)
...
-        if ($returnType !== null && !in_array($returnType, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed']) && !TypeHandling::isSimpleType($returnType)) {
-            $returnType = '\\' . $returnType;
-        }
-        if ($method->isDeclaredReturnTypeNullable()) {
-            $returnType = '?' . $returnType;
+        $returnType= $method->getDeclaredReturnType();
+        $applyLeadingSlashIfNeeded = function (string $type): string {
+            if (!in_array($type, ['self', 'null', 'callable', 'void', 'iterable', 'object', 'mixed'])
+                && !TypeHandling::isSimpleType($type)
+                && !TypeHandling::isLiteralType($type)
```
grebaldi: The type range covered here is not enough. As per https://www.php.net/manual/en/language.types.type-system.php, `resource`, `never`, `parent` and `static` must also be covered.

Also: I think that L1415 should be encapsuled in `TypeHandling` as well (namely through the newly suggested `isUserDefinedType` method).

grebaldi (changes requested):
Hi @mficzel,

I was able to reproduce the issue and to verify that your change works as expected. While browsing the Code however, I saw some odd semantics in the `TypeHandling` class and thus discovered that the original bug has a whole other dimension :scream: 

Long story short: There are still PHP return type names that would be destroyed during Proxy Class building (I was able to create such a class). I added some code comments to describe the issue more thoroughly.

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
mficzel: @grebaldi the idea of literal type comes from php itself https://www.php.net/manual/de/language.types.type-system.php but I find it weird myself and would be fine with just adding true false null to the simple types.

Neos.Flow/Classes/Reflection/ReflectionService.php
mficzel: Makes sense. Never and void still need to be handled separately as they cannot be used in union types. 

Also I just noticed that union types can contain intersection types in brackets … but it would be fair to ignore this in a bugfix for Neos 7.3

mficzel:
@grebaldi, thank you for your detailed feedback. I think i covered the cases you found: 

- I moved `true` and `false` to the isSimpleType method as i agree that the semantic of isLiteral vs isLiteralType is bad. As SimpleTypes are no concept of PHP we can aswell just add the values there.
- I added the missing exceptions for `static`, `parent`, `never` and `resource` to the exception list in the ReflectionService

I 100% agree to your suggestions for the additional methods in TypeHandling but opted to not introduce those in a bugfix but would add those rather as a feature to 8,3 after this fix was upmerged.

Also i did not handle the cases when never or void are wrongly used in a union type specially as it is correct that the reflection does not modify those types and php will throw the correct errors anyways. So i consider the current behavior in this regard already ok.

What also is not handled yet is nesting of intersection types inside union types. I consider this too much for this bugfix. This should be solved with a feature for a Neos release.

Neos.Flow/Classes/Reflection/ReflectionService.php
mficzel: Reading again parsing docblocks to reflections is exactly the core of your suggestion. So i agree to that.

grebaldi (approved):
Hi @mficzel,

thanks for your adjustments! LGTM :)

> I 100% agree to your suggestions for the additional methods in TypeHandling but opted to not introduce those in a bugfix but would add those rather as a feature to 8,3 after this fix was upmerged.

I opened an issue to cover this: https://github.com/neos/flow-development-collection/issues/2945

mficzel:
@robertlemke @albe @fcool a second approval would be great

fcool:
Will definitely test and check it at latest tomorrow morning!

fcool (approved):
Just a comment: Flow is still broken, if you use a type hint of `object` in the signature of a proxied method. :(

Error is: `PHP Fatal error:  Type declaration 'object' must be unqualified in ...`

But this is probably out of scope of this Bugfix, so I'm fine with creating a new issue for this behavior.

mficzel:
Just noticed that almost the same problems occur when union types or other types flow does not yet handle correctly are used in method parameters ... see https://github.com/neos/flow-development-collection/blob/7.3/Neos.Flow/Classes/Reflection/ReflectionService.php#L1834 

I wonder how to approach this, probably #2945 is needed for that.

albe (approved):

albe:
> I wonder how to approach this, probably #2945 is needed for that.

Yeah, anything beyond the (incomplete) scope of this bugfix will need much deeper change. So as long as this change doesn't build up more stumbling blocks for that bigger change (which I don't see), we should get it in IMO.

PR #2950, opened 2023-01-04
BUGFIX: Declare properties in proxy classes to prevent PHP 8.2 deprecations

### LINKED ISSUE #2946: BUG: (PHP 8.2 Support) Creation of dynamic property [...]::$Flow_Injected_Properties is deprecated

### Current Behavior

Running a basic Neos.Demo installation on PHP 8.2 currently leads to this (repeated) PHP deprecation message:
```
Deprecated: Creation of dynamic property Neos\Flow\ResourceManagement\Streams\ResourceStreamWrapper::$Flow_Injected_Properties is deprecated in /tmp/neos/Development/SubContextddev/Cache/Code/Flow_Object_Classes/Neos_Flow_ResourceManagement_Streams_ResourceStreamWrapper.php on line 591
```

Indeed, dynamic properties have been deprecated in PHP 8.2:
>  The creation of dynamic properties is deprecated to help avoid mistakes and typos, unless the class opts in by using the `#[\AllowDynamicProperties]` attribute. `stdClass` allows dynamic properties.
>
> Usage of the __get/__set magic methods is not affected by this change.

(see: https://www.php.net/releases/8.2/en.php)

### Expected Behavior

There is no deprecation message.

### Steps To Reproduce

* Set up a Neos Distribution running Neos.Demo on PHP 8.2
* See deprecation message on homepage

### Environment

```markdown
- Flow: 8.2
- PHP: 8.2
```

### Anything else?

_No response_

paxuclus:
This patch resolves deprecation notices on PHP 8.2:

## Dynamic properties in proxy classes
There are some properties that are dynamically declared in proxy classes:
* `Flow_Injected_Properties`
* `Flow_Object_PropertiesToSerialize`
* `Flow_Persistence_RelatedEntities`: this is mostly used inside of the `ObjectSerializationTrait`, so I thought it might make more sense to declare it there instead of adding a property using the `ProxyClassBuilder`

Resolves #2946

## `parent` inside of closures
There are some methods that will be checked against `is_callable` with `parent::`:
* `parent::Flow_Aop_Proxy_buildMethodsAndAdvicesArray`
* `parent::__wakeup`

**Review instructions**
* Set up a flow distribution on PHP 8.2
* Run tests and make sure no deprecation warnings are thrown

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/DependencyInjection/PropertyInjectionTrait.php
+++ b/Neos.Flow/Classes/ObjectManagement/DependencyInjection/PropertyInjectionTrait.php
@@ -16,6 +16,8 @@ namespace Neos\Flow\ObjectManagement\DependencyInjection;
  */
 trait PropertyInjectionTrait
 {
+    protected array $Flow_Injected_Properties = [];
+
     /**
      * Does a property injection lazily with fallbacks.
      * Used in proxy classes.
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
@@ -27,6 +27,10 @@ use Neos\Utility\Arrays;
  */
 trait ObjectSerializationTrait
 {
+    protected array $Flow_Object_PropertiesToSerialize = [];
+
+    protected ?array $Flow_Persistence_RelatedEntities = null;
+
     /**
      * Code to find and serialize entities on sleep
      *
@@ -43,11 +47,14 @@ trait ObjectSerializationTrait
             if (in_array($propertyName, [
                 'Flow_Aop_Proxy_targetMethodsAndGroupedAdvices',
                 'Flow_Aop_Proxy_groupedAdviceChains',
-                'Flow_Aop_Proxy_methodIsInAdviceMode'
+                'Flow_Aop_Proxy_methodIsInAdviceMode',
+                'Flow_Persistence_RelatedEntities',
+                'Flow_Object_PropertiesToSerialize',
+                'Flow_Injected_Properties',
             ])) {
                 continue;
             }
-            if (isset($this->Flow_Injected_Properties) && is_array($this->Flow_Injected_Properties) && in_array($propertyName, $this->Flow_Injected_Properties)) {
+            if (property_exists($this, 'Flow_Injected_Properties') && is_array($this->Flow_Injected_Properties) && in_array($propertyName, $this->Flow_Injected_Properties)) {
                 continue;
             }
             if ($reflectionProperty->isStatic() || in_array($propertyName, $transientProperties)) {
@@ -72,7 +79,7 @@ trait ObjectSerializationTrait
                     }
                 }
                 if ($this->$propertyName instanceof PersistenceMagicInterface && !Bootstrap::$staticObjectManager->get(PersistenceManagerInterface::class)->isNewObject($this->$propertyName) || $this->$propertyName instanceof OrmProxy) {
-                    if (!property_exists($this, 'Flow_Persistence_RelatedEntities') || !is_array($this->Flow_Persistence_RelatedEntities)) {
+                    if (!isset($this->Flow_Persistence_RelatedEntities)) {
                         $this->Flow_Persistence_RelatedEntities = [];
                         $this->Flow_Object_PropertiesToSerialize[] = 'Flow_Persistence_RelatedEntities';
                     }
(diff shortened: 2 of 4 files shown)


kitsunet:
Good idea +++ But I guess we need to ensure those properties do not get serialised seeing the test issues.

paxuclus:
@kitsunet the tests pass now. What about the target branch? Is 8.0 fine or should this go into 7.3?

markusguenther:
Good to see that coming. Had today the issue that my Neos 9.0 tryout throw the error:

```
PHP Deprecated:  Creation of dynamic property Neos\Flow\ResourceManagement\Streams\ResourceStreamWrapper::$Flow_Injected_Properties is deprecated in /Volumes/Development/Playground/neos-90/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/Neos_Flow_ResourceManagement_Streams_ResourceStreamWrapper.php on line 591
```

Fixed that now temporary with `#[\AllowDynamicProperties]` in the class, but that is just my workaround :D

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -116,7 +123,7 @@ private function Flow_searchForEntitiesAndStoreIdentifierArray(string $path, mix
                 $this->Flow_searchForEntitiesAndStoreIdentifierArray($path . '.' . $key, $value, $originalPropertyName);
             }
         } elseif ($propertyValue instanceof PersistenceMagicInterface && !Bootstrap::$staticObjectManager->get(PersistenceManagerInterface::class)->isNewObject($propertyValue) || $propertyValue instanceof DoctrineProxy) {
-            if (!property_exists($this, 'Flow_Persistence_RelatedEntities') || !is_array($this->Flow_Persistence_RelatedEntities)) {
+            if (!property_exists($this, 'Flow_Persistence_RelatedEntities') || !isset($this->Flow_Persistence_RelatedEntities) || !is_array($this->Flow_Persistence_RelatedEntities)) {
```
mhsdesign: It seems we don’t need `property_exists` anymore ;)

and we should def look up if property_exists is used elsewhere ein the codebase (as it might not work anymore)

mhsdesign:
After discussing I think we should target 7.3 https://neos-project.slack.com/archives/C04Q0TC15/p1677852202036989

Neos.Flow/Classes/ObjectManagement/DependencyInjection/PropertyInjectionTrait.php
```diff
@@ -16,6 +16,8 @@
  */
 trait PropertyInjectionTrait
 {
+    protected $Flow_Injected_Properties = [];
```
bwaidelich: I think this can have a type hint!?
```suggestion
    protected array $Flow_Injected_Properties = [];
```

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -28,6 +28,10 @@
  */
 trait ObjectSerializationTrait
 {
+    protected $Flow_Object_PropertiesToSerialize = [];
```
bwaidelich:
```suggestion
    protected array $Flow_Object_PropertiesToSerialize = [];
```

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -28,6 +28,10 @@
...
+
+    protected $Flow_Persistence_RelatedEntities = null;
```
bwaidelich:
```suggestion
    protected ?array $Flow_Persistence_RelatedEntities = null;
```

Neos.Flow/Tests/Functional/ObjectManagement/ProxyCompilerTest.php
```diff
@@ -194,7 +194,9 @@ public function complexPropertyTypesArePreserved()
         $reflectionClass = new ClassReflection(Fixtures\PHP8\ClassWithUnionTypes::class);
         /** @var PropertyReflection $property */
         foreach ($reflectionClass->getProperties() as $property) {
-            if ($property->getName() !== 'propertyA' && $property->getName() !== 'propertyB') {
+            if (str_starts_with($property->getName(), 'property') &&
```
bwaidelich: Why is this addition required?

paxuclus: By explicitly declaring the `Flow_*` properties in the traits, `$reflectionClass->getProperties()` will now also return those new properties, which breaks this test. 
I'm not really happy with this change but simply checking for the property name was the easiest way to fix the test.

This test could also be refactored to be a bit stricter with the property names that are checked (e.g. only look at `$propertyC`, `$propertyD` and `$propertyE` of the [fixture class](https://github.com/neos/flow-development-collection/blob/8.3/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP8/ClassWithUnionTypes.php#L28))

kitsunet: Incidentally I just created the same fix and stumbled on the same test failure, it makes sense.

kitsunet:
Hey, we just discussed this further on our code sprint and I would move this up to the 8.0 branch so we can be sure to have PHP 8.x requirements in place for this and add some small code cleanups I did along the way.

PR #2953, opened 2023-01-11
FEATURE: Add methods `uriInterfaceFor` and `buildUriInterface` to UriBuilder

mficzel:
The new methods return the result as a `Psr\Http\Message\UriInterface` object instead of a string and thus are much easier to process further. Other than that the behavior is not modified as the heavy lifting always had been done by the router that already returned UriInterface objects

The existing methods `uriFor` and `build` that still exist are now wrappers around the new methods that only convert the UriInterface objects to string.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
+++ b/Neos.Flow/Classes/Mvc/Routing/UriBuilder.php
@@ -19,6 +19,7 @@ use Neos\Flow\Mvc\ActionRequest;
 use Neos\Flow\Mvc\Routing\Dto\ResolveContext;
 use Neos\Flow\Mvc\Routing\Dto\RouteParameters;
 use Neos\Utility\Arrays;
+use Psr\Http\Message\UriInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


mficzel:
@bwaidelich how about this very basic approach to get UriInterfaces out of the current UriBuilder.

bwaidelich:
@mficzel I think it's a great idea to investigate alternatives to the (maybe over-sophisticated) approach we discussed with #2744 but I have some concerns/questions:
- why would I use this instead of `new Uri($uriBuilder->uriFor(...))`?
- can we come up with a better method name? IMO `uriInterfaceFor` is very misleading to someone that is not familiar with the context
- I still like the idea of a dedicated `ActionUriBuilder` that solves some other quirks along the lines.. With 9.0 we would have the chance to introduce it together with the dedicated `NodeUriBuilder`.. Maybe we can simplify #2744 slightly?

mficzel:
@bwaidelich i agree and will close this.

PR #2961, opened 2023-02-22
TASK: extending phpstorm autocomplete information, focused on attributes

fcool:
This only adds some autocomplete hints to PHP-Storm, especially the Validate and Scope - Attributes.

**Checklist**

### DIFF
--- a/Neos.Flow/.phpstorm.meta.php
+++ b/Neos.Flow/.phpstorm.meta.php
@@ -15,4 +15,20 @@ namespace PHPSTORM_META {
         \Neos\Flow\Core\Bootstrap::getEarlyInstance(),
         map(['' => '@'])
     );
+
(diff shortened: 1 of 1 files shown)


mhsdesign:
Noice :D

i dont know how far behing phpdoc is with enabling this (which in theorie should make this phpstorm meta obsolete one day as in typescript/jsdoc i also can declare such logic inline)

but as phpstorm is still far behing understanding our code like vscode understands typescript/jsdoc, i think we still need to use a specific phpstorm meta right?

fcool:
At least there I know no possibility to narrow down the expected arguments to a known set of arguments. Like this patch does for the "Scope" - Attribute (which now will let you easiliy select the three available scopes) and for "type" of the Validate argument, bringing the flow provided validators in Auto-Complete. (Of course "custom validators" will not be in the suggestion list, but you may still use custom values, without PHPStorm complaining.

kdambekalns (commented):
Looks good, but when I tried to add a `Scope` annotation to test this, nothing new happened. Anything else I might have missed, except changing the `.phpstorm.meta.php` file?

I tried it with https://plugins.jetbrains.com/plugin/7320-php-annotations **disabled**, but that didn't change anything.

fcool:
Sadly it does not work with Annotations. But with Attributes - and for them it is irrelevant if the Annotation Plugin is installed or not.

kdambekalns:
Ah, I see, *attributes*… Fine!

kdambekalns (approved):
Works

PR #2964, opened 2023-02-24
FEATURE: Allow optimization of proxy building

### LINKED ISSUE #1539: Don't create proxies for classes that don't need it
Flow creates a proxy class for every PHP class of a package with [enabled object management](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/ObjectManagement.html#enabling-other-package-classes-for-object-management) (all Flow packages by default).

This allows the corresponding class to use annotation based Dependency Injection and to be targeted by [AOP aspects](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/AspectOrientedProgramming.html)

Instead a proxy class should only be generated if it is really needed (= class uses DI or is targetted by AOP aspects).

### Background

The reason for this feature is (apart from performance) that certain construct don't work with proxy classes.
For example the following code
```php
<?php
final class SomeClass
{
    private function __construct(string $foo)
    {
        // ...
    }
}
```

leads to the proxy class code
```php
final class SomeClass extends SomeClass_Original implements \Neos\Flow\ObjectManagement\Proxy\ProxyInterface {

    // ...

    public function __construct()
    {
        $arguments = func_get_args();
        if (!array_key_exists(0, $arguments)) throw new \Neos\Flow\ObjectManagement\Exception\UnresolvedDependenciesException('Missing required constructor argument $foo in class ' . __CLASS__ . '. Note that constructor injection is only support for objects of scope singleton (and this is not a singleton) – for other scopes you must pass each required argument to the constructor yourself.', 1296143788);
        call_user_func_array('parent::__construct', $arguments);
    }
    // ...
}
```

kitsunet:
Adds the new configuration setting

Neos.Flow.object.proxyOptimizationLevel

which defaults to 1, as a base line for safe optimizations that will be added in the future. 0 means no optimization and results in the same as before this change. Everything over 1 will increasingly add optimizations that could be breaking in special use cases. See comments in the `Settings.Object.yaml` of Neos.Flow.

Resolves: #1539

**Checklist**

### DIFF
--- a/Neos.Flow/Configuration/Settings.Object.yaml
+++ b/Neos.Flow/Configuration/Settings.Object.yaml
@@ -23,3 +23,8 @@ Neos:
       # by specifying corresponding regular expressions that don't match classes to exclude.
       #
       includeClasses: []
+
+      # 0 means no optimizations get enabled, the higher the number the more optimizations but the more it _could_ be breaking in special use cases.
+      # 1 excludes object serialization helpers for classes that only have simple type properties and have no injections, should be very safe
+      # 2 excludes object serialization helpers for classes that only have simple type properties (but still injections, so configuration injections will get serialized)
+      proxyOptimizationLevel: 1
--- a/Neos.Utility.Arrays/Classes/Arrays.php
+++ b/Neos.Utility.Arrays/Classes/Arrays.php
@@ -199,7 +199,7 @@ abstract class Arrays
      * @return mixed The value found, NULL if the path didn't exist (note there is no way to distinguish between a found NULL value and "path not found")
      * @throws \InvalidArgumentException
      */
-    public static function getValueByPath(array &$array, $path)
+    public static function getValueByPath(array $array, $path)
     {
         if (is_string($path)) {
             $path = explode('.', $path);
--- a/Neos.Flow/Classes/Aop/AspectContainer.php
+++ b/Neos.Flow/Classes/Aop/AspectContainer.php
@@ -12,6 +12,7 @@ namespace Neos\Flow\Aop;
  */
 
 use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Aop\Builder\ClassNameIndex;
 use Neos\Flow\Aop\Pointcut\Pointcut;
 
 /**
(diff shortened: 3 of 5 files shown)


kitsunet:
Draft for now, need to fix tests and want to do some more optimization, but htis works for the described type of classes.

Neos.Flow/Configuration/Settings.Object.yaml
```diff
@@ -23,3 +23,8 @@ Neos:
       # by specifying corresponding regular expressions that don't match classes to exclude.
       #
       includeClasses: []
+
+      # 0 means no optimizations get enabled, the higher the number the more optimizations but the more it _could_ be breaking in special use cases.
+      # 1 excludes object serialization helpers for classes that only have simple type properties and have no injections, should be very safe
```
mhsdesign, line 28: Thanks first for investigating the proxy building ^^

Hmmm this could be real bad for the flow eco system, as when i require a package which uses dtos with `new self` (when it was developed on 1) but the project i require it in is on level 0 it woulndt work.

Neos.Utility.Arrays/Classes/Arrays.php
```diff
@@ -199,7 +199,7 @@ public static function array_reduce(array $array, string $function, $initial = n
      * @return mixed The value found, NULL if the path didn't exist (note there is no way to distinguish between a found NULL value and "path not found")
      * @throws \InvalidArgumentException
      */
-    public static function getValueByPath(array &$array, $path)
+    public static function getValueByPath(array $array, $path)
```
mhsdesign, line 202: hu scary.... btw its a bit odd that we use recursion in this helper ... shouldnt be a foreach with references be faster?

kitsunet: IDK I think it went between different variants over hte years, fact is the reference seems unnecessary here and I hate references, so I removed it, probably should take that out of this change htough...

Neos.Flow/Configuration/Settings.Object.yaml
kitsunet: That is true I guess. We could just decide on optimizations we want and skip the setting, but I think it might be very helpful to introduce more aggressive optimizations that won't work on any code base necessarily.

sorenmalling: I'm having a hard time understanding this - could the documentation be updated to tell about this optimized proxy? 
And is there a real usecase of different levels? @mhsdesign points out a case that could lead to different experience (works on my machine-syndrome)

kitsunet: yeah, I see this will be a difficult sell 😂 I will remove the setting and just put in the respective optimizations I find sensible 😆

mhsdesign:
Btw i lately found out that classes ending with "Exception" are not proxied ^^ wondered always why `new self()` was working here :P

Neos.Flow/Configuration/Settings.Object.yaml
sorenmalling: I love awesome optimization for everybody/all cases 🚀❤️

robertlemke:
@kitsunet what do you think, is this feature still needed, now that #3049 etc. are in place?

kitsunet:
Integrated in the big refactoring, makes more sense, so closing.

PR #2970, opened 2023-03-03
FEATURE: Separate RouteConfiguration from Router

### LINKED ISSUE #2948: BUG: RoutingCommandController injects Router class instead of Interface

### Current Behavior

The `RoutingCommandController` injects the default `Router` implementation. This causes a implementation, implements `RouterInterface` and configures it in `Objects.yaml` does not affect the output of the RoutingCommandController commands.

For example, overiding the default way of the route to gather routes to match, it not displated in the CLI as expected

### Expected Behavior

The CLI tool `routing:list` should inject the interface and let the `Objects.yaml` take care of confiure the implementation.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Flow: All supported versions
- PHP:
```

### Anything else?

_No response_

sorenmalling:
This separates the Routes configuration from the router by introducing a `RoutesProviderInterface` which will be used by the router implementation together with a `ConfigurationRoutesProvider` that implements the current configuration from Routes.yaml. 

Switching out the internal implementation of the `RoutesProviderInterface` can be done via Objects.yaml to add custom behaviour. But be aware that this is not covered by our api promises. All Implementations should include the routes provided by the `ConfigurationRoutesProvider`.

This change also makes sure, that the RouteCommandController uses the current `RoutesProviderInterface` implementation, instead of hard coded Flow router. That ensures that all Routes available to the router are now also visible to route cli-commands.

Fixes: #2948

**Upgrade instructions**

This change removes the methods `getRoutes` and `addRoute` from the Router that previously were mainly used in functional-tests as they were never part of the Router Interface. 

To adjust for that the existing utility `FunctionalTestCase->registerRoute` method has to be used instead of `FunctionalTestCase->router->addRoute`.

The method `Router::setRoutesConfiguration`, which was also previously used for internal testing has been removed without official replacement. You _could_ technically inject a custom routes provider to do so but be aware that this is internal behaviour.

**Review instructions**

Run the ./flow routing:list command - you will see the list as expected

**Checklist**


Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
```diff
@@ -0,0 +1,120 @@
...
+ * @Flow\Scope("singleton")
+ */
+final class RouteConfiguration
+{
+    /**
+     * Array of routes to match against
+     *
+     * @var Routes
+     */
+    private Routes $routes;
```
bwaidelich: Does this work currently?

I think it has to be nullable, since it is not initialized during construction time
```suggestion
    private ?Routes $routes = null;
```

bwaidelich: btw: you can omit the `routesCreated` field and just replace the check in line 51 with
```php
if ($this->routes === null) {
    $this->routes = $this->createRoutesFromConfiguration();
}

File: Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php [hunk 2]
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Creates \Neos\Flow\Mvc\Routing\Route objects from the injected routes
+     * configuration.
+     *
+     * @return void
+     * @throws InvalidRouteSetupException
+     */
+    protected function createRoutesFromConfiguration(): void
```
bwaidelich: I would suggest:
```suggestion
    private function createRoutesFromConfiguration(): Routes
```

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
```diff
@@ -0,0 +1,120 @@
...
+                $route->setHttpMethods($routeConfiguration['httpMethods']);
+            } else {
+                if (isset($routesWithHttpMethodConstraints[$uriPattern]) && $routesWithHttpMethodConstraints[$uriPattern] === true) {
+                    throw new InvalidRouteSetupException(sprintf('There are multiple routes with the uriPattern "%s" and "httpMethods" option set. Please specify accepted HTTP methods for all of these, or adjust the uriPattern', $uriPattern), 1365678432);
+                }
+                $routesWithHttpMethodConstraints[$uriPattern] = false;
+            }
+            $routes[] = $route;
+        }
+        $this->routes = Routes::fromArray($routes);
```
bwaidelich: could then be changed to
```suggestion
        return Routes::fromArray($routes);
```

Neos.Flow/Classes/Command/RoutingCommandController.php
```diff
@@ -40,15 +39,9 @@ class RoutingCommandController extends CommandController
...
+     * @var RouteConfiguration
      */
-    protected $configurationManager;
-
-    /**
-     * @Flow\Inject
-     * @var Router
-     */
-    protected $router;
+    protected $routeConfiguration;
```
bwaidelich: Instead of making the `RouteConfiguration` an explicit concept, what about adding a `Router::getRoutes()`? (not to the `RouterInterface` since the whole CommandController won't work for custom implementations anyways)

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
```diff
@@ -0,0 +1,120 @@
...
+                }
+                $routesWithHttpMethodConstraints[$uriPattern] = true;
+                $route->setHttpMethods($routeConfiguration['httpMethods']);
+            } else {
+                if (isset($routesWithHttpMethodConstraints[$uriPattern]) && $routesWithHttpMethodConstraints[$uriPattern] === true) {
+                    throw new InvalidRouteSetupException(sprintf('There are multiple routes with the uriPattern "%s" and "httpMethods" option set. Please specify accepted HTTP methods for all of these, or adjust the uriPattern', $uriPattern), 1365678432);
+                }
+                $routesWithHttpMethodConstraints[$uriPattern] = false;
+            }
+            $routes[] = $route;
```
bwaidelich: What about moving all that logic to some 
```php
Routes::fromConfiguration(array $configuration)
```
method that calls some
```php
Route::fromConfiguration(array $configuration)
```
method internally?

This way, this class turns into a really simple layer between configuration manager & router

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
bwaidelich: well, with the suggested changes, the whole class could be changed to sth. like
```php
final class RouteConfiguration
{
    private ?Routes $routes = null;

    public function __construct(
        private readonly ConfigurationManager $configurationManager
    ) {}

    public function getRoutes(): Routes
    {
        if ($this->routes === null) {
            $this->routes = Routes::fromConfiguration($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES));
        }
        return $this->routes;
    }
}
```

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
```diff
@@ -0,0 +1,120 @@
...
+        }
+        $this->routes = Routes::fromArray($routes);
+
+        $this->emitRoutesCreated($this->routes);
```
bwaidelich: I would suggest not to expose the internals and make this class an extension point.
Instead we could add a signal to the `Router` itself and/or introduce a very simple interface like:
```php
interface RouteConfiguration {
    public function getRoutes(): Routes;
}
```

(even though `RoutesProvider` or `RoutesSource` might be a better name)

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -81,6 +53,11 @@ class Router implements RouterInterface
      */
     protected $lastResolvedRoute;
 
+    public function __construct(
+        private readonly Routes $routes
```
bwaidelich: OK, this is a flaw I'm afraid.
The whole point of the configuration manager injection was to _avoid having to parse the routes for every request_.
The router would only build the route objects if it comes across a route it hasn't seen (i.e. cached) before.
By injecting the parsed routes, we defeat that purpose and this will be a _huge_ performance hit.


But I think we can work around this by adding one layer of abstraction, with the mentioned interface that defers loading & parsing of the routes until requested.

Neos.Flow/Tests/Unit/Mvc/Routing/RouterTest.php
```diff
@@ -69,7 +70,7 @@ class RouterTest extends UnitTestCase
      */
     protected function setUp(): void
     {
-        $this->router = $this->getAccessibleMock(Router::class, ['dummy']);
+        $this->router = $this->getMockBuilder(Router::class)->setConstructorArgs([Routes::empty()])->getMock();
```
bwaidelich: I think you have removed some important tests

bwaidelich (dismissed):
I still like the direction, but I stumbled upon a major issue: This change removes the lazy loading & parsing of routes

(see https://github.com/neos/flow-development-collection/pull/2970#discussion_r1124244940)

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
mhsdesign:
```php
final class RouteConfiguration
{
    private ?Routes $routes = null;

    public function __construct(
        private readonly ConfigurationManager $configurationManager
    ) {}

    public function getRoutes(): Routes
    {
        return $this->routes ??= Routes::fromConfiguration($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES));
    }
}
```

lets go

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
mhsdesign: fun fact it doesnt have to be null able (as long as you use isset() or `??` to check on it)

Neos.Flow/Classes/Mvc/Routing/Routes.php
```diff
@@ -0,0 +1,52 @@
...
+final class Routes implements \IteratorAggregate
+{
+    /**
+     * @var array<int, Route>
+     */
+    private readonly array $routes;
+
+    private function __construct(
+        array $routes
+    ) {
```
mhsdesign, line 20:
```suggestion
    private function __construct(
        Route ... $routes
    ) {
```
for strict typing (calling side now needs adjustments aswell (`new self(...$routesArray)`) )

Neos.Flow/Classes/Mvc/Routing/Routes.php
```diff
@@ -0,0 +1,52 @@
...
+        array_unshift($routes, $route);
+        return self::fromArray($routes);
+    }
+
+    public function append(Route $route): self
+    {
+        $routes = $this->routes;
+        $routes[] = $route;
+        return self::fromArray($routes);
+    }
```
mhsdesign, line 46:
```suggestion
    public function prepend(Route $route): self
    {
        return new self($route, ...$this->routes);
    }

    public function append(Route $route): self
    {
        return new self(...$this->routes, $route);
    }
```

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
bwaidelich: True, it could even be simplified to a oneliner
https://3v4l.org/JBB4r#v8.1.9

Neos.Flow/Classes/Mvc/Routing/Routes.php
mhsdesign: Variadic `Route ...$routes` for strict runtime types

```suggestion
    private function __construct(
        Route ...$routes
    ) {
        $this->routes = $routes;
    }

    public static function fromArray(array $routes): self
    {
        return new self(...$routes);
    }

    public static function empty(): self
    {
        return new self();
    }

    public function prepend(Route $route): self
    {
        return new self($route, ...$this->routes);
    }

    public function append(Route $route): self
    {
        return new self(...$this->routes, $route);
    }
```

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
mhsdesign: exactly what i suggested here: https://github.com/neos/flow-development-collection/pull/2970#discussion_r1124256695 :P

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
sorenmalling: 🙈 evil oneliners 😄

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
sorenmalling: To clarify: The extension point, is to take a package like this

https://github.com/wbsply/route-annotation

and give it a point to hook into the configuration. In package above, I override the RouterInterface implementation to add annotated routes. 

What you would suggest, is to give the extension possibility to the `Router` class? 

@bwaidelich  Can you write a scratch file of your idea and I can adapt the PR :)

Neos.Flow/Classes/Mvc/Routing/Routes.php
sorenmalling: Love it!

Neos.Flow/Classes/Mvc/Routing/RouteConfiguration.php
sorenmalling: It does not return anything

> I still like the direction, but I stumbled upon a major issue: This change removes the lazy loading & parsing of routes

Please but some more words to it, I really want to understand it, but don't at the moment 🙃

bwaidelich:
@sorenmalling here's my suggestion briefly scratched out:

```php
interface RoutesProvider
{
    public function getRoutes(): Routes;
}

final class ConfigurationRoutesProvider implements RoutesProvider
{
    private ?Routes $routes = null;

    public function __construct(
        private readonly ConfigurationManager $configurationManager
    ) {}

    public function getRoutes(): Routes
    {
        return $this->routes ??= Routes::fromConfiguration($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES));
    }
}

final class Router implements RouterInterface
{
    public function __construct(
        private readonly RoutesProviderInterface $routesProvider,
    ) {}

    // ...
}
```

The extension point would be a custom implementation of the `RoutesProviderInterface`, e.g.:

```php
final class MyRoutesProvider implements RoutesProvider
{
    private ?Routes $routes = null;

    public function __construct(
        private readonly ConfigurationRoutesProvider $configurationRoutesProvider,
    ) {}

    public function getRoutes(): Routes
    {
        if ($this->routes === null) {
            $this->routes = $this->appendCustomRoutes($this->configurationRoutesProvider->getRoutes());
        }
        return $this->>routes;
    }

    private function appendCustomRoutes(Routes $routes): Routes
    {
         // load routes from annotations etc.
    }
}
```

sorenmalling:
> The extension point would be a custom implementation of the `RoutesProviderInterface`, e.g.:

How about we put the extension point into the RoutesLoader class?

Routes is a configuration, the Routes have been loaded, now extend that - we avoid having a new Provider concept, that requires overriding (which will lead to dropping the Yaml configurated routes, right?)

sorenmalling:
@bwaidelich @mhsdesign Tried my best to implement suggestion and to address the injection issue. Will you re-review that concept of `fromConfiguration` implemented now?

Neos.Flow/Classes/Mvc/Routing/Routes.php
```diff
@@ -16,33 +18,47 @@ final class Routes implements \IteratorAggregate
...
+                }
+                $routesWithHttpMethodConstraints[$uriPattern] = true;
+                $route->setHttpMethods($routeConfiguration['httpMethods']);
+            } else {
+                if (isset($routesWithHttpMethodConstraints[$uriPattern]) && $routesWithHttpMethodConstraints[$uriPattern] === true) {
+                    throw new InvalidRouteSetupException(sprintf('There are multiple routes with the uriPattern "%s" and "httpMethods" option set. Please specify accepted HTTP methods for all of these, or adjust the uriPattern', $uriPattern), 1365678432);
+                }
+                $routesWithHttpMethodConstraints[$uriPattern] = false;
+            }
+            $routes->append($route);
```
mhsdesign: this cant work right? as this is immutable ;)

either `$routes = $routes->append($route);`

or create a separate array and do `new self(...$routes)` at the end

mhsdesign (dismissed):

Neos.Flow/Classes/Mvc/Routing/Routes.php
sorenmalling: True

mficzel:
FYI: this needs a rebase after #2925 is merged

Neos.Flow/Classes/Command/RoutingCommandController.php
```diff
@@ -100,7 +95,7 @@ public function listCommand(): void
     public function showCommand(int $index): void
     {
         /** @var Route[] $routes */
-        $routes = $this->router->getRoutes();
+        $routes = Routes::fromConfiguration($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES));
```
mhsdesign: we can use flow inject https://github.com/neos/flow-development-collection/blob/a77393a42c45c2456124ded78fea54d0727613a1/Neos.Flow/Classes/Annotations/InjectConfiguration.php#L59
or objects yaml pick as you fit

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -151,29 +119,6 @@ public function getLastMatchedRoute()
...
-        return $this->routes;
-    }
-
-    /**
-     * Manually adds a route to the beginning of the configured routes
-     *
-     * @param Route $route
-     * @return void
-     */
-    public function addRoute(Route $route)
```
mhsdesign, line 172: this can be considered breaking

mhsdesign:
Hi ;) 

we just looked at your pr on the sprint and so far we really like it. As rm im wondering if youd like to finish this for 8.3

Neos.Flow/Classes/Mvc/Routing/Router.php
sorenmalling: Nope - because it was never a part of the RouterInterface

sorenmalling:
@bwaidelich Since the time of writing https://github.com/neos/flow-development-collection/pull/2970/#issuecomment-1453449187 I seem to understand what your point/idea with that Provider was.

Implemented a basic provider for now, adding a extension point for others

Neos.Flow/Classes/Command/RoutingCommandController.php
```diff
@@ -73,8 +66,9 @@ public function listCommand(): void
     {
         $this->outputLine('<b>Currently registered routes:</b>');
         $rows = [];
+        $routes = $this->routesProvider->getRoutes();
         /** @var Route $route */
-        foreach ($this->router->getRoutes() as $index => $route) {
+        foreach ($routes as $index => $route) {
```
bwaidelich: Why a separate variable?
This could be `foreach ($this->routesProvider->getRoutes() as $index => $route)`

This will spare one assigned variable. Besides the `@var` annotation should be obsolete now if `Routes` is typed correctly to return `array<int, Route>`

Neos.Flow/Classes/Command/RoutingCommandController.php
```diff
@@ -203,8 +197,10 @@ public function resolveCommand(string $package, string $controller = null, strin
         /** @var Route|null $resolvedRoute */
         $resolvedRoute = null;
         $resolvedRouteNumber = 0;
+
+        $routes = $this->routesProvider->getRoutes();
         /** @var int $index */
-        foreach ($this->router->getRoutes() as $index => $route) {
+        foreach ($routes as $index => $route) {
```
bwaidelich: same here

Neos.Flow/Classes/Command/RoutingCommandController.php
```diff
@@ -292,8 +288,9 @@ public function matchCommand(string $uri, string $method = null, string $paramet
         /** @var Route|null $matchedRoute */
         $matchedRoute = null;
         $matchedRouteNumber = 0;
+        $routes = $this->routesProvider->getRoutes();
         /** @var int $index */
-        foreach ($this->router->getRoutes() as $index => $route) {
+        foreach ($routes as $index => $route) {
```
bwaidelich: and here :)

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -124,11 +90,12 @@ public function route(RouteContext $routeContext): array
...
             return $cachedRouteResult;
         }
-        $this->createRoutesFromConfiguration();
+
         $httpRequest = $routeContext->getHttpRequest();
 
+        $routes = $this->routesProvider->getRoutes();
         /** @var $route Route */
-        foreach ($this->routes as $route) {
+        foreach ($routes as $route) {
```
bwaidelich: here as well

Neos.Flow/Classes/Mvc/Routing/Router.php
```diff
@@ -194,10 +138,10 @@ public function resolve(ResolveContext $resolveContext): UriInterface
             return $cachedResolvedUriConstraints->applyTo($resolveContext->getBaseUri(), $resolveContext->isForceAbsoluteUri());
         }
 
-        $this->createRoutesFromConfiguration();
+        $routes = $this->routesProvider->getRoutes();
 
         /** @var $route Route */
-        foreach ($this->routes as $route) {
+        foreach ($routes as $route) {
```
bwaidelich: same same

Neos.Flow/Classes/Mvc/Routing/Routes.php
```diff
@@ -0,0 +1,68 @@
...
+    public function prepend(Route $route): self
+    {
+        return new self($route, ...$this->routes);
+    }
+    public function append(Route $route): self
+    {
+        return new self(...$this->routes, $route);
+    }
+
+    public function getIterator(): Traversable
```
bwaidelich, line 71: With this annotation the thing should become more type aware:
```suggestion
    /**
     * @return Traversable<int, Route>
     **/
    public function getIterator(): Traversable
```

(it might also require `@return Traversable|array<int, Route>` for IDEs to pick it up correctly)

Neos.Flow/Classes/Mvc/Routing/Routes.php
```diff
@@ -0,0 +1,68 @@
...
+                $routesWithHttpMethodConstraints[$uriPattern] = false;
+            }
+            $routes = $routes->append($route);
+        }
+    }
+
+    public static function empty(): self
+    {
+        return new self();
+    }
```
bwaidelich, line 56:
```suggestion
    }

```

Neos.Flow/Classes/Mvc/Routing/Routes.php
```diff
@@ -0,0 +1,68 @@
...
+    public function prepend(Route $route): self
+    {
+        return new self($route, ...$this->routes);
+    }
```
bwaidelich, line 61:
```suggestion
    }

```

Neos.Flow/Classes/Mvc/Routing/RoutesProviderInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Neos\Flow\Mvc\Routing;
+
```
bwaidelich, line 6: Some doc comments would be nice

bwaidelich (commented):
Thanks, this is exactly what I had in mind!
Just a few - mostly nitpicky - comments and it would be nice if you could update the description of this PR to document this more (so that it can be automatically picked up by the changelog)

Also we should probably add a phrase or two to the documentation

Neos.Flow/Classes/Mvc/Routing/Routes.php
```diff
@@ -0,0 +1,68 @@
...
+        $routes = self::empty();
+        $routesWithHttpMethodConstraints = [];
+        foreach ($configuration as $routeConfiguration) {
+            $route = Route::fromConfiguration($routeConfiguration);
+
+            $uriPattern = $route->getUriPattern();
+
+            if (isset($routeConfiguration['httpMethods'])) {
+                if (isset($routesWithHttpMethodConstraints[$uriPattern]) && $routesWithHttpMethodConstraints[$uriPattern] === false) {
+                    throw new InvalidRouteSetupException(sprintf('There are multiple routes with the uriPattern "%s" and "httpMethods" option set. Please specify accepted HTTP methods for all of these, or adjust the uriPattern', $uriPattern), 1365678427);
```
mhsdesign, line 32: imo validation belongs to the actual constructor, so that these constrains are also validated for `prepend` `append` and any other future methods ;)

sorenmalling: @mhsdesign So adding a `validate` method after the assigning of `$this->routes` in the private constructor?

bwaidelich: > imo validation belongs to the actual constructor

Oh yes, that's a very good point!

I'd say something along the lines of:

```php
    private function __construct(
        Route ...$routes
    ) {
        $routesWithHttpMethodConstraints = [];
        foreach ($this->routes as $route) {
            $uriPattern = $route->getUriPattern();
            if ($route->hasHttpMethodConstraints()) {
                if ($routesWithHttpMethodConstraints[$uriPattern] ?? null === false) {
                    throw new InvalidRouteSetupException(sprintf('There are multiple routes with the uriPattern "%s" and "httpMethods" option set. Please specify accepted HTTP methods for all of these, or adjust the uriPattern', $uriPattern), 1365678427);
                }
                $routesWithHttpMethodConstraints[$uriPattern] = true;
        } else {
            if ($routesWithHttpMethodConstraints[$uriPattern] ?? null === true) {
                throw new InvalidRouteSetupException(sprintf('There are multiple routes with the uriPattern "%s" and "httpMethods" option set. Please specify accepted HTTP methods for all of these, or adjust the uriPattern', $uriPattern), 1365678432);
            }
            $routesWithHttpMethodConstraints[$uriPattern] = false;
        }
    }
```

And then change this method to just:
```php
    public static function fromConfiguration(array $configuration): self
    {
        $routes = [];
        foreach ($configuration as $routeConfiguration) {
            $routes[] = Route::fromConfiguration($routeConfiguration);
        }
        return new static(...$routes);
    }
```

(untested)

mhsdesign: Exactly what i had in mind ❤️

mhsdesign:
Fyi the currently removed method `$this->router->addRoute($route);` is used in test :(

in those test wed need to mock the routeProvider to only return the wanted route ...

mficzel:
Am trying to adjust this pr to get this finally ready and I have a question. 

As i understand this is a prerequisite for configuring routes via annotation. However i am not sure how this is supposed to be done. Is the plan to create a MetaRoutesProvider that combines the ConfigurationRoutesProvider with a future AnnotationRoutesProvider or is there a clever trick in here i have missed. The signals that are mentioned in the pr description have been removed in the meantime.

I would love to see annotation routes + policies as a framework feature.

mhsdesign:
> Is the plan to create a MetaRoutesProvider that combines the ConfigurationRoutesProvider with a future AnnotationRoutesProvider or is there a clever trick in here i have missed. The signals that are mentioned in the pr description have been removed in the meantime.

yes something along the lines. The interface `RoutesProviderInterface` might be implemented by an `DelegatingRoutesProvider` which internally delegates to the configuration based source and the annotation source. (maybe the sub providers could even be set by configuration)

mficzel:
@mhsdesign I see two waaknesses in the current approach i would like to address as well so the planned follow ups can build the features they want.

1. I personally would like to see extensibility beyond objects.yaml built in if we refactor this part (that was the intention of the signal at first). Maybe we have a list of RouteProviders configured via Settings that can be extended (think middlewares) or something similar that fits flow-framework. That would also define the order of routeProviders and thus routes.

2. We currently loose the @position information (currently only between package routes) as far as i see and it would be impossible for a secondary provider to put routes in between routes from other providers). Not sure that is something we do want but we could either preserve the @position information and do the sorting at the end or allow to insert Routes at specific places in between and pass a routes object around currently there is only prepend and append. That would be more like a `RoutesProcessor::process(Routes $routes): Routes' than a routes provider. 

As i just start to dig deeper it is possible you already thought about that and how to address that.

mhsdesign:
Hi @mficzel

indeed i did discuss point 1 with basti once. I had the same opinion as you but i was convinced that this pr should mostly be considered a cleanup pr for disentangling router from config. Overriding the provider interface should not be considered api as of this stage. Experimental packages like https://github.com/wbsply/route-annotation (to configure routes via annotations) could of course use this internal api to easier hack into the system but this is at this stage not recommended - and even less recommended would be to have two packages providing other route configuration sources.

> Note that a middleware like approach is already doable by not replacing the interface via objects yaml, but by using flows aop to `@Flow\Around` the interface `getRoutes` which will be enhanced by the package's provider. That way - theoretically - multiple packages could ship their routing providers. (Which is still not a good idea :P)

About 2. We didnt think of this really. And thats probably mostly because we didnt consider the api really a 100% perfect extensibility point.

I think that this change alone is currently quite polished and should go directly into the core. Any further adjustments to make the api actually fool proof and usable should be discussed and implemented in another pr as this alone is a super valuable cleanup.

mficzel:
@mhsdesign almost finished is an interesting state ... this state here simply does not work yet after the last refactorings anymore.

Regarding the extension points. For experimental package i agree that it is fine to have an extension point to replace the whole configuration via obejcts.yaml. However as far as i understand we agreed that policies and routes via annotations would be a very valuable addition and we mainly discussed the how an not the if. That is why i would like to have an understanding what kind of extension point we would want to use ourself in future. Just to avoid optimizing in the wrong direction.

If we would place this directly to the ConfigurationRoutesProvider that would be fine with me but since that name suggests something different i would expect a second AnnotationRoutesProvider.

mhsdesign:
> almost finished is an interesting state ... this state here simply does not work yet after the last refactorings anymore.

haha okay fair enough - i was just defending this pr ^^

> Just to avoid optimizing in the wrong direction.

That makes also sense as we dont want to put in to much effort into this if we know we have to rehaul it again.

> If we would place this directly to the ConfigurationRoutesProvider that would be fine with me but since that name suggests something different i would expect a second AnnotationRoutesProvider.

I agree - it would be cleaner to have them split up.

--------

I would like to have your input @bwaidelich as well on this matter or maybe we can discuss this at times in slack ;)

mficzel:
As i see it this pr now only is a cleanup of the route configuration (wich is fine) but it does not solve the original goal of making the route configuration extensible to later implement attribute/annotation based routes.

mficzel:
@sorenmalling @bwaidelich @mhsdesign I finally got the code and the tests working which was not that easy as functional tests did quite a lot of $router->addRoute and $router->getRoutes. 

relevant adjustments to the code i made:
- Support php 8.0 as this has Neos 8.x as target (maybe there will be a 8.4) so no public readonly properties here
- `Routes->append` and. `Routes->prepend` are renamed to `Routes->withAppendedRoute` / `Routes->withPrependedRoute` to ensure the immutability is clear to the outside.
- cache lifetime and tags via Routes.yaml from as implemented in #2925 are merged from 8.3

If we consider the adjustments to the router breaky the target branch should be adjusted to 9.0.

Neos.Flow/Tests/FunctionalTestCase.php
```diff
@@ -369,7 +379,14 @@ protected function registerRoute($name, $uriPattern, array $defaults, $appendExc
         if ($httpMethods !== null) {
             $route->setHttpMethods($httpMethods);
         }
-        $this->router->addRoute($route);
+
+        // the old router->addRoute prepended the routes so this is needed for consistency
+        $this->routes = $this->routes->withPrependedRoute($route);
```
mhsdesign: the register route will now always clear the default routes as the default provider from the testing context is not used. I imagined that we would merge the runtime registered routes with the actual.

And a possibly way to archive this is to have a `TestingRoutesProvider` with a static $routes cache ... we of course have to reset it on `shutDown`

mhsdesign:
wow thank you so much @mficzel ;)

Neos.Flow/Tests/FunctionalTestCase.php
mficzel: Not really sure what "default" routes for a functional testcase would be. To me it totally makes sense that the router for functional testcases  only knows the routes he is explicitly told about.

mhsdesign: i have written and seen tests where the router is required to know the full routing configuration. This is essential for e2e testing a route for example with flows testing browser.

mhsdesign: if you like i could implement my suggestion

mficzel: I see use for a method like `registerDefaultRoutes` in the FunctionalTestcase.php that uses the RoutesProvider to calculate the routes. That would be called explicitly in e2e tests. In other cases i see this as a very strange side effect.

mhsdesign: This is funny. Wellllll.... I am pretty sure that the behaviour as i described it is intended and wanted, but on the other hand i was surprised that the tests dont fail. I found a clue. I searched for a registered test route in `Testing/Routes.yaml` and found `neos/flow/test/http/foo` for example. This is referenced and tested in the `Neos.Flow/Tests/Functional/Mvc/RoutingTest.php`.

When running the test it actually doesnt fail but is rather skipped :D
This is caused by a check in https://github.com/neos/flow-development-collection/blob/86d95b38bd618ca68e6dd8d9c400367ccef2f232/Neos.Flow/Tests/Functional/Mvc/RoutingTest.php#L49-L61 which tries to avoid making the test fail if it is executed in a client project i suppose.

e.g without this: https://github.com/neos/flow-development-collection/blob/86d95b38bd618ca68e6dd8d9c400367ccef2f232/.github/workflows/build.yml#L201C1-L203C36

Neos.Flow/Classes/Mvc/Routing/Routes.php
mhsdesign: Adjusted ;)

Neos.Flow/Tests/FunctionalTestCase.php
mhsdesign: Yes i can confirm that with my adjustments the tests are working again.

mhsdesign:
I solved all open discussions and code suggestions, introduced a TestingRoutesProvider and targeted 9.0 this is now nearly finished.

Neos.Flow/Tests/Functional/Mvc/RoutingTest.php
```diff
@@ -382,37 +375,4 @@ public function uriPathPrefixIsRespectedInRoute()
...
-            ]
-        ];
-        $this->router->setRoutesConfiguration($routesConfiguration);
-        $baseUri = new Uri('http://localhost');
-        $actualResult = $this->router->resolve(new ResolveContext($baseUri, $routeValues, false, '', RouteParameters::createEmpty()));
-        self::assertSame('/custom/uri/pattern', (string)$actualResult);
-
-        // reset router configuration for following tests
-        $this->router->setRoutesConfiguration(null);
-    }
```
mhsdesign: this test must be reenabled

Neos.Flow/Tests/Unit/Mvc/Routing/RouterTest.php
```diff
@@ -99,101 +102,13 @@ protected function setUp(): void
         $this->mockActionRequest = $this->getMockBuilder(ActionRequest::class)->disableOriginalConstructor()->getMock();
     }
 
-    /**
-     * @test
-     */
-    public function resolveCallsCreateRoutesFromConfiguration()
```
mhsdesign, line 105: we should carefully review if not too many testcases were removed here and if they were copied to the routing provider tests

Neos.Flow/Tests/Unit/Mvc/Routing/RoutesTest.php
```diff
@@ -0,0 +1,89 @@
...
+                'name' => 'Route 1',
+                'uriPattern' => 'route1/{@package}/{@controller}/{@action}(.{@format})',
+                'defaults' => ['@format' => 'html']
+            ],
+            [
+                'name' => 'Route 2',
+                'uriPattern' => 'route2/{@package}/{@controller}/{@action}(.{@format})',
+                'defaults' => ['@format' => 'html'],
+                'appendExceedingArguments' => true,
+                'cache' => ['lifetime' => 10000, 'tags' => ['foo', 'bar']]
```
mhsdesign: not all fields are unit tested like "httpMethods"

Neos.Flow/Tests/Unit/Mvc/Routing/RouterTest.php
```diff
@@ -454,85 +351,12 @@ public function routeSetsLastMatchedRoute()
...
-        $router->route($routeContext);
+        $this->router->route($routeContext);
 
-        self::assertSame($mockRoute2, $router->getLastMatchedRoute());
-    }
-
-    /**
-     * @test
-     */
-    public function routeLoadsRoutesConfigurationFromConfigurationManagerIfNotSetExplicitly()
```
mhsdesign, line 467: these tests have been removed because `setRoutesConfiguration` was removed.

Neos.Flow/Tests/Unit/Mvc/Routing/RouterTest.php
mhsdesign: The tests were either obsolete, testing nothing, or moved to ConfigurationProviderTest (new) or RoutesTest also new.

mhsdesign (approved):
As this was a joined effort and i committed as well count me as half of a review :)

Thank you so much @sorenmalling for the initial improvement and @mficzel for making it work after the long pause. (And of course @bwaidelich for the discussions ^^)

Lets merge it to 9.0

mficzel (approved):
The testing route provider could be moved to tests but other than that.

Neos.Flow/Classes/Mvc/Routing/TestingRoutesProvider.php
```diff
@@ -0,0 +1,57 @@
...
+
+use Neos\Flow\Annotations as Flow;
+
+/**
+ * Extends the routing to be able to add custom routes at runtime.
+ *
+ * @internal implementation detail. Please use {@see FunctionalTestCase::registerRoute} instead.
+ */
+#[Flow\Scope("singleton")]
+final class TestingRoutesProvider implements RoutesProviderInterface
```
mhsdesign, line 24: as discussed with @mficzel this class resides here correctly.

`Neos\Flow\Tests` has no proxy building and the `TestingPrivilegeManager` also lives in `Classes`.

mhsdesign:
Fyi phpstan 3 catched a bug in `routing:show`, but that will be fixed here:

https://github.com/neos/flow-development-collection/pull/3261

PR #2972, opened 2023-03-05
FEATURE: Add TypeConverters to convert `BackedEnum` back and forth to `string|int`

mficzel:
The change adds Enum Support to Flow property mapping which allows to use Enums as Node properties in Neos aswell. As enums will likely be used with the SelectBoxEditor a change on the Neos-Ui side is required aswell to support selecting the different cases of an enum on via select box.

The implementation checks the existence for the "BackedEnum" class first as this is only available from PHP 8.1.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Property/TypeConverter/BackedEnumToIntegerConverter.php
+++ b/Neos.Flow/Classes/Property/TypeConverter/BackedEnumToIntegerConverter.php
@@ -1,0 +1,57 @@
+<?php
+declare(strict_types=1);
+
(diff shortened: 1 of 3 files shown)


mficzel:
@mhsdesign you see a chance for this to make it into 8.3 … will probably need some support on the ui side but could enable Enum node properties.

mficzel:
Will use this in a custom package rather than adding this to core

PR #2976, opened 2023-03-08
Feature: Refactor EelUtility & DefaultContext Syntax (make ProtectedContextInterface optional)

Jan3k3y:
**Upgrade instructions**

**Review instructions**

**Checklist**


Neos.Eel/Classes/Utility.php
```diff
@@ -61,50 +76,70 @@ public static function getDefaultContextVariables(array $configuration)
...
+            $allowed = [...$allowed, ...$defaultContextEntry->getAllowedMethods()];
+            $defaultContextVariables = Arrays::setValueByPath(
+                $defaultContextVariables,
+                $defaultContextEntry->paths,
+                $defaultContextEntry->toContextValue()
             );
-        };
+        }
+
+        $defaultContext = new ProtectedContext($defaultContextVariables);
```
mhsdesign: or 

```php
ProtectedContext::fromDefaultContextConfiguration(DefaultContextConfiguration $defaultContext): self
```
 as api?

Neos.Eel/Classes/Utility/EelHelperDefaultContextEntry.php
```diff
@@ -0,0 +1,91 @@
...
+        $isHelperContextAware = array_search(ProtectedContextAwareInterface::class, class_implements($className), true);
+        if ($isHelperContextAware && $allowedMethods !== []) {
+            throw new \DomainException(
+                "EEL Helper '$className' should not implement ProtectedContextAwareInterface and have allowedMethods configured.",
+                1678353296292
+            );
+        }
+        if (!$isHelperContextAware && $allowedMethods === []) {
+            throw new \DomainException(
+                "Plain Helper '$className' should have allowedMethods or ProtectedContextAwareInterface configured.",
```
mhsdesign, line 44: This is the only breaking thing if you will ^^

Neos.Eel/Tests/Unit/Utility/LegacyUtilityTest.php
```diff
@@ -0,0 +1,186 @@
...
+            ],
+            "Foo" => [
+                "Bar" => [
+                    "Example" => [
+                        "className" => ExampleHelper::class,
+                        "allowedMethods" => "*"
+                    ],
+                ]
+            ],
+            "__internalLegacyConfig" => [
```
mhsdesign, line 129: or 
"__internalLegacyProtectedContextAwareHelper"

mhsdesign:
instead of `allowedMethods` per helper we might want to consider introducing a generall section for allowedMethods where we can allow certain methods of certain classes selectively:

current idea of this pr:
```yaml
defaultContext:
  Example:
    className: Foo\\Bar\\ExampleHelper
    allowedMethods: "*"
  
```

even better?
```yaml

defaultContext:
  # general section of helpers ...
  Example: Foo\\Bar\\ExampleHelper
  
  # new "special key" we can also prefix with @ or sth ...
  allowedMethods:
    # the equivalent of the above
    Foo\\Bar\\ExampleHelper: "*"

    # allow all helpers of a namespace
    Foo\\Bar\\.*Helper: "*"

   # or even allow basic "safe objects"

   Psr\\Http\\Message\\Uri: "with*"

   # or even go as far as allowing all with calls starting with with (considering they might be immutable like get)
   *: "with*"
```

... just food for thought what do you say @kitsunet 



<details><summary>My brain dump</summary>
<p>

```php
$config = [
    "Example" => ExampleHelper::class,
    "allowedMethods" => [
        ExampleHelper::class => "*"
    ]
];

$allowed = AllowedMethods::fromConfiguration($config["allowedMethods"]);

$allowed->isAllowed(new ExampleHelper());


// so either foreach entry and do an instance of check, con: what if the specific implementation has different config than the interface ... this would need to be merded
// or get all implementations for interface and all extended classes and build an complete map for this
// or find a way of attaching this attribute via objectmanager instead of eel specific?
```

</p>
</details>

kitsunet:
I do like the idea of the separate config quite a lot, it could open up making configurations for objects retrieved from helpers without them being context objects themselves. That would be pretty neat.

PR #2991, opened 2023-03-18
TASK: Use union types in ObjectAccess::setProperty and remove exception

sorenmalling:
This change makes use of union types from PHP8 in the ObjectAccess::setProperty method.

It removes superfluous condition and exceptions testing for argument value types.

**Checklist**

- [] Tests have been created, run and adjusted as needed


mficzel:
Like the change in general. Tests should be green ofc.

Consider this non-breaky especially since still \InvalidArgumentExceptions will be thrown.

sorenmalling:
> Consider this non-breaky especially since still \InvalidArgumentExceptions will be thrown.

Where do you find that InvalidArgumentException is still thrown?

The test is changed to support TypeError, as this is the semnatic error that PHP throws

mficzel:
InvalidArgumentException is thrown by php once arguments do not match the specified type. Maybe I got this wrong.

Generally I would consider it non breaky as long as the same conditions lead to an exception. The exact type of the exception imho would not define a breaking change.

mhsdesign:
Yes I agree it would be pretty hacky and sorry stupid to rely on an exception in this case, as it’s way easier to check before hand.

And yes it will be a TypeError but it doesn’t really matter ;)

mhsdesign:
Hopefully fixed everything and targeted 9.0 :D

PR #2997, opened 2023-03-24
TASK: Use route parameters in MVC Dispatcher

sorenmalling:
Use the RouteParameters in the MVC Dispatcher to determine if there is a controller given and simplify the responsiblity of `resolveController`

### DIFF
--- a/Neos.Flow/Classes/Mvc/Dispatcher.php
+++ b/Neos.Flow/Classes/Mvc/Dispatcher.php
@@ -13,11 +13,13 @@ namespace Neos\Flow\Mvc;
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Configuration\Exception\NoSuchOptionException;
+use Neos\Flow\Http\ServerRequestAttributes;
 use Neos\Flow\Log\PsrLoggerFactoryInterface;
 use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\Mvc\Controller\ControllerInterface;
 use Neos\Flow\Mvc\Controller\Exception\InvalidControllerException;
 use Neos\Flow\Mvc\Exception\InfiniteLoopException;
+use Neos\Flow\Mvc\Exception\NoMatchingRouteException;
 use Neos\Flow\Mvc\Exception\StopActionException;
 use Neos\Flow\Mvc\Exception\ForwardException;
 use Neos\Flow\Mvc\Exception\UnsupportedRequestTypeException;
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Mvc/Dispatcher.php
```diff
@@ -96,6 +98,9 @@ public function injectFirewall(FirewallInterface $firewall)
      */
     public function dispatch(ActionRequest $request, ActionResponse $response)
     {
+        if ($request->getHttpRequest()->getAttribute(ServerRequestAttributes::ROUTING_PARAMETERS) === null) {
+            throw new NoMatchingRouteException();
```
sorenmalling, line 102: This part deserve some attention, but this is to conceptualize the idea

bwaidelich (commented):
Can you please clarify which problem this PR is solving?
`$request->getHttpRequest()->getAttribute(ServerRequestAttributes::ROUTING_PARAMETERS)` is no public API and IMO we should carefully consider whether we want to change that

sorenmalling:
> Can you please clarify which problem this PR is solving?

@kitsunet and I had a chat in slack leading to this

image

https://neos-project.slack.com/archives/C04V4C6B0/p1679648986047279

sorenmalling:
> `$request->getHttpRequest()->getAttribute(ServerRequestAttributes::ROUTING_PARAMETERS)` is no public API and IMO we should carefully consider whether we want to change that

What is not a public API in this code example?

bwaidelich:
> @kitsunet and I had a chat in slack leading to this

Thanks for the clarification! I still fail to understand the scenario completely to be honest.

If I create a route like
```yaml
-
  uriPattern: '{@package}/{@controller}'
```

and navigate to `/invalid/controller`, I get a 404 response as I would expect.

With your change applied, I also get a 404. But this time it is thrown _before_ `\Neos\Flow\Security\Authorization\FirewallInterface::blockIllegalRequests()` is invoked – Not sure if this is to be expected.
Also, the controller resolving code is no longer invoked within the dispatch loop – so the behavior of "forwards" might have changed. Maybe to the better, but I don't completely understand the problem yet.

> What is not a public API in this code example?

By default there are three HTTP Middlewares involved in the MVC routing & dispatching:

1. `RoutingMiddleware` invokes `Router::route()` and adds the results to the `routingResults` request attribute
2. `SecurityEntryPointMiddleware` reads this attribute, creates an `ActionRequest` and adds that as `actionRequest` attribute to the request
3. `DispatchMiddleware` reads the `actionRequest` attribute and passes it to the dispatcher

`\Neos\Flow\Mvc\Dispatcher::dispatch()` is the only part in this process that is `@api` – everything before _could_ be replaced (even though that's not very likely I assume)

bwaidelich:
@sorenmalling you closing this without further comments make me think that you didn't agree to my points!? Please feel free to get in touch if you think this needs further discussion! I didn't mean to step on anyones toes

sorenmalling:
@bwaidelich no toes stepped on 🤗

PR #2998, opened 2023-03-27
BUGFIX: Change type hint of returned security logger

sorenmalling:
The type hint for the `$securityLogger` was referring to the `PsrLoggerFactoryInterface` instead of the correct `LoggerInterface`


kdambekalns (approved):

sorenmalling:
This was never really merged but only approved?

kdambekalns:
> This was never really merged but only approved?

Seems like it… And since it's only a "developer hint", nobody noticed it's missing. 🙈 I probably thought you would (could) merge this yourself.

I tried changing the base to 7.3 here in GitHub directly, but that brings more "interesting" changes than expected. Would you adjust this as needed? Otherwise I can do it as well…

kdambekalns:
I created a new PR against 7.3.

PR #3002, opened 2023-04-01
FEATURE: Make base URI configuration obsolete

mhsdesign:
solves #2157

see https://github.com/neos/flow-development-collection/pull/2744#discussion_r1151998706
for our mile long discussion.

will update this pr description, once we want to merge it ^^

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Http/BaseUriProvider.php
+++ b/Neos.Flow/Classes/Http/BaseUriProvider.php
@@ -18,6 +18,7 @@ class BaseUriProvider
     /**
      * THe possibly configured Flow base URI.
      *
+     * @deprecated with Flow 8.3 - use the FLOW_BASE_URI env or HTTP request instead
      * @Flow\InjectConfiguration(package="Neos.Flow", path="http.baseUri")
      * @var string|null
      */
@@ -32,7 +33,7 @@ class BaseUriProvider
     /**
      * Get the configured framework base URI.
      *
-     * @return Uri|null
+     * @deprecated with Flow 8.3 - use the FLOW_BASE_URI env or HTTP request instead
      */
     private function getConfiguredBaseUri(): ?UriInterface
     {
@@ -64,7 +65,8 @@ class BaseUriProvider
 
     /**
      * Gives the best possible base URI with the following priority:
-     * - configured base URI
+     * - environment variable FLOW_BASE_URI
+     * - configured base URI (deprecated with Neos 8.3)
      * - generated base URI from currently active server request
      * - generated base URI from specified $fallbackRequest
      *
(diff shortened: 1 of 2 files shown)


bwaidelich:
@kitsunet wrote (in #2744):

> But how is this different than the setting, apart from making it harder for some folks to use the feature?

And I have to admit, that I have the same feeling when looking at this PR..
Maybe it's just not time to deprecated this setting as we don't seem to have a viable replacement yet..!?

sorenmalling:
Putting in a thought about baseuri(s):

I have specific packages that would address specific (sub)domains - so "admin.domain.tld" to a admin package, "app.domain.tld" to the "application" package.

Gives me thought about: 
 - a baseuri can be relevant down to a specific package - not just for the "distribution".

So what is the real issue that we are hoping to tackle and benefit from (can't find a "final result/wish" of such thougts).

bwaidelich:
> So what is the real issue that we are hoping to tackle and benefit from

Exactly the one you describe: A *global* base URI does not work for Multi-Site-Neos Installations and Flow Apps with dynamic (sub) domains. So it would make sense to replace the mechanism with a more application specific API.

For Neos that's actually already the case – the `baseUri` config is not useful there, because we use the `Domain` entity for this.

sorenmalling:
> So it would make sense to replace the mechanism with a more application specific API.

First idea, a middleware with a "provider" concept like from #2970 with a default implementation, reading from the Settings.yaml and returning a Uri object positioned before any routing resolving, so it can even be taken into account at that part?

bwaidelich:
A middleware will only work for HTTP requests (for which you can mostly use the current URI anyways). There are more discussions in the original issue (#2744)

kitsunet:
I am btw. super happy if we _can_ actually replace the baseUri. I wonder though, after having had a break, should we maybe embrace the idea of having some kind of framework way to know what host to use for building a URL? Basically what neos does with domain objects as a framework concept that you can extend for your app as needed? then one way _could_ still be a configuraiton setting implementation.

bwaidelich:
>  should we maybe embrace the idea of having some kind of framework
>  way to know what host to use for building a URL?

Yes, that would be cool. In fact we already _have_ that thing with the `BaseUriProvider` – The problem is, that this is a global singleton. And if it _wasn't_, we'd have to find ways to make that accessible wherever it could be needed (example: render an email body with absolute links via Fusion on the CLI).

OTOH maybe the base URI itself is not required, but a properly configured URI _builder_.

Maybe the [ActionUriBuilder](https://github.com/neos/flow-development-collection/pull/2744) could become the replacement for the annoying `ControllerContext`.. For example in the Fusion `RuntimeFactory`
```php
public function create($fusionConfiguration, ControllerContext $controllerContext = null)
```
could probably be replaced by something like
```php
public function create(array $fusionConfiguration, ActionUriBuilder $actionUriBuilder)
```

kitsunet:
Sounds good to me, yes!

mhsdesign:
Nice i just found out how we do it in the cli rendering for flowpack decoupled content store

https://github.com/Flowpack/Flowpack.DecoupledContentStore/blob/6adb8e04246f5fc9b8471b656db7ddf131474745/Classes/Transfer/Resource/Target/MultisiteFileSystemSymlinkTarget.php#L16-L38

😭 -> so ideally one should not need to go to such lengths and need such global configurations.

PR #3014, opened 2023-04-10
FEATURE: Multiple shorthand identifiers in one translation id string

sbruggmann:
Refactored to handle multiple shorthand identifiers in one translation id string.
Does match the functionality in https://github.com/neos/neos-ui/pull/3326

And added an optional fallbackLocale to handle needs from https://github.com/neos/neos-development-collection/pull/4180


crydotsnake:
@sbruggmann Could you maybe check the StyleCI issue?

mhsdesign (changes requested):
As i wrote here https://github.com/neos/neos-ui/pull/3326#pullrequestreview-3856796171 i think its not really sensible to implement multi translation resolution with an api as suggested here. During the translation its not known where the provided arguments should go to and also which fallback is specified for which translation id. 
I think it might be too complex to solve here at the core of the translation framework as this should stay relatively simple to use?

PR #3024, opened 2023-04-29
BUGFIX: Allow private ip addresses in trusted proxies

### LINKED ISSUE #3013: BUG: TrustedProxiesMiddleware should not filter private ip ranges

### Current Behavior

Working behind a reverse proxy on intranets don't respect X-FORWARD-FOR.

### Expected Behavior

TrustedProxiesMiddleware should not filter FILTER_FLAG_NO_PRIV_RANGE on line 265 in Neos.Flow/Classes/Http/Middleware/TrustedProxiesMiddleware.php

### Steps To Reproduce

Try to access an Backendserver over (Reverse)Proxy over the intranet.

### Environment

```markdown
- Flow:
- PHP:
```

### Anything else?

_No response_

NamioL:
**Upgrade instructions**

**Review instructions**

Stop filtering private IP addresses from the trusted proxies middleware

Fix: https://github.com/neos/flow-development-collection/issues/3013

**Checklist**


robertlemke:
@NamioL Thanks for your contribution! As discussed during the Contribution Day, I think that this makes sense, as it solves problems typically seen in Docker-based development environments, such as Local Beach or DDEV. Please amend the git commit message and mention what your patch fixes. Also mention "Resolves #3013" in your commit message and please adjust the subject line so that it starts with "BUGFIX:" (without brackets).

For other reviewers:

Without this patch, Trusted Proxies would have to be set to "*". I'm not completely sure, why we did not allow IP addresses from the private range for clients (browsers).

The only situation I can think of where this patch causes issues is, if there are multiple remote addresses provided by the proxy header and the first one being a private IP address. In that case, Flow would previously return use the 2nd (or 3rd …) address as the remote address, because private addresses would be skipped. With this patch, the private address would now be used.

kitsunet:
I would really like to see an end to end testcase and a better description of the real world issues this fixes. This is the _client_ ip address which AFAIK is not used at all in the framework, but only in userland code. IMHO for typical applications a client ip should never ever be from a private range, also for security reasons. This is even documented eg. in the MDN description for `X-Forwarded-For`. 

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#selecting_an_ip_address

fcool (approved):

fcool:
I disaggree. Think of an Web application Firewall in the same network as the webserver. The Firewall will act as reverse proxy - and it's very probable, that it is on the same net, as the server. And this net is perfectly fine to be local.

PR #3031, opened 2023-05-02
TASK: Drop tests checking notices/warnings are not supressed

kdambekalns:
These were added years ago to make sure the suppression of notices and warnings when including cached PHP files does not come back.

Since the way to check for those  will be removed in PhpUnit 10, and does not come back, I decided to completely remove the tests.

**Review instructions**

With PhpUnit 9 the deprecation warning on those tests should go away, and the tests should no longer fail on PhpUnit 10.

**Checklist**

### DIFF
--- a/Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
+++ b/Neos.Cache/Tests/Unit/Backend/FileBackendTest.php
@@ -566,40 +566,6 @@ class FileBackendTest extends BaseTestCase
     /**
      * @test
      */
-    public function requireOnceDoesNotSwallowPhpWarningsOfTheIncludedFile()
-    {
-        $this->expectWarning();
-        $mockCache = $this->createMock(AbstractFrontend::class);
-        $mockCache->expects(self::atLeastOnce())->method('getIdentifier')->will(self::returnValue('UnitTestCache'));
-
-        $backend = $this->prepareDefaultBackend();
-        $backend->setCache($mockCache);
-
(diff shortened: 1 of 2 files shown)


kdambekalns:
Ping…

mhsdesign:
I cant really judge the situation, but if we still have that functionality it feels odd to just remove the tests.

What about building an own assert helper method, based on what phpunit was doing - leveraging the php error handler:

untested:
```php
    public static function expectErrorNumber(callable $callable, int $expectedErrorNumber): void
    {
        $errorHandlerWasCalled = false; # actually use self::expect once to be called or sth
        set_error_handler(
            static function ($errorNumber, $errorString) use(&$errorHandlerWasCalled, $expecteErrorNumber)
            {
                $errorHandlerWasCalled = true;
                self::assertSame($expectedErrorNumber, $errorNumber);
                return;
            }
        );

        $result = $callable();

        restore_error_handler();

        self::assertTrue($errorHandlerWasCalled);

        return $result;
    }
```

bwaidelich:
I agree to @mhsdesign – it would be a shame to lose tests for this important scenario.
Also could you point to the documentation that mentions the removal of this feature from PHPUnit?

kdambekalns:
> I agree to @mhsdesign – it would be a shame to lose tests for this important scenario. Also could you point to the documentation that mentions the removal of this feature from PHPUnit?

See "Test Outcome versus Test Issues" on https://phpunit.de/announcements/phpunit-10.html and the issue https://github.com/sebastianbergmann/phpunit-documentation-english/issues/330

PR #3040, opened 2023-05-08
!!! TASK: Require PHP 8.2

robertlemke:
The minimum requirement for the Flow Framework, including all packages of its distribution, was raised to PHP 8.2.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.0', '8.1', '8.2']
+        php-versions: ['8.2']
         # see https://mariadb.com/kb/en/mariadb-server-release-dates/
         # this should be a current release, e.g. the LTS version
         mariadb-versions: ['10.6']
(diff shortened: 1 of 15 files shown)


fcool:
What about the split repositories like "Utility\MediaTypes" "Utility\Files" and so on?

robertlemke:
> What about the split repositories like "Utility\MediaTypes" "Utility\Files" and so on?

I currently don't plan to change anything there which would require PHP 8.2, so we could leave them as they are, in case the packages are used elsewhere without Flow. On the other hand, that would increase the complexity for maintaining this repository, so maybe bump PHP to 8.2 for all of them, too?

fcool:
I'm sure we will mess up, if we would use different minimal requirements. So I would probably go for changing them all.

If we were going to maintain them differently, we probably needed to set up tests for all of them in all PHP variants, as we are not going to test them together with the main repo any more.

So to save us work, from which probably next to no one would benefit, my opionion is: Let's keep it simple.

And as we are going to raise all their version numbers to 9.0.0 it is absolutely fine to introduce higher dependencies for new Majors.

robertlemke:
@fcool Makes sense – I force-pushed a new commit which raises the PHP requirement for all packages of the distribution.

I'm curious what Psalm will say this time and what to do about potential errors …

robertlemke:
I need to adjust dependencies and code related to TYPO3 Fluid for this to work.

fcool (approved):

kdambekalns (commented):
Post-merge 🎉 from me!

PR #3041, opened 2023-05-09
BUGFIX: Move access to objectAccess of TemplateObjectAccessInterface into getByPath

dlubitz:
... as accessors are not used anymore for variable provider within fluid, starting v2.8.0.

Due to the missing accessors the `objectAccess` of `TemplateObjectAccessInterface` didn't get called anymore, so the result of the `getByPath` method was an object of `FusionPathProxy` instead of an rendered string.

See: 
https://github.com/TYPO3/Fluid/compare/2.7.4...2.8.0#diff-a0aa72aa19d9eb57cdb9a4dcd344c3706d75ae7ca408286f91a846e495b3c766L122
https://github.com/TYPO3/Fluid/compare/2.7.4...2.8.0#diff-a0aa72aa19d9eb57cdb9a4dcd344c3706d75ae7ca408286f91a846e495b3c766L341
https://github.com/TYPO3/Fluid/compare/2.7.4...2.8.0#diff-a0aa72aa19d9eb57cdb9a4dcd344c3706d75ae7ca408286f91a846e495b3c766L312

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php
@@ -11,8 +11,6 @@ namespace Neos\FluidAdaptor\Core\ViewHelper;
  * source code.
  */
 
-use Neos\Utility\Exception\PropertyNotAccessibleException;
-use Neos\Utility\ObjectAccess;
 use Neos\FluidAdaptor\Core\Parser\SyntaxTree\TemplateObjectAccessInterface;
 use TYPO3Fluid\Fluid\Core\Variables\StandardVariableProvider;
 use TYPO3Fluid\Fluid\Core\Variables\VariableProviderInterface;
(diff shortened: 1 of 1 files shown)


kdambekalns (approved):
Looks good, so fine if it works for you… 😇

robertlemke (approved):
Looks good, quickly tested this also with an application using Fluid.

kdambekalns:
As I found out today, this does not work when you use `{image.title}` in Fluid and `image` is a `FusionPathProxy`:

1. `StandardVariableProvider` checks for `ArrayAccess`, `FusionPathProxy` provides that. `offsetExists()` and `offsetGet()`  only check for the offset directly in the partial fusion tree – there we have no title
2. Now `$subject` is an object (our `FusionPathProxy`), but has no matching get/set/is/property
3 Our `TemplateVariableContainer` is getting `null` back, and thus exits in https://github.com/neos/flow-development-collection/blob/9.0/Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php#L37 – that results in `null` again.
4. `title` ends up as `null`

PR #3045, opened 2023-05-14
BUGFIX: Uncached ConfigurationManager

mhsdesign:
### BUGFIX: Proper uncached configurationManager mode

It is purposely not allowed to disable the cache at runtime (when you have a configuration manager at hand)

The usage to create a configuration manager without caching, you need to have your own request handler and boot only this step:

```php
Scripts::initializeConfiguration($this->bootstrap, false);
```

---

### BUGFIX: ConfigurationManager with disabled cache doesn't replace environment variables in setting

@kitsunet and me need this for building https://github.com/neos/setup/pull/59 at super early boot time - pre compile time.

We want to use the config manager without cache, which currently has a bug and you cant really disable the cache unless using reflection to set `temporaryDirectoryPath` to null.

The config manager with disabled cache doesnt replace environment variables in settings.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Configuration/ConfigurationManager.php
+++ b/Neos.Flow/Classes/Configuration/ConfigurationManager.php
@@ -382,6 +382,9 @@ class ConfigurationManager
      */
     protected function loadConfigurationsFromCache(): void
     {
+        if ($this->temporaryDirectoryPath === null) {
+            return;
+        }
         $cachePathAndFilename = $this->constructConfigurationCachePath();
         /** @noinspection UsingInclusionReturnValueInspection */
         $configurations = @include $cachePathAndFilename;
@@ -553,6 +556,7 @@ class ConfigurationManager
      */
     protected function constructConfigurationCachePath(): string
     {
+        assert($this->temporaryDirectoryPath !== null, 'ConfigurationManager::$temporaryDirectoryPath must not be null.');
         $configurationCachePath = $this->temporaryDirectoryPath . 'Configuration/';
         return $configurationCachePath . str_replace('/', '_', (string)$this->context) . 'Configurations.php';
     }
--- a/Neos.Flow/Classes/Configuration/ConfigurationManager.php
+++ b/Neos.Flow/Classes/Configuration/ConfigurationManager.php
@@ -252,7 +252,9 @@ class ConfigurationManager
                 return new ObjectsLoader(new YamlSource());
             case self::CONFIGURATION_PROCESSING_TYPE_POLICY:
                 $policyLoader = new PolicyLoader(new YamlSource());
-                $policyLoader->setTemporaryDirectoryPath($this->temporaryDirectoryPath);
+                if ($this->temporaryDirectoryPath) {
+                    $policyLoader->setTemporaryDirectoryPath($this->temporaryDirectoryPath);
+                }
                 return $policyLoader;
             case self::CONFIGURATION_PROCESSING_TYPE_ROUTES:
                 return new RoutesLoader(new YamlSource(), $this);
--- a/Neos.Flow/Classes/Configuration/ConfigurationManager.php
+++ b/Neos.Flow/Classes/Configuration/ConfigurationManager.php
@@ -154,7 +154,7 @@ class ConfigurationManager
     /**
      * An absolute file path to store configuration caches in. If not set, no cache will be active.
      *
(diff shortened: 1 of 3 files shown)


Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -423,6 +423,11 @@ protected function processConfigurationType(string $configurationType)
             $this->writeConfigurationCacheFile($cachePathAndFilename, $configurationType);
             $this->replaceConfigurationForConfigurationType($configurationType, $cachePathAndFilename);
             @unlink($cachePathAndFilename);
+        } else {
+            // in case the cache is disabled (implicitly, because temporaryDirectoryPath is null)
```
kitsunet, line 438: This was in fact never a planned use case ;)

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -423,6 +423,11 @@ protected function processConfigurationType(string $configurationType)
...
+            // we need to still handle replacing the environment variables like `%FLOW_PATH_ROOT%` in the configuration
+            $configuration = $this->unprocessedConfiguration[$configurationType];
+            $this->configurations[$configurationType] = eval('return ' . $this->replaceVariablesInPhpString(var_export($configuration, true)) . ';');
```
kitsunet, line 441: mmmmmmmmmmmm, ev(i|a)l. Do we really need this for the checks? I guess so 🤔

Neos.Flow/Classes/Configuration/ConfigurationManager.php
mhsdesign:
```
    /**
     * An absolute file path to store configuration caches in. If null no cache will be active.
     *
     * @var string
     */
    protected $temporaryDirectoryPath;
```

this has been there since EVER!

mhsdesign: you build it ^^ more or less accidental: https://github.com/neos/flow-development-collection/commit/85a7f15664cca6dfba7911b1ed8ee34cc2296102

mhsdesign: and later the feature Runtime evaluation of env and constants in Configuration was added https://github.com/neos/flow-development-collection/commit/1ca7951743ac780bfb6edb83435532eafabd2ef3#diff-83398c2914c2ce46b6e55c9821468814652a275d332692492bb6edbbb84607f9R617

kitsunet: Yep, I know, and I _think_ there was some low level use case at that time that required it to work that way. I would rather close that road than open it :D

kitsunet: Although, the more I think about it, why even not, might be useful to have this in the long run, I really don't like it but well...

kitsunet:
Not beautiful, but lets us use the health check really low level and fast. Might be useful in other situations as well...

kitsunet (approved):

mhsdesign:
Last psalm failure is unrelated

Neos.Flow/Classes/Configuration/ConfigurationManager.php
```diff
@@ -152,11 +152,11 @@ class ConfigurationManager
     protected $cacheNeedsUpdate = false;
 
     /**
-     * An absolute file path to store configuration caches in. If null no cache will be active.
+     * An absolute file path to store configuration caches in. If not set, no cache will be active.
      *
-     * @var string
+     * @var ?string
```
fcool: this is no valid php doctype - in doctypes there are only union types - so there it should be `string|null`

fcool (approved):
Found a minor issue in a phpdoc annotation - I would be fine with ignoring it - but even better would be probably to have it "correct"

Neos.Flow/Classes/Configuration/ConfigurationManager.php
mhsdesign: well i didn't know that actually thanks ^^

mhsdesign:
```suggestion
     * @var string|null
```

PR #3047, opened 2023-05-15
TASK: Only add serialization entities code if needed

robertlemke:
Proxy classes created by the Dependency Injection Proxy Class Builder now only contain code related to serialization and deserialization of related entities if needed.

The code is only rendered if one of the following conditions is met:

- The class is annotated with Entity
- The class is annotated with Scope("session")

Despite the previous condition, the code will not be rendered if the following condition is true:

- The class already has a __sleep() method (we assume that the developer wants to take care of serialization themself)

As part of this change, the generated code related to serialization was slightly adjusted for stricter type handling.

related: #1539

**Review instructions**

- try to find an existing application which relies on serialization of related entities, for example a Flow application which uses ORM with relations or uses entities in a session scope.
- remove all caches and then access your application in a browser using the current Flow 9 branch (without this patch)
- create a backup of the Cache/Code/Flow_Object_Classes directory
- switch to a branch with this change, remove all caches and access the application again in a browser
- use a diff tool (e.g. Kaleidoscope) to compare both cache directories to see what is now different
- check if your application still works

### DIFF
--- a/Neos.Flow/Classes/Cache/CacheFactory.php
+++ b/Neos.Flow/Classes/Cache/CacheFactory.php
@@ -57,18 +57,20 @@ class CacheFactory extends \Neos\Cache\CacheFactory
 
     /**
      * @param CacheManager $cacheManager
-     * @Flow\Autowiring(enabled=false)
+     *
+     * @Flow\Autowiring (enabled=false)
      */
-    public function injectCacheManager(CacheManager $cacheManager)
+    public function injectCacheManager(CacheManager $cacheManager): void
     {
         $this->cacheManager = $cacheManager;
     }
 
     /**
      * @param EnvironmentConfiguration $environmentConfiguration
-     * @Flow\Autowiring(enabled=false)
+     *
+     * @Flow\Autowiring (enabled=false)
      */
-    public function injectEnvironmentConfiguration(EnvironmentConfiguration $environmentConfiguration)
+    public function injectEnvironmentConfiguration(EnvironmentConfiguration $environmentConfiguration): void
     {
         $this->environmentConfiguration = $environmentConfiguration;
     }
--- a/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
@@ -97,7 +97,7 @@ class ProxyClassBuilder
             if ($proxyClass === false) {
                 continue;
             }
-            $this->logger?->debug(sprintf('Building dependency injection proxy for "%s"', $className), LogEnvironment::fromMethodName(__METHOD__));
+            $this->logger->debug(sprintf('Building dependency injection proxy for "%s"', $className), LogEnvironment::fromMethodName(__METHOD__));
 
             $constructor = $proxyClass->getConstructor();
             $constructor->addPreParentCallCode($this->buildSetInstanceCode($objectConfiguration));
@@ -184,8 +184,10 @@ class ProxyClassBuilder
             return '';
         }
 
+        $scopeAnnotation = $this->reflectionService->getClassAnnotation($className, Flow\Scope::class);
+
         $doBuildCode =  $this->reflectionService->getClassAnnotation($className, Flow\Entity::class) !== null;
-        $doBuildCode = $doBuildCode || $this->reflectionService->getClassAnnotation($className, Flow\Scope::class)?->value === 'session';
+        $doBuildCode = $doBuildCode || ($scopeAnnotation->value ?? 'prototype') === 'session';
         if ($doBuildCode === false) {
             return '';
         }
@@ -503,9 +505,12 @@ class ProxyClassBuilder
      * @param string $className Name of the class to inject into
      * @param string $propertyName Name of the property to inject
      * @param string $preparedSetterArgument PHP code to use for retrieving the value to inject
-     * @return array PHP code
+     *
+     * @return null|string[] PHP code
+     *
+     * @psalm-return array{0?: string}|null
      */
-    protected function buildSetterInjectionCode($className, $propertyName, $preparedSetterArgument)
+    protected function buildSetterInjectionCode($className, $propertyName, $preparedSetterArgument): array|null
     {
         $setterMethodName = 'inject' . ucfirst($propertyName);
         if ($this->reflectionService->hasMethod($className, $setterMethodName)) {
--- a/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
@@ -13,11 +13,14 @@ namespace Neos\Flow\ObjectManagement\DependencyInjection;
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Configuration\ConfigurationManager;
+use Neos\Flow\Configuration\Exception\InvalidConfigurationTypeException;
+use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\ObjectManagement\CompileTimeObjectManager;
 use Neos\Flow\ObjectManagement\Configuration\Configuration;
 use Neos\Flow\ObjectManagement\Configuration\ConfigurationArgument;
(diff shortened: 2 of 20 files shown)


kitsunet:
I guess I need to find a breaking case now and backport it in the process. I am pretty sure this is too aggressive, as eg. DTOs (from forms for example) with attached entities can happen, that then end up in a session. At that point the chain is broken. But who knows, maybe it's just a theoretical case. I will have a look.

robertlemke:
> I guess I need to find a breaking case now and backport it in the process. I am pretty sure this is too aggressive, as eg. DTOs (from forms for example) with attached entities can happen, that then end up in a session. At that point the chain is broken. But who knows, maybe it's just a theoretical case. I will have a look.

If those rather exotic cases exist, we could introduce an annotation which allows for explicitly turning object serialization on. Or we add a configuration option (similar to what you suggested), which tells Flow to always generate object serialization code, as if this optimization here would not happen. That setting could be turned off in 9.0 but can be turned on for backwards compatibility.

But of course it would be better to have a clearer understanding of possible edge cases, so we can support them automatically, if feasible.

kitsunet:
Yep, hopefully not needed. Does this need any of the other changes you made recently, or is it a standalone change? because the projects I have in mind I cannot just quickly update to 9.0, so I would want to try and apply this to some older flow(s).

robertlemke:
> Yep, hopefully not needed. Does this need any of the other changes you made recently, or is it a standalone change? because the projects I have in mind I cannot just quickly update to 9.0, so I would want to try and apply this to some older flow(s).

It could work as a cherry-pick, because the recent changes in Reflection Service were basically about strict types. But I haven't tried that …

robertlemke:
I'm not sure what to do about all the Psalm complaints. For example, should I really introduce `@psalm-param class-string $className` everywhere? I fear that these corrections would start a row of more complaints, which would go beyond the scope of this PR.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1900,6 +1925,10 @@ public function freezePackageReflection($packageKey): void
 
     /**
      * Filter an array of entries where keys are class names by being in the given package namespace.
+     *
+     * @param (int|string) $packageKey
```
kdambekalns: Where the parentheses suggested by PhpStorm? I'd write it in "plain PHP syntax"…

```suggestion
     * @param int|string $packageKey
```

Same for the other places where this is used (reading the diff "backwards", so…)

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1725,6 +1742,10 @@ protected function forgetChangedClasses(): void
 
     /**
      * Forgets all reflection data related to the specified class
+     *
+     * @param (int|string) $className
+     *
+     * @psalm-param array-key $className
```
kdambekalns, line 1748: Huh? Why `array-key`?

Neos.Flow/Classes/Cache/CacheFactory.php
```diff
@@ -57,18 +57,20 @@ class CacheFactory extends \Neos\Cache\CacheFactory
 
     /**
      * @param CacheManager $cacheManager
-     * @Flow\Autowiring(enabled=false)
+     *
+     * @Flow\Autowiring (enabled=false)
```
kdambekalns, line 61: Why? I could live with the empty line, but the space before the parentheses looks very wrong!

Neos.Flow/Classes/Cache/CacheFactory.php
```diff
@@ -57,18 +57,20 @@ class CacheFactory extends \Neos\Cache\CacheFactory
...
+    public function injectCacheManager(CacheManager $cacheManager): void
     {
         $this->cacheManager = $cacheManager;
     }
 
     /**
      * @param EnvironmentConfiguration $environmentConfiguration
-     * @Flow\Autowiring(enabled=false)
+     *
+     * @Flow\Autowiring (enabled=false)
```
kdambekalns, line 71: Why? I could live with the empty line, but the space before the parentheses looks very wrong!

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -213,40 +166,71 @@ protected function buildSetInstanceCode(Configuration $objectConfiguration)
...
-        \$result = \$this->Flow_serializeRelatedEntities(\$transientProperties, \$propertyVarTags);\n";
+        if ($this->reflectionService->hasMethod($className, '__sleep')) {
+            return '';
         }
-        return $code;
+
+        $scopeAnnotation = $this->reflectionService->getClassAnnotation($className, Flow\Scope::class);
+
+        $doBuildCode =  $this->reflectionService->getClassAnnotation($className, Flow\Entity::class) !== null;
+        $doBuildCode = $doBuildCode || ($scopeAnnotation->value ?? 'prototype') === 'session';
```
kdambekalns, line 190:
```suggestion
        $isEntity =  $this->reflectionService->getClassAnnotation($className, Flow\Entity::class) !== null;
        $doBuildCode = $isEntity || ($scopeAnnotation->value ?? 'prototype') === 'session';
```

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -688,4 +670,15 @@ protected function compileStaticMethods($className, ProxyClass $proxyClass)
...
+
+    private function skipBuildForClass(string $className, string $objectName): bool
+    {
+        if ($className === '' || $this->compiler->hasCacheEntryForClass($className) === true) {
+            return true;
+        }
+        if ($objectName !== $className || $this->reflectionService->isClassAbstract($className)) {
+            return true;
+        }
+        return false;
```
kdambekalns, line 682: This could be simpified like "in the other place" I saw this in this PR…

kdambekalns (commented):
- I support the concept
- Reading the diff I left some comments/questions
- I have not (yet) tested the change

Neos.Flow/Classes/Reflection/ReflectionService.php
robertlemke: I tried letting Psalm resolve a few issues which went completely havoc and apparently I forgot to revert these changes at some places. This was not intentional.

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
robertlemke: The meaning is actually "should I build the code"? And that variable name is used elsewhere in proxy class building, so I used that here again for consistency.

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
robertlemke: Fair enough … I'll polish this during the next ProxyClassBuilder refactoring.

Neos.Flow/Classes/Reflection/ReflectionService.php
robertlemke: No idea, Psalm added it. I'll go through all doc comments and clean up in a separate PR.

Neos.Flow/Classes/Cache/CacheFactory.php
robertlemke: I'll fix that in a separate PR.

kdambekalns:
> DTOs (from forms for example) with attached entities can happen

Those should still work, as the attached entities would have serialization code and the DTO should be serializable "by plain PHP" then, no?

robertlemke:
> Those should still work, as the attached entities would have serialization code and the DTO should be serializable "by plain PHP" then, no?

But I guess that waking them up will fail then. Or magically work … something interested people might want to try out.

kitsunet (approved):

kdambekalns (approved):
Nitpicks could be solved later, too…

kdambekalns:
> But I guess that waking them up will fail then. Or magically work … something interested people might want to try out.

If serialization works, wakeup should work, too… 🤔

We'll see (or may never know) 😇

Issue #3049, opened 2023-05-15
Only add constructor injection code if needed


robertlemke:
Currently Flow generates proxy classes with constructor injection code for all classes which have a constructor. Classes like read models however never need constructor injection and Flow needlessly creates proxy classes for those as well.

Therefore, change the Proxy Class Builder so that only proxy classes with constructor injection code are generated if the original class is a singleton, because only singleton original classes are supported anyway and read models will never be singletons.

related: #1539

PR #3050, opened 2023-05-15
TASK: Only add constructor injection code if needed

robertlemke:
The proxy class builder now skips code generation for constructor injection code if the given original class is prototype, no user-defined object configuration exists and all potentially autowired constructor arguments are prototypes or simple values. This change should result in a significantly less amount of proxy classes generated in most modern Flow projects.

resolves: #3049
resolves: #1539

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
@@ -97,7 +97,7 @@ class ProxyClassBuilder
             if ($proxyClass === false) {
                 continue;
             }
-            $this->logger?->debug(sprintf('Building dependency injection proxy for "%s"', $className), LogEnvironment::fromMethodName(__METHOD__));
+            $this->logger->debug(sprintf('Building dependency injection proxy for "%s"', $className), LogEnvironment::fromMethodName(__METHOD__));
 
             $constructor = $proxyClass->getConstructor();
             $constructor->addPreParentCallCode($this->buildSetInstanceCode($objectConfiguration));
@@ -184,8 +184,10 @@ class ProxyClassBuilder
             return '';
         }
 
+        $scopeAnnotation = $this->reflectionService->getClassAnnotation($className, Flow\Scope::class);
+
         $doBuildCode =  $this->reflectionService->getClassAnnotation($className, Flow\Entity::class) !== null;
-        $doBuildCode = $doBuildCode || $this->reflectionService->getClassAnnotation($className, Flow\Scope::class)?->value === 'session';
+        $doBuildCode = $doBuildCode || ($scopeAnnotation->value ?? 'prototype') === 'session';
         if ($doBuildCode === false) {
             return '';
         }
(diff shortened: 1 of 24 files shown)


robertlemke:
Note: This PR requires (and contains) https://github.com/neos/flow-development-collection/pull/3047 which should be merged first.

kitsunet:
Sounds like a reasonable assumption, I was quite convinced we do this somewhere, but if not, great!

mhsdesign:
So in case there should be a `Flow\Scope("singleton")`, forgetting it seems to matters now?

as i assume from the description, that neither class will get a proxy and in turn an injection of `SomeService` or `new SomeService()` will not work (because the params are not wired)?
```php
class SomeService
{
   public function __construct(private readonly $otherService OtherService) {}
}

class OtherService {}
```

i certainly have some code doing that ^^
... but maybe we can even write an migration for that ...

the distinction doesnt seem like a bad decision and it would force people to actually use the singleton scope more - and I guess this is good so I’m in favour of this idea too.

robertlemke:
> So in case there should be a `Flow\Scope("singleton")`, forgetting it seems to matters now?

Yes, your example would not be autowired anymore. I think we can justify that change because the advantages outweigh the disadvantages, but I'm not sure how a meaningful automatic migration could be implemented.

mhsdesign:
The actual diff is rather small ^^: https://github.com/robertlemke/flow-development-collection/compare/1539-dont-create-proxies-for-classes-that-dont-need-it...robertlemke:flow-development-collection:3049-only-add-constructor-injection-code-if-needed#diff-e60ad6f2cb9f1f0603166db88fc40906fa480caf078cac14efaa52ff41f3ccaeR20


(count me as a half review ^^)

robertlemke:
I can think of more edge cases when a proxy class is created but not needed. I suggest to optimize the proxy class builder in a follow-up PR in order to keep the review process as simple as possible.

albe:
Haven't looked through the change yet, but I like it from the description!

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -312,7 +318,7 @@ protected function buildConstructorInjectionCode(Configuration $objectConfigurat
             $index++;
         }
 
-        return $code;
+        return $doBuildCode ? $code : '';
```
kdambekalns, line 321: The actual building of `$code` could be skipped, if the check was done after the `foreach ($argumentConfigurations as  …` loop…

kdambekalns (approved):
Except a small optimization comment this looks good. How old was that `TODO` comment about checking for prototypes?

mhsdesign:
> How old was that TODO comment about checking for prototypes?

was wondering the same at first, but it was just introduced with https://github.com/neos/flow-development-collection/pull/3047 [here](https://github.com/neos/flow-development-collection/pull/3047/files#diff-8515e53b9cee4f4ad7e1fe36ff941a83d25992a174d5643a28ea6d9a6aad8e9bR228) :D

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
kitsunet: can be done later, merging to move forward :)

kitsunet (approved):

Issue #3025, opened 2023-05-02
Add support for readonly classes


robertlemke:
https://www.php.net/releases/8.2/en.php#readonly_classes

PR #3051, opened 2023-05-15
FEATURE: Add support for readonly classes

robertlemke:
Flow now respects readonly classes during proxy class building and makes sure that proxy classes are readonly as well.

resolves: #3025

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
@@ -215,7 +215,7 @@ class ProxyClass
             $classModifier = 'final ';
         }
         if ($this->reflectionService->isClassReadonly($this->fullOriginalClassName)) {
-            $classModifier = 'readonly ';
+            $classModifier .= 'readonly ';
         }
 
         $constantsCode = $this->renderConstantsCode();
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
@@ -17,9 +17,8 @@ use Neos\Flow\Reflection\ReflectionService;
 
 /**
  * Representation of a Proxy Class during rendering time
- *
- * @Flow\Proxy(false)
  */
+#[Flow\Proxy(false)]
 class ProxyClass
 {
     /**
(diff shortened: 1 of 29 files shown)


robertlemke:
Note that this PR is based on https://github.com/neos/flow-development-collection/pull/3050 which needs to be merged first.

Diff: https://github.com/robertlemke/flow-development-collection/compare/3049-only-add-constructor-injection-code-if-needed...robertlemke:flow-development-collection:3025-add-support-for-readonly-classes

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/ValueObjectClassA.php
```diff
@@ -0,0 +1,36 @@
...
+ * source code.
+ */
+
+/**
+ * A readonly class in the style of a value object
+ */
+readonly class ValueObjectClassA implements \JsonSerializable
+{
+    public function __construct(
+        readonly public string $value,
```
mhsdesign, line 20: sorry nitpicky: i think we can drop the `readonly` in the property then

kdambekalns (approved):
Looks good!

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/ValueObjectClassA.php
kdambekalns: It's to make sure it's really, really, really readonly! 😇

kdambekalns: Honestly, yes – but I'd mix it, and leave one in. Just to have yet another edge-case…

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyClass.php
```diff
@@ -216,6 +214,9 @@ public function render()
         } elseif ($this->reflectionService->isClassFinal($this->fullOriginalClassName)) {
             $classModifier = 'final ';
         }
+        if ($this->reflectionService->isClassReadonly($this->fullOriginalClassName)) {
+            $classModifier = 'readonly ';
```
kitsunet: wouldn't that override the `final` modifiers, if both existed?

kdambekalns: Good point. And would `abstract readonly` be a thing?

kitsunet: both are allowed, yes

kitsunet:
```suggestion
            $classModifier .= 'readonly ';
```

kitsunet (approved):

PR #3052, opened 2023-05-16
BUGFIX: Skip proxy for optional straight values

robertlemke:
When a promoted property was an optional straight value, the proxy class builder decided to create a proxy class because it could be a straight value configured in the object configuration via Objects.yaml. Flow now checks the value of the given argument and only triggers proxy class building if the argument is not null. That way, Flow will not build useless proxies for typical read models which expect a mix of objects and straight values in their constructor.

related: #1539
related: #3049

### DIFF
--- a/Neos.Flow/Tests/Functional/ObjectManagement/DependencyInjectionTest.php
+++ b/Neos.Flow/Tests/Functional/ObjectManagement/DependencyInjectionTest.php
@@ -17,6 +17,7 @@ use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\FinalClassWithDependenc
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\Flow175\ClassWithTransitivePrototypeDependency;
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PrototypeClassA;
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PrototypeClassH;
+use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PrototypeClassI;
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\SingletonClassA;
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\ValueObjectClassA;
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\ValueObjectClassB;
@@ -326,6 +327,22 @@ class DependencyInjectionTest extends FunctionalTestCase
     {
         $object = new PrototypeClassH(
             new ValueObjectClassA('foo'),
+            new ValueObjectClassB('bar')
+        );
+        self::assertNotInstanceOf(ProxyInterface::class, $object);
+
+        $object = new PrototypeClassA();
+        self::assertInstanceOf(ProxyInterface::class, $object);
+    }
+
+    /**
+     * @test
+     */
+    public function noProxyClassIsGeneratedForPrototypeClassesWithOptionalStraightValues(): void
+    {
+        $object = new PrototypeClassI(
+            new ValueObjectClassA('foo'),
+            null,
             new ValueObjectClassB('bar')
         );
         self::assertNotInstanceOf(ProxyInterface::class, $object);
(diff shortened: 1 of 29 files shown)


robertlemke:
Note that this PR is based on https://github.com/neos/flow-development-collection/pull/3051 which needs to be merged first.

Diff: https://github.com/robertlemke/flow-development-collection/compare/3025-add-support-for-readonly-classes...robertlemke:flow-development-collection:bugfix/skip-proxy-for-optional-straight-values

albe:
Sidenode: I was somehow weirded off by the `straight` nomenclatur because I couldn't immediately decypher what a "straight value" is supposed to be (I've been absent from the Flow codebase for a bit now).
After checking https://github.com/neos/flow-development-collection/blob/d14198d03d42a0f406565c50d85bcff6dad0f69e/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php#L239-L247 I get that it's a direct PHP value as opposed to a settings or object configuration `reference`. Not sure if there could a better naming, but in any case not a topic for this PR. Just wanted to mention it

robertlemke:
> I get that it's a direct PHP value as opposed to a settings or object configuration `reference`. Not sure if there could a better naming, but in any case not a topic for this PR. Just wanted to mention it

I agree, it's some 15-year-old wording there, but I chose to not change too much unrelated at the same time. I could imagine that, when everything we aim for is working nicely, we go for a round of cosmetic / conceptual / architectural changes. But even then we need to outweigh our urge for cleanliness with stability and consistency. For example, I'd like to rename classes in the proxy class namespace – the ProxyClassBuilder is actually a ProxyClassGenerator. We'll see … first things first!

kdambekalns:
StyleCI has been fixed already in the `readonly` class PR…

kdambekalns:
> "straight value" is supposed to be

Would `native PHP literal` be better (to change it to later)? Or does that not quite fit?

Neos.Flow/Tests/Functional/ObjectManagement/DependencyInjectionTest.php
```diff
@@ -333,4 +334,20 @@ public function noProxyClassIsGeneratedForPrototypeClassesWithOnlyPrototypeConst
...
+     * @test
+     */
+    public function noProxyClassIsGeneratedForPrototypeClassesWithOptionalStraightValues(): void
+    {
+        $object = new PrototypeClassI(
+            new ValueObjectClassA('foo'),
+            null,
+            new ValueObjectClassB('bar')
+        );
+        self::assertNotInstanceOf(ProxyInterface::class, $object);
```
kdambekalns, line 348: What about

```php
$object = new PrototypeClassI(
    new ValueObjectClassA('foo'),
    'John Doe',
    new ValueObjectClassB('bar')
);
```

Would that be a proxy, and why?

kdambekalns (commented):
Admittedly I am confused about why only `null` is checked…

robertlemke:
> Admittedly I am confused about why only `null` is checked…

Yes, I worried that the code is not expressive enough – I added a comment but then removed it again.

The assumption is this: if the value is not empty, it has been set by an object configuration (usually via Objects.yaml) and we need to create a proxy in order to inject the argument. If it is null, the argument configuration only exists because Flow tries to auto-wire the arguments. That alone would not justify creating a proxy.

Neos.Flow/Tests/Functional/ObjectManagement/DependencyInjectionTest.php
robertlemke: No matter what you pass there – if null or a string – there wouldn't be a proxy, because that argument is passed during runtime when proxies were already built.

See my other comment for more details, I hope that helps.

kitsunet (approved):
Makes sense to me, but then I know this code.

kdambekalns (approved):

albe:
> > "straight value" is supposed to be
> 
> Would `native PHP literal` be better (to change it to later)? Or does that not quite fit?

That's formally the most correct description I guess? Though a bit bulky. `value literal`? Dunno. Let's discuss that in a to-be-created separate PR

Issue #3042, opened 2023-05-10
Introduce laminas code for proxy method rendering


robertlemke:
As a preparation for the refactoring of Flow's proxy class building and the reflection service, introduce [laminas/laminas-code](https://github.com/laminas/laminas-code) for rendering the methods in our proxy classes.

mhsdesign:
Hi - just out of interest (i bet you have this thought out already), but wasnt there a performance problem with laminas? https://github.com/neos/flow-development-collection/pull/2701#issuecomment-1078739551 or as i assume the performance problem was about laminas code reflection and not laminas/laminas-code?

robertlemke:
> Hi - just out of interest (i bet you have this thought out already), but wasnt there a performance problem with laminas? [#2701 (comment)](https://github.com/neos/flow-development-collection/pull/2701#issuecomment-1078739551) or as i assume the performance problem was about laminas code reflection and not laminas/laminas-code?

Yes, that's the challenge I need to solve, but I'm on it.

PR #3057, opened 2023-05-17
BUGFIX: Union types in proxy classes

### LINKED ISSUE #2941: BUG: Use of union types breaks proxy building, creates broken PHP code

robertlemke:
Flow's proxy class building now supports union types in method signatures.

This change merely adds a test which proves that the feature is working. The actual implementation was part of #3042.

resolves: #2941


robertlemke:
**NOTE:** https://github.com/neos/flow-development-collection/pull/3064 must be merged before this one.

robertlemke:
Since this only contains a test, I merged this now and appreciate any post-merge review.

Issue #3062, opened 2023-05-25
BUG: Authentication is broken with new Proxy Class Building


bwaidelich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Authenicating (with a session bound auth token) leads to an exception
```
Serialization of 'Closure' is not allowed
```

### Expected Behavior

no exception :)

## 1. Configure an authentication provider (e.g. using the `UsernamePassword` token) with an entry point

for example
```yaml
Neos:
  Flow:
    security:
      authentication:
        providers:
          'Acme.SomePackage:Default':
            provider: PersistedUsernamePasswordProvider
            token: UsernamePassword
            entryPoint: WebRedirect
            entryPointOptions:
              routeValues:
                '@package': Acme.SomePackage
                '@controller': Authentication
                '@action': login
```

## 2. Authenticate

---

Alternatively, the Flow version can be upgraded in a running Neos installation

### Environment

```markdown
- Flow: 9.0-dev
- PHP: 8.2
```

### Anything else?

Slack debugging session: https://neos-project.slack.com/archives/C3MCBK6S2/p1685021294539839

robertlemke:
It's not directly to be reproduced like you mentioned, you first need to apply another patch. Here's what I did:

- composer create-project neos/neos-base-distribution
- modify composer.json to use 9.0.x-dev: `"neos/flow-development-collection": "9.0.x-dev as 8.3.1"`
- composer update, check that `Framework` indeed contains 9.0.x-dev (6cfe7d8d14f304a1a8d3b65a93dd1ff154fb768d)
- `doctrine:migrate`, create a user etc
- open Neos in a browser

What I see then, using PHP 8.2, are a lot of errors like:

> Deprecated: Creation of dynamic property Neos\Flow\ResourceManagement\Streams\ResourceStreamWrapper::$Flow_Injected_Properties is deprecated in /application/Data/Temporary/Development/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_ResourceManagement_Streams_ResourceStreamWrapper.php on line 567 

To solve that, cherry-pick and merge conflicts of this PR by @kitsunet : https://github.com/neos/flow-development-collection/pull/3032

Then try to open the Neos backend via https://thedomain/neos – you will end up on this page:


![Image](https://github.com/neos/flow-development-collection/assets/95582/caf086e6-cf97-415f-a629-fb73c704d846)

[...]

PR #3063, opened 2023-05-26
BUGFIX: Create serialization code for transient properties

robertlemke:
Due to a recent optimization, Flow was not generating `__sleep()` methods for classes which are not either entities or were configured with a session scope. This led to errors in classes which were using the `@Transient` annotation to exclude certain properties from serialization. Therefore, Flow now also generates proxy classes with `__sleep()` methods if the original class contains such annotations.

Resolves: #3062


bwaidelich (approved):
Yay!
I don't really understand the code to be honest but I can confirm that this fixes the authentication issue with Neos 9 \o/

So +1 not by reading :)

kitsunet:
post merge +1

PR #3064, opened 2023-05-26
TASK: Use Laminas Code for proxy method rendering

robertlemke:
Flow now uses laminas/laminas-code for rendering proxy methods. The Dependency Injection Proxy Class Builder was refactored and the classes ProxyConstructor and ProxyMethod were replaced by new implementations called ProxyConstructorGenerator and ProxyMethodGenerator respectively.

Resolves: #3042


robertlemke:
Notes for review:

- this is a refactoring and is not supposed to introduce new features or solve existing bugs
- the biggest change is that `ProxyMethod` and `ProxyConstructor` were completely rewritten to use Lamina's `MethodGenerator` for rendering methods, including method signatures
- a lot of attention and work went into comparing generated proxy classes of Flow 8.3 with this implementation; minor improvements were made along the refactoring, so the classes will not match 100%
- the ProxyMethod test was removed because the contained features are now provided by `MethodGenerator` which is already tested; however, it makes sense to add additional tests in a follow-up
- I tested this implementation with Neos 8.3 and a big Flow application; the things I tested worked all fine
- there is still code in AOP which renders method signatures (for constructor injection), not using Laminas Code. I plan to cover that in a separate PR
- this implementation may or may not be faster or slower than before – I plan to tackle performance tests separately
- I'm not sure what to do about the new Psalm complaints

kitsunet (approved):
By reading and tested on a project

ahaeslich (approved):
Tested in our neos project ... and found a neos bug 🙈

robertlemke:
Thanks a lot for your reviews and testing!

Issue #3059, opened 2023-05-17
Replace `self` with `static` in proxy classes


robertlemke:
Factory methods may use code like `new self()` for creating a new instance. However, this code will break if the class is proxied, as `self` will still refer to the original class, which does not contain DI / AOP code. The compiler could therefore replace `self` keywords with `static` in the rendered proxy class file to work around this issue.

mhsdesign:
Sweet

bwaidelich:
as already commented on [Slack](https://neos-project.slack.com/archives/C050KKBEB/p1684399979909129?thread_ts=1684223801.762719&cid=C050KKBEB):

> rewriting `self` to `static` automagically seems dangerous to me due to the different semantics

I might be wrong on this one and I think we should definitely experiment with it to get to a stage were we can basically proxy any PHP class.

A safer alternative could be to rewrite (`private function __construct` to `protected function __construct`) (that would also fix the broken proxy class mentioned in #3060).

If everything turns out to be too flaky, I'd rather prevent those classes from being proxied at all.

nezaniel:
I think this is the reason why I always have to add Proxy(false) to all my classes with static constructors; Might be a non-issue since they are probably no longer proxied anyway. Will test this thoroughly in the coming weeks

mhsdesign:
Hi i just stumbled upon this (as i was searching, if i can use the `self` to `static` replacements and the `private` to `protected` rewrites now in neos 9 dev ^^).

Sorry for asking this, but im wondering if we should just disable proxy logic for `private` constructor classes?
Currently the proxy building seemed to got a bit more complex with the two prs #3074 and #3072. And possibly also slower as we now leverage `debug_backtrace` and `token_get_all`.

```php
class SomeService
{
   #[Flow\Inject]
   protected OtherService $otherThing;

   private function __construct() {}

   public static function create() { return new self(); }
}
```

i would argue this is hardly a usecase? (And i dont know if constructor injection should work, but i think there would be no reason for the developer that it _should_ work ^^ ...)

Currently most private constructors i see are in `readonly` models, where we need no proxy building. If im right, i think we have now the heuristic to do no proxy building here? https://github.com/neos/flow-development-collection/issues/3049 - so this `self` and `private` replace is only useful for services?

```php
// no proxy class?
final readonly class Dto
{
     private function __construct(public string $bar) {}

     public static function fromString(string $bar) { return new self($bar); }
}

[...]

PR #3070, opened 2023-06-06
FEATURE: Exclude classes from constructor autowiring

robertlemke:
Classes can now explicitly be excluded from constructor autowiring through a new setting.

The setting accepts an array of fully qualified class names, each class name being a regular expression. Classes of scope prototype which expect objects to be passed to their constructor are usually considered for autowiring which results in a proxy class being generated.

This option allows to exclude classes from this process. This is useful for classes like data transfer objects, read models, commands, events and value objects which usually don't rely on dependency injection.

Flow cannot reliably detect weather a prototype class depends on autowiring for constructor arguments or not. Use this option to optimize your application to avoid the small but measurable overhead of proxy generation for those kinds of classes.

Note that if there are other reasons than constructor injection which require a proxy class to be generated, the proxy class will be generated no matter what.

This change partly reverts #3050 because now proxy classes _are_ generated for prototype classes by default. Otherwise a lot of existing Flow applications would not work correctly anymore.

resolves: #3049

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
@@ -20,6 +20,7 @@ use Neos\Flow\ObjectManagement\Configuration\Configuration;
 use Neos\Flow\ObjectManagement\Configuration\ConfigurationArgument;
 use Neos\Flow\ObjectManagement\Configuration\ConfigurationProperty;
 use Neos\Flow\ObjectManagement\Exception as ObjectException;
+use Neos\Flow\ObjectManagement\Exception\InvalidObjectConfigurationException;
 use Neos\Flow\ObjectManagement\Exception\UnknownObjectException;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 use Neos\Flow\ObjectManagement\Proxy\Compiler;
@@ -45,6 +46,10 @@ class ProxyClassBuilder
     protected LoggerInterface $logger;
     protected ConfigurationManager $configurationManager;
     protected CompileTimeObjectManager $objectManager;
+
+    /**
+     * @var array<Configuration>
+     */
     protected array $objectConfigurations = [];
 
     public function injectReflectionService(ReflectionService $reflectionService): void
@@ -206,7 +211,7 @@ class ProxyClassBuilder
 
         $scopeAnnotation = $this->reflectionService->getClassAnnotation($className, Flow\Scope::class);
         $transientProperties = $this->reflectionService->getPropertyNamesByAnnotation($className, Flow\Transient::class);
-        $injectedProperties =  $this->reflectionService->getPropertyNamesByAnnotation($className, Flow\Inject::class);
+        $injectedProperties = $this->reflectionService->getPropertyNamesByAnnotation($className, Flow\Inject::class);
 
         $doBuildCode = $this->reflectionService->getClassAnnotation($className, Flow\Entity::class) !== null;
         $doBuildCode = $doBuildCode || (count($transientProperties) > 0);
@@ -251,16 +256,17 @@ class ProxyClassBuilder
      *
      * @throws InvalidConfigurationTypeException
      * @throws UnknownObjectException
+     * @throws InvalidObjectConfigurationException
      */
     protected function buildConstructorInjectionCode(Configuration $objectConfiguration): string
     {
-        $doBuildCode = $objectConfiguration->getScope() !== Configuration::SCOPE_PROTOTYPE;
+        $doReturnCode = $objectConfiguration->getScope() !== Configuration::SCOPE_PROTOTYPE;
 
         $assignments = [];
-
         $argumentConfigurations = $objectConfiguration->getArguments();
         $constructorParameterInfo = $this->reflectionService->getMethodParameters($objectConfiguration->getClassName(), '__construct');
         $argumentNumberToOptionalInfo = [];
+
         foreach ($constructorParameterInfo as $parameterInfo) {
             $argumentNumberToOptionalInfo[($parameterInfo['position'] + 1)] = $parameterInfo['optional'];
         }
@@ -283,7 +289,7 @@ class ProxyClassBuilder
                 switch ($argumentConfiguration->getType()) {
                     case ConfigurationArgument::ARGUMENT_TYPES_OBJECT:
                         if ($argumentValue instanceof Configuration) {
-                            $doBuildCode = true;
+                            $doReturnCode = true;
                             $argumentValueObjectName = $argumentValue->getObjectName();
                             $argumentValueClassName = $argumentValue->getClassName();
                             if ($argumentValueClassName === null) {
@@ -304,22 +310,20 @@ class ProxyClassBuilder
                                 throw new UnknownObjectException('The object "' . $argumentValue . '" which was specified as an argument in the object configuration of object "' . $objectConfiguration->getObjectName() . '" does not exist.', 1264669967);
                             }
                             $assignments[$argumentPosition] = $assignmentPrologue . '\Neos\Flow\Core\Bootstrap::$staticObjectManager->get(\'' . $argumentValue . '\')';
-                            if ($this->objectConfigurations[$argumentValue]->getScope() !== Configuration::SCOPE_PROTOTYPE) {
-                                $doBuildCode = true;
-                            }
+                            $doReturnCode = true;
                         }
                         break;
 
                     case ConfigurationArgument::ARGUMENT_TYPES_STRAIGHTVALUE:
                         $assignments[$argumentPosition] = $assignmentPrologue . var_export($argumentValue, true);
                         if ($argumentValue !== null) {
-                            $doBuildCode = true;
+                            $doReturnCode = true;
                         }
                         break;
 
                     case ConfigurationArgument::ARGUMENT_TYPES_SETTING:
                         $assignments[$argumentPosition] = $assignmentPrologue . '\Neos\Flow\Core\Bootstrap::$staticObjectManager->get(\Neos\Flow\Configuration\ConfigurationManager::class)->getConfiguration(\Neos\Flow\Configuration\ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, \'' . $argumentValue . '\')';
-                        $doBuildCode = true;
+                        $doReturnCode = true;
                         break;
                 }
             }
@@ -344,7 +348,7 @@ class ProxyClassBuilder
             $index++;
         }
 
-        return $doBuildCode ? $code : '';
+        return $doReturnCode ? $code : '';
     }
 
     /**
--- a/Neos.Flow/Configuration/Settings.Object.yaml
+++ b/Neos.Flow/Configuration/Settings.Object.yaml
@@ -28,8 +28,8 @@ Neos:
 
         # Explicitly exclude classes from constructor autowiring.
         #
-        # This setting accepts an array of fully qualified class names, each class name
-        # being a regular expression.
+        # This setting accepts an array of regular expressions, each matching one or more fully
+        # qualified class names.
         #
         # Classes of scope prototype which expect objects to be passed to their constructor
         # are usually considered for autowiring which results in a proxy class being generated.
@@ -37,7 +37,7 @@ Neos:
         # like data transfer objects, read models, commands, events and value objects which
         # usually don't rely on dependency injection.
         #
-        # Flow cannot reliably detect weather a prototype class depends on autowiring for
+        # Flow cannot reliably detect wether a prototype class depends on autowiring for
         # constructor arguments or not. Use this option to optimize your application to avoid
         # the small but measurable overhead of proxy generation for those kinds of classes.
         #
--- a/Neos.Flow/Configuration/Settings.Object.yaml
+++ b/Neos.Flow/Configuration/Settings.Object.yaml
@@ -22,4 +22,32 @@ Neos:
       # reflected by default. You can however exclude specific (or all) classes from Flow packages
       # by specifying corresponding regular expressions that don't match classes to exclude.
       #
-      includeClasses: []
+      includeClasses: [ ]
+
+      dependencyInjection:
+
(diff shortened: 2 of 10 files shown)


Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -304,22 +310,20 @@ protected function buildConstructorInjectionCode(Configuration $objectConfigurat
                                 throw new UnknownObjectException('The object "' . $argumentValue . '" which was specified as an argument in the object configuration of object "' . $objectConfiguration->getObjectName() . '" does not exist.', 1264669967);
                             }
                             $assignments[$argumentPosition] = $assignmentPrologue . '\Neos\Flow\Core\Bootstrap::$staticObjectManager->get(\'' . $argumentValue . '\')';
-                            if ($this->objectConfigurations[$argumentValue]->getScope() !== Configuration::SCOPE_PROTOTYPE) {
-                                $doBuildCode = true;
-                            }
+                            $doReturnCode = true;
```
robertlemke, line 313: This is the line which reverts the decision we made earlier to not build proxy classes for prototypes at all.

bwaidelich:
Just two questions before getting deeper into this one:

1. Is there a specific reason, that this is configured in `Settings.yaml` rather than `Objects.yaml`?
2. Is it correct, that I have to add that configuration if I use a DTO like the following:

```php
final readonly class StringBasedDto {
  public function __construct(public string $value) {}
}
final readonly class CompoundDto {
  public function __construct(public StringBasedDto $foo, public StringBasedDto $bar) {}
}
```
?

Neos.Flow/Configuration/Settings.Object.yaml
```diff
@@ -22,4 +22,32 @@ Neos:
...
       #
-      includeClasses: []
+      includeClasses: [ ]
+
+      dependencyInjection:
+
+        # Explicitly exclude classes from constructor autowiring.
+        #
+        # This setting accepts an array of fully qualified class names, each class name
+        # being a regular expression.
```
bwaidelich: It's not actually FQN IMO, maybe something like:
```suggestion
        # This setting accepts an array of regular expressions, each matching one or more fully
        # qualified class names.
```

Neos.Flow/Configuration/Settings.Object.yaml
```diff
@@ -22,4 +22,32 @@ Neos:
...
+        #
+        # Classes of scope prototype which expect objects to be passed to their constructor
+        # are usually considered for autowiring which results in a proxy class being generated.
+        # This option allows to exclude classes from this process. This is useful for classes
+        # like data transfer objects, read models, commands, events and value objects which
+        # usually don't rely on dependency injection.
+        #
+        # Flow cannot reliably detect weather a prototype class depends on autowiring for
```
bwaidelich:
```suggestion
        # Flow cannot reliably detect wether a prototype class depends on autowiring for
```

Neos.Flow/Configuration/Settings.Object.yaml
```diff
@@ -22,4 +22,32 @@ Neos:
...
+        # constructor arguments or not. Use this option to optimize your application to avoid
+        # the small but measurable overhead of proxy generation for those kinds of classes.
```
bwaidelich, line 42: This is slightly misleading. At least for me the performance overhead is not the main issue, but that this won't work with proxy classes:

```php
final readonly Foo {
  public function withBar(): self {
     return new self(...);
  }
}
```

or is that fixed differently somehow?

bwaidelich (commented):
Just a few more (partly nitpicky) comments for now

In general I have to say that I'm slightly unsure about this to be honest.
Personally I would continue using the `#[Flow\Proxy(false)]` attribute rather than adding a Setting somewhere in the project that – if missing or wrong – could lead to the program breaking (or worse, behaving weirdly).

But I might have misunderstood the reasons for the whole setting in the first place

kitsunet:
@bwaidelich I argued somewhat into this direction, because I think we all agree less magic is better. The ultimate result would be IMHO to disable autowiring everywhere and you have to annotate when you want it in a constructor. But that again would be super breaking, so this is a bit of a middle ground. We cannot auto optimize this, but you can disable it quickly on a wide range of classes if you want to, if you do not use the setting no harm is done, you just have more proxies in you cache. But if you use it you can optimize the amount of proxies without needing to touch single classes.

bwaidelich:
> I argued somewhat into this direction

Into which direction, what are you referring to?

> The ultimate result would be IMHO to disable autowiring everywhere and you have to annotate
> when you want it in a constructor

Maybe this is a naive wish but to me the ultimate result would be that proxy classes are only created if they are actually needed and that it fails at compile time with a meaningful exception if that can't be resolved.

I seem to have missed some discussions but I fail to comprehend

> This change partly reverts https://github.com/neos/flow-development-collection/pull/3050 because now proxy classes are
> generated for prototype classes by default. Otherwise a lot of existing Flow applications would not work correctly anymore.

e.g. I think it's totally OK to disallow
```php
class SomeClassWithConstructorArgs {
  public function __construct(string $foo) {}
}
$someClass = new SomeClassWithConstructorArgs();
```
because even the IDE will scream at you.
And if I do:

```php
#[Flow\Scope('singleton')]
class SomeSingleton {
  public function __construct(SomeClassWithConstructorArgs $dep) {}
}
```
I wonder why I would need a proxy class for `SomeClassWithConstructorArgs`.


Also, do we really inject prototypes that much in the code?
What about making _that_ a configuration option and deprecating it for a transition phase?

Again, I probably miss a lot of the details – but I'm afraid that most people won't understand this new setting

robertlemke:
@bwaidelich I'll try to answer your questions:

First, let's get on the same page regarding why this has nothing to do `Proxy(false)` and your `withBar()` example. There are still cases / PHP 8 language constructs, which are not supported by proxy classes and which will break. For example, I'm sure that there will be issues with private constructors. These cases need to be tackled separately from this pull requests. Flow must either decide to not create a proxy class or throw an exception if something cannot be supported. There should hardly ever be a situation where you need to set `Proxy(false)`.

Now, what is this change then about?

I tried to provide an implementation which does not break existing code which relies on constructor injection using prototypes be supported for classes also being prototype. There are several examples of this in Flow and Neos and many applications based on Flow. It's debatable if this is a good style, but it has been supported for years, so we can't just remove it. For half a day I tried going that path and refactor an application to fit, but it turned out to be a lot of wor.

So, if we continue to support prototypes being injected to prototypes, we need to generate constructor injection code for these cases. However, we can't detect reliably, if you actually want to use constructor injection or if this is just a read model or value object. So, by default, we need to generate that constructor injection code and thus a proxy class. A proxy class for such a case would look like this (removed some comments for brevity):

```php
#…
use Flownative\Beach\Domain\ValueObject\InstanceIdentifier;
use Flownative\Beach\Domain\ValueObject\Name;

class RenameInstance_Original
{
    public function __construct(
        readonly public InstanceIdentifier $instanceIdentifier,
        readonly public Name $newInstanceName
    ) {
    }
}

final class RenameInstance extends RenameInstance_Original implements \Neos\Flow\ObjectManagement\Proxy\ProxyInterface {

    public function __construct()
    {
        $arguments = func_get_args();
        if (!array_key_exists(0, $arguments)) $arguments[0] = \Neos\Flow\Core\Bootstrap::$staticObjectManager->get('Flownative\Beach\Domain\ValueObject\InstanceIdentifier');
        if (!array_key_exists(1, $arguments)) $arguments[1] = \Neos\Flow\Core\Bootstrap::$staticObjectManager->get('Flownative\Beach\Domain\ValueObject\Name');
        if (!array_key_exists(0, $arguments)) throw new \Neos\Flow\ObjectManagement\Exception\UnresolvedDependenciesException('Missing required constructor argument $instanceIdentifier in class ' . __CLASS__ . '. Note that constructor injection is only support for objects of scope singleton (and this is not a singleton) – for other scopes you must pass each required argument to the constructor yourself.', 1296143788);
        if (!array_key_exists(1, $arguments)) throw new \Neos\Flow\ObjectManagement\Exception\UnresolvedDependenciesException('Missing required constructor argument $newInstanceName in class ' . __CLASS__ . '. Note that constructor injection is only support for objects of scope singleton (and this is not a singleton) – for other scopes you must pass each required argument to the constructor yourself.', 1296143788);
    }
}
```

In this case, this is useless code. It adds a small overhead which is neglectible in smaller projects but may be more significant in bigger projects. Therefore, I added the setting which allows for _optimizing_ for these cases. Again, this setting is not needed to make something work or avoid errors. It is only about optimizing performance.

There are multiple ways to optimize:
1. add an annotation / attribute to the constructor:
   ```php
   #[Flow\Autowiring(false)]
   public function __construct(
        readonly public InstanceIdentifier $instanceIdentifier,
        readonly public Name $newInstanceName
    ) {
    }
   ```
2. setting `#[Flow\Proxy(false)]`  for the whole class
3. disabling autowiring via _Objects.yaml_
4. disabling autowiring via patterns through the new setting in _Settings.yaml_

Now, why is the new option a **setting** and not being done via objects configuration?

The structure of _Objects.yaml_ expects an object name on the first level of the hierarchy. There is no support for configuring multiple objects with one statement using wildcards or regular expressions. I shortly looked into making that possible, but it's a medium nightmare in terms of implementation but also from a user's perspective.

I was reluctant to introduce a setting, but at the moment I think that this is the best compromise. It does not allow you to configure arbitrary object properties but is clearly only meant for disabling constructor injection by using an array of patterns. You can use it, if your application has a clear pattern for classes which don't require constructor injection or if you cannot easily modify the code (for example, because it is a third party library).

In Beach, for example, it's sufficient to add this configuration to get rid of all unnecessary proxy classes:

```yaml
    object:
      dependencyInjection:
        excludeClassesFromConstructorAutowiring:
          - '^Flownative\\Beach\\Domain\\.*\\Command\\.*$'
          - '^Flownative\\Beach\\Domain\\.*\\Event\\.*$'
          - '^Flownative\\Beach\\Domain\\.*\\Projection\\.*$'
```

That's much easier than going through > 100 classes and adding attributes / annotations there.

robertlemke:
PS: The reason why I consider `Proxy(false)` as a last resort and suggest to use `Autowiring(false)`, the new setting etc. if possible is, that by disabling proxy class building completely, you also loose any AOP-related features, including security aspects. Especially with libraries / extensions you might give users a last-resort-extensibility-option via AOP, to fix something which you have not foreseen.

kitsunet:
Thanks robert, was about to write a longer exlpanation as well. 

I would like to add that:

> prototypes being injected to prototypes

might seem odd, but there are IMHO valid (and bad) examples in Flow core already. This gets especially interesting with third party code, which doesn't know about our scopes. Prototypes created by factories, stuff that can be created based on settings, I think there are good examples of this, where an object shouldn't be or doesn't have to be a singleton yet makes sense to get injected as you don't want to hardcode this dependency. (Some examples are difficult because some conditions apply, but Caches and Loggers could be such examples).

bwaidelich:
@robertlemke @kitsunet Thanks for these detailed explanations, it makes much more sense to me now.

I can't tell how much the impact of these proxy classes really is (did you measure it?) but if it's "just" about performance, I think it's fine to have a somewhat "esoteric" setting for this.

> might seem odd, but there are IMHO valid (and bad) examples in Flow core already

valid or bad? I guess _both_ ;)
Do you have any at hand, I find that highly suspicious – but that's just out of curiosity. I didn't mean to suggest removing support for this. I just wonder wether we need to support `new ClassWithPrototypeConstructorInjections()` rather than `$objectManager->get(ClassWithPrototypeConstructorInjections::class)`. And for the latter I question that we need a proxy class.

But, with your explanations, I guess that's not really related to this PR

> The reason why I consider Proxy(false) as a last resort
> [...] you also loose any AOP-related features, including security aspects
> [...] give users a last-resort-extensibility-option via AOP, to fix something

Also not directly related, but in the meantime I tend to create dedicated non-Flow-packages for the core domain logic and integration packages with all the DI and AOP support because I deliberately don't want any 3rd party to interfere with that.

It might just be a personal thing, but IMO the code style and architecture got so much cleaner and easier to maintain that way. Even (or especially) for libraries I think that we should embrace a defensive programming style and rely on DI/AOP for the integration.
For fixes/patches I would much prefer people to create PRs and/or pull them in via [composer-patches](https://github.com/cweagans/composer-patches)

robertlemke:
> I just wonder wether we need to support `new ClassWithPrototypeConstructorInjections()` rather than `$objectManager->get(ClassWithPrototypeConstructorInjections::class)`. And for the latter I question that we need a proxy class.

For the latter we currently do need a proxy class, considering how the whole dependency injection feature is implemented. For example, property injection is only possible with a proxy class – if you don't want to use reflection during runtime.

> It might just be a personal thing, but IMO the code style and architecture got so much cleaner and easier to maintain that way. Even (or especially) for libraries I think that we should embrace a defensive programming style and rely on DI/AOP for the integration.
For fixes/patches I would much prefer people to create PRs and/or pull them in via [composer-patches](https://github.com/cweagans/composer-patches)

Personally, I fully agree! However, I assume that there are others who still like to be able to do these things. We might want to discuss this further during a sprint and see if we can get rid of some worst-practices in the future.

kdambekalns:
The explanations given on the why/how (not) should go into the documentation, I think… 😉

bwaidelich:
> Personally, I fully agree! However, I assume that there are others who still like to be able to do these things

Absolutely. I was just wondering whether we should do these "micro-optimizations" for them.
Then again, I don't know how micro this really _is_ and I got you side-tracked enough already :)

kitsunet:
> For the latter we currently do need a proxy class, considering how the whole dependency injection feature is implemented. For example, property injection is only possible with a proxy class – if you don't want to use reflection during runtime.

We _could_ however implement that in the future, because we have all the information from compiletime already, the objectmanager could very well do the constructor injection if we wanted to.

Neos.Flow/Configuration/Settings.Object.yaml
robertlemke: This will be fixed with https://github.com/neos/flow-development-collection/issues/3059

kitsunet (approved):
I'd say lets do it for now, we can still iterate on this later on.

Issue #3058, opened 2023-05-17
Support private constructors in proxy classes


robertlemke:
When a proxy class is built for an original class containing a private constructor, the proxy class file must change the original constructor to `protected` and then create a new constructor which is `private`.

PR #3072, opened 2023-06-08
FEATURE: Support private constructors in proxy classes

robertlemke:
Flow now can correctly build proxy classes for classes with private constructors. Previously, such classes caused errors and proxy class building had to be disabled with the `Proxy(false)` annotation. Now classes with private constructors can take advantage of setter and property injection and are considered for advices through the AOP framework.

Resolves: #3058

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyConstructorGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyConstructorGenerator.php
@@ -16,6 +16,8 @@ use Neos\Flow\ObjectManagement\DependencyInjection\ProxyClassBuilder;
 
 final class ProxyConstructorGenerator extends ProxyMethodGenerator
 {
+    private ?string $originalVisibility = null;
+
     public function __construct($name = null, array $parameters = [], $flags = self::FLAG_PUBLIC, $body = null, $docBlock = null)
     {
         if ($docBlock === null) {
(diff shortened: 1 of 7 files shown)


Neos.Flow/Classes/ObjectManagement/Proxy/ProxyConstructorGenerator.php
```diff
@@ -87,14 +95,66 @@ protected function buildAssignMethodArgumentsCode(): string
...
+        return <<<PHP
+        \$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2);
+        if (isset(\$backtrace[1]) &&
+            !is_subclass_of(\$backtrace[1]['class'], \\{$this->fullOriginalClassName}::class) &&
+            !is_subclass_of(\\{$this->fullOriginalClassName}::class, \$backtrace[1]['class']) &&
+            \$backtrace[1]['class'] !== '{$this->fullOriginalClassName}' &&
+            \$backtrace[1]['class'] !== '{$fullOriginalClassNameWithSuffix}'
+        ) {
+            throw new \\Error('Call to {$originalVisibilityString} {$this->fullOriginalClassName}::__construct() from invalid context', 1686153840);
+        }
```
bwaidelich, line 157: AFAIR `debug_backtrace` is not very performant (I didn't double check!) and I know, this is only done if there's a proxy class, so it might not be an issue. But I wonder: Do we really need to check whether a private constructor is invoked from the "outside"?
I mean, static analyzers and the IDE will scream at you, if you do – and if you really _do_ want to invoke the constructor, you could probably still circumvent the check with some reflection magic, anyways

robertlemke: I think it's not super important to have this check (even though I worked hard to get this running). My motivation is mostly that proxy classes should not change the behavior of the proxied code if at all possible, so don't get bad surprises. 

Two made up scenarios:

1. not using or ignoring your IDE, you program something calling a private constructor which works fine because the class is proxied due to a security advice. When the advice does not match / is removed some day, this breaks, and you'll only notice it during runtime
2. a user instantiates a class in their code; the maintainer of that class at some point decides to make the constructor private, but the user won't notice that because it still works

🤷‍♂️

bwaidelich: yes, very good points – And doctrine uses the same kind of checks in their [deprecations](https://github.com/doctrine/deprecations/blob/eac01fbdf4e4a3635169b056608730ec0930e188/lib/Doctrine/Deprecations/Deprecation.php#L129) package, so the performance hit is probably not too bad

kitsunet (approved):
Not a big fan of the backtrace thing, but it works, we used it in other places before, so lets support this.

PR #3074, opened 2023-06-08
FEATURE: Replace self with static in proxy classes

robertlemke:
Factory methods which use code like new self() for creating a new instance are now handled correctly in proxy classes. The compiler automatically replaces "self" keywords with "static" in the rendered proxy class file to make this possible.

This implementation has not been optimized for performance.

Resolves: #3059

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Exception/ProxyCompilerException.php
+++ b/Neos.Flow/Classes/ObjectManagement/Exception/ProxyCompilerException.php
@@ -1,0 +1,18 @@
+<?php
+namespace Neos\Flow\ObjectManagement\Exception;
+
+/*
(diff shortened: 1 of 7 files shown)


robertlemke:
Notes for review:

- this pull requests is based on https://github.com/neos/flow-development-collection/pull/3072 which should be merged first
- the implementation is not optimized for speed yet, for example, the tokenizing needed for this replacement could be combined with that of `replaceClassName()`. But this needs some more thorough thinking and certainly a day of attention to make properly
- even though this wasn't optimized for speed, I did not notice any notable speed issues in Development context. I didn't measure though

kitsunet (approved):

Issue #3079, opened 2023-06-14
BUG: Upgrading http-factories from 7.3.8 to 7.3.12 *downgrades* GuzzleHttp


jakobmerrild:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When upgrading the dependency on `neos/http-factories` from `7.3.8` to `7.3.12` the dependency on `guzzlehttp/psr7` is downgraded from `^2.1.1` to `^1.8.4`

I suspect this is because the `7.3.8` tag seems to accidentally be on a commit on the `8.x` branch https://github.com/neos/http-factories/blob/ddf639fb7d39460d87acd48c66b4e0c98be56a78/composer.json#L12
while `7.3.12` seems to once again be on the `7.x` branch: https://github.com/neos/http-factories/blob/8184b3a1c92add547400e3ef063b1095c87d7808/composer.json#L12

### Expected Behavior

I generally expect transitive dependencies to go in the same direction as dependencies on `neos` packages, i.e. if I bump my `neos` version transitive dependencies stay the same version or are bumped and vice versa.

### Environment

```markdown
- Flow: 7.3.12
- PHP: 8.1
```

dlubitz:
Thanks for your contribution. Looks like an upmerge issue. Was handled properly for later versions.

https://github.com/neos/http-factories/blob/8.0.9/composer.json

dlubitz:
Sorry for the inconvenience, but 7.3.8 seems to be a broken tag with wrong dependencies merged. With 7.3.9 everything looks fine.

PR #3081, opened 2023-06-15
BUGFIX: Revert changes to dependenies due to upmerge

dlubitz:
Due to an upmerge the dependencies where set to lower versions as before. This reverts this change to the old version values.

Fixes: #3079

### DIFF
--- a/Neos.Http.Factories/composer.json
+++ b/Neos.Http.Factories/composer.json
@@ -7,7 +7,7 @@
         "MIT"
     ],
     "require": {
-        "php": "^8.0",
+        "php": "^7.1 || ^8.0",
         "psr/http-factory": "^1.0",
         "guzzlehttp/psr7": "^1.8.4 || ^2.1.1"
     },
--- a/Neos.Http.Factories/composer.json
+++ b/Neos.Http.Factories/composer.json
@@ -7,9 +7,9 @@
         "MIT"
     ],
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^8.0",
         "psr/http-factory": "^1.0",
-        "guzzlehttp/psr7": "^1.8.4"
+        "guzzlehttp/psr7": "^1.8.4 || ^2.1.1"
     },
     "autoload": {
         "psr-4": {


Neos.Http.Factories/composer.json
```diff
@@ -7,9 +7,9 @@
         "MIT"
     ],
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^8.0",
```
bwaidelich: This would change the required PHP version in a patch level release!?

dlubitz: Basically... yes. But it was already that in [7.3.8](https://github.com/neos/http-factories/blob/7.3.8/composer.json). Was just relaxed due to a upmerge issue.

bwaidelich: But that change is in the `neos/http-factories` package, so Flow 7.3.x could still be installed with PHP < 8. With this. Or am I missing something here?

dlubitz: IDK. I guess so. But I was just checking the history of this package composer.json and was assuming this is right, as it was released like that before.

kdambekalns: As of https://github.com/neos/http-factories/commit/a15a3853780bbb37617f0080e05ea954f00f983b, to be precise.

But… that cannot be correct, as https://docs.neos.io/guide/installation-development-setup/system-requirements#php-version-matrix as well as https://github.com/neos/flow-development-collection/blob/1d64e6a72c2e419e7dd9a5546bf4ac0735a70001/composer.json say it's compatible with PHP 7.3 and up.

So it seems that back then this was a mistake.

dlubitz: I see! But `guzzlehttp/psr7@^2.1.1` is fine?

dlubitz: So actually this commit must not have been gone into 7.3:
https://github.com/neos/http-factories/commit/a15a3853780bbb37617f0080e05ea954f00f983b

kdambekalns: > I see! But `guzzlehttp/psr7@^2.1.1` is fine?

Huh… not sure, would need to look it up.

> So actually this commit must not have been gone into 7.3: [neos/http-factories@a15a385](https://github.com/neos/http-factories/commit/a15a3853780bbb37617f0080e05ea954f00f983b)

I'd say so, yes!

dlubitz: @kdambekalns I changed the dependencies accordingly. Could you please approve it, if this is fine for you?

mhsdesign:
As far as i understand, we accidentally released a 7.3.8 which was a trojan horse and actually contained a 8.x

And now when people seemingly upgrade from 7.3.8 to 7.3.9 or above, they actually downgrade from 8.x to 7.3.9 and thus the dependencies are again "correctly" defined.

But this means that this pr doesnt revert anything but only loosens the constraints so people getting used to their trojan horse installation are not robbed of their newer dependencies?

kdambekalns (commented):
Still not sure…

- The addition of `guzzlehttp/psr7` in version `^2.1.1` would be new, nothing that ever existed, see git blame (screenshot below). Why is it needed, what does it do?
- The PR title is no longer correct, pleae adjust when the previous question has been answered.

image

dlubitz:
@kdambekalns You are right, But something weired happend to the 7.3.8 tag. It contains a commit, that doesn't exists in 7.3 branch. This commit is named `MERGE: Merge branch '7.3' into 8.0`, which seems totally wrong for a 7.3,x tag. Anyways...

Commits of branch 7.3: https://github.com/neos/flow-development-collection/commits/7.3/Neos.Http.Factories/composer.json
Commits of tag 7.3.8: https://github.com/neos/flow-development-collection/commits/7.3.8/Neos.Http.Factories/composer.json

 I'll close this one and the issue.

kdambekalns:
Right, as was pointed out here, the 7.3.8 was tagged off the `8.2` branch, as far as I can tell… 🙈

In the following image orange is `8.0`, yellow is `7.3` and red is `master`:
branch-point

At some point that was branched to be `8.2`, and the release was done on it:

tag-point

The tag now lives "disconnected" in the Git repository…

PR #3086, opened 2023-06-16
FEATURE: InjectConfiguration for constructor arguments

### LINKED ISSUE #3077: FEATURE: Configure settings directly in a constructor argument
```php
public function __construct(
    #[Flow\Configuration("fusion.autoInclude")]
    private array $autoIncludeConfiguration,
) {
}
```

this syntax combines the benefit, of both worlds:

- inject configuration is easily readable and configurable (you dont need to add a comment that you use Objects.yaml)
- constructor injection + objects yaml is easily testable -> as you can just specifiy a different parameter 

_Originally posted by @mhsdesign in https://github.com/neos/flow-development-collection/issues/2966#issuecomment-1445154289_
            

also related: https://github.com/neos/flow-development-collection/issues/3080

mhsdesign:
Flow now supports InjectConfiguration attributes for constructor arguments which allows for injecting configuration, such as settings, via the constructor. Compared to property injection, constructor injection results in more portable and better testable code.

Resolves #3077

### DIFF
--- a/Neos.Flow/Classes/Annotations/InjectConfiguration.php
+++ b/Neos.Flow/Classes/Annotations/InjectConfiguration.php
@@ -11,11 +11,11 @@ namespace Neos\Flow\Annotations;
  * source code.
  */
 
-use Neos\Flow\Configuration\ConfigurationManager;
 use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
+use Neos\Flow\Configuration\ConfigurationManager;
 
 /**
- * Used to enable property injection for configuration including settings.
+ * Used to enable property and constructor argument injection for configuration including settings.
  *
  * Flow will build Dependency Injection code for the property and try
  * to inject the configured configuration.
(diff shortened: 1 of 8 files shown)


Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -99,6 +99,7 @@ class ReflectionService
     protected const DATA_PROPERTY_ANNOTATIONS = 15;
     protected const DATA_PROPERTY_VISIBILITY = 24;
     protected const DATA_PROPERTY_TYPE = 26;
+    protected const DATA_PROPERTY_PROMOTED = 26;
```
dlubitz: This looks suspicios! Both const have the same value.

mhsdesign: Upsi 😄 Nice 👌

mhsdesign:
```php
$reflected = new \ReflectionClass($this);
$constants = $reflected->getConstants();
$duplicateConstantValues = array_diff_assoc($constants, array_unique($constants, SORT_NUMERIC));
if ($duplicateConstantValues !== []) {
    throw new \CompileError(sprintf('Class "%s" constants must have unique values. Doubles found: %s', __CLASS__, json_encode($duplicateConstantValues)));
}
```

ich würde ja gerne sowas einbauen als sanity check aber es scheint, dass mehrere konstanten sich einen wert teilen.

```
Class "Neos\Flow\Reflection\ReflectionService" constants must have unique values.
Doubles found: {"DATA_INTERFACE_IMPLEMENTATIONS":1,"DATA_CLASS_INTERFACES":2,"DATA_CLASS_SUBCLASSES":3,"DATA_PARAMETER_SCALAR_DECLARATION":24,"DATA_PARAMETER_ANNOTATIONS":25}
```

und das jetzt anzupassen wäre bestimmt sinvoll, aber man müsste auch die cache clearen ... aber das macht man so oder so ab und zu?

Was sagt ihr @kitsunet @robertlemke?

Neos.Flow/Classes/Annotations/InjectConfiguration.php
```diff
@@ -49,21 +49,33 @@ final class InjectConfiguration
...
+    public readonly string $type;
 
     public function __construct(?string $path = null, ?string $package = null, ?string $type = null)
     {
         $this->path = $path;
         $this->package = $package;
-        if ($type !== null) {
-            $this->type = $type;
+        $this->type = $type ?? ConfigurationManager::CONFIGURATION_TYPE_SETTINGS;
+    }
```
mhsdesign: we should put the validation, that package cannot be used together with type if its a non setting.

previous error message

> The InjectConfiguration annotation for property "injectedViewsConfiguration" in class "Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\ClassWithInjectedConfiguration" specifies a "package" key for configuration type "Views", but this is only supported for injection of "Settings".

doctrine will help us make a simple message readable:

> [Creation Error] An error occurred while instantiating the annotation @Flow\InjectConfiguration declared on property Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\ClassWithInjectedConfiguration::$injectedViewsConfiguration: "Invalid usage of "package" with configuration type "Views". Using "package" is only valid for "Settings".".

Neos.Flow/Classes/Reflection/ReflectionService.php
bwaidelich: I'd say: Better catch those typos with tests *g

robertlemke (approved):
I added documentation, fixed a few cosmetic glitches and squashed the three commits into one with a meaningful commit message. Tested it with a real project.

kitsunet (approved):

dlubitz (approved):

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassL.php
```diff
@@ -0,0 +1,27 @@
...
+use Neos\Flow\Annotations as Flow;
+
+/**
+ * A readonly class of scope prototype with settings injected in the constructor
+ */
+readonly class PrototypeClassL
+{
+    public function __construct(
+        #[Flow\InjectConfiguration(path: 'tests.functional.settingInjection.someSetting')]
+        public string $value
```
mhsdesign, line 23: this feature needs a followup as `private` properties cannot be set:

> The property "headerComment" in class "HtmlDocumentFactory" must not be private when annotated for configuration injection.

Packages/Framework/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
  Line: 613

`protected` works too but... flow seems to guard that as the property is promoted

Issue #3078, opened 2023-06-14
BUG: Instantiation of singleton via `new` creates new object and overrides instance in objectmanager


mhsdesign:
yes, i could reproduce this, see my failing tests at: https://github.com/neos/flow-development-collection/pull/3093

the question is, whether this is intended behaviour, and if not how could we efficiently restrict it?

this would work

```php
class Lol_Original
{
    public function __construct()
    {
    }
}

class Lol extends Lol_Original
{
    private function __construct()
    {
        parent::__construct();
    }

    /**
     * Autogenerated Proxy Method
     */
    public static function __unsafeCreateSingletonInstance(): self
    {
        return new self();
    }
}

// doesnt work
// $instance = new Lol();

$instance = Lol::__unsafeCreateSingletonInstance();
```

PR #3093, opened 2023-06-17
!!! WIP BUGFIX instantiation of singletons with new should fail?

mhsdesign:
see https://github.com/neos/flow-development-collection/issues/3078

@robertlemke i talked with you about this behaviour and i could reproduce it as described to you in this test.
What do you say, is this intended behaviour?

it does makes sense from a technical point of view, why it behaves like this, because the proxy contains this line in the constructor:

```php
if (get_class($this) === 'Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\SingletonClassA') \Neos\Flow\Core\Bootstrap::$staticObjectManager->setInstance('Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\SingletonClassA', $this);
```

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Tests/Functional/ObjectManagement/DependencyInjectionTest.php
+++ b/Neos.Flow/Tests/Functional/ObjectManagement/DependencyInjectionTest.php
@@ -160,6 +160,35 @@ class DependencyInjectionTest extends FunctionalTestCase
     /**
      * @test
      */
+    public function singletonCannotBeInstantiatedViaNew(): void
+    {
+        $this->expectException(\Throwable::class);
+        new Fixtures\SingletonClassA();
+    }
+
+    /**
+     * TODO remove this test, only to proof our currently invalid? situation.
+     *
+     * @test
+     */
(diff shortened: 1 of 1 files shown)


mhsdesign:
hi what do you say @robertlemke what should be the correct behavior?

robertlemke:
In my opinion, the correct behavior should be that an exception is thrown if you try to instantiate a class marked as Singleton the second time. Ideally you would even get an exception when trying it with *new* instead of using the Object Manager.

One way I could imagine this to be implemented is like this:

- ObjectManager::get($objectName) is called – either directly or indirectly via Dependency Injection
- if there is already an instance of $objectName, it is returned
- otherwise, an internal flag is set, that $objectName is now going to be instantiated
- the class is instantiated
- the internal flag is unset the the instance is returned

Now, if you try to instantiate the class with new, the constructor in the proxy class calls a new method ObjectManager::isBeingInstantiated($objectName) which checks the internal flag. If that is false, the proxy class constructor throws an exception.

PR #3101, opened 2023-06-23
BUGFIX: Replacement proxy methods rendered again

### LINKED ISSUE #3099: BUG: Performance `StreamWrapperAdapter` is not proxied

### Current Behavior

StreamWrapperAdapter is not proxied.

### Expected Behavior

The StreamWrapperAdapter gets StreamWrapperInterface implementations compiled into the proxy in Production context to speed up the bootstrap.

### Steps To Reproduce

install Flow 9.0 run help in Production context and check proxy classes.

### Environment

```markdown
- Flow: 9.0.x-dev
- PHP: 8.2.5
```

### Anything else?

_No response_

kitsunet:
This fixes a bug introduced in d939e6b8 switching to laminuas-code. A proxy method can replace the full body of an existing method or even be a fully new method, in which case only `body` will be set in the proxy method. We still want those to be generated. This for example currently breaks the CompileStatic feature, as those methods do not get rendered anymore resulting in worse performance in Production context compared to before.

This fix renders a proxy method also when a body was set for it, but still skips it if neither pre/post nor body is set.

It also enabled CompileStatic in Testing Context so that it is testable and adds a test to make sure it works as intended.

Fixes: #3099

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
@@ -154,7 +154,7 @@ class ProxyClassBuilder
             $constructor->addPostParentCallCode($this->buildLifecycleInitializationCode($objectConfiguration, ObjectManagerInterface::INITIALIZATIONCAUSE_CREATED));
             $constructor->addPostParentCallCode($this->buildLifecycleShutdownCode($objectConfiguration, ObjectManagerInterface::INITIALIZATIONCAUSE_CREATED));
 
-            if ($this->objectManager->getContext()->isProduction()) {
+            if ($this->objectManager->getContext()->isProduction() || $this->objectManager->getContext()->isTesting()) {
                 $this->compileStaticMethods($className, $proxyClass);
             }
         }
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -108,7 +108,7 @@ class ProxyMethodGenerator extends MethodGenerator
      */
     public function willBeRendered(): bool
     {
-        return ($this->addedPreParentCallCode !== '' || $this->addedPostParentCallCode !== '');
+        return ($this->addedPreParentCallCode !== '' || $this->addedPostParentCallCode !== '' || $this->body !== '');
     }
 
     /**
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -23,11 +23,14 @@ class ProxyMethodGenerator extends MethodGenerator
 
     protected string $fullOriginalClassName = '';
 
+    protected bool $bodyWasOverridden = false;
+
     public static function fromReflection(\Laminas\Code\Reflection\MethodReflection $reflectionMethod): static
     {
         $instance = parent::fromReflection($reflectionMethod);
         assert($instance instanceof static);
         $instance->fullOriginalClassName = $reflectionMethod->getDeclaringClass()->getName();
+        $instance->bodyWasOverridden = false;
         return $instance;
     }
 
@@ -108,7 +111,7 @@ class ProxyMethodGenerator extends MethodGenerator
      */
     public function willBeRendered(): bool
     {
-        return ($this->addedPreParentCallCode !== '' || $this->addedPostParentCallCode !== '' || $this->body !== '');
+        return ($this->addedPreParentCallCode !== '' || $this->addedPostParentCallCode !== '' || $this->bodyWasOverridden);
     }
 
     /**
@@ -160,5 +163,11 @@ class ProxyMethodGenerator extends MethodGenerator
             return '';
         }
         return 'parent::' . $methodName . '(' . $this->buildMethodParametersCode($fullClassName, $methodName, false) . ");\n";
+    }
+
+    public function setBody($body): static
+    {
+        $this->bodyWasOverridden = true;
+        return parent::setBody($body);
     }
 }
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -11,6 +11,7 @@ namespace Neos\Flow\ObjectManagement\Proxy;
  * source code.
  */
 
+use Laminas\Code\Generator\DocBlockGenerator;
 use Laminas\Code\Generator\MethodGenerator;
 
 /**
@@ -31,10 +32,13 @@ class ProxyMethodGenerator extends MethodGenerator
         return $instance;
     }
 
-    public static function copyMethodSignature(\Laminas\Code\Reflection\MethodReflection $reflectionMethod): static
+    public static function copyMethodSignatureAndDocblock(\Laminas\Code\Reflection\MethodReflection $reflectionMethod): static
     {
         $instance = parent::copyMethodSignature($reflectionMethod);
         assert($instance instanceof static);
+        if ($reflectionMethod->getDocComment() !== false) {
+            $instance->setDocBlock(DocBlockGenerator::fromReflection($reflectionMethod->getDocBlock()));
+        }
         $instance->fullOriginalClassName = $reflectionMethod->getDeclaringClass()->getName();
         return $instance;
     }
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -23,14 +23,11 @@ class ProxyMethodGenerator extends MethodGenerator
 
     protected string $fullOriginalClassName = '';
 
-    protected bool $bodyWasOverridden = false;
-
     public static function fromReflection(\Laminas\Code\Reflection\MethodReflection $reflectionMethod): static
     {
         $instance = parent::fromReflection($reflectionMethod);
         assert($instance instanceof static);
         $instance->fullOriginalClassName = $reflectionMethod->getDeclaringClass()->getName();
-        $instance->bodyWasOverridden = false;
         return $instance;
     }
 
@@ -73,6 +70,11 @@ class ProxyMethodGenerator extends MethodGenerator
         if ($this->body === '') {
             $this->body = $this->renderBodyCode();
         }
+
+        if ($this->body === '') {
+            return '';
+        }
+
         return parent::generate();
     }
 
@@ -111,7 +113,7 @@ class ProxyMethodGenerator extends MethodGenerator
      */
     public function willBeRendered(): bool
     {
-        return ($this->addedPreParentCallCode !== '' || $this->addedPostParentCallCode !== '' || $this->bodyWasOverridden);
+        return ($this->addedPreParentCallCode !== '' || $this->addedPostParentCallCode !== '' || $this->body !== '');
     }
 
     /**
@@ -163,11 +165,5 @@ class ProxyMethodGenerator extends MethodGenerator
             return '';
         }
         return 'parent::' . $methodName . '(' . $this->buildMethodParametersCode($fullClassName, $methodName, false) . ");\n";
-    }
-
-    public function setBody($body): static
-    {
-        $this->bodyWasOverridden = true;
-        return parent::setBody($body);
     }
 }
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -89,7 +89,7 @@ class ProxyMethodGenerator extends MethodGenerator
         }
 
         $callParentMethodCode = $this->buildCallParentMethodCode($this->fullOriginalClassName, $this->name);
-        $returnTypeIsVoidOrNever = ((string)$this->getReturnType() === 'void' || (string)$this->getReturnType() === 'never' );
+        $returnTypeIsVoidOrNever = ((string)$this->getReturnType() === 'void' || (string)$this->getReturnType() === 'never');
         $code = $this->addedPreParentCallCode;
         if ($this->addedPostParentCallCode !== '') {
             if ($returnTypeIsVoidOrNever) {
--- a/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
+++ b/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
@@ -22,6 +22,6 @@ class PrototypeClassK
      */
     public static function compiledStaticallyMethod(): float
     {
-        return microtime(true);
+        return bin2hex(random_bytes(10));
     }
 }
--- a/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
+++ b/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
@@ -9,7 +9,7 @@ use Neos\Flow\Utility\Algorithms;
  */
 class PrototypeClassK
 {
-    public function getMicrotime(): float
+    public function getToken(): string
     {
         return static::compiledStaticallyMethod();
     }
--- a/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
+++ b/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
@@ -1,0 +1,27 @@
+<?php
+namespace Neos\Flow\Tests\Functional\ObjectManagement\Fixtures;
+
+use Neos\Flow\Annotations as Flow;
+
+/**
+ * A class to test static compile functionality
+ */
+class PrototypeClassK
+{
+    public function getMicrotime(): float
+    {
+        return static::compiledStaticallyMethod();
+    }
+
+    /**
+     * Method that should get static compiled into the proxy, saving some processing power in production,
+     * but also providing the exact same result on every call.
+     *
+     * @return float
+     * @Flow\CompileStatic
+     */
+    public static function compiledStaticallyMethod(): float
+    {
+        return microtime(true);
+    }
+}
--- a/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
+++ b/Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
@@ -2,6 +2,7 @@
 namespace Neos\Flow\Tests\Functional\ObjectManagement\Fixtures;
 
 use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Utility\Algorithms;
 
 /**
  * A class to test static compile functionality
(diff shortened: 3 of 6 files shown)


kitsunet:
Mmm, should probably see if I can come up with a test for this?

kitsunet:
I don#t quite understand the test failures, as I get them in the 9.0 branch without my changes as well...

kitsunet:
Ah, damn I see the problem.... 

When we call `getMethod()` with an existing method name  on a proxy we will get a `\Neos\Flow\ObjectManagement\Proxy\ProxyMethodGenerator` filled from reflection, which includes `$this->body` being filled with the original method body. So with this change every method that we ever create a proxy method generator for will end up with the original body. I suppose we actually should most of the time create a `ProxyMethodGenerator` without the original body (see  `\ProxyManager\Generator\MethodGenerator::fromReflectionWithoutBodyAndDocBlock`).

Or include a distinction somewhere between original body and overwritten body.

kitsunet:
This still does funky things with the proxies, but the bug is fixed with it. Investigating further....

kitsunet:
This seems to generate proper proxies for me, fixed a small thing with empty `__wakeup` being generated along the way. Neos 9.0 works fine. Still not sure what's going on with those tests....

robertlemke (approved):
In both, Development and Production, context, after applying this change, proxy methods which were generated to implement an AOP advice now contain the full doc block of the original method, including description and annotations.

In Production context, `StreamWrapperAdapter` is now proxied and contains an optimized version of `getStreamWrapperImplementationClassNames()`.

Therefore, it should be safe to merge this pull request. I'm not sure about the speed implications due to rendering all the new doc blocks. On the other hand, it's probably better to do so, as the proxied methods now contain all original annotations. I would guess that this potentially fixes other problems we haven't experienced yet.

(I compared the cache directories of our biggest application at Flownative with and without the patch applied)

mhsdesign:
id like to understand 😅 can you please add an example proxy class with before and after in the description or a test? ^^

And maybe we should also add inline comments or something to prevent any future regressions ...

robertlemke:
Here's an excerpt from a diff of the files in the cache directory – left side without this patch, right side with this patch applied. Blue highlighted files were changed, the red highlighted file did not exist without this patch:

image

As an example, here's the code diff of `AuthenticationProviderManager`:

image

robertlemke:
PS: Yes, this fix should please come with a test.

kitsunet:
> Here's an excerpt from a diff of the files in the cache directory – left side without this patch, right side with this patch applied....

> 
> As an example, here's the code diff of `AuthenticationProviderManager`:
> image

Screenshot 2023-06-26 at 10 08 39

This is my local AuthenticationProviderManager proxy, with the patch, and the result without docblock seems logical according to the code. 

`\Neos\Flow\ObjectManagement\Proxy\ProxyMethodGenerator::fromReflection` will copy the docblock in the parent call, while I can't find anything about docblocks in the corresponding `copyMethodSignature` call. 

The question is, do we want the docblock or not? Currently we do copy it over in < 9.0 versions, but I question if it makes sense? But it shouldn't be too hard to add if we want to.

kdambekalns:
> The question is, do we want the docblock or not? Currently we do copy it over in < 9.0 versions, but I question if it makes sense?

We used to transfer over some annotations, but at some point consciously decided to copy the full docblock as it makes sure everything is in place that was added to the original code. Let's see… ah, there you go: https://github.com/neos/flow-development-collection/pull/2533

robertlemke:
> This is my local AuthenticationProviderManager proxy, with the patch, and the result without docblock seems logical according to the code.

Yes, that *would* seem logical, but the behavior is different in my case!

Consider this screenshot:

image

Note that the left side is the Production cache directory generated with 9.0-dev and the right side is the cache generated using your cache. With the patch applied there *is* a copied doc comment!

Am I doing something wrong or can you confirm this behavior?

robertlemke:
(I deleted a previous comment which was wrong … I mixed up directories here)

I would say that it is important having the doc comments in the proxy and that this is the reason why I didn't choose to use `copyMethodSignature()` in the first place.

After proxy classes are generated, the Reflection Service updates its reflection data again. For example, if AOP advised a method, reflection is updated and *then* dependency injection is applied. Therefore I assume (without testing) that with this patch applied, DI may be broken and reflection might be wrong for advised methods.

I assume that we need the docblocks in the proxy methods unless we can prove the contrary.

kitsunet:
alright, will add docblocks again and a test :) Any hints regarding the failing tests?

robertlemke (approved):
LGTM

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Method that should get static compiled into the proxy, saving some processing power in production,
+     * but also providing the exact same result on every call.
+     *
+     * @return float
+     * @Flow\CompileStatic
+     */
+    public static function compiledStaticallyMethod(): float
+    {
+        return microtime(true);
```
mhsdesign: might be better to use random string instead as if your mashine is fast enough it might yield the same result either way?

kitsunet: I guess in theory? How long of a random string do we want to use?

kitsunet:
```suggestion
        return bin2hex(random_bytes(10));
```

kitsunet: something like this looks fine to you @mhsdesign

mhsdesign: yeah i have no clue ;) sometimes we also use `md5(uniqid(mt_rand(), true))` but im not a cryptographer ^^ pick what you like :P

kdambekalns: What about [`Neos\Flow\Utility\Algorithms::generateRandomToken(10)`](https://github.com/neos/flow-development-collection/blob/15591ccdd9bc661765fbe3a25f51de6bb4d7038f/Neos.Flow/Classes/Utility/Algorithms.php#L68C28-L74)?
```suggestion
        return \Neos\Flow\Utility\Algorithms::generateRandomToken(10);
```
(plus a `use` statement)

kdambekalns: Did you resolve this before or after my comment, @kitsunet?

kdambekalns:
`Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PrototypeClassK::compiledStaticallyMethod(): Return value must be of type float, string returned` sounds like a problem… 😎

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php
kdambekalns: Applied my suggestion while fixing the test…

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
```diff
@@ -31,10 +32,13 @@ public static function fromReflection(\Laminas\Code\Reflection\MethodReflection
         return $instance;
     }
 
-    public static function copyMethodSignature(\Laminas\Code\Reflection\MethodReflection $reflectionMethod): static
+    public static function copyMethodSignatureAndDocblock(\Laminas\Code\Reflection\MethodReflection $reflectionMethod): static
```
kdambekalns, line 35: Was that method unused before this change? It has its name changed, but its use in this PR is "new". 🤔

kdambekalns (approved):
Looks good, left a question. Hoping for the tests to pass.

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
kitsunet: Indeed, some places (AOP) use the parent copyMethodSignature directly, this one was unused, I think it may have been in preparation of future changes but I guess we always want the docblock and if not can re-add a copy of it.

kitsunet:
> `Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PrototypeClassK::compiledStaticallyMethod(): Return value must be of type float, string returned` sounds like a problem… 😎

Thanks for the fix!

bwaidelich:
This PR seems to break all doctrine proxies in testing context (see https://neos-project.slack.com/archives/C3MCBK6S2/p1692608042742949?thread_ts=1692608042.742949&cid=C3MCBK6S2)

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -154,7 +154,7 @@ public function build(): void
             $constructor->addPostParentCallCode($this->buildLifecycleInitializationCode($objectConfiguration, ObjectManagerInterface::INITIALIZATIONCAUSE_CREATED));
             $constructor->addPostParentCallCode($this->buildLifecycleShutdownCode($objectConfiguration, ObjectManagerInterface::INITIALIZATIONCAUSE_CREATED));
 
-            if ($this->objectManager->getContext()->isProduction()) {
+            if ($this->objectManager->getContext()->isProduction() || $this->objectManager->getContext()->isTesting()) {
```
bwaidelich, line 157: I don't understand this check and it probably is the root cause for the regression.

In general we should avoid different behaviors based on the context IMO so to avoid tests getting less robust

kitsunet: Interesting question indeed, the compile static always had different behavior in production and development, I wonder though why I had to add this for testing, I'll investigate. The point of static compile Is to bake in expensive operations that cannot change during runtime (eg. all inerface implementations). It did work fine until the recent changes in the framework. this general switch is years old.

bwaidelich:
Here's a minimal example to reproduce the regression:

```php
<?php
namespace Neos\Flow\Command;

use Neos\Flow\Security\AccountRepository;

class TestCommandController extends CommandController
{

    public function __construct(private AccountRepository $accountRepository)
    {
        parent::__construct();
    }

    public function testCommand(): void
    {
        \Neos\Flow\var_dump($this->accountRepository->findAll());
    }
}
```

```
FLOW_CONTEXT=Testing ./flow test:test
```

returns
```
Flow Variable Dump
Neos\Flow\Persistence\Doctrine\QueryResult(empty)
```

and logs the following exception (for some reason the exception is just logged for me and not rendered, but that's probably a separate issue):

```
Exception #1542792708 in line 133 of Packages/Libraries/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php: Failure while fetching class schema class "Neos\Flow\Security\Account": No class schema found for "Neos\Flow\Security\Account".

34 Neos\Flow\Persistence\Doctrine\Mapping\Driver\FlowAnnotationDriver_Original::loadMetadataForClass("Neos\Flow\Security\Account", Neos\Flow\Persistence\Doctrine\Mapping\ClassMetadata)
33 Doctrine\ORM\Mapping\ClassMetadataFactory::doLoadMetadata(Neos\Flow\Persistence\Doctrine\Mapping\ClassMetadata, NULL, false, array|0|)
32 Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::loadMetadata("Neos\Flow\Security\Account")
31 Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getMetadataFor("Neos\Flow\Security\Account")
30 Doctrine\ORM\EntityManager::getClassMetadata("Neos\Flow\Security\Account")
29 Doctrine\ORM\Query\Parser::RangeVariableDeclaration()
28 Doctrine\ORM\Query\Parser::IdentificationVariableDeclaration()
27 Doctrine\ORM\Query\Parser::FromClause()
26 Doctrine\ORM\Query\Parser::SelectStatement()
25 Doctrine\ORM\Query\Parser::QueryLanguage()
24 Doctrine\ORM\Query\Parser::getAST()
23 Doctrine\ORM\Query\Parser::parse()
22 Doctrine\ORM\Query::parse()
21 Doctrine\ORM\Query::_doExecute()
20 Doctrine\ORM\AbstractQuery::executeIgnoreQueryCache(NULL, 4)
19 Doctrine\ORM\AbstractQuery::execute(NULL, 4)
18 Doctrine\ORM\AbstractQuery::getSingleResult(4)
17 Doctrine\ORM\AbstractQuery::getSingleScalarResult()
16 Neos\Flow\Persistence\Doctrine\Query_Original::count()
15 Neos\Flow\Persistence\Doctrine\QueryResult_Original::count()
14 count(Neos\Flow\Persistence\Doctrine\QueryResult)
13 Neos\Flow\Error\Debugger::renderArrayDump(Neos\Flow\Persistence\Doctrine\QueryResult, 1, true, true)
12 Neos\Flow\Error\Debugger::renderDump(Neos\Flow\Persistence\Doctrine\QueryResult, 0, true, true)
11 Neos\Flow\var_dump(Neos\Flow\Persistence\Doctrine\QueryResult)
10 Neos\Flow\Command\TestCommandController_Original::testCommand()
9 Neos\Flow\Cli\CommandController_Original::callCommandMethod()
8 Neos\Flow\Cli\CommandController_Original::processRequest(Neos\Flow\Cli\Request, Neos\Flow\Cli\Response)
7 Neos\Flow\Cli\Dispatcher_Original::dispatch(Neos\Flow\Cli\Request, Neos\Flow\Cli\Response)
6 Neos\Flow\Cli\CommandRequestHandler::Neos\Flow\Cli\{closure}()
5 Closure::__invoke()
4 Neos\Flow\Security\Context_Original::withoutAuthorizationChecks(Closure)
3 Neos\Flow\Cli\CommandRequestHandler::handleRequest()
2 Neos\Flow\Core\Bootstrap::run()
1 require("Packages/Framework/Neos.Flow/Scripts/flow.php")

Previous exception: Exception #1295973082 in line 184 of Data/Temporary/Testing/Cache/Code/Flow_Object_Classes/Neos_Flow_Persistence_Doctrine_Mapping_Driver_FlowAnnotationDriver.php: No class schema found for "Neos\Flow\Security\Account".

35 Neos\Flow\Persistence\Doctrine\Mapping\Driver\FlowAnnotationDriver_Original::getClassSchema("Neos\Flow\Security\Account")
34 Neos\Flow\Persistence\Doctrine\Mapping\Driver\FlowAnnotationDriver_Original::loadMetadataForClass("Neos\Flow\Security\Account", Neos\Flow\Persistence\Doctrine\Mapping\ClassMetadata)
33 Doctrine\ORM\Mapping\ClassMetadataFactory::doLoadMetadata(Neos\Flow\Persistence\Doctrine\Mapping\ClassMetadata, NULL, false, array|0|)
32 Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::loadMetadata("Neos\Flow\Security\Account")
31 Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getMetadataFor("Neos\Flow\Security\Account")
30 Doctrine\ORM\EntityManager::getClassMetadata("Neos\Flow\Security\Account")
29 Doctrine\ORM\Query\Parser::RangeVariableDeclaration()
28 Doctrine\ORM\Query\Parser::IdentificationVariableDeclaration()
27 Doctrine\ORM\Query\Parser::FromClause()
26 Doctrine\ORM\Query\Parser::SelectStatement()
25 Doctrine\ORM\Query\Parser::QueryLanguage()
24 Doctrine\ORM\Query\Parser::getAST()
23 Doctrine\ORM\Query\Parser::parse()
22 Doctrine\ORM\Query::parse()
21 Doctrine\ORM\Query::_doExecute()
20 Doctrine\ORM\AbstractQuery::executeIgnoreQueryCache(NULL, 4)
19 Doctrine\ORM\AbstractQuery::execute(NULL, 4)
18 Doctrine\ORM\AbstractQuery::getSingleResult(4)
17 Doctrine\ORM\AbstractQuery::getSingleScalarResult()
16 Neos\Flow\Persistence\Doctrine\Query_Original::count()
15 Neos\Flow\Persistence\Doctrine\QueryResult_Original::count()
14 count(Neos\Flow\Persistence\Doctrine\QueryResult)
13 Neos\Flow\Error\Debugger::renderArrayDump(Neos\Flow\Persistence\Doctrine\QueryResult, 1, true, true)
12 Neos\Flow\Error\Debugger::renderDump(Neos\Flow\Persistence\Doctrine\QueryResult, 0, true, true)
11 Neos\Flow\var_dump(Neos\Flow\Persistence\Doctrine\QueryResult)
10 Neos\Flow\Command\TestCommandController_Original::testCommand()
9 Neos\Flow\Cli\CommandController_Original::callCommandMethod()
8 Neos\Flow\Cli\CommandController_Original::processRequest(Neos\Flow\Cli\Request, Neos\Flow\Cli\Response)
7 Neos\Flow\Cli\Dispatcher_Original::dispatch(Neos\Flow\Cli\Request, Neos\Flow\Cli\Response)
6 Neos\Flow\Cli\CommandRequestHandler::Neos\Flow\Cli\{closure}()
5 Closure::__invoke()
4 Neos\Flow\Security\Context_Original::withoutAuthorizationChecks(Closure)
3 Neos\Flow\Cli\CommandRequestHandler::handleRequest()
2 Neos\Flow\Core\Bootstrap::run()
1 require("Packages/Framework/Neos.Flow/Scripts/flow.php")
```

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
bwaidelich: > this general switch is years old

I'm not sure how it's related. All I can say is that the regression seems to be gone without that check

bwaidelich:
For the record:

With `FLOW_CONTEXT=Development` and `FLOW_CONTEXT=Production` everything works as expected

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
kitsunet: yep, understood! Maybe it was unintended 🤷 I'll backtrace

bwaidelich:
I'm really surprised that the CI didn't catch this..
I wonder if we should pin the `neos-development-collection` to a version before this merge?

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
mhsdesign: no sorry uups i might have given you the idea ^^ https://neos-project.slack.com/archives/C04V4C6B0/p1687865311078709

the purpose was to make the test in `Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PrototypeClassK.php` work.

we basically check that the method is now pure and not a side effect anymore.

dlubitz: @kitsunet @mhsdesign @bwaidelich Do we still consider this as an issue? Maybe we should create a dedicated ticket then.

For some reason, it's not an issue in "my" behat tests anymore, but I don't know why. And we don't know, why it's not an issue at all in the flow and neos tests - but is easily reproducible with Bastians "CommandController".

kitsunet: Maybe we should roll this back together with the test and see if we can come up with a way to enable it for one test, recompile do test and flush/reset afterwards. It might just make sense to introduce this as a setting, that way it would be easier to set it on a case by case basis and people could disable it if it causes issues.

mhsdesign:
hmmm @bwaidelich could you please extract this information about hte regression into an issue so we can continue the discussion there and dont miss this topic?

bwaidelich:
=> #3190

PR #3102, opened 2023-06-23
9.0 FEATURE: Add `unique` flowQuery operation

mficzel:
This operation applies `array_unique` to the current flowQuery context.

While the same could previously achieved via `Array.unique()` the flow query operation can be placed in an operation chain without extra wrapping.

**Review instructions**

There is also a node specific implementation of the `unique` operation in https://github.com/neos/neos-development-collection/pull/4355

I know the php code looks oldish but the style is in line with the other flowQuery operations around. 

**Checklist**

### DIFF
--- a/Neos.Eel/Classes/FlowQuery/Operations/UniqueOperation.php
+++ b/Neos.Eel/Classes/FlowQuery/Operations/UniqueOperation.php
@@ -1,0 +1,43 @@
+<?php
+namespace Neos\Eel\FlowQuery\Operations;
+
+/*
+ * This file is part of the Neos.Eel package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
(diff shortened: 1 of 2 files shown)


mficzel:
Since psalm is lamenting about unrelated things i consider the test fine.

mhsdesign (approved):

bwaidelich (approved):
I don't _really_ see the need for this because it could be achieved with a simple `@process.unique = ${Array.unique(value)}` AFAIS but the code looks good and simple enough to me

mficzel:
@bwaidelich this is parallel to the node specific `unique` operation that we need. So this is not technically needed but improves the consistency.

bwaidelich:
Thanks for the clarification, that makes a lot of sense!

PR #3106, opened 2023-06-30
 BUGFIX: Publish Resources On Windows Without Admin Rights

mhsdesign:
**Tested on windows and works ^^**

see https://discuss.neos.io/t/running-neos-flow-on-windows-10-success/2752 and especially thanks to @creative-resort for the hint https://discuss.neos.io/t/running-neos-flow-on-windows-10-success/2752/6?u=marc

Now you should be able to setup a Neos on Windows without the admin shell ^^

`flow flow:cache:warmup` should _just_ work on a fresh install ^^

We currently only need to link whole directories, thats why the "hack" works.

previously it would fail to publish `Neos.Flow\Resources\Public` into `Web\_Resources\Static\Packages\Neos.Flow`

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
+++ b/Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
@@ -166,7 +166,14 @@ class FileSystemSymlinkTarget extends FileSystemTarget
                 $result = Files::createRelativeSymlink($sourcePathAndFilename, $targetPathAndFilename);
             } else {
                 $temporaryTargetPathAndFilename = $targetPathAndFilename . '.' . Algorithms::generateRandomString(13) . '.tmp';
-                symlink($sourcePathAndFilename, $temporaryTargetPathAndFilename);
+                if (PHP_OS_FAMILY === 'Windows' && is_dir($sourcePathAndFilename)) {
+                    // on windows creating "normal" symlinks is only allowed as admin or with SeCreateSymbolicLinkPrivilege
+                    // in case we only want to symlink a dir, we can instead create directory junctions, which is allowed for non admins
+                    // see https://github.com/git-for-windows/git/wiki/Symbolic-Links
+                    exec(sprintf('mklink /j %s %s', escapeshellarg($temporaryTargetPathAndFilename), escapeshellarg($sourcePathAndFilename)));
+                } else {
+                    symlink($sourcePathAndFilename, $temporaryTargetPathAndFilename);
+                }
                 $result = rename($temporaryTargetPathAndFilename, $targetPathAndFilename);
             }
         } catch (\Exception $exception) {


dlubitz:
Uff, this is a tough one. 
I can imagine this might open other symlink issues, which are not handled here. 
And I don't the drawbacks of junctions in comparison to symlinks.
Also what about users, which want to use symlinks (with supporting environment)?

Actually, when I was using Windows for Neos development, the symlinks where a "one time" issue you had to solve/configure. Afterwards I never got in touch with it. Maybe we just need to improve documentation here. On the other hand, it would be great to get it working without any tweeks to the host system.

dlubitz:
What about a dedicated `FileSystemWindowsJunctionTarget`?

mhsdesign:
i could also imagine first trying `symlink` ignoring all warnings and if this failed - only then try the windows special case. That would still allow me to use the windows admin shell to create "real" symlinks.

I just want to make flow work out of the box on windows, as many newcomers just start there.

mhsdesign:
image (2)

The new setup tool will now warn the windows user ;) 

https://github.com/neos/neos-development-collection/issues/4243

mhsdesign:
We should also ignore any errors inside `symlink` via `@` so the whole flow booting doesnt fail on windows "just" because the static resource collections cant be published (a little intense ^^)


https://github.com/neos/flow-development-collection/blob/9b1c66fb94f1544177b96cc9a3aed6a60fe806e5/Neos.Flow/Classes/Package.php#L105-L110

static resources are always republished in development mode when the file monitor detects changes inside `Flow_PublicResourcesFiles` or when the cache is flushed.

So i think its fine to just log a warning that they coudnt be published, and the user has to either A: run publish in an admin shell, or properly configure the windows user.

C: would be to incorporate this fix, or fall back to copying files instead of symlinking but as you said @dlubitz it might be more of a one time configuration issue ;)

kdambekalns:
> We should also ignore any errors inside `symlink` via `@` so …

Generally speaking the whole DNA of Flow & Neos loathes the shut-up operator. Flow uses it 36 times in the whole codebase (still too many times), and Neos _not at all_. So, please only if there is no other way. 🙏

mhsdesign:
> So, please only if there is no other way. 

@kdambekalns yes i agree - its just - well this is the alternative:

```php
private function isSymlinkAllowedForUser(): bool
{
  if (PHP_OS_FAMILY !== 'Windows') {
    return true;
  }
  exec('whoami /priv', $output);
  $isUserAllowedToCreateSymlinks = str_contains(join("\n", $output), 'SeCreateSymbolicLinkPrivilege');
  return $isUserAllowedToCreateSymlinks;
}
```

which is way slower and harder to read and possible not really accurate than the stfu `@`

Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
```diff
@@ -166,7 +166,14 @@ private function publish(string $targetPathAndFilename, string $sourcePathAndFil
                 $result = Files::createRelativeSymlink($sourcePathAndFilename, $targetPathAndFilename);
             } else {
                 $temporaryTargetPathAndFilename = $targetPathAndFilename . '.' . Algorithms::generateRandomString(13) . '.tmp';
-                symlink($sourcePathAndFilename, $temporaryTargetPathAndFilename);
+                if (PHP_OS_FAMILY === 'Windows' && is_dir($sourcePathAndFilename)) {
+                    // on windows creating "normal" symlinks is only allowed as admin or with SeCreateSymbolicLinkPrivilege
+                    // in case we only want to symlink a dir, we can instead create directory junctions, which is allowed for non admins
+                    // see https://github.com/git-for-windows/git/wiki/Symbolic-Links
+                    exec(sprintf('mklink /j %s %s', escapeshellarg($temporaryTargetPathAndFilename), escapeshellarg($sourcePathAndFilename)));
```
albe, line 173: Waiting on what edgecase this will break on, because `escapeshellarg` on Windows is haunted (it actually just replaces some characters with whitespace instead of escaping them - see https://www.php.net/manual/en/function.escapeshellarg.php). Also, escaping in Windows command shells is a hot mess in itself:
> The % character is one exception to this rule, it is a valid NTFS filename character but also will be evaluated as a variable name even within a quoted string.
...
> When the shell is running in [EnableDelayedExpansion](https://ss64.com/nt/delayedexpansion.html) mode the ! character is used to denote a variable and so must be escaped (twice)
...
> The \ escape can cause problems with quoted directory paths that contain a trailing backslash because the closing quote " at the end of the line will be escaped \".

https://ss64.com/nt/syntax-esc.html

So we might be better off just preg-replacing the important special chars with `^`, putting double quotes around the two paths and making sure they end with a double-backslash if they do end with a backslash.

fcool (approved):

Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
fcool: Yay. But I guess Marc Henry has a point here - I myself has no windows availabale to test, but I'm fine be reading.

dlubitz:
@mhsdesign Can we close this? I think I didn't saw any complain about this in last 3 years 😬 Or shall we get this merged?

mhsdesign:
Jup now _everyone_ should be using WSL by now and if people still use plain windows with XAMPP they cannot be rescued. 

Lets continue in https://github.com/neos/flow-development-collection/issues/3111 or make a new issue that we might just want to phase out windows support?

Issue #3112, opened 2023-07-16
FEATURE: Improve api / refactor `Scripts::executeCommand` 


mhsdesign:
Scripts::executeCommand has currently an odd, i suppose historically evolved api https://github.com/neos/flow-development-collection/blob/1531a8125ad41e62324c7a85e440c14c1cb768ac/Neos.Flow/Classes/Core/Booting/Scripts.php#L682

1. its hard to get the actual output of the command, currently one needs to `ob_get_clean();` which seems pretty hacky.
2. its not obvious at first what the behavior on error is. The returned status code is actually irrelevant - it will always be true because otherwise we throw an exceptions.
3. the doc commend `$outputResults if false the output of this command is only echoed if the execution was not successful` is lying. In case of an error the output is converted into an exception

mhsdesign:
@kitsunet what do you think about adding a new parameter `&$output` to [`executeCommand`](https://github.com/neos/flow-development-collection/blob/1531a8125ad41e62324c7a85e440c14c1cb768ac/Neos.Flow/Classes/Core/Booting/Scripts.php#L682) (via a bugfix release) or are you fine with:


```php
            ob_start();

            try {
                Scripts::executeCommand(
                    'neos.setup:setup:executeruntimehealthchecks',
                    $this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'Neos.Flow')
                );
            } catch (SubProcessException $subProcessException) {
                exit(1);
            }

            // hack see: https://github.com/neos/flow-development-collection/issues/3112
            $json = ob_get_clean();
```

see https://github.com/neos/setup/blob/439277485f0172dfc3b196ca6f43cd15fbcdbbc5/Classes/RequestHandler/SetupCliRequestHandler.php#L136

kdambekalns:
Agree.

1. Add an `$output` by-reference variable
2. Always return `true`
3. I'd say this works fine as is, and the docblock should be adjusted

mhsdesign:
Fyi the behaviour to return `false` was removed 10 years ago 😂 https://github.com/neos/flow-development-collection/commit/527790b182abe99a2f4be242a6932e6eb0687ee9

can we just remove the return type and have `void`? Hmm or mark the returntype as deprecated?

bwaidelich:
Slightly related: I realized the other day, that an exception in the invoked command leads to an exit code of `0`..

[...]

PR #3118, opened 2023-07-18
BUGFIX: Document `Scripts::executeCommand` properly

mhsdesign:
Related: #3112

`Scripts::executeCommand` has currently an odd, i suppose historically evolved api https://github.com/neos/flow-development-collection/blob/1531a8125ad41e62324c7a85e440c14c1cb768ac/Neos.Flow/Classes/Core/Booting/Scripts.php#L682

1. its not obvious at first what the behavior on error is. The returned status code is actually irrelevant - it will always be true because otherwise we throw an exceptions.
2. the doc commend `$outputResults if false the output of this command is only echoed if the execution was not successful` is lying. In case of an error the output is converted into an exception

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -675,16 +675,16 @@ class Scripts
      * @param array $settings The Neos.Flow settings
      * @param boolean $outputResults Echo the commands output on success
      * @param array $commandArguments Command arguments
-     * @param array|null &$output If the output argument is present, then the specified array will be filled with every line of output from the command. {@see exec()}
      * @return true Legacy return value. Will always be true. A failure is expressed as a thrown exception
      * @throws Exception\SubProcessException The execution of the sub process failed
      * @api
      */
-    public static function executeCommand(string $commandIdentifier, array $settings, bool $outputResults = true, array $commandArguments = [], ?array &$output = []): bool
+    public static function executeCommand(string $commandIdentifier, array $settings, bool $outputResults = true, array $commandArguments = []): bool
     {
         $command = self::buildSubprocessCommand($commandIdentifier, $settings, $commandArguments);
         // Output errors in response
         $command .= ' 2>&1';
+        $output = [];
         exec($command, $output, $result);
         if ($result !== 0) {
             if (count($output) > 0) {
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -673,16 +673,16 @@ class Scripts
      *
      * @param string $commandIdentifier E.g. neos.flow:cache:flush
      * @param array $settings The Neos.Flow settings
-     * @param boolean $outputResults if false the output of this command is only echoed if the execution was not successful
+     * @param boolean $outputResults Echo the commands output on success
      * @param array $commandArguments Command arguments
-     * @return boolean true if the command execution was successful (exit code = 0)
+     * @param array|null &$output If the output argument is present, then the specified array will be filled with every line of output from the command. {@see exec()}
+     * @return true Legacy return value. Will always be true. A failure is expressed as a thrown exception
+     * @throws Exception\SubProcessException The execution of the sub process failed
      * @api
-     * @throws Exception\SubProcessException if execution of the sub process failed
      */
-    public static function executeCommand(string $commandIdentifier, array $settings, bool $outputResults = true, array $commandArguments = []): bool
+    public static function executeCommand(string $commandIdentifier, array $settings, bool $outputResults = true, array $commandArguments = [], ?array &$output = []): bool
     {
         $command = self::buildSubprocessCommand($commandIdentifier, $settings, $commandArguments);
-        $output = [];
         // Output errors in response
         $command .= ' 2>&1';
         exec($command, $output, $result);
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -673,16 +673,16 @@ public static function initializeResources(Bootstrap $bootstrap)
      *
      * @param string $commandIdentifier E.g. neos.flow:cache:flush
      * @param array $settings The Neos.Flow settings
-     * @param boolean $outputResults if false the output of this command is only echoed if the execution was not successful
+     * @param boolean $outputResults Echo the commands output on success
      * @param array $commandArguments Command arguments
-     * @return boolean true if the command execution was successful (exit code = 0)
+     * @param array|null &$output If the output argument is present, then the specified array will be filled with every line of output from the command. {@see exec()}
```
```suggestion
     * @param array &$output If the output argument is present, then the specified array will be filled with every line of output from the command. {@see exec()}
```

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -673,16 +673,16 @@ public static function initializeResources(Bootstrap $bootstrap)
...
+     * @return true Legacy return value. Will always be true. A failure is expressed as a thrown exception
+     * @throws Exception\SubProcessException The execution of the sub process failed
      * @api
-     * @throws Exception\SubProcessException if execution of the sub process failed
      */
-    public static function executeCommand(string $commandIdentifier, array $settings, bool $outputResults = true, array $commandArguments = []): bool
+    public static function executeCommand(string $commandIdentifier, array $settings, bool $outputResults = true, array $commandArguments = [], ?array &$output = []): bool
```
```suggestion
    public static function executeCommand(string $commandIdentifier, array $settings, bool $outputResults = true, array $commandArguments = [], array &$output = []): bool
```

kdambekalns (commented):
I think the `$output` doesn't need to be nullable. It should always be an array, you either leave it out (default value makes it an array) or you pass an array. No?

Neos.Flow/Classes/Core/Booting/Scripts.php
mhsdesign: i think it has to so you can pass an uninitialized variable into it like in `preg_match`

`Scripts::executeCommand(..., $output)` 

with your approach:

`Scripts::executeCommand(..., $output = [])`

kdambekalns: > so you can pass an uninitialized variable

_That_ is ugly as hell. I did not understand what you were saying first, as I never did that. I always initialize variables… but yes, you are right. Passing in an unitialized variable means passing `null`, and that needs to be allowed then.

I still prefer

```php
$output = [];
Scripts::executeCommand(..., $output);
```

over

```php
Scripts::executeCommand(..., $output);
```

Especially as `$output` **is appended to** by `exec()`, so explicitly initializing such variables is safer. 

But hey, do as you like. In this case, following "accepted practices" seems reasonable. 👍

kdambekalns (approved):

kitsunet:
Meh, by reference is fugly, maybe we should just rather introduce a proper new method in an actual release than doing this?

Neos.Flow/Classes/Core/Booting/Scripts.php
mhsdesign: 👍 actually, hate to tell you, but in flow and neos we also usually pass in `$matches` uninitialized:

https://github.com/neos/flow-development-collection/blob/24297f3ab69357855d243a52a16bc0e06f725720/Neos.Flow/Classes/ResourceManagement/ResourceManager.php#L456-L465

and im using this approach also when writing new code ^^

mhsdesign:
i would not oppose. This 10 year old code had it coming to be refactored ;)

Flags like `$outputResults` to control legacy behavior could be removed. Odd return types as well and the methods should not take the flow settings as argument ;) Rather be unary or in this case with only two args.


```php
class SubProcessHelper // to be named
{
    private function __construct(
        private array $settings
    ) {
    }
    
    public static function fromFlowConfiguration(array $settings): self
    {
        return new self($settings);
    }

    /**
     * Executes the given command as a sub-request to the Flow CLI system.
     *
     * @param string $commandIdentifier E.g. neos.flow:cache:flush
     * @param array $commandArguments Command arguments
     * @return array containing every line of output from the command. {@see exec()}
     * @throws Exception\SubProcessException The execution of the sub process failed
     * @api
     */
    public static function executeCommand(string $commandIdentifier, array $commandArguments = []): array;

    public static function executeCommandAsync(string $commandIdentifier, array $commandArguments = []);

    public static function buildPhpCommand(): string;
}
```

bwaidelich:
> maybe we should just rather introduce a proper new method in an actual release than doing this?

I agree. But if we do so, we should consider using https://symfony.com/doc/current/components/process.html internally – I used that in some tests and it seems to be much more stable.

Alternatively https://github.com/reactphp/child-process is a really interesting approach, but I doubt that we are ready to go all in ReactPHP just yet :)

mhsdesign:
I moved our discussion over to https://github.com/neos/flow-development-collection/issues/3112 ;)

As said in slack: the code has been ugly for over 10years - we have all the time to fix it properly in a major

so i will refrain from doing any quick "stupid" hot fix changes for now - eg no new $output parameter.

This pr will only address the wrong doc block.

mhsdesign:
i reverted the new parameter ;)

fcool (approved):

PR #3119, opened 2023-07-18
BUGFIX: improve performance of `Scripts::buildPhpCommand`

mhsdesign:
https://github.com/neos/flow-development-collection/pull/2491 seems to cause a tiny minor performance regression. Booting minimal flow (only configuration manager), and running Scripts::buildPhpCommand went from `0.025s` to `0.044s` on my super fast machine ... it might be a more measurable difference on slower systems.

using `exec` and invoking another php process is just slower than not to.

Following changes were made:

A: runtime cache the buildPhpCommand
B: remove the overhead of a second `exec` by using first native realpath and only fallback in case of an open base dir restriction to `exec`

see https://github.com/neos/flow-development-collection/pull/2491#issuecomment-1639964461

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -60,7 +60,7 @@ use Psr\Http\Message\RequestInterface;
  */
 class Scripts
 {
-    protected static ?string $buildPhpCommand = null;
+    protected static ?string $builtPhpCommand = null;
 
     /**
      * Initializes the Class Loader
@@ -777,8 +777,8 @@ class Scripts
      */
     public static function buildPhpCommand(array $settings): string
     {
-        if (isset(static::$buildPhpCommand)) {
-            return static::$buildPhpCommand;
+        if (isset(static::$builtPhpCommand)) {
+            return static::$builtPhpCommand;
         }
 
         $subRequestEnvironmentVariables = [
@@ -818,7 +818,7 @@ class Scripts
 
         static::ensureWebSubrequestsUseCurrentlyRunningPhpVersion($command);
 
-        return static::$buildPhpCommand = $command;
+        return static::$builtPhpCommand = $command;
     }
 
     /**
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -60,7 +60,8 @@ use Psr\Http\Message\RequestInterface;
  */
 class Scripts
 {
-    protected static ?string $builtPhpCommand = null;
+    /** @var string */
+    protected static $builtPhpCommand = null;
 
     /**
      * Initializes the Class Loader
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -60,6 +60,8 @@
  */
 class Scripts
 {
+    protected static ?string $buildPhpCommand = null;
```
```suggestion
    protected static ?string $builtPhpCommand = null;
```
as it _has been built_ if it's in there. 😎

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -775,6 +777,10 @@ protected static function buildSubprocessCommand(string $commandIdentifier, arra
      */
     public static function buildPhpCommand(array $settings): string
     {
+        if (isset(static::$buildPhpCommand)) {
+            return static::$buildPhpCommand;
```
```suggestion
        if (isset(static::$builtPhpCommand)) {
            return static::$builtPhpCommand;
```

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -812,7 +818,7 @@ public static function buildPhpCommand(array $settings): string
 
         static::ensureWebSubrequestsUseCurrentlyRunningPhpVersion($command);
 
-        return $command;
+        return static::$buildPhpCommand = $command;
```
```suggestion
        return static::$builtPhpCommand = $command;
```

kdambekalns (changes requested):
Mind the tense! `buildPhpCommand` sounds like a flag or command, `builtPhpCommand` at least tells me something has been done already…

Also, the tests fail – might be a "test-only" side effect, but please investigate.

Neos.Flow/Classes/Core/Booting/Scripts.php
mhsdesign: yes mister english teacher ;) ^^

mhsdesign:
> Also, the tests fail – might be a "test-only" side effect, but please investigate.

all adjusted ;)

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -60,6 +60,8 @@
  */
 class Scripts
 {
+    protected static ?string $builtPhpCommand = null;
```
mhsdesign:
```suggestion
    /** @var string */
    protected static $builtPhpCommand = null;
```

fcool (approved):

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -849,8 +856,13 @@ protected static function ensureCLISubrequestsUseCurrentlyRunningPhpBinary($phpB
             return;
         }
 
-        exec(PHP_BINARY . ' -r "echo realpath(PHP_BINARY);"', $output);
-        $realPhpBinary = $output[0];
+        // stfu to avoid possible open_basedir restriction https://github.com/neos/flow-development-collection/pull/2491
+        $realPhpBinary = @realpath(PHP_BINARY);
```
mhsdesign, line 860: alternatively we can just check if `ini_get('open_basedir')` is not empty and then assume `realpath` will not work ...

kdambekalns (approved):

PR #3126, opened 2023-08-02
BUGFIX: Pin doctrine/orm to <2.16.0

dlubitz:
After release of 2.16.0 of doctrine/orm the order of created objects has changed. 
See: https://github.com/doctrine/orm/issues/10864

From Slack: https://neos-project.slack.com/archives/C050KKBEB/p1690915423960539

Until this got fixed or we could fix this on our end we need to pin to a version below 2.16.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "psr/http-server-middleware": "^1.0",
         "psr/http-server-handler": "^1.0",
         "ramsey/uuid": "^3.0 || ^4.0",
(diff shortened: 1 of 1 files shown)


ahaeslich (approved):

albe:
~~Post-merge approve :)~~

Err: Need to adjust the composer.json in the Package(s) as well, e.g. https://github.com/neos/flow-development-collection/blob/7.3/Neos.Flow/composer.json#L39 - always did that wrong myself as well

PR #3128, opened 2023-08-03
BUGFIX: Deduplication on prePersist

dlubitz:
Doctrine refactored their IdentityMap, so it checks for duplicated objects already on persist instead of on flush before.

The way Flow works with ValueObjects allows creating multiple object instances, but they are equal on there persistance identifier. To prevent doctrine from throwing exceptions, because the objects are equal on their persistance identifier we remove one value objects from the IdentityMap if there is an equal one to persist.

```
        $post1 = new Fixtures\Post();
        $post1->setAuthor(new Fixtures\TestValueObject('Some Name'));

        $post2 = new Fixtures\Post();
        $post2->setAuthor(new Fixtures\TestValueObject('Some Name'));

        $this->postRepository->add($post1);
        $this->postRepository->add($post2); // <-- doctrine would throw an exception here
        $this->persistenceManager->persistAll();
```

Until now, flow removes the duplicated value objects on before flush from the list of scheduled insertions, but this is to late now.

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationAndDeDuplicationListener.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationAndDeDuplicationListener.php
@@ -75,7 +75,7 @@ class ObjectValidationAndDeDuplicationListener
 
         if ($classSchema !== null && $classSchema->getModelType() === ClassSchema::MODELTYPE_VALUEOBJECT) {
             foreach ($identityMap[$className] ?? [] as $objectInIdentityMap) {
-                if ($objectInIdentityMap == $object) {
+                if ($this->persistenceManager->getIdentifierByObject($objectInIdentityMap) === $this->persistenceManager->getIdentifierByObject($object)) {
                     $unitOfWork->removeFromIdentityMap($objectInIdentityMap);
                 }
             }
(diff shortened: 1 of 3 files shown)


albe:
Looks good and makes things work with doctrine/orm 2.16 again 👍

fcool:
Have you checked if we can remove our logic with the most ugly unset now?

If so,  the general quality would rise and I would love to have the code deleted. It proved to be quite whacky and neeeded several patches in the last years.

This is the block I'm reffering:
https://github.com/neos/flow-development-collection/blob/e0f4aa0bd69982bc8bf03724070821ad3e9f39f1/Neos.Flow/Classes/Persistence/Doctrine/ObjectValidationAndDeDuplicationListener.php#L87

dlubitz:
@fcool Yes, I checked that. But this is still needed to check for exisiting VOs in the database. As the VOs are not properly loaded through the entity manager, it doesn't know about them. So it leads to a duplicated entry in the database.

fcool (approved):

kdambekalns (approved):
Looks good and sounds reasonable. Too bad the functionality is "duplicated" in a way…

dlubitz:
@kdambekalns Duplicated for deduplication, so... it's fine I guess 😄

PR #3130, opened 2023-08-05
TASK: Update .phpstorm.meta.php to work with psr `ContainerInterface`

mhsdesign:
Currently only php storm autocompletion is supported for the `\Neos\Flow\ObjectManagement\ObjectManagerInterface::get`, but sometimes you dont want to pass the whole object manager around but only a psr `\Psr\Container\ContainerInterface` (which our `ObjectManagerInterface` extends, as we are psr compatible.

This fix allows the same autocompletion for simple psr container.

Test it like:

```php
/** @var $objectMngr \Neos\Flow\ObjectManagement\ObjectManagerInterface */
$objectMngr->

// and

/** @var $container \Psr\Container\ContainerInterface */
$container->
```

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/.phpstorm.meta.php
+++ b/Neos.Flow/.phpstorm.meta.php
@@ -5,6 +5,11 @@
  */
 
 namespace PHPSTORM_META {
+    override(
+        \Psr\Container\ContainerInterface::get(),
+        map(['' => '@'])
+    );
+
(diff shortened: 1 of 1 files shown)


kdambekalns (approved):

mhsdesign:
On the other hand it we dont need to provide this helper.
E.g. for phpstan some plugin is required as well, so if someone uses the `ContainerInterface` directly it will not be thaat straight forward either.

I can recommend to introduce a utility method like:

```php
/**
 * @template T of object
 *
 * @return T
 */
private function getObject(string $className): object
{
    return $this->container->get($className);
}
```

PR #3133, opened 2023-08-11
BUGFIX: add 'validationErrorMessage' as a supported option

### LINKED ISSUE #3691: BUG: Regexvalidator option throws error on apply

### Current Behavior

If I use the `validationErrorMessage` option for the `Neos.Neos/Validation/RegularExpressionValidator` on property, introduced here: https://github.com/neos/neos-ui/pull/2996.

And then try to apply a change to this property, I get an `Unsupported validation option(s) found`-error:
Screenshot_2023-06-30_at_16 43 10

### Expected Behavior

I would expect to be able to use this option without issues.

### Steps To Reproduce

Add the validationErrorMessage option to a RegularExpressionValidator, used as a property validator in the yaml NodeType definition.

Example:
```
      validation:
        'Neos.Neos/Validation/RegularExpressionValidator':
          regularExpression: /^(?:http(s)?:\/)?(\/)+[\w\-\.~:\/\?#[\]@!\$&\(\)\*\+=]*$/i
          validationErrorMessage: 'Please enter a valid relative or absolute URL.'
```

Try to apply a change to the property using this validation.
=> Error

### Environment

_No response_

### Anything else?

_No response_

MiauzGenau:
Add 'validationErrorMessage' as an supported option, for the RegularExpressionValidator.

resolved: neos/neos-ui#3691 

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Validation/Validator/RegularExpressionValidator.php
+++ b/Neos.Flow/Classes/Validation/Validator/RegularExpressionValidator.php
@@ -24,7 +24,8 @@ class RegularExpressionValidator extends AbstractValidator
      * @var array
      */
     protected $supportedOptions = [
-        'regularExpression' => ['', 'The regular expression to use for validation, used as given', 'string', true]
+        'regularExpression' => ['', 'The regular expression to use for validation, used as given', 'string', true],
+        'validationErrorMessage' => ['', 'The error message to show for validation, if the regular expression validation fails', 'string', false]
     ];
 
     /**


crydotsnake:
Thanks for the fix.

Is this bug related too Neos 9? otherwise the PR should target the Neos 7.3 branch.

MiauzGenau:
Create new PR on 7.3 branch here: https://github.com/neos/flow-development-collection/pull/3134

fcool:
If the aim is to change the error message, there are plenty of possibilities. You could use the code and provide an own translation. Or you could override the translation flow provides. Why would you have the message changeable for each validator instance? 
And given the use case is convincing: Why only with the regex validator? And not with all validators?

Neos.Flow/Classes/Validation/Validator/RegularExpressionValidator.php
```diff
@@ -24,7 +24,8 @@ class RegularExpressionValidator extends AbstractValidator
      * @var array
      */
     protected $supportedOptions = [
-        'regularExpression' => ['', 'The regular expression to use for validation, used as given', 'string', true]
+        'regularExpression' => ['', 'The regular expression to use for validation, used as given', 'string', true],
+        'validationErrorMessage' => ['', 'The error message to show for validation, if the regular expression validation fails', 'string', false]
```
kitsunet, line 28: This is only enough to fullfill the needs of the Neos validation system so we do not get an exception here, but in fact this new supported option should impact eg.
https://github.com/neos/flow-development-collection/pull/3133/files#diff-2ecda0a3f7d78fb28fd5cffe20976014c4281e728f4193e505844d380144e9f1R43

otherwise it makes absolutely no sense from the PHP side of things, as nothing happens when I set this option.

PR #3134, opened 2023-08-11
BUGFIX: add 'validationErrorMessage' as a supported option

MiauzGenau:
Add 'validationErrorMessage' as an supported option, for the RegularExpressionValidator.

resolved: https://github.com/neos/neos-ui/issues/3691

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Validation/Validator/RegularExpressionValidator.php
+++ b/Neos.Flow/Classes/Validation/Validator/RegularExpressionValidator.php
@@ -24,7 +24,8 @@ class RegularExpressionValidator extends AbstractValidator
      * @var array
      */
     protected $supportedOptions = [
-        'regularExpression' => ['', 'The regular expression to use for validation, used as given', 'string', true]
+        'regularExpression' => ['', 'The regular expression to use for validation, used as given', 'string', true],
+        'validationErrorMessage' => ['', 'The error message to show for validation, if the regular expression validation fails', 'string', false]
     ];
 
     /**


kdambekalns (changes requested):
And what about actually using it, too?

fcool:
If the aim is to change the error message, there are plenty of possibilities. You could use the code and provide an own translation. Or you could override the translation flow provides. Why would you have the message changeable for each validator instance?
And given the use case is convincing: Why only with the regex validator? And not with all validators?

Are you sure, this is a bugfix and not a feature?

kdambekalns:
> If the aim is to change the error message, there are plenty of possibilities.

Obviosuly the option was added to the UI in https://github.com/neos/neos-ui/pull/2996 but that (now?) breaks with Flow.

As to "why only here", then probably because most other default messages are good enough, but "does not match pattern" holds next to no useful information. To which I can agree. 🙃

fcool:
Ah... and that's why it (the new message) is not used in Flow itself.

But that feels like a bug in Neos, doesn't it? Flow shouldn't allow a property to be changed, because of Neos setting it "accidently", to fire own logic? Maybe we should add an additional validator in Neos then?

kdambekalns:
> But that feels like a bug in Neos, doesn't it? Flow shouldn't allow a property to be changed, because of Neos setting it "accidently", to fire own logic? Maybe we should add an additional validator in Neos then?

- Well, I can only assume it used to work–and now breaks due to Flow being more strict? 🤷‍♂️
- If this fixes a bug, but introduces weird behaviour (the message being "supported", but not used), I am against it.
- An additional validator feels weird, given the usecase is valid for (pure) Flow applications, too.

MiauzGenau:
I created a new PR: https://github.com/neos/flow-development-collection/pull/3141
As a FEATURE draft.
For now I only did it for the RegularExpressionValidator, if this looks good/ better to you I would also adjust the other validators and close this PR.

mhsdesign:
> But that feels like a bug in Neos, doesn't it? Flow shouldn't allow a property to be changed, because of Neos setting it "accidently", to fire own logic? Maybe we should add an additional validator in Neos then?

yeeeees true. The current situation is suboptimal and flow should not suffer from this. Agreed ^^.
We are currently evaluating better concepts than using flows validators for node properties, so they can be validated in the cr: https://github.com/neos/neos-development-collection/issues/4631

The "auto fire" was introduced via https://github.com/neos/neos-ui/pull/2351 and actually caused much headache, as the javascript validators and phpvalidators are not always the same and validate differently. This new message feature is just an example.

As this feels indeed quite hacky on php land

> If the aim is to change the error message, there are plenty of possibilities. You could use the code and provide an own translation. Or you could override the translation flow provides. Why would you have the message changeable for each validator instance?

i would say the triggering logic in the neos ui php code should be adjusted. As this neos ui property server validation is already super hacky (see https://github.com/neos/neos-ui/pull/2351) we should fix it there (as bugfix) and keep all the ugly code at one place and when the time comes remove it.

TLTR: I would like to close this in favour of fixing it in the Ui

kitsunet (changes requested):
Pre-emptive block as we definitely should not merge this, if anything the feature PR linked in here or fix it in the Neos UI (probabably better). This change would introduce an option that has absoluitely no use or result in Flow and will be very misleading.

lorenzulrich:
@mhsdesign How would you fix this in the UI? It seems that all validator options are passed to the PHP validator, regardless of whether they are "UI-only" or also valid for the PHP validation.

Wouldn't it make sense to distinguish between the kind of option?

```
'Vendor.Site:Content.YourNodeTypeName':
  properties:
    lastName:
      type: string
      ui:
		label: 'Last Name'
        inspector:
          group: 'contactData'
      validation:
        'Neos.Neos/Validation/RegularExpressionValidator':
          regularExpression: '...'
          neosUi:
            validationErrorMessage: 'You must set a last name.'
```

mhsdesign:
@lorenzulrich i moved the discussion and my answer over here: https://github.com/neos/neos-ui/issues/3691#issuecomment-1898100119

Issue #3137, opened 2023-08-15
BUG: interceptedRequest lacks critical data


kdambekalns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A [longer description is available](https://discuss.neos.io/t/redirect-to-intercepted-request-resulting-in-wrong-url/6383?u=kdambekalns), but it boils down to this:

- When an intercepted request is used as redirect target after login (using `$this->redirectToRequest()` in an MVC `AbstractController` subclass), any argument namespace from the "current" request is applied. This leads to "wrong" URLs being built.
- It is not possible to simply do `$this->uriBuilder->setRequest($originalRequest)` at this point, as this leads to an error when `getHttpRequest()` is called on said request.
- That is caused by the serialized request dropping it's (HTTP) _parent request_ (in [`ActionRequest.__sleep()`](https://github.com/neos/flow-development-collection/blame/6f8f0d747f2a8d5fa513f337cc079cda2b723c1f/Neos.Flow/Classes/Mvc/ActionRequest.php#L731-L744)) since https://github.com/neos/flow-development-collection/commit/44107b85f01b1f8be3c449b829292aeda9a3c13b (sadly the issue for this is lost in the digital nirvana…) and the _root request_ (which was added way after said `__sleep()` method.)

### Expected Behavior

It should be possible to simply use `$this->redirectToRequest($originalRequest)` – it clearly defines the intended target.

### Steps To Reproduce

See linked post above…

### Environment

```markdown
- Flow: 7.3+
- PHP: does not matter
```

kdambekalns:
There are two potential ways to fix this:

- **Serialize root and parent request** and then **set the intercepted request** on the `UriBuilder`. I tried this locally, and it solves the issue. No further tests for side effects have been done, though.
- **Store the requested URI** instead of the request. As we [only intercept GET](https://github.com/neos/flow-development-collection/pull/1695) requests and an URI is all we need for redirection, that would be enough.

kdambekalns:
PR #3138 implements the first potential solution described above…

kdambekalns:
The second solution would be breaking (if there is no intercepted **request** anymore) or would need to deprecate the current handling and possible store an `interceptedRequestUri` along with the request. Removal of the current way of handling this could then be done for 9.0 (or 10.0, or…)

[...]

PR #3138, opened 2023-08-15
BUGFIX: Fix `redirectToRequest()` to use given request

kdambekalns:
Makes it possible to really use `redirectToRequest()` with an "intercepted request" from authentication handling.

See #3137 and [this discuss.neos.io post](https://discuss.neos.io/t/redirect-to-intercepted-request-resulting-in-wrong-url/6383?u=kdambekalns) for details.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Mvc/ActionRequest.php
+++ b/Neos.Flow/Classes/Mvc/ActionRequest.php
@@ -736,10 +736,19 @@ class ActionRequest implements RequestInterface
      */
     public function __sleep()
     {
-        $properties = ['controllerPackageKey', 'controllerSubpackageKey', 'controllerName', 'controllerActionName', 'arguments', 'internalArguments', 'pluginArguments', 'argumentNamespace', 'format', 'dispatched'];
-        if ($this->parentRequest instanceof ActionRequest) {
-            $properties[] = 'parentRequest';
-        }
-        return $properties;
+        return [
+            'controllerPackageKey',
+            'controllerSubpackageKey',
+            'controllerName',
+            'controllerActionName',
+            'arguments',
+            'internalArguments',
+            'pluginArguments',
+            'argumentNamespace',
+            'format',
+            'dispatched',
+            'parentRequest',
+            'rootRequest'
+        ];
     }
 }
--- a/Neos.Flow/Classes/Mvc/ActionRequest.php
+++ b/Neos.Flow/Classes/Mvc/ActionRequest.php
@@ -729,8 +729,7 @@ class ActionRequest implements RequestInterface
     }
 
     /**
-     * We provide our own __sleep method, where we serialize all properties *except* the parentRequest if it is
-     * a HTTP request -- as this one contains $_SERVER etc.
+     * We provide our own __sleep method to control serialization.
      *
      * @return array
      */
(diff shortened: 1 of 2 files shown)


kdambekalns:
This "brings back" data that @skurfuerst explicitly excluded with https://github.com/neos/flow-development-collection/commit/44107b85f01b1f8be3c449b829292aeda9a3c13b back in 2013. Probably a thorough check for negative side-effects is advisable…

bwaidelich:
Potentially breaking if the HTTP request contains attributes that are not serializable

kdambekalns:
> Potentially breaking if the HTTP request contains attributes that are not serializable

Anything "suspicious" you expect?

bwaidelich:
> Anything "suspicious" you expect?

Not really, and usually(tm) you don't put complex objects into request attributes.

But the other day I did something like this:
```php
$request = $request->withAttribute('AuthenticatedAccount', $this->authenticatedAccount);
```
in a behat feature context in order to simulate authenticated requests

Neos.Flow/Classes/Mvc/ActionRequest.php
```diff
@@ -736,10 +736,19 @@ public function __clone()
      */
     public function __sleep()
```
mhsdesign, line 736: the doc block would need some adjustment as well ;)

kdambekalns: ✅

kdambekalns:
> Not really, and usually(tm) you don't put complex objects into request attributes.

See!? 😎

> But the other day I did something like this:

Ok. Now, would you expect this to be still there when you get that back as "intercepted request"? I am pretty sure I would want anything I put into a drawer to still be there after closing and opening it again. 😉

But yes, I guess this can have weird side effects. And I'd still love to check that historical issue behind the change that @skurfuerst did 10 years ago that added the `_sleep()`. 🤷

How do we proceed?

- Hope this works for real (and not just in the tests)?
- Hope for many people to actually test it in their actual projects?
- Or do we scrap it and go for "store just the URI"?
- Or do we go for your idea of storing the HTTP request? That at least would bring us back to square one ("_Potentially breaking if the HTTP request contains attributes that are not serializable_")

skurfuerst:
Hey everybody,

I just read the thread and the pull request I appearently did 10 years ago 🤣 😄  From reading the code, I could imagine the original problem could resurface, but on the other hand things*tm have changed quite a bit in the last 10 years; f.e. we use PSR Request which will behave differently.

To me the cleanest solution would be storing the request URI (as you all suggested) and deprecate the other one now + remove it for Flow 9; but I could also live with trying out the change above in practice and see if we see any major backlash. F.e. @fcool has big Flow projects where issues could surface by changing this.

All the best, Sebastian

bwaidelich:
> Now, would you expect this to be still there when you get that back as "intercepted request"?

No, but the code would now fail when trying to serialize without a simple escape hatch (probably not in my case and I don't intercept requests in my example anyways, but in similar situations maybe?)

> To me the cleanest solution would be storing the request URI (as you all suggested) and deprecate the
> other one now + remove it for Flow 9;

+1
IMO we don't even need to remove it with Flow 9 already because it does not hurt to store both (IMO it _does_ hurt to start a session for that, but that's a different topic)

PR #3140, opened 2023-08-24
BUGFIX: Only set distinct on count clause if explicitely set to improve performance

Sebobo:
F.e. Postgres has performance issues with large datasets and the DISTINCT clause. In a test this change reduced the query time of a count query for ~900.000 entities by >80%.

In a custom project this affected their Neos Media.UI in which the following results were found:

* Count all assets |  580ms ->  260ms
* Query 20 assets  |  690ms ->  350ms
* Query 100 assets |  990ms -> 650ms
* Module load | 1900ms -> 1400ms

**Review instructions**

Everything should work the same, as https://github.com/neos/flow-development-collection/pull/415 already sets the distinct flag where (possibly) necessary.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/CountWalker.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/CountWalker.php
@@ -49,7 +49,7 @@ class CountWalker extends TreeWalkerAdapter
 
         $AST->selectClause->selectExpressions = [
             new SelectExpression(
-                new AggregateExpression('count', $pathExpression, true),
+                new AggregateExpression('count', $pathExpression, $AST->selectClause->isDistinct),
                 null
             )
         ];


albe:
I guess that's a good solution 👍. Regarding #415 I just noticed that it was actually reverted due to issues: #1293, but that should not have too much influence on this. Implicit joins were returning duplicates before, so counts on them were off. Now they're at least faster at that :D

albe (approved):

albe:
Or wait, I was never checking a `$query->count()` on implicit join results, but just the `count($result)`... https://github.com/neos/flow-development-collection/pull/1293/files#diff-b389146f7ff3bedefca76487b8b3421d40ebf42cb4f24e854d129897dacc03a8L269
So maaaaybeee this is indeed a bit more breaky than my last comment suggests. But breaky in the sense it behaves differently than before, though more logical, i.e. `->count()` would return 2 now, when there is a duplicate result/2 result rows.

Sebobo:
Setting a query as distinct should always be explicitly set and not assumed in any case?
Doctrine already does more magic than wished for in some cases.

Sebobo:
Just got results from my customers production system (Postgres, 850.000 assets) based only on this change:

Bildschirmfoto 2023-08-25 um 09 04 16

kdambekalns (approved):
LGTM

kitsunet (approved):

Issue #3142, opened 2023-08-29
BUG: "Only variables should be passed by reference" error in DI


bwaidelich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using `Objects.yaml` to inject [virtual objects](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/ObjectManagement.html#sect-virtual-objects) into other objects, the following PHP notice is generated (leading to an exception at build time):

> Notice: Only variables should be passed by reference in Packages/Framework/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php line 294

### Expected Behavior

..well, no exception of course :)

### Steps To Reproduce

Create an `Objects.yaml` configuration like this:

```yaml
'SomePackage:SomeVirtualObject':
  className: Some\Package\SomeObject
  scope: singleton
  factoryObjectName: Some\Package\SomeFactory

Some\Package\SomeOtherObject:
  scope: singleton
  arguments:
    1:
      object: 'SomePackage:SomeVirtualObject'
```

### Environment

- Flow: 7.3-8.3
- PHP: 8.x

### Anything else?

The root cause is the signature of [Arrays::getValuesByPath()](https://github.com/neos/flow-development-collection/blob/dc7214653b8e1ca9e4e3f0b8fc5cfb293bf5eac7/Neos.Utility.Arrays/Classes/Arrays.php#L202) that expects the array to be passed by reference for some reason.

In version 9 the issue is most probably be solved, because we replaced

```php
$settings = Arrays::getValueByPath($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS), array_shift($settingPath));
```

with

```php
$settings = $this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, array_shift($settingPath));
```

([source](https://github.com/neos/flow-development-collection/blob/9.0/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php#L306))

PR #3143, opened 2023-08-29
BUGFIX: Prevent method call in ProxyClassBuilder

bwaidelich:
Fixes: #3142


kdambekalns (approved):
by 👀

mhsdesign (approved):
we dont seem to need this change anymore technically right since https://github.com/neos/flow-development-collection/pull/3144 is merged?

but its a tiny clean up so why not +1

Issue #3043, opened 2023-05-10
FEATURE: Settings Framework: Better/readable setting validation


mhsdesign:
wouldnt it be interesting to have better runtime setting validation for flow?

currently it mostly looks like this:

```php
@Flow\InjectConfiguration(package="Neos.Neos", "foo")
protected string $foo;
```

and in case of an faulty type we would get a runtime error like: cannot assign type null to type string in property $foo.

It would be great if the exception message would contain the actual setting path which was set faulty to make the experience better.

cases where we inject a whole "subtree" of the config, an array mostly work fine at first:

```php
@Flow\InjectConfiguration(package="Neos.Neos", "foo")
protected array $foo;
```

but later when traversing the array we implicitly assume the structure of the array without having validated it, which leads to really obscure error messages.

another case for obscure errors are cases where we currently pass an array of options to a view or a cache backend: 
```yaml
MY_CACHE:
  backend: Neos\Cache\Backend\PdoBackend
  backendOptions:
    username: '%env:SOME_ENV%'
```
in case the enviroment is not set, the fallback is false which leads to an odd error at first:

```
Argument 1 passed to Neos\Cache\Backend\PdoBackend::setUsername() must be of the type string, bool given, called neos/cache/Classes/Backend/AbstractBackend.php on line 102
```

i would rather like to see:

```
expected configuration MY_CACHE.backendOptions.username to be string, false given.
```

all the mentioned cases i believe are solveable by passing a ProxySetting object around, which nows its path and value, and can be asserted to resolve to for example a string. If an unexpected value is given at runtime we can (knowing the config path) throw a nice readable exception.

The idea is not mine, but ive seen it from @grebaldi in his experiment: https://github.com/PackageFactory/extractor

What are your thoughs?

grebaldi:
> [...] and in case of an faulty type we would get a runtime error like: cannot assign type null to type string in property $foo.

This could be analysed statically too :)

mhsdesign:
See for example usecase our new ContentRepositoryRegistry https://github.com/neos/neos-development-collection/blob/33e8c220628395e8bf03538cfa1ebf7c4852cca6/Neos.ContentRepositoryRegistry/Classes/ContentRepositoryRegistry.php#L131

We validate a lot of settings ourselves, and i think the framework should help us, so that we dont have to use so many primitive checks.

What do you say? Is it flows responsibility or not?

bwaidelich:
[...]

PR #3149, opened 2023-09-02
FEATURE: Add `getAccessorByPath` to `Neos\Utility\Arrays` for type safe accessing of array values

### LINKED ISSUE #3164: FEATURE: Add accessor for type safe access to `Arrays` utility

### Description

It is nice to access arrays like Settings by path but dealing with unexpected types requires a lot of checking. Especially when code is to be statically analyzed.

### Possible Solution

Add an accessor to the arrays utility that does the type checking and throws errors when expectations are not matched.

mficzel:
_**Please note that this is an experimental feature and the API is not stable yet.**_

The array utility allows to create a type safe accessor via `Arrays::getAccessorByPath($arrayValue, 'your.path')`. The accessor provides the following methods that will either return the requested type or throw a `\UnexpectedValueException`.  

* `int(): int`
* `float(): float`
* `number(): int|float`
* `string(): string`
* `classString(): string` - with annotation for class-string
* `array(): array`
* `instanceOf(string $className): object` - with annotation for dynamic type
* `intOrNull(): ?int`
* `floatOrNull(): ?float`
* `numberOrNull(): null|int|float`
* `stringOrNull(): ?string`
* `classStringOrNull(): ?string` - with annotation for class-string | null
* `arrayOrNull(): ?array`
* `instanceOfOrNull(string $className): ?object` - with annotation for dynamic type | null

This will allow to write code that accesses settings via pathes without checking every level for existence still beeing type safe and accessible for static analysis.

This can be used together with settingInjection.

```php
public function injectSettings(array $settings): void
{
   $this->limit = Arrays::getAccessorByPath($settings, 'limit')->intOrNull();
}
```

Resolves: #3164

**Review instructions**

It may look inefficient to manually throw TypeErrors that in many cases would be thrown automatically because of the declared return types. However this is not a performance issue as those are never on the happy-path and the created TypeError provides additional informations to help understand and fix problems faster.

Inspired by https://github.com/PackageFactory/extractor

**Checklist**

### DIFF
--- a/Neos.Utility.Arrays/Classes/Arrays.php
+++ b/Neos.Utility.Arrays/Classes/Arrays.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Neos\Utility;
 
 /*
@@ -194,28 +195,38 @@ abstract class Arrays
     /**
      * Returns the value of a nested array by following the specifed path.
      *
-     * @param array &$array The array to traverse as a reference
+     * @param array $array The array to traverse
      * @param array|string $path The path to follow. Either a simple array of keys or a string in the format 'foo.bar.baz'
      * @return mixed The value found, NULL if the path didn't exist (note there is no way to distinguish between a found NULL value and "path not found")
-     * @throws \InvalidArgumentException
      */
-    public static function getValueByPath(array &$array, $path)
+    public static function getValueByPath(array $array, array|string $path): mixed
     {
-        if (is_string($path)) {
-            $path = explode('.', $path);
-        } elseif (!is_array($path)) {
-            throw new \InvalidArgumentException('getValueByPath() expects $path to be string or array, "' . gettype($path) . '" given.', 1304950007);
-        }
-        $key = array_shift($path);
-        if (isset($array[$key])) {
-            if (count($path) > 0) {
-                return (is_array($array[$key])) ? self::getValueByPath($array[$key], $path) : null;
-            } else {
-                return $array[$key];
-            }
-        } else {
+        $pathSegments = is_string($path) ? explode('.', $path) : $path;
+        $pathSegment = array_shift($pathSegments);
+
+        if (!isset($array[$pathSegment])) {
             return null;
         }
+        if ($pathSegments === []) {
+            return $array[$pathSegment];
+        }
+        return is_array($array[$pathSegment])
+            ? self::getValueByPath($array[$pathSegment], $pathSegments)
+            : null;
+    }
+
+    /**
+     * Returns a type safe accessor for a value in a nested array by following the specifed path.
+     *
+     * @param array $array The array to traverse
+     * @param array|string $path The path to follow. Either a simple array of keys or a string in the format 'foo.bar.baz'
+     * @return ValueAccessor
+     */
+    public static function getAccessorByPath(array $array, array|string $path): ValueAccessor
+    {
+        $pathinfo = is_array($path) ? implode('.', $path) : $path;
+        $value = self::getValueByPath($array, $path);
+        return new ValueAccessor($value, $pathinfo);
     }
 
     /**
--- a/Neos.Utility.Arrays/Classes/ValueAccessor.php
+++ b/Neos.Utility.Arrays/Classes/ValueAccessor.php
@@ -23,9 +23,9 @@ class ValueAccessor
     ) {
     }
 
-    private function createTypeError($message): \TypeError
+    private function createTypeError($message): \UnexpectedValueException
     {
-        return new \TypeError(get_debug_type($this->value) . ' ' . $message . ($this->pathinfo ? ' in path ' . $this->pathinfo : ''));
+        return new \UnexpectedValueException(get_debug_type($this->value) . ' ' . $message . ($this->pathinfo ? ' in path ' . $this->pathinfo : ''));
     }
 
     public function int(): int
@@ -86,10 +86,10 @@ class ValueAccessor
      */
     public function instanceOf(string $className): object
     {
-        if (is_a($this->value, $type, false)) {
+        if ($this->value instanceof $className) {
             return $this->value;
         }
-        throw $this->createTypeError(sprintf('is not of type class %s', $type));
+        throw $this->createTypeError(sprintf('is not an instance of %s', $className));
     }
 
     public function intOrNull(): ?int
@@ -150,9 +150,9 @@ class ValueAccessor
      */
     public function instanceOfOrNull(string $className): ?object
     {
-        if (is_a($this->value, $type, false) || is_null($this->value)) {
+        if ($this->value instanceof $className || is_null($this->value)) {
             return $this->value;
         }
-        throw $this->createTypeError(sprintf('is not of type class %s or null', $type));
+        throw $this->createTypeError(sprintf('is not an instance of %s or null', $className));
     }
 }
--- a/Neos.Utility.Arrays/Classes/ValueAccessor.php
+++ b/Neos.Utility.Arrays/Classes/ValueAccessor.php
@@ -1,0 +1,158 @@
+<?php
+
+namespace Neos\Utility;
+
+/*
+ * This file is part of the Neos.Utility.Arrays package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+/**
+ * Type safe accessing of values from nested arrays
+ */
+class ValueAccessor
+{
+    public function __construct(
+        public readonly mixed   $value,
+        public readonly ?string $pathinfo = null
+    ) {
+    }
+
+    private function createTypeError($message): \TypeError
+    {
+        return new \TypeError(get_debug_type($this->value) . ' ' . $message . ($this->pathinfo ? ' in path ' . $this->pathinfo : ''));
+    }
+
+    public function int(): int
+    {
+        if (is_int($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not an integer");
+    }
+
+    public function float(): float
+    {
+        if (is_float($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not a float");
+    }
+
+    public function number(): int|float
+    {
+        if (is_int($this->value) || is_float($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not a number");
+    }
+
+    public function string(): string
+    {
+        if (is_string($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not a string");
+    }
+
+    /**
+     * @return class-string
+     */
+    public function classString(): string
+    {
+        if (is_string($this->value) && class_exists($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not a class-string");
+    }
+
+    public function array(): array
+    {
+        if (is_array($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not an array");
+    }
+
+    /**
+     * @template T
+     * @param class-string<T> $type
+     * @return object&T
+     */
+    public function object(string $type): object
+    {
+        if (is_a($this->value, $type, false)) {
+            return $this->value;
+        }
+        throw $this->createTypeError(sprintf('is not of type class %s', $type));
+    }
+
+    public function intOrNull(): ?int
+    {
+        if (is_int($this->value) || is_null($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not an integer or null");
+    }
+
+    public function floatOrNull(): ?float
+    {
+        if (is_float($this->value) || is_null($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not a float or null");
+    }
+
+    public function numberOrNull(): null|int|float
+    {
+        if (is_int($this->value) || is_float($this->value) || is_null($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not a number or null");
+    }
+
+    public function stringOrNull(): ?string
+    {
+        if (is_string($this->value) || is_null($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not a string or null");
+    }
+
+    /**
+     * @return class-string|null
+     */
+    public function classStringOrNull(): ?string
+    {
+        if ((is_string($this->value) && class_exists($this->value)) || is_null($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not a class-string");
+    }
+
+    public function arrayOrNull(): ?array
+    {
+        if (is_array($this->value) || is_null($this->value)) {
+            return $this->value;
+        }
+        throw $this->createTypeError("is not an array or null");
+    }
+
+    /**
+     * @template T
+     * @param class-string<T> $type
+     * @return (object&T)|null
+     */
+    public function objectOrNull(string $type): ?object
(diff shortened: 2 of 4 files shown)


mficzel:
~~Tests are not yet complete. Will finalize if the concept is seen as valuable.~~

@grebaldi i tried to apply a very simplified rendition of your extractor pattern to the arrays utility. Do you think this makes sense?

grebaldi:
@mficzel:
> Do you think this makes sense?

Definitely! :)

I like the idea of checking the type at the leaf-end. It's less cluttery than the original extractor.

The only thing I'd like to suggest is to pass along the `$path` (as optional), so `ValueAccessor` can incorporate it in its error messages :)

mhsdesign:
Nice idea to move the https://github.com/PackageFactory/extractor to the core somewhat ;)

I also thought a bit along the side how we could provide better flow setting validation please let me know what you think of https://github.com/neos/flow-development-collection/issues/3043

And currently im also experimenting with type safe accessors for fusion runtime variables: https://github.com/neos/neos-development-collection/issues/4475

mficzel:
@mhsdesign this was actually the result of me looking at your fusion runtime context prs and not liking to introduce type safe array access only there. How about using this Accessor for the fusion runtime, Neos\Utility\Arrays is already used?

Neos.Utility.Arrays/Tests/Unit/ArraysTest.php
```diff
@@ -108,6 +108,103 @@ public function getValueByPathReturnsNullIfThePathHasMoreSegmentsThanTheGivenArr
...
+     */
+    public function getAccessorByPathTypeErrorContainsPathForNonMatchingTypes()
+    {
+        $array = ['Foo' => ['Bar' => ['Baz' => 'the value']]];
+        try {
+            self::assertNull(Arrays::getAccessorByPath($array, ['Foo', 'Bar', 'Baz'])->int());
+        } catch (\Error $error) {
+            self::assertInstanceOf(\TypeError::class, $error);
+            self::assertStringEndsWith('in path Foo.Bar.Baz', $error->getMessage());
+        }
```
bwaidelich: The `self::assertNull()` is probably not evaluated.
Also, instead of `try/catch` you could use the built-in methods to assert exceptions:
```suggestion
        $this->expectException(\TypeError::class);
        $this->expectExceptionMessage('in path Foo.Bar.Baz');
        Arrays::getAccessorByPath($array, ['Foo', 'Bar', 'Baz'])->int();
```

Neos.Utility.Arrays/Classes/Arrays.php
```diff
@@ -218,6 +219,34 @@ public static function getValueByPath(array &$array, $path)
...
 
+    /**
+     * Returns a type safe accessor for a value in a nested array by following the specifed path.
+     *
+     * @param array $array The array to traverse as a reference
+     * @param array|string $path The path to follow. Either a simple array of keys or a string in the format 'foo.bar.baz'
+     * @param string|null $pathinfoSoFar
+     * @return ValueAccessor
+     */
+    public static function getAccessorByPath(array $array, array|string $path, string $pathinfoSoFar = null): ValueAccessor
```
bwaidelich: I struggle a bit with the last argument that has no doc comments and should probably only be used internally for the recursion AFAIS.
What about separating the internal part from the API (possibly via inline closure

Neos.Utility.Arrays/Tests/Unit/ArraysTest.php
```diff
@@ -108,6 +108,103 @@ public function getValueByPathReturnsNullIfThePathHasMoreSegmentsThanTheGivenArr
...
+     */
+    public function getAccessorByPathTypeErrorContainsPathForNonExistingPathed()
+    {
+        $array = ['Foo' => ['Bar' => ['Baz' => 'the value']]];
+        try {
+            self::assertNull(Arrays::getAccessorByPath($array, ['Foo', 'Bar', 'Baz', 'Bux'])->intOrNull());
+        } catch (\Error $error) {
+            self::assertInstanceOf(\TypeError::class, $error);
+            self::assertStringEndsWith('in path Foo.Bar.Baz.Bux', $error->getMessage());
+        }
```
bwaidelich: No exception is thrown in this case (another reason not to use `try/catch` in those tests)
```suggestion
        self::assertNull(Arrays::getAccessorByPath($array, ['Foo', 'Bar', 'Baz', 'Bux'])->intOrNull());
```

Neos.Utility.Arrays/Tests/Unit/ArraysTest.php
```diff
@@ -108,6 +108,103 @@ public function getValueByPathReturnsNullIfThePathHasMoreSegmentsThanTheGivenArr
...
+     */
+    public function getAccessorByPathTypeErrorContainsPathForNonMatchingTypesOnRoot()
+    {
+        $array = ['Foo' => 'string'];
+        try {
+            self::assertNull(Arrays::getAccessorByPath($array, ['Foo'])->int());
+        } catch (\Error $error) {
+            self::assertInstanceOf(\TypeError::class, $error);
+            self::assertStringEndsWith('in path Foo', $error->getMessage());
+        }
```
bwaidelich: same here
```suggestion
        $this->expectException(\TypeError::class);
        $this->expectExceptionMessage('in path Foo');
        Arrays::getAccessorByPath($array, ['Foo'])->int();
```

Neos.Utility.Arrays/Classes/Arrays.php
```diff
@@ -218,6 +219,34 @@ public static function getValueByPath(array &$array, $path)
...
+                return is_array($array[$key])
+                    ? self::getAccessorByPath($array[$key], $path, $pathinfo)
+                    : new ValueAccessor(null, $pathinfo);
+            } else {
+                return new ValueAccessor($array[$key], $pathinfo);
+            }
+        } else {
+            return new ValueAccessor(null, $pathinfo);
+        }
+    }
```
bwaidelich: There are a few possibilities for early returns that reduce cognitive load:

```suggestion
     * @param string|null $pathinfoSoFar
     * @return ValueAccessor
     */
    public static function getAccessorByPath(array $array, array|string $path, string $pathinfoSoFar = null): ValueAccessor
    {
        $pathSegments = is_string($path) ? explode('.', $path) : $path;
        $pathSegment = array_shift($pathSegments);
        $pathinfo = is_string($pathinfoSoFar) ? $pathinfoSoFar . '.' . $pathSegment : $pathSegment;
        if (!isset($array[$pathSegment])) {
            return new ValueAccessor(null, $pathinfo);
        }
        if ($pathSegments === []) {
            return new ValueAccessor($array[$pathSegment], $pathinfo);
        }
        return is_array($array[$pathSegment])
            ? self::getAccessorByPath($array[$pathSegment], $pathSegments, $pathinfo)
            : new ValueAccessor(null, $pathinfo);
    }
```

But I also struggle a bit with the last argument that has no doc comments and should probably only be used internally for the recursion AFAIS.
What about turning this into a non-recursive function like:

```suggestion
     * @return ValueAccessor
     */
    public static function getAccessorByPath(array $array, array|string $path, string $pathinfoSoFar = null): ValueAccessor
    {
        $pathSegments = is_string($path) ? explode('.', $path) : $path;
        $value = $array;
        $processedSegments = [];
        do {
            $pathSegment = array_shift($pathSegments);
            if (!array_key_exists($pathSegment, $value)) {
                $value = null;
                break;
            }
            $processedSegments[] = $pathSegment;
            $value = $value[$pathSegment];
            if (!is_array($value)) {
                if ($pathSegments !== []) {
                    $value = null;
                }
                break;
            }
        } while ($pathSegments !== []);
        return new ValueAccessor($value, implode('.', $processedSegments));
    }
```


Alternatively, we could consider reworking the existing `getValueByPath()` such that we could reuse it here:

```php
public static function getValueByPath(array $array, array|string $path, string &$processedPath = null): mixed
{
    $pathSegments = is_string($path) ? explode('.', $path) : $path;
    $value = $array;
    $processedSegments = [];
    do {
        $pathSegment = array_shift($pathSegments);
        if (!array_key_exists($pathSegment, $value)) {
            $value = null;
            break;
        }
        $processedSegments[] = $pathSegment;
        $value = $value[$pathSegment];
        if (!is_array($value)) {
            if ($pathSegments !== []) {
                $value = null;
            }
            break;
        }
    } while ($pathSegments !== []);
    $processedPath = implode('.', $processedSegments);
    return $value;
}
```

With that in place we could simplify the new method to just
```php
public static function getAccessorByPath(array $array, array|string $path): ValueAccessor
{
    $value = self::getValueByPath($array, $path, $processedPath);
    return new ValueAccessor($value, $processedPath);
}
```

(applying the last suggestion leads to one failing test because `getValueByPath()` used to throw an `InvalidArgumentException` instead of a `TypeError` – but that behavior should be unified anyways, IMO)

bwaidelich (commented):
I left a couple of comments.

Apart from those I wonder whether the "Arrays" package is really the right place for the `ValueAccessor` - OTOH I wouldn't know where to put it instead.

In addition to the suggested code unifications we should unify errors for `getValueByPath()` and `getAccessorByPath()` IMO (see inline comments)

Neos.Utility.Arrays/Classes/Arrays.php
mficzel: Adjusted to use getValueByPath internally which then used your early return suggestion.

Neos.Utility.Arrays/Tests/Unit/ArraysTest.php
mficzel: adjusted as requested

Neos.Utility.Arrays/Tests/Unit/ArraysTest.php
mficzel: adjusted as requested

Neos.Utility.Arrays/Classes/Arrays.php
mficzel: Removed that by internally using getValueByPath.

Neos.Utility.Arrays/Tests/Unit/ArraysTest.php
mficzel: adjusted as requested

mficzel:
@bwaidelich i also struggled deciding where to put this but neos/utility-arrays is correct as this package is concerned with unstructured array heaps in general. `neos/utility-objecthandling` would be okish aswell but seems not as fitting to me.

Another good solution would be to extract `neos\utility-typehandling` from `neos/utility-objecthandling` but that would introduce another utility package.

mficzel:
PR is ready and psalm complains about unrelated stuff

Neos.Utility.Arrays/Classes/ValueAccessor.php
```diff
@@ -0,0 +1,158 @@
...
+ */
+class ValueAccessor
+{
+    public function __construct(
+        public readonly mixed   $value,
+        public readonly ?string $pathinfo = null
+    ) {
+    }
+
+    private function createTypeError($message): \TypeError
```
grebaldi: I'm not sure whether `\TypeError` is appropriate for this. Besides the semantics, `\TypeError` bypasses Flow's error handling, leading to a rather unpleasant error message like this:

image

As opposed to:
image

(using the SPL exception [`\UnexpectedValueException`](https://www.php.net/manual/en/class.unexpectedvalueexception.php) which strikes me as the closest semantic equivalent)

Neos.Utility.Arrays/Classes/ValueAccessor.php
```diff
@@ -0,0 +1,158 @@
...
+        }
+        throw $this->createTypeError("is not an array");
+    }
+
+    /**
+     * @template T
+     * @param class-string<T> $type
+     * @return object&T
+     */
+    public function object(string $type): object
```
grebaldi:
```suggestion
    /**
     * @template T
     * @param class-string<T> $className
     * @return object&T
     */
    public function instanceOf(string $className): object
```

Sorry that I keep heckling stuff because of naming preferences :sweat_smile:, but as I went though the methods, this was the only one that didn't tell me immediately what it wanted to check. 

I believe that `instanceOf` would meet common expectations better.

Neos.Utility.Arrays/Classes/ValueAccessor.php
```diff
@@ -0,0 +1,158 @@
...
+        }
+        throw $this->createTypeError("is not an array or null");
+    }
+
+    /**
+     * @template T
+     * @param class-string<T> $type
+     * @return (object&T)|null
+     */
+    public function objectOrNull(string $type): ?object
```
grebaldi:
```suggestion
    /**
     * @template T
     * @param class-string<T> $className
     * @return (object&T)|null
     */
    public function instanceOfOrNull(string $className): ?object
```

see above

grebaldi (commented):
Hi @mficzel,

I tested your change locally and it works as expected :+1: 

I added two minor nitpicks in the comments, but I consider neither of these blockers. So, if you wish to keep the current implementation as it is, just say so and I'll approve :)

Neos.Utility.Arrays/Classes/ValueAccessor.php
mhsdesign: Yes I too was wondering if it would be more helpful to have a class thrown inheriting from exception. 

Regular errorhandling would otherwise fail and it would be seen as programming error when indeed is was a validation exception

mficzel: @grebaldi Not opinionated there at all. Simply used TypeError as this us what php would throw if one would simply use the return types. Did not much thought into it really.

Would you consider `\UnexpectedValueException` a good solution or should this be a Flow Exception?

mhsdesign: the neos utility package has no dependency on flow as far as i know.

Neos.Utility.Arrays/Classes/ValueAccessor.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * @template T
+     * @param class-string<T> $className
+     * @return object&T
+     */
+    public function instanceOf(string $className): object
+    {
+        if (is_a($this->value, $className, false)) {
```
mhsdesign: there was a super big problem with `is_a` i remember me and @nezaniel talking about it ... at least in our cases `class_implements` was better...

but how about we do a simple `$this-value instanceof $className` then id feel better :D

Neos.Utility.Arrays/Classes/ValueAccessor.php
mficzel: So then let us use \UnexpectedValueException

Neos.Utility.Arrays/Classes/ValueAccessor.php
mficzel: Adjusted as requested so @mhsdesign feels better ;-)

mhsdesign: :D thänks ^^

grebaldi (approved):
Thanks for the adjustments, @mficzel! LGTM :+1:

bwaidelich (approved):

Neos.Utility.Arrays/Classes/ValueAccessor.php
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Type safe accessing of values from nested arrays
+ */
+class ValueAccessor
+{
+    public function __construct(
+        public readonly mixed   $value,
+        public readonly ?string $pathinfo = null
+    ) {
+    }
```
mhsdesign, line 24: i would like to propose a follow up, where we make this constructor internal and expose those two static factories:

```
// @internal
public static function fromValueWithPathInfo

// @api
public static function fromValue
```

Issue #3120, opened 2023-07-18
BUG: RouteTags::createFromArray uses code deprecated in PHP8.2


nezaniel:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

RouteTags::createFromArray is implemented as
```php
public static function createFromArray(array $tags): self
 {
        array_walk($tags, 'static::validateTag');
        return new static($tags);
 }
```

This throws the following deprecation error using PHP 8.2: _Use of "static" in callables is deprecated_

### Expected Behavior

no deprecation errors :)

### Steps To Reproduce

call RouteTags::createFromArray

### Environment

```markdown
- Flow: 9.0
- PHP: 8.2
```

mhsdesign:
Link: https://github.com/neos/flow-development-collection/blob/4358297c60a0b6266900e447757a3de2388b2fb8/Neos.Flow/Classes/Mvc/Routing/Dto/RouteTags.php#L74C28-L74C34

`array_walk($tags, static::validateTag(...));` should do, but id actually prefer a simple for each 😅

bwaidelich:
FYI: I just opened up the WEB IDE and did the fix, it was almost faster than writing a comment :)
=> https://github.com/neos/flow-development-collection/pull/3156

mhsdesign:
Hahaha thännks ^^

PR #3156, opened 2023-09-07
BUGFIX: Prevent deprecation warning in `RouteTags::createFromArray()`

bwaidelich:
Fixes: #3120


mhsdesign (approved):
But only almost faster ❤️

PR #3117, opened 2023-07-18
TASK: Avoid potential deprecation warning in StringHelper

kdambekalns:
`str_replace()` expects strings, but Eel with it's loose typing might pass in different types.

**Checklist**

- [x] Code follows the PSR-2 coding style
- [ ] ~Tests have been created, run and adjusted as needed~
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [x] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [x] Reviewer - The first section explains the change briefly for change-logs
- [ ] ~Reviewer - Breaking Changes are marked wit `!!!` and have upgrade-instructions~

PR #3167, opened 2023-09-19
BUGFIX: Revert "TASK: Avoid potential deprecation warning in StringHelper"

### LINKED ISSUE #3166: BUG: Array to string conversion after Update

### Current Behavior

Currently, in `Neos\Eel\Helper\StringHelper`, the function `replace($string, $search, $replace)` casts each attribute to a string.

The problem with this is as follows:
Before the update, you could also specify an array as `$search` parameter, since 
`array|string $search, array|string $replace, array|string $subject` can be specified at `str_replace`. https://www.php.net/manual/en/function.str-replace.php

So now if you specified an array as `$search` attribute, an error will be thrown.

### Expected Behavior

It should not be thrown an exception after the update.
You should still be allowed to pass an array as an attribute, as before.

### Steps To Reproduce

**Install Neos 7.3.14**

Implement a function like this:
```js
// ReplaceSpecialChars.fusion
root {
    @process.replaceShy = ${String.replace(value, Array.keys(Configuration.setting('**.Neos.Base.specialCharReplacements')), Configuration.setting('**.Neos.Base.specialCharReplacements'))}
    @process.replaceShy.@if.notInBackend = ${!documentNode.context.inBackend}
}
```

The Config: 
```yaml
**:
  Neos:
    Base:
      testSystem: true
      googleTagManagerDataLayerEnabled: false
      specialCharReplacements:
        '|shy|': '&shy;'
        '|zwj|': '&zwj;'
        '|thinsp|': '&thinsp;'
        '|ensp|': '&ensp;'
        '|emsp|': '&emsp;'
        '|nbsp|': '&nbsp;'
        '|nbdash|': '&#8209;'
```

**Update to Neos 7.3.15**

### Environment

```markdown
- Flow: 7.3.15
- PHP: 8.1
```

### Anything else?

Log: [202309190959221e3bbe.txt](https://github.com/neos/flow-development-collection/files/12658008/202309190959221e3bbe.txt)

Slack: https://neos-project.slack.com/archives/C050C8FEK/p1695110373297909

image (2)

## Possible fix:
**You can find a possible fix for it at #3167**

LvckyAPI:
## Informations
**Based on #3166** 
Resolves #3166 

```php
str_replace()
```
`str_replace()` supports arrays as attributes.
We should allow users to use the function correctly.
https://www.php.net/manual/de/function.str-replace.php

This reverts commit 1390d88103a6be6ddfa827eb834b0f666f5ea67b. in 028e2e33570a20a835eff37021aa55b1b331a406
And updates the docs comment. 5ad36d9d925509697f33700dddf9c09d385b4aa7

**Checklist**

- [x] ~~Tests have been created, run and adjusted as needed~~

### DIFF
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -326,7 +326,7 @@ class StringHelper implements ProtectedContextAwareInterface
      */
     public function replace($string, $search, $replace)
     {
-        return str_replace((string)$search, (string)$replace, (string)$string);
+        return str_replace($search, $replace, (string)$string);
     }
 
     /**
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -319,14 +319,24 @@ class StringHelper implements ProtectedContextAwareInterface
      *
      * Note: this method does not perform regular expression matching, @see pregReplace().
      *
-     * @param string|array $string The input string|array
+     * @param string|array $subject The input string|array
      * @param string|array $search A search string|array
-     * @param string $replace A replacement string
-     * @return string The string with all occurrences replaced
+     * @param string|array $replace A replacement string|array
+     * @return string|array The subject with all occurrences replaced
      */
-    public function replace($string, $search, $replace)
+    public function replace($subject, $search, $replace)
     {
-        return str_replace($search, $replace, (string)$string);
+        if (!is_array($subject) && !is_string($subject)) {
+            $subject = (string)$subject;
+        }
+        if (!is_array($search) && !is_string($search)) {
+            $search = (string)$search;
+        }
+        if (!is_array($replace) && !is_string($replace)) {
+            $replace = (string)$replace;
+        }
+
+        return str_replace($search, $replace, $subject);
     }
 
     /**
--- a/Neos.Eel/Classes/Helper/StringHelper.php
+++ b/Neos.Eel/Classes/Helper/StringHelper.php
@@ -319,8 +319,8 @@ class StringHelper implements ProtectedContextAwareInterface
      *
      * Note: this method does not perform regular expression matching, @see pregReplace().
      *
-     * @param string $string The input string
-     * @param string $search A search string
+     * @param string|array $string The input string|array
+     * @param string|array $search A search string|array
      * @param string $replace A replacement string
      * @return string The string with all occurrences replaced
      */


mhsdesign:
Thanks for reporting the issue and the fix ❤️ 

- what do you say @kdambekalns why dont we allow array? I must say i dont understand where any possible deprecation warning the method might be coming from, which you tried to fix with https://github.com/neos/flow-development-collection/pull/3117 ?
I guess you want to cast `int` to a string, but we need to preserve arrays as is?

kdambekalns (changes requested):
_Oh my…_ PHP, you gotta love it. 😎

An array is allowed for search, replace – and subject, so **let's allow an array for the subject**, too. From https://www.php.net/str_replace:

>  If search and replace are arrays, then str_replace() takes a value from each array and uses them to search and replace on subject. If replace has fewer values than search, then an empty string is used for the rest of replacement values. If search is an array and replace is a string, then this replacement string is used for every value of search. The converse would not make sense, though.
>
> If search or replace are arrays, their elements are processed first to last. 
>
> If subject is an array, then the search and replace is performed with every entry of subject, and the return value is an array as well. 

But **if the passed valus are not array or string, we must cast to string** (most common case in Fusion/Eel is probably `null` coming in.)

kdambekalns:
Ah, and of course the return value can be an array, too. 🤪

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -319,14 +319,14 @@ public function pregSplit($string, $pattern, $limit = -1)
...
      * Note: this method does not perform regular expression matching, @see pregReplace().
      *
-     * @param string $string The input string
-     * @param string $search A search string
+     * @param string|array $string The input string|array
+     * @param string|array $search A search string|array
      * @param string $replace A replacement string
      * @return string The string with all occurrences replaced
      */
     public function replace($string, $search, $replace)
```
kdambekalns:
```suggestion
     * @param string|array $subject The input string|array
     * @param string|array $search A search string|array
     * @param string|array $replace A replacement string|array
     * @return string|array The subject with all occurrences replaced
     */
    public function replace($subject, $search, $replace)
    {
        if (!is_array($subject) && !is_string($subject)) {
            $subject = (string)$subject;
        }
        if (!is_array($search) && !is_string($search)) {
            $search = (string)$search;
        }
        if (!is_array($replace) && !is_string($replace)) {
            $replace = (string)$replace;
        }
```

Neos.Eel/Classes/Helper/StringHelper.php
```diff
@@ -319,14 +319,14 @@ public function pregSplit($string, $pattern, $limit = -1)
...
     {
-        return str_replace((string)$search, (string)$replace, (string)$string);
+        return str_replace($search, $replace, (string)$string);
```
kdambekalns:
```suggestion

        return str_replace($search, $replace, $subject);
```

LvckyAPI:
I will apply the requested changes from your review very soon.

Thank you :)

LvckyAPI:
I've implemented your suggestions @kdambekalns 
Thank you 😄

mhsdesign (approved):
Finde by 👀 

It should be actually quite easy to add a test case for this (there are unit tests already) - id really like that for the array case! ^^

mhsdesign (changes requested):
I like https://github.com/neos/flow-development-collection/pull/3180 better because it has tests ^^

kdambekalns:
Thanks for bringing this up and proposing a PR – even though we merged a different one, don't give up! 😉

PR #3172, opened 2023-09-22
BUGFIX: Sanitize uploaded svg files from suspicious content

mficzel:
Adding an internal methods `isSanitizingRequired` and `sanitizeImportedFileContent` to the resourceManager. The import is adjusted to first determine the mediaType of an imported resource to decide wether sanitizing is needed which for now happens only for SVG files. If no sanitizing is needed the code will perform as before by passing streams or filenames around.

If suspicious content was removed from a warning is logged that mentions the remove data and line. The sanitizing is done using "enshrined/svg-sanitize" that is used by other cms aswell.

The initial implementation will only sanitize SVG files as those can contain malicious scripts. In future this should be expanded to a feature that allows registering of custom sanitizing functions.

The sanitizing logic itself ist basically the same as what is done by typo3 here: https://github.com/TYPO3/typo3/blob/357b07064cf2c7f1735cfb8f73ac4a7248ab040e/typo3/sysext/core/Classes/Resource/Security/SvgSanitizer.php

This addresses the issue described here: https://nvd.nist.gov/vuln/detail/CVE-2023-37611

**Review Instructions**

The change adds quite a bit of complexity to the importResource method to avoid loading the file content into ram whenever possible. As this method accepts filenames and resources this leads to quite some nested checking. I consider this kindoff necessary as one does not want to read a full video file into php ram to check wether it may be an svg. 

Better suggestions are welcome.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/ResourceManager.php
+++ b/Neos.Flow/Classes/ResourceManagement/ResourceManager.php
@@ -11,10 +11,12 @@ namespace Neos\Flow\ResourceManagement;
  * source code.
  */
 
+use enshrined\svgSanitize\Sanitizer;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
+use Neos\Utility\MediaTypes;
 use Neos\Utility\ObjectAccess;
 use Neos\Flow\ResourceManagement\Storage\StorageInterface;
 use Neos\Flow\ResourceManagement\Storage\WritableStorageInterface;
@@ -160,13 +162,29 @@ class ResourceManager
         $collection = $this->collections[$collectionName];
 
         try {
-            $resource = $collection->importResource($source);
-            if ($forcedPersistenceObjectIdentifier !== null) {
-                ObjectAccess::setProperty($resource, 'Persistence_Object_Identifier', $forcedPersistenceObjectIdentifier, true);
-            }
-            if (!is_resource($source)) {
+            if (is_resource($source)) {
+                $mediaType = MediaTypes::getMediaTypeFromResource($source);
+                if ($this->isSanitizingRequired($mediaType)) {
+                    $content = stream_get_contents($source);
+                    $resource = $this->importResourceFromContent($content, '', $collectionName, $forcedPersistenceObjectIdentifier);
+                } else {
+                    $resource = $collection->importResource($source);
+                }
+            } else {
+                $resource = fopen($source, 'rb');
+                $mediaType = MediaTypes::getMediaTypeFromResource($resource);
+                fclose($resource);
+                if ($this->isSanitizingRequired($mediaType)) {
+                    $content = file_get_contents($source);
+                    $resource = $this->importResourceFromContent($content, '', $collectionName, $forcedPersistenceObjectIdentifier);
+                } else {
+                    $resource = $collection->importResource($source);
+                }
                 $pathInfo = UnicodeFunctions::pathinfo($source);
                 $resource->setFilename($pathInfo['basename']);
+            }
+            if ($forcedPersistenceObjectIdentifier !== null) {
+                ObjectAccess::setProperty($resource, 'Persistence_Object_Identifier', $forcedPersistenceObjectIdentifier, true);
             }
         } catch (Exception $exception) {
             throw new Exception(sprintf('Importing a file into the resource collection "%s" failed: %s', $collectionName, $exception->getMessage()), 1375197120, $exception);
@@ -204,6 +222,11 @@ class ResourceManager
 
         if (!isset($this->collections[$collectionName])) {
             throw new Exception(sprintf('Tried to import a file into the resource collection "%s" but no such collection exists. Please check your settings and the code which triggered the import.', $collectionName), 1380878131);
+        }
+
+        $mediaType = MediaTypes::getMediaTypeFromFileContent($content);
+        if ($this->isSanitizingRequired($mediaType)) {
+            $content = $this->sanitizeImportedFileContent($mediaType, $content, $filename);
         }
 
         /* @var CollectionInterface $collection */
@@ -606,6 +629,44 @@ class ResourceManager
 
             $this->collections[$collectionName] = new Collection($collectionName, $this->storages[$collectionDefinition['storage']], $this->targets[$collectionDefinition['target']], $pathPatterns, $filenames);
         }
+    }
+
+    /**
+     * Decide weather the given media-type has to be sanitized
+     * for now this only checks svg file to solve the issue here https://nvd.nist.gov/vuln/detail/CVE-2023-37611
+     *
+     * @todo create a feature from this and allow to register code for sanitizing file content before importing
+     */
+    protected function isSanitizingRequired(string $mediaType): bool
+    {
+        return $mediaType === 'image/svg+xml';
+    }
+
+    /**
+     * Sanitize file content and remove content that is suspicious
+     * for now this only checks svg file to solve the issue here https://nvd.nist.gov/vuln/detail/CVE-2023-37611
+     *
+     * @todo create a feature from this and allow to register code for sanitizing file content before importing
+     */
+    protected function sanitizeImportedFileContent(string $mediaType, string $content, $filename = ''): string
+    {
+        if ($mediaType === 'image/svg+xml') {
+            // @todo: Simplify again when https://github.com/darylldoyle/svg-sanitizer/pull/90 is merged and released.
+            $previousXmlErrorHandling = libxml_use_internal_errors(true);
+            $sanitizer = new Sanitizer();
+            $sanitizedContent = $sanitizer->sanitize($content);
+            libxml_clear_errors();
+            libxml_use_internal_errors($previousXmlErrorHandling);
+            $issues = $sanitizer->getXmlIssues();
+            if ($issues && count($issues) > 0) {
+                if ($sanitizedContent === false) {
+                    throw new Exception('Sanitizing of suspicious file "' . $filename . '" failed during import.', 1695395560);
+                }
+                $content = $sanitizedContent;
+                $this->logger->warning(sprintf('Imported file "%s" contained suspicious content and was sanitized.', $filename), $issues);
+            }
+        }
+        return $content;
     }
 
     /**
--- a/Neos.Utility.MediaTypes/Classes/MediaTypes.php
+++ b/Neos.Utility.MediaTypes/Classes/MediaTypes.php
@@ -1831,6 +1831,21 @@ abstract class MediaTypes
     }
 
     /**
+     * Returns a Media Type based on the given resource
+     *
+     * @param resource $resource The resource to determine the media type from
+     * @return string The IANA Internet Media Type
+     */
+    public static function getMediaTypeFromResource($resource): string
+    {
+        if (!is_resource($resource)) {
+            throw new \TypeError('Argument "resource" has to be a resource');
+        }
+        $mediaType = self::trimMediaType(mime_content_type($resource));
+        return isset(self::$mediaTypeToFileExtension[$mediaType]) ? $mediaType : 'application/octet-stream';
+    }
+
+    /**
      * Returns the primary filename extension based on the given Media Type.
      *
      * @param string $mediaType The IANA Internet Media Type, for example "text/html"
--- a/Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
+++ b/Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
@@ -10,6 +10,7 @@ namespace Neos\Utility\MediaTypes\Tests\Unit;
  * information, please view the LICENSE file which was distributed with this
  * source code.
  */
+
 use Neos\Utility\MediaTypes;
 
 /**
--- a/Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
+++ b/Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
@@ -10,7 +10,6 @@ namespace Neos\Utility\MediaTypes\Tests\Unit;
  * information, please view the LICENSE file which was distributed with this
  * source code.
  */
-
 use Neos\Utility\MediaTypes;
 
 /**
@@ -66,6 +65,22 @@ class MediaTypesTest extends \PHPUnit\Framework\TestCase
         $filePath = __DIR__ . '/Fixtures/' . $filename;
         $fileContent = is_file($filePath) ? file_get_contents($filePath) : '';
         self::assertSame($expectedMediaType, MediaTypes::getMediaTypeFromFileContent($fileContent));
+    }
+
+    /**
+     * @test
+     * @dataProvider filesAndMediaTypes
+     */
+    public function getMediaTypeFromResource(string $filename, string $expectedMediaType)
+    {
+        $filePath = __DIR__ . '/Fixtures/' . $filename;
+        $resource = is_file($filePath) ? fopen($filePath, 'rb') : fopen('data://text/plain,', 'rb');
+        if ($resource !== false) {
+            self::assertSame($expectedMediaType, MediaTypes::getMediaTypeFromResource($resource));
+            fclose($resource);
+        } else {
+            $this->fail('fixture ' . $filePath . ' could not be read');
+        }
     }
 
     /**
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -51,7 +51,8 @@
 
         "composer/composer": "^2.2.8",
 
-        "egulias/email-validator": "^2.1.17 || ^3.0"
+        "egulias/email-validator": "^2.1.17 || ^3.0",
+        "enshrined/svg-sanitize": "^0.16.0"
     },
     "require-dev": {
         "vimeo/psalm": "~4.30.0",
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "neos/composer-plugin": "^2.0",
         "composer/composer": "^2.2.8",
         "egulias/email-validator": "^2.1.17 || ^3.0",
+        "enshrined/svg-sanitize": "^0.16.0",
         "typo3fluid/fluid": "~2.7.0",
         "guzzlehttp/psr7": "^1.8.4",
         "ext-mbstring": "*"


mficzel:
I would rather see this in the resource manager but since the codePath uses $resourceManager->importResource i am not sure it would be wise to be placed there as this would check every resource even those created internally like image crops.

I assume it somehow more correct to validate the uploaded files then the importedResources.

Also i would like to show a nicer error message. Did not manage that yet as Errors or exceptions do not end up in the Media module.

mficzel:
It should be discussed wether this is actually expected behavior as there are possible valid use cases for scripts inside of svgs as animations like https://www.svgator.com/ i am aware of a former customer that actually used this.

Neos.Flow/Classes/ResourceManagement/ResourceTypeConverter.php
```diff
@@ -306,8 +308,35 @@ protected function handleUploadedFile(UploadedFileInterface $source, PropertyMap
...
+            $issues = $sanitizer->getXmlIssues();
+            if ($issues && count($issues) > 0) {
+                $messages = array_map(function (array $issue) {
+                    return $issue['message'] . ' in line ' . $issue['line'] . '.';
+                }, $issues);
+                $this->logger->warning('Blocked upload of suspicious file', ['issues' => $issues] + LogEnvironment::fromMethodName(__METHOD__));
+
+                return new FlowError('The upload of file "' . strip_tags($pathInfo['basename']) . '" was blocked. ' . implode(' ', $messages) . '  As it could pose a security risk it cannot be imported.', 1695361773);
+            }
+        }
```
dlubitz: Looks good for me, but I would suggest to spend a method like `checkForVulnerableCodeInSVG()` or similar for better reading.

dlubitz (dismissed):

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -608,6 +603,35 @@ protected function initializeCollections()
...
+            $sanitizedContent = $sanitizer->sanitize($content);
+            libxml_clear_errors();
+            libxml_use_internal_errors($previousXmlErrorHandling);
+            $issues = $sanitizer->getXmlIssues();
+            if ($issues && count($issues) > 0) {
+                if ($sanitizedContent === false) {
+                    throw new Exception('Sanitizing of suspicious file "' . $filename . '" failed during import.', 1695395560);
+                }
+                $content = $sanitizedContent;
+                $this->logger->debug(sprintf('Imported file "%s" contained suspicious content and was sanitized.', $filename), $issues);
```
mhsdesign: debug log might be too silent i think? How about at least an info?

```suggestion
                $this->logger->info(sprintf('Imported file "%s" contained suspicious content and was sanitized.', $filename), $issues);
```

kdambekalns (commented):
LGTM… 👍 by 👀

kdambekalns (dismissed):

kdambekalns:
We should also point out that people can use CSP to block such attacks. Back then I noted this down:

> As far as protecting against this goes, my first thought was "we can't do anything, those files are opened directly." Which might be not wrong, but we can probably do something. At least point people to the need to configure CSP headers to be sent with SVG? That's one thing mentioned in https://digi.ninja/blog/svg_xss.php that could help.

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
mficzel: You are right, i think that justifies a warning

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -151,30 +153,21 @@ protected function initialize()
...
             }
-            if (!is_resource($source)) {
-                $pathInfo = UnicodeFunctions::pathinfo($source);
-                $resource->setFilename($pathInfo['basename']);
+        } else {
+            $content = file_get_contents($source);
+            /** @var string $filename */
+            $filename = UnicodeFunctions::pathinfo($source, PATHINFO_FILENAME);
+            if ($content === false) {
+                throw new Exception(sprintf('Could not read the file from "%s" as resource.', $source), 1695390671);
```
dlubitz:
```suggestion
                throw new Exception(sprintf('Could not read the file from "%s" as resource.', $source), 1695390672);
```
Shouldn't this exception get it's own error code?

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -151,30 +153,21 @@ protected function initialize()
...
-
-        /* @var CollectionInterface $collection */
-        $collection = $this->collections[$collectionName];
-
-        try {
-            $resource = $collection->importResource($source);
-            if ($forcedPersistenceObjectIdentifier !== null) {
-                ObjectAccess::setProperty($resource, 'Persistence_Object_Identifier', $forcedPersistenceObjectIdentifier, true);
+        if (is_resource($source)) {
+            $content = stream_get_contents($source);
```
dlubitz: But this will load all resources fully into the memory, right? Wouldn't this have a huge impact for large uploads like videos?

dlubitz (commented):
I wonder, how does this impact cloud storages like S3?

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
mficzel: That somehow is the price for sanitizing ...

mficzel:
> I wonder, how does this impact cloud storages like S3?

I do not think that cdns are affected in a special way especially since this is in the resourceManager and not the storage. The side effect of loading the whole file is probably the major one to consider.

mficzel:
@dlubitz removed your approval and re-requested review as the approval was from before the refactoring and you still commented valid points. Was nut sure it was intended that the review was still in place and wanted to prevent a merge wrong assumptions.

dlubitz:
Thanks @mficzel 
TBH I struggle with the change, just because of the in-memory loading of all resources. There is a good reason, why we didn't did this before. But atm I have no better idea how to solve the issue without loading the resources into the memory. And maybe it's not an issue at all, but it feels like a big one.

Was thinking about something async, but this would go beyond the scope of this fix.

Furthermore we discussed a mechanism to disable the sanitize, if someone wants to. This would work as discussed with a aspect, but we wouldn't get the old behavior back with it. It would still load all resources into memory, even without the sanitize.

Sorry for "blocking" this, but I'd love to see more opinions on this. WDYT?

mficzel:
@fcool can you add your opinion since i know you have usecases where video files are uploaded to Neos.

mficzel:
@dlubitz that is valuable input no need to apologize. If you see any way around this i would be glad aswell. Maybe you can help limiting this behavior to svg files. 

We could try to detect the mimetype before reading the full content. That way we could limit the behavior to 'image/svg-xml' files. Not sure what would be a viable way to achieve that especially for streams where we have no filename at hand. But even with a filename i would rather look into the content and not fully trust the extension.

mficzel:
finfo::buffer (which is used internally https://github.com/neos/flow-development-collection/blob/9.0/Neos.Utility.MediaTypes/Classes/MediaTypes.php#L1826C2-L1831C3 ) does detect svg files also when only the first bytes are read.

```php
$fileInfo = new \finfo(FILEINFO_MIME);
$content = <<<EOF
  <?xml version="1.0" standalone="no"?>
    <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
    <svg version="1.1" baseProfile="full" xmlns="http://
EOF;
var_dump($fileInfo->buffer($content));
```

We could also add a method get mediaTypeFromStream to the MediaTypesUtility ... no clue how much the mime_content_type will actually read.
```php
    public static function getMediaTypeFromStream(resource $stream): string
    {       
        return mime_content_type($stream) ??: 'application/octet-stream';
    }
```

fcool:
Sorry, I'm a bit damaged currently. Can you elaborate, why this change might effect video uploads especially? The resource, which gets imported was loaded into RAM completely before, too.

@dlubitz has concerns about loading the complete collection, if I understood everything... which I seem to be overreading again and again

mficzel:
@fcool i expect video files to be the largest fils and be the most lilkely to cause issues with memory usage. The old implementation used streams or copy during importResource and tried to avoid reading the whole file into ram. I am not sure wether this was better memory wise and am not dealing with such problems usually as i am seldom involved in ops.

As i understood @dlubitz reading the whole collection was not an issue but only file in memory thing. By referring to "in-memory loading of all resources" he meant every resource that is uploaded is loaded into php memory.

~~I pushed an a follow up commit that tries to determine the mimetype before deciding to fully read the content. Much uglier and i would prefer the first solution it would be fine php-memory wise.~~

mficzel:
Removed the pre sanitizing check and converted it to a pr for easier comparison ´. This change here is back to the original implementation with the possible memory side effects.

Alternative pr with media type check before full reading and sanitizing https://github.com/neos/flow-development-collection/pull/3174 ... i think this is ugly and even if it is improved adds a lot of complexity.

If $someone with a usecase that uploads large files and imports them as resources would give the implementation here a test we can decide wether we gat away with a simple and clean solution or have to live with the added complexity.

dlubitz:
@fcool I was referring the loading of each resource (independent from type) into the memory. 

> The resource, which gets imported was loaded into RAM completely before, too.

ATM we just pass a stream resource and use `stream_copy_to_stream` to write the target file. So nothing "huge" is loaded into memory. That's why I'm a bit afraid of this change, as I can't predict the impact.

dlubitz:
I double checked that and can confirm that it adds the file size as memory usage on top of the current memory usage.

mficzel:
I adjusted the pr to avoid reading the file content into ram when possible. Especially the importResource method now looks a bit weird which is a consequence of an api like `string|resource $source A URI (can therefore also be a path and filename) or a PHP resource stream(!)`.

Please check this and also check wether i really did sort all those if / else craft properly.

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -160,18 +162,36 @@ public function importResource($source, $collectionName = ResourceManager::DEFAU
...
                 $resource->setFilename($pathInfo['basename']);
             }
         } catch (Exception $exception) {
             throw new Exception(sprintf('Importing a file into the resource collection "%s" failed: %s', $collectionName, $exception->getMessage()), 1375197120, $exception);
         }
 
+        if ($forcedPersistenceObjectIdentifier !== null) {
+            ObjectAccess::setProperty($resource, 'Persistence_Object_Identifier', $forcedPersistenceObjectIdentifier, true);
+        }
+
```
dlubitz: If you move this up into the try catch, the error handling stays the same as before.
(Up to line 186/187)

dlubitz (approved):
@mficzel Thank you very much. This looks pretty good by reading. I just added one little thing. 
I'll test it later.

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -160,18 +162,36 @@ public function importResource($source, $collectionName = ResourceManager::DEFAU
...
+                } else {
+                    $resource = $collection->importResource($source);
+                }
+            } else {
+                $resource = fopen($source, 'rb');
+                $mediaType = MediaTypes::getMediaTypeFromResource($resource);
+                fclose($resource);
+                if ($this->isSanitizingRequired($mediaType)) {
+                    $content = file_get_contents($source);
+                    /** @var string $filename */
```
mhsdesign:
```suggestion
```

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -160,18 +162,36 @@ public function importResource($source, $collectionName = ResourceManager::DEFAU
...
+                    $resource = $this->importResourceFromContent($content, '', $collectionName, $forcedPersistenceObjectIdentifier);
```
mhsdesign, line 180:
```suggestion
                    $resource = $this->importResourceFromContent($content, $source, $collectionName, $forcedPersistenceObjectIdentifier);
```

mhsdesign:
Thanks looks good and is still quite readable thanks.

Count me for now as half review ^^

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
mficzel: addressed

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
mficzel: done as suggested

mficzel:
Adjusted the mentioned points

kdambekalns:
> It should be discussed wether this is actually expected behavior as there are possible valid use cases for scripts inside of svgs as animations like https://www.svgator.com/ i am aware of a former customer that actually used this.

That could indeed have side effects. If this had a switch, people could disable sanitation when they apply some other means of blocking such attacks (e.g. a CSP…)

kdambekalns (approved):
Looks good by careful 👀, good to go. See my last comment on making this optional, though.

mficzel:
@kdambekalns in my tests the animations survived to my surprise. Consider this safe for now. Also i want to avoid injecting features in a bugfix release. We can still add an opt in flag once we are notified of negative effects.

Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
```diff
@@ -68,6 +68,22 @@ public function getMediaTypeFromFileContent(string $filename, string $expectedMe
...
+     */
+    public function getMediaTypeFromResource(string $filename, string $expectedMediaType)
+    {
+        $filePath = __DIR__ . '/Fixtures/' . $filename;
+        $resource = is_file($filePath) ? fopen($filePath, 'rb') : fopen('data://text/plain,', 'rb');
+        if ($resource !== false) {
+            self::assertSame($expectedMediaType, MediaTypes::getMediaTypeFromResource($resource));
+            fclose($resource);
+        } else {
+            $this->fail('fixture ' . $filePath . ' could not be read');
```
simonschaufi, line 83: Does the resource not be closed here as well?

Neos.Utility.MediaTypes/Classes/MediaTypes.php
```diff
@@ -11,6 +11,7 @@
  * source code.
  */
 
+
```
simonschaufi: i guess this can be deleted, or?

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -608,6 +631,44 @@ protected function initializeCollections()
...
+        return $mediaType === 'image/svg+xml';
+    }
+
+    /**
+     * Sanitize file content and remove content that is suspicious
+     * for now this only checks svg file to solve the issue here https://nvd.nist.gov/vuln/detail/CVE-2023-37611
+     *
+     * @todo create a feature from this and allow to register code for sanitizing file content before importing
+     */
+    protected function sanitizeImportedFileContent(string $mediaType, string $content, $filename = ''): string
```
simonschaufi, line 651: would it not make sense to extract this method into an own class the same way what TYPO3 does?

mficzel: Eventually yes. But not in a bugfix. That is why there is "@todo create a feature from this" in there

kitsunet:
Thanks for taking care, aaaaand I am sorry that I come in so late, I won't block this, but I find sanitizer the wrong approach. IMHO either we straight up block SVG uploads with scripts (which would still mean looking into them :/) or we disable SVGs altogther by default and document at the point where you enable them that you must take care of sending appropriate CSP to prevent inline scripts or include the SVGs in a safe way according to the article on how to make this safe. 
I feel sanitizing is just a kind of arms race against more cleverly disguised scripts.

Neos.Flow/Classes/ResourceManagement/ResourceManager.php
simonschaufi: ok

Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
mficzel: No, the else branch is used when fopen returned false because it could not open the file in the first place.

Neos.Utility.MediaTypes/Classes/MediaTypes.php
mficzel: correct is removed

mficzel:
@kitsunet i first tried to block the upload by throwing an exception. The same library can also tell you wether an svg smells fishy. Problem was that the media utility did not show an understandable error message when an upload failed.

In the end sanitizing the svg was almost the same code and did not require any user interaction or changes in other parts than resource management.

Blocking SVG alltogether seems no viable to me at all. This is a pretty common image format nowadays.

Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
```diff
@@ -10,7 +10,6 @@
  * information, please view the LICENSE file which was distributed with this
  * source code.
  */
-
```
simonschaufi: this new line is ok. I think other files are the same way with one empty line after the header comment

mficzel:
AFK for the next days. Feed free to adjust / merge this as you like. 

@kitsunet if you find a way to show a meaningful error message in the media module by throwing an Exception in https://github.com/neos/flow-development-collection/pull/3172/files#diff-189d35abcc40be6c39fd0d0e15c088658da7c96ee3985b4e2afb6061dd6f3aa9R662 i would appreciate that and gladly abandon the idea of silent sanitizing. Maybe I just did not find the correct exception type. Media module always complained about http errors or something else not suitable for editors

Neos.Utility.MediaTypes/Tests/Unit/MediaTypesTest.php
kdambekalns: Yes, we want that line.

kitsunet:
I don't think there is an easy way without messing with the whole module and JS, so not blocking, lets do this, better this than nothing.

kdambekalns:
I'll merge this. For the "reject instead of sanitize" I created an issue, but this is better than the current state, arguably.

kdambekalns:
What, wait? How can an unrelated Redis test suddenly fail, and only on PHP8?

kdambekalns:
> What, wait? How can an unrelated Redis test suddenly fail, and only on PHP8?

https://github.com/neos/flow-development-collection/pull/3194

mficzel:
Rebase or just merge this after the redis issue was solved?

kdambekalns:
> Rebase or just merge this after the redis issue was solved?

LGTM 😎

kdambekalns:
Dang, that would have been nice for the releases I just tagged… 🤪

mficzel:
Just to be sure ... this was the redis issue not this pr?

PR #3174, opened 2023-09-24
TASK: Check media type before reading whole file

mficzel:
Attempt to improve: https://github.com/neos/flow-development-collection/pull/3172

Try to read as little as possible of the resources during import before deciding wether the media type needs sanitizing and therefore the full content has to be read into a variable

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/ResourceManager.php
+++ b/Neos.Flow/Classes/ResourceManagement/ResourceManager.php
@@ -11,10 +11,12 @@ namespace Neos\Flow\ResourceManagement;
  * source code.
  */
 
+use enshrined\svgSanitize\Sanitizer;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Log\Utility\LogEnvironment;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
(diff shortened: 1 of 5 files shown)


mficzel:
A thing that i dislike in this approach is the complexity and the case that it relies on the method `mime_content_type` accepting a resource which is documented and works but not specificly explained in the documentation. The documentation only mentions filenames beeing passed but resources are accepted aswell. Also, If mime_content_type will read the full content we would not gain anything

dlubitz:
Tested `mime_content_type` locally with steams. It doesn't add something noticeable to the memory usage and seems to work quite well. It detected my test files even if I tried with "wrong" extensions.

mficzel:
Then we should will try to get this approach to look nice :-/

mhsdesign:
Hmm having the correct csp in place should also prevent this issue - as it only occurs when opening the svg directly: https://digi.ninja/blog/svg_xss.php

Thank you both so much for taking care and also making sure we dont implement anything not performant!

mficzel:
Not sure csp can cover 100% of the cases. Especially when Resourcs are served directly by the webserver.

Issue #3169, opened 2023-09-21
META: Cleanup behat stuff


mhsdesign:
- [x] revert hack https://github.com/neos/neos-ui/pull/3569/commits/103afcf25ed28a65eebe7c69fddbfc63f7f163d7
- [x] revert hack https://github.com/neos/redirecthandler-neosadapter/pull/60/files/892f8a4e6b618d935c709103cd6ebdf5253eb504#r1354586677
- [x] https://github.com/neos/flow-development-collection/pull/3175
- [x] https://github.com/neos/flow-development-collection/pull/3173
- [x] https://github.com/neos/neos-development-collection/pull/4543
- [x] https://github.com/neos/neos-development-collection/pull/4544 (cleans up also behat stuff partially)
- [x] https://github.com/neos/behat/pull/35
- [x] https://github.com/neos/neos-development-collection/pull/4669
- [x] https://github.com/neos/neos-development-collection/pull/4669

public usages that we need to consider:

`FlowContextTrait` in every package that has behat tests:

- `Flowpack.DecoupledContentStore`
- `Neos.RedirectHandler.NeosAdapter`

### Packages using "scarier" functionality:

`SecurityOperationsTrait` in https://github.com/Flowpack/Flowpack.DecoupledContentStore/blob/6adb8e04246f5fc9b8471b656db7ddf131474745/Tests/Behavior/Features/Bootstrap/FeatureContext.php#L49 -> requires `$isolated = false`

And a lot of better to be cleaned up stuff in the `Neos.RedirectHandler.NeosAdapter` version 9

### Utility package for actual e2e tests based up on neos e2e "api"

- https://github.com/sandstorm/Sandstorm.E2ETestTools

- https://github.com/cron-eu/neos-behat/blob/cf5628722862373783827bd513d2f5af4413f524/Tests/Behat/FeatureContextBase.php#L30

mhsdesign:
The last open task is to get rid of our `@isolated` stuff but thats for another time. https://github.com/neos/flow-development-collection/issues/3170

Our behat things are now cleaned up ^^

PR #3173, opened 2023-09-24
TASK: document and deprecate flows internal isolated behat tests

mhsdesign:
Related https://github.com/neos/flow-development-collection/issues/3170

The infrastructure is quite complex and not in relation to those two tests. That's why we declare it ready to be removed.

**Upgrade instructions**

**Review instructions**

**Checklist**

- [ ] Code follows the PSR-2 coding style
- [ ] Tests have been created, run and adjusted as needed
- [ ] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [ ] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [ ] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked wit `!!!` and have upgrade-instructions

PR #3175, opened 2023-09-24
Task: minor behat test cleanup

mhsdesign:
resolves partially https://github.com/neos/flow-development-collection/issues/3169

**Upgrade instructions**

**Review instructions**

Instead of using `neos/behat` which installs selenium and stuff in another composer installation directory, we use behat as intended in the same composer installation.

**Checklist**

### DIFF
--- a/Neos.Flow/Tests/Behavior/behat.yml.dist
+++ b/Neos.Flow/Tests/Behavior/behat.yml.dist
@@ -1,7 +1,3 @@
-# Behat distribution configuration
-#
-# Override with behat.yml for local configuration.
-#
 default:
   autoload:
     '': "%paths.base%/Features/Bootstrap"
@@ -11,17 +7,3 @@ default:
         - "%paths.base%/Features/Security"
       contexts:
         - FeatureContext:
-
-  extensions:
-    Behat\MinkExtension:
-      files_path: Features/Resources
-      show_cmd: 'open %s'
-      goutte: ~
-      selenium2: ~
-
-      # Project base URL
-      #
-      # Use BEHAT_PARAMS="extensions[Behat\MinkExtension\Extension][base_url]=http://neos.local/" for configuration during
-      # runtime.
-      #
-      # base_url: http://localhost/
--- a/Neos.Flow/Tests/Functional/Command/BehatTestHelper.php
+++ b/Neos.Flow/Tests/Functional/Command/BehatTestHelper.php
@@ -14,17 +14,11 @@ namespace Neos\Flow\Tests\Functional\Command;
 require_once(FLOW_PATH_PACKAGES . '/Framework/Neos.Flow/Tests/Behavior/Features/Bootstrap/IsolatedBehatStepsTrait.php');
 require_once(FLOW_PATH_PACKAGES . '/Framework/Neos.Flow/Tests/Behavior/Features/Bootstrap/SecurityOperationsTrait.php');
 
-use Neos\Flow\Tests\Behavior\Features\Bootstrap\IsolatedBehatStepsTrait;
-use Neos\Flow\Tests\Behavior\Features\Bootstrap\SecurityOperationsTrait;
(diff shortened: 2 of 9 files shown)


Neos.Flow/Tests/Behavior/behat.yml.dist
bwaidelich, line 1: using a `behat.yml.dist` is a common pattern to allow for local modifications and IIRC there is some automatic fallback in place that looks for `behat.yml` and uses `behat.yml.dist` if the former does not exist.

OTOH I wouldn't know what to change for a local installation since this does not contain any setup specific parameters..

Neos.Flow/Tests/Functional/Command/BehatTestHelper.php
```diff
@@ -35,13 +29,11 @@
...
 class BehatTestHelper
 {
-    use IsolatedBehatStepsTrait;
-
     use SecurityOperationsTrait;
 
-    /**
-     * @var Bootstrap
-     */
+    protected $isolated = false;
```
bwaidelich, line 34: Not new with this change, but I really wonder why we use so many `protected` fields. It makes the whole construct more fragile IMO

bwaidelich (commented):
Very nice changes, thanks for taking care. Just a couple of inline comments

Neos.Flow/Tests/Functional/Command/BehatTestHelper.php
mhsdesign: yes it does - but this is completely internal and will be removed (see https://github.com/neos/flow-development-collection/pull/3173)

Neos.Flow/Tests/Behavior/behat.yml.dist
mhsdesign: hmm 😂  so what do you propose? `behat.yml` feels more correct, as no one should override it dude

Neos.Flow/composer.json
```diff
@@ -60,7 +60,8 @@
     "require-dev": {
         "vimeo/psalm": "~4.30.0",
         "mikey179/vfsstream": "^1.6.10",
-        "phpunit/phpunit": "~9.1"
+        "phpunit/phpunit": "~9.1",
+        "behat/behat": "^3.5"
```
dlubitz: Why do we need `behat/behat` in Neos.Flow AND in the development-collection? From my POV we just need it in the  development-collection and just put the tests in Neos.Flow

mhsdesign: ill revert that ;)

Neos.Flow/Tests/Behavior/behat.yml.dist
mhsdesign: there is actually no documentation of the `yml.dist` in the official docs https://docs.behat.org/en/latest/user_guide/configuration.html#behat-yml so lets drop it.

Neos.Flow/composer.json
mhsdesign:
```suggestion
        "phpunit/phpunit": "~9.1"
```

mhsdesign: fixed

Neos.Flow/Tests/Behavior/behat.yml.dist
bwaidelich: OK, I might have confused it with `.env.dist`

mhsdesign:
Okay @kitsunet liked this idea in slack 

> yes, very much in favor of simplifying that and using behat natively

so id count this as a review. @bwaidelich then lets merge this?

dlubitz (approved):

dlubitz:
I just don't get what the `.composer.json` is for. 🤷‍♂️ 

Edit: Ok, figured out. 👍

kitsunet (approved):

kitsunet:
Would you do a merge conflict fix, then I guess we could add this.

PR #3181, opened 2023-09-26
WIP: FEATURE: move FlowContextTrait into neos/flow

mhsdesign:
**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/TestSuite/Behavior/Behat/FlowBootstrapTrait.php
+++ b/Neos.Flow/Classes/TestSuite/Behavior/Behat/FlowBootstrapTrait.php
@@ -1,0 +1,191 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Flow\TestSuite\Behaviour\Behat;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Doctrine\DBAL\DBALException;
+use Doctrine\ORM\EntityManagerInterface;
+use Neos\Flow\Cli\CommandRequestHandler;
+use Neos\Flow\Core\Booting\Scripts;
+use Neos\Flow\Core\Bootstrap;
+use Neos\Flow\Mvc\Routing\Router;
+use Neos\Flow\ObjectManagement\ObjectManagerInterface;
+use Neos\Flow\Persistence\Doctrine\Service;
+use Neos\Flow\Persistence\PersistenceManagerInterface;
+use Neos\Flow\Security\Policy\PolicyService;
+
+trait FlowBootstrapTrait
+{
+    /**
+     * @var Bootstrap
+     */
+    static protected $bootstrap;
+
+    /**
+     * @var Router
+     */
+    protected $router;
+
+    /**
+     * @var ObjectManagerInterface
+     */
+    protected $objectManager;
+
+    /**
+     * @var \Doctrine\DBAL\Schema\Schema
+     */
+    protected static $databaseSchema;
+
+    /**
+     * Create a flow bootstrap instance
+     */
+    protected function initializeFlow(): Bootstrap
+    {
+        require_once(__DIR__ . '/../../../Core/Bootstrap.php');
+        if (!defined('FLOW_PATH_ROOT')) {
+            define('FLOW_PATH_ROOT', realpath(__DIR__ . '/../../../../..') . '/');
+        }
+        // The new classloader needs warnings converted to exceptions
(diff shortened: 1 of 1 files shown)


bwaidelich:
As discussed: I would recommend to keep the `neos/behat` package, because it is a good starting point and a perfect place for Behat specific documentation and utilities.

But we should radically clean it up for 9.0 with just a simple way to bootstrap Flow via some `BehatFlowBootstrapper` or a `FlowBootstrapTrait` along the lines of:

```php
<?php
declare(strict_types=1);

use Doctrine\DBAL\Schema\Schema;
use Neos\Flow\Configuration\ConfigurationManager;
use Neos\Flow\Core\Booting\Exception\SubProcessException;
use Neos\Flow\Core\Booting\Scripts;
use Neos\Flow\Core\Bootstrap;
use Neos\Flow\Exception;
use Neos\Flow\ObjectManagement\ObjectManagerInterface;
use Neos\Flow\Persistence\PersistenceManagerInterface;
use Neos\Flow\Tests\FunctionalTestRequestHandler;

trait FlowContextTrait
{

    /**
     * @var Bootstrap
     */
    private static $bootstrap;

    /**
     * @var Schema
     */
    private static $databaseSchema;

    /**
     * Create a flow bootstrap instance
     * @internal
     */
    private static function initializeFlow(): Bootstrap
    {
        if (self::$bootstrap !== null) {
            // already initialized
            return self::$bootstrap;
        }
        require_once dirname(__DIR__, 5) . '/Packages/Framework/Neos.Flow/Classes/Core/Bootstrap.php';
        if (!defined('FLOW_PATH_ROOT')) {
            define('FLOW_PATH_ROOT', dirname(__DIR__, 5) . '/');
        }
        // The new classloader needs warnings converted to exceptions
        if (!defined('BEHAT_ERROR_REPORTING')) {
            define('BEHAT_ERROR_REPORTING', E_ALL);
        }
        self::$bootstrap = new Bootstrap('Testing/Behat');
        Scripts::initializeClassLoader(self::$bootstrap);
        Scripts::initializeSignalSlot(self::$bootstrap);
        Scripts::initializePackageManagement(self::$bootstrap);
        self::$bootstrap->setActiveRequestHandler(new FunctionalTestRequestHandler(self::$bootstrap));
        self::$bootstrap->buildRuntimeSequence()->invoke(self::$bootstrap);
        return self::$bootstrap;
    }

    /**
     * @AfterSuite
     */
    public static function shutdownFlow(): void
    {
        if (self::$bootstrap !== null) {
            self::$bootstrap->shutdown('Runtime');
        }
    }

    /**
     * Persist any changes
     */
    protected function persistAll(): void
    {
        $this->getObject(PersistenceManagerInterface::class)->persistAll();
        $this->getObject(PersistenceManagerInterface::class)->clearState();
    }

    /**
     * If you only want to get an object instance you can use getObject() instead
     * @see getObject
     *
     * @return ObjectManagerInterface
     */
    protected function getObjectManager(): ObjectManagerInterface
    {
        try {
            return self::initializeFlow()->getObjectManager();
        } catch (Exception $e) {
            throw new \RuntimeException('Could not initialize Flow Object Manager', 1549385035, $e);
        }
    }

    /**
     * @template T of object
     *
     * @return T
     */
    protected function getObject(string $className): object
    {
        return $this->getObjectManager()->get($className);
    }

    /**
     * Executes the specified $command in the context of the application to test
     */
    protected function executeCommand(string $command, array $commandArguments): void
    {
        /** @var ConfigurationManager $configurationManager */
        $configurationManager = $this->getObject(ConfigurationManager::class);
        /** @noinspection PhpUnhandledExceptionInspection */
        $flowSettings = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'Neos.Flow');
        try {
            Scripts::executeCommand($command, $flowSettings, false, $commandArguments);
        } catch (SubProcessException $e) {
            throw new RuntimeException(sprintf('An exception was thrown while executing command "%s"; %s', $command, $e->getMessage()), 1549290029, $e);
        }
    }
}
```

Neos.Flow/Classes/TestSuite/Behavior/Behat/FlowBootstrapTrait.php
```diff
@@ -0,0 +1,191 @@
...
+        return $bootstrap;
+    }
+
+    public function getObjectManager(): ObjectManagerInterface
+    {
+        return $this->objectManager;
+    }
+
+    /**
+     * @AfterSuite
```
kitsunet, line 84: If we switch this to attributes in the future we would create a dependency on behat, therefore I think we want a separate package even if it's just this trait?

mhsdesign: Yesssesses it’s in neos/be hat already. This idea is overhauled and can be closed. We decided against this.

Issue #3159, opened 2023-09-12
BUG: Exception handling differs depending on "trigger point"


kdambekalns:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

One example is the handling of login forms:

- when entering wrong credentials, a "nice" error message is show, no exception dump is written
- when tampering with the HMAC, the output is a bare error message, an exception is logged, an exception dump is written

Other "wrong" requests show the same behaviour, e.g. when changing the controller reference of a POST request.

### Expected Behavior

The error shown to the user should be consistent.

### Environment

```markdown
- Flow: 7.3 & up
```

### Anything else?

It is generally fine that Flow throws an error here. But the Neos backend login that is “tested” quite often, like any other reachable login page on the web, should in my opinion be a bit more resilient to such tampered requests. Showing the user the same or a similar error message like for a failed login would be OK. Optionally just logging this but not writing out exception dumps would be great. Optionally even ignoring that in the logs should be possible.

All the above applies to Production context. Having more verbose output in Development context is wanted/needed, of course.

Related issue from the past: #2681 (and fix in #2685)

kdambekalns:
- Testcase: Neos-Login
- Context: Production
- Neos 7.3
- PHP 8.1

## Scenarios

1. When the Neos login form is submitted with a changed value for the controller or action in `__referrer[@…`], a 404 is returned when changin the controller to something non-existent, when changing the action to e.g. `logout`, the form is shown again. No log message, no trace file. This is fine… ✅
2. When the `__trustedProperties` of a request is tampered with, the browser show a plain text message: `Invalid HMAC submitted`. A trace file is written, the log says:
   `24-02-23 09:23:46 2212       NOTICE    Neos.Flow            Property mapping configuration failed due to HMAC errors. Exception #1320830018 in line 127 of /application/Data/Temporary/Production/SubContextBeach/SubContextInstance/Cache/Code/Flow_Object_Classes/Neos_Flow_Mvc_Controller_MvcPropertyMappingConfigurationService.php: The given string was not appended with a valid HMAC. - See also: 2024022309234484b44b.txt`
3. If in scenario 3 the `__referrer[arguments]` (present e.g. after a failed login) is tampered with (and not the `trustedProperties`), a "400 Bad Request" error screen is the result. A trace file is written, the log says:
[...]

PR #3187, opened 2023-09-28
TASK: Clean up stored throwable dumps

### LINKED ISSUE #3158: FEATURE: Rotate / clean up exception dumps similar to log files

### Description

Flow rotates it's log files as configured, but does not do anything like that to the exception dumps it writes to disk. This is something that can be unexpected and unless monitored could lead to issues with disk space and general "information overload".

### Possible Solution

When rotating log files, also check exception dumps and remove the ones older than a certain threshold and/or if there are more than a certain number of those.

kdambekalns:
Whenever a new dump is written, check the existing dumps and remove those that are older than allowed or exceed the maximum count.

By default nothing is cleaned up.

Resolves #3158

**Review instructions**

Should remove old dump files as configured…

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
+++ b/Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
@@ -36,6 +36,20 @@ class FileStorage implements ThrowableStorageInterface
     protected $storagePath;
 
     /**
+     * The maximum age of throwable dump in seconds, 0 to disable cleaning based on age
+     *
+     * @var int
+     */
+    protected $maximumThrowableDumpAge = 0;
+
+    /**
+     * The maximum number of throwable dumps to store, 0 to disable cleaning based on count
+     *
+     * @var int
+     */
+    protected $maximumThrowableDumpCount = 0;
+
+    /**
      * Factory method to get an instance.
      *
      * @param array $options
@@ -44,18 +58,24 @@ class FileStorage implements ThrowableStorageInterface
     public static function createWithOptions(array $options): ThrowableStorageInterface
     {
         $storagePath = $options['storagePath'] ?? (FLOW_PATH_DATA . 'Logs/Exceptions');
-        return new static($storagePath);
+        $maximumThrowableDumpAge = $options['maximumThrowableDumpAge'] ?? 0;
+        $maximumThrowableDumpCount = $options['maximumThrowableDumpCount'] ?? 0;
+        return new static($storagePath, $maximumThrowableDumpAge, $maximumThrowableDumpCount);
     }
 
     /**
      * Create new instance.
      *
      * @param string $storagePath
+     * @param int $maximumThrowableDumpAge
+     * @param int $maximumThrowableDumpCount
      * @see createWithOptions
      */
-    public function __construct(string $storagePath)
+    public function __construct(string $storagePath, int $maximumThrowableDumpAge, int $maximumThrowableDumpCount)
     {
         $this->storagePath = $storagePath;
+        $this->maximumThrowableDumpAge = $maximumThrowableDumpAge;
+        $this->maximumThrowableDumpCount = $maximumThrowableDumpCount;
 
         $this->requestInformationRenderer = static function () {
             // The following lines duplicate Scripts::initializeExceptionStorage(), which is a fallback to handle
@@ -133,6 +153,8 @@ class FileStorage implements ThrowableStorageInterface
         if (!file_exists($this->storagePath) || !is_dir($this->storagePath) || !is_writable($this->storagePath)) {
             return sprintf('Could not write exception backtrace into %s because the directory could not be created or is not writable.', $this->storagePath);
         }
+
+        $this->cleanupThrowableDumps();
 
         // FIXME: getReferenceCode should probably become an interface.
         $referenceCode = (is_callable([
@@ -232,5 +254,37 @@ class FileStorage implements ThrowableStorageInterface
         }
 
         return $output;
+    }
+
+    /**
+     * Cleans up existing throwable dumps when they are older than the configured
+     * maximum age or the oldest ones exceeding the maximum number of dumps allowed.
+     */
+    protected function cleanupThrowableDumps(): void
+    {
+        if ($this->maximumThrowableDumpAge > 0) {
+            $cutoffTime = time() - $this->maximumThrowableDumpAge;
+
+            /** @var \SplFileInfo $directoryEntry */
+            $iterator = new \DirectoryIterator($this->storagePath);
+            foreach ($iterator as $directoryEntry) {
+                if ($directoryEntry->isFile() && $directoryEntry->getCTime() < $cutoffTime) {
+                    unlink($directoryEntry->getRealPath());
+                }
+            }
+        }
+
+        if ($this->maximumThrowableDumpCount > 0) {
+            // this returns alphabetically ordered, so oldest first, as we have a date/time in the name
+            $existingDumps = glob(Files::concatenatePaths([$this->storagePath, '*']));
+            $existingDumpsCount = count($existingDumps);
+            if ($existingDumpsCount > $this->maximumThrowableDumpCount) {
+                $dumpsToRemove = array_slice($existingDumps, 0, $existingDumpsCount - $this->maximumThrowableDumpCount);
+
+                foreach ($dumpsToRemove as $dumpToRemove) {
+                    unlink($dumpToRemove);
+                }
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


kdambekalns:
I admit:

- I did not run this code once, yet 🙈
- It might look like a feature, but I want this in 7.3, as it can avoid disks being filled

@mrimann – feel free to test, if you dare…

kdambekalns:
Fixed PHP 7.3 compatibility for Flow 7.3… 🙄🤣

mrimann:
I tried, but failed. I'm not 100% sure if I tested wrong, or if the code does not yet do what I'd have expected.

My test-setup was:
- fresh flow-base-distribution, 7.3.x branch
- in a DDEV environment with PHP 7.4.x
- composer install
- empty DB + applied all migrations
- copied over the 2 changed files from this pull request

Then I tried to provoke Flow into writing exceptions by calling `/foobar` via the browser -> shows an error, but does not write out an Exception / Dump file.

Next I "broke" the DB by removing an entry from the migrations table - re-applying the same migration now throws an error and writes a Dump file.

But no exceptions are removed, even if I configure the `maximumThrowableDumpCount: 2`.

(I verified that editing those settings works by temporarily setting the `storagePath` option to something else -> this immediately worked. I also tried to circumvent any problems with the configuration injection, by adding the line `$this->maximumThrowableDumpCount = 2;` right before the `if ()` that checks if that setting is >0 - but this also does not remove any of the exception files).

@kdambekalns sorry - but hope this helps a bit.

Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
```diff
@@ -233,4 +255,35 @@ protected function renderRequestInfo()
...
+                if ($directoryEntry->isFile() && $directoryEntry->getCTime() < $cutoffTime) {
+                    unlink($directoryEntry->getRealPath());
+                }
+            }
+        }
+
+        if ($this->maximumThrowableDumpCount > 0) {
+            // this returns alphabetically ordered, so newest first, as we have a date/time in the name
+            $existingDumps = glob($this->storagePath);
+            if (count($existingDumps) > $this->maximumThrowableDumpCount) {
```
ahaeslich: `count($existingDumps)` will always return 1 as `$this->storagePath` contains the folder (e.g. `/myFlowPath/Data/Logs/Exceptions`) instead of a glob pattern (e.g. `/myFlowPath/Data/Logs/Exceptions/*`).

ahaeslich (changes requested):
I tested it in a 8.3 installation and using `maximumThrowableDumpAge` worked as expected.

Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
```diff
@@ -233,4 +255,35 @@ protected function renderRequestInfo()
...
+                $dumpsToRemove = array_slice($existingDumps, $this->maximumThrowableDumpCount);
```
ahaeslich: On second view: did you want to remove files in this case regardless of their creation date?

Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
kdambekalns: Oh my… I actually remember writing this in my head:
```php
$existingDumps = glob($this->storagePath . '/*');
```
and then stopping because I wondered about `$this->storagePath` having a trainling slash or not and how `glob` orders things. That got me distracted enough, and I forgot…

Thanks for seeing this! ❤️

Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
kdambekalns: Yes, I don't care at that point. Because: The dumps files

- are ordered by name, see , where it says
   > GLOB_NOSORT -  […]. When this flag is not used, the pathnames are sorted alphabetically 
- and the name starts with a timestamp generated through `date('YmdHis', $time)`
- so the older files come first

Which makes the code almost correct… 😇 The maximum should be left, not removed, so it should be

```php
$dumpsToRemove = array_slice($existingDumps, -$this->maximumThrowableDumpCount);
```

to slice from the end of the array, see https://3v4l.org/FUGWC

kdambekalns: And now of course those are the ones that should **not** be deleted. 🤪

kdambekalns: Like this: https://3v4l.org/46LTA

kdambekalns:
@mrimann wrote:
> But no exceptions are removed, even if I configure the `maximumThrowableDumpCount: 2`.

As @ahaeslich pointed out, exactly that limit was implemented wrong… Now it should work!

Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
ahaeslich: > I actually remember writing this in my head

A classic. I definitly can relate to that 😄

ahaeslich (approved):
> It might look like a feature, but I want this in 7.3, as it can avoid disks being filled

One can easily argue that it is a bug that disks are filled with exceptions without a cleanup mechanism😉. No objection on my part on targeting 7.3.

Btw: `maximumThrowableDumpCount` works now 👍

kitsunet (changes requested):
I am against putting this into a released version, it changes the constructor of the FileStorage, api or not, this might very well be used in userland code and breaking it in a bugfix release is a nogo.

mrimann:
Thanks @kdambekalns for the second version of the change. I could now successfully test it's behaviour (with the same test-setup as outlined in the previous comment). The only functional thing I noticed is that setting `maximumThrowableDumpCount` to e.g. "2", results in 3 Dump files remaining on the disk (the two allowed ones + the new one that is written after cleaning up). But I think that's OK and still improves the situation and avoids filling up the disk with dump files.

Regarding the branch to target, I'm with @kitsunet and against adding this to older releases as it adds new config options - and the new feature of cleaning up is off by default (so just updating to the latest patchlevel release "won't solve a problem" (without manually changing some settings).

Instead I'd prefer to see a solution to the second issue https://github.com/neos/flow-development-collection/issues/3159 that I've initially brought up, backported to Neos 7.x (once there's a solution) instead of this here.

kdambekalns:
@kitsunet It changes what the constructor does internally, in a very careful way. What kind of breaking behavior do you have in mind? 🤔

@mrimann So far I did not pick "sane defaults" precisely to not change the behavior, I'd love to, though. Just not sure what would be good and somehow match the log file rotation defaults. 🤷

About the "off-by-one, true… But with anything in real-life that probably doesn't matter. 😇

mrimann:
@kdambekalns who knows what "sane" is :-) IMO good defaults ("good" as in "acceptable by large parts of the community") could be to
- remove dumps older than 30 days (if they have not been looked after within 30 days, no one ever will)
- remove older dumps in case there are >10'000 dumps on the disk (a high limit, but not that high that the disk will fill up from the dump files)

ahaeslich:
@kdambekalns & @kitsunet which version do we want to target now?

kitsunet (approved):
Lets merge it like this.

Neos.Flow/Configuration/Settings.Log.yaml
```diff
@@ -58,3 +58,7 @@ Neos:
         optionsByImplementation:
           'Neos\Flow\Log\ThrowableStorage\FileStorage':
             storagePath: '%FLOW_PATH_DATA%Logs/Exceptions'
+            # The maximum age of throwable dump in seconds, 0 to disable cleaning based on age, default 30 days
```
bwaidelich, line 61: > default 30 days

is that true?

Neos.Flow/Configuration/Settings.Log.yaml
```diff
@@ -58,3 +58,7 @@ Neos:
...
+            maximumThrowableDumpAge: 0
+            # The maximum number of throwable dumps to store, 0 to disable cleaning based on count, default 10.000
```
bwaidelich, line 63: > default 10.000

And that? I don't see any code for this default behavior

Neos.Flow/Configuration/Settings.Log.yaml
ahaeslich: We discussed at the sprint to disable this feature by default to not change the behaviour in a path release. But I guess we missed changing the comments 🙈

Neos.Flow/Configuration/Settings.Log.yaml
ahaeslich: See here. It was the original default and will be introduced in Neos 9: https://github.com/neos/flow-development-collection/pull/3213

bwaidelich: Thanks for the update (and for the record: comments for 7.3-8.x will be fixed with #3269)

PR #3203, opened 2023-10-31
BUGFIX: __wakeup must not return

kdambekalns:
Generated proxy classes return something in `__wakeup`, but that is not allowed since PHP 8.0. Root cause for this is the fact, that we check for `void` but take an "undefined return type" as "something".

This fixes the issue by checking for `'void' || null`…

**Review instructions**

The `__wakeup()` method in the generated proxy class must not contain `return $result` anymore.

**Checklist**


robertlemke (approved):
Good catch!

kdambekalns:
Ok, as I feared this is not as simple as it seemed. To be continued…

PR #3211, opened 2023-11-01
FEATURE: Add more information for object arguments in debugging

### LINKED ISSUE #3165: `Exception` log should contain value of simple dtos / read models
I have the feeling that since using more and more dtos it gets harder and harder to make sense of the written exceptions.
So instead of just saying this function was called with a `NodeTypeName` object (which is obvious as the method signature likely only allows it) - it would be really helpful knowing its value like `NodeTypeName(Foo.Bar)`

Same for more complex things like `Node`, here the id would be a start ^^

kitsunet:
For stacktraces in exceptions and logs we now render some representation of content for objects to ease debugging with DTOs.

Specifically we will try to obtain a string representation for such an object by using either in this order:

- a string cast if __toString() is available
- json_encode if it is JsonSerializable
- json_encode on the array of public properties

For readability json_encode will be limited to the first level, also all of those string representations will be cut off after 100 characters.

If any of those options works we will also shorten the className to avoid this output becoming overly long.

Note that we use JSON_PARTIAL_OUTPUT_ON_ERROR to make sure some output is provided. This might lead to partial or weird outputs depending on the object structure, but might still provide pointers for debugging.

Fixes: #3165

### DIFF
--- a/Neos.Flow/Classes/Error/Debugger.php
+++ b/Neos.Flow/Classes/Error/Debugger.php
@@ -307,6 +307,8 @@ class Debugger
             if (!$plaintext) {
                 $dump = '<a href="#o' . $objectIdentifier . '" onclick="document.location.hash=\'#o' . $objectIdentifier . '\'; return false;" class="debug-seeabove" title="see above">' . $dump . '</a>';
             }
+        } else {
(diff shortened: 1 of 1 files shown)


kdambekalns (approved):
👍 by 👀

PR #3215, opened 2023-11-03
TASK: PEG Parser declares properties

kitsunet:
Prevents deprecation warnings for dynamic properties.


dlubitz (approved):

kdambekalns (approved):
Looks innocent. But is the message right in saying "PHP 7"?

kitsunet:
@kdambekalns I admit I was lazy and should've checked which PHP version Flow 7.3 still supports (7.3) but yes if you compare the two commits, the first one I tried to be modern and added typehints that worked for me but obviously do not on php 7.3 so I downgraded them to PHP 7(.3) compatibility.

PR #3222, opened 2023-11-07
TASK: Use generics via @template instead of PHPSTORM_META

mhsdesign:
Since the php universe evolved im gonna try https://github.com/neos/flow-development-collection/pull/2753 again ;)

Adds typings to:

\Neos\Flow\ObjectManagement\ObjectManagerInterface::get()

and

\Neos\Flow\Core\Bootstrap::getEarlyInstance()

by usage of the @template tag: https://phpstan.org/writing-php-code/phpdocs-basics#generics

This feature is supported by phpstorm, psalm and phpstan and also used widely in Neos 9

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/ObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/ObjectManager.php
@@ -180,8 +180,9 @@ class ObjectManager implements ObjectManagerInterface
      * Returns a fresh or existing instance of the object specified by $objectName.
      *
      * @template T of object
-     * @param class-string<T> $objectName The name of the object to return an instance of
+     * @param class-string<T>|string $objectName The name of the object to return an instance of
      * @param mixed ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
+     * @phpstan-return ($objectName is class-string<T> ? T : object) The object instance
      * @return T The object instance
      * @throws Exception\CannotBuildObjectException
      * @throws Exception\UnknownObjectException if an object with the given name does not exist
@@ -370,7 +371,8 @@ class ObjectManager implements ObjectManagerInterface
      * registered yet.
      *
      * @template T of object
-     * @param class-string<T> $objectName The object name
+     * @param class-string<T>|string $objectName The object name
+     * @phpstan-return ($objectName is class-string<T> ? T|null : object|null) The object instance or null
      * @return T|null The object instance or null
      */
     public function getInstance($objectName)
--- a/Neos.Flow/Classes/ObjectManagement/ObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/ObjectManager.php
@@ -179,9 +179,10 @@ class ObjectManager implements ObjectManagerInterface
     /**
      * Returns a fresh or existing instance of the object specified by $objectName.
      *
-     * @param string $objectName The name of the object to return an instance of
+     * @template T of object
+     * @param class-string<T> $objectName The name of the object to return an instance of
      * @param mixed ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
-     * @return object The object instance
+     * @return T The object instance
      * @throws Exception\CannotBuildObjectException
      * @throws Exception\UnknownObjectException if an object with the given name does not exist
      * @throws \InvalidArgumentException
@@ -368,8 +369,9 @@ class ObjectManager implements ObjectManagerInterface
      * Returns the instance of the specified object or NULL if no instance has been
      * registered yet.
      *
-     * @param string $objectName The object name
-     * @return object The object or NULL
+     * @template T of object
+     * @param class-string<T> $objectName The object name
+     * @return T|null The object instance or null
      */
     public function getInstance($objectName)
     {
--- a/Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
+++ b/Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
@@ -43,9 +43,10 @@ interface ObjectManagerInterface extends ContainerInterface
      * new keyword and Singleton objects should rather be injected by some type of
      * Dependency Injection.
      *
-     * @param string $objectName The name of the object to return an instance of
-     * @param mixed[] ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
-     * @return object The object instance
+     * @template T of object
+     * @param class-string<T> $objectName The name of the object to return an instance of
+     * @param mixed ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
+     * @return T The object instance
      * @api
      */
     public function get($objectName, ...$constructorArguments);
--- a/Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
+++ b/Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
@@ -44,8 +44,9 @@ interface ObjectManagerInterface extends ContainerInterface
      * Dependency Injection.
      *
      * @template T of object
-     * @param class-string<T> $objectName The name of the object to return an instance of
+     * @param class-string<T>|string $objectName The name of the object to return an instance of
      * @param mixed ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
+     * @phpstan-return ($objectName is class-string<T> ? T : object) The object instance
      * @return T The object instance
      * @api
      */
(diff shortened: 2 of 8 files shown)


mhsdesign:
Im not sure i got all the places where generics make sense. But its a good start :D

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -362,6 +362,10 @@ public function getDefaultImplementationClassNameForInterface(string $interfaceN
      * Searches for and returns all class names of implementations of the given object type
      * (interface name). If no class implementing the interface was found, an empty array is returned.
      *
+     * @template T of object
+     * @param class-string<T> $interfaceName
+     * @return list<class-string<T>>
+     *
```
kdambekalns:
```suggestion
```

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
```diff
@@ -56,6 +57,7 @@ public function get($objectName, ...$constructorArguments);
      * @param string $objectName
      * @return bool
      * @see isRegistered
+     * don throw
```
kdambekalns: ?
```suggestion
```

kdambekalns (commented):
Yeah, why not…

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
mhsdesign: hmm it was late and this was meant to be a todo - upsi.

The PsrContainer actual describes what kind of exceptions can be thrown, but we dont keep this api promise by throwing custom exceptions. Question is, if we should throw the right exceptions?

bwaidelich: I would suggest to address this in a separate PR so that this one stays a purely "cosmetic" change

bwaidelich (approved):
Nice one, I like!

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
kdambekalns: 🤷‍♂️

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
```diff
@@ -43,9 +43,10 @@ public function getContext();
      * new keyword and Singleton objects should rather be injected by some type of
      * Dependency Injection.
      *
-     * @param string $objectName The name of the object to return an instance of
-     * @param mixed[] ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
-     * @return object The object instance
+     * @template T of object
+     * @param class-string<T> $objectName The name of the object to return an instance of
```
mficzel: How will this behave with virtual objects that only exist in objects.yaml. May be a non-issue as those are unlikely to show up in any php code but should be considered.

I may be wrong about how virtual objects are handled.

mficzel (commented):
I generally like this a lot and probably am an explanation about virtual objects away from aproving.

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
bwaidelich: Very good point. I wanted to ask the same thing but forgot about it.

It triggers a warning when used with virtual object names it seems: https://phpstan.org/r/62e1adbd-b442-423c-a4d4-4838e9a004b3

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -368,8 +369,9 @@ public function hasInstance($objectName): bool
      * Returns the instance of the specified object or NULL if no instance has been
      * registered yet.
      *
-     * @param string $objectName The object name
-     * @return object The object or NULL
+     * @template T of object
+     * @param class-string<T> $objectName The object name
+     * @return T|null The object instance or null
```
bwaidelich, line 376:
```suggestion
     * @return ($objectName is class-string ? T|null : object|null) The object instance or null
```

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
```diff
@@ -43,9 +43,10 @@ public function getContext();
...
+     * @param mixed ...$constructorArguments Any number of arguments that should be passed to the constructor of the object
+     * @return T The object instance
```
bwaidelich, line 50:
```suggestion
     * @return ($objectName is class-string ? T: object) The object instance
```

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -368,8 +369,9 @@ public function hasInstance($objectName): bool
      * Returns the instance of the specified object or NULL if no instance has been
      * registered yet.
      *
-     * @param string $objectName The object name
-     * @return object The object or NULL
+     * @template T of object
+     * @param class-string<T> $objectName The object name
```
bwaidelich:
```suggestion
     * @param class-string<T>|string $objectName The class or virtual object name
```

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
bwaidelich:
```suggestion
     * @param class-string<T>|string $objectName The class or virtual object name to return an instance of
```

bwaidelich (changes requested):
I added suggestions to support [virtual objects](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartIII/ObjectManagement.html#virtual-objects) using the PHPStan notation for [Conditional return types](https://phpstan.org/writing-php-code/phpdoc-types#conditional-return-types)

See https://phpstan.org/r/1017a495-661c-4929-9f33-8f83065ac1f4

mhsdesign:
@bwaidelich i was about to say: good point with virtual objects but we should avoid that ^^ - but your solution seems quite clever! Question is if phpstorm/psalm also supports this or do we have to use `@phpstan-return conditional stuff` and `@return T`

bwaidelich:
> Question is if phpstorm/psalm also supports this

I gave it a shot and it seems to work:
image

An additional `@return object|null` would do no harm though

mhsdesign:
> See https://phpstan.org/r/1017a495-661c-4929-9f33-8f83065ac1f4

Fyi it has to be `@return ($objectName is class-string<T>  ` (not `@return ($objectName is class-string `) otherwise phpstan doesnt work.

```php
    /**
     * @template T of object
     * @param class-string<T>|string $objectName The object name
     * @return ($objectName is class-string<T> ? T : object) The object instance
     */
```

your previous code would always say object (also for actual passed classes) - will test this also in phpstorm ^^

mhsdesign:
Hmm @bwaidelich and @mficzel it seems that the proposed super magic cow powder syntax is not supported by phpstorm yet: (newGetInstance works but not crazyNewGetInstance)

```php
/**
 * @template T of object
 * @param class-string<T> $objectName The object name
 * @return T The object instance
 */
public function newGetInstance($objectName)
{
    return $this->objects[$objectName]['i'] ?? throw new \Exception();
}

/**
 * @template T of object
 * @param class-string<T>|string $objectName The object name
 * @return ($objectName is class-string<T> ? T : object) The object instance
 */
public function crazyNewGetInstance($objectName)
{
    return $this->objects[$objectName]['i'] ?? throw new \Exception();
}
```


i would propose that people have to write this in the super rare cases, when they access an actual virtual object:

```php
/** 
 * @var MyVirtualObject $virtual
 * @phpstan-ignore-next-line
 */
$virtual = $objectManager->get('My.Virtual');
```

Neos.Flow/Classes/ObjectManagement/ObjectManagerInterface.php
mhsdesign: adjusted ;) cool that this is possible :D

mhsdesign:
With https://github.com/neos/flow-development-collection/pull/3222/commits/d4ed04889c41cc964304050c3757de618000dfb5 it will also work with virtual object by just returning `object` in those cases.

bwaidelich (approved):
Works like a charm, thanks for taking care!

## PHPStan:

https://phpstan.org/r/795a1071-4d3f-43bc-b578-6b5c53fd116c

## PHPStorm:

image

*Note:* For virtual objects, the IDE interprets the string as class name, but that's not worse than before I think:

image

kitsunet (approved):
very nice!

PR #3223, opened 2023-11-08
TASK: Remove persistence clone magic

mhsdesign:
This removed the code that set `Flow_Persistence_clone` in entities or value objects when they were `clone`d.

As dynamic properties are deprecated with PHP 8.2, this caused warnings and will eventually break.

Since this was (re)-introduced in Flow 2 via 90cb65827c1550e9144e9f83b9231b430c106660 to support custom backends in the geenric persistence layer of Flow, like the (now outdated) `TYPO3.CouchDB`, we felt it is best to remove it.

**Upgrade instructions**

If you rely on this, you need to adjust your code. Chances are, if you still need this, you use the generic peristsnece layer, which is gone in Flow 7 aready (see https://github.com/neos/flow-development-collection/pull/1769 and https://github.com/neos/flow-development-collection/pull/2262). So, you have other problems to solve, anyway…

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
+++ b/Neos.Flow/Classes/Persistence/Aspect/PersistenceMagicAspect.php
@@ -140,19 +140,4 @@ class PersistenceMagicAspect
         $proxy = $joinPoint->getProxy();
         ObjectAccess::setProperty($proxy, 'Persistence_Object_Identifier', sha1($serializedSource), true);
     }
-
-    /**
-     * Mark object as cloned after cloning.
-     *
-     * Note: this is not used by anything in the Flow base distribution,
-     * but might be needed by custom backends (like Neos.CouchDB).
-     *
-     * @param JoinPointInterface $joinPoint
-     * @return void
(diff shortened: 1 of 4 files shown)


mhsdesign:
My question would be, if all the code of 90cb65827c1550e9144e9f83b9231b430c106660 is only required [to keep CouchDB working](https://github.com/andreaswolf/TYPO3.CouchDB/blob/873e619da4d5cd9d14c88189a80853a88eb2e82b/Classes/TYPO3/CouchDB/Persistence/Backend/CouchDbBackend.php#L429-L441), and now we broke it - by removing this one feature - shouldnt we remove all of it?

kdambekalns (approved):

kdambekalns:
> My question would be, if all the code of [90cb658](https://github.com/neos/flow-development-collection/commit/90cb65827c1550e9144e9f83b9231b430c106660) […] shouldnt we remove all of it?

Generic persistence is long gone… AFAICT this was the last remaining bit.

kitsunet (approved):

PR #3229, opened 2023-11-09
!!! TASK: Introduce `TargetInterface::onPublish` callback

mhsdesign:
Currently every implementation of the `TargetInterface::publishCollection` should declare a second parameter: `callable $callback = null` which not part of the interface, but used by convention. This pattern causes trouble when using phpstan and also it’s not best practice. To improve this code and preserve the usecase partially the interface now allows to register `onPublish` callbacks, which should be called when `publishCollection` is run:

```php
interface TargetInterface
{
     // ...

     /**
      * @param \Closure(int $iteration): void $callback Function called after each resource publishing
      */
     public function onPublish(\Closure $callback): void;
}
```

**Upgrade instructions**

In case you are using the callback, you need to adjust the calling side:

```diff
- $fileSystemTarget->publishCollection($staticCollection, $myPublicationCallback);
+ $fileSystemTarget->onPublish($myPublicationCallback);
+ $fileSystemTarget->publishCollection($staticCollection);
```

Also note that the second parameter `$object` will not be passed anymore. The callback only contains the `$iteration` as one and only parameter.

Additionally the method `iterate(…)` in the `ResourceRepository` has been removed, replace it by iterating over the result of `findAllIterator()` directly.

**Checklist**


bwaidelich:
I like the possibility, but to me the implementation is not really self-explanatory at all.
For example "Function called after each object" – what is that supposed to mean?

I would suggest to add a simple event-listener mechanic instead along the lines of:
```php
enum TargetEventType {
  case PUBLISH_COLLECTION;
}

interface TargetEventListener {
  public function on(TargetEventType $eventType, Closure $callback): void;
}
```
and allow supporting targets to implement it.


For events on the getters I really wonder about the use case!?

kitsunet:
@bwaidelich that woudl be neat, I think what @mhsdesign tried to do was to expose the implicit behavior that already exists, see eg. \Neos\Flow\ResourceManagement\Target\FileSystemTarget::publishCollection()
Given that it was never in the interface I guess we could just fix our implementations and make this a (semi) breaking change? Would be nice to get this cleaner.

mhsdesign:
thanks for your input just to be clear, im not trying to defend this code, and im just trying to make it more transparent.

the current state is, that the targets and storages have by convention a nullalbe callback. Passing additional arguments to a method is not good practice and phpstan will forbid this:

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------ 
  Line   Neos.Flow/Classes/Command/ResourceCommandController.php                                                                                         
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------ 
  101    Method Neos\Flow\ResourceManagement\Target\TargetInterface::publishCollection() invoked with 2 parameters, 1 required.                          
         ✏️ Neos.Flow/Classes/Command/ResourceCommandController.php:101  
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------ 

 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Neos.Flow/Classes/ResourceManagement/Collection.php                                                                                         
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
  167    Method Neos\Flow\ResourceManagement\Storage\StorageInterface::getObjectsByCollection() invoked with 2 parameters, 1 required.               
         ✏️  Neos.Flow/Classes/ResourceManagement/Collection.php:167  
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php                                                                                         
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------- 
  199    Method Neos\Flow\ResourceManagement\CollectionInterface::getObjects() invoked with 1 parameter, 0 required.                                              
         ✏️  Neos.Flow/Classes/ResourceManagement/Target/FileSystemTarget.php:199  
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

the root cause seems to be in the `ResourceCommandController` https://github.com/neos/flow-development-collection/blob/31f5dff4fed437b658a48f6a05df47258fa4dcf0/Neos.Flow/Classes/Command/ResourceCommandController.php#L101 were we call `$this->persistenceManager->clearState();` every 1000nth time.

But i have used the callback logic to have a nice progress bar in the cli which advances.

bwaidelich:
Ah right, thanks for the clarification!
I heard that Github will add an "create summary" button for PRs - this will be helpful ;)

kdambekalns (dismissed):
Marked as breaking but no clear instructions in the PR message.

(Generally speaking, more love for the message, e.g. removing cruft from the template, would be cool, @mhsdesign 😇)

Neos.Flow/Classes/ResourceManagement/Storage/StorageInterface.php
```diff
@@ -52,17 +52,19 @@ public function getStreamByResourcePath($relativePath);
     /**
      * Retrieve all Objects stored in this storage.
      *
+     * @param callable(int $iteration, StorageObject $object): void $callback Function called after each object
```
kdambekalns:
```suggestion
     * @param callable(int $iteration, StorageObject $object): void $callback Function called after each iteration
```

Neos.Flow/Classes/ResourceManagement/CollectionInterface.php
```diff
@@ -81,9 +81,10 @@ public function publish();
     /**
      * Returns all internal data objects of the storage attached to this collection.
      *
+     * @param callable(int $iteration, StorageObject $object): void $callback Function called after each object
```
kdambekalns:
```suggestion
     * @param callable(int $iteration, StorageObject $object): void $callback Function called after each iteration
```

kdambekalns (commented):
Maybe those suggestions "fix" what @bwaidelich mentioned about the comment being unclear?

kitsunet:
For me part of the problem is to actually bake this in the interface when we know this is not the nicest way.

mhsdesign:
I mean its not that important to me. We could also just `@phpstan-ingnore-next-line todo we do stuff here`

bwaidelich:
I checked the implementation again and.. wow is that quirky :)
* `StorageInterface::getObjects()` and `StorageInterface::getObjectsByCollection()` are apparently never called with a `$callback` in the Flow & Neos core
* `CollectionInterface::getObjects()` is only called with the `$callback` from within `publishCollection()` methods of `ProxyAwareTargetTrait` and `FileSystemTarget`.
* `TargetInterface::publishCollection()` is called with an `$callback` in `ResourceCommandController::publishCommand()` to clear persistence state every 1000th iteration

So maybe my idea is not too far off:

Replace
```php
public function publishCollection(CollectionInterface $collection, callable $callback = null)
{
    // ...
    foreach ($collection->getObjects($callback) as $object) {
```
with
```php
public function publishCollection(CollectionInterface $collection)
{
    // ...
    foreach ($collection->getObjects() as $object) {
       $this->invokeOnPublishCallbacks($object);
```
in `ProxyAwareTargetTrait` and `FileSystemTarget` and then add some `onPublish(Closure $callback): void` to...

* a) ...the `TargetInterface` (= breaking change)
* b) ...a new interface (e.g. `TargetEventListener`) and implement that from the `FileSystemTarget` and `ProxyAwareTargetTrait` (= backwards compatible, but classes using the trait should implement the interface)
* c) ...`FileSystemTarget` and `ProxyAwareTargetTrait` (without interface)

kdambekalns:
I'd love to see this "fixed" now that we are aware and have a major around the corner that allows us to break things. 😇

~I'd vote for b) from https://github.com/neos/flow-development-collection/pull/3229#issuecomment-1804216203~

Update: Hm… our targets would no longer call the callback then, I guess? So that's breaking without being breaking?

Maybe I'd rather vote for a) from https://github.com/neos/flow-development-collection/pull/3229#issuecomment-1804216203

bwaidelich:
> our targets would no longer call the callback then, I guess? So that's breaking without being breaking?

Currently it's not documented and we only invoke it from `ResourceCommandController::publishCommand()` to clear persistence state..
But yes, extending the `TargetInterface` makes custom implementations at least aware of that change.

mhsdesign:
Tadaaa: (with a sleep statement 😂 - but i had a case lately where i needed this ^^)

https://github.com/neos/flow-development-collection/assets/85400359/a9f6696b-672c-448c-8445-69445fccbeee

Neos.Flow/Classes/ResourceManagement/Storage/FileSystemStorage.php
```diff
@@ -161,14 +158,6 @@ public function getObjectsByCollection(CollectionInterface $collection, callable
...
             });
             yield $object;
-            if ($callback !== null) {
-                call_user_func($callback, $iteration, $object);
-            }
-            $iteration++;
-        }
-        if ($iteration === 0) {
-            // required because if the collection is empty, this should still return an empty generator
-            yield from [];
```
mhsdesign, line 171: i doubt that this https://github.com/neos/flow-development-collection/pull/2167 fixed this. Should not be necessary but ill write a test ^^

mhsdesign: see added test ;)

mhsdesign:
Okay so i implemented `a)` and it seems to work see my gif above ^^

This is now ready for review again. Thanks for your input.

Neos.Flow/Classes/Command/ResourceCommandController.php
```diff
@@ -94,13 +94,21 @@ public function publishCommand(string $collection = null)
             }
 
             foreach ($collections as $collection) {
+                $progressIndicator = new ProgressIndicator($this->output->getOutput());
                 try {
                     /** @var CollectionInterface $collection */
                     $this->outputLine('Publishing resources of collection "%s"', [$collection->getName()]);
+                    $progressIndicator->start('Published 0');
```
kdambekalns: If this is used in automated ways, the output might be unexpected. What about adding a `quiet` flag, so it can at least be disabled?

Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
```diff
@@ -32,18 +32,19 @@ class FileSystemSymlinkTarget extends FileSystemTarget
...
-     * @param callable $callback Function called after each resource publishing
      * @return void
      */
-    public function publishCollection(CollectionInterface $collection, callable $callback = null)
+    public function publishCollection(CollectionInterface $collection)
     {
         $storage = $collection->getStorage();
         if ($storage instanceof PackageStorage) {
             foreach ($storage->getPublicResourcePaths() as $packageKey => $path) {
+                // FIXME we cannot invoke the callback here as we just symlink the whole folder,
```
kdambekalns:
```suggestion
                // TODO we cannot invoke the callback here as we just symlink the whole folder,
```
TODO is potentially easier to find (by humans and tools alike)

Also, what could we potentially improve here?

Neos.Flow/Classes/Command/ResourceCommandController.php
mhsdesign: sure. I think @robertlemke was thinking about generic `--quiet` and `--no-input` for a while ... but for this case we could start with an explicit `$quiet` parameter

mhsdesign: adjusted

Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php
mhsdesign: Fixed it by actually calling the callback so 

`flow resource:publish --collection static`

will show

```
Publishing resources of collection "static"
 - Published 11
```

The 11 will in that case not reflect the number of real storageObjects but rather how many folders have been symlinked.

Neos.Flow/Classes/ResourceManagement/Target/SymlinkedFolderResourceObject.php
```diff
@@ -0,0 +1,62 @@
...
+declare(strict_types=1);
+
+namespace Neos\Flow\ResourceManagement\Target;
+
+use Neos\Flow\ResourceManagement\ResourceMetaDataInterface;
+
+/**
+ * @internal only for the callback {@see TargetInterface::onPublish()}
+ */
+class SymlinkedFolderResourceObject implements ResourceMetaDataInterface
```
mhsdesign: This might need some discussion and feel odd to you.
Its a necessity of the FileSystemSymlinkTarget to satisfy the `onPublish` callbacks

mhsdesign:
Ping ;) This working and ready for review and merge ;)

bwaidelich (approved):
Nice one, thanks. +1 by reading

mhsdesign:
@kdambekalns it would be great if you could re-review this as well ;)

mhsdesign:
Okay ill merge this tomorrow if no one opposes

Neos.Flow/Classes/ResourceManagement/Target/SymlinkedFolderResourceObject.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Neos\Flow\ResourceManagement\Target;
+
+use Neos\Flow\ResourceManagement\ResourceMetaDataInterface;
+
+/**
+ * @internal only for the callback {@see TargetInterface::onPublish()}
```
bwaidelich: Not entirely true, it is used above in Neos.Flow/Classes/ResourceManagement/Target/FileSystemSymlinkTarget.php – and that makes me wonder

bwaidelich (commented):
I looked over this once again and I stumbled upon the `SymlinkedFolderResourceObjectForPublishingCallback` that only seems to exist in order to satisfy static types and is not really useful at all (because it throws exceptions for almost every method invokation).

IMO we should either:
a) define a base interface that only contains the `getFilename()` method
b) omit the 2nd parameter from the callback (but IIRC that is currently used?)

mhsdesign:
we (@kitsunet and @bwaidelich) discussed to opt in for option b

> omit the 2nd parameter from the callback

as the parameter is not used in the source code.

kitsunet:
post merge thumb up

robertlemke:
I'm late to the discussion. Thanks a lot for the initiative, but 👎 from my side to this particular implementation.

I think that the method name `onPublish()` is misleading: From reading it, I would suspect that it is an event listener which is called whenever something is published. But in fact it is a method for *registering* an event listener. And introducing just one method to that interface feels a bit short-sighted (what if there are more interesting events in the future – those would require more breaking changes).

My suggestion:

- go the extra mile and revert this again (I also need to revert changes then in made in other packages)
- go for the `TargetEventListenerInterface` approach @bwaidelich suggested.

To simplify things, a standard `TargetEventListener` class can be provided:

```php
foreach ($collections as $collection) {
  $progressIndicator = new ProgressIndicator($this->output->getOutput());
  try {
    $this->outputLine('Publishing resources of collection "%s"', [$collection->getName()]);
    $progressIndicator->start('Published 0');
    $target = $collection->getTarget();
    $lastIteration = 0;
    $target->addListener(
      TargetEventListener::withOnPublish(
        function ($iteration) use ($progressIndicator, &$lastIteration) {
          $progressIndicator->advance();
          $progressIndicator->setMessage(sprintf('Published %s', $iteration));
          $this->clearState($iteration);
          $lastIteration = $iteration;
        });
      )
    );
  …
```

bwaidelich:
> I think that the method name onPublish() is misleading

The name comes from similar `on*` callbacks (in JS etc).
It's not great but from the using side it is pretty self-explanatory IMO:
```php
$target->onPublish($this->doSomething(...));
```

Problem with a new interface is, that things might not work properly if 3rd parties don't implement the interface (e.g. we persist changes every x files in the core – that simply won't be triggered if the interface is not implemented).

So IMO the breaking change was OK in this case.

All options had their down-sides and this whole part would need some rewrite because there are a lot more oddities in there. However, feel free to go the extra mile if you think it's worth it!

robertlemke:
> Problem with a new interface is, that things might not work properly if 3rd parties don't implement the interface (e.g. we persist changes every x files in the core – that simply won't be triggered if the interface is not implemented).

Right now, most publishing targets are just implementing a dummy onPublish() method to satisfy the interface. But there's no information about what targets should do, for example the `onPublish()` method in the interface is not documented.

Anyway, you all discussed this in more detail, so these were just my 2ct and if you agree that it's a good compromise then let's leave it at that.

mhsdesign:
I must agree that there are some valid points on both sides.
I dont think its necessary to revert this pr but i would like a new pr on top of this were we can discuss the direction, otherwise this discussion will just be swallowed by the web.

Thanks for your input robert!

mhsdesign:
Tried to improve the doc block a bit via https://github.com/neos/flow-development-collection/pull/3270

mhsdesign:
Btw i found accidentally the right commit where this was introduced:

https://github.com/neos/flow-development-collection/commit/879fba19f93d0a8628682698e57da9f1b58ad7d4

Maybe we can just rethink this whole commit and do things differently?

Without the questionable callback to clear doctrine states?

...But im not that deep into that.

PR #3232, opened 2023-11-11
!!! FEATURE: WIP Dispatcher and controller return `ActionResponse` (simpler controller pattern)

kitsunet:
will not be merged directly into 9.0 but included in this mvc overhaul pr: https://github.com/neos/flow-development-collection/pull/3311

This change needs an accompanying adjustment to Neos to adjust the
PluginImplementation as well as Modules.

~The new `SimpleActionController` gives you a direct and simple way to
route an ActionRequest and return an ActionReponse with nothing in
between. Routing should work just like with other ActionControllers.~

This is breaking if you implemented your own ControllerInterface
or overwrote or expect some of the api methods of the ActionController.
We now use a direct pattern of f(ActionRequest) => ActionResponse
in more places. Adjusting existing controllers should be easy though.
Additionally implementing your own dispatch loop (don't do this) will
need adjustments.

We discourage to manipulate `$this->reponse` in controllers,
although it should still work fine in actions for now, please consider
other options.

```php
class Dispatcher
{
    public function dispatch(ActionRequest $request): ActionResponse;
}
```


kitsunet:
Probably still has a few rough edges but I would like to discuss this before putting more time into it. I made it so that for probably the 90% userland case nothing should change, but open options, the few people that did overwrite/mess around with the interface and methods I changed shoudl have an easy time adapting. we could go way further with this paying for it with more breakiness.

Neos.Flow/Classes/Mvc/Controller/ControllerInterface.php
```diff
@@ -34,12 +34,11 @@ interface ControllerInterface
      * Processes a general request. The result can be returned by altering the given response.
      *
      * @param ActionRequest $request The request object
-     * @param ActionResponse $response The response, modified by the controller
-     * @return void
+     * @return ActionResponse $response The response, modified by the controller
```
mhsdesign:
```suggestion
     * @return ActionResponse $response The response, created by the controller
```

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
```diff
@@ -0,0 +1,47 @@
...
+     * @param ActionRequest $request
+     * @return ActionResponse
+     * @throws NoSuchActionException
+     * @throws InvalidSlotException
+     */
+    public function processRequest(ActionRequest $request): ActionResponse
+    {
+        $request->setDispatched(true);
+        $actionMethodName = $this->resolveActionMethodName($request);
+        return $this->$actionMethodName($request);
```
mhsdesign: i think we should still ensure that the action is public, maybe at runtime by calling it from another class (maybe via an anonymous class?)

kitsunet: Yeah I thought about it, but felt this to be an overprotective precaution, after all it needs to end in `Action` anyways to be even callable, and the signature needs to fit otherwise it will just die as well, and in these simple controllers what else should there be.

And I think it would be actually nicer code wise to have it be not public?

mhsdesign:
Hi :D Thanks for this wip approach to have something concrete to discuss!. This new controller interface and the SimpleActionController look both real good. But i see that we have to go through some lengths to achieve this response manipulating "feature".

Im not entirely sure if this is the best way to have a common interface for both legacy and new controllers. I think the code manipulation of the `response` is currently not perfectly readable but we can rely on it working and i think the readability decreases for our legacy layer when we use exception throwing with response as field for control flow (clever idea though ^^).

How about to keep our legacy controller code base as is and instead introduce a new interface?

```php
interface LegacyActionControllerInterface
{
     public function processRequest(ActionRequest $request, ActionResponse $response): void;
} 
```

```php
interface SimpleActionControllerInterface
{
     public function processRequest(ActionRequest $request): ActionResponse;
} 
```

I also have been thinking about an controller which would work entirely only with psr requests and responses so that our users can apply their knowledge from other php frameworks and dont have to panic: `header(), echo "", die()`
That can  work if the dispatcher could figure out which method has to be called on the controller and the interface would just be a simple marker:


```php
/** implement your actions by the pattern {$name}Action, process the psr requet and return a response */
interface PsrControllerInterface
{
   // @example
   // public function indexAction(ServerRequest $request): Response;
} 
```

kitsunet:
But then you have a lot of weird checks and contortions in the lower layers? As in the dispatcher and any other code (plugins, widgets etc.) need to suddenly deal with two ways to call a controller. IDK I find the readability not too bad actually. And now the weirdness happens at least only inside the ActionController. And then we can step by step reduce the surface of this weirdness down to only the action and declare that the response is only globally available during the action method call. Could even be something we can do right now, not sure how much of our core contorllers do something beyond that.

Again PSR is a separate concern and not part of the dispatcher, that is a lower level we would have to build separate infrastructure for.

mficzel:
I really like this approach. No property mapping and no mimetype negotiation to produce unexpected trouble \o/.

@mhsdesign i think that this makes more sense with FlowActionRequest / Response as otherwise it would be more like a middleware. The middleware approach will probably be the path anyway $someone chooses if one does not want to learn a new framework.

kitsunet:
I think I do see a way in the future, we could move everything the ActionController does into the userland with services you can call. Like we deprecate _regular_ actions at some point and say you just make it a SimpleController action and use the following services to get the same results (ala propertymapper -> validator -> view) and then the old code is just wrapped by a new simple controller action with a few lines of boilerplate to init the stuff.

Future backwards compat controller way suggestion:
This would come in a future PR then when we decide to deprecate ActionController at some point.

```
class MyOldStyleController extends ActionController {
    // your classic old action method that would need to be migrated
    public function oldStyleAction(MyObject $foo, string $bar) {
        // do stuff
        $this->view->....
        return ...;
    }
}
```

```
class MyOldStyleController extends SimpleActionController {
    // provides "legacy" functionality like mapAndValidate for arguments
    use ActionControllerCompatibility;

    // the new public action method that re-implements the magic and wraps the logic of the old one
    public function oldStyleAction(ActionRequest $request) {
         $preparedArguments = $this->mapAndValidate('oldStyleActionMethod', $request); // provided by ActionControllerCompatibility trait
         $preparedView = $this->prepareViewFor('oldStyleActionMethod', $request); // provided by ActionControllerCompatibility trait
         // we would provide request and view here as additional arguments as to avoid the object state.
         $result = $this->oldStyleActionMethod(...$preparedArguments, $request , $view);
         return $this->mapResultsAndViewToReponse($result, $view); // provided by ActionControllerCompatibility trait
    }

    // your classic old action method with extended signature to provide request and view, can basically do the same as before
    public function oldStyleActionMethod(MyObject $foo, string $bar, $request, $view) {
        // do stuff
        $view->....
        return ...;
    }
}
```

Total sketch code obviously but probably something we could rector...

Neos.Flow/Classes/Mvc/Controller/SpecialResponsesSupport.php
```diff
@@ -0,0 +1,99 @@
...
+     * Sends the specified HTTP status immediately.
+     *
+     * @param integer $statusCode The HTTP status code
+     * @param ?string $statusMessage A custom HTTP status message
+     * @param string $content Body content which further explains the status
+     * @param ActionResponse $response
+     * @return never
+     * @throws StopActionException
+     */
+    protected function reponseThrowsStatus(int $statusCode, ?string $statusMessage = null, string $content = '', ?ActionResponse $response =  null): never
```
dlubitz:
```suggestion
    protected function responseThrowsStatus(int $statusCode, ?string $statusMessage = null, string $content = '', ?ActionResponse $response =  null): never
```

Neos.Flow/Classes/Mvc/Controller/SpecialResponsesSupport.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * @param ActionResponse $response
+     * @param string $message
+     * @param int $code
+     * @return never
+     * @throws StopActionException
+     */
+    protected function throwStopActionWithReponse(ActionResponse $response, string $message = '', int $code = 0): never
```
dlubitz:
```suggestion
    protected function throwStopActionWithResponse(ActionResponse $response, string $message = '', int $code = 0): never
```

Neos.Flow/Classes/Mvc/Controller/SpecialResponsesSupport.php
```diff
@@ -0,0 +1,99 @@
...
+     * @return ActionResponse
+     * @throws StopActionException
+     */
+    protected function responseRedirectsToUri(UriInterface $uri, int $delay = 0, int $statusCode = 303, ?ActionResponse $response = null): ActionResponse
+    {
+        $response = $response ?? new ActionResponse;
+
+        if ($delay < 1) {
+            $response->setRedirectUri($uri, $statusCode);
+            $this->throwStopActionWithReponse($response, '', 1699478812);
```
dlubitz:
```suggestion
            $this->throwStopActionWithResponse($response, '', 1699478812);
```

Neos.Flow/Tests/Unit/Mvc/Controller/AbstractControllerTest.php
```diff
@@ -350,10 +357,11 @@ public function myIndexAction(): void
...
         try {
-            $controller->processRequest($this->mockActionRequest, $this->actionResponse);
-        } catch (StopActionException) {
-            Assert::assertSame('the_uri', $this->actionResponse->getRedirectUri()?->__toString());
-            Assert::assertSame(303, $this->actionResponse->getStatusCode());
+            $controller->processRequest($this->mockActionRequest);
+        } catch (StopActionException $exception) {
+            $actionReponse = $exception->response;
+            Assert::assertSame('the_uri', $actionReponse->getRedirectUri()?->__toString());
+            Assert::assertSame(303, $actionReponse->getStatusCode());
```
dlubitz:
```suggestion
            $actionResponse = $exception->response;
            Assert::assertSame('the_uri', $actionResponse->getRedirectUri()?->__toString());
            Assert::assertSame(303, $actionResponse->getStatusCode());
```

Neos.Flow/Tests/Unit/Mvc/Controller/AbstractControllerTest.php
```diff
@@ -309,10 +312,11 @@ public function myIndexAction(): void
...
```
dlubitz:
```suggestion
            $actionResponse = $exception->response;
            Assert::assertSame('the_uri', $actionResponse->getRedirectUri()?->__toString());
            Assert::assertSame(303, $actionResponse->getStatusCode());
```

Neos.Flow/Tests/Unit/Mvc/DispatchMiddlewareTest.php
```diff
@@ -78,10 +79,14 @@ protected function setUp(): void
      */
     public function processDispatchesTheRequest()
     {
+        $testContentType = 'audio/ogg';
         $this->mockHttpRequest->method('getQueryParams')->willReturn([]);
-        $this->mockDispatcher->expects(self::once())->method('dispatch')->with($this->mockActionRequest);
+        $testReponse = new ActionResponse();
+        $testReponse->setContentType($testContentType);
+        $this->mockDispatcher->expects(self::once())->method('dispatch')->with($this->mockActionRequest)->willReturn($testReponse);
```
dlubitz:
```suggestion
        $testResponse = new ActionResponse();
        $testResponse->setContentType($testContentType);
        $this->mockDispatcher->expects(self::once())->method('dispatch')->with($this->mockActionRequest)->willReturn($testResponse);
```

Neos.FluidAdaptor/Tests/Unit/Core/Widget/AjaxWidgetMiddlewareTest.php
```diff
@@ -119,6 +120,8 @@ protected function setUp(): void
         $this->inject($this->ajaxWidgetMiddleware, 'hashService', $this->mockHashService);
 
         $this->mockDispatcher = $this->getMockBuilder(Dispatcher::class)->getMock();
+        $actionReponse = new ActionResponse();
+        $this->mockDispatcher->expects(self::any())->method('dispatch')->willReturn($actionReponse);
```
dlubitz:
```suggestion
        $actionResponse = new ActionResponse();
        $this->mockDispatcher->expects(self::any())->method('dispatch')->willReturn($actionResponse);
```

Neos.FluidAdaptor/Tests/Unit/Core/Widget/AjaxWidgetMiddlewareTest.php
```diff
@@ -177,7 +180,8 @@ public function handleDispatchesActionRequestIfWidgetContextIsPresent()
         $mockActionRequest = $this->getMockBuilder(ActionRequest::class)->disableOriginalConstructor()->getMock();
         $this->mockActionRequestFactory->method('prepareActionRequest')->willReturn($mockActionRequest);
 
-        $this->mockDispatcher->expects(self::once())->method('dispatch');
+        $actionReponse = new ActionResponse();
+        $this->mockDispatcher->expects(self::once())->method('dispatch')->willReturn($actionReponse);
```
dlubitz:
```suggestion
        $actionResponse = new ActionResponse();
        $this->mockDispatcher->expects(self::once())->method('dispatch')->willReturn($actionResponse);
```

Neos.Flow/Classes/Mvc/Controller/SpecialResponsesSupport.php
```diff
@@ -0,0 +1,99 @@
...
+        $response->setStatusCode($statusCode);
+        if ($content === '') {
+            $content = sprintf(
+                '%s %s',
+                $statusCode,
+                $statusMessage ?? ResponseInformationHelper::getStatusMessageByCode($statusCode)
+            );
+        }
+        $response->setContent($content);
+        $this->throwStopActionWithReponse($response, $content, 1558088618);
```
dlubitz:
```suggestion
        $this->throwStopActionWithResponse($response, $content, 1558088618);
```

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -311,22 +322,17 @@ protected function redirectToRequest(ActionRequest $request, $delay = 0, $status
...
-        } else {
-            $this->response->setStatusCode($statusCode);
-            $this->response->setContent('<html><head><meta http-equiv="refresh" content="' . (int)$delay . ';url=' . $uri . '"/></head></html>');
+        if (!$uri instanceof UriInterface) {
+            $uri = new Uri($uri);
         }
-        throw new StopActionException();
+
+        $response = $this->responseRedirectsToUri($uri, $delay, $statusCode, $this->response);
+        $this->throwStopActionWithReponse($response, '', 1699716808);
```
dlubitz:
```suggestion
        $this->throwStopActionWithResponse($response, '', 1699716808);
```

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -351,25 +357,29 @@ protected function throwStatus(int $statusCode, $statusMessage = null, $content
             );
         }
         $this->response->setContent($content);
-        throw new StopActionException($content, 1558088618);
+        $this->throwStopActionWithReponse($this->response, $content, $statusCode);
```
dlubitz:
```suggestion
        $this->throwStopActionWithResponse($this->response, $content, $statusCode);
```

dlubitz (commented):
I like that idea. I just would challange the name "SimpleActionController", but I have no better idea at the moment.

(Sorry for the typo change-requests at this point... I didn't expect that much and I was afraid it would be forgotten in the end)

kitsunet:
Note that if you want to test this with Neos you need to apply the patch for it as well 
https://github.com/neos/neos-development-collection/pull/4738

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -216,30 +229,7 @@ protected function forward($actionName, $controllerName = null, $packageKey = nu
...
-     */
-    protected function forwardToRequest(ActionRequest $request)
-    {
-        $packageKey = $request->getControllerPackageKey();
-        $subpackageKey = $request->getControllerSubpackageKey();
-        if ($subpackageKey !== null) {
-            $packageKey .= '\\' . $subpackageKey;
-        }
-        $this->forward($request->getControllerActionName(), $request->getControllerName(), $packageKey, $request->getArguments());
+        $this->forwardToRequset($nextRequest);
```
kitsunet: Haha, the typo makes this work... going to fix...

mficzel:
@kitsunet like this change still by reading but wonder wether the Fusion ControllerContext will still be able to manipulate the action response as before?

Side note: This is used in fusion form here https://github.com/neos/fusion-form/blob/main/Classes/Runtime/FusionObjects/RuntimeFormImplementation.php#L75-L89 and here https://github.com/neos/fusion-form/blob/main/Classes/Runtime/FusionObjects/RuntimeFormImplementation.php#L171-L175 mostly for setting location headers in redirect actions.

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
mhsdesign: I think `public` makes more sense, as otherwise code sniffing things will immediately say its unused?

Also it feels like some kind of entry point, even if not a conventional method...

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
```diff
@@ -0,0 +1,49 @@
...
+ * Provides most direct access to our request/response abstraction,
+ * the request comes directly from the dispatcher and goes directly back to it.
+ *
+ * For helpers to facilitate throws, forwards, redirects:
+ * @see SpecialResponsesSupport
+ *
+ * Views or other processing needs to be added to your controller as needed,
+ * helpers will be suggested here as they become available.
+ */
+class SimpleActionController implements ControllerInterface
```
mhsdesign, line 19: ha as a new class we can finally avoid making mistakes concerning what api is and whats not :D

I guess extending is okay xD - no just joking - but then we should probably mark it `@api`, also as there is relatively little code in `processRequest` and `resolveActionMethodName`, we _can_ consider making both final, to limit api scope, as creating an own plain controller by implementing the `ControllerInterface` is now super simple.

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
```diff
@@ -0,0 +1,49 @@
...
+        return $this->$actionMethodName($request);
+    }
+
+    /**
+     * Resolves and checks the current action method name
+     *
+     * @return string Method name of the current action
+     * @throws NoSuchActionException
+     */
+    protected function resolveActionMethodName(ActionRequest $request): string
```
mhsdesign: ha as a new class we can finally avoid making mistakes concerning what api is and whats not :D

I guess extending is okay xD - no just joking - but then we should probably mark it `@api`, also as there is relatively little code in `processRequest` and `resolveActionMethodName`, we _can_ consider making both final, to limit api scope, as creating an own plain controller by implementing the `ControllerInterface` is now super simple.

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @param ActionRequest $request
+     * @return ActionResponse
+     * @throws NoSuchActionException
+     * @throws InvalidSlotException
+     */
+    public function processRequest(ActionRequest $request): ActionResponse
+    {
+        $request->setDispatched(true);
```
mhsdesign: why does the action controller have to set it to dispatched if the dispatcher could do it too? Feels more clean and will make it even easier to have an own implementation for the `ControllerInterface` as there would be 0 boiler plate code ❤️

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
mhsdesign: using `Simple*` as pattern because the initial implementation blow up too much is okayish, but could maybe be improved?

Pro `SimpleActionController`:
- its obvious when people look for action controller they are able to find this one
- psr _simple_ cache uses the same naming pattern

Con:
- indicates that somewhere in the codebase there is probably codesmell :P

But i was not able to come with an alternative naming so this is probably best ^^

Neos.Flow/Classes/Mvc/Controller/SpecialResponsesSupport.php
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Sends the specified HTTP status immediately.
+     *
+     * @param integer $statusCode The HTTP status code
+     * @param string $content Body content which further explains the status the body of a given response will be overwritten if this is not empty
+     * @param ActionResponse|null $response The response to use or null for an empty response with the given status and message or content
+     * @return never
+     * @throws StopActionException
+     */
+    protected function responseThrowsStatus(int $statusCode, string $content = '', ?ActionResponse $response =  null): never
```
mhsdesign: do we want that people override those helpers? Otherwise they should be `final` wdym?

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
kitsunet: Fair point, I 'll change it.

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Resolves and checks the current action method name
+     *
+     * @return string Method name of the current action
+     * @throws NoSuchActionException
+     */
+    protected function resolveActionMethodName(ActionRequest $request): string
+    {
+        $actionMethodName = $request->getControllerActionName() . 'Action';
```
mhsdesign: do we want to keep the `*Action` pattern, because it still happens to me that i miss the suffix and it takes some time to figure out again what i missed 😅 , to fix this we could just improve the `routing:match` command or the like to emit a warning, or wait we cannot as this is highly dynamic and we have to have an actual request at hand.

Neos.Flow/Classes/Mvc/Exception/StopActionException.php
```diff
@@ -23,4 +25,5 @@
  */
 class StopActionException extends \Neos\Flow\Mvc\Exception
 {
+    public ActionResponse $response;
 }
```
mhsdesign: as the `StopActionException` is only created at one place, and immediately enriched with its action request, i would strongly prefer to make it mandatory in the constructor and also immutable. Also we should document _why_ the request is attached at all and what its meaning is.

The same applies for the `ForwardException`, which is also only created once.

_But_ if you feel this is to breaky right now and will mudden your change im fine with creating an additional pr to fix this before the 9.0 release.

Additonally while were at it, to make our code more stable, it think its fair to say that these both kind of exceptions are final, as our framework would probably just break if one tries to inherit from them. I also dont think it makes sense for the  `StopActionException` to be non final just that the `ForwardException` can inherit from that. _Can_ be separate final objects imo.

Neos.Flow/Tests/Functional/Http/Fixtures/Controller/FooController.php
```diff
@@ -11,16 +11,19 @@
  * source code.
  */
 
+use Neos\Flow\Mvc\ActionResponse;
 use Neos\Flow\Mvc\Controller\AbstractController;
 
 class FooController extends AbstractController
```
mhsdesign, line 17:
```suggestion
class FooController implements ControllerInterface
```

if we override everything there is not much gain

Neos.Flow/Classes/Mvc/Controller/ActionController.php
```diff
@@ -624,14 +631,14 @@ public static function getPublicActionMethods($objectManager)
      * @return ViewInterface the resolved view
      * @throws ViewNotFoundException if no view can be resolved
      */
-    protected function resolveView()
+    protected function resolveView(ActionRequest $request)
```
mhsdesign, line 635: i think you explained, but i forgot. Why do we pass in here the $request explicitly, while at other places we use `$this->request` and `$this->response`?

mhsdesign (changes requested):
Hi i had a little in depth look. Thanks for your cleanup ❤️, i think i changed my mind this _does_ make the code better :D
I know i asked many questions but provided little answer, and thats mean. So i would like to have a call with you about this so we can move quickly to the right solutions and eventually merge this little beast.

mhsdesign:
My proposed separation of the `ForwardException` and the `StopActionException` really makes sense and makes the code less quirky, i created a pr for that ontop of yours so you can review that :P https://github.com/kitsunet/flow-development-collection/pull/3

Neos.Flow/Classes/Mvc/Exception/UnsupportedRequestTypeException.php
```diff
@@ -14,7 +14,7 @@
 /**
  * An "Unsupported Request Type" exception
  *
- * @api
+ * @deprecated since Flow 9.0
  */
 class UnsupportedRequestTypeException extends \Neos\Flow\Mvc\Exception
```
mhsdesign, line 19: i could not find any instantiation anymore, only imports and dead `@throws`. So it should be safe to remove this exception. (Certainly in another pr) What do you mean?

mhsdesign:
I few more notes about slots and the cli infrastructure i stumbled upon:

Ha i found out the cli dispatching stuff works exactly like before our mvc refactoring, but its internal so i guess we can leave it for now? https://github.com/neos/flow-development-collection/blob/9ec290133b3da3d23068cf99898bb6392ab7e442/Neos.Flow/Classes/Cli/CommandControllerInterface.php#L20

The cli dispatcher actually invokes the mvc slots (with wrong parameters) and i wondered if this is really a wise idea to do?
https://github.com/neos/flow-development-collection/blob/add3b78597c5a8a090205c4ac8f4352ea9d00d43/Neos.Flow/Classes/Cli/Dispatcher.php#L89-L93
Seems to be some kind of legacy layer after the [split](https://github.com/neos/flow-development-collection/commit/cf55b180c953de8d02adb680216f5c24a3524237#diff-2c9408e74a8ac737f84468e74c23956d2057e64196e66b97281905d8697226ca)?

$some people might actually use it though see here https://github.com/neos/flow-development-collection/pull/2529 ^^

And we use the `afterControllerInvocation` ourselves to persist object during safe requests: https://github.com/neos/flow-development-collection/blob/984971c4e18228420812bbc80768bf399ae207dd/Neos.Flow/Classes/Package.php#L68-L80

-------

In the mentioned review pr, i now also adjusted the signals to not except any empty responses but rather null in that case. Feels more sane and people using this might have a chance of figuring out why stuff isnt working as before. A passed object which can be modified, but shouldnt is an easy footgun.

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
kitsunet: We _could_ make this a little service I guess, then you either use it or replace it as needed...

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
kitsunet: I wondered the same thing, it's a bit more hidden in the ActionController, but distilled down like here it seems glaringly obvious something is wrong LOL. I will dig a bit into history as I guess something (that obviously no one cares about) was lost in the PSR HTTP refactoring. I think (as we can see in the dispatch loop) there was a possibility before for a controller to gain control but decide it does not finally dispatch the request and thus the dispatch loop would continue and try to find another controller to take care, so it was the controllers responsibility to set this, but I am not sure what conditions it previously attached to it being dispatched. I'll have a look or provide maybe a little pre PR to this that just removes the thing?

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
kitsunet: Yeah happy for any other name LOL. 

We could just call it WebController 😆

Neos.Flow/Classes/Mvc/Controller/SpecialResponsesSupport.php
kitsunet: right, no, no overwriting I think, you can just use your own code if you need some other behavior.

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
kitsunet: Mmm, yeah I thought about changing it, and I wouldn't mind that, for me I would actually say just make it an attribute? just allowing any method (without marker like an attribute) seems dangerous-ish especially if someone doesn't do proper sec policies. Also this then wouldn't match any existing policies as they all assume Action, so I just kept it that way.

Neos.Flow/Classes/Mvc/Exception/StopActionException.php
kitsunet: Good plan 👍  I would just do it I think, it's a major version and I doubt too many instantiate those.

Neos.Flow/Classes/Mvc/Controller/ActionController.php
kitsunet: I ultimately want to get rid of `$this->request` and `$this->response` so took steps were I felt it is not too breaky. We can go over this and decide were we want to make that boundary and were not. I know it looks a bit hack right now, but I think whatever we already dare to clamp down right now makes sense to do as it eases potential refactoring (like making everything views related a separate service).

Neos.Flow/Classes/Mvc/Exception/UnsupportedRequestTypeException.php
kitsunet: Right, why not.

kitsunet:
@mhsdesign cli dispatch is a bit of a mess, and good point we should probably breaking fix it for this release. it has been this way ever since we split the two. It was one dispatcher at some point but that really didn't work well api wise, so I ripped it apart brutally with some dirt left behind....

kitsunet:
TODOS:

- [x] Stop and ForwardException constructors
  - https://github.com/neos/flow-development-collection/pull/3232#discussion_r1466770004
  - patch: https://github.com/kitsunet/flow-development-collection/pull/3
- ⏩ fix cli signal dispatch to action mvc mess -> https://github.com/neos/flow-development-collection/pull/3296
- ⏩ fix action request dispatch == true smell -> https://github.com/neos/flow-development-collection/pull/3301
- [ ] https://github.com/kitsunet/flow-development-collection/pull/4
- [ ] remove exceptions from controllerInterface signature? https://github.com/neos/flow-development-collection/pull/3232#issuecomment-1913304554
- [ ] upgrade guide: https://github.com/neos/flow-development-collection/pull/3232#issuecomment-1913166370

FOLLOWUPS:

- ⏩ Remove references to `UnsupportedRequestTypeException` -> https://github.com/neos/neos-development-collection/pull/4858

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
mhsdesign: LOL. It just hit me. The most obvious one to also solve my above comments... is just use a trait which will put in the functionality of mapping action name to method name :D that seems to be it :D horray?.

And then one implement the `ControllerInterface` _that_ is sexy?

kitsunet: that would work indeed, I guess @bwaidelich will hit us over the head about traits but I like it too 😂

Neos.Flow/Classes/Mvc/Exception/StopActionException.php
mhsdesign: see https://github.com/kitsunet/flow-development-collection/pull/3 :D

kitsunet:
> @kitsunet like this change still by reading but wonder wether the Fusion ControllerContext will still be able to manipulate the action response as before?
> 
> Side note: This is used in fusion form here https://github.com/neos/fusion-form/blob/main/Classes/Runtime/FusionObjects/RuntimeFormImplementation.php#L75-L89 and here https://github.com/neos/fusion-form/blob/main/Classes/Runtime/FusionObjects/RuntimeFormImplementation.php#L171-L175 mostly for setting location headers in redirect actions.

Mmm I see, yeah that won't work and we want to get rid of the controller context anyways. I'll create a little test setup and see what could be done.

After thought, it _should_ work if you set it up that way. Given that you don't have a controllercontext by default you would have to create one for the view anyways as it stands right now, and then you could just grab it from the view again at the end and get your information back. bit more pedestrian but I don't see why that wouldn't work.

Neos.Flow/Tests/Functional/Http/Fixtures/Controller/FooController.php
mhsdesign: i was dumb, we use `initializeController` 😂  sorry reverted.

Neos.Flow/Classes/Mvc/Exception/UnsupportedRequestTypeException.php
mhsdesign: But please in a separate PR 😅 otherwise this will become a slight big ball of diff stuff 😂

Neos.Flow/Classes/Mvc/Controller/ActionController.php
mhsdesign: Thats a good plan.

Im just looking where the variables are still used. And they leak also into other object. So a lot of things are stateful:

- $this->request
- $this->response
- $this->controllerContext
- $this->uriBuilder
- $this->arguments

with that many arguments, its just a question of time until one of our utility traits like `SpecialResponsesSupport` accesses - accidentally e.g. in a bugfix - any of them and were doomed.

Maybe we should take reinforcements like making the trait's methods static?

mhsdesign: The AbstractController uses this state in

- getControllerContext
- addFlashMessage (needs the request for the storage)
- forward as base action to merge the new params onto
- redirect, a resetted uribuilder and to set a default format based on the current request
- throwStatus, to with the current response to use the body if empty but overrule its status
- mapRequestArgumentsToControllerArguments sets the $this->arguments (can be solved easily because should only be called in `processRequest`


ActionController

- callActionMethod which doesnt make use of the passed parameters??
- handleTargetNotFoundError idk what this is
- addErrorFlashMessage uses the controller context
- forwardToReferringRequest uses the request->getReferringRequest
- getFlattenedValidationErrorMessage


And i also noticed that "replacing" the `$this->request` is probably not a good idea and should be disallowed via readonly? Mutating by object reference is the only way allowed.

mhsdesign: My experiment to make them final: https://github.com/neos/flow-development-collection/pull/3283/commits/72700b56d4671544dc1bc6858c35717b21374eec

There was only one place where we ourselves replace the variable for real, but i solved it with a `mergeIntoParent`.

mhsdesign: But as i feared that means that controllers cant be singletons anymore (why are they declared as such even in general???)

The only usecase of a controller being a singleton would be for forwarded requests, but that is hacky as hell to then rely on internal state :D

Aaaand it seems in widgets we rely on that the controller's process to be called multiple times.

https://github.com/neos/flow-development-collection/blob/465c80be0b6753356fe6e2c71178405970bf1fea/Neos.FluidAdaptor/Classes/Core/Widget/AbstractWidgetViewHelper.php#L241

mhsdesign: Would be cool to just put the request and reponse into the Controller via constructor and then process().

kitsunet: I would actually just like to not provide a response at all.... I don#t know any use case were we put anytihng but an empty response in there. Felt a bit too breaking to change it here though. we coud make a separate PR for that? 👿

kitsunet:
Ok so, I went back in time a bit and think we should just remove the whole "dispatched" thing on the request, I might just do it with this change and call this wohle thing a MVC overhaul or something.
Turns out we always set dispatched true in the MVC controller BUT we did use both states in the CLI controller, as both used the same dispatcher. Now that this is separate the property is completely useless IMHO.
Additionally in MVC we create new requests for forwards (which are basically the main reason for dispatch loop), so then the new request isn't dispatched automatically, while in CLI previously we would change the cli request for forwarding to another command. I think that is were it all comes from. So I lean heavily towards remove. I will have a go and see what the tests say.

mhsdesign:
Yes that sounds good thanks for the deep dive :D

This commit seems to confirm this: https://github.com/neos/flow-development-collection/commit/94dc7932f9c21851874a5ec1639ad929e11721a2#diff-1c9031af82104aa2f0a79a351adb135216b78da806f2f5b0fc7da5776eef34bcL227

The `ForwardException` was introduced wich now holds the `$nextRequest` and the dispatched(False) was never used in http mvc again. (Except maybe a little silently due to magic in `ActionRequest::__clone`).

I would have been interested in the _why_ but it totally makes sense for our case and it fits perfectly together :D

mhsdesign:
WIP TODO upgrade notes:

- MVC controll-flow-exceptions:
  - The `StopActionException` and `ForwardException` are both final and independant objects, so in the unlikely case that you catched the `StopActionException` you now need to catch both:
`} catch (StopActionException|ForwardException $exception) {`
  - The `StopActionException` must be constructed via `createForResponse` and the action response must be passed.
  - The `ForwardException` must be constructed via `createForNextRequest` and the next response must be passed.
- `Dispatcher` signals:
  - The parameter `$response` was removed from the `beforeControllerInvocation` signal, as it would be just always empty at that point and misleading and any mutations would be ignored.
  - The parameter `$response` was made nullable for the `afterControllerInvocation` signal, as it's not set after a `forwardToRequest`.
  - Also, it's to be noted, that modifying the response in an `afterControllerInvocation` is not always going to take effect, for example if the dispatcher is still in the loop.
- `Dispatcher::processRequest` changed signature (`return parent::processRequest($request);`)

kitsunet:
Aaaand another unraveling, we do not _really_ need Stop/ForwardActionException, we could just return.... For StopAction it's super easy, just return the (redirect/error/whatever) response. For ForwardAction something special is needed, it could stay an exception or we could attach the next request in some way to a response and just return that 🤔

mhsdesign:
Ha why does it always take time to fall into place :D I like this line of thought and that would reduce the api surface of the controller again.

And the signature would change to

```
public function processRequest(ActionRequest $request): ActionResponse|ForwardToNextActionRequest;
```

right? Wasnt that what basti suggested in:

https://neos-project.slack.com/archives/C04Q0TC15/p1703858685808609?thread_ts=1703852389.168549&cid=C04Q0TC15

----------

Additionally i did some small tweaking of things around:

https://github.com/kitsunet/flow-development-collection/pull/4

commit for commit is quite simple but it summs up to something, if you dont like that to go all in here we probably could split this up, but then again it feels nice to have all breaking notes on one pr?

mhsdesign:
Also i just stumbled upon the `RequestInterface`. It says "Contract for a dispatchable request." but due to the strict ActionRequest type in the dispatcher this is untrue now. Id say due to all the typings against ActionRequest the interface is completely useless and should be at least deprecated if not removed.


EDIT: Followup https://github.com/neos/flow-development-collection/pull/3295

kitsunet:
I have this whole removal of dispatched sitting around here and now wonder if I shoulnd't rather make this a separate PR, I think this PR will become a huge mess fast and no one wants to review huge messes 🙈

mhsdesign:
Hmm i was actually thinking of branching 9.0 off to `dispatcher-refactoring` :D but that wouldnt be easier to review as well.

I think multiple prs is the way to go and we could possibly create depending prs where we agree to merge them not directly into this but wait first for this one to be merged.

For that workflow it would be better if this pr was made from a neos branch and not on a fork ^^ so depending prs would reside here in neos...

EDIT:

As one cannot change the source of a pr, i created a temporary branch in neos `temporary-mvc-refactoring-target-branch` where all depending prs should be made. All followups and stuff. Once this pr is merged into 9.0, we can retarget all prs to 9.0 as well and delete the temporary branch.

mhsdesign:
Fyi this is my super small review pr https://github.com/kitsunet/flow-development-collection/pull/4 (one commit) that should ideally go directly into this pr as the part have already been touched.

Further adjustments to e.g. introduce trait instead of simple action controller: (comment https://github.com/neos/flow-development-collection/pull/3232#discussion_r1467385933)
Were extracted in an own pull request:

https://github.com/neos/flow-development-collection/pull/3297

I would like to propose to revert the introduction of the `SimpleActionController` so we can discuss in a new pr what to do without blocking this pr and without having anything undiscussed merged.

mhsdesign:
Aaaand im also not so happy with some of the naming choices in `SpecialResponsesSupport`. And would like to discuss this in a separate pr as well 🤔 (sorry :D)

Anything that is serious public api should deserve its own discussion and not be swallowed in a big change like this ...
If you like i could take your to separate this out into a new pr? (Feel free to force push it onto this pr https://github.com/neos/flow-development-collection/pull/3298)

I would really like to get the basis of this change in rather quickly, as i prepared already some followups ^^

Neos.Flow/Classes/Mvc/Controller/AbstractController.php
```diff
@@ -234,12 +243,8 @@ protected function forward($actionName, $controllerName = null, $packageKey = nu
      */
     protected function forwardToRequest(ActionRequest $request)
     {
-        $packageKey = $request->getControllerPackageKey();
-        $subpackageKey = $request->getControllerSubpackageKey();
-        if ($subpackageKey !== null) {
-            $packageKey .= '\\' . $subpackageKey;
-        }
-        $this->forward($request->getControllerActionName(), $request->getControllerName(), $packageKey, $request->getArguments());
+        $nextRequest = clone $request;
```
mhsdesign, line 245: we only need to clone as far as i know, because that will unset the dispatch state? But that will be hopefully be fixed via https://github.com/neos/flow-development-collection/pull/3301

Neos.Flow/Classes/Mvc/Controller/ActionController.php
mhsdesign: Discussion moved here https://github.com/neos/flow-development-collection/pull/3293

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
mhsdesign: The ActionController will be part of https://github.com/neos/flow-development-collection/pull/3297

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
mhsdesign: Implemented here https://github.com/neos/flow-development-collection/pull/3297

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
mhsdesign: Part of https://github.com/neos/flow-development-collection/pull/3301

Neos.Flow/Classes/Mvc/Controller/SimpleActionController.php
mhsdesign: Part of https://github.com/neos/flow-development-collection/pull/3297

mhsdesign:
I moved the simple controller over [here](https://github.com/neos/flow-development-collection/pull/3297) and the special response stuff [there](https://github.com/neos/flow-development-collection/pull/3298) 

After merging https://github.com/kitsunet/flow-development-collection/pull/4 into here and deciding if https://github.com/neos/flow-development-collection/pull/3294 should go in here or separate (`ActionResponse` vs `ResponseInterface`) i think this is mergeable. :D

mhsdesign:
We discussed that we like the idea to return psr responses instead of an `ActionResponse`

This PR will be merged into `feature/dispatcher-returns-psr-responses` from which we will open a blank new PR against 9.0.

This will be done to also merge https://github.com/neos/flow-development-collection/pull/3294 into that branch, and have one final place for last reviews.

This PR with 76 comments escalated a bit, but after we stripped out the controversial parts into own prs most of the discussion that took place here is not related anymore to the core changes right now.

PR #3233, opened 2023-11-12
BUGFIX: Make new object debug output more robust

kitsunet:
Unfortunately magic methods are tricky and __toString is no exception, a check if it's callable can result in true if the magic __call method is implemented but then the results of this call are completely undefined and therefore catching errors and continuing with other options is a good safeguard here.

Noticed this when I had an error in the `Mvc\Arguments` implementation which declares __call.

Followup to https://github.com/neos/flow-development-collection/pull/3211

### DIFF
--- a/Neos.Flow/Classes/Error/Debugger.php
+++ b/Neos.Flow/Classes/Error/Debugger.php
@@ -467,7 +467,10 @@ class Debugger
     protected static function getObjectSnippetPlaintext(object $object): string
     {
         if (is_callable([$object, '__toString'])) {
-            return self::getObjectShortName($object) . '|' . self::truncateObjectOutput((string)$object) . '|';
+            try {
+                $string = (string)$object;
+                return self::getObjectShortName($object) . '|' . self::truncateObjectOutput($string) . '|';
+            } catch (\Throwable $_) {/* This can happen if what was callable was not actually __toString (a magic __call() will do that) we can try other ways to get information. */}
         }
 
         if ($object instanceof \JsonSerializable) {
--- a/Neos.Flow/Classes/Error/Debugger.php
+++ b/Neos.Flow/Classes/Error/Debugger.php
@@ -466,7 +466,7 @@ class Debugger
 
     protected static function getObjectSnippetPlaintext(object $object): string
     {
-        if (method_exists([$object, '__toString'])) {
+        if (method_exists($object, '__toString')) {
             try {
                 $string = (string)$object;
                 return self::getObjectShortName($object) . '|' . self::truncateObjectOutput($string) . '|';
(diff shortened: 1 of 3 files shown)


mhsdesign:
Haha xD had this https://github.com/neos/flow-development-collection/pull/3233/commits/18c202c42bff07ec14fc2de52b3eb795f19046d2 around locally already and added it here :D

Neos.Flow/Classes/Error/Debugger.php
```diff
@@ -467,7 +467,10 @@ public static function getCodeSnippet(string $filePathAndName, int $lineNumber,
     protected static function getObjectSnippetPlaintext(object $object): string
     {
         if (is_callable([$object, '__toString'])) {
```
mhsdesign: maybe `method_exist` but yeah i get the point while rendering an exception trace there are more important manners than to find out that the __toString implementation of some stupid object doesnt work. lIke our eel context which hasnt worked for ages ...

kdambekalns (approved):

ahaeslich:
Are we keeping this as draft until we find all problems?

kitsunet:
Ups, no sorry

Neos.Flow/Classes/Error/Debugger.php
kitsunet: Ah mmm yes maybe we change that anyways...

kitsunet:
```suggestion
        if (method_exists([$object, '__toString'])) {
```

mhsdesign:
@kitsunet i hope its fine that i fixed the EEL Context within this pr?

kitsunet:
> @kitsunet i hope its fine that i fixed the EEL Context within this pr?

Extremely unrelated, why ^^ but shrug, fine for me

kitsunet:
Do I get another checkmark though?

Neos.Flow/Classes/Error/Debugger.php
```diff
@@ -466,8 +466,11 @@ public static function getCodeSnippet(string $filePathAndName, int $lineNumber,
 
     protected static function getObjectSnippetPlaintext(object $object): string
     {
-        if (is_callable([$object, '__toString'])) {
-            return self::getObjectShortName($object) . '|' . self::truncateObjectOutput((string)$object) . '|';
+        if (method_exists($object, '__toString')) {
+            try {
+                $string = (string)$object;
+                return self::getObjectShortName($object) . '|' . self::truncateObjectOutput($string) . '|';
+            } catch (\Throwable $_) {/* This can happen if what was callable was not actually __toString (a magic __call() will do that) we can try other ways to get information. */}
```
mhsdesign:
```suggestion
            } catch (\Throwable $_) {/* This can happen if what was callable was not actually __toString (a magic __call() will do that) we can try other ways to get information. */
            }
```

mhsdesign (approved):

PR #3239, opened 2023-11-18
!!! FEATURE: Extract `LowLevelFrontendInterface` and implement PSR-6 and PSR-16 as cache frontends to be used via `Caches.yaml`

mficzel:
While Neos had methods to create PSR caches those were seldomly used as they worked quite differently to other caches that were configured via `Caches.yaml`.

This change introduces implementations of PSR-6 Cache and PSR-16 SimpleCache that can be used as normal flow cache-frontends. To make this possible the low level functionality of object instantiation and garbage collection and flushing was extracted into a `LowLevelFrontendInterface` while the `FrontenInterface` on to defines the interaction with the caches.

```yaml
Vendor_Package_SimpleCache:
  frontend: Neos\Cache\Frontend\Psr\SimpleCache\SimpleCacheFrontend
  backend: Neos\Cache\Backend\FileBackend

Vendor_Package_Cache:
  frontend: Neos\Cache\Frontend\Psr\Cache\CacheItemPoolFrontend
  backend: Neos\Cache\Backend\FileBackend

```

Tho help to ensuring which cache interface you are working with the method `getCache` of the `CacheManager` now got an optional second argument that allows to specify the classname of the expected frontend implementation.

The old methods to create PSR caches are deprecated and marked for removal with Neos Flow 10.

**Upgrade instructions**

If Caches are configured via Caches.yaml and injected via Objects.yaml no changes are needed. 

If the CacheManager is used to obtain certain caches it is important to note that the the method `getCache` now returns `LowLevelFrontendInterface` instead of `FrontendInterface`. It will still return the configured cache implementation from `Caches.yaml` but you may have to add the expected className as second argument to convince your static analysis.

**Review instructions**

A secondary non-official goal of this pr is allowing to implement other cache interfaces like the following as flow cache and be managed by the cache manager.

```php
    public function resolve(string $identifier, \Closure $updateClosure, array $tags = [], ?int $lifeTime = null, ?int $gracePeriod = null, ?int $lockPeriod = null): mixed
```

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Frontend/AbstractFrontend.php
+++ b/Neos.Cache/Classes/Frontend/AbstractFrontend.php
@@ -13,81 +13,14 @@ namespace Neos\Cache\Frontend;
  * source code.
  */
 
-use Neos\Cache\Backend\BackendInterface;
-use Neos\Cache\Backend\TaggableBackendInterface;
 
 /**
  * An abstract cache
  *
  * @api
  */
-abstract class AbstractFrontend implements FrontendInterface
+abstract class AbstractFrontend extends AbstractLowLevelFrontend implements FrontendInterface
 {
-    /**
-     * Identifies this cache
-     * @var string
-     */
-    protected $identifier;
-
-    /**
-     * @var BackendInterface
-     */
-    protected $backend;
-
-    /**
-     * Constructs the cache
-     *
-     * @param string $identifier A identifier which describes this cache
-     * @param BackendInterface $backend Backend to be used for this cache
-     * @throws \InvalidArgumentException if the identifier doesn't match PATTERN_ENTRYIDENTIFIER
-     */
-    public function __construct(string $identifier, BackendInterface $backend)
-    {
-        if (preg_match(self::PATTERN_ENTRYIDENTIFIER, $identifier) !== 1) {
-            throw new \InvalidArgumentException('"' . $identifier . '" is not a valid cache identifier.', 1203584729);
-        }
-        $this->identifier = $identifier;
-        $this->backend = $backend;
-    }
-
-    /**
-     * Initializes this frontend
-     *
-     * The backend is connected with this frontend in initializeObject(), not in __construct(), because the Cache Factory
-     * needs an opportunity to register the cache before the backend's setCache() method is called. See
-     * CacheFactory::create() for details. A backend (for example the SimpleFileBackend) may rely on the cache already
-     * being registered at the CacheManager when its setCache() method is called.
-     *
-     * @return void
-     */
-    public function initializeObject()
-    {
-        // FIXME: This can be removed in next major, since the Backend gets the Frontend set in the cache factory now.
-        $this->backend->setCache($this);
-    }
-
-    /**
-     * Returns this cache's identifier
-     *
-     * @return string The identifier for this cache
-     * @api
-     */
-    public function getIdentifier(): string
-    {
-        return $this->identifier;
-    }
-
-    /**
-     * Returns the backend used by this cache
-     *
-     * @return BackendInterface The backend used by this cache
-     * @api
-     */
-    public function getBackend(): BackendInterface
-    {
-        return $this->backend;
-    }
-
     /**
      * Checks if a cache entry with the specified identifier exists.
      *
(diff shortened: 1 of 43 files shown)


kitsunet:
Beautiful thanks for preparing this!

mficzel:
@kitsunet @kdambekalns after stumbling over the limitations of our cache interface again i dug out the old idea of having a more basic cache interface we recntly discussed. 

Alongside making the use of PSR Caches as easy as any other cache this would allow us to implement caches like https://github.com/sitegeist/Sitegeist.StaleMate directly as a cache frontend which i consider quite valuable.

Neos.Flow/Classes/Cache/CacheManager.php
```diff
@@ -177,19 +178,24 @@ public function registerCache(FrontendInterface $cache, bool $persistent = false
     /**
      * Returns the cache specified by $identifier
      *
+     * @template T of object
      * @param string $identifier Identifies which cache to return
-     * @return FrontendInterface The specified cache frontend
+     * @param class-string<T>|null $className The expected CacheManager for the cache frontend
+     * @return (T is null ? LowLevelFrontendInterface : T && LowLevelFrontendInterface) The specified cache frontend
```
mficzel: I still struggle a bit with the syntax for return types. PHPStan seems to understand that but it may be a semi-geniuis  idea alltogether.

mhsdesign: Hmm i see your point. Maybe its better to delay this discussion in a separate pr as this is a general problem we face in many places see FusionRuntime but also the ValueAccessor and we should discuss our philosophy once in an issue and draw out a path for all parts that would profit from this solution. Otherwise i fear this discussion might take over in this pr and scatter our focus from the important parts ;)

mficzel: @mhsdesign done as suggested.

Neos.Cache/Classes/Frontend/AbstractFrontend.php
```diff
@@ -13,81 +13,14 @@
...
-use Neos\Cache\Backend\BackendInterface;
-use Neos\Cache\Backend\TaggableBackendInterface;
 
 /**
  * An abstract cache
  *
  * @api
  */
-abstract class AbstractFrontend implements FrontendInterface
+abstract class AbstractFrontend extends AbstractLowLevelFrontend implements FrontendInterface
```
kitsunet, line 22: Love it overall, just wonder if we should go for composition instead of inheritance? Should be feasible I assume...

kitsunet (commented):
Thanks for taking care, would love to get this into 9.0, as written, I think composing our Frontend with a lowlevel one inside might be nicer in the long run :)

But I probably wouldn't block it if that takes too long.

Neos.Cache/Classes/Frontend/AbstractFrontend.php
mficzel: Thought about that but did not have a really bright idea what building blocks one would want for that.

kitsunet: Right, I will install the change and have a look at a suggestion :)

mficzel: Since keeping the abstract Frontend in place is a good ides to limit breakiness anyways i did not give it too too much effort either.

kitsunet: Closing this, I misread this LOL, I also see no big point to compose here :D

kitsunet:
Had a look as promised, my suggestion about composition was 🤦 I just misread what was going on. I don#t think it's necessary, but overall I find the code a bit convoluted now, I wrote up a whole lot here about conceptually splitting the interfaces but however I turn it, there is no beautiful solution with different needs pulling and pushing at this code. Maybe the LowLevelInterface could have a different name ala `CacheMaintenanceInterface` to put it apart from the FrontendInterface, but that's the best I can come up with...

kitsunet (approved):
Giving this a preliminary +1 even though comment above applies.

mficzel:
@kitsunet for the psr-frontends it would work to have them as a wrapper around the current frontends. For flexibility i would like to be able to implement frontends with a totally different interface like the closures the symfony cache uses. 

For that (+ async cache population)  i tried the wrapper approach here https://github.com/sitegeist/Sitegeist.StaleMate but this is even more convoluted. That is why i ended here. 

I did not know how to describe this properly in the pr description. Will add another line to the review instructions.

mhsdesign:
If you like I would also try to look in depth into this and give you a full review. But that may take some time and I might have some questions :P (i don't know what to make out of the lowlevel naming). Maybe we can discuss this at the sprint or next Friday?
I hope you're okay with delaying it a bit longer;)

PR #3240, opened 2023-11-20
BUGFIX: No useless read in `FileBackend` (improve performance)

fdsis:
`FileBackend::findIdentifiersByTags` now early returns if `$tags` is empty. Otherwise it would read every cache entry completely unnecessarily from the filesystem. 

<details>
  <summary>Profile before</summary>

Screenshot_1

</details>

<details>
  <summary>Profile after</summary>

Screenshot_2

Screenshot_3

</details>

**Checklist**

- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html) |  7.3 does not have the method in `FileBackend` so using 8.0 instead

### DIFF
--- a/Neos.Cache/Classes/Backend/FileBackend.php
+++ b/Neos.Cache/Classes/Backend/FileBackend.php
@@ -245,11 +245,15 @@ class FileBackend extends SimpleFileBackend implements PhpCapableBackendInterfac
      * specified tags.
      *
      * @param string[] $tags The tags to search for
-     * @return string[] An array with identifiers of all matching entries. An empty array if no entries matched
+     * @return string[] An array with identifiers of all matching entries. An empty array if no entries matched or no tags were provided
      * @api
      */
     public function findIdentifiersByTags(array $tags): array
     {
+        if (empty($tags)) {
+            return [];
+        }
+
         $entryIdentifiers = [];
         $now = $_SERVER['REQUEST_TIME'];
         $cacheEntryFileExtensionLength = strlen($this->cacheEntryFileExtension);


kitsunet (changes requested):
Good catch thank you for this! 
Although our currently supported version is 7.3, so it should either go in there or in 9.0 as next upcoming version depending if it's considered a bugfix or feature. I will sleep on it. Currently I would say it is a feature, BUT I also see no drawback as it's absolutely clear that there is no result with no tags, therefore I could live with it going into previous versions. 7.3 then though.

Sebobo:
As this type of cache regularly causes issues for projects, any improvement should be considered a bugfix if the behavior doesn't change IMO. So 7.3 should be the target.

mhsdesign (approved):
Thanks for your first pr ;)It seems the target is already correctly chosen, as this was introduced with 8.0: https://github.com/neos/flow-development-collection/commit/590a968adb48750c8399a00d79a19d88ac2f6df8

And i very much consider this a safe (performance) bugfix ;)

Neos.Cache/Classes/Backend/FileBackend.php
```diff
@@ -245,11 +245,15 @@ public function findIdentifiersByTag(string $searchedTag): array
      * specified tags.
      *
      * @param string[] $tags The tags to search for
-     * @return string[] An array with identifiers of all matching entries. An empty array if no entries matched
+     * @return string[] An array with identifiers of all matching entries. An empty array if no entries matched or no tags were provided
      * @api
      */
     public function findIdentifiersByTags(array $tags): array
     {
+        if (empty($tags)) {
```
kdambekalns, line 253:
```suggestion
        if ($tags === []) {
```

kdambekalns (approved):
A "proper comparison" beats `empty(…)`, but other than that: 👍

Thanks for your first contribution!

Rgearding bug or feature: I edited the title, as I clearly this as a bugfix. No input means no output here, so…

kdambekalns:
@fdsis If you'd like to practice your test-writing skills, you could add a unit test to verify the new behavior. 😉

kitsunet (approved):

Neos.Cache/Classes/Backend/FileBackend.php
mhsdesign: hmm true but in this case its actually fully equivalent ^^ so im fine with merging it either way.

mhsdesign:
I think writing tests is not that easy as you need to mock internals and assert them to be not be called. But if the implementation changes any time the test will become worthless as it asserts that something that is maybe not used anymore will not be used and thus the test will give a false positive and pass. (maybe that makes sense ^^)

kitsunet:
Note to us, we _could_ apply a similar check to findIdentifiersByTag() in 7.3 as that _should_ never have a result if the tag is an empty string.

Neos.Cache/Classes/Backend/FileBackend.php
fdsis: I chose `empty` solely for readability [because `empty($tags)` gets parsed as `!$tags` anyway](https://github.com/php/php-src/blob/98b207ed8aa815afbda777dd338ff1847c7e9cff/Zend/zend_compile.c#L9587). So performance wise it should make no difference

kitsunet: I think there is always leeway for some personal preference, I also like to use empty because it's just an easy to parse "keyword" in a condition.

mhsdesign: Linking to the actual line in `php-src` 😂 pro gamer move 💯

kdambekalns:
> I think writing tests is not that easy as you need to mock internals 

Just make sure `readCacheFile()` is never called when `$tags` is an empty array.

> the test will become worthless

Like it is now this fixes a "built-in brake" without any assertion this won't come back. I have seen those things resurface, you know… 🤷

PR #3244, opened 2023-11-29
FEATURE: Add `Flow\InjectCache` Attribute / Annotation for property injection

mficzel:
In many cases an `Objects.yaml` is created just to inject caches which can feel a bit cumbersome as one already had specified the cache in `Caches.yaml`.

To address this the new `@Flow\InjectCache` annotation allows to assign a cache frontend of a configured cache directly to a property without having to configure the `Objects.yaml` at all.

```php
    #[Flow\InjectCache(identifier: 'Flow_Mvc_Routing_Resolve')]
    protected VariableFrontend $cache;
```

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
@@ -12,6 +12,7 @@ namespace Neos\Flow\ObjectManagement\DependencyInjection;
  */
 
 use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Cache\CacheManager;
 use Neos\Flow\Configuration\ConfigurationManager;
 use Neos\Flow\Configuration\Exception\InvalidConfigurationTypeException;
 use Neos\Flow\Log\Utility\LogEnvironment;
@@ -45,6 +46,7 @@ class ProxyClassBuilder
     protected Compiler $compiler;
     protected LoggerInterface $logger;
     protected ConfigurationManager $configurationManager;
+    protected CacheManager $cacheManager;
     protected CompileTimeObjectManager $objectManager;
 
     /**
@@ -65,6 +67,11 @@ class ProxyClassBuilder
     public function injectConfigurationManager(ConfigurationManager $configurationManager): void
     {
         $this->configurationManager = $configurationManager;
+    }
+
+    public function injectCacheManager(CacheManager $cacheManager): void
+    {
+        $this->cacheManager = $cacheManager;
     }
 
     #[Flow\Autowiring(false)]
@@ -395,6 +402,13 @@ class ProxyClassBuilder
                     }
                     $commands = array_merge($commands, $this->buildPropertyInjectionCodeByConfigurationTypeAndPath($objectConfiguration, $propertyName, $configurationType, $propertyValue['path']));
                     break;
+                case ConfigurationProperty::PROPERTY_TYPES_CACHE:
+                    $cacheIdentifier = $propertyValue['identifier'];
+                    if (!array_key_exists($cacheIdentifier, $this->cacheManager->getCacheConfigurations())) {
+                        throw new UnknownObjectException('The cache injection specified for property "' . $propertyName . '" in the object configuration of object "' . $objectConfiguration->getObjectName() . '" refers to the unknown cache identifier "' . $cacheIdentifier . '".', 1701280300);
+                    }
+                    $commands = array_merge($commands, $this->buildPropertyInjectionCodeByCacheIdentifier($objectConfiguration, $propertyName, $cacheIdentifier));
+                    break;
             }
             $injectedProperties[] = $propertyName;
         }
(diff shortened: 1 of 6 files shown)


Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -513,6 +527,25 @@ public function buildPropertyInjectionCodeByConfigurationTypeAndPath(Configurati
...
+     *
+     * @param Configuration $objectConfiguration Configuration of the object to inject into
+     * @param string $propertyName Name of the property to inject
+     * @param string $cacheIdentiier the configuration type of the injected property (one of the ConfigurationManager::CONFIGURATION_TYPE_* constants)
+     * @return array PHP code
+     */
+    public function buildPropertyInjectionCodeByCacheIdentifier(Configuration $objectConfiguration, string $propertyName, string $cacheIdentiier): array
+    {
+        $className = $objectConfiguration->getClassName();
+        $preparedSetterArgument = '\Neos\Flow\Core\Bootstrap::$staticObjectManager->get(\Neos\Flow\Cache\CacheManager::class)->getCache(\'' . $cacheIdentiier . '\')';
```
mhsdesign:
```suggestion
     * @param string $cacheIdentifier the configuration type of the injected property (one of the ConfigurationManager::CONFIGURATION_TYPE_* constants)
     * @return array PHP code
     */
    public function buildPropertyInjectionCodeByCacheIdentifier(Configuration $objectConfiguration, string $propertyName, string $cacheIdentifier): array
    {
        $className = $objectConfiguration->getClassName();
        $preparedSetterArgument = '\Neos\Flow\Core\Bootstrap::$staticObjectManager->get(\Neos\Flow\Cache\CacheManager::class)->getCache(\'' . $cacheIdentifier . '\')';
```

and i also was wondering if `"'" . $bla . "'"` is good enough as we never validate that a cache identifier might not contain a `'` itself - but would work 99% of the time so fine ^^

mficzel: Regarding the "'" we do pretty much the same for settings nowadays. The current solution would be to make the cache identifier a valueObject but this is beyond the scope here. 

The suggested changes are applied

Sebobo (changes requested):
Looks great by reading, thx!

Can you add the annotation also to the `AnnotationReference.rst`.

mficzel:
@Sebobo i assume this is autogenerated "This reference was automatically generated from code on 2023-11-22" so this should appear automagically.

Sebobo (approved):

Sebobo:
Ah ok then nvm :)

kdambekalns:
I might be trying to open a certain well-known box here, but what about targeting 8.3? 😬

mficzel:
@kdambekalns that is a clear feature so would at max be for an 8.4. Not a fan of that either as 9.0 looks really promising already.

kdambekalns:
> Not a fan of that either as 9.0 looks really promising already.

**Flow** 9.0 is fine, yeah. 🤪 Fine with me either way, it's a nice feature.

kdambekalns (approved):
👍 by 👀

kitsunet:
SORRY super late comment but I have been mulling this in my head for a while until I came up with a good description/thought...

We already have the virtual object feature Bastian created years ago, this works well but means you need to configure a virtual object for your cache in Objects.yaml before you can inject it. What IF we were to provide virtual object configurations for each available cache automatically, looks a bit ugly but ala `Neos.Flow:Cache:Neos_Fusion_Content`  given that the configuration is formulaic this should be rather simple and make this work with "regular" inject annotations/attributes.

mficzel:
@kitsunet yeah that would be possible but would also be imho be a bit weird as it requires an identifier that is not the the same as the one you configured in caches yaml.

Not sure i like this better than the current approach here since `#[Flow\InjectCache]` is pretty much in line to what we do with `#[Flow\InjectConfiguration]`.

kitsunet:
Yeah just an idea, I mostly see the issue that we start "collecting" Inject* things, Logs would be another obvious candidate. Configuration is a bit special IMHO because it returns this array. If we _could_ allow something like `Inject("Neos.Flow:Setting:My.Package.Setting")` I would kinda be in favor to just distill it down to Inject, but I think wildcards won't work easily with the virtual object syntax.

mficzel:
@kitsunet is this a merely academic discussion? If you do not like this at all please add a "request change" review as this can be merged with the current reqviews.

Regarding logs the simple case ther is to inject a PSR Logger (resolves to system logger) wich works fine 99% of the time. Since caches are always custom configured this will not work here. 

Offcourse it would be possible to create a default configuration for a "VariableFrontend" and "StringFrontend" to work with plain `@Flow\Inject` but that would be SharedCaches with tons of identifier collisions so a really bad idea anyways.

BTW: Did not merge this yet (even before the latest discussion) to be able to change target to a possible 8.4 branch if we decide to do such a thing.

kitsunet:
I am sorry if it sounds academic, I certainly won't block it when we have a decision about the branch.

I think features like this are part of a future "direction" of how we want to develop this framework. Once this feature is in circulation we will have a hard time walking back on it. So I think it's important to consider the future we plan for. 

You misunderstand my idea for the Caches with inject, I agree a generic cache like the logs would be bad. But that virtual object idea could work fine. I slightly prefer that direction as it seems like introducing less Flow specific "magic" stuff (which is definitely arguable as both are magic in their own way). Ultimately I don't know, it just seems to me that we need to consider the Framework direction for this change in addition to the general feature which I think is a great idea in general.

KarstenDambekalns-eaton:
@kitsunet has a point here. I just started using virtual objects for logger injection (🙈) and like the way it works. So that might indeed be a "generic" option we already have?

mficzel:
I personally would like to add more php-attributes to flow in future since objects yaml is good for configuring external packages but mostly weird in project code. Even i always have to look up the syntax there.

I think we already agreed that we want `Flow\Route` and `Flow\Policy` annotations eventually to avoid Routes.yaml and Policy.yaml in simple cases (project code). Even an `injectLogger("customLogger")` attribute seems not absurd to me as it would be a nice incentive to not dump everything into the system logger.

Also any magic solution to generate virtual classes for caches on the fly would have to be implemented while the attributes pr is already there and ready to merge.

kitsunet (approved):

kitsunet:
Valid arguments thank you :)

kitsunet:
Now I was a bit quick as I wanted to show my agreement, I guess IF we do 8.4 we can always backport this. I don't see it atm though even if I still like the idea, but there seems to be not much enthusiasm about it.

PR #3250, opened 2023-12-05
BUGFIX: Assume content exists, if stream size is unknown

kdambekalns:
If a PSR7 stream is returned from an `ActionController` action, no content arrives at the client, if the stream has an unknown size.

Why is that? Because the check in our `ActionResponse` in `hasContent()` is implemented like this: 

    $this->content->getSize() > 0

If a stream returns `null` here, because the size is unknown, we should assume content exists...

There should be no change in behavior, even if the stream is in fact empty. Because that would lead to `hasContent()` returning `true`, and the HTTP response being built in `ActionResponse` would get the stream as content. When being delivered that would evaluate to "a stream from an empty string", so there will be (again) no difference, if you look at what the `MessageTrait`does if the internal stream is `null`:

```php
    public function getBody(): StreamInterface
    {
        if (!$this->stream) {
            $this->stream = Utils::streamFor('');
        }

        return $this->stream;
    }
```

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
+++ b/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php
@@ -228,7 +228,7 @@ abstract class ResponseInformationHelper
             }
         }
 
-        if (!$response->hasHeader('Content-Length')) {
+        if (!$response->hasHeader('Content-Length') && $response->getBody()->getSize() !== null) {
             $response = $response->withHeader('Content-Length', $response->getBody()->getSize());
         }
 
--- a/Neos.Flow/Classes/Mvc/ActionResponse.php
+++ b/Neos.Flow/Classes/Mvc/ActionResponse.php
@@ -345,6 +345,7 @@ final class ActionResponse
      */
     private function hasContent(): bool
     {
-        return $this->content->getSize() > 0;
+        $contentSize = $this->content->getSize();
+        return $contentSize === null || $contentSize > 0;
     }
 }


kdambekalns:
The `StreamInterface` says:

```php
    /**
     * Get the size of the stream if known.
     *
     * @return int|null Returns the size in bytes if known, or null if unknown.
     */
    public function getSize();
```

If someone has the feeling this might be a breaking change, let me know. I'd be interested!

kitsunet (approved):
I'd say that makes sense, did you try with an empty response stream? as in empty string?

kdambekalns:
> did you try with an empty response stream? as in empty string?

No. That would behave as before, the size is known and is 0, so…

Or do you refer to a stream without content that has an unkown size… hm. That would lead to `hasContent()` returning `true`, and the HTTP response being built in `ActionResponse` would get the stream as content. When being delivered that would evaluate to "a stream from an empty string", so there will be (again) no difference, if you look at what the `MessageTrait`does if the internal stream is `null`:

```php
    public function getBody(): StreamInterface
    {
        if (!$this->stream) {
            $this->stream = Utils::streamFor('');
        }

        return $this->stream;
    }
```

kitsunet:
Fair point, was probably too late to reason about it. What about https://github.com/neos/flow-development-collection/blob/401c861d354aa2b4d25ba5797d28e510104d3942/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php#L232 (called via StandardsComplianceMiddleware at the end), that might need adaption to not write content length if it's a stream?

kdambekalns:
> What about
> 
> https://github.com/neos/flow-development-collection/blob/401c861d354aa2b4d25ba5797d28e510104d3942/Neos.Flow/Classes/Http/Helper/ResponseInformationHelper.php#L232
> (called via StandardsComplianceMiddleware at the end), that might need adaption to not write content length if it's a stream?

In my actual usecase that works fine, i.e. there is no Content-Length header being sent when the stream size is unknown. That `null` -valued header is obviously filtered out at some point.

kdambekalns:
> In my actual usecase that works fine

… but why not add a check. ✅

kitsunet (approved):

robertlemke (approved):
LGTM

PR #3252, opened 2023-12-20
TASK: Correct Flow composer.json

kitsunet:
The upmerge commit 42e3fd7886d5bed317511a2046d4119867216923 wrongly overwrote major parts of Flows composer.json, introducing older versions of PHP and psr/log as well as removing dependencies on some other psr packages. This change corrects the issue and needs to be upmerged accordingly.

The changes were never merged into the collection composer.json so that the issue was not noticed in development environments.

### DIFF
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -6,7 +6,7 @@
     "license": ["MIT"],
 
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^8.0",
 
         "ext-zlib": "*",
         "ext-SPL": "*",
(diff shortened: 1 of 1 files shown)


kitsunet:
Will merge without votes in 24h latest.

kitsunet:
Note if someone is following my footsteps, the update to psr/log came with 
https://github.com/neos/flow-development-collection/pull/2792

and if you look at the commit it is supposedly gone into 7.3.8 which would suggest that something was already wrong in the 7.3 branch, yet that is not true. 7.3.8 was a broken release it included stuff from the 8.0 branch. This was corrected in 7.3.9.
https://github.com/neos/flow-development-collection/blob/7.3.8/composer.json

Therefore this should be the origin of the misaligned versions and the place to fix it.

markusguenther (approved):
Sorry :/

dlubitz:
I was a bit confused, why the PR is against 8.0 instead 7.3. But Ok, in 7.3 we already fixed it, but never fixed it properly in 8.0.
See: 
* https://packagist.org/packages/neos/flow#8.0.7
* https://packagist.org/packages/neos/flow#8.0.8

So looks good to me!

dlubitz (approved):

ahaeslich:
thx @kitsunet for taking care 💖

PR #3254, opened 2023-12-23
TASK: Refactor and optimize of session data storage

### LINKED ISSUE #525: Optimize session cache data writes
Jira issue originally created by user @hlubek:

The session cache backend currently writes data back for every _access_ (including read-only requests) at the end of each request.

For Neos this involves the following keys (individual cache entries) for **each request** (also in the frontend), while having an active session:
- lastVisitedNode
- TYPO3_Flow_Security_Accounts
- TYPO3_Flow_Object_ObjectManager

We could greatly optimize I/O and cache file contention by checking on a cache frontend level if the data was actually changed. The idea is to store md5 sums of the data that was fetched (`get`) and comparing that on subsequent `set` operations and doing nothing if they are equal. Doing that on the frontend layer would not require an additional serialize call.

 Jira-URL: https://jira.neos.io/browse/FLOW-147

mficzel:
Before the SessionMetadata and SessionData was written with every request which caused network traffic, storage wear and also made race conditions much more likely when parallel requests changed session data. 

In total this can reduce the number of write operations on the Session caches by 80-90% removing storage and network load as those caches always are persistent and shared across clusters.

1. The improvements are on top of the Neos 9 already reducing the Flow_Session_Storage write load by not always storing the "lastVisitedNode" in the session.
2. The improvements mostly occur after sending the result "onShutdown" so this will not improve single requests but overall performance and number of parallel requests.

### Problem 

The session metadata (SessionId, StorageId, LastActivity .. ) is usually written at shutdown of every single request to the session metadata cache even when nothing changed. 

### Optimization 

This is optimized by the new setting `Neos.Flow.session.updateMetadataThreshold` that allows to configure the interval for updating the session metadata when nothing but the `lastActivityTimestamp` has changed. This removes lots of cache writes and avoids network traffic or storage wear. The session metadata is also converted to a single value object that combines SessionID, StorageID and lastActivityTimestamp.

### Problem

Session data is written to the Flow_Session_Storage cache once Session->putData is called. In case of flow this mostly is the Objects of scope `@Flow\Scope("session")` that are stored on shutdown. Those objects are sometimes modified but during most requests nothing changes here and at the end of the request the same data is added to the cache again with another redundant cache write. 

### Optimization

The SessionDataStore optimizes this be keeping a hash of the value all previously read keys and avoids writes if the serialized content that is stored yields the same hash. That way only once session-data was actually changed the session objects are actually written to the cache. This also lessens the probability of some race conditions drastically that can occur when multiple parallel requests work on the same session.

The following redundant behavior was also removed:
- All session metadata records in the cache were previously tagged with `session` for iterating over them again. This is replaced by the `retrieveAll` method.
- The current authentication providers were always stored in the session data as `Neos_Flow_Security_Accounts ` but were unused  
- Do we want to release this as 8.4 or 9.0? In case of 9.0 the SessionMetaData ValueObject will be adjusted to php 8.1 style.

Resolves: https://github.com/neos/flow-development-collection/issues/525

**Upgrade instructions**

**Review instructions**

Some questions during the review could be:
- Are there better ways to determine written objects are modified than comparing hashes of serialized values?: I did not find one.
- Should the comparison of written data with hashes of existing data be implemented in the cache frontend instead?: I think this would consume to much memory we have lots of cache items.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Session/Data/SessionIdentifier.php
+++ b/Neos.Flow/Classes/Session/Data/SessionIdentifier.php
@@ -17,6 +17,10 @@ use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Utility\Algorithms;
 
 /**
+ * The session identifier that in a string form is written in the session cookie
+ * and identifies a SessionMetadata object from the SessionMetaDataStore that represents
+ * a stored session without the key value store.
+ *
  * @Flow\Proxy(false)
  * @internal
  */
--- a/Neos.Flow/Classes/Session/Data/SessionIdentifier.php
+++ b/Neos.Flow/Classes/Session/Data/SessionIdentifier.php
@@ -23,7 +23,7 @@ use Neos\Flow\Utility\Algorithms;
 class SessionIdentifier
 {
     private function __construct(
-        public /*readonly*/ $value
+        public readonly string $value
     ) {
     }
 
--- a/Neos.Flow/Classes/Session/Data/SessionIdentifier.php
+++ b/Neos.Flow/Classes/Session/Data/SessionIdentifier.php
@@ -1,0 +1,44 @@
+<?php
+declare(strict_types=1);
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+namespace Neos\Flow\Session\Data;
+
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Utility\Algorithms;
+
+/**
+ * @Flow\Proxy(false)
+ * @internal
+ */
+class SessionIdentifier
+{
+    private function __construct(
+        public /*readonly*/ $value
+    ) {
+    }
+
+    public static function createFromString(string $value): self
+    {
+        return new self($value);
+    }
+
+    public static function createRandom(): self
+    {
+        return new self(Algorithms::generateRandomString(32));
+    }
+
+    public function equals(SessionIdentifier $other): bool
+    {
+        return $this->value === $other->value;
+    }
+}
--- a/Neos.Flow/Classes/Session/Data/SessionMetaData.php
+++ b/Neos.Flow/Classes/Session/Data/SessionMetaData.php
@@ -1,0 +1,78 @@
+<?php
+declare(strict_types=1);
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+namespace Neos\Flow\Session\Data;
+
+use Neos\Flow\Annotations as Flow;
+
+/**
+ * @Flow\Proxy(false)
+ */
+class SessionMetaData
+{
+    protected int $lastActivityTimestamp;
+    protected string $storageIdentifier;
+
+    /**
+     * @var string[]
+     */
+    protected array $tags;
+
+    /**
+     * @param string $storageIdentifier
+     * @param int $lastActivityTimestamp
+     * @param string[] $tags
+     */
+    public function __construct(string $storageIdentifier, int $lastActivityTimestamp, array $tags)
+    {
+        $this->lastActivityTimestamp = $lastActivityTimestamp;
+        $this->storageIdentifier = $storageIdentifier;
+        $this->tags = $tags;
+    }
+
+    public static function fromClassicArrayFormat(array $data): self
+    {
+        return new self(
+            $data['storageIdentifier'],
+            $data['lastActivityTimestamp'],
+            $data['tags']
+        );
+    }
+
+    public function withLastActivityTimestamp(int $lastActivityTimestamp): self
+    {
+        return new self(
+            $this->storageIdentifier,
+            $lastActivityTimestamp,
+            $this->tags
+        );
+    }
+
+    public function getLastActivityTimestamp(): int
+    {
+        return $this->lastActivityTimestamp;
+    }
+
+    public function getStorageIdentifier(): string
+    {
+        return $this->storageIdentifier;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getTags(): array
+    {
+        return $this->tags;
+    }
+}
--- a/Neos.Flow/Classes/Session/Data/SessionMetaData.php
+++ b/Neos.Flow/Classes/Session/Data/SessionMetaData.php
@@ -14,14 +14,18 @@ declare(strict_types=1);
 namespace Neos\Flow\Session\Data;
 
 use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Utility\Algorithms;
 
 /**
  * @Flow\Proxy(false)
  */
 class SessionMetaData
 {
-    protected int $lastActivityTimestamp;
+    protected string $sessionIdentifier;
+
     protected string $storageIdentifier;
+
+    protected int $lastActivityTimestamp;
 
     /**
      * @var string[]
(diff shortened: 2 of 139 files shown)


kitsunet:
Great idea <3

mficzel:
Keep this as draft for now until we decide wether this should go to 9.0 or a hypothetical 8.4. In case of 9.0 would refactor the SessionMetaData Value object to PHP 8.1 style.

fcool:
Yay! The idea is great - but I fear this won't work with Entites used in sessions - as we reverted my change, that serializiation becomes repeatable.

So if one uses serialize on an object used in session, the session will again "always be modified".

But before seeing only the troubles... I'll try to prove my assumption based on this PR.

The general idea and implementation are looking VEEERY neat

mficzel:
@fcool as i understand you are referring to #2009 .. while this is valid you would have to either call serialize twice for some reason or have this as an argument and async jobqueue call somewhere. Sounds like an edge case to me, do i understand that correct?

Sidenote: I personally would avoid entities in methods calls that are asyncronously handled by the jobqueue and only use identifiers as arguments there.

fcool:
Exactly. 
But no, "one" would only need to call it once. The session would be "twice". 

So if you put your objects, which are stored in session for whatever reason in a cache or use the serialize call for other reasons, the session - the "second" call, will not get the expected result.

And its no "edge case". Neos itself does it (storing the Assets to display or already displayed in session for example)

mficzel:
Will try to do some benchmark to see wether this prevents this from beeing effective at all or wether it only is not always effective.  In general it should still work as before but may not be able to optimize in certain cases.

mficzel:
Just took some totally non scientific numbers during a Neos session:  Login to Neos, change a text on the homepage and another page, open 4 other documents and log out again.

- Before: 
  - Flow_Session_MetaData: get 11 , set 33
  - Flow_Session_Storage: get 26, set 23
- After: 
  - Flow_Session_MetaData: get 10 , set 3
  - Flow_Session_Storage: get 20, set 13

That looks like at least 50% less many writes to the Flow_Session_Storage and 90% less to Flow_Session_MetaData.  And that with backend login and logout which has quite some legitimate stuff to write. I would expect the numbers to be even higher outside of the Neos Backend as this stores the last visited node all the time.

mficzel:
@fcool Tested a bit and had to adjust the code a little for writes when no read was performed before (lastVisitedNode). To me it seems to work at least as good as before with quite significantly reduced writes. If there are still some edge cases left this should not make the situation for anyone worse so imho this still seems to be the right direction.

PS: Maybe we should discourage to use entities in session data and jobqueue calls in favor of plain value objects and identifiers ... seems a good practice anyways.

Sebobo:
Sadly cannot give any good feedback on this as I never looked into the Session related code. But sounds great!
Or is there something I could test with a large project and check for differences?

mficzel:
@Sebobo as you have much experience with optimizing caches you might be able to measure compare the performance impact especially for larger projects. Also it would be good to compare impact for FlowApplications (with Sessions) that are not the Neos Backend.

PS: We had flow sessions kill a bigger galera cluster because of the current non optimized writing behavior. So this is not purely theoretical.

Neos.Flow/Classes/Session/Data/SessionMetaData.php
```diff
@@ -0,0 +1,140 @@
...
+     * @param string $sessionIdentifier
+     * @param string $storageIdentifier
+     * @param int $lastActivityTimestamp
+     * @param string[] $tags
+     */
+    public function __construct(
+        public /** readonly */ string $sessionIdentifier,
+        public /** readonly */ string $storageIdentifier,
+        public /** readonly */ int $lastActivityTimestamp,
+        public /** readonly */ array $tags
```
mhsdesign: .

Neos.Flow/Classes/Session/Data/SessionMetaData.php
```diff
@@ -0,0 +1,140 @@
...
+    {
+        if ($this->sessionIdentifier != $other->sessionIdentifier) {
+            return false;
+        }
+
+        if ($this->storageIdentifier != $other->storageIdentifier) {
+            return false;
+        }
+
+        if (array_diff($this->tags, $other->tags) || array_diff($other->tags, $this->tags)) {
```
mhsdesign: this seems not super optimised ...

mhsdesign: adjusted via. https://github.com/neos/flow-development-collection/pull/3254/commits/1a16e7e075d9ad3e3320a13da73ddf25355e18f4

mficzel:
close and reopen to trigger tests

mficzel:
I rebased this on Flow 9 and created fresh stats. 

This pr yields a 85% resp. 90% improvement in write (set) cache operations for the session caches in this arguably simple scenario while all the other call numbers remain the same.

procedure: Login * refresh homepage 10x * logout

before:
```
+---------------------------------+-------+
| key                             | num   |
+---------------------------------+-------+
| Flow_Session_MetaData_has       | 23    |
| Flow_Session_MetaData_get_num   | 47    |
| Flow_Session_MetaData_set_num   | 25    |
| Flow_Session_MetaData_remove    | 2     |
| Flow_Session_Storage_has        | 24    |
| Flow_Session_Storage_get_num    | 24    |
| Flow_Session_Storage_set_num    | 47    |
+---------------------------------+-------+
```

after:
```
+---------------------------------+-------+
| key                             | num   |
+---------------------------------+-------+
| Flow_Session_MetaData_has       | 23    |
| Flow_Session_MetaData_get_num   | 48    |
| Flow_Session_MetaData_set_num   | 4     |
| Flow_Session_MetaData_remove    | 2     |
| Flow_Session_Storage_has        | 25    |
| Flow_Session_Storage_get_num    | 24    |
| Flow_Session_Storage_set_num    | 5     |
+---------------------------------+-------+
```

Neos.Flow/Classes/Session/Data/SessionIdentifier.php
```diff
@@ -0,0 +1,44 @@
...
+use Neos\Flow\Utility\Algorithms;
+
+/**
+ * @Flow\Proxy(false)
+ * @internal
+ */
+class SessionIdentifier
+{
+    private function __construct(
+        public /*readonly*/ $value
```
mhsdesign: can be _real_ readonly now with Flow 9

mhsdesign (commented):
Thank you already for the hard work :D
Fyi for other reviewers: It might not seem like it but i reviewed the change already together with Wilhelm and Martin so the last commits _were_ already the adjustments:

- `TASK: Adjust comparison operators in SessionMetadata`
- `TASK: Rename SessionDataStore to SessionKeyValueStore and mark both internal`
- `TASK: Introduce StorageIdentifier value object`
- `TASK: Introduce SessionIdentifier value object`

and we also discussed that the optimisation sorely in the intention for `lastVisitedNode` will not be needed for 9.0 so thanks for following up on that and reverting this `TASK: Remove weirdness that is not needed in neos 9` again ❤️ 

I found that the properties can now be finally _real_ readonly properties and i would like to have a simple comment for `StorageIdentifier` `SessionIdentifier` what they represent (for us flow devs). Besides these nitpicks i will look into the adjustments above you made to refresh my memory and give you my approval soonish.

Neos.Flow/Classes/Session/Data/SessionIdentifier.php
mficzel: You are right, i forgot about that. Will go over again and also add proper argument and return types to the whole session classes.

mficzel: is adjusted

Neos.Flow/Classes/Session/Data/StorageIdentifier.php
```diff
@@ -0,0 +1,44 @@
...
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+namespace Neos\Flow\Session\Data;
+
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Utility\Algorithms;
+
+/**
```
mhsdesign, line 19: i would like to have a simple comment for StorageIdentifier SessionIdentifier what they represent (for us flow devs). 

After your last explanation i already forgot about their meaning a little 😅 as im not that deep into these things ^^

mficzel: Added comments to explain as good as i can ... the distinction between session and storage id has been there beforehand and while i see it i can only guess the real reason and or necessity.

mhsdesign: Thank youuuu!!

kitsunet (approved):

mhsdesign (approved):
Im not really sure that we still have to upcast old sessions from 8.3 or 9.0-dev previously via `createFromSessionIdentifierStringAndOldArrayCacheFormat` but as the logic is now in place and to be marked as depreacted and to be removed with Flow 10 we can surely keep it.

Thanks for putting the work into it and also for adjusting the code as requested ❤️

robertlemke (approved):
Thanks a lot

This looks good to me, makes sense and did work for me using some application I tested it with.

PR #2999, opened 2023-03-30
TASK: Relax symfony component constraints to allow version 6.0

mficzel:
Adjust symfony/(yaml, dom-crawler, console) composer constraints from `^5.1` to `^5.1||^6.0`. Since symfony 6.0 is symfony 5.4 without deprecations (like php 7. support) we should be fine and no changes are needed.

**Upgrade instructions**

Any package that relies on features in those packages that are removed in 6.0 should have a version constraint for that which is good practice anyways.

**Review instructions**

According to the symfony release notes the 6.0 is symfony 5.4 with all deprecations removed. Unless we use deprecated functionality in places where i could not find it we should be safe in relaxing the dependency.  See https://www.heise.de/news/PHP-Framework-Symfony-6-0-erscheint-zeitgleich-mit-Version-5-4-6279941.html - german)

**Checklist**

- [x] Code follows the PSR-2 coding style
- [x] Tests have been created, run and adjusted as needed
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [ ] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [ ] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked wit `!!!` and have upgrade-instructions

PR #3255, opened 2024-01-04
TASK: Correct symfony dependencies in Flow composer.json

ahaeslich:
The upmerge commit `e2b895120eb00a6c2a352ce22d84f0302b6c3c71` wrongly removed symfony 6.0 in the version constraints of `neos/flow`.

Relates: #2999

### DIFF
--- a/Neos.Flow/composer.json
+++ b/Neos.Flow/composer.json
@@ -48,9 +48,9 @@
         "doctrine/common": "^3.0.3",
         "doctrine/annotations": "^1.12",
 
-        "symfony/yaml": "^5.1",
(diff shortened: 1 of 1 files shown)


dlubitz (approved):
Ohh, my bad. I guess this was resulting because of the other "self.version" changes 🙈

ahaeslich:
> Ohh, my bad. I guess this was resulting because of the other "self.version" changes 🙈

No worries. At least we noticed it now. 

And I could combine that in a pair-programming session on how to find out why packages are downgraded during an update 😎 .

PR #3260, opened 2024-01-16
TASK: Fix some nullable php doc types

mhsdesign:
I ran phpstan level 3 once on flow. And it seems we dont specifiy the nullable types correctly, but we document them in the doc string.
So i wrote a little helping script that would add the `|null` php doc annotation to all `@param` and `@return` types if we specify `or NULL` in the message. I carefully reviewed every change and made additionally some manual changes and corrected things. This is completely non breaking as only doc comments are being touched.

This will help for migrating to phpstan level 3.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Cli/Response.php
+++ b/Neos.Flow/Classes/Cli/Response.php
@@ -118,7 +118,7 @@ class Response
     /**
      * Sets color support / styled output to yes, no or auto detection
      *
-     * @param boolean $colorSupport true, false or NULL (= autodetection)
+     * @param boolean|null $colorSupport true, false or NULL (= autodetection)
      * @return void
      */
     public function setColorSupport(bool $colorSupport): void
--- a/Neos.Flow/Classes/Cli/Response.php
+++ b/Neos.Flow/Classes/Cli/Response.php
@@ -47,12 +47,12 @@ class Response
     private $content = '';
 
     /**
-     * @var
+     * @var bool|null
      */
     private $colorSupport;
 
     /**
-     * @var
+     * @var int
      */
     private $outputFormat = self::OUTPUTFORMAT_STYLED;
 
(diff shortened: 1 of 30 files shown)


dlubitz (approved):
LGTM

Neos.Flow/Classes/Cli/Response.php
```diff
@@ -118,10 +118,10 @@ public function getContent(): string
     /**
      * Sets color support / styled output to yes, no or auto detection
      *
-     * @param boolean $colorSupport true, false or NULL (= autodetection)
+     * @param boolean|null $colorSupport true, false or NULL (= autodetection)
      * @return void
      */
-    public function setColorSupport(bool $colorSupport): void
+    public function setColorSupport(bool|null $colorSupport): void
```
kitsunet, line 124: Already has a type hint so I don't see how anyone would use it with NULL at this point and no one has missed it, I think it would be better to remove the reference to NULL from the doc block. We might consider deprecating this altogether and just auto detecting even.

kitsunet (approved):
see comment, otherwise LGTM

Neos.Flow/Classes/Cli/Response.php
mhsdesign: Yes one part of me agrees and i initially thought so too. But the whole class is build to allow `$colorSupport` to be null, and it is the default. This reset is unlikely to be used - especially as the `setColorSupport` is not even used currently - but most of our cli stuff should probably refactored and i just wanted to restore it to a consistent state.

mhsdesign: Oh i overread this

> We might consider deprecating this altogether and just auto detecting even.

this is the way.

Will do in a followup.

kitsunet: yeah fine by me.

PR #3263, opened 2024-01-17
TASK: Remove dead condition in `FlowAnnotationDriver`

mhsdesign:
phpStan rightfully complains in the `FlowAnnotationDriver` about this dead code.

We foreach over the `$tableAnnotation->uniqueConstraints` as `$uniqueConstraint`. But in Line 251 we reassign the variable `$uniqueConstraint` to `['columns' => $uniqueConstraint->columns];` and thus the check `if (!empty($uniqueConstraint->options))` will always fail.

The code was introduced as fix via https://github.com/neos/flow-development-collection/commit/012f0e76a3169ed9b30b2dd23698718c4d550782, but it could never have worked as intended.

One can only guess but this might be the actual intention of the code.

```php
foreach ($tableAnnotation->uniqueConstraints as $uniqueConstraint) {
    $someOtherVariableName = ['columns' => $uniqueConstraint->columns];
    if (!empty($uniqueConstraint->options)) {
        $someOtherVariableName['options'] = $uniqueConstraint->options;
    }
    if (!empty($uniqueConstraint->name)) {
        $primaryTable['uniqueConstraints'][$uniqueConstraint->name] = $someOtherVariableName;
    } else {
        $primaryTable['uniqueConstraints'][] = $someOtherVariableName;
    }
}
```

But as this behaviour is now untouched since Flow 2, it might be already functioning correctly - and thus the dead conditions were removed.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
@@ -249,14 +249,7 @@ class FlowAnnotationDriver implements DoctrineMappingDriverInterface, PointcutFi
             if ($tableAnnotation->uniqueConstraints !== null) {
                 foreach ($tableAnnotation->uniqueConstraints as $uniqueConstraint) {
                     $uniqueConstraint = ['columns' => $uniqueConstraint->columns];
-                    if (!empty($uniqueConstraint->options)) {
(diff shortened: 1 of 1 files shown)


kdambekalns:
I think it was not intended back then. The loop in line 150 uses `as $indexAnnotation`, but the loop in line 161 uses `as $uniqueConstraint` – if it had used `…Annotation` as well, it would have worked as intended, instead of overwriting itself, so to speak.

So, maybe fix it, instead of drop it?

PR #3217, opened 2023-11-04
TASK: Level up to phpstan 2 (Flow 9 adjustments)

mhsdesign:
The upgrade to phpstan level two was introduced via https://github.com/neos/flow-development-collection/pull/3264, this holds the flow 9 specific adjustments.

Related (phpstan level 1) https://github.com/neos/flow-development-collection/pull/3216

**Upgrade instructions**

**Review instructions**

**Checklist**

- [ ] Code follows the PSR-2 coding style
- [ ] Tests have been created, run and adjusted as needed
- [ ] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [ ] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [ ] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked wit `!!!` and have upgrade-instructions

PR #3218, opened 2023-11-05
TASK: Migrate to PHPStan for Flow 8

mhsdesign:
This is a backport of https://github.com/neos/flow-development-collection/pull/3216

Adds PHPStan level 1 to the whole Flow code base and CI.
Psalm was removed.

**Checklist**

- [x] Code follows the PSR-2 coding style
- [x] Tests have been created, run and adjusted as needed
- [ ] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [x] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [x] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked wit `!!!` and have upgrade-instructions

PR #3264, opened 2024-01-17
TASK: Phpstan level 2 for Flow 8.3 and ignore to be fixed things

mhsdesign:
Flow 9 Pr https://github.com/neos/flow-development-collection/pull/3217

**Upgrade instructions**

**Review instructions**

**Checklist**


Neos.Flow/Classes/Command/ResourceCommandController.php
```diff
@@ -232,13 +233,13 @@ public function cleanCommand()
...
         $thumbnailRepository = class_exists(ThumbnailRepository::class) ? $this->objectManager->get(ThumbnailRepository::class) : null;
-        $mediaPackagePresent = $assetRepository && $thumbnailRepository && $this->packageManager->isPackageAvailable('Neos.Media');
+        $mediaPackagePresent = $this->packageManager->isPackageAvailable('Neos.Media');
 
         if (count($brokenResources) > 0) {
             foreach ($brokenResources as $key => $resourceIdentifier) {
                 $resource = $this->resourceRepository->findByIdentifier($resourceIdentifier);
                 $brokenResources[$key] = $resource;
-                if ($mediaPackagePresent) {
+                if ($mediaPackagePresent && class_exists(AssetRepository::class) && class_exists(ThumbnailRepository::class)) {
```
dlubitz: Why not 
```suggestion
                if ($mediaPackagePresent && $assetRepository!==null && $thumbnailRepository!==null) {
```

Neos.Flow/Classes/Command/ResourceCommandController.php
```diff
@@ -280,7 +281,7 @@ public function cleanCommand()
                         ]);
                         $resource->disableLifecycleEvents();
                         $this->resourceRepository->remove($resource);
-                        if ($mediaPackagePresent) {
+                        if ($mediaPackagePresent && class_exists(AssetRepository::class) && class_exists(ThumbnailRepository::class)) {
```
dlubitz:
```suggestion
                        if ($mediaPackagePresent && $assetRepository!==null && $thumbnailRepository!==null) {
```

Neos.FluidAdaptor/Classes/Core/Cache/CacheAdaptor.php
```diff
@@ -68,7 +68,7 @@ public function flush($name = null)
         if ($name !== null) {
             return $this->flowCache->remove($name);
         } else {
-            return $this->flowCache->flush();
+            $this->flowCache->flush();
```
dlubitz, line 71: I don't get, why we should remove the return here. But if we need to, we should remove both.

dlubitz (commented):
Looks good, but I left two questions.

Neos.Eel/Classes/CompilingEvaluator.php
```diff
@@ -103,7 +103,9 @@ protected function generateEvaluatorCode($expression)
 
         if ($result === false) {
             throw new ParserException(sprintf('Expression "%s" could not be parsed.', $expression), 1344513194);
+            /** @phpstan-ignore-next-line */
```
grebaldi:
```suggestion
            /** @phpstan-ignore-next-line $parser->pos appears to be undeclared, because instantiation of $parser is hidden as well */
```

I believe we should comment each occurrence of `/** @phpstan-ignore-next-line */` to give future people (including our future selves) a fair chance to understand and resolve the issue eventually.

L100 would need treatment as well, though:
```php
/** @phpstan-ignore-next-line `CompilingEelParser` has been explicitly excluded from analysis, because it contains generated code that is impossible to fix */
```

(Don't know if my assumption in that comment is entirely correct though)

Neos.Eel/Classes/CompilingEvaluator.php
```diff
@@ -103,7 +103,9 @@ protected function generateEvaluatorCode($expression)
...
         } elseif ($parser->pos !== strlen($expression)) {
+            /** @phpstan-ignore-next-line */
```
grebaldi:
```suggestion
            /** @phpstan-ignore-next-line $parser->pos appears to be undeclared, because instantiation of $parser is hidden as well */
```

Neos.Eel/Classes/Helper/MathHelper.php
```diff
@@ -94,7 +94,7 @@ public function getSQRT2()
      * @param float $x A number
      * @return float The absolute value of the given value
      */
-    public function abs($x = 'NAN')
+    public function abs($x = NAN)
```
grebaldi, line 97: Excellent! I've never heard of `NAN` before. It's not documented as a pre-defined constant, but documented under "floating point numbers": https://www.php.net/manual/en/language.types.float.php#language.types.float.nan

Great choice to use it here!

Neos.Eel/Classes/InterpretedEvaluator.php
```diff
@@ -35,7 +35,9 @@ public function evaluate($expression, Context $context)
 
         if ($res === false) {
             throw new ParserException(sprintf('Expression "%s" could not be parsed.', $expression), 1344514198);
+            /** @phpstan-ignore-next-line */
```
grebaldi:
```suggestion
            /** @phpstan-ignore-next-line $parser->pos appears to be undeclared, because it is declared up in the parent class chain of `InterpretedEelParser`, wherein multiple classes (e.g. `EelParser` and `AbstractEelParser`) are explicitly excluded from analysis, because they contain generated code that is impossible to fix */
```

Neos.Eel/Classes/InterpretedEvaluator.php
```diff
@@ -35,7 +35,9 @@ public function evaluate($expression, Context $context)
...
         } elseif ($parser->pos !== strlen($expression)) {
+            /** @phpstan-ignore-next-line */
```
grebaldi:
```suggestion
            /** @phpstan-ignore-next-line $parser->pos appears to be undeclared, because it is declared up in the parent class chain of `InterpretedEelParser`, wherein multiple classes (e.g. `EelParser` and `AbstractEelParser`) are explicitly excluded from analysis, because they contain generated code that is impossible to fix */
```

Neos.Eel/Classes/Validation/ExpressionSyntaxValidator.php
```diff
@@ -28,7 +28,9 @@ protected function isValid($value)
 
         if ($result === false) {
             $this->addError('Expression "%s" could not be parsed.', 1421940748, [$value]);
+            /** @phpstan-ignore-next-line */
```
grebaldi, line 31:
```suggestion
            /** @phpstan-ignore-next-line $parser->pos appears to be undeclared, because instantiation of $parser is hidden as well */
```

Again, here L25 would need treatment as well:
```php
/** @phpstan-ignore-next-line `EelParser` has been explicitly excluded from analysis, because it contains generated code that is impossible to fix */
```

Neos.Eel/Classes/Validation/ExpressionSyntaxValidator.php
```diff
@@ -28,7 +28,9 @@ protected function isValid($value)
...
         } elseif ($parser->pos !== strlen($value)) {
+            /** @phpstan-ignore-next-line */
```
grebaldi, line 33:
```suggestion
            /** @phpstan-ignore-next-line $parser->pos appears to be undeclared, because instantiation of $parser is hidden as well */
```

Neos.Flow/Classes/Aop/Advice/AdviceChain.php
```diff
@@ -56,7 +57,9 @@ public function proceed(JoinPointInterface &$joinPoint)
         if ($this->adviceIndex < count($this->advices)) {
             $result = $this->advices[$this->adviceIndex]->invoke($joinPoint);
         } else {
-            $result = $joinPoint->getProxy()->Flow_Aop_Proxy_invokeJoinpoint($joinPoint);
+            /** @var ProxyInterface $proxy */
```
grebaldi: This strikes me as weird. So, as if `JoinPointInterface` declares the wrong return type for `getProxy`. If we expect this to always be an `Aop\ProxyInterface`, shouldn't that be made explicit in the `JoinPointInterface` ?:

https://github.com/neos/flow-development-collection/blob/65eea14d531158c3bf079739c6824ab99ab4027b/Neos.Flow/Classes/Aop/JoinPointInterface.php#L20-L25

```diff
-      * @return \Neos\Flow\ObjectManagement\Proxy\ProxyInterface
+      * @return \Neos\Flow\Aop\ProxyInterface
```

I mean, `Aop\ProxyInterface` *does* extend `ObjectManagement\Proxy\ProxyInterface`:
https://github.com/neos/flow-development-collection/blob/68923ac7af238e6109a5021f49f7a548e9238c79/Neos.Flow/Classes/Aop/ProxyInterface.php#L18

So, imho this should be adjusted in `JoinPointInterface`  rather than casting it here.

Neos.Flow/Classes/Cli/Response.php
```diff
@@ -47,12 +47,12 @@ class Response
     private $content = '';
 
     /**
-     * @var
+     * @var boolean|int
```
grebaldi:
```suggestion
     * @var bool|int
```

PSR-12:
> Short form of type keywords MUST be used i.e. bool instead of boolean, int instead of integer etc.
https://www.php-fig.org/psr/psr-12/#25-keywords-and-types

Neos.Flow/Classes/Cli/Response.php
```diff
@@ -118,6 +118,7 @@ public function getContent(): string
     /**
      * Sets color support / styled output to yes, no or auto detection
      *
+     * @deprecated and will be removed https://github.com/neos/flow-development-collection/pull/3262
```
grebaldi: Can't judge, whether it's okay to introduce this deprecation within the scope of this PR. But, it sure feels sneaky :sweat_smile: 

On second thought: Is this maybe a leftover, since you were backporting from your PR targeting 9.0?

Neos.Flow/Classes/Command/ResourceCommandController.php
grebaldi: I second that. `class_exists` is more expensive (and happening in a loop at this point). Null-checks are better.

Additionally, this logic should probably be moved back into `$mediaPackagePresent` as well - to save the `isPackageAvailable` call as well. I assume this was the original intention in L236 and should imho not be touched by a PR that addresses  static analysis only.

Neos.Flow/Classes/Log/ThrowableStorage/FileStorage.php
```diff
@@ -266,8 +266,8 @@ protected function cleanupThrowableDumps(): void
         if ($this->maximumThrowableDumpAge > 0) {
             $cutoffTime = time() - $this->maximumThrowableDumpAge;
 
-            /** @var \SplFileInfo $directoryEntry */
             $iterator = new \DirectoryIterator($this->storagePath);
+            /** @var \DirectoryIterator $directoryEntry */
```
grebaldi:
```suggestion
```

Not necessary imho. phpstan should figure this out on its own. (But does it even care on level 2?)

grebaldi (commented):
Hi @mhsdesign,

I did a full review and added a bunch of comments. Nothing deal-breaking I suppose, mostly karma stuff. The only truly irritating spot is the one that @dlubitz pointed out already.

Also, I can confirm that phpstan analysis on level 2 works locally. Thanks a lot for taking the time to create those fixes :100: 

You may go ahead and address/resolve/dismiss my comments as you see fit and I'll give you a :+1:  in return :)

Neos.Flow/Classes/Aop/Advice/AdviceChain.php
mhsdesign: Yes the interface extends the other. But the extended `\Neos\Flow\Aop\ProxyInterface` seems super internal as it contains this method `Flow_Aop_Proxy_invokeJoinpoint`, and if we adjust this doc comment int `getProxy` which is called by userland people might suddenly start using to internal things like `Flow_Aop_Proxy_invokeJoinpoint`. Thats why the cast.

Neos.FluidAdaptor/Classes/Core/Cache/CacheAdaptor.php
mhsdesign: PhpStan doesnt like to return `void` when returned from another method. Will adjust to add a `return null` below

Neos.Flow/Classes/Cli/Response.php
mhsdesign: No i was trying to sneak it in, but its not that important will sneak it out again ^^

Neos.Eel/Classes/CompilingEvaluator.php
mhsdesign: Yes good idea, well anything regarding the parsers is a lost cause :D

Neos.Flow/Classes/Command/ResourceCommandController.php
mhsdesign: Yess ill be trying the null checks. Its not super straight forward to make phpstan work in this awkward situation.

The "original" intention is not older than a few weeky see https://github.com/neos/flow-development-collection/pull/3218 so that why i changed it so careless ;)

I probably now spend more time on trying phpstan to ignore this code than i would have to fix it properly 😂

mhsdesign:
Thank you for your reveiws @dlubitz and @grebaldi. I adjusted everything ;)

dlubitz (approved):

dlubitz:
Thanks for taking care!

grebaldi (approved):
Thanks a lot, @mhsdesign! :+1: 

btw: Nice idea aliasing that `InternalAopProxyInterface` - much clearer now :)

mhsdesign:
Thank you for your quick re-reviews ;)

PR #3265, opened 2024-01-17
FEATURE: Consider PHP attributes in proxy method building

### LINKED ISSUE #3075: FEATURE: Consider PHP attributes in proxy class and method building

### Description

Proxy classes and methods which are generated through `ProxyClassGenerator` and `ProxyMethodGenerator` respectively don't support PHP 8 attributes yet. For example, while generating a proxy method, the DocBlock including the annotations are preserved and thus generated for the new method, but the attributes of the original method are not. This leads to errors when using attributes instead of annotations in certain situations.

One way to produce an error due to this missing feature is to use a `Defer` attribute from the Flowpack Job Queue package. While generating the proxy classes, Flow will recognize the attribute and generate an advice for the `DeferAspect`. However, the proxy method does not have the attribute anymore and the ReflectionService therefore can't provide the aspect with any arguments of the `Defer` attribute. The error will be something like "Attempt to read property 'queueName' on null".

### Possible Solution

In Flow 9 we are using `laminas/laminas-code` for proxy class and method generation. There is an unfinished pull request (https://github.com/laminas/laminas-code/pull/145) which introduces support for attributes for classes. We may want to help them finishing that pull request and also provide support for generating attributes for methods.

robertlemke:
Added support for preserving PHP 8 attributes in generated proxy class methods. This feature enables correct argument passing from attributes to proxied methods which allows developers to use attributes instead of annotations in most cases.

Resolves #3075

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -281,7 +281,11 @@ class ProxyMethodGenerator extends MethodGenerator
         $formattedArguments = [];
 
         foreach ($arguments as $key => $value) {
-            $formattedArguments[] = "{$key}: " . $this->formatAttributeValue($value, $methodName);
+            if (is_int($key)) {
+                $formattedArguments[] = $this->formatAttributeValue($value, $methodName);
+            } else {
+                $formattedArguments[] = "{$key}: " . $this->formatAttributeValue($value, $methodName);
+            }
         }
 
         return implode(', ', $formattedArguments);
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -125,7 +125,7 @@ class ProxyMethodGenerator extends MethodGenerator
             . $this->getName() . '(';
 
         $output .= implode(', ', array_map(
-            static fn(ParameterGenerator $parameter): string => $parameter->generate(),
+            static fn (ParameterGenerator $parameter): string => $parameter->generate(),
             $this->getParameters()
         ));
 
@@ -329,7 +329,8 @@ class ProxyMethodGenerator extends MethodGenerator
                 $this->getFullOriginalClassName(),
                 $methodName,
                 get_debug_type($value)
-            ), 1705501433
+            ),
+            1705501433
         );
     }
 }
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -13,14 +13,19 @@ namespace Neos\Flow\ObjectManagement\Proxy;
 
 use Laminas\Code\Generator\DocBlockGenerator;
 use Laminas\Code\Generator\MethodGenerator;
+use Laminas\Code\Generator\ParameterGenerator;
 
 /**
- * Generator for proxy methods
+ * Class ProxyMethodGenerator
+ *
+ * This class is responsible for generating proxy methods that can be used as method interceptors.
+ * It extends the MethodGenerator class.
  */
 class ProxyMethodGenerator extends MethodGenerator
 {
     protected string $addedPreParentCallCode = '';
     protected string $addedPostParentCallCode = '';
+    protected string $attributesCode = '';
 
     /** @var class-string|null */
     protected ?string $fullOriginalClassName = null;
@@ -30,6 +35,7 @@ class ProxyMethodGenerator extends MethodGenerator
         $instance = parent::fromReflection($reflectionMethod);
         assert($instance instanceof static);
         $instance->fullOriginalClassName = $reflectionMethod->getDeclaringClass()->getName();
+        $instance->attributesCode = $instance->buildAttributesCode($reflectionMethod);
         return $instance;
     }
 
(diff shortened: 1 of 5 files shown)


Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
```diff
@@ -174,4 +242,78 @@ protected function buildCallParentMethodCode(string $fullClassName, string $meth
...
+            $formattedArrayElements = implode(', ', array_map(function ($key, $value) {
+                return is_int($key)
+                    ? $this->formatAttributeValue($value)
+                    : "\"{$key}\" => " . $this->formatAttributeValue($value);
+            }, array_keys($value), $value));
+            return "[{$formattedArrayElements}]";
+        }
+
+        // Fallback for any other types (shouldn't happen with PHP attributes)
+        return '';
```
kitsunet: Do we want to except here instead? Just to be sure so that potential future expansions of the syntax get to an exception instead of resulting in weird proxies?

I am probably fine either way, but I could imagine e.g. floats being added which would then fall through to here...

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
```diff
@@ -174,4 +242,78 @@ protected function buildCallParentMethodCode(string $fullClassName, string $meth
...
+     * @param array $arguments An array of arguments for attributes.
+     *
+     * @return string The formatted arguments as a string.
+     */
+    private function formatAttributesArguments(array $arguments): string
+    {
+        $formattedArguments = [];
+
+        foreach ($arguments as $key => $value) {
+            $formattedArguments[] = "{$key}: " . $this->formatAttributeValue($value);
```
kitsunet: According to the docs something like this here is valid:

`#[JsonSerialize('foobar')]`

so we probably need to have a test for this kind of value without key and properly handle it here.

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/ClassWithPhpAttributes.php
```diff
@@ -21,4 +21,12 @@
 #[SampleAttribute(ClassWithPhpAttributes::class, ['foo' => 'bar'])]
 class ClassWithPhpAttributes
 {
+    /**
+     * @return void
+     */
+    #[Flow\Around(pointcutExpression: "method(somethingImpossible())")]
```
kitsunet, line 27: Now I wonder, I think we always create the "Flow" use statement in proxy classes, but what about other namespaces, do you know if we copy all use statements over to the proxy, otherwise this might fail?

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
robertlemke: Yes, good point, I pushed a commit handling that and added support for floats.

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/ClassWithPhpAttributes.php
robertlemke: This is independent from any imports, the proxy method uses a fully qualified attribute name including the absolute namespace.

kitsunet: oh I see, neat then!

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
robertlemke: Thanks! I added support for that, too.

kitsunet (approved):
Very nice, thanks for the quick additions :)

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
```diff
@@ -83,7 +100,59 @@ public function generate(): string
...
+        }
+
+        $output .= $this->getVisibility()
+            . ($this->isStatic() ? ' static' : '')
+            . ' function '
+            . ($this->returnsReference() ? '& ' : '')
+            . $this->getName() . '(';
+
+        $output .= implode(', ', array_map(
+            static fn(ParameterGenerator $parameter): string => $parameter->generate(),
```
kdambekalns:
```suggestion
            static fn (ParameterGenerator $parameter): string => $parameter->generate(),
```

Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
```diff
@@ -174,4 +243,93 @@ protected function buildCallParentMethodCode(string $fullClassName, string $meth
...
+            }, array_keys($value), $value));
+            return "[{$formattedArrayElements}]";
+        }
+        throw new UnsupportedAttributeException(
+            sprintf(
+                'Failed rendering proxy method %s::%s because an attribute contained an unsupported value type (%s)',
+                $this->getFullOriginalClassName(),
+                $methodName,
+                get_debug_type($value)
+            ), 1705501433
```
kdambekalns:
```suggestion
            ),
            1705501433
```

Neos.Flow/Tests/Functional/ObjectManagement/ProxyCompilerTest.php
```diff
@@ -23,6 +25,8 @@
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PHP81\BackedEnumWithMethod;
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PrototypeClassA;
 use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\PrototypeClassK;
+use Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\SampleAttribute;
```
kdambekalns:
```suggestion
```

kdambekalns (commented):
StyleCI fixes…

kdambekalns (approved):
👍 by 👀

PR #3268, opened 2024-01-18
TASK: Improve help text for doctrine:migrate

robertlemke:
Most users probably didn't know that it is possible to use "prev" or "next" as version names for `./flow doctrine:migrate --version'. This is now documented as part of the help message and additionally the version alias "previous" is automatically converted to "prev" internally.


ahaeslich:
Just noticed: should we have merged this into 8.3 🤔

PR #3249, opened 2023-12-05
Revert "BUGFIX: Sanitize uploaded svg files for suspicious contents"

dlubitz:
As discussed in team meeting, this reverts commit a1642ef31f19a974f34a302c98c13c77b6422ba1
as it raises issues when workong with streams (https://github.com/neos/flow-development-collection/issues/3246).

Original PR: https://github.com/neos/flow-development-collection/pull/3172

See also: https://github.com/neos/flow-development-collection/issues/3248

PR #3273, opened 2024-01-21
TASK: Followup 8.3 Revert "BUGFIX: Sanitize uploaded svg files for suspicious contents"

mhsdesign:
Followup #3249

This reverts commit https://github.com/neos/flow-development-collection/commit/a1642ef31f19a974f34a302c98c13c77b6422ba1 fully for 8.3

In the upmerge https://github.com/neos/flow-development-collection/commit/e2b895120eb00a6c2a352ce22d84f0302b6c3c71#diff-b54379d1fdcdc7a8433b179ba5a080f729e68a559cd2a6203e9d1c49b7789c3e this composer json adjustment in Flow got lost.


composer.json
```diff
@@ -16,6 +16,7 @@
         "ext-reflection": "*",
         "ext-xml": "*",
         "ext-xmlreader": "*",
+        "neos/composer-plugin": "^2.0",
```
dlubitz, line 19: @mhsdesign Can we revert the change in order? As this will lead to other merge conflicts on next upmerge from 7.3.

dlubitz:
Thanks for fixing that. Somehow this slipped through.

composer.json
mhsdesign: im afraid that this is correct. I didnt do this manually but the composer manifest merger did this.
Also there will be possibly only one last upmerge from 7.3 so i guess this is all fine.

kitsunet: Comes from here ( https://github.com/neos/flow-development-collection/pull/3273/files#diff-b54379d1fdcdc7a8433b179ba5a080f729e68a559cd2a6203e9d1c49b7789c3eR32 ), so (new) order should be correct, I assume someone merged a change with manually modified composer.json before, hint always run the merger for PR-ing ;)

dlubitz (approved):

PR #3275, opened 2024-01-24
TASK: `resource:clean` followup #1678

mhsdesign:
While reading the code, looking for improvement, it seems tedious that we `getIdentifierByObject` just to `findByIdentifier` a few lines later.

This happened due to a funny history of back and forth.

At first - 2014 - `resource:clean` was introduced looping over the PersistentResource: https://github.com/neos/flow-development-collection/commit/8a1ce0fba6cb0bf301f971a6d7d5675e0c038d75

Then - 2016 - it was decided to save the sha1 and loop over them and retrieve the asset via `findOneBySha1`: https://github.com/neos/flow-development-collection/commit/879fba19f93d0a8628682698e57da9f1b58ad7d4

But that did not improve the situation as described in https://github.com/neos/flow-development-collection/pull/1678 and was removed again - 2019.

So in functionality we made a full round, im just here to followup on the last fix to restore the full state syntactically as it was once though of.

### DIFF
--- a/Neos.Flow/Classes/Command/ResourceCommandController.php
+++ b/Neos.Flow/Classes/Command/ResourceCommandController.php
@@ -273,8 +273,8 @@ class ResourceCommandController extends CommandController
                 }
             }
             $response = null;
-            while (!in_array($response, ['y', 'n', 'c'])) {
-                $response = $this->output->ask('<comment>Do you want to remove all broken resource objects and related assets from the database? (y/n/c) </comment>');
+            while (!in_array($response, ['y', 'n'])) {
+                $response = $this->output->ask('<comment>Do you want to remove all broken resource objects and related assets from the database? (y/n) </comment>');
             }
 
             switch ($response) {
@@ -318,9 +318,6 @@ class ResourceCommandController extends CommandController
                     break;
                 case 'n':
                     $this->outputLine('Did not delete any resource objects.');
-                    break;
-                case 'c':
-                    $this->outputLine('Stopping. Did not delete any resource objects.');
                     $this->quit(0);
                     break;
             }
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Command/ResourceCommandController.php
```diff
@@ -221,6 +221,7 @@ public function cleanCommand()
         $resourcesCount = $this->resourceRepository->countAll();
         $this->output->progressStart($resourcesCount);
 
+        /** @var list<PersistentResource> $brokenResources */
```
bwaidelich, line 224: what's that type hint, is `list` supported by common IDEs or should this be:

```suggestion
        /** @var array<PersistentResource> $brokenResources */
```

mhsdesign: yes it is well understood and its a short form of `array<int, mixed>`. I use it to denote further that the keys are numeric and (should) be simply ascending without holes. See `array_is_list`.

For example `array<int, string>` might mean that its a map of constant values to something else, of course the type is the same for phpstan, but visually its not.

bwaidelich: TIL, nice

bwaidelich (approved):
+1 by reading

Neos.Flow/Classes/Command/ResourceCommandController.php
```diff
@@ -275,8 +273,8 @@ public function cleanCommand()
                 }
             }
             $response = null;
-            while (!in_array($response, ['y', 'n', 'c'])) {
-                $response = $this->output->ask('<comment>Do you want to remove all broken resource objects and related assets from the database? (y/n/c) </comment>');
```
mhsdesign, line 279: it seems the `cancel` option was meant to be a hard stop while `n` would ask for each image.
This was NEVER implement that way not even for the first draft robert did. Thats why i removed `cancel` as its equivalent to `no`.

PR #3278, opened 2024-01-25
!!! TASK: Fix `TextIterator::following` and `preceding`

mhsdesign:
Accidentally they have been typed wrongly. First in phpdoc, which is harmless and later actual types introduced in https://github.com/neos/flow-development-collection/commit/70b671228ee4f66c54fb7fbfa390aac12b5a71c5#diff-947f5937b1e181a6e4ae7bb23349d22d839b073a07104b884c08583cc12f63df enforced that.

The tests didnt fail, because as strict types were not enabled php just cast the int's to string.

The tests, also casting when using assertEquals, didnt notice that.

This is required in preparation for https://github.com/neos/flow-development-collection/pull/3261

### DIFF
--- a/Neos.Utility.Unicode/Classes/TextIterator.php
+++ b/Neos.Utility.Unicode/Classes/TextIterator.php
@@ -1,4 +1,7 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Neos\Utility\Unicode;
 
 /*
@@ -195,13 +198,13 @@ class TextIterator implements \Iterator
     }
 
     /**
-     * Returns the next elment following the character of the original string
+     * Returns the offset of the next element following the character of the original string
      * given by its offset
      *
      * @param integer $offset The offset of the character
-     * @return string The element following this character
+     * @return int The offset of the element following this character
      */
-    public function following(int $offset): string
+    public function following(int $offset): int
     {
         $this->rewind();
         while ($this->valid()) {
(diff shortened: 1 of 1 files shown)


kitsunet:
Another one of those packages we should actually check if we use them and if we still need them

kitsunet (approved):

kdambekalns (approved):
The new return value `-1` puzzles me, where did you get that?

Then again, `TextIterator` did not even survive the end of PHP6, the only proper mentionI could find was http://zmievski.org/2006/07/all-the-little-pieces-or-textiterator-in-php-6/ 🤣

mhsdesign:
> The new return value -1 puzzles me, where did you get that?

@kdambekalns if the iterator is not valid, an error would have been thrown while accessing the undeclared variable `$currentElement`. So i thought why not just return -1 instead of fatal error :D

But then again i have no clue about the code, will probably never use it and have no idea why one would use it ^^

kdambekalns:
Oh, for use-cases see the linked article from 2006. 🙂 Actually makes sense. 

But ok, fine with me. I had hoped you found what the original method would have returned in that case…

mhsdesign:
Well now you challenged me xD
I looked a bit into that supposed `TextIterator` and what happened to php 6 in general. And there is very little information on the web. And it all seemed to vanish :D

The behaviour of the `TextIterator` can be inferred from the original php6 code which resides in the `experimental/first_unicode_implementation` in the php-src repo.

Sadly there are only about 4 meaningless tests like that one: https://github.com/php/php-src/blob/experimental/first_unicode_implementation/ext/unicode/tests/iterator_001.phpt

But there is a little documentation: https://github.com/php/php-src/blob/experimental/first_unicode_implementation/README.UNICODE#L561-L582

And of course one can look into the c code of the `TextIterator` where there are also nice comments <3:
https://github.com/php/php-src/blob/experimental/first_unicode_implementation/ext/unicode/unicode_iterators.c#L1208-L1248


For real experiments one can actually play with this docker image to see php6 in action: https://github.com/wizaplace/docker-php-6-apache

But i dont want to dug in there deeper than that :P

mhsdesign:
Followup discussion: https://github.com/neos/flow-development-collection/issues/3279

PR #3286, opened 2024-01-27
!!! FEATURE: `ViewInterface` returns PSR `StreamInterface`

mhsdesign:
Neos adjustments https://github.com/neos/neos-development-collection/pull/4856

- the views are now independent of the `ControllerContext`
  - `ViewInterface::setControllerContext` is not part of the interface anymore and will only be called on demand
- the `ActionRequest` if necessary can be accessed via the variable "request" (like "settings")
- `ViewInterface::canRender` was required for fallbackviews which have been removed long ago, and so this artefact will be removed as well.
- !!! the return type is now forced to be either a `ResponseInterface` or a `StreamInterface`. Simple strings must be wrapped in a psr stream! (see `StreamFactoryTrait::createStream`)

Related to https://github.com/neos/flow-development-collection/pull/3232


mhsdesign:
@kitsunet That might be in conjunction with https://github.com/neos/flow-development-collection/pull/3232 an actual step to make the controllerContext obsolete?

But to still support martins use case and have the `response` always available https://github.com/neos/flow-development-collection/pull/3232#issuecomment-1910563163 i introduced a new `ActionMessages` dto, wich i pass secretly to the view under the key `actionMessages` like we do it with `settings`:

```php
 final readonly class ActionMessages
 {
     private function __construct(
         public ActionRequest $request,
         public ActionResponse $response
     ) {
     }

     public static function create(ActionRequest $request, ActionResponse $response)
     {
         return new self($request, $response);
     }
 }
```

This is the secret passing logic where we make the `setControllerContext` optional and only call it if existent:

```php
$this->view->assign('actionMessages', ActionMessages::create($this->request, $this->response));
if (method_exists($this->view, 'setControllerContext')) {
    $this->view->setControllerContext($this->controllerContext);
}
```
One _could_ say that nothing changed but i think the `ActionMessages` is an improvement over the controller context as it offers less weird things.

Of course it would be ideal to not have to pass the pre-`response` anywhere but just work with the request - like my first commit shows, ~but that wouldnt allow many complex usecases.~ Edit no see my next comment...

Neos.Flow/Classes/Mvc/View/JsonView.php
```diff
@@ -200,7 +195,10 @@ public function setConfiguration(array $configuration)
      */
     public function render()
     {
-        $this->controllerContext->getResponse()->setContentType('application/json');
+        /** @var ActionMessages $actionMessages */
```
mhsdesign: the json view should probably just return a new Response instead of that.

... then only fusion uses the ActionResponse ...

mhsdesign:
> Of course it would be ideal to not have to pass the pre-response anywhere but just work with the request - like my first commit shows, but that wouldnt allow many complex usecases. 

yes that would be the goal. To only pass the request but not the Response outside of the view and its possible!!!

In this pr i found a way to make the neos fusion view work with plugins and fusion forms despite having no controller context at hand. https://github.com/neos/neos-development-collection/pull/4856

It works by creating a dummy controller context in the view and later rendering out that response from the view. That gives us the possibility to actually decouple views and controllers a lot. Simply by using the feature to render a psr response from a view.

The same i did in a much smaller scope for the json view, so that the content type is set automatically.

My prior `ActionMessages` gedöns has been reverted.

Neos.Flow/Classes/Mvc/View/JsonView.php
```diff
@@ -12,11 +12,13 @@
...
  */
 
+use GuzzleHttp\Psr7\Response;
+use GuzzleHttp\Psr7\Utils;
 use Neos\Flow\Annotations as Flow;
-use Neos\Flow\Mvc\Controller\ControllerContext;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
 use Neos\Utility\ObjectAccess;
 use Neos\Utility\TypeHandling;
+use Psr\Http\Message\ResponseInterface;
```
mhsdesign, line 21: deprecate

Neos.Flow/Classes/Mvc/View/ViewInterface.php
```diff
@@ -53,12 +54,9 @@ public function assign($key, $value);
...
 
     /**
-     * Tells if the view implementation can render the view for the given context.
-     *
-     * @param ControllerContext $controllerContext
-     * @return boolean true if the view has something useful to display, otherwise false
+     * The view is satisfied by the assigned variables and can render
      */
-    public function canRender(ControllerContext $controllerContext);
+    public function canRender(): bool;
```
mhsdesign: todo remove?

Neos.Flow/Classes/Mvc/View/ViewInterface.php
mhsdesign, line 1: todo only return psr stream or psr responses!

mhsdesign:
About `ViewInterface::canRender`
    
Initially `canRender` was introduced to allow special logic in a controller to determine the correct view out of multiple options. https://github.com/neos/flow-development-collection/commit/cb8fa242150ccdd8d5f9e7121c70c896efaaa62b


But with https://github.com/neos/flow-development-collection/commit/d69bcfe290a04b4079c8a764bb8a2c97b72d13d8 the canRender() checks were removed and leaves the method with zero usages and no _real_ implementation.

Neos.Flow/Classes/Mvc/View/ViewInterface.php
```diff
@@ -52,21 +52,12 @@ public function assign($key, $value);
      */
     public function assignMultiple(array $values);
```
mhsdesign: todo add strict types also here?

Neos.Flow/Tests/Functional/Mvc/ViewsConfiguration/Fixtures/TemplateView.php
```diff
@@ -70,31 +62,19 @@ public function assign($key, $value)
...
-     */
-    public function canRender(ControllerContext $controllerContext)
-    {
-        return true;
-    }
-
     /**
      * Renders the empty view
      *
      * @return string An empty string
```
kitsunet, line 73:
```suggestion
     * @return StreamInterface An empty string
```

Neos.Flow/Classes/Mvc/View/ViewInterface.php
```diff
@@ -26,53 +25,45 @@ interface ViewInterface
     /**
      * Sets the current controller context
      *
+     * @deprecated if you absolutely need access to the current request please assign a variable.
```
kitsunet, line 28: Will this even show in IDEs with the function commented out? bit nasty I guess 🤷

kitsunet (approved):
I think the more important headline this PR should have is View returns Stream, because that will hit many peoples feet updating to this. I think it's a good direction, minor comments, probably overlooked osmething, probably something needs fixing in the future, but we should do it.

Neos.Flow/Classes/Mvc/View/ViewInterface.php
mhsdesign: documenting this via 
> @method setControllerContext(ControllerContext $controllerContext) @deprecated
on the view-interface will not have the desired effect of annotating it as deprecated but would still allow goto clicking.

In the end all views extend the AbstractView which still contains a definition for `setControllerContext` (but deprecated) so i guess its fine?

Neos.Flow/Tests/Unit/Mvc/Controller/ActionControllerTest.php
```diff
@@ -201,45 +201,14 @@ public function processRequestThrowsExceptionIfRequestedActionIsNotPublic()
...
-     */
-    public function processRequestInjectsSettingsToView()
+    public function processRequestInjectsSettingsAndRequestToView()
     {
         $this->actionController = $this->getAccessibleMock(ActionController::class, ['resolveActionMethodName', 'initializeActionMethodArguments', 'initializeActionMethodValidators', 'resolveView', 'callActionMethod']);
         $this->actionController->method('resolveActionMethodName')->willReturn('indexAction');
 
         $this->inject($this->actionController, 'objectManager', $this->mockObjectManager);
         $this->inject($this->actionController, 'controllerContext', $this->mockControllerContext);
+        $this->inject($this->actionController, 'request', $this->mockRequest);
```
mhsdesign, line 211: One last question, do we want to use `request` as naming here or preserve that slot for a pretty psr request, despite that in Fusion request is an ActionRequest (which will not change anytime?)

Do we all agree to keep the ActionRequest for some time and have it hold all the context, like matched controller, base uri and routing parameters?

Neos.Flow/Classes/Error/ProductionExceptionHandler.php
```diff
@@ -39,7 +40,15 @@ protected function echoExceptionWeb($exception)
...
                 try {
-                    echo $this->buildView($exception, $this->renderingOptions)->render();
+                    $stream = $this->buildView($exception, $this->renderingOptions)->render();
+                    if ($stream instanceof ResponseInterface) {
+                        /**
+                         * The http status code will already be sent, and we are only currently interested in the content stream
+                         * Thus, we unwrap the repose here:
+                         */
+                        $stream = $stream->getBody();
+                    }
```
mhsdesign, line 50: @Sebobo we just talked about that the `FusionExceptionView`'s headers are removed by this logic.
What would be the desired logic?

We generate the `header` from the exceptions status code currently.

Sebobo: Can't we here get the headers part now too and apply it to the response?

mhsdesign: Yeah we can add the headers but which status code should we use? The one inferred from the FlowException / 500 or the one from the view?

Maybe: use from the view if it’s NOT 200 and else fallback to the current logic?

mhsdesign: See https://github.com/neos/neos-development-collection/issues/4964 and can be fixed as follow up. Behaviour did not change yet, but we have the possibility to fix it ;)

Neos.Flow/Classes/Mvc/View/JsonView.php
mhsdesign: Now that we deprecated the json view ... what is our answer to code like this? :O

```php
/**
 * @var array
 */
protected $viewFormatToObjectNameMap = [
    'html' => FusionView::class,
    'json' => JsonView::class,
];
```

do we need a new simple json view? :P

cc @kitsunet

PR #3287, opened 2024-01-28
WIP: Experiment: Create action response from http response

### LINKED ISSUE #2492: ActionResponse::replaceHttpResponse behaves unexpectedly
Related to #2478

The `replaceHttpResponse()` should make sure that the ContentType (and any other headers really) that is potentially set inside a HttpResponse which is returned from a view should be applied. But apparently it isn't - sooo... what is happening then?
Only thing I could imagine is this:
- the `ActionResponse` (`$this->response`) already has a ContentType set (e.g. from content negotiation)
- after replacing the `HttpResponse`, the `ActionResponse` gets merged into the `HttpResponse` (inside the dispatch middleware - see https://github.com/neos/flow-development-collection/blob/7.0/Neos.Flow/Classes/Mvc/DispatchMiddleware.php#L46 resp. https://github.com/neos/flow-development-collection/blob/7.0/Neos.Flow/Classes/Mvc/ActionResponse.php#L314)
- at that place, the ContentType from the view returned `HttpResponse` is overwritten again

So given that, I think even though this fixes this single symptom, there is a bigger general logic issue below that probably needs to be fixed - i.e. a `replaceHttpResponse()` not preventing the `ActionResponse` from overwriting properties in that HttpResponse again. Maybe that is even intended for some properties of the `ActionResponse`? Dunno. It basically boils down to what is expectation for the response of an action like this:

```php
public function weirdResponseAction()
{
    $this->response->setContentType('application/xml');
    $this->response->setRedirectUri(new Uri('somewhereelse'), 304);
    $this->response->setContent('Hello World!');
    $this->response->setCookie(new Cookie(...));
    $this->response->setHttpHeader('X-My-Header', 'My-Header-Value');
    $this->response->replaceHttpResponse(new Response(200, ['Content-Type' => 'application/json'])); // which is equivalent to letting the view render a PSR Response
}
```
- should the response status be 200 or 304?
- should the response header `Location` be set to indicate a redirect? (should probably be in sync with the above)
- should the response content be "Hello World!" or empty?
- should the response content type be "application/json" or "application/xml"?
- should the response header `X-My-Header` be set?
- should the response contain a `Set-Cookie` header?
- should any of the above change if the `replaceHttpResponse()` call happened earlier?

So is `replaceHttpResponse()` a full/partial/conditional escape hatch from the `ActionResponse` abstraction?

mhsdesign:
Resolves: #2492

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Mvc/ActionResponse.php
+++ b/Neos.Flow/Classes/Mvc/ActionResponse.php
@@ -1,6 +1,7 @@
 <?php
 namespace Neos\Flow\Mvc;
 
+use GuzzleHttp\Psr7\Uri;
 use GuzzleHttp\Psr7\Utils;
 use Neos\Flow\Http\Cookie;
 use Psr\Http\Message\ResponseInterface;
@@ -52,11 +53,6 @@ final class ActionResponse
      * @var array
      */
     protected $headers = [];
-
-    /**
-     * @var ResponseInterface|null
-     */
-    protected $httpResponse;
 
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/Mvc/ActionResponse.php
```diff
@@ -231,13 +227,36 @@ public function getContentType(): string
...
+        $location = $httpResponse->getHeader('Location');
+        if (!empty($location)) {
+            $this->setRedirectUri(new Uri(reset($location)), $httpResponse->getStatusCode());
+            $httpResponse = $httpResponse->withoutHeader('Location');
+        } else {
+            $this->setStatusCode($httpResponse->getStatusCode());
+        }
+
+        foreach ($httpResponse->getHeader('Set-Cookie') as $value) {
+            $cookie = Cookie::createFromRawSetCookieHeader($value);
```
mhsdesign, line 252: not sure if its wise to introduce there the first actual usage: https://github.com/neos/flow-development-collection/issues/3288

kitsunet: I would mid term want to get rid of the whole Cookie object or remodel it maybe? IDK, sure a DTO to make getCookieParams/withCookieParams a bit nicer to use would be great, but right now the cookie object stems from our own http implementation way back. So anything we can cut from it is a win IMHO 😂

mhsdesign:
As we want to remove the `ActionResponse` and operate on `ResponseInterface` directly we don't need the broken `replaceHttpResponse` handling any further, i will close this pr.

See https://github.com/neos/flow-development-collection/issues/3289

Issue #1892, opened 2020-01-06
MVC Dispatcher is emitted even in CLI Dispathcer (Slack Conversation)


sorenmalling:
**Søren Malling:rainbowsheep: 3:12 PM**
I seem to have found a change in FLOW 6.0 I can't find mentioned in the changelog or other docs. When the Neos\Flow\Security\Context class is injected with @Flow\Inject I get the following exeption in the CLI
```
Cannot build object "Neos\Flow\Security\Authentication\TokenAndProviderFactoryInterface" because constructor injection is not available in the compile time Object Manager.
Refactor your code to use setter injection instead. Configuration source:
automatically registered class. Build stack:
Lelesys\SuperUser\Service\SuperUserService, Neos\Flow\Security\Context
```

What changed on this subject? (edited) 

**Christian Müller 3:16 PM**
nothing, you try to use it during compile time, that was never allowed
the refactoring just split things, so while before Security\Context was probably not ossible at compile time, now it's the TokenAndProviderFactoryInterface

**Søren Malling:rainbowsheep: 3:19 PM**
Can you enlighten me a bit? where Compile time kick in when i run the ./flow  command and this Lelesys\SuperUser\Service\SuperUserService (which is `@Flow\Scope("singleton")` ) is injected in a Web controller and mentioned in a signalSlot dispatcher registration in Package.php. No exception before upgrading from Flow 5 to Flow 6 so trying to track down the real issue

**Christian Müller 3:21 PM**
interesting
yeah signal slot might be the issue
compile time is a "run" of flow where proxies are build
so no injection available etc.
calling flow surely can trigger a compiletime "run" if the caches are empty

**Søren Malling:rainbowsheep: 3:24 PM**
Got it! `
`$dispatcher->connect('Neos\Flow\Mvc\Dispatcher', 'afterControllerInvocation', 'Lelesys\SuperUser\Service\SuperUserService', 'appendLogoutLink');`

This registration caused it
Didn't think about the signal slot dispatcher was a issue :slightly_smiling_face: Thanks!

**Christian Müller 3:25 PM**
that is weird though
if you can I would further track that, because why would the MVC Dispatcher be instantiated in a compile time (CLI) request?
there must be some reason for that
since 6.0 CLI has a separate dispatcher
so the MVC dispatcher is not even needed for a CLI request

**Søren Malling:rainbowsheep: 3:27 PM**
The MVC dispatcher i used in the Cli Requesthandler - could that be why ?

[...]

PR #3296, opened 2024-01-30
!!! TASK: `Mvc\Dispatcher::afterControllerInvocation` will not be called for CLI requests anymore.

mhsdesign:
Reverts that the cli dispatcher invokes the mvc slots dispatcher slots with cli requests.

When there was one dispatcher the slot `afterControllerInvocation` was fired for both cli and web request. (Seemingly only ever at Runtime?)

Then with the split of web and cli dispatchers this legacy layer was introduced:

https://github.com/neos/flow-development-collection/commit/cf55b180c953de8d02adb680216f5c24a3524237#diff-2c9408e74a8ac737f84468e74c23956d2057e64196e66b97281905d8697226ca

Now during a short time the `Mvc\Dispatcher::afterControllerInvocation` signal was now also called for cli request during compile time.

With this bugfix https://github.com/neos/flow-development-collection/pull/2529 the initial behaviour was restored again. For cli request it will only fire at runtime, and web request are either way always runtime.

This breaking change cleans up the legacy layer.

- The original signals `Mvc\Dispatcher` `'afterControllerInvocation'` and `'beforeControllerInvocation'`
  Will be only invoked for action requests.
They still only fire at runtime, as web requests happen at runtime.

- The with Flow 6.0 introduced signals `Cli\Dispatcher` `'afterControllerInvocation'` and `'beforeControllerInvocation'`
  Will still be only invoked for cli requests.
Will still fired either at compile or runtime, as cli requests can happen always.

**Upgrade instructions**

In case you used the MVC signals and relied on it to also be invoked for CliRequest, you need to connect as well to the respective Cli\Dispatcher signal. But keep in mind that you might need to check if flow is in runtime as this signal will be also fired for compile time unlike the mvc signal before.

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Cli/Dispatcher.php
+++ b/Neos.Flow/Classes/Cli/Dispatcher.php
@@ -77,6 +77,7 @@ class Dispatcher
      * @param Response $response
      * @param CommandControllerInterface $controller
      * @return void
+     * @Flow\Signal
      */
     protected function emitBeforeControllerInvocation(Request $request, Response $response, CommandControllerInterface $controller)
     {
@@ -90,6 +91,7 @@ class Dispatcher
      * @param Response $response
      * @param CommandControllerInterface $controller
      * @return void
+     * @Flow\Signal
      */
     protected function emitAfterControllerInvocation(Request $request, Response $response, CommandControllerInterface $controller)
     {
--- a/Neos.Flow/Classes/Cli/Dispatcher.php
+++ b/Neos.Flow/Classes/Cli/Dispatcher.php
@@ -5,7 +5,6 @@ use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Cli\Exception\InfiniteLoopException;
 use Neos\Flow\Cli\Exception\InvalidCommandControllerException;
 use Neos\Flow\Cli\Exception\StopCommandException;
-use Neos\Flow\ObjectManagement\CompileTimeObjectManager;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 
 /**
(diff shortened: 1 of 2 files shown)


mhsdesign:
Related https://github.com/neos/flow-development-collection/pull/3232#issuecomment-1911647934

and @kitsunet 's comment:

> cli dispatch is a bit of a mess, and good point we should probably breaking fix it for this release. it has been this way ever since we split the two. It was one dispatcher at some point but that really didn't work well api wise, so I ripped it apart brutally with some dirt left behind....


And https://github.com/neos/flow-development-collection/issues/1892

Neos.Flow/Classes/Cli/Dispatcher.php
```diff
@@ -104,16 +93,6 @@ protected function emitBeforeControllerInvocation(Request $request, Response $re
      */
```
kitsunet, line 95: These still need `@Flow\Signal`

Neos.Flow/Classes/Cli/Dispatcher.php
```diff
@@ -104,16 +93,6 @@ protected function emitBeforeControllerInvocation(Request $request, Response $re
...
     protected function emitAfterControllerInvocation(Request $request, Response $response, CommandControllerInterface $controller)
```
mhsdesign, line 96: todo add actual @Flow\Signal

kdambekalns:
> Seemingly only ever at Runtime?

Yes, because (I guess) at compile time there should be no request being handled… even CLI. No?

kdambekalns (commented):
I think it's fine (mind the missing annotation, thouhg).

kdambekalns (approved):
👍 by 👀

kdambekalns:
Another review for this one, maybe?

kitsunet (approved):

PR #3297, opened 2024-01-30
EXPERIMENT: FEATURE: Introduce trait `ActionToMethodDelegation` to create own simple controller.

mhsdesign:
The name of the previously suggested `SimpleActionController` looked for improvement.

The `Simple*` prefix suggest, that the initial implementation blew up too much. But ideally a name would be more obvious and not suggest that there is another _complex_ action controller.

 I could not come up with a better name, but we can make the controllers more composable.

 This change introduces a trait which implements the logic of the previous `SimpleActionController`.
 That way people can implement the `ControllerInterface` directly and use a trait that will map action to method name.

```php
 class SimpleActionTestController implements ControllerInterface
 {
     use ActionToMethodDelegation;

     public function addTestContentAction(ActionRequest $actionRequest): ActionResponse
     {
         $response = new ActionResponse();
         $response->setContent('Simple');
         return $response;
     }
 }
```

**Upgrade instructions**

**Review instructions**

See also
- https://github.com/neos/flow-development-collection/pull/3232#discussion_r1466751316

**Checklist**


kitsunet:
Haha, well, I wanted to do this after I finish my work stuff, well, good good. Btw, this ignores the public method thing you [noted](https://github.com/neos/flow-development-collection/pull/3232#discussion_r1390419285) on the original simple controller chnage before, and I wanted to comment that that needs reflection which is nasty, so I would rather not check visibility.

mhsdesign:
I was unsure to do this in this pr or let this trait first be destroyed by someone who dislikes it :D

Actually i _did_ plan on using reflection as its super fast nowadays (faster than property mapping :D)

https://stackoverflow.com/questions/4160901/how-to-check-if-a-function-is-public-or-protected-in-php

```
Is Callable way: 0.0033209323883057ms
Reflection way: 0.0052220821380615ms
```

mhsdesign:
Aaand i really think that we should keep this change out of the main pr https://github.com/neos/flow-development-collection/pull/3232#issuecomment-1916592740 right now (also the class `SimpleActionController`) as this might otherwise get 90% of the discussion focus and the rest of that big pr is lost.

The introduction of `process(): Response` might be enough ^^ for one pr ^^

kitsunet:
> I was unsure to do this in this pr or let this trait first be destroyed by someone who dislikes it :D
> 
> Actually i _did_ plan on using reflection as its super fast nowadays (faster than property mapping :D)
> 
> https://stackoverflow.com/questions/4160901/how-to-check-if-a-function-is-public-or-protected-in-php
> 

Neat, well then, great, we can check callable and public in one go.

kitsunet:
We should/could use this also for the existing Controller?!

mhsdesign:
> We should/could use this also for the existing Controller?!

Yes sounds great. And im already glad this is an own pr 😅

mhsdesign:
We also discussed to keep the `*Action` pattern or not: https://github.com/neos/flow-development-collection/pull/3232#discussion_r1466758693 but it seems its here to stay :)

> do we want to keep the `*Action` pattern, because it still happens to me that i miss the suffix and it takes some time to figure out again what i missed 😅 , to fix this we could just improve the `routing:match` command or the like to emit a warning, or wait we cannot as this is highly dynamic and we have to have an actual request at hand.

> Mmm, yeah I thought about changing it, and I wouldn't mind that, for me I would actually say just make it an attribute? just allowing any method (without marker like an attribute) seems dangerous-ish especially if someone doesn't do proper sec policies. Also this then wouldn't match any existing policies as they all assume Action, so I just kept it that way.

PR #3295, opened 2024-01-30
TASK: Deprecate Flow's RequestInterface

mhsdesign:
Not to be confused with the psr requests interface.

TLTR:

Since the separation of http and cli requests the interface is completely unused. And strict types against its only implementation, the ActionRequest make it impossible to implement a custom request.

--------

_Initially_ the interface was introduced with https://github.com/neos/flow-development-collection/commit/2a269253c0345f6d2ec369d663214a6117df7a59 at beginning of time. At that time this was done probably to be able to type against the interface instead of the common `Request` class, that the Cli and Web Request both extended.

With the removal of many methods in its interface https://github.com/neos/flow-development-collection/commit/8c40ff2baf83158f6ce085e5b50c9b6400418fa6 and with the full separation of action and cli requests https://github.com/neos/flow-development-collection/pull/1552 the interface became useless.

The dispatcher and controller interface are both typed against the `ActionRequest` and so it's impossible to implement a custom request.

**Upgrade instructions**

Please reference the `ActionRequest` instead in typehints.

**Review instructions**

**Checklist**

- [ ] Code follows the PSR-2 coding style
- [ ] Tests have been created, run and adjusted as needed
- [ ] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [ ] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [ ] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked with `!!!` and have upgrade-instructions

PR #3301, opened 2024-02-02
!!! TASK: Remove dispatched state from ActionRequest

kitsunet:
> I will dig a bit into history as I guess something (that obviously no one cares about) was lost in the PSR HTTP refactoring. I think (as we can see in the dispatch loop) there was a possibility before for a controller to gain control but decide it does not finally dispatch the request and thus the dispatch loop would continue and try to find another controller to take care, so it was the controllers responsibility to set this, but I am not sure what conditions it previously attached to it being dispatched.

**Upgrade instructions**

**Review instructions**

**Checklist**


Neos.Flow/Classes/Mvc/RequestInterface.php
mhsdesign, line 1: the interface could alternatively be removed with https://github.com/neos/flow-development-collection/pull/3295 as i prepared over there already a nice description to the _why_ -> edit added it as separate commit

Neos.Flow/Classes/Mvc/Dispatcher.php
```diff
@@ -123,28 +123,29 @@ public function dispatch(ActionRequest $request): ActionResponse
...
+                return $response;
             } catch (StopActionException $stopActionException) {
                 $this->emitAfterControllerInvocation($request, $stopActionException->response, $controller);
                 $response = $stopActionException->response;
-                if (!$request->isMainRequest()) {
-                    $request = $request->getParentRequest();
+                if ($request->isMainRequest()) {
+                    return $response;
                 }
+                $request = $request->getParentRequest();
```
mhsdesign: as discussed we concluded that this logic might be faulty and actually do harm without the `isDispatched` in place.

background:

the dispatcher is either used for main request, in which case this logic is not needed,
or for subrequest like in a fusion plugin

scenario:

a subrequest, while being handled, throws a `StopActionException`.
which is catched here. We check if its a main request and get the main request which is now to be handled in the loop, like a forward.

Thesis:

Now it will actually render the outer response inside the plugin
But previously because the main request was dispatched === true, it would not be handled.

So in theory this is dead code, and now must be removed.

mhsdesign:
We further discussed that the `ActionReponse` doesnt need to be mutable anymore, and all its setters should be either deprecated or made internal.

One must use the `ForwardException` with attached `$nextRequest` instead to keep the dispatch loop going.

Neos.Flow/Classes/Mvc/Dispatcher.php
mhsdesign: Yes the thesis seems to be correct. Using `throwStatus` in a plugin will result in the main request rendering being started again an again until the memory is exhausted.

Removing this code will actually make the tests `dispatchIgnoresStopExceptionsForFirstLevelActionRequests` and `dispatchCatchesStopExceptionOfActionRequestsAndRollsBackToTheParentRequest` fail, so i wanted to research about the original intentions.
It seems that with the introduction https://github.com/neos/flow-development-collection/commit/ff050845e3659c08189026e374334b579865e32c#diff-450585af3bb27be2be268192637242195c5d6308512fd53dac9fa248ff319686 of sub request, this logic was introduced. (perviously all first level `StopActionException` where silently swallowed)

```php
} catch (\F3\FLOW3\MVC\Exception\StopActionException $stopActionException) {
    if ($request instanceof \F3\FLOW3\MVC\Web\SubRequest && $request->isDispatched()) {
        throw $stopActionException;
    }
}
```

mhsdesign: i removed the 

```php
if (!$request->isMainRequest()) {
    $request = $request->getParentRequest();
}
```

part in https://github.com/neos/flow-development-collection/pull/3301/commits/0ed5f28a0e574882c38e065d1bb66e5e4c407d1f ;)

Neos.Flow/Tests/Unit/Mvc/DispatcherTest.php
```diff
@@ -136,24 +132,13 @@ protected function setUp(): void
...
-        $this->dispatcher->dispatch($this->mockActionRequest);
-    }
-
     /**
      * @test
      */
     public function dispatchIgnoresStopExceptionsForFirstLevelActionRequests()
     {
-        $this->mockParentRequest->expects(self::exactly(2))->method('isDispatched')->willReturnOnConsecutiveCalls(false, true);
+        $this->markTestSkipped('TODO currently fails.');
```
mhsdesign: todo fix test

Neos.Flow/Tests/Unit/Mvc/DispatcherTest.php
```diff
@@ -166,12 +151,11 @@ public function dispatchIgnoresStopExceptionsForFirstLevelActionRequests()
      */
     public function dispatchCatchesStopExceptionOfActionRequestsAndRollsBackToTheParentRequest()
     {
-        $this->mockActionRequest->expects(self::atLeastOnce())->method('isDispatched')->willReturn(false);
-        $this->mockParentRequest->expects(self::atLeastOnce())->method('isDispatched')->willReturn(true);
+        $this->markTestSkipped('TODO currently not implemented to state what it says. Should fail.');
```
mhsdesign: todo fix test as well (we need to redefine this behaviour)

kitsunet: This is tricky, and might point to something we are breaking here that shouldn't be broken 🧐

kitsunet: `AndRollsBackToTheParentRequest` mmmm

mhsdesign: my commit description says following ... maybe it helps :D

**_TASK: Remove obsolete "RollsBackToTheParentRequest" logic in dispatcher_**

As discussed we concluded that this logic might be faulty and actually do harm without the `isDispatched` in place:

Background:

the dispatcher is either used for main request, in which case this logic is not needed, or for subrequest like in a fusion plugin

Scenario:

A subrequest, while being handled, throws a `StopActionException`.
which is caught here. We check if it's a main request and get the main request which is now to be handled in the loop, like a forward.

Thesis:

Now it will actually render the outer response inside the plugin
But previously because the main request was dispatched === true, it would not be handled.

Example:

Using `throwStatus` in a plugin will result in the main request rendering being started again and again until the memory is exhausted.

mhsdesign (approved):

PR #3305, opened 2024-02-04
Experiment; WIP: BUGFIX: bugfix/use-md5paths-instead-classnames-to-flush-correctly

### LINKED ISSUE #3303: BUG: Reflection cache for class is not flushed on file deletion

mhsdesign:
resolves #3303

but this is just an experiment. I was under the impression that we cannot find out the namespace of a removed php file, but we can do a reverse lookup by taking all possible psr-4 autoloading configs into account and then doing the flush. See https://github.com/neos/flow-development-collection/pull/3383

**Upgrade instructions**

**Review instructions**

**Checklist**


mhsdesign:
superseeded by https://github.com/neos/flow-development-collection/pull/3383

Issue #3317, opened 2024-02-14
BUG: It's not possible to configure a custom CacheFactory in Flow


Sebobo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The method used to access the configured CacheFactory implementation 
never returns a proper result as the raw configuration is accessed which
is split by package and the array access therefore cannot return a classname.

### Expected Behavior

It is possible to configure a CacheFactory via Objects.yaml and Flow uses it.

### Steps To Reproduce

Configure any custom CacheFactory via Objects.yaml.

### Environment

```markdown
- Flow: <= 9.0
- PHP: *
```

PR #3318, opened 2024-02-14
BUGFIX: Use configured cache factory

Sebobo:
The previously used method to select the configured CacheFactory class from the Objects.yaml never returned a result as the raw configuration split by package name was accessed, therefore the fallback was used and it was not possible to provide a custom CacheFactory.

Resolves: #3317

### DIFF
--- a/Neos.Flow/Classes/Cache/CacheManager.php
+++ b/Neos.Flow/Classes/Cache/CacheManager.php
@@ -11,6 +11,7 @@ namespace Neos\Flow\Cache;
  * source code.
  */
 
+use Neos\Cache\CacheFactoryInterface;
 use Neos\Flow\Annotations as Flow;
 use Neos\Cache\Backend\FileBackend;
 use Neos\Cache\Exception\DuplicateIdentifierException;
@@ -37,7 +38,7 @@ use Psr\Cache\CacheItemPoolInterface;
 class CacheManager
 {
     /**
-     * @var CacheFactory
+     * @var CacheFactoryInterface
      */
     protected $cacheFactory;
 
@@ -100,10 +101,10 @@ class CacheManager
     }
 
     /**
-     * @param CacheFactory $cacheFactory
+     * @param CacheFactoryInterface $cacheFactory
      * @return void
      */
-    public function injectCacheFactory(CacheFactory $cacheFactory): void
+    public function injectCacheFactory(CacheFactoryInterface $cacheFactory): void
     {
         $this->cacheFactory = $cacheFactory;
     }
--- a/Neos.Flow/Classes/Cache/CacheManager.php
+++ b/Neos.Flow/Classes/Cache/CacheManager.php
@@ -497,6 +497,7 @@ class CacheManager
         $backend = isset($this->cacheConfigurations[$identifier]['backend']) ? $this->cacheConfigurations[$identifier]['backend'] : $this->cacheConfigurations['Default']['backend'];
         $backendOptions = isset($this->cacheConfigurations[$identifier]['backendOptions']) ? $this->cacheConfigurations[$identifier]['backendOptions'] : $this->cacheConfigurations['Default']['backendOptions'];
         $persistent = isset($this->cacheConfigurations[$identifier]['persistent']) ? $this->cacheConfigurations[$identifier]['persistent'] : $this->cacheConfigurations['Default']['persistent'];
+        // @phpstan-ignore-next-line - $persistent is not yet part of the CacheFactoryInterface
         $cache = $this->cacheFactory->create($identifier, $frontend, $backend, $backendOptions, $persistent);
         $this->registerCache($cache, $persistent);
     }
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -377,8 +377,6 @@ class Scripts
         $cacheManager->injectEnvironment($environment);
         $cacheManager->injectCacheFactory($cacheFactory);
 
-        $cacheFactory->injectCacheManager($cacheManager);
-
         $bootstrap->setEarlyInstance(CacheManager::class, $cacheManager);
         $bootstrap->setEarlyInstance(CacheFactory::class, $cacheFactory);
     }
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -356,10 +356,15 @@ class Scripts
         $configurationManager = $bootstrap->getEarlyInstance(ConfigurationManager::class);
         $environment = $bootstrap->getEarlyInstance(Environment::class);
 
-        $cacheFactoryObjectConfiguration = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_OBJECTS, CacheFactoryInterface::class);
-        $cacheFactoryClass = isset($cacheFactoryObjectConfiguration['className']) ? $cacheFactoryObjectConfiguration['className'] : CacheFactory::class;
+        $cacheFactoryClass = null;
+        $cacheFactoryObjectConfiguration = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_OBJECTS);
+        foreach ($cacheFactoryObjectConfiguration as $objectConfiguration) {
+            if (isset($objectConfiguration[CacheFactoryInterface::class])) {
+                $cacheFactoryClass = $objectConfiguration[CacheFactoryInterface::class]['className'] ?? CacheFactory::class;
+            }
+        }
 
-        /** @var CacheFactory $cacheFactory */
+        /** @var CacheFactoryInterface $cacheFactory */
         $cacheFactory = new $cacheFactoryClass($bootstrap->getContext(), $environment, $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_SETTINGS, 'Neos.Flow.cache.applicationIdentifier'));
 
         $cacheManager = new CacheManager();
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -356,11 +356,14 @@ class Scripts
         $configurationManager = $bootstrap->getEarlyInstance(ConfigurationManager::class);
         $environment = $bootstrap->getEarlyInstance(Environment::class);
 
-        $cacheFactoryClass = null;
+        // Workaround to find the correct CacheFactory implementation at compile time.
+        // We can rely on the $objectConfiguration being ordered by the package names after their loading order.
+        // Normally this wiring would be done for proxy building a similar way, see ConfigurationBuilder.
+        $cacheFactoryClass = CacheFactory::class;
         $cacheFactoryObjectConfiguration = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_OBJECTS);
         foreach ($cacheFactoryObjectConfiguration as $objectConfiguration) {
-            if (isset($objectConfiguration[CacheFactoryInterface::class])) {
-                $cacheFactoryClass = $objectConfiguration[CacheFactoryInterface::class]['className'] ?? CacheFactory::class;
+            if (isset($objectConfiguration[CacheFactoryInterface::class]['className'])) {
+                $cacheFactoryClass = $objectConfiguration[CacheFactoryInterface::class]['className'];
             }
         }
 
--- a/Neos.Flow/Classes/Cache/CacheFactory.php
+++ b/Neos.Flow/Classes/Cache/CacheFactory.php
@@ -39,13 +39,6 @@ class CacheFactory extends \Neos\Cache\CacheFactory
     protected $context;
 
     /**
-     * A reference to the cache manager
-     *
-     * @var CacheManager
-     */
-    protected $cacheManager;
-
-    /**
      * @var Environment
      */
     protected $environment;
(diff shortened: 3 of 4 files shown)


Sebobo:
If there is a better way to access the merged config, I'm happy about a suggestion 🙂

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -356,10 +356,15 @@ public static function initializeCacheManagement(Bootstrap $bootstrap)
         $configurationManager = $bootstrap->getEarlyInstance(ConfigurationManager::class);
         $environment = $bootstrap->getEarlyInstance(Environment::class);
 
-        $cacheFactoryObjectConfiguration = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_OBJECTS, CacheFactoryInterface::class);
-        $cacheFactoryClass = isset($cacheFactoryObjectConfiguration['className']) ? $cacheFactoryObjectConfiguration['className'] : CacheFactory::class;
+        $cacheFactoryClass = null;
+        $cacheFactoryObjectConfiguration = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_OBJECTS);
+        foreach ($cacheFactoryObjectConfiguration as $objectConfiguration) {
+            if (isset($objectConfiguration[CacheFactoryInterface::class])) {
+                $cacheFactoryClass = $objectConfiguration[CacheFactoryInterface::class]['className'] ?? CacheFactory::class;
```
mhsdesign:
```suggestion
        // workaround to find the correct cache manager implementation at compile time.
        // we can rely on the $objectConfiguration being ordered by the package names after its loading order.
        // normally this wiring would be done for proxy building a similar way, see ConfigurationBuilder.
        $cacheFactoryClass = CacheFactory::class;
        $cacheFactoryObjectConfiguration = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_OBJECTS);
        foreach ($cacheFactoryObjectConfiguration as $objectConfiguration) {
            if (isset($objectConfiguration[CacheFactoryInterface::class]['className'])) {
                $cacheFactoryClass = $objectConfiguration[CacheFactoryInterface::class]['className'];
```

Neos.Flow/Classes/Cache/CacheManager.php
mhsdesign, line 1: ha it seems we do overload the interface here (`this->cacheFactory->create`), thats why phpstan (rightfully) complains.
The last argument `$persistent` is NOT part of the interface and now passed by good faith.
So either we extend the interface to include this param (which _would_ be breaking under normal circumstances, but id say currently not as the interface is until this change just useless).

Neos.Flow/Classes/Core/Booting/Scripts.php
bwaidelich: With Marc Henrys comment, this is a bit more clear to me but I still find this change suspicious. We iterate over the all configurations and override the `$cacheFactoryClass` variable whenever a `className` is specified!?

> // we can rely on the $objectConfiguration being ordered by the package names after its loading order.

Is there really order guarantee? and is it correct that a _later_ value overrides a former (and if not, why don't we break the loop upon match)?

This screams for a test, but it's probably not so easy to test – given it happens so early in the bootstrap!?

Sebobo: The configuration manager returns the configs in loading order, so a later appearing config overwrites the classname. Therefore I cannot break the loop early.
No clue right now on how to write a test for this change.

bwaidelich: Thanks for the clarification.
I might be able to help with the tests, but apart from those I feel that iterating over all object configurations for _every_ bootstrap is an overkill.. I hope that we can somehow access merged configuration or a find different way to configure the cache factory

mhsdesign: on the otherhand phps jit will probably do its magic. We loop over all flow packages that is not that much maybe 15?

bwaidelich: I'm more concerned about memory footprint than timing. For my (bare bones) Flow installation this is an array with 185 entries that is loaded into RAM.

Not sure if that's a problem.. But the whole thing makes me wonder: Why don't we merge object configurations like all the others?

Sebobo: This currently happens in the `CompileTimeObjectManager` 😔

The array is anyway loaded as it is used later on by the ObjectManager?

bwaidelich:
Some more little issues I found:
* [The Script calls `injectCacheManager()`](https://github.com/neos/flow-development-collection/blob/0ebc3bc1e9f71f1da0d3d8c2e759dc3c427ea1cf/Neos.Flow/Classes/Core/Booting/Scripts.php#L372) which is not defined in the interface
* [We set the early instance of the class instead of the factory](https://github.com/neos/flow-development-collection/blob/0ebc3bc1e9f71f1da0d3d8c2e759dc3c427ea1cf/Neos.Flow/Classes/Core/Booting/Scripts.php#L375)

Neos.Flow/Classes/Core/Booting/Scripts.php
mhsdesign: The array is structured by package i think to have a hint for debugging where what comes from

Sebobo:
> Some more little issues I found:
> 
> * [The Script calls `injectCacheManager()`](https://github.com/neos/flow-development-collection/blob/0ebc3bc1e9f71f1da0d3d8c2e759dc3c427ea1cf/Neos.Flow/Classes/Core/Booting/Scripts.php#L372) which is not defined in the interface
> * [We set the early instance of the class instead of the factory](https://github.com/neos/flow-development-collection/blob/0ebc3bc1e9f71f1da0d3d8c2e759dc3c427ea1cf/Neos.Flow/Classes/Core/Booting/Scripts.php#L375)

Thx for looking thoroughly into this! 

So what would be the proper solution for this? Modify the interface that was probably never used by anyone as it didn't work as expected or is there a safer way?

bwaidelich:
> So what would be the proper solution for this?

I don't know why the `CacheFactory` would need an instance of the `CacheManager` and it seems to be [unused](https://github.com/neos/flow-development-collection/blob/0ebc3bc1e9f71f1da0d3d8c2e759dc3c427ea1cf/Neos.Flow/Classes/Cache/CacheFactory.php#L46).
And I don't know why we have `\Neos\Flow\Cache\CacheFactory` and `\Neos\Cache\CacheFactory`
@kitsunet might know more since he

Neos.Flow/Classes/Core/Booting/Scripts.php
```diff
@@ -356,10 +356,15 @@ public static function initializeCacheManagement(Bootstrap $bootstrap)
         $configurationManager = $bootstrap->getEarlyInstance(ConfigurationManager::class);
         $environment = $bootstrap->getEarlyInstance(Environment::class);
 
-        $cacheFactoryObjectConfiguration = $configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_OBJECTS, CacheFactoryInterface::class);
```
kitsunet, line 359: Was that always the case? I mean why the hell do we separte the object configuration by package LOL. But that's for another PR / ticket.

Neos.Flow/Classes/Core/Booting/Scripts.php
kitsunet: Was that always the case? I mean why the hell do we separate the object configuration by package LOL. But that's for another PR / ticket

kitsunet: Note: Yes was always the case, how unexpected 🤷 My bad. In that case I guess objects configuration is just bad for anything that needs to perform and we should switch to do this via settings. but also how the hell do we work with this in proxy building and such, why is it not merged aaahhhhhh

kitsunet:
The factory is needed to create the actual Cache instance, see `\Neos\Flow\Cache\CacheManager::createCache` 

Manager vs. Factory are the separating high level layers between the external package (has a factory) and the flow syntactics "injected" via the manager which takes care of the environment that abstracts the - flow specfiic - application context away and deals with settings.

Neos.Flow/Classes/Core/Booting/Scripts.php
kitsunet: the comment is wrong btw. !!! this muddies the waters we are searching for a cache FACTORY implementation as to potentially allow other implementations outside of neos/cache (our flow independent package). the Manager is "the manager", there should only be that specific one for flow. it could be final even in a future iteration.

`// workaround to find the correct cache manager implementation at compile time.`

->
`
// workaround to find the correct cache factory implementation at compile time.`

kitsunet: So with the error below, that's a nasty bit, we overwrite the factory implementation in Flow, I forgot about that part and the cacheManaegr is indeed not used in this overwritten factory, not sure why it exists in the first place... Bad decisions from the past come haunting LOL. I think what we do there in the constructor should just move to the manager and we should get rid of this implementation that implements stuff on top of the interface and thus breaks the static analysis below.

bwaidelich:
> The factory is needed to create the actual Cache instance, see \Neos\Flow\Cache\CacheManager::createCache

@kitsunet my question was, why the `CacheFactory` needs the `CacheManager` though. It seems not to be used.

kitsunet:
yep sorry, too many different comments @bwaidelich see in code comment about that. I agree it should not need it and it doesn't even use it.

Sebobo:
Thanks for the suggestions, I adjusted the code and now only the `$oersistent` interface mismatch is left to solve.

kdambekalns (commented):
Looks good to me – how do we proceed with the `$persistent` question?

bwaidelich:
Can we should add the `$persistent` argument to the [CacheFactoryInterface](https://github.com/neos/cache/blob/8.3/Classes/CacheFactoryInterface.php):
```php
public function create(string $cacheIdentifier, string $cacheObjectName, string $backendObjectName, array $backendOptions = []): FrontendInterface;
```
to
```php
public function create(string $cacheIdentifier, string $cacheObjectName, string $backendObjectName, array $backendOptions = [], bool $persistent = false): FrontendInterface;
```

It's a bit nasty because it adds this "persistent"-quirk to the `neos/cache` Package, but the other options don't seem to work out:
* We can't add a new marker interface because it would be incompatible with the `CacheFactoryInterface`
* We could check for the actual implementation as in `$cache = $this->cacheFactory instanceof FlowCacheFactory ? $this->cacheFactory->create($identifier, $frontend, $backend, $backendOptions, $persistent) : $this->cacheFactory->create($identifier, $frontend, $backend, $backendOptions);` but that would pose a risk of new implementations not to support persistent caches

mhsdesign:
or we ignore this error via phpstan :D

I am more critical about the ram usage as noted by bastian: https://github.com/neos/flow-development-collection/pull/3318#discussion_r1489220453

kdambekalns:
> I am more critical about the ram usage

That's for compile time, no? I'd say that's not really an issue here…

kitsunet:
We just need to separate key/value store abstraction (aka persistent cache) and actual cache and that interface problem resolves, so I wouldn't worry too much about it. 

What's that about RAM usage? Is the respective comment deleted? I can't find mention of ram usage.

mhsdesign:
updated the link but

> I'm more concerned about memory footprint than timing. For my (bare bones) Flow installation this is an array with 185 entries that is loaded into RAM.

> Not sure if that's a problem.. But the whole thing makes me wonder: Why don't we merge object configurations like all the others?

so we should definitely open a followup issue and move the discussion how we can fix the merged config to make think a little better. Possibly this can be fixed for 8.3 but at least safely for 9.0

kitsunet:
We could... static compile that value 🤷 

Wait no, we cannot, because where. Unless we uhhh, make a little class that basically encapsulates what happens in Scripts and make that compile static. a CacheFactoryFactory kinda LOL

mhsdesign:
Otherwise it would be a bit against all the microoptimisation DNA of flow... not saying that for example we always need them but that would have been no question to static compile ten years ago i think?
Im not really that deep into performance and memory and live the dangerous live of just getting a faster machine if its slow (at least i use a lot of value objects and annoy my garbage collector) :D But i guess having an eye on these think especially for the framework is a good thing. But on the other-hand all those probably thought out optimisations _could_ turn out to be useless because php just does things differently faster than expected and it might make things just worse or slower -> ala premature optimization is the root of all evil.  So in the end what we really would need is some baseline to detect performance and memory usage detection but idk how reliable that would be and its def out of scope of this pr that just poke into a beehive.

kitsunet:
I dislike anything that happens for EVERY single call, and this is one of those things, if we can ake it any better we should.

Sebobo:
Keep in mind, my change only adds the loop, not the config retrieval. That was already loaded in the previous version.

mhsdesign:
okay very well thats true. is there a way to faster loop in reverse over the entries? As we _could_ break directly when we find a declaration as the loading order is given.

bwaidelich:
> Keep in mind, my change only adds the loop, not the config retrieval

Ah, good point – Maybe we can improve that in a separate PR so that we don't block your change!

kitsunet (approved):
what basti said, therefore approving as this generally is the right direction

bwaidelich:
the static analysis is still failing though (I'd be fine with some `@phpstan-ignore` comments for now, if we create an issue to improve that persistent-supporting-caching-interface-thingy)

Sebobo:
OK I now ignore the warning. PHPStan still fails due to an unrelated unchanged file.

mhsdesign (approved):

mhsdesign:
@kitsunet actually found _the_ way so please review the followup :D https://github.com/neos/flow-development-collection/pull/3322

PR #3319, opened 2024-02-15
BUGFIX: Normalize source type for converting floats to ValueObjects

paxuclus:
This change allows the ScalarTypeToObjectConverter to convert from floats to ValueObjects. The type converter uses `gettype()` which will return `double` for `floats`. As the constructor argument type of the ValueObject is passed through `TypeHandling::normalizeType()`, it will always be `float` and never `double` which will lead to floats not being converted by this type converter.

**Checklist**


mhsdesign (approved):
Thank you so much for your contribution ;)

mhsdesign:
... Also thanks for the test ;) ❤️ 

The phpstan failure is unrelated, will update 8.3 accordingly

PR #3322, opened 2024-02-21
TASK: Use `getClassNameByObjectName` instead of manually traversing `…

mhsdesign:
…CONFIGURATION_TYPE_OBJECTS`

We were to blind to see it in https://github.com/neos/flow-development-collection/pull/3318 but thankfully christian noticed it the last second:

> For the respective use case we already HAVE a way better way to do this, namely:

> `\Neos\Flow\ObjectManagement\ObjectManagerInterface::getClassNameByObjectName()`, this should just return the _right_ implementation as in, the chosen one the objectmanager would inject if you asked it, and that's all we need and we have an object manager, at least compiletime which also can answer this question.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -356,16 +356,7 @@ class Scripts
         $configurationManager = $bootstrap->getEarlyInstance(ConfigurationManager::class);
         $environment = $bootstrap->getEarlyInstance(Environment::class);
 
-        // Workaround to find the correct CacheFactory implementation at compile time.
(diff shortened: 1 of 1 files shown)


mhsdesign:
@kitsunet sadly not:

> The Object Manager is not available at this stage of the bootstrap run.

Sebobo:
Jepp, doesn't work.

kitsunet:
That is disappointing :/
Right because the ObjectManager needs the Flow_Object_Configuration cache which contains exactly the information we need 🐓🥚

kitsunet:
Nothing we can fix short term I guess :/

PR #3324, opened 2024-03-03
FEATURE: Add `Flow\Policy` Attribute/Annotation

### LINKED ISSUE #2060: Allow to specify policies per action through annotations

mficzel:
The `Flow\Policy` attribute allows to assign the required policies (mostly roles) directly on the affected method. 
This allows to avoid dealing with Policy.yaml in projects in simple cases where is sometimes is annoying to look up the exact syntax for that.

Hint: While this is a very convenient way to add policies in project code it should not be used in libraries/packages that expect to be configured for the outside. In such cases the policy.yaml is still preferred as it is easier to overwrite. Also Policy yaml offers features like parameter filters that are not available via annotations.

Usage:

```php
use Neos\Flow\Security\Authorization\Privilege\PrivilegeInterface;

class ExampleController extends ActionController
{
    /**
     * By assigning a policy with a role argument access to the method is granted to the specified role
     */
    #[Flow\Policy(role: 'Neos.Flow:Everybody')]
    public function everybodyAction(): void
    {
    }

    /**
     * By specifying the permission in addition and the DENY and ABSTAIN can be configured aswell
     * Flow\Policy attributes can be assigned multiple times if multiple roles are to be configured
     */
    #[Flow\Policy(role: 'Neos.Flow:Administrator', permission: PrivilegeInterface::GRANT)]
    #[Flow\Policy(role: 'Neos.Flow:Anonymous', permission: PrivilegeInterface::DENY)]
    public function adminButNotAnonymousAction(): void
    {
    }
}
```

The package: `Meteko.PolicyAnnotation` by @sorenmalling implemented the same ideas earlier.

Resolves: #2060

**Upgrade instructions**

**Review instructions**

See: #3325 resolving #2059, both solutions should work hand in hand

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Annotations/Policy.php
+++ b/Neos.Flow/Classes/Annotations/Policy.php
@@ -1,0 +1,41 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Flow\Annotations;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
+use Neos\Flow\Security\Authorization\Privilege\PrivilegeInterface;
+
+/**
+ * Adds a policy configuration to a method
+ *
+ * This is a convenient way to add policies in project code
+ * but should not be used in libraries/packages that shall be
+ * configured for different use cases.
+ *
+ * @Annotation
+ * @NamedArgumentConstructor
+ * @Target({"METHOD"})
+ */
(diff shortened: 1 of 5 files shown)


mficzel:
A better name for the `PolicyAnnotationService` would be appreciated

mficzel:
Just discovered that the signal in the policy service is called to late for any policies that require proxy building. Will try to refactor this to the PolicyConfigurationLoader that seems to be called in time.

Strange is that the roles and policies show up in the `./flow security:*` commands without beeing effective.

PR #3311, opened 2024-02-06
!!! FEATURE: Dispatcher psr overhaul

mhsdesign:
To ease the implementation of a custom controller now use a direct pattern of `f(ActionRequest) => PsrResponseInterface`.
This is breaking if you implemented your own `ControllerInterface` or overwrote low level parts and methods of the `ActionController`.

```php
class Dispatcher
{
    public function dispatch(ActionRequest $request): ResponseInterface;
}

interface ControllerInterface
{
    public function processRequest(ActionRequest $request): ResponseInterface;
}
```

Also, it's discouraged to manipulate `$this->response` in controllers (the `ActionResponse` is deprecated), although it's still supported in actions for now, please consider returning a new response instead.

<details>

<summary>
Explanation of the legacy MVC response object.
</summary>

Previously Flows MVC needed a single mutable response which was passed from dispatcher to controllers
and even further to the view and other places via the controller context: `ControllerContext::getResponse()`.
This allowed to manipulate the response at every place.
With the dispatcher and controllers now directly returning a response, the mutability is no longer required.
Additionally, the abstraction offers naturally nothing, that cant be archived by the psr response,
as it directly translates to one: `ActionResponse::buildHttpResponse()`
So you can and should use the immutable psr {@see ResponseInterface} instead where-ever possible.
For backwards compatibility, each controller will might now manage an own instance of the action response
via `$this->response` `AbstractController::$response` and pass it along to places.
But this behaviour is deprecated!
Instead, you can directly return a PSR repose `\GuzzleHttp\Psr7\Response` from a controller:

```php
public function myAction()
{
    return (new Response(status: 200, body: $output))
        ->withAddedHeader('X-My-Header', 'foo');
}
```

</details>

Further the `ForwardException` does not extend the `StopActionException` anymore, meaning a try catch must be adjusted.

This is the main PR and it contains
- https://github.com/neos/flow-development-collection/pull/3232
- https://github.com/neos/flow-development-collection/pull/3294

Followup: https://github.com/neos/flow-development-collection/pull/3301
Resolves: https://github.com/neos/flow-development-collection/issues/3289

This change needs an accompanying adjustment to Neos to adjust the
PluginImplementation as well as Modules: https://github.com/neos/neos-development-collection/pull/4738

**Upgrade instructions**

WIP Upgrade notes: https://github.com/neos/flow-development-collection/pull/3232#issuecomment-1913166370

**Review instructions**

**Checklist**

- [ ] Code follows the PSR-2 coding style
[...]

PR #3325, opened 2024-03-03
FEATURE: Add `Flow\Route` Attribute/Annotation

### LINKED ISSUE #2059: Support defining routes on the actions via annotation

mficzel:
The `Flow\Route` attribute allows to define routes directly on the affected method. This allows to avoid dealing with Routes.yaml in projects in simple cases where is sometimes is annoying to look up the exact syntax for that.

Usage:

```php

class ExampleController extends ActionController
{
    #[Flow\Route(uriPattern:'my/path', httpMethods: ['GET'])]
    public function someAction(): void
    {
    }

    #[Flow\Route(uriPattern:'my/other/b-path', defaults: ['test' => 'b'])]
    #[Flow\Route(uriPattern:'my/other/c-path', defaults: ['test' => 'c'])]
    public function otherAction(string $test): void
    {
    }
}
```

To use annotation routes packages have to register the `AttributeRoutesProviderFactory` in `Neos.Flow.mvc.routes` with Controller classNames or patterns.

Settings.yaml:
```yaml
Neos:
  Flow:
    mvc:
      routes:
        Vendor.Example.attributes:
          position: 'before Neos.Neos'
          providerFactory: \Neos\Flow\Mvc\Routing\AttributeRoutesProviderFactory
          providerOptions:
            classNames:
              - Vendor\Example\Controller\ExampleController
```

This pr also adds the general option to register `provider` and `providerOptions` in  the Setting `Neos.Flow.mvc.routes` which was required obviously.

The package: `WebSupply.RouteAnnotation` by @sorenmalling implemented similar ideas earlier.

Resolves: #2059

**Upgrade instructions**

**Review instructions**

Alsow see: #3324 resolving #2060, both solutions ideally would work hand in hand

**Checklist**


mficzel:
This one needs a little discussion:

1. Shall we expose `appendExceedingArguments`
2. Shall we expose `cachePeriod` and `cacheTags`
3. Is it problematic that defaults could overwrite @package etc

Neos.Flow/Classes/Mvc/Routing/AnnotationRoutesProvider.php
```diff
@@ -0,0 +1,85 @@
...
+                foreach ($annotations as $annotation) {
+                    if ($annotation instanceof Flow\Route) {
+                        $configuration = [
+                            'uriPattern' => $annotation->uriPattern,
+                            'defaults' => Arrays::arrayMergeRecursiveOverrule(
+                                [
+                                    '@package' => $controllerPackageKey,
+                                    '@subpackage' => $subPackage,
+                                    '@controller' => substr($controllerName, 0, -10),
+                                    '@action' => substr($methodName, 0, -6),
```
mhsdesign: lets discuss this `ActionController` magic ;)

edit: see proposed behaviour in https://github.com/neos/flow-development-collection/pull/3325#issuecomment-2016594513

bwaidelich:
FYI: Related discussion: https://discuss.neos.io/t/rfc-future-of-routing-mvc-in-flow/6535

mficzel:
@mhsdesign @bwaidelich i am a bit frustraded how this turns out. While i like the suggestions on discuss the approach "we have to rewrite routing first" basically excludes this from beeing in Neos 9 (this close to a release we should not introduce any big changes we do not strictly need) which means we are stuck with Routes.yaml and in extension Policy.yaml for 9.0

I also see no easy way to only include part 1 of the discuss post (configure route provideres in seettings `mvc.routes`) into the RouteConfiguration that was just refactored in #2970. If you see a clear way to include this i might work on this but i do'nt short of refactoring again which i have no motivation to atm.

Regarding binding the current implementation to `Action` i tend to not see this as a real problem as this is how routing works now and it makes sense to implement exactly that now. Once routing gets more smart we can adjust that but even then we will surely want to support ActionControllers even if we allow other alternatives. So i see no way the Route Annotation would become wrong in future, only the way the attributes are handled would have to change together with routing.

bwaidelich:
@mficzel I'm sorry that you're frustrated. But I think this is due to a misunderstanding:
I'm in no way trying to block or slow down this PR – I just linked the writeup of the things we discussed yesterday because they are related.

IMO there is no reason not to proceed with this one, we can always refactor it to use a different integration mechanism lateron.

There's only two issues that we should tackle/decide upon IMO:

* from a DX perspective, I would love to have the routing cache issue fixed (i.e. that routing caches are flushed when a route attribute is changed – or is that the case already?)
* yesterday we were in favor of keeping the explicit routing activation & ordering via Settings, and this change would circumvent this in the current form AFAIS

> I also see no easy way to only include part 1 of the discuss post (configure route provideres in seettings mvc.routes) into the RouteConfiguration

I think it would actually be pretty straightforward to skip the `CombinedRoutesProvider` and instead support a setting of
```yaml
provider: '<providerClassName>'
```
in our default `ConfigurationRoutesProvider`

mficzel:
@bwaidelich the handling of the whole `mvc.routes` setting currently happens before the combined settings are passed to the ConfigurationRoutesProvider ... that is why there is currently no way to get in between at a specific point. 

Also specifying a provider via setting is probably not that trivial as one will have to pass at leas the package key and for that either allow options or pass the key from the configuration in any case. Imho it would make sense to just include searching for route attributes into what happens when you add a package key to `mvc.routes`

bwaidelich:
Let's talk, I have some ideas and I think that they can solve the remaining issues without a lot of work

mficzel:
Regarding Cache flushing i added this class to the annotationCacheFlusher not sure this actually works as i could not find prior uses.

bwaidelich:
@mficzel A little bit hacky, but with the following change:

```diff
diff --git a/Neos.Flow/Classes/Configuration/Loader/RoutesLoader.php b/Neos.Flow/Classes/Configuration/Loader/RoutesLoader.php
index a02a5c152..8b2229c31 100644
--- a/Neos.Flow/Classes/Configuration/Loader/RoutesLoader.php
+++ b/Neos.Flow/Classes/Configuration/Loader/RoutesLoader.php
@@ -96,6 +96,13 @@ class RoutesLoader implements LoaderInterface
             if (isset($routeFromSettings['suffix'])) {
                 $subRoutesConfiguration['suffix'] = $routeFromSettings['suffix'];
             }
+            if (isset($routeFromSettings['provider'])) {
+                $routeDefinitions[] = [
+                    'name' => $packageKey,
+                    'provider' => $routeFromSettings['provider'],
+                ];
+                continue;
+            }
             $routeDefinitions[] = [
                 'name' => $packageKey,
                 'uriPattern' => '<' . $subRoutesName . '>',
@@ -128,6 +135,10 @@ class RoutesLoader implements LoaderInterface
             }
             $mergedSubRoutesConfiguration = [$routeConfiguration];
             foreach ($routeConfiguration['subRoutes'] as $subRouteKey => $subRouteOptions) {
+                if (isset($subRouteOptions['provider'])) {
+                    $mergedRoutesConfiguration[] = $subRouteOptions;
+                    continue;
+                }
                 if (!isset($subRouteOptions['package'])) {
                     throw new ParseErrorException(sprintf('Missing package configuration for SubRoute in Route "%s".', ($routeConfiguration['name'] ?? 'unnamed Route')), 1318414040);
                 }
diff --git a/Neos.Flow/Classes/Mvc/Routing/ConfigurationRoutesProvider.php b/Neos.Flow/Classes/Mvc/Routing/ConfigurationRoutesProvider.php
index f876008e8..a4567c559 100644
--- a/Neos.Flow/Classes/Mvc/Routing/ConfigurationRoutesProvider.php
+++ b/Neos.Flow/Classes/Mvc/Routing/ConfigurationRoutesProvider.php
@@ -6,6 +6,8 @@ namespace Neos\Flow\Mvc\Routing;
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Configuration\ConfigurationManager;
+use Neos\Flow\Configuration\Loader\RoutesLoader;
+use Neos\Flow\ObjectManagement\ObjectManagerInterface;
 
 /**
  * @Flow\Scope("singleton")
@@ -15,13 +17,26 @@ final class ConfigurationRoutesProvider implements RoutesProviderInterface
     private ConfigurationManager $configurationManager;
 
     public function __construct(
-        ConfigurationManager $configurationManager
+        ConfigurationManager $configurationManager,
+        private ObjectManagerInterface $objectManager,
     ) {
         $this->configurationManager = $configurationManager;
     }
 
     public function getRoutes(): Routes
     {
-        return Routes::fromConfiguration($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES));
+        $routes = [];
+        foreach ($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES) as $routeConfiguration) {
+            if (isset($routeConfiguration['provider'])) {
+                $provider = $this->objectManager->get($routeConfiguration['provider']);
+                assert($provider instanceof RoutesProviderInterface);
+                foreach ($provider->getRoutes() as $route) {
+                    $routes[] = $route;
+                }
+                continue;
+            }
+            $routes[] = Route::fromConfiguration($routeConfiguration);
+        }
+        return Routes::create(...$routes);
     }
 }
```

I can register custom providers via

```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Foo':
          provider: 'Some\Package\SomeRoutesProvider'
          position: 'start'
```

we could also add `providerOptions`

bwaidelich:
Re your discussion points from [2 weeks ago](https://github.com/neos/flow-development-collection/pull/3325#issuecomment-1975255985):

> Shall we expose appendExceedingArguments

I would say no because they are too easily confused with query parameters. We could always add it lateron

> Shall we expose cachePeriod and cacheTags

Personally I don't really care, but IMO there is no reason _not_ to (but it could always be added lateron)

> Is it problematic that defaults could overwrite @Package etc

That would be solved with an explicit activation like suggested above


ps: if you force push changes it makes it really hard to follow the progress

mficzel:
@bwaidelich it works now but is ugly as hell for now ... moved most of the logic of routes configuration loader into the ConfigurationRoutesProvider and integrated the annotation handling there. Have to figure out how to seperate and test this later. 

Also i think that adding "pathPrefix" and "additionalRouteProviders" to the Setting `mvc.routes` can and should be be done separately and that flow route annotations should be used for all packages that are configured in `mvc.routes` without configuring an extra route provider.

Neos.Flow/Classes/Mvc/Routing/ConfigurationRoutesProvider.php
```diff
@@ -6,22 +6,37 @@
...
     {
-        return Routes::fromConfiguration($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES));
+        $routes = [];
+        foreach ($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES) as $routeConfiguration) {
+            if (isset($routeConfiguration['provider'])) {
+                $provider = $this->objectManager->get($routeConfiguration['provider']);
+                if ($provider instanceof RoutesProviderWithOptionsInterface) {
+                    $provider = $provider->withOptions($routeConfiguration['providerOptions']);
+                }
+                assert($provider instanceof RoutesProviderInterface);
```
bwaidelich: I know we do this in other places too (so I'd be perfectly fine with it) but just to bring it up: This could lead to weird errors when failed in production because `assert` usually is disabled there

bwaidelich (commented):
~Just one mini nitpick that we can safely ignore IMO. Apart from that +1 by reading.~ (*edit:* I found some more issues I would like to bring up, see comment below)

But we should add some docs & tests with this PR, I'll gladly help with those!

mficzel:
@bwaidelich agree and will add docs after cleaning up and fixing the tests. Ran into an issue with the partner policy pr #3324 that has to be solved first as we should release those together if possible.

Neos.Flow/Classes/Mvc/Routing/RouteAnnotationRoutesProvider.php
```diff
@@ -0,0 +1,107 @@
...
+namespace Neos\Flow\Mvc\Routing;
+
+use Neos\Flow\Mvc\Exception\InvalidActionNameException;
+use Neos\Flow\Mvc\Routing\Exception\InvalidControllerException;
+use Neos\Flow\ObjectManagement\ObjectManagerInterface;
+use Neos\Flow\Reflection\ReflectionService;
+use Neos\Flow\Annotations as Flow;
+use Neos\Utility\Arrays;
+
+class RouteAnnotationRoutesProvider implements RoutesProviderWithOptionsInterface
```
bwaidelich: Shouldn't it be `RouteAttributeRoutesProvider` or even just `AttributeRoutesProvider`? and it could be `final` since there is no method to override

Neos.Flow/Classes/Mvc/Routing/RouteAnnotationRoutesProvider.php
```diff
@@ -0,0 +1,107 @@
...
+        public readonly ObjectManagerInterface $objectManager,
+        public readonly array $classNames = [],
+    ) {
+    }
+
+    /**
+     * @param array<string, mixed> $options
+     * @return $this
+     */
+    public function withOptions(array $options): static
```
bwaidelich: This is a pragmatic approach that works due to our crazy dependency injection that can deal with prototype injection. But it can lead to weird (caching) issues if the object is called before it is "optionized" and IMHO a factory pattern would be the more explicit way to solve this, i.e.:

```php
interface RoutesProviderFactoryInterface {
    public function create(array $options): RoutesProviderInterface
}

#[Flow\Scope('singleton')]
final class AttributeRoutesProviderFactory implements RoutesProviderFactoryInterface {
    public function __construct(
        public readonly ReflectionService $reflectionService,
        public readonly ObjectManagerInterface $objectManager,
    ) {}

    public function create(array $options): RouteAnnotationRoutesProvider {
        return new RouteAnnotationRoutesProvider($this->reflectionService, $this->objectManager, $options['classNames']);
    }
}

final class RouteAnnotationRoutesProvider implements RoutesProviderInterface {
    public function __construct(
        public readonly ReflectionService $reflectionService,
        public readonly ObjectManagerInterface $objectManager,
        public readonly array $classNames,
    ) {}
    
    // ...
}
```

as a consequence we should specify the *factory*  instead of the *provider*  class in the settings:

```yaml
providerFactory: Some\Namespace\AttributeRoutesProviderFactory
providerOptions:
  foo: bar
```

That might seem cumbersome but we use that pattern a lot in Neos 9 and IMO it works very well

bwaidelich (commented):
Upon second pass and reading your comment I stumbled upon some more details

@mficzel thanks for taking care!

> it works now but is ugly as hell for now ...

Which part do you think is ugly?
The implicit nesting of `RoutesProvider` is maybe a little weird, but I don't think that the code is overly complex or ugly

> Also i think that adding "pathPrefix" and "additionalRouteProviders" to the Setting mvc.routes can and should be be done separately

Yes, I totally agree!

> that flow route an notations should be used for all packages that are configured in mvc.routes without configuring an extra route provider.

I see that this is more convenient. But it also introduces this magic implicit ordering issue.
I thought we agreed that routes should always be registered explicitly.

And IMO that would mean, that if you want to mix `Routes.yaml`-routes and attribute-driven routes in the _same_ package, you would need to add two entries in your settings and I think that's fine because it's an unlikely scenario.

Now having to opt in for all controller class names seems overly complex indeed for those common cases:

```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Some.Package': true
        'Some.Package:fromAttributes':
          handler: 'Neos\Flow\Mvc\Routing\RouteAnnotationRoutesProvider'
          handlerOptions:
            classNames:
              - 'Some\Package\Controller\MyController'
              - 'Some\Package\Controller\MyOtherController'
```

So I would suggest, that the minimal configuration for this scenario should look like this:

```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Some.Package': true
        'Some.Package:fromAttributes':
          handlerFactory: 'Neos\Flow\Mvc\Routing\AttributeRoutesProviderFactory'
          handlerOptions:
              packageKey: 'Some.Package'
```

Potentially with some magic sprinkled in (I don't think it's helpful but I would be fine with):

```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Some.Package': true
        'Some.Package:fromAttributes':
          handlerFactory: 'Neos.Flow:AttributeRoutesProvider'
```
...which would resolve the handler factory FQN to `<Package.Key>/Routing/<Name>Factory` and extract the package key from the settings key

mficzel:
@bwaidelich yes we agreed that package routes are always configured via `mvc.routes`. The idea to include both routes.yaml and annotation routes via `mvc.routes` was before that and is replaced by now. It totally makes sense to only include Package/Routes.yaml or annotation routes which is ensured by configuring a routes provider. 

The code is imho fine but testing needs a little work now to catch up to those changes.

Neos.Flow/Classes/Mvc/Routing/ConfigurationRoutesProvider.php
mficzel: Refactored to \InvalidArgumentException

Neos.Flow/Classes/Mvc/Routing/RouteAnnotationRoutesProvider.php
mficzel: Will adjust that tomorrow ... that will avoid the RouteProviderWithOptions problem that is weird aswell ... i just wanted to avoid builderFactoryGeneratorProviders

Neos.Flow/Classes/Mvc/Routing/RouteAnnotationRoutesProvider.php
mficzel: Agreed and adjusted

Neos.Flow/Classes/Mvc/Routing/RouteAnnotationRoutesProvider.php
mficzel: Slowly getting the hang of the providerFactory pattern ... guess it is mostly a thing to get used to. Is adjusted as requested

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1273,7 +1273,7 @@ protected function reflectClassMethod(string $className, MethodReflection $metho
             if (!isset($this->classesByMethodAnnotations[$annotationClassName][$className])) {
                 $this->classesByMethodAnnotations[$annotationClassName][$className] = [];
             }
-            $this->classesByMethodAnnotations[$annotationClassName][$className][] = $methodName;
+            $this->classesByMethodAnnotations[$annotationClassName][$className][$methodName] = $methodName;
```
mficzel, line 1276: This change ensures that methods with multiple annotations do not appear multiple times in methodsAnnotatedWith ... pretty sure this ha no unwanted side effects

Neos.Flow/Classes/Configuration/Loader/RoutesLoader.php
```diff
@@ -128,6 +136,10 @@ protected function mergeRoutesWithSubRoutes(array $packages, ApplicationContext
             }
             $mergedSubRoutesConfiguration = [$routeConfiguration];
             foreach ($routeConfiguration['subRoutes'] as $subRouteKey => $subRouteOptions) {
+                if (isset($subRouteOptions['provider'])) {
```
bwaidelich:
```suggestion
                if (isset($subRouteOptions['providerFactory'])) {
```

Neos.Flow/Classes/Mvc/Routing/ConfigurationRoutesProvider.php
```diff
@@ -6,22 +6,36 @@
...
 
     public function getRoutes(): Routes
     {
-        return Routes::fromConfiguration($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES));
+        $routes = [];
+        foreach ($this->configurationManager->getConfiguration(ConfigurationManager::CONFIGURATION_TYPE_ROUTES) as $routeConfiguration) {
+            if (isset($routeConfiguration['providerFactory'])) {
+                $providerFactory = $this->objectManager->get($routeConfiguration['providerFactory']);
+                if (!$providerFactory instanceof RoutesProviderFactoryInterface) {
+                    throw new \InvalidArgumentException(sprintf('The configured route providerFactory "%s" does not implement the "%s"', $routeConfiguration['provider'], RoutesProviderFactoryInterface::class), 1710784630);
```
bwaidelich:
```suggestion
                    throw new \InvalidArgumentException(sprintf('The configured route providerFactory "%s" does not implement the "%s"', $routeConfiguration['providerFactory'], RoutesProviderFactoryInterface::class), 1710784630);
```

bwaidelich (commented):
Lovely! Two little fixes (please double check, done from my phone *g)

Neos.Flow/Classes/Mvc/Routing/RouteAnnotationRoutesProvider.php
bwaidelich: That's great. And it will allow us to extend the feature just by tweaking the factory.
For example it could support an option 'packageKey' and translate it to class names without having to change the provider itself

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
```diff
@@ -0,0 +1,102 @@
...
+
+final class AttributeRoutesProvider implements RoutesProviderInterface
+{
+    /**
+     * @param array<string> $classNames
+     */
+    public function __construct(
+        public readonly ReflectionService $reflectionService,
+        public readonly ObjectManagerInterface $objectManager,
+        public readonly array $classNames = [],
```
bwaidelich: With the factory pattern we don't need to use this optional param
```suggestion
        public readonly array $classNames,
```
I would even suggest to test against empty array in the constructor and fail in that case

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
```diff
@@ -0,0 +1,102 @@
...
+                            'defaults' => Arrays::arrayMergeRecursiveOverrule(
+                                [
+                                    '@package' => $controllerPackageKey,
+                                    '@subpackage' => $subPackage,
+                                    '@controller' => $controller,
+                                    '@action' => $action,
+                                    '@format' => 'html'
+                                ],
+                                $annotation->defaults ?? []
+                            )
```
mhsdesign, line 140: martin:

> Is it problematic that defaults could overwrite @Package etc

yes id say so, we should probably reverse the order and use a normal `array_merge` as we dont need to merge nested things.

Or even better if we throw an exception if we encounter `@package` in `$annotation->defaults`

... and i dont understand bastis point ^^

> That would be solved with an explicit activation like suggested above

Neos.Flow/Classes/Mvc/Routing/RouteAnnotationRoutesProvider.php
mhsdesign: Thanks @mficzel for keeping up with all the todos :D

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,39 @@
...
+ * @Target({"METHOD"})
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class Route
+{
+    public function __construct(
+        public readonly string $uriPattern,
+        public readonly ?string $name = null,
+        public readonly ?array $httpMethods = null,
+        public readonly ?array $defaults = null,
```
mhsdesign: Why do we actually need to expose the `defaults`  here? i cant think of a usecase that could be solved here? Or wait, we allow full functionality and interpolation in routes right?  `{banane}` but we might need to limit the feature set, as `{@action}` makes no sense for example

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        $routes = [];
+        $annotatedClasses = $this->reflectionService->getClassesContainingMethodsAnnotatedWith(Flow\Route::class);
+
+        foreach ($annotatedClasses as $className) {
+            $includeClassName = false;
+            foreach ($this->classNames as $classNamePattern) {
+                if (fnmatch($classNamePattern, $className, FNM_NOESCAPE)) {
+                    $includeClassName = true;
+                }
```
mhsdesign:
```suggestion
                if (fnmatch($classNamePattern, $className, FNM_NOESCAPE)) {
                    $includeClassName = true;
                    break;
                }
```

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
```diff
@@ -0,0 +1,102 @@
...
+
+namespace Neos\Flow\Mvc\Routing;
+
+use Neos\Flow\Mvc\Exception\InvalidActionNameException;
+use Neos\Flow\Mvc\Routing\Exception\InvalidControllerException;
+use Neos\Flow\ObjectManagement\ObjectManagerInterface;
+use Neos\Flow\Reflection\ReflectionService;
+use Neos\Flow\Annotations as Flow;
+use Neos\Utility\Arrays;
+
```
mhsdesign, line 24:
```suggestion

/**
 * Allows to annotate controller methods with route configurations
 *
 * Implementation:
 *
 * Flows routing configuration is declared via \@package, \@subpackage, \@controller and \@action (as well as the format \@format)
 * The first three options will resolve to a fully qualified class name {@see \Neos\Flow\Mvc\ActionRequest::getControllerObjectName()}
 * which is instantiated in the dispatcher {@see \Neos\Flow\Mvc\Dispatcher::dispatch()}
 *
 * The latter \@action option will be treated internally by each controller.
 * By convention and implementation of the default ActionController inside processRequest
 * {@see \Neos\Flow\Mvc\Controller\ActionController::callActionMethod()} will be used to concatenate the "Action" suffix
 * to the action name and invoke it internally with prepared method arguments.
 * The \@action is just another routing value while the doest not really know about "actions" from the "outside" (dispatcher).
 *
 * Creating routes by annotation must make a few assumptions to work.
 * As not every FQ class name is representable via the routing configuration (e.g. the class has to end with "Controller"),
 * only classes can be annotated that reside in a correct location and have the correct suffix.
 * Otherwise, an exception will be thrown as the class is not discoverable by the dispatcher.
 *
 * The routing annotation is placed at methods.
 * It is validated that the annotated method ends with "Action" and a routing value with the suffix trimmed will be generated.
 * Using the annotations on any controller makes the assumption that the controller will delegate the request to the dedicate
 * action by depending "Action".
 * This thesis is true for the ActionController.
 *
 * As discussed in https://discuss.neos.io/t/rfc-future-of-routing-mvc-in-flow/6535 we want to refactor the routing values
 * to include the fully qualified controller name, so it can be easier generated without strong restrictions.
 * Additionally, the action mapping should include its full name and be guaranteed to called.
 * Either by invoking the action in the dispatcher or by documenting this feature as part of a implementation of a ControllerInterface
 */
```

wrote a little something ;)

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,39 @@
...
+ * @Annotation
+ * @NamedArgumentConstructor
+ * @Target({"METHOD"})
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class Route
+{
+    public function __construct(
+        public readonly string $uriPattern,
+        public readonly ?string $name = null,
```
mhsdesign: name is only cosmetic right? maybe we should annotate that?
... would be great to have some documentation here ^^

Neos.Flow/Classes/Annotations/Route.php
mhsdesign:
```suggestion
        public readonly array $httpMethods = [],
        public readonly array $defaults = [],
```

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
mficzel: Am a bit undecided ... we could forbid the keys "@package", "@subpackage", "@controller" and "@action" but i am not totally sure it is worth the effort. We could even forbid any keys starting with "@" if we find a way to infer formats other than "html".

Neos.Flow/Classes/Annotations/Route.php
mficzel: If you want to create a specific route with a parameter you would use that. Think about special pathes for a language.

Neos.Flow/Classes/Annotations/Route.php
mficzel: The whole name thingy in routes is pretty cosmetic anyways. Not sure we have to do something special here.

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
mhsdesign: ahh i see `@format` should be allowed from the outside, well i dindt had that on my scope :D

> am not totally sure it is worth the effort

no i would say it is, that way we make sure that we have a well defined api which doesn not allow random stuff and escape its sandbox ^^ makes it much easier to reason about when refactoring, as one would see, a yes no this usecase is not supported.

Neos.Flow/Classes/Annotations/Route.php
mhsdesign: i think we have to limit `uriPattern` to not allow `{@action}` or `{@controller}` as this would be super weird and buggy ^^

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
mhsdesign: Maybe `@format` should also get its own option in the annotation

mficzel: I now filter out '@package', '@subpackage', '@controller', '@action' from defaults via route. Not sure wether the concept of "format" lives long enough to justify adding it as a special option to the attribute.

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
```diff
@@ -0,0 +1,138 @@
...
+                                [
+                                    '@package' => $controllerPackageKey,
+                                    '@subpackage' => $subPackage,
+                                    '@controller' => $controller,
+                                    '@action' => $action,
+                                    '@format' => 'html'
+                                ],
+                                $defaults
+                            )
+                        ];
```
mhsdesign: for simplicity ^^

```suggestion
                        $configuration = [
                            'name' => $controllerPackageKey . ' :: ' . $controller . ' :: ' . ($annotation->name ?: $action),
                            'uriPattern' => $annotation->uriPattern,
                            'httpMethods' => $annotation->httpMethods,
                            'defaults' => array_merge(
                                $annotation->defaults,
                                [
                                    '@package' => $controllerPackageKey,
                                    '@subpackage' => $subPackage,
                                    '@controller' => $controller,
                                    '@action' => $action,
                                    '@format' => $annotation->defaults['@format'] ?? 'html'
                                ]
                            )
                        ];
```

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
```diff
@@ -0,0 +1,138 @@
...
+                $annotations = $this->reflectionService->getMethodAnnotations($className, $methodName, Flow\Route::class);
+                foreach ($annotations as $annotation) {
+                    if ($annotation instanceof Flow\Route) {
+                        $controller = substr($controllerName, 0, -10);
+                        $action = substr($methodName, 0, -6);
+                        $defaults = array_filter(
+                            $annotation->defaults ?? [],
+                            fn($key) => !in_array($key, ['@package', '@subpackage', '@controller', '@action']),
+                            ARRAY_FILTER_USE_KEY
+                        );
```
bwaidelich: I would suggest to detect those directly in the `Route` attribute implementation and throw.
This will simplify this code and notify users about the mis-usage

Neos.Flow/Classes/Annotations/Route.php
mficzel: Added some doc-comments

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
mhsdesign: Yes even better than what i suggested with https://github.com/neos/flow-development-collection/pull/3325#discussion_r1535954865

Neos.Flow/Classes/Annotations/Route.php
mficzel: Added a check for that

Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
mficzel: Implement that

bwaidelich (approved):
I love it. Thanks for bearing with us and sticking through this!
+1 by reading

mhsdesign (changes requested):
Before merge i like to adjust one nitpick myself (and actually experiment with this feature for real ^^) (ill add a pr on top of this)

Additionally i want to discuss with you martin the last final touch like

that `Flow\Route` might rather be called `Flow\Action` so that we can have `Flow\Endpoint` for a rest controller and etc. But i guess this is best to be discussed in a huddle ;)

mficzel:
Flow/action is a bit misleading as we are configuring the route. A possible name would be Flow/ActionRoute

bwaidelich:
IMO `Flow/Route` is fine.
You can always have a custom `MyPackage/Route`

mhsdesign:
Ill just write it down:

1.) `ActionController` + method annotation

This is the main usecase: We want to support annotations on an `ActionController` (in my example below as `Flow\Action`).
As "legacy" behaviour here, to satisfy `ActionController::callActionMethod`, we require the action method to end with `Action` and trim the suffix.
The expected route values are listed as comment:

```php
class MyThingController extends ActionController
{
    // @package My.Package
    // @controller MyThing
    // @action some
    #[Flow\Action(path: 'foo')]
    public function someAction()
    {
    }
}
```

2.) `ControllerInterface` + method annotation (proposed behaviour not implemented here right now)

If the `Flow\Action` annotation is placed on a custom controller, i think it should behave a little less magic.
(Custom controllers are currently unattractive but via the dispatcher overhaul they will be simpler to use #3311)
In comparison when being placed onto an `ActionController` we might not enforce the 'Action' suffix and pass the method name 1 by 1 to the `@action` route value:

```php
class MySimpleController implements ControllerInterface
{
     // @package My.Package
     // @controller MySimple
     // @action myMethod
    #[Flow\Action(path: 'foo')]
    private function myMethod(ActionRequest $request, ActionResponse $response)
    {
    }

    public function processRequest(ActionRequest $request, ActionResponse $response)
    {
        $this->{$request->getControllerActionName()}($request, $response);
    }
}
```

3.) `ControllerInterface` + class annotation

And as part of a **future scope**, once `@action` is truly optional: https://discuss.neos.io/t/rfc-future-of-routing-mvc-in-flow/6535/5

We can (and probably should) add a class based annotation like:

```php
// @package My.Package
// @controller MyRest
#[Flow\Endpoint(path: 'foo')]
class MyRestController extends RestController
{
    private function get()
    {
    }

    private function post()
    {
    }
}
```
`Flow\Endpoint` can only be placed once per class at its top and `Flow\Action` must not be used in this case.

-------

Other things:

In the future as proposed here https://discuss.neos.io/t/rfc-future-of-routing-mvc-in-flow/6535/5, we will have it easier as we can just generate `@controller My\Package\MyCustomController` as FQN

Additionally i seem to outnumbered by bernhard and christian as they do not want to enforce action methods to be public, and i now agree ^^. 

------

The namings `Flow\Action` and `Flow\Endpoint` are not super well founded, bernhard and me just though of them ;)
`Flow\ActionRoute` and `Flow\EndpointRoute` or whatever would also be cool, but i think i like to have those concepts separated and not mixed into one `Flow\Route` ... or maybe that is also fine?

bwaidelich:
`Flow\Action` does not make sense to me.
I think we should just go for `Flow\Route` and allow the attribute on methods that have the `Action` suffix for now.

As a follow-up we could extend the behavior to allow...

* the attribute on classes (IMO that should be done with the [suggested minimal controller implementation](https://discuss.neos.io/t/rfc-future-of-routing-mvc-in-flow/6535/5) of `__invoke(Request): Response`)
* the attribute on methods of other classes (for that it would set `@controller` to the FQN once that is supported by the routing framework)

A `Flow\Endpoint` is speculation at this point, it could as well come from a 3rd party – but in any case it does not conflict with `Flow\Route` IMO

mhsdesign:
> I think we should just go for `Flow\Route` and allow the attribute on methods that have the `Action` suffix for now.

Jain if we make it purely specific for the `ActionController` i agree. All other / custom controllers would otherwise not be supported at first.

for custom controllers i wouldnt want any suffix magic and have in `getControllerActionName` the whole method name without trimming.

bwaidelich:
>  All other / custom controllers would otherwise not be supported at first.

at _first_. I don't think it's fair to block a new feature because it does not include another feature

> for custom controllers i wouldnt want any suffix magic and have in getControllerActionName the whole method name without trimming.

I agree – we can just add a check for the controller base class and omit the magic if it's not a Flow ActionController. But IMO this can be done in a separate PR

mhsdesign:
> I don't think it's fair to block a new feature because it does not include another feature

Agreed.

> I agree – we can just add a check for the controller base class and omit the magic if it's not a Flow ActionController.

okay so exactly as described above in point 2?

> 2.) `ControllerInterface` + method annotation (proposed behaviour not implemented here right now)


but regarding:

> But IMO this can be done in a separate PR

i dont see a reason for that, a pr should IMO be self contained and complete. But dont worry _i_ will prepare the necessary adjustments as it was _my_ idea. Dont want to chase anyone around ^^ 


-------

Regarding `Flow\Endpoint`, i was actually referring to attributes on classes and i wanted to discuss to have two separate annotations instead of one almighty.

But it seems we want to go the route (hatahahah) of allowing the annotation to be placed at some point on methods _and_ classes?

```php
class MyThingController extends ActionController
{
    // @package My.Package
    // @controller MyThing
    // @action some
    #[Flow\Route(path: 'foo')]
    public function someAction()
    {
    }
}
```

and

```php
// @package My.Package
// @controller MyRest
#[Flow\Route(path: 'foo')]
class MyRestController extends RestController
{
    private function get()
    {
    }

    private function post()
    {
    }
}
```

kitsunet (approved):
Thanks a lot, this was a struggle, I actually would rather hit merge now than later, I think @mhsdesign concerns can and should all be tackled in a different PR, it's important to remmeber that just because we decided on a redirection it's not reality yet and the controller chnages are not fully in not even speaking about documenting them. So IMHO totally fine if this doesn't work correctly for _other_ controllers that are not fully there yet.

kitsunet:
So giving this until tomorrow morning, then anything else becomes a separate PR as this here will be merged.

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
```diff
@@ -729,10 +729,66 @@ two more options you can use:
...
+
+      #[Flow\Route(uriPattern:'my/other/b-path', defaults: ['test' => 'b'])]
+      #[Flow\Route(uriPattern:'my/other/c-path', defaults: ['test' => 'c'])]
+      public function otherAction(string $test): void
+      {
+      }
+  }
+
+To find the annotation and tp specify the order of routes this has to be used together with the
+`\Neos\Flow\Mvc\Routing\AttributeRoutesProviderFactory` as `providerFactory` in  Setting `Neos.Flow.mvs.routes`
```
mhsdesign: typo

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,51 @@
...
+ * @Target({"METHOD"})
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class Route
+{
+    /**
+     * @param string $uriPattern The uri-pattern for the route without leading '/'. Must not contain `{@action}` or `{@controller}`.
+     * @param string|null $name (default null) The name ouf the route as it shows up in the route:list command
+     * @param array $httpMethods (default []) List of http verbs like 'GET', 'POST', 'PUT', 'DELETE', if not specified 'any' is used
+     * @param array $defaults (default []) Values to set for this route. Dan define arguments but also specify the `@format` if required.
```
mhsdesign: typo

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,51 @@
...
+        public readonly array $httpMethods = [],
+        public readonly array $defaults = [],
+    ) {
+        if (str_contains($uriPattern, '{@controller}') || str_contains($uriPattern, '{@action}')) {
+            throw new \DomainException(sprintf('It is not allowed to override {@controller} or {@action} in route annotations "%s"', $uriPattern), 1711129634);
+        }
+        if (in_array(array_keys($defaults), ['@package', '@subpackage', '@controller', '@action'])) {
+            throw new \DomainException(sprintf('It is not allowed to override @package, @controller, @subpackage and @action in route annotation defaults "%s"', $uriPattern), 1711129638);
+        }
+    }
```
mhsdesign: ill add a unit test for that ...

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
```diff
@@ -729,10 +729,66 @@ two more options you can use:
...
+      public function someAction(): void
+      {
+      }
+
+      #[Flow\Route(uriPattern:'my/other/b-path', defaults: ['test' => 'b'])]
+      #[Flow\Route(uriPattern:'my/other/c-path', defaults: ['test' => 'c'])]
+      public function otherAction(string $test): void
+      {
+      }
+  }
```
mhsdesign, line 775: ill try to add a functional test where once can see the feature 1 to 1 in action. Currently the unit tests are cool to have but really obscure :D

Neos.Flow/Classes/Annotations/Route.php
```diff
@@ -0,0 +1,51 @@
...
+     * @param array $httpMethods (default []) List of http verbs like 'GET', 'POST', 'PUT', 'DELETE', if not specified 'any' is used
+     * @param array $defaults (default []) Values to set for this route. Dan define arguments but also specify the `@format` if required.
+     */
+    public function __construct(
+        public readonly string $uriPattern,
+        public readonly ?string $name = null,
+        public readonly array $httpMethods = [],
+        public readonly array $defaults = [],
+    ) {
+        if (str_contains($uriPattern, '{@controller}') || str_contains($uriPattern, '{@action}')) {
```
mhsdesign: also prohibit 

`@package, @subpackage`

mhsdesign (dismissed):
While i adjust the above ill also take care of these things ... lets see if i pass your constraints of tomorrow, but i have a lot going on at the moment

mhsdesign:
@kitsunet and me agreed to just allow this feature at first on the `ActionController` so we can merge fast and keep the extensive followup discussion separate ;)

That being said, there _will_ be a followup, and we are still developing Flow 9 so this must not necessary be the end of it - but even if it were and we only find a sane solution in Flow 9.1 it should be acceptable as this feature is itself a milestone.

I will try to provide the constraints and the adjustments above today unless someone else does it ;)

Neos.Flow/Tests/Unit/Mvc/Routing/AttributeRoutesProviderTest.php
```diff
@@ -0,0 +1,134 @@
...
+            ->method('getMethodsAnnotatedWith')
+            ->with('Vendor\Example\Controller\ExampleController', Flow\Route::class)
+            ->willReturn(['specialAction']);
+
+        $this->mockReflectionService->expects($this->once())
+            ->method('getMethodAnnotations')
+            ->with('Vendor\Example\Controller\ExampleController', 'specialAction', Flow\Route::class)
+            ->willReturn([
+                new Flow\Route(uriPattern: 'my/path'),
+                new Flow\Route(uriPattern: 'my/other/path', name: 'specialRoute', defaults: ['test' => 'foo'], httpMethods: ['get', 'post'])
```
mhsdesign: lowercase httpMethods will not work as we do case sensitive comparison here: https://github.com/neos/flow-development-collection/blob/c6e32474c0fa0ce1940c2d2e6fb9aa7f87f3d566/Neos.Flow/Classes/Mvc/Routing/Route.php#L490

we might as well validate that in the constructor

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
```diff
@@ -729,10 +729,66 @@ two more options you can use:
 With ``suffix`` you can specify a custom filename suffix for the SubRoute. The ``variables`` option allows you to
 specify placeholders in the SubRoutes (see `Nested Subroutes`_).
 
+It also is possible to specify a `providerFactory` and (optional) `providerOptions` to generate the subroutes via the
```
mhsdesign, line 732: via are the `providerOptions` optional in this case? I think this is the general chapter which should explain how to use a custom `RoutesProviderFactory`?

bwaidelich: A custom provider might not need custom options and in this case is given an empty array, see https://github.com/neos/flow-development-collection/blob/ee8a8127e1d77908b5f44786a1ba4d6f2500c583/Neos.Flow/Classes/Mvc/Routing/ConfigurationRoutesProvider.php#L31

I think that totally makes sense and it's the way we do it for a lot of other similar configuration (patternOptions, postprocessorOptions, editorOptions, ...)

mhsdesign: Oke I thought so, will update the example so it’s not bound to the AttributeRoutes thing and more generic

Neos.Flow/Classes/Configuration/Loader/RoutesLoader.php
```diff
@@ -84,20 +84,28 @@ public function load(array $packages, ApplicationContext $context): array
...
             }
-            $subRoutesName = $packageKey . 'SubRoutes';
-            $subRoutesConfiguration = ['package' => $packageKey];
+            if (isset($routeFromSettings['providerFactory'])) {
+                $routeDefinitions[] = [
+                    'name' => $configurationKey,
+                    'providerFactory' => $routeFromSettings['providerFactory'],
+                    'providerOptions' => $routeFromSettings['providerOptions'] ?? [],
+                ];
+                continue;
```
mhsdesign, line 97: i was trying to write some functional test for this but doesnt this `continue` here prohibit that `Neos.Flow` provides _real_ routes.yaml AND routes from annotation?

Currently we use

```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Neos.Flow': TRUE
```

but when changing this to

```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Neos.Flow':
          providerFactory: ...
```

this will not include the defined routes of _routes.yaml_ no?

Or is the idea to use a "virtual" package / separate key like

```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Neos.Flow:FromAnnotation':
          providerFactory: ...
```

kitsunet: I thought the idea is the latter

mhsdesign: Yeah though so too and also the code looks like it. Will document this as recommended.

bwaidelich: If you want both (which is unlikely for 3rd party packages) you'll write:
```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Neos.Flow': true
        'Neos.Flow.attributes':
          providerFactory: ...
```

bwaidelich: But a 3rd party package would not use the Neos.Flow namespace ofc and instead for example:
```yaml
Neos:
  Flow:
    mvc:
      routes:
        'Some.Package':
          before: 'Neos.Flow'
          providerFactory: ...
```

bwaidelich:
image

Can we get a grip on this one or what is still blocking that from being merged, @mhsdesign ?

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
```diff
@@ -729,10 +729,67 @@ two more options you can use:
...
+`Neos\Flow\Mvc\Routing\RoutesProviderFactoryInterface` and the ``Neos\Flow\Mvc\Routing\RoutesProviderInterface`.
+
+.. code-block:: yaml
+
+  Neos:
+    Flow:
+      mvc:
+        routes:
+          Vendor.Example:
```
kitsunet:
```suggestion
          Vendor.Example.attributes:
```

Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Routing.rst
```diff
@@ -729,10 +729,67 @@ two more options you can use:
...
+To find the annotation and tp specify the order of routes this has to be used together with the
+`\Neos\Flow\Mvc\Routing\AttributeRoutesProviderFactory` as `providerFactory` in  Setting `Neos.Flow.mvc.routes`
+
+.. code-block:: yaml
+
+  Neos:
+    Flow:
+      mvc:
+        routes:
+          Vendor.Example:
```
kitsunet:
```suggestion
          Vendor.Example.attributes:
```

kitsunet:
I'll merge this as soon as tests are done, everything else can really happen in follow ups.

mhsdesign:
I wish you were so eager with other prs ;)
I just managed to add a functional tests and improved documentation a little. Which ill rebase now on 9.0 and make a followup.

Two questions:

1) Also i stumbled over the thing that the `AttributeRoutesProvider`'s are NOT cached and always computed at runtime (unless the router has the route cached). Is this anticipated?

2) Its special that the `AttributeRoutesProvider` and `ConfigurationRoutesProvider` share one interface as registering the `ConfigurationRoutesProvider` via the settings would cause infinite recursion.

I would say we need one `CentralRoutesProviderInterface` implementation and possible several `AttributeRoutesProvider` implementations.

That means: `ConfigurationRoutesProvider` will be renamed to `DefaultCentralRoutesProvider` or sth. we add the `CentralRoutesProviderInterface` and the rest can stay as is.

The naming is of course discussable but i think it makes more sense. Also by that we can clearly document the one interface as api and describe how it can be used.

Does this direction make sense to you?

bwaidelich:
> I wish you were so eager with other prs ;)

Likewise :)
My concern was not to get this merged asap but to avoid this from absorbing more resources from other pressing topics that affect existing core logic.
Besides we should IMHO only block PRs from others if we have sincere concerns or want to discuss matters that couldn't easily be done in follow-ups.

> Also i stumbled over the thing that the AttributeRoutesProvider's are NOT cached [...]
> unless the router has the route cached [...]

But that should almost always be the case, right? I hope that the route is only evaluated if it hasn't been hit before and I'd say that's totally fine

> Its special that the AttributeRoutesProvider and ConfigurationRoutesProvider share one interface

Yes, it's special and a bit weird. But I don't think that it's a problem we have to solve:
- You can globally replace the routes provider if you want to
- The provided `ConfigurationRoutesProvider` supports evaluating nested providers

IMO that's just a matter of documentation and it's quite a low-level API anyways

bwaidelich:
FYI: Unfortunately this does not seem to work in production context, see #3400

kitsunet:
#3401 is the bugfix just to have it refrenced here as well

PR #3329, opened 2024-03-13
TASK: Remove code related to PHP < 8.2

### LINKED ISSUE #3085: TASK: Remove dead code (`if (PHP_MAJOR_VERSION >= 8)`)
in some places, like the reflection service, we check if we are on php 8, but we dont need to anymore as we require currently 8.2, so we can get rid of some conditional logic.

The same is true for some $this->markTest as skipped in case php is not v8

robertlemke:
Code supporting backwards-compatibility with PHP versions earlier than 8.2 were removed, since the minimum required version for Flow is 8.2.

Resolves #3085

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
@@ -237,12 +237,6 @@ class CompileTimeObjectManager extends ObjectManager
                 }
                 if ($package instanceof FlowPackageInterface && $shouldRegisterFunctionalTestClasses) {
                     foreach ($package->getFunctionalTestsClassFiles() as $fullClassName => $path) {
-                        if (PHP_VERSION_ID <= 80000 && str_contains($fullClassName, '\\PHP8\\')) {
-                            continue;
-                        }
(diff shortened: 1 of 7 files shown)


ahaeslich (approved):

Neos.Flow/Classes/ObjectManagement/CompileTimeObjectManager.php
```diff
@@ -237,12 +237,6 @@ protected function registerClassFiles(array $packages): array
                 }
                 if ($package instanceof FlowPackageInterface && $shouldRegisterFunctionalTestClasses) {
                     foreach ($package->getFunctionalTestsClassFiles() as $fullClassName => $path) {
-                        if (PHP_VERSION_ID <= 80000 && str_contains($fullClassName, '\\PHP8\\')) {
```
mhsdesign, line 240: do we want to comment this out instead for preparation of PHP83 fixtures 😅

robertlemke: Nah.

robertlemke: Never leave dead code, in my opinion.

PR #3330, opened 2024-03-13
!!! TASK: Modernize and clean up session-related code

robertlemke:
The session-related code in Flow was updated to use modern PHP features and attributes. Classes are also declared as strict and a few minor bugs which surfaced due to type strictness were fixed along the way.

This change is breaking for anyone who implemented their own implementation of `SessionInterface` or `SessionManagerInterface`because parameter and return types were added. It's very easy to solve though.

### DIFF
--- a/Neos.Flow/Classes/Session/Aspect/LoggingAspect.php
+++ b/Neos.Flow/Classes/Session/Aspect/LoggingAspect.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 namespace Neos\Flow\Session\Aspect;
 
 /*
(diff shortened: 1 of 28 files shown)


Neos.Flow/Classes/Session/SessionManager.php
```diff
@@ -11,80 +12,39 @@
...
-
-    /**
-     * @Flow\Inject(name="Neos.Flow:SystemLogger")
-     * @var LoggerInterface
-     */
-    protected $logger;
+    protected ?SessionInterface $currentSession = null;
+    protected array $remoteSessions = [];
+
+    public function __construct(
```
mficzel, line 32: Not so sure about that as this mixes autowired properties with values from objects.yaml

Why not use `#[Flow\Injectconfiguration(... )]` for the single properties. 

Also the Objects.yaml makes it harder to understand the actual configuration pathes used if $someone reads the code

Neos.Flow/Classes/Session/TransientSession.php
```diff
@@ -184,11 +141,11 @@ public function close()
     /**
      * Explicitly destroys all session data
      *
-     * @param string $reason A reason for destroying the session – used by the LoggingAspect
+     * @param string|null $reason A reason for destroying the session – used by the LoggingAspect
      * @return void
      * @throws Exception\SessionNotStartedException
      */
-    public function destroy($reason = null)
+    public function destroy(string $reason = null): void
```
mficzel: Can we type such stuff as `?string` ... i kinda dislike implicit nullability. 

Also there is an accepted pr to deprecate implicit nullability from php: https://php.watch/rfcs/deprecate-implicitly-nullable-types

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -488,6 +488,22 @@ Neos\Flow\Session\Data\SessionMetaDataStore:
 
 Neos\Flow\Session\SessionManagerInterface:
   className: Neos\Flow\Session\SessionManager
+  arguments:
```
mficzel: As stated before ... i tend to prefer `#[Flow\InjectConfiguration]` Attributes

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -488,6 +488,22 @@ Neos\Flow\Session\Data\SessionMetaDataStore:
...
+    3:
+      setting: Neos.Flow.session.garbageCollection.probability
+    4:
+      setting: Neos.Flow.session.garbageCollection.maximumPerRun
+    5:
+      setting: Neos.Flow.session.inactivityTimeout
+
+Neos\Flow\Session\SessionManager:
```
mficzel: As stated before ... i tend to prefer `#[Flow\InjectConfiguration]` Attributes

mficzel (changes requested):
I totally like by reading but i consider injecting the settings from Objects.yaml and mixing the autowired with configured arguments really strange. Also injecting Settings via objects yaml really makes it harder to get where this comes from.

More of a nitpick is wether we use implict nullable arguments. I can accept that.

Neos.Flow/Classes/Session/TransientSession.php
kdambekalns: But be careful, `?string $reason` is **not** the same as `?string $reason = null`, see https://3v4l.org/aR0bZ

Neos.Flow/Classes/Session/Data/SessionMetaDataStore.php
```diff
@@ -14,27 +14,25 @@
...
 
 /**
  * @internal
  */
 class SessionMetaDataStore
 {
-    protected VariableFrontend $cache;
-
     protected const TAG_PREFIX = 'customtag-';
+    protected const GARBAGE_COLLECTION_CACHE_IDENTIFIER = '_garbage-collection-running';
```
kdambekalns, line 30: Not against it, but I am pretty sure this was a deliberate naming choice:

- it is a "garbage-collection"-cache-identifier,
- not a "garbage-collection-cache"-identifier

🤣

mficzel: I chose that name and I am sure I did not put that much thought into it :-)

Neos.Flow/Classes/Session/Aspect/LoggingAspect.php
```diff
@@ -81,76 +58,58 @@ public function logResume(JoinPointInterface $joinPoint)
...
+    public function logRenewId(JoinPointInterface $joinPoint): mixed
     {
-        /** @var SessionInterface $session */
         $session = $joinPoint->getProxy();
-        $oldId = $session->getId();
         $newId = $joinPoint->getAdviceChain()->proceed($joinPoint);
-        if ($session->isStarted()) {
-            $this->logger->info(sprintf('%s: Changed session id from %s to %s', $this->getClassName($joinPoint), $oldId, $newId));
+        if ($session instanceof SessionInterface) {
+            $oldId = $session->getId();
```
kitsunet: Doesn't look like we need $oldId outside the next "if", so this could again be a combination instanceof & isStarted "if" here?

Neos.Flow/Classes/Session/SessionManager.php
kitsunet: I would be fine either way, like this it's easier to test. Generally I am for not optimizing for flow magic.

kitsunet (approved):
I like, my comments should not block a merge.

Neos.Flow/Classes/Session/SessionManager.php
robertlemke: I took a little extra-mile and reviewed the `InjectConfiguration` support for constructor arguments (https://github.com/neos/flow-development-collection/pull/3086), so we can use that instead of the Objects.yaml-hassle.

Neos.Flow/Classes/Session/TransientSession.php
robertlemke: Good catch, I think we should (and I will) change this to `?string` instead, now that we're at it.

Neos.Flow/Classes/Session/Aspect/LoggingAspect.php
robertlemke: Will fix.

mficzel (approved):
Beautiful and i can confirm that it works on my machine ;-)

PR #3336, opened 2024-03-28
!!! TASK: Refactor uri helpers

mhsdesign:
See #3316

**Upgrade instructions**

The following methods were removed from the `UriHelper` as they are obsolete and not used.
Its unlikely that the functionality is known and simple to implement yourself.
- `\Neos\Flow\Http\Helper\UriHelper::getUsername`
- `\Neos\Flow\Http\Helper\UriHelper::getPassword`
- `\Neos\Flow\Http\Helper\UriHelper::parseQueryIntoArguments`

The method `\Neos\Flow\Http\Helper\UriHelper::uriWithArguments` was renamed to `\Neos\Flow\Http\Helper\UriHelper::uriWithQueryParameters` to distinct between route values and query parameters which are not the same.
Also it will encode the query parameters after [PHP_QUERY_RFC1738](https://www.php.net/manual/de/url.constants.php#constant.php-query-rfc1738).

**Review instructions**

The pr #3316 introduced a new uri helper while we already had one actually. This pr combines the two and cleans things up.

To ensure the logic of `uriWithAdditionalQueryParameters` is not duplicated and possibly handled elsewhere differently the helper is now also used internally by the uriConstraints.

Also the method has been renamed to better fit the previous sibling `uriWithArguments`.

The removed methods are dead code and previously introduced once with Flow 5.1: https://github.com/neos/flow-development-collection/commit/85408589462b7530180d3dce2858500f29f94bbe
As part of replacements for the old (now removed) flow Uri implementation:

`Neos\Flow\Http\Uri::getUsername` -> `\Neos\Flow\Http\Helper\UriHelper::getUsername`
`Neos\Flow\Http\Uri::getPassword` -> `\Neos\Flow\Http\Helper\UriHelper::getPassword`
`Neos\Flow\Http\Uri::getArguments` -> `\Neos\Flow\Http\Helper\UriHelper::parseQueryIntoArguments`

So maybe these methods _are_ known in fact and it would be a bit mean to remove them just because i felt like it and we dont use / test them?

**Checklist**


bwaidelich (approved):
Lovely, thanks for your efforts!

mhsdesign:
I dont know if i was to eager deleting stuff so i summon @kitsunet to please have a look at it and veto if plausible :D

Neos.Flow/Classes/Http/Helper/UriHelper.php
```diff
@@ -11,15 +11,21 @@
...
  */
 
+use Neos\Utility\Arrays;
 use Psr\Http\Message\UriInterface;
 
 /**
  * Helper to extract information from Uris.
  */
-abstract class UriHelper
+final class UriHelper
```
kitsunet, line 20: I'd personally prefer the other helper, this Helper namespace seems rather pointless.

mhsdesign: Yeah but there are already a few helpers inside: https://github.com/neos/flow-development-collection/tree/9.0/Neos.Flow/Classes/Http/Helper

and it would be a little less breaking (though I think no one uses the helper)

The other helper (from the merge yesterday) resided directly in the http namespace but at that point I was unaware of the other helpers

mhsdesign: I don’t exactly love the namespace either but idk it does no harm as well.

Btw _you_ added the uri helper as part of the psr7 migration in this place ;) so better keep that location?

kitsunet: Right, I guess better keep the older one, I also just realized that. Shame on me for creating that useless namespace, although back then I planned for more helpers to support using PSR-7 in Flow (but pretty much everything else exists in guzzle so no need for them). tl;dr; kk lets keep this.

kitsunet (approved):

Issue #3331, opened 2024-03-14
BUG: Calling QueryResult->count() returns incorrect number of results when criterions on related entities are used


tvt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Calling `count()` on a `QueryResult` that use properties of related entities in criterions returns wrong counts.

It's caused by the fix in #3140.

The SQL generated by the `Query` uses `LEFT JOIN`s for criterions on related entities. If there are more than one related entity that matches the criterions the SQL will return multiple rows if the `DISTINCT` clause is not added.

I hope this makes sense, but otherwise the test case should speak for itself.

### Expected Behavior

Calling `count()` on a `QueryResult` should give back the same number as getting the entities from the `QueryResult` and counting them.

### Steps To Reproduce

Easiest way is to add the following test case to `QueryTest.php`.

It'll fail in the last assert of the test case.

```
    /**
     * @test
     */
    public function subpropertyQueriesFailCountAlias()
    {
        $testEntityRepository = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\TestEntityRepository();
        $testEntityRepository->removeAll();

        $testEntity = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\TestEntity;
        $testEntity->setName('Flow');

        $subEntity1 = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\SubEntity;
        $subEntity1->setContent('foo');
        $subEntity1->setParentEntity($testEntity);
        $testEntity->addSubEntity($subEntity1);
        $this->persistenceManager->add($subEntity1);

        $subEntity2 = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\SubEntity;
        $subEntity2->setContent('foo');
        $subEntity2->setParentEntity($testEntity);
        $testEntity->addSubEntity($subEntity2);
        $this->persistenceManager->add($subEntity2);

        $testEntityRepository->add($testEntity);

        $this->persistenceManager->persistAll();

        $query = new Query(\Neos\Flow\Tests\Functional\Persistence\Fixtures\TestEntity::class);

        $constraint = $query->logicalAnd($query->equals('subEntities.content', 'foo'));
        $result = $query->matching($constraint)->execute();

        $count = $result->count();
        $arrayCount = $result->toArray();

        self::assertEquals(1, count($arrayCount), 'This correctly returns 1');
        self::assertEquals(1, $count, 'this returns 2');
    }
```

[...]

PR #3342, opened 2024-04-16
BUGFIX: Fix count for query across OneToMany joins

tvt:
The `Query->count` now returns the correct count when a criterion is added on a OneToMany relation.

**Review instructions**

The problem is described in details in #3331.

**Checklist**

### DIFF
--- a/Neos.Flow/Tests/Functional/Persistence/Doctrine/QueryTest.php
+++ b/Neos.Flow/Tests/Functional/Persistence/Doctrine/QueryTest.php
@@ -353,6 +353,45 @@ class QueryTest extends FunctionalTestCase
         self::assertEquals([$testEntity2], $unserializedQuery->execute()->toArray());
     }
 
+    /**
+     * @test
+     */
+    public function countReturnsCorrectNumberOfEntities()
+    {
+        $testEntityRepository = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\TestEntityRepository();
+        $testEntityRepository->removeAll();
+
+        $testEntity = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\TestEntity;
+        $testEntity->setName('Flow');
+
+        $subEntity1 = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\SubEntity;
+        $subEntity1->setContent('foo');
+        $subEntity1->setParentEntity($testEntity);
+        $testEntity->addSubEntity($subEntity1);
+        $this->persistenceManager->add($subEntity1);
+
+        $subEntity2 = new \Neos\Flow\Tests\Functional\Persistence\Fixtures\SubEntity;
+        $subEntity2->setContent('foo');
+        $subEntity2->setParentEntity($testEntity);
+        $testEntity->addSubEntity($subEntity2);
+        $this->persistenceManager->add($subEntity2);
+
+        $testEntityRepository->add($testEntity);
+
+        $this->persistenceManager->persistAll();
+
+        $query = new Query(\Neos\Flow\Tests\Functional\Persistence\Fixtures\TestEntity::class);
+
+        $constraint = $query->logicalAnd($query->equals('subEntities.content', 'foo'));
+        $result = $query->matching($constraint)->execute();
+
+        $count = $result->count();
+        $arrayCount = $result->toArray();
+
+        self::assertEquals(1, count($arrayCount), 'This correctly returns 1');
+        self::assertEquals(1, $count, 'this returns 2');
+    }
+
     protected function assertQueryEquals(Query $expected, Query $actual)
     {
         self::assertEquals($expected->getConstraint(), $actual->getConstraint());
--- a/Neos.Flow/Classes/Persistence/Doctrine/CountWalker.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/CountWalker.php
@@ -63,7 +63,8 @@ class CountWalker extends TreeWalkerAdapter
         $AST->orderByClause = null;
     }
 
-    private function isDistinctRequired(): bool {
+    private function isDistinctRequired(): bool
+    {
         foreach ($this->getQueryComponents() as $queryComponent) {
             if (isset($queryComponent['relation']['type']) && $queryComponent['relation']['type'] === ClassMetadataInfo::ONE_TO_MANY) {
                 return true;
--- a/Neos.Flow/Classes/Persistence/Doctrine/CountWalker.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/CountWalker.php
@@ -8,6 +8,7 @@ namespace Neos\Flow\Persistence\Doctrine;
  * with this package in the file License-BSD.txt.                         *
  *                                                                        */
 
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Query\AST\AggregateExpression;
 use Doctrine\ORM\Query\AST\PathExpression;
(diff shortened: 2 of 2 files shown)


kdambekalns (changes requested):
As this effectively just reverts https://github.com/neos/flow-development-collection/pull/3140 without any connection to https://github.com/neos/flow-development-collection/pull/415 I think this needs to be discussed.

I see the bug, but can't we find a way to fix it _and_ keep the performance improvement?

kitsunet:
IMHO we need to make this an option to be decided in userland code, the question is rather what should be the default.

kdambekalns:
But that option would be making the query distinct, no? So, do we "just" need to document it?

Or do we "play it safe" by keepig `true` unless switched off somehow (to "enable performance"?)

But AFAIK we cannot decide between "distinct flag not set" and "distinct flag set to false", can we? So that would mean adding (yet) something Flow-specific.

Sebobo:
I wrote in Slack back then an assumption: 
`The problem is basically whether we join on a 1-1 or many-1 relationship which doesn’t need to be distinct, or we join on a 1-many right?`
Which could be a basis for an auto detection.

But having two methods like doctrine does it in https://www.doctrine-project.org/projects/doctrine-orm/en/3.0/reference/query-builder.html#the-expr-class with `count` and `countDistinct` would be the another option.

bwaidelich:
My 2ct:
Please lets not make it a configuration option or "solve" it via documentation. Both are just coward ways to push the burden of API design to the userland :)
I think two distinct (no pun intended) methods make the most sense, and personally I would prefer the default (`count()`) to behave like I would expect in most cases (i.e. returning the same result as if I would count the items of the iterated resultset)

tvt:
@kdambekalns sure, it reverts the incorrect code but also adds a test case, so the same bug can't be reintroduced.

I understand there's a performance consideration as well, but I would expect that correctness overrules performance.

The performance can probably be fixed without reintroducing bugs, but I'm not deep into the Flow Doctrine code, so I don't know how.

Also if you're having performance issues with `count`, why not write DQL or native SQL instead of introducing bugs to the framework?

Sebobo:
> Also if you're having performance issues with `count`, why not write DQL or native SQL instead of introducing bugs to the framework?

Nobody wanted to introduce a bug.

The problem is that the performance issues are also triggered by using the official APIs of Flow and Neos. Going around that and using custom queries is not an easy task and creates issues when upgrading Flow/Neos. I know that because I spent a ton of time on performance stuff in the last years. And initially you don't always know what's going on. So one can spend days debugging after a project has been running perfectly fine for a long time and at some point reached the critical mass and doesn't scale anymore

Similar issue with our caching. Many Neos projects get super slow after some weeks when integrators don't know that the filesystem caches will kill their app at some point. Now you can say: "Hey integrators, properly read the docs", or we provide defaults that scale until you do your super fancy thing, because that means, its now your job to also care about the performance of it.

I expect a framework with its APIs to handle a million db entries. Which is a laughable amount actually.

Now if we adjust all the public APIs to use native queries and make that safe, the next dev will again use the query builder and run into the same issue at some point.

Don't get me wrong, I really want to have a proper fix for this, that fulfils all requirements.

tvt:
> Don't get me wrong, I really want to have a proper fix for this, that fulfils all requirements.

I agree that would be the best solution, but until then I suggest that the `count` function returns the correct result. Why would you even need a `count` if whether it's correct or not doesn't matter?

> [...] and creates issues when upgrading Flow/Neos.

Tell me about it. We're stuck on 7.3.19 because of this bug being introduced. It wasn't easy to track down.

KarstenDambekalns-eaton:
> I think two distinct (no pun intended) methods make the most sense

Did someone already try to find out how "pure Doctrine" users usually handle this? Or is this somehow a problem that we have created ourselves by adding to many layers around Doctrine in Flow?

kdambekalns:
> I think two distinct (no pun intended) methods make the most sense

Did someone already try to find out how "pure Doctrine" users usually handle this? Or is this somehow a problem that we have created ourselves by adding to many layers around Doctrine in Flow?

tvt:
What about this then @kdambekalns?

Neos.Flow/Tests/Functional/Persistence/Doctrine/QueryTest.php
```diff
@@ -353,6 +353,45 @@ public function comlexQueryWithJoinsCanBeExecutedAfterDeserialization()
...
+
+        $constraint = $query->logicalAnd($query->equals('subEntities.content', 'foo'));
+        $result = $query->matching($constraint)->execute();
+
+        $count = $result->count();
+        $arrayCount = $result->toArray();
+
+        self::assertEquals(1, count($arrayCount), 'This correctly returns 1');
+        self::assertEquals(1, $count, 'this returns 2');
+    }
```
bwaidelich, line 393: This test _won't_ use the `$AST->selectClause->isDistinct = true`, right?
Would it make sense to add one test that _does_? (or vice versa)

tvt: It will. `subEntities` is a `OneToMany` relation of `TestEntity`.

You can run the test with `$AST->selectClause->isDistinct = true` removed and it'll fail.

bwaidelich:
How shall we proceed with this one?

My gut feeling is that whatever we decide/detect is to some extend intransparent and potentially leads to errors.

@Sebobo you have valid performance concerns and those have to be addressed.

What do you folks think of adding that second `count` method like doctrine does? Wouldn't that solve the issue?

tvt:
I just want to make sure we're all on the same page @bwaidelich. As I see it this fixes a bug without introducing any performance problems, except for situations that currently have the possibility of returning wrong counts.

If this fix causes performance issues for anybody, I think the bigger issue is that they are currently getting wrong counts.

bwaidelich:
> As I see it this fixes a bug without introducing any performance problems

I'm not into ORM at all (and try not to use it if possible) so I can't judge whether the fix catches all cases or might even introduce new, different, bugs (no allegation I simply can't tell out of ignorance to the topic)

>  I think the bigger issue is that they are currently getting wrong counts

I agree. But we have to take those decisions carefully nevertheless, because it means that very specific patch level versions of Flow behave differently.

If your fix solves both, performance and inconsistency issues, it should be merged IMO but if @Sebobo or @kdambekalns or anyone else still has concerns, those should be taken seriously

tvt:
Any feedback @kdambekalns and @Sebobo? I don't think I can do much more here. 

This should only have an impact on performance on the queries that currently returns wrong counts.

Sebobo:
@tvt thanks for the change, I will test it in the use case where I hat the biggest issues and give feedback.

One thing: according to our roadmap this change should target 8.3 and not 7.3.

tvt:
Hmm, what's the easiest way to target 8.3 instead? Should I create a new PR?

bwaidelich:
> Hmm, what's the easiest way to target 8.3 instead? Should I create a new PR?

You could change the target branch in this PR, but that usually leads to a mess.
A new PR would be better, you can probably just cherry-pick your commits:
```
git cherry-pick da13be260fce86d03d6d42c5157e7b6a785297b9
git cherry-pick eea4d2b90d123238790a537cfaa003d8b4763c14
git cherry-pick 521440396b3695396251ed332acd2060102ad71d
```

tvt:
I've recreated the PR for 8.3 here: #3345.

kdambekalns:
Thanks for pushing this through!

Hint: When going "up", changing the base of a PR works fine more often than not.

PR #3344, opened 2024-04-18
BUGFIX: Add exponential backoff when reading from Redis

### LINKED ISSUE #3284: BUG: Error when Redis not ready

### Current Behavior

In a project we are seing exceptions after a deployment. The cause is a failing doctrine proxy compilation step early on, that in turn is caused by the `RedisBackend` trying to read Redis, but that just says: `LOADING Redis is loading the dataset in memory`.

### Expected Behavior

It should "Just Work" (tm) 🙃

### Steps To Reproduce

_No response_

### Environment

```markdown
- Flow:
- PHP:
```

### Anything else?

_No response_

kdambekalns:
When Redis is not ready (yet) when trying to read from the cache, this makes Flow wait and retry up to 8 times with an exponentially growing back-off time.

Fixes #3284

**Review instructions**

This might be hard to check in real life, sorry…

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -170,7 +170,18 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
      */
     public function get(string $entryIdentifier): string|bool
     {
-        return $this->uncompress($this->redis->get($this->getPrefixedIdentifier('entry:' . $entryIdentifier)));
+        $tries = 1;
+        do {
+            try {
+                return $this->uncompress($this->redis->get($this->getPrefixedIdentifier('entry:' . $entryIdentifier)));
+            } catch (\RedisException  $exception) {
+                if ($tries > 8) {
+                    throw $exception;
+                }
+                usleep($tries^2 * 100000);
+                $tries++;
+            }
+        } while (true);
     }
 
     /**
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -178,7 +178,7 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
                 if ($tries > 8) {
                     throw $exception;
                 }
-                usleep($tries^2 * 100000);
+                usleep($tries**2 * 100000);
                 $tries++;
             }
         } while (true);


Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -170,7 +170,18 @@ private function calculateExpires(string $tag, int $lifetime): int
      */
     public function get(string $entryIdentifier): string|bool
     {
-        return $this->uncompress($this->redis->get($this->getPrefixedIdentifier('entry:' . $entryIdentifier)));
+        $tries = 1;
+        do {
+            try {
+                return $this->uncompress($this->redis->get($this->getPrefixedIdentifier('entry:' . $entryIdentifier)));
+            } catch (\RedisException  $exception) {
```
bwaidelich, line 177: Seems dangerous to me.. Isn't there at least a more specific exception that is thrown when "Redis is not ready (yet)"?

kdambekalns: No, you can only inspect the error message.

But then again, even if it's something else, why not try again? What could be dangerous?

bwaidelich: I'm not sure but I've seen so many errors in the past due to silently caught exceptions that those make me suspicious.
Also we have seen some performance issues with exponential back offs in the past (and are replacing those in Neos 9). For example: Some error in Redis will now lead to a 20s+ delay (if I counted correctly) until it is displayed. A wrong configuration could easily build up to kill the server upon deployments.


What about moving this logic to the [getStatus()](https://github.com/neos/flow-development-collection/blob/db497e5d7e903c1fdfd412ec40e37b316cd97ada/Neos.Cache/Classes/Backend/RedisBackend.php#L578) implementation and make sure that it is called upon deployment?
And/or implement the `WithSetupInterface` and put it there

kdambekalns: Yeah, maybe logging the exceptions would be good – after all, at that point nothing _should_ go wrong.

About killing something on a deployment: The use case for me here is exactly a deployment failure. 🙃

Adding something else to explicitly call won't happen (in "my" current project), as we have that workaround already: Call `doctrine:compileproxies` after deployment until no error appears… 🙈

bwaidelich: > About killing something on a deployment: The use case for me here is exactly a deployment failure.

Right, it fixes the error you described in the issue, but it might create a new one for misconfigured backends.
I won't block this, but I'm not a big fan as you might be able to tell :)
What about turning this into a composition, i.e. introduce some `RetryBackend` that can be wrapped around any other backend (similar to the `MultiBackend`)?

kdambekalns: That is also an option…

robertlemke: I think that the PDO and Redis backend should both have a automatic retry implementation like this by default. In modern hosting environments, there will be movement and it can happen that a Redis is gone for a second or two once in a while. Therefore, it would be nice if a Flow dev wouldn't have to think about this and her application would just work.

I see the point of @bwaidelich that sometimes retries can worsen things. You need to apply the "circuit braker" design pattern for those cases. However, it depends …

I wouldn't like to introduce another wrapper backend since it makes the setup complicated and I think that a reasonable retry is what *most* people want.

How about making this configurable (on / off and number of retries) using the same option names for RedisBackend, PDOBackend and any other backend which could profit from this?

PR #3347, opened 2024-05-02
FEATURE: Option to add an username to the RedisBackend configuration

vcg-development:
Redis has support for authentication which supplies an username+password combination. This PR extends the current RedisBackend so that an username can be supplied in addition to the password. The underlying library already the authentication, hence it simply needs to be configured and passed.

### DIFF
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -34,6 +34,7 @@ use Neos\Error\Messages\Result;
  *  - port:            The TCP port of the redis server (will be ignored if connecting to a socket)
  *  - database:        The database index that will be used. By default,
  *                     Redis has 16 databases with index number 0 - 15
+ *  - username:         The username needed for the redis clients to connect to the server (hostname)
  *  - password:        The password needed for redis clients to connect to the server (hostname)
  *  - batchSize:       Maximum number of parameters per query for batch operations
  *
@@ -68,6 +69,8 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
     protected int $port = 6379;
 
     protected int $database = 0;
+
+    protected string $username = '';
 
     protected string $password = '';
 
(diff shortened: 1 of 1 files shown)


kdambekalns (commented):
Thanks! Some kind of test would be great… Then again, I am not sure we even test the password functionality. 🙄

kdambekalns (approved):
👍 by 👀

kdambekalns:
Two more things that would be great:

- bring this into Flow 8.4
- document the `username` option in `Neos.Flow/Documentation/TheDefinitiveGuide/PartIII/Caching.rst`

Neos.Cache/Classes/Backend/RedisBackend.php
```diff
@@ -535,7 +543,8 @@ private function getRedisClient(): \Redis
             }
         }
 
-        if ($this->password !== '' && !$redis->auth($this->password)) {
+        $credentials = array_values(array_filter([$this->username, $this->password]));
```
robertlemke, line 546: I'd suggest to make this code a bit more readable – as not everyone is familiar with `array_filter()` and the fact that `$redis->auth()` accepts both a string or an array (see [docs](https://github.com/phpredis/phpredis/?tab=readme-ov-file#auth)).

Also, I'd prefer strict comparisons over weak ones like `empty()'.

How about something like this?

```php
        if ($this->username !== '' && $this->password !== '') {
            $result = $redis->auth([$this->username, $this->password]);
            if ($result === false) {
                throw new CacheException(sprintf('Redis authentication failed, using username "%s" and a %s bytes long password.', $this->username, strlen($this->password)), 1725607160);
            }
        } elseif ($this->password !== '') {
            $result = $redis->auth($this->password);
            if ($result === false) {
                throw new CacheException(sprintf('Redis authentication failed, using a %s bytes long password.', strlen($this->password)), 1502366200);
            }
        } else {
            $result = $redis->auth();
            if ($result === false) {
                throw new CacheException('Redis authentication with empty password failed.', 1725607324);
            }
        }
```

vcg-development:
Thx for the feedback.  I will re-open a merge request for flow 8.4 with the suggested changes this afternoon :).

robertlemke:
Thanks for taking care, I'm happy that we'll have that new feature!

PR #3348, opened 2024-05-03
BUGFIX: Pass composer event to update and install scripts

jonnitto:
Currently it is not possible to get the composer event in install and update scripts. With this fix it is possible if the script should be interactive, for example: `$event->getIO()->isInteractive()`

### DIFF
--- a/Neos.Flow/Classes/Composer/InstallerScripts.php
+++ b/Neos.Flow/Classes/Composer/InstallerScripts.php
@@ -89,11 +89,11 @@ class InstallerScripts
         }
 
         if ($operation instanceof InstallOperation && isset($packageExtraConfig['neos/flow']['post-install'])) {
-            self::runPackageScripts($packageExtraConfig['neos/flow']['post-install']);
+            self::runPackageScripts($packageExtraConfig['neos/flow']['post-install'], $event);
         }
 
         if ($operation instanceof UpdateOperation && isset($packageExtraConfig['neos/flow']['post-update'])) {
-            self::runPackageScripts($packageExtraConfig['neos/flow']['post-update']);
+            self::runPackageScripts($packageExtraConfig['neos/flow']['post-update'], $event);
         }
     }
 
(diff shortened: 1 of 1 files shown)


mhsdesign (approved):
Hmm by that we increase our api surface of course and expose composer stuff but its low level either way and when it helps why not?

Neos.Flow/Classes/Composer/InstallerScripts.php
```diff
@@ -121,10 +121,11 @@ protected static function copyDistributionFiles(string $installerResourcesDirect
      * Calls a static method from it's string representation
      *
      * @param string $staticMethodReference
+     * @param PackageEvent $event
      * @return void
      * @throws Exception\InvalidConfigurationException
      */
-    protected static function runPackageScripts(string $staticMethodReference): void
+    protected static function runPackageScripts(string $staticMethodReference, PackageEvent $event): void
```
bwaidelich, line 128: because this method is protected, this is a potentially breaking change.. Not very likely that someone overrode this, but maybe it should be marked as such to make sure?

mhsdesign: Id say this is def not public api because `InstallerScripts` is not api, this is not `public` and frankly i see no way to extend this class and use it?

And more importantly we use `self::runPackageScripts` vs `static::runPackageScripts` which makes it completely impossible to override this part ;)

bwaidelich: I agree that it's not likely. But the `InstallerScripts` class is no final and this method is `protected` – so code that did
```php
class MyInstallerScripts extend InstallerScripts {
  protected static function runPackageScripts(string $staticMethodReference): void {
    InstallerScripts::runPackageScripts($staticMethodReference);
    // my custom logic
  }
}
```
would now break.

Another reason to use `final` classes and `private` methods by default.

Anyways, not blocking this.
Not +1ing it either, because I don't understand the usecase fully yet

jonnitto: The usecase to pass the event is following: I've a `neos-build` package who copy files to root folder, but it ask the user if a) it should copy it, b) override existing files and c) if the user use JavaScript or TypeScript. As `neos-build` packages can't have flow commands, it is done via installer scripts. In the future there will be some more options like which type of CSS framework they want to use and if the root should be the project or a custom package (If you want to develop site package shipped will all settings to build the frontend assets)

Without the event, it is not possible to check, if the install script is running in non-interactive mode

bwaidelich (approved):
Thanks for the explanation, makes sense to me

kitsunet:
yeah I am not concerned binding us to composer, there is no alternative anyways.

kitsunet (approved):

PR #3032, opened 2023-05-02
BUGFIX: No usage of dynamic properties

kitsunet:
This patch resolves deprecation notices on PHP 8.2:

## Dynamic properties in proxy classes
There are some properties that are dynamically declared in proxy classes:
* `Flow_Injected_Properties`
* `Flow_Object_PropertiesToSerialize`
* `Flow_Persistence_RelatedEntities`: this is mostly used inside of the `ObjectSerializationTrait`, so I thought it might make more sense to declare it there instead of adding a property using the `ProxyClassBuilder`

Resolves #2946

## `parent` inside of closures
There are some methods that will be checked against `is_callable` with `parent::`:
* `parent::Flow_Aop_Proxy_buildMethodsAndAdvicesArray`
* `parent::__wakeup`

**Review instructions**
* Set up a flow distribution on PHP 8.2
* Run tests and make sure no deprecation warnings are thrown

**Checklist**

- [x] Code follows the PSR-2 coding style
- [x] Tests have been created, run and adjusted as needed
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [x] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [x] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked wit `!!!` and have upgrade-instructions

PR #3351, opened 2024-05-11
BUGFIX: Adjust to Php 83 `get_parent_class` deprecation

mhsdesign:
see https://www.php.net/manual/en/function.get-parent-class.php

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
+++ b/Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
@@ -409,7 +409,7 @@ class ProxyClassBuilder
             $proxyClass->addProperty($propertyName, var_export($propertyIntroduction->getInitialValue(), true), $propertyIntroduction->getPropertyVisibility(), $propertyIntroduction->getPropertyDocComment());
         }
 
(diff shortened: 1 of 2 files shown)


mhsdesign:
Must have gone missing while upmerging https://github.com/neos/flow-development-collection/pull/3032 i guess

lorenzulrich:
@mhsdesign I suspect that this change has regressions. In https://github.com/neos/flow-development-collection/pull/3406, I backported this change to 8.3 and applied it as a patch in an installation.

Now, when I call an own repository that **extends** a Neos repository and call a method that creates a query, I get this error:

```bash
Call to a member function createQueryForType() on null

  Type: Error
  File: Packages/Framework/Neos.Flow/Classes/Persistence/Repository.php
  Line: 137
```

So somehow DI fails in this case. I suspect it is because PHP <8.3 `get_parent_class()` is not identical to PHP 8.3 `get_parent_class($this)`.

https://www.php.net/manual/en/function.get-parent-class.php#refsect1-function.get-parent-class-changelog

lorenzulrich:
It seems that using `method_exists(parent::class` as mentioned [here](https://php.watch/versions/8.3/get_class-get_parent_class-parameterless-deprecated#get_parent_class-replacement) seems to work in my situation. However I'd be surprised if that part of Flow wasn't covered by a test.

kdambekalns:
As mentioned in the post you linked, `parent::class` fails without a parent and the proper replacement would be `get_parent_class($this)`. Also the PHP documentation says:

>  previously, calling this function inside a class returned the name of that class. 

And inside a class, `$this` is, well, "that class". Also, https://3v4l.org/6TtkP shows no difference for PHP < 8.3

kitsunet:
Given that this here happens in a proxy class we can safely assume a parent::class cannot fail?!

kdambekalns:
Yep, in this case, true. I was more trying to prove that the change as a reason for the "mthod call on null" seems… strange?!

PR #3358, opened 2024-05-25
BUGFIX: Do proper resolving of FusionPathProxy

### LINKED ISSUE #3357: BUG: Fluid does not correctly resolve path references on FusionPathProxy

### Current Behavior

This is probably a variant of an issue thought to be fixed with #3041.

As I found out today, this does not work when you use {image.title} in Fluid and image is a FusionPathProxy:

1. `StandardVariableProvider` checks for `ArrayAccess`, `FusionPathProxy` provides that. `offsetExists()` and `offsetGet()` only check for the offset directly in the partial fusion tree – there we have no title
2. Now `$subject` is an object (our `FusionPathProxy`), but has no matching get/set/is/property
3. Our `TemplateVariableContainer` is getting null back, and thus exits in https://github.com/neos/flow-development-collection/blob/129306a2f5374c07c6b2450b1a5ccf649f16892a/Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php#L37 – that results in null again.
4. title ends up as null

### Expected Behavior

The title should be taken from the assigned image.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Flow: 9.0.0-beta10
- PHP: 8.3
```

### Anything else?

_No response_

kdambekalns:
Using `{image.title}` in Fluid when the image is a `FusionPathProxy` does not work. The result is simply `null` instead of the image title.

This change fixes that by moving more code down into our own custom `TemplateVariableContainer` from the `StandardVariableProvider`.

Fixes neos/flow-development-collection#3357

**Review instructions**

The fixed issue contains instructions on how to reproduce this.

**Checklist**

### DIFF
--- a/Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php
+++ b/Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Neos\FluidAdaptor\Core\ViewHelper;
 
 /*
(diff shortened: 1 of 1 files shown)


Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php
```diff
@@ -13,32 +14,77 @@
...
 use TYPO3Fluid\Fluid\Core\Variables\StandardVariableProvider;
-use TYPO3Fluid\Fluid\Core\Variables\VariableProviderInterface;
 
 /**
  * Provides the variables inside fluid template. Adds TemplateObjectAccessInterface functionality.
  *
  * @api
  */
-class TemplateVariableContainer extends StandardVariableProvider implements VariableProviderInterface
+class TemplateVariableContainer extends StandardVariableProvider
```
kdambekalns, line 23: The interface is implemented by the extended class already…

Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php
```diff
@@ -13,32 +14,77 @@
...
+                if (method_exists($subject, $getMethod)) {
+                    $subject = $subject->$getMethod();
+                    continue;
+                }
+                $isMethod = 'is' . $upperCasePropertyName;
+                if (method_exists($subject, $isMethod)) {
+                    $subject = $subject->$isMethod();
+                    continue;
+                }
+                $hasMethod = 'has' . $upperCasePropertyName;
```
mhsdesign, line 64: is it worth a consideration to use object access here?
And also can it be that we lost certain features over time in fluid like https://github.com/neos/flow-development-collection/pull/108?
I cant really judge and am not that invested into it and neither know whats best. But it seems its a bit back and forth ... we seemed to had already such manual implementation once and it got lost it seems?

dlubitz:
Looks reasonable to me.

dlubitz (approved):

Neos.FluidAdaptor/Classes/Core/ViewHelper/TemplateVariableContainer.php
kdambekalns: I did very consciously copy the method as unchanged as possible, to make the difference clearly visible and easy to understand. This will make maintenance easier for our future selves.

About #108 – yes, it seems that was lost. But given the decline of Fluid, I have little interest in fixing that nowish, and certainly not in the scope of this PR…

Issue #2967, opened 2023-02-24
Feature: Add more useful/modern eel helper methods (see es2020)


mhsdesign:
This list can become endless but a good start would be to have these readily at hand:

```
String.includes()
Array.includes()
```

The spec of the to be added methods should be taken from MDN docs as usual.

NamioL:
Do you know if it is still relevant? Can I prepare PR for that?

mhsdesign:
oh yes very much :D id love to see a pr @NamioL ❤️

Kleisli:
I still have an conceptual question here..

If we stick to the JS interfaces, as suggested, we are missing handy options like the $strict parameter of in_array()
- could we include the $strict parameter anyway, just after $fromIndex?
- we could also add another Helper `Array.includesStrict()`
- should we have two different versions, one according to the PHP interface and one according to the JS interface? I doubt that, but with `String.substring()` and `String.substr()` we have exactly the same.

My opinion: If there are similar functions in PHP and JS, I would combine the parameters as in the first bullet. and if they are incompatible, like substring/substr I would keep or add both.

mhsdesign:
hmm regarding flags. I always dislike flags. Especially when designing an API or encapsulating lower level internals which we do here i think we should really have to expose a feature alternating flag. For most if not many cases it would be in my opinion much better to set a sane default which should be strict mode for `array_includes` by default.

About substr vs substring according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr both actually existed but substr is deprecated.

So we should definitely always check against https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array etc ;)

PR #3359, opened 2024-06-01
FEATURE add `includes` Eel Helpers for String and Array

Kleisli:
resolved:https://github.com/neos/flow-development-collection/issues/2967
Added two new EelHelpers following the interface of the JavaScript functions String.prototype.includes() and Array.prototype.includes()
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Eel/Classes/Helper/ArrayHelper.php
+++ b/Neos.Eel/Classes/Helper/ArrayHelper.php
@@ -27,6 +27,27 @@ use Neos\Eel\ProtectedContextAwareInterface;
  */
 class ArrayHelper implements ProtectedContextAwareInterface
 {
+
+    /**
+     * determines whether an array includes a certain value among its entries,
+     * returning true or false as appropriate.
+     *
+     * If fromIndex is non-negative, the sequence will start at that offset in the array.
+     * If fromIndex is negative, the sequence will start that far from the end of the array.
+     *
+     * @param iterable $array
+     * @param mixed $searchElement The value to search for.
+     * @param int $fromIndex Zero-based index at which to start searching
+     * @return bool
+     */
+    public function includes(iterable $array, mixed $searchElement, int $fromIndex = 0): bool
+    {
+        if ($array instanceof \Traversable) {
+            $array = iterator_to_array($array);
+        }
+        return in_array($searchElement, (array)array_slice($array, $fromIndex));
+    }
+
     /**
      * Concatenate arrays or values to a new array
      *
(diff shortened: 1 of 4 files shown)


mhsdesign:
FYI just made this wip because it has to land in 8.4 but were working on branching it of

Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -27,6 +27,27 @@
...
+     * @param mixed $searchElement The value to search for.
+     * @param int $fromIndex Zero-based index at which to start searching
+     * @return bool
+     */
+    public function includes(iterable $array, mixed $searchElement, int $fromIndex = 0): bool
+    {
+        if ($array instanceof \Traversable) {
+            $array = iterator_to_array($array);
+        }
+        return in_array($searchElement, (array)array_slice($array, $fromIndex));
```
mhsdesign, line 48: we should probably use the 3 param (strict) to true or is the proposed way closer to ECMASCRIPT?

... also the `(array)` casts looks a little weird and im wondering in general if we should optimize and early return for the `fromIndex=0` case?

mhsdesign (commented):
Thanks for working on this!. I havent checked but the `includes` was just an example case.
Are there more helpful helpers that were missing?

Also i think we talked about that `String.substr` is deprecated in ecmascript and we should probably also add all deprecations like this here.

Or if you dont have the time for it we could of course only include those two helpers ... ^^

Neos.Eel/Classes/Helper/ArrayHelper.php
```diff
@@ -27,6 +27,27 @@
...
+     *
+     * @param iterable $array
+     * @param mixed $searchElement The value to search for.
+     * @param int $fromIndex Zero-based index at which to start searching
+     * @return bool
+     */
+    public function includes(iterable $array, mixed $searchElement, int $fromIndex = 0): bool
+    {
+        if ($array instanceof \Traversable) {
+            $array = iterator_to_array($array);
```
sjsone, line 46:
```php
iterator_to_array($array, false);
```

Should be used so duplicate keys will not be overwritten as we do not care about keys here.  
Most of the time this shouldn't be a problem but if someone uses a home brewed `\Traversable` and re-uses keys that's a very hard bug to find.

mhsdesign: jup good catch, i also had this dilemma once and its really nasty to debug ^^

PR #3360, opened 2024-06-05
BUGFIX: ReflectionService must always initialize

kitsunet:
The ReflectionService lazy loads reflection data from cache, but every method making use of the data needs to call initialize.
This change adds missing calls that seem to never happen first in regular flow applications. Still better to prevent broken reflection in case we optimize other uses in the future.

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1153,6 +1153,10 @@ class ReflectionService
      */
     public function getClassSchema($classNameOrObject)
     {
+        if (!$this->initialized) {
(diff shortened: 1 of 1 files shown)


kitsunet:
Note a lot can be made nicer in ReflectionService, I deliberately didn't touch anything else though, as with the 9.0 -> 8.4 backport things will change here and I would do another pass in 9.0 to remove freezing packages and some other cleanup.

grebaldi (approved):
Makes sense, :+1: by :eyes:

bwaidelich (approved):

kdambekalns (commented):
Post-merge 👍

Issue #3364, opened 2024-06-10
BUG: Send HTTP status code 500 when Bootstrap::ensureRequiredEnvironment() fails


christophlehmann:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`Flow could not create the directory "/var/www/html/Data/Persistent". Please check the file permissions manually or run "sudo ./flow flow:core:setfilepermissions" to fix the problem. (Error #1347526553)`.

This critical error message results with exit code `1` (= not successful) on cli but on a web request i get a HTTP 200 response.

### Expected Behavior

I get a HTTP 500 response for such failing web requests. This can be achieved with `http_response_code(500);`

### Environment

```markdown
- Neos: 8.3
- PHP: 8.3
```

PR #3365, opened 2024-06-10
BUGFIX: Streamline exit(1) to web response with HTTP status code 500

christophlehmann:
With this critical error messages like the following now sends the HTTP status code 500 instead 200 in order to state something is not okay on the server side:

`Flow could not create the directory "/var/www/html/Data/Persistent". Please check the file permissions manually or run "sudo ./flow flow:core:setfilepermissions" to fix the problem. (Error #1347526553)`

Resolved: #3364

**Checklist**


kitsunet (approved):
Looks like a great solution, thanks, this is all early enough to not interfere with anything else.

kitsunet:
I would merge this in 24h if no one else has a look here.

kitsunet:
This technically belongs into 8.3 unless there are strong reasons for it to go somewhere else, as everythign else is in security fix only mode already. I would do that on merge then.

kdambekalns (approved):

PR #3367, opened 2024-06-11
FEATURE: Introduce --help flag option for existing CLI commands

crydotsnake:
**Upgrade instructions**

_None_

**Review instructions**

This change introduces a new way of using the help function for existing Flow/Neos CLI commands.

Currently you always used:

```bash
./flow help user:create
```

With this change it is possible to use the new `--help` flag as an alternative at **the end** of a CLI command:

```bash
./flow user:create --help
```

But of course the first way will also still work!

### Demo

https://github.com/neos/flow-development-collection/assets/39345336/4f93f5cf-0435-4344-b37a-0a76b6df7824

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Cli/RequestBuilder.php
+++ b/Neos.Flow/Classes/Cli/RequestBuilder.php
@@ -118,13 +118,6 @@ class RequestBuilder
         $request = new Request();
         $request->setControllerObjectName(HelpCommandController::class);
 
-        if (is_array($commandLine) && in_array('--help', $commandLine, true)) {
-            $request->setControllerCommandName('help');
-            $request->setArguments(['commandIdentifier' => $commandLine[0]]);
-            $request->setControllerObjectName(HelpCommandController::class);
-            return $request;
-        }
-
         if (is_array($commandLine) === true) {
             $rawCommandLineArguments = $commandLine;
         } else {
@@ -147,14 +140,23 @@ class RequestBuilder
                 }
             }
         }
-        if (count($rawCommandLineArguments) === 0) {
+        $firstArgument = count($rawCommandLineArguments) ? trim(array_shift($rawCommandLineArguments)) : null;
+        if (
+            $firstArgument === null
+            || $firstArgument === '--help'
+        ) {
             $request->setControllerCommandName('helpStub');
 
             return $request;
         }
-        $commandIdentifier = trim(array_shift($rawCommandLineArguments));
+        if (in_array('--help', $rawCommandLineArguments, true)) {
+            $request->setControllerCommandName('help');
+            $request->setArguments(['commandIdentifier' => $firstArgument]);
+            return $request;
+        }
+
         try {
-            $command = $this->commandManager->getCommandByIdentifier($commandIdentifier);
+            $command = $this->commandManager->getCommandByIdentifier($firstArgument);
         } catch (CommandException $exception) {
             $request->setArgument('exception', $exception);
             $request->setControllerCommandName('error');
@@ -195,6 +197,9 @@ class RequestBuilder
         $requiredArguments = [];
         $optionalArguments = [];
         foreach ($commandMethodParameters as $parameterName => $parameterInfo) {
+            if ($parameterName === 'help') {
+                throw new \RuntimeException(sprintf('The option --help is reserved in %s::%s', $controllerObjectName, $commandMethodName), 1730715152);
+            }
             if ($parameterInfo['optional'] === false) {
                 $requiredArguments[strtolower($parameterName)] = [
                     'parameterName' => $parameterName,
--- a/Neos.Flow/Classes/Cli/RequestBuilder.php
+++ b/Neos.Flow/Classes/Cli/RequestBuilder.php
@@ -118,8 +118,7 @@ class RequestBuilder
         $request = new Request();
         $request->setControllerObjectName(HelpCommandController::class);
 
-        // @TODO: Can this be implemented even better/cleaner?
-        if (in_array('--help', $commandLine)) {
+        if (is_array($commandLine) && in_array('--help', $commandLine, true)) {
             $request->setControllerCommandName('help');
             $request->setArguments(['commandIdentifier' => $commandLine[0]]);
             $request->setControllerObjectName(HelpCommandController::class);
--- a/Neos.Flow/Classes/Cli/RequestBuilder.php
+++ b/Neos.Flow/Classes/Cli/RequestBuilder.php
@@ -118,6 +118,14 @@ class RequestBuilder
         $request = new Request();
         $request->setControllerObjectName(HelpCommandController::class);
 
+        // @TODO: Can this be implemented even better/cleaner?
+        if (in_array('--help', $commandLine)) {
+            $request->setControllerCommandName('help');
+            $request->setArguments(['commandIdentifier' => $commandLine[0]]);
+            $request->setControllerObjectName(HelpCommandController::class);
+            return $request;
+        }
+
         if (is_array($commandLine) === true) {
             $rawCommandLineArguments = $commandLine;
         } else {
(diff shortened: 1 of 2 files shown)


crydotsnake:
## Open TODOs

I'm still thinking about if i could implement this better.

Also i still have to think about how to improve the description for the `--help` flag. Because it is not even with the other options. It looks different because the description for the `--help` attribute is implemented in the HelpCommandController. And not like the others directly in the method signature.

WindowsTerminal_48GwdZh4d2

And also the tests needs to be adjusted.

jonnitto:
As it is not possible to combine `--help` with other options, I would remove the description.

Neos.Flow/Classes/Cli/RequestBuilder.php
```diff
@@ -118,6 +118,14 @@ public function build($commandLine): Request
         $request = new Request();
         $request->setControllerObjectName(HelpCommandController::class);
 
+        // @TODO: Can this be implemented even better/cleaner?
```
kitsunet: I don't really see how, AFAIK we do not have a concept for such global things, but I'd say whatever _next_ thing we want to implement that is global (-v(v(v)) maybe???) we can then think about a better structure.

kitsunet (commented):
I like it, pretty simple really, I would approve but wonder if we shouldn't safeguard users who actually used help as argument in their command because that would be pretty breaking for them.

crydotsnake:
> but wonder if we shouldn't safeguard users who actually used help as argument in their command because that would be pretty breaking for them.

Good point! i havent thought about this so far.. How could we best do this?

Neos.Flow/Classes/Cli/RequestBuilder.php
crydotsnake: I agree

Neos.Flow/Classes/Cli/RequestBuilder.php
```diff
@@ -118,6 +118,14 @@ public function build($commandLine): Request
...
+        if (in_array('--help', $commandLine)) {
```
dlubitz: Tests are failing, because, `$commandLine` could be a string. It's not guaranteed to be an array.

crydotsnake:
> but wonder if we shouldn't safeguard users who actually used help as argument in their command because that would be pretty breaking for them.

@kitsunet Do you maybe have an idea, how we could solve this the best way?

dlubitz:
I think @kitsunet is right and we can't provide safely this global flag without potentially breaking dedicated help arguments of the commands.

Neos.Flow/Classes/Cli/RequestBuilder.php
```diff
@@ -118,6 +118,13 @@ public function build($commandLine): Request
         $request = new Request();
         $request->setControllerObjectName(HelpCommandController::class);
 
+        if (is_array($commandLine) && in_array('--help', $commandLine, true)) {
+            $request->setControllerCommandName('help');
+            $request->setArguments(['commandIdentifier' => $commandLine[0]]);
+            $request->setControllerObjectName(HelpCommandController::class);
+            return $request;
+        }
```
mhsdesign: i would put this logic at below line 155 of `$commandIdentifier` so we can use this variable and also use `$rawCommandLineArguments` meaning we support the `$commandLine` being a string possibly.

Also the case `flow --help` will with this implementation result in an error as there is no command id named `--help`. The general help should be shown i suppose.

In the end this will make `--help` an reserved argument and we have to agree on that (which is likely but still breaking? so `!!!`?)

Also this raw lookup against `--help` seems a little hacky. Like flow does support this short notation of boolean parameters but `--help=true` or `--help=false` are technically possible and will now still end up being passed to the final controller do we want that?

bwaidelich:
Thanks for this contribution. This is a feature that I'm missing forever. It's small, but it really enhances UX.

If you're concerned about existing commands with a `--help` argument we could target `9.0`.
But maybe we can easily detect these cases and ignore the global flag for those?

Lastly, I agree to @mhsdesign's comments about the check against `--help` – but personally I would be OK with that if the choice is hacky but working or not at all :)

crydotsnake:
> Thanks for this contribution. This is a feature that I'm missing forever. It's small, but it really enhances UX.
> 
> If you're concerned about existing commands with a `--help` argument we could target `9.0`. But maybe we can easily detect these cases and ignore the global flag for those?
> 
> Lastly, I agree to @mhsdesign's comments about the check against `--help` – but personally I would be OK with that if the choice is hacky but working or not at all :)

> But maybe we can easily detect these cases and ignore the global flag for those?

Yes, that would be awesome. As i think this is really a good improvement.

robertlemke:
I wonder if there's actually any Flow application out there which implemented its own `--help` argument. And even if there is, I think it would be totally fine if this new feature would just override the custom implementation and that's it. It shouldn't cause an error, but I personally would be totally fine, if the custom argument implementation would just be ignored.

In that sense, what's missing to get this into 8.4? I'd love to have it in soon, so we can making plans for a release.

mhsdesign:
Yeah sorry i didnt meant to side track it with asking what happens if `--help` is a real argument. 
> In the end this will make `--help` an reserved argument and we have to agree on that (which is likely but still breaking? so `!!!`?)

But by that i was just trying to make a point that some things are not thought out yet ^^


> i would put this logic at below line 155 of `$commandIdentifier` so we can use this variable and also use `$rawCommandLineArguments` meaning we support the `$commandLine` being a string possibly.
> 
> Also the case `flow --help` will with this implementation result in an error as there is no command id named `--help`. The general help should be shown i suppose.
> 
>  ...
> 
> Also this raw lookup against `--help` seems a little hacky. Like flow does support this short notation of boolean parameters but `--help=true` or `--help=false` are technically possible and will now still end up being passed to the final controller do we want that?

And the other points are still valid imo and have not been tackled.

Neos.Flow/Classes/Cli/RequestBuilder.php
crydotsnake: > i would put this logic at below line 155 of $commandIdentifier so we can use this variable and also use $rawCommandLineArguments meaning we support the $commandLine being a string possibly.

So my check should also check if `$rawCommandLineArguments` is a string? Dont fully understand what you mean.

> Also the case flow --help will with this implementation result in an error as there is no command id named --help. The general help should be shown i suppose.

100%. Is there a way to trigger the default `flow help` via PHP?

robertlemke (changes requested):
Generally fine, but the "--help" help text doesn't fit well into the rest of the description, because there's no indentation:

image

I'll push a fix in a minute ...

robertlemke:
It now adds the following at the bottom of the general help screen (list of commands):
image

And the command-specific help looks like this:
image

Let's merge this?

robertlemke (approved):

Neos.Flow/Classes/Cli/RequestBuilder.php
```diff
@@ -188,6 +197,9 @@ protected function parseRawCommandLineArguments(array $rawCommandLineArguments,
         $requiredArguments = [];
         $optionalArguments = [];
         foreach ($commandMethodParameters as $parameterName => $parameterInfo) {
+            if ($parameterName === 'help') {
+                throw new \RuntimeException(sprintf('The option --help is reserved in %s::%s', $controllerObjectName, $commandMethodName), 1730715152);
+            }
```
mhsdesign, line 202: is this oke? :D

mhsdesign (approved):
btw in b9fc7eafdd5a6898d297b0e4b1f57324df9ab1ac i also adjusted a few things earlier:

- allows `flow --help`
- restricts that no command controller uses `$help` as custom option
- remove duplicate set `HelpCommandController`
- centralise $firstArgument "parsing"
- respect that $commandLine could be a string

Does not handle `flow cache:list --help=true` as this might be bungus :D

... so definitely +1 and tested this ... and if youre all okay with my `RuntimeException('The option --help is reserved ...')` then lets have it :)

crydotsnake:
Thanks u both! @mhsdesign @robertlemke ❤️ Sorry i havent continue working on that 😅

PR #3371, opened 2024-06-22
!!!TASK: Bootstrap instantiates request handlers internally

kitsunet:
It seems unnecessarily complex to have code register a request handler as build instance. Instead it is now registered by class name. The Flow Bootstrap will then create instances while determining the best matching request handler. To facilitate this being lazy the RequestHandlerInterface::getPriority method is now declared static so that we can decide beforehand if the priority is higher than any previously found best match.
Depending on possibly request handlers and specifically in all preselected cases this all reduces the amount of created objects and possibly the amount of canHandle calls. A microptimization but which affects all requests. Primary concern was to make registration easier and internalize instantiation.

As small simplification the
`Bootstrap->setPreselectedRequestHandlerClassName` will also register the given class name as possible request handler removing one step.

This is breaking if you have your own request handler implementation. The `getPrioritty` method needs to declare an `int` return type and must be static. Additionally the registration happens via `Bootstrap->registerRequestHandlerClassName(YourRequestHandler::class)`.
Also the RequestHandler now needs a static constructor 
`RequestHandlerInterface::fromBootstrap(Bootstrap $bootstrap): self` that the Bootstrap will call when instantiating one.

If the RequestHandler for some reason was used as a bridge to the outside world and the instance carried more information besides the bootstrap, you can "sneak in" your instance with anything you want by setting it via 
`Bootstrap->setEarlyInstance(YourRequestHandler::class, $yourInstance)` and this will be used instead of creating a new instance via the static constructor. You still need to register the class as possible request handler though.

### DIFF
--- a/Neos.Flow/Classes/Core/Bootstrap.php
+++ b/Neos.Flow/Classes/Core/Bootstrap.php
@@ -31,28 +31,22 @@ use Neos\Utility\Files;
  */
 class Bootstrap
 {
-    const RUNLEVEL_COMPILETIME = 'Compiletime';
-    const RUNLEVEL_RUNTIME = 'Runtime';
+    public const RUNLEVEL_COMPILETIME = 'Compiletime';
+    public const RUNLEVEL_RUNTIME = 'Runtime';
+
+    protected ApplicationContext $context;
 
     /**
-     * @var ApplicationContext
+     * @var array<int, class-string>
      */
-    protected $context;
+    private array $requestHandlerClassNames = [];
 
     /**
-     * @var array<class-string<RequestHandlerInterface>, RequestHandlerInterface>
+     * @var class-string<RequestHandlerInterface>|null
      */
-    protected $requestHandlers = [];
+    protected string|null $preselectedRequestHandlerClassName = null;
 
-    /**
-     * @var class-string<RequestHandlerInterface>
-     */
-    protected $preselectedRequestHandlerClassName;
-
-    /**
-     * @var RequestHandlerInterface
-     */
-    protected $activeRequestHandler;
+    protected RequestHandlerInterface $activeRequestHandler;
 
     /**
      * The same instance like $objectManager, but static, for use in the proxy classes.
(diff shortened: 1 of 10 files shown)


kitsunet:
### This cannot ever be green on CI as the necessary changes to the BuildEssentials (register/preselect testing request handlers) are a hard block in either direction, whichever comes first breaks it. 

So you would have to run tests locally or trust me and we fix anything that comes up after....

Neos.Flow/Classes/Core/Bootstrap.php
```diff
@@ -404,27 +403,54 @@ public function getObjectManager(): ObjectManagerInterface
...
+     * @param class-string $className
+     * @return RequestHandlerInterface
+     */
+    private function getRequestHandlerInstance(string $className): RequestHandlerInterface
+    {
+        if (!in_array($className, $this->requestHandlerClassNames, true)) {
+            throw new FlowException(sprintf('A request handler "%s" was never registered', $className), 1719060769);
+        }
+
+        $requestHandler = new $className($this);
```
mhsdesign: im not so found of the implicit `public function __construct(Bootstrap $bootstrap)` part now ... at least its not declared in the `RequestHandlerInterface` ... maybe we should introduce a `public static function fromBootstrap(Bootstrap $bootstrap): self;` ?

mhsdesign (commented):
thanks for this ;) We also stumbled upon this complexity for the `neos/setup` part ... this change does make things easier. Though we do have to live with the fact that the 6.0.0 tag of `neos/setup` will not be compatible with Flow 9.0 anymore which it is currently and tagged: https://github.com/neos/setup/blob/ecd193d814aea502aad71b0907f193de08fc9d0d/composer.json#L8 (im stupid sorry lol).

Neos.Flow/Classes/Core/Bootstrap.php
kitsunet: yeah good point, now that I mess with the interface anyways, makes sense. Added now

PR #3375, opened 2024-07-04
FEATURE: Support doctrine/dbal 2.x and 3.x

bwaidelich:
Related: #2637

### DIFF
--- a/Neos.Flow/Classes/Command/DatabaseCommandController.php
+++ b/Neos.Flow/Classes/Command/DatabaseCommandController.php
@@ -124,8 +124,8 @@ class DatabaseCommandController extends CommandController
 
         $statements[] = 'ALTER DATABASE ' . $this->connection->quoteIdentifier($this->persistenceSettings['backendOptions']['dbname']) . ' CHARACTER SET ' . $characterSet . ' COLLATE ' . $collation;
 
-        $tableNames = $this->connection->createSchemaManager()->listTableNames();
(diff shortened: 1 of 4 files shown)


kitsunet:
I will say I am not _really_ up for figuring  out cross compatibility for exceptions and stuff that was deprecated, IMHO it's a HUGE PITA to do so and I woudl rahter clean cut it and say 9.0 uses DBAL ^3(.3) and that's that.

bwaidelich:
@kitsunet I would love to discuss matters with you, but my points briefly:
- IMO a framework should have as little and as relaxed 3rd party dependencies to other packages as possible
- it's a bit unfortunate, that the Flow core has a dependency to `doctrine/dbal` at _all_ but that's obviously nothing we can change in the [forseeable future](https://discuss.neos.io/t/could-we-build-a-version-of-flow-without-doctrine-orm-database-requirement/2187)
- My goal is not to make Flow 9.0 compatible with an ancient doctrine/dbal 2.x but rather: allow Flow 8.4 to be used with doctrine/dbal 3.x. And as far as I see it is quite doable (given how small [your PR](https://github.com/neos/flow-development-collection/pull/2637) and this one is)
- I don't understand why you fear cross-compatibility for exceptions?

bwaidelich:
This was just a test, the actual goal would be to make `8.4` _forwards-compatible_  (#3377)

PR #3379, opened 2024-07-08
!!!TASK: Deprecate outdated doctrine functionality

kitsunet:
In preparation of Flow 9 we deprecate some functionality we expose but that was deprecated in doctrine already, so we will remove it.

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Repository.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Repository.php
@@ -161,6 +161,8 @@ abstract class Repository extends EntityRepository implements RepositoryInterfac
      * @param IterableResult $iterator
      * @param callable|null $callback
      * @return \Generator
+     *
+     * @deprecated Will be removed with Flow 9. The {@see findAllIterator} will return an \iterable directly
      */
     public function iterate(IterableResult $iterator, callable $callback = null): ?\Generator
     {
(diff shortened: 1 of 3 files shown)


kdambekalns (approved):
👍

Neos.Flow/Classes/Persistence/Doctrine/Repository.php
```diff
@@ -161,6 +161,8 @@ public function findAllIterator(): IterableResult
      * @param IterableResult $iterator
      * @param callable|null $callback
      * @return \Generator
+     *
+     * @deprecated Will be removed with Flow 9. The {@see findAllIterator} will return an \iterable directly
      */
     public function iterate(IterableResult $iterator, callable $callback = null): ?\Generator
```
mhsdesign, line 167: nice :D thx for removing this stuffy. Fyi we also have to remove it from the thumbnail repository and others in Neos.Neos ;)

mhsdesign (approved):

Neos.Flow/Classes/Persistence/Doctrine/Repository.php
kitsunet: Yep I know, I know! I would add this to the respective Neos 9 dbal changes then, but first wanted to get this in to be sure I _can_ actually remove it 😆

PR #3383, opened 2024-07-27
BUGFIX: #3303 Flush reflection cache for removed php classes

### LINKED ISSUE #3303: BUG: Reflection cache for class is not flushed on file deletion

mhsdesign:
resolves #3303

There are a few layers to this bugfix.

1. Previously we ignored deleted files (see `file_exists`) and did not flush the `Flow_Reflection_Status` cache accordingly as we are not able to read the namespace from a deleted php file. Now we do a best effort reverse lookup by going through all psr-4 autoload configurations and calculate a namespace under which composer would have found this file.

2. We introduce a state `removedReflectionDataClasses` like `updatedReflectionData` which will be used to check if we have to update the `ReflectionData` cache.

**Upgrade instructions**

**Review instructions**

### How to test

We want to test `getAllImplementationClassNamesForInterface` as by the bugreport. And we can do this simply because we use it at a few places.

1. Open the Neos.Demo site (in development context)
2. Remove a not-so-important FlowQuery operation like `ReferencePropertyOperation` (which is not used on the homepage)
3. Reload the homepage
4. Expect everything be ok

Without this fix there will be an error thrown:

```
Exception in line 58 of Data/Temporary/Development/Cache/Code/Flow_Object_Classes/Neos_Eel_FlowQuery_OperationResolver.php: Class "Neos\ContentRepository\NodeAccess\FlowQueryOperations\ReferencePropertyOperation" not found

69 Neos\Eel\FlowQuery\OperationResolver_Original::buildOperationsAndFinalOperationNames(Neos\Flow\ObjectManagement\ObjectManager)
68 Neos\Eel\FlowQuery\OperationResolver_Original::initializeObject(1)
```

### How the fix works

With the fix applied we get the information that `"/Users/neos/Packages/Neos/Neos.ContentRepository.NodeAccess/Classes/FlowQueryOperations/ReferencePropertyOperation.php"` was removed. Then we can resolve that it belongs to the package `Neos.ContentRepository.NodeAccess` and reverse calculate its namespace/cache identifier (with underscores): `Neos_ContentRepository_NodeAccess_FlowQueryOperations_ReferencePropertyOperation`. Then we can flush the cache accordingly.

### Todo

- [ ] check if this is symlink safe! Like for user package in DistributionPackages

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Cache/CacheManager.php
+++ b/Neos.Flow/Classes/Cache/CacheManager.php
@@ -20,6 +20,8 @@ use Neos\Cache\Frontend\FrontendInterface;
 use Neos\Cache\Frontend\VariableFrontend;
 use Neos\Flow\Configuration\ConfigurationManager;
 use Neos\Flow\Log\Utility\LogEnvironment;
+use Neos\Flow\Monitor\ChangeDetectionStrategy\ChangeDetectionStrategyInterface;
+use Neos\Flow\Package\PackageManager;
 use Neos\Flow\Utility\Environment;
 use Neos\Utility\Files;
 use Neos\Flow\Utility\PhpAnalyzer;
@@ -46,6 +48,11 @@ class CacheManager
      * @var ConfigurationManager
      */
     protected $configurationManager;
+
+    /**
+     * @var PackageManager
+     */
+    protected $packageManager;
 
     /**
      * @var LoggerInterface
@@ -116,6 +123,11 @@ class CacheManager
     public function injectConfigurationManager(ConfigurationManager $configurationManager): void
     {
         $this->configurationManager = $configurationManager;
+    }
+
+    public function injectPackageManager(PackageManager $packageManager): void
+    {
+        $this->packageManager = $packageManager;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Neos.Flow/Classes/Cache/CacheManager.php
```diff
@@ -352,15 +364,35 @@ protected function flushClassCachesByChangedFiles(array $changedFiles): void
...
-                continue;
+            $classNameWithUnderscores = null;
+            if ($status !== ChangeDetectionStrategyInterface::STATUS_DELETED) {
+                $fileContents = file_get_contents($pathAndFilename);
+                $className = (new PhpAnalyzer($fileContents))->extractFullyQualifiedClassName();
+                if ($className === null) {
+                    continue;
+                }
+                $classNameWithUnderscores = str_replace('\\', '_', $className);
+            } else {
```
mhsdesign, line 375: technically we can discuss if we always want to use the psr-4 reverse lookup instead of parsing the file via `PhpAnalyzer` ... but for that we would need to know how stable this logic is... for example regarding symlinks

kitsunet:
I am quite opposed to adding more places where we lookup package classes/class loading to do magic stuff, I think we need to get rid of that concept. 
But more importantly, this screams performance bottleneck with what it does (although I just looked at the code as is and this just adds to what is already there). I think the md5 thing was nicer overall. Probably we should streamline this all to be independent from looking into the class file...

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1833,6 +1837,8 @@ protected function forgetClass($className): void
 
         unset($this->classReflectionData[$className]);
         unset($this->classesCurrentlyBeingForgotten[$className]);
+
+        $this->removedReflectionDataClasses[$className] = true;
```
mhsdesign, line 1841: todo check if `forgetClass` is also called in other occasions and if this flag makes sense in that case.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -2092,7 +2098,7 @@ protected function saveProductionData(): void
      */
     protected function updateReflectionData(): void
```
grebaldi, line 2099: `$this->removedReflectionDataClasses` should be reset to `[]` in here, right?

PR #3386, opened 2024-08-09
FEATURE: Introduce EEL tracer for handling Neos9 deprecations

mhsdesign:
Related https://github.com/neos/neos-development-collection/issues/5022

In todays weekly @bwaidelich and me discussed a concrete way how to log deprecations like `node.indentifier` in Neos 8.4 and 9.0

The idea is to add a tracer to eel, that will be implemented in Fusion. Technically we would need to add an abstraction to Neos.Fusion as well to not access the Node from there as this is architecturally illegal but to simplify the code and in light that this is just considered for temporary time we propose to implement it as such:

```php
<?php

namespace Neos\Fusion\Core;

use Psr\Log\LoggerInterface;

final class Neos9RuntimeMigrationTracer implements EelInvocationTracerInterface
{
    /** @Flow\Inject */

    private const DEPRECATED_NODE_PROPERTIES = [
        'identifier' => true,
        'nodetype' => true,
        // ...
    ];

    public function __construct(
        private readonly string $eelExpression,
        private readonly bool $showMercy
    ) {
    }

    public function recordPropertyAccess(object $object, string $propertyName): void
    {
        if (
            $object instanceof \Neos\ContentRepository\Domain\Model\Node
            && array_key_exists(strtolower($propertyName), self::DEPRECATED_NODE_PROPERTIES)
        ) {
            $this->logDeprecationOrThrowException(
                sprintf('"node.%s" is deprecated in "%s"', $propertyName, $this->eelExpression)
            );
        }
    }

    public function recordMethodCall(object $object, string $methodName): void
    {
    }

    private function logDeprecationOrThrowException(string $message): void
    {
        if ($this->showMercy) {
            $this->logger->debug($message);
        } else {
            throw new \RuntimeException($message);
        }
    }
}
```

and instantiate this `Neos9RuntimeMigrationTracer` (name tbd) in `\Neos\Fusion\Core\Runtime::evaluateEelExpression`

```diff
- return EelUtility::evaluateEelExpression($expression, $this->eelEvaluator, $contextVariables);
+ $tracer =.$this->settings['enableDeprecationTracer'] ? new Neos9RuntimeMigrationTracer($expression, $this->settings['strictEelMode'] ?? false) : null;
+ return EelUtility::evaluateEelExpression($expression, $this->eelEvaluator, $contextVariables, $tracer);
```

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Eel/Classes/EelInvocationTracerInterface.php
+++ b/Neos.Eel/Classes/EelInvocationTracerInterface.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Neos\Eel;
+
+/**
+ * @internal experimental tracer for eel. Could be used for example to collect and log deprecations.
+ */
+interface EelInvocationTracerInterface
+{
+    public function recordPropertyAccess(object $object, string $propertyName): void;
+
+    public function recordMethodCall(object $object, string $methodName): void;
+}
--- a/Neos.Eel/Classes/EelInvocationTracerInterface.php
+++ b/Neos.Eel/Classes/EelInvocationTracerInterface.php
@@ -9,5 +9,13 @@ interface EelInvocationTracerInterface
 {
     public function recordPropertyAccess(object $object, string $propertyName): void;
 
-    public function recordMethodCall(object $object, string $methodName): void;
+    /**
+     * @param array<int, mixed> $arguments
+     */
+    public function recordMethodCall(object $object, string $methodName, array $arguments): void;
+
+    /**
+     * @param array<int, mixed> $arguments
+     */
+    public function recordFunctionCall(callable $function, string $functionName, array $arguments): void;
 }
--- a/Neos.Eel/Classes/Context.php
+++ b/Neos.Eel/Classes/Context.php
@@ -27,16 +27,12 @@ use Neos\Utility\ObjectAccess;
 class Context
 {
     /**
-     * @var mixed
-     */
-    protected $value;
-
-    /**
      * @param mixed $value
      */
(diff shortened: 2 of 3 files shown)


kitsunet (changes requested):
Before/after performance in a reasonably big site please ;) I think as small as this is, it will be a performance hit.

mhsdesign:
Of course if Fusion implements a stupid tracer that is way too slow by doing to many things there will be a performance hit, BUT as i have drawn out above we want to set the tracer fully to `null` in production and thus the optional chaining will just be a true noop:

```php
$this->tracer?->recordMethodCall($this->value, $method);
```

I also considered introducing some NoopTracer but that would be slower i guess.

Neos.Eel/Classes/EelInvocationTracerInterface.php
```diff
@@ -0,0 +1,13 @@
...
+namespace Neos\Eel;
+
+/**
+ * @internal experimental tracer for eel. Could be used for example to collect and log deprecations.
+ */
+interface EelInvocationTracerInterface
+{
+    public function recordPropertyAccess(object $object, string $propertyName): void;
+
+    public function recordMethodCall(object $object, string $methodName): void;
```
mhsdesign: add recordFunctionCall and introduce arguments

kitsunet (approved):
Performance discussed, I am fine with that ✅

I woulnd't call it `showMercy`, but that is not part of this change so fine as the example :D 

I guess lets bring it in...

PR #3387, opened 2024-08-29
TASK: Use `SCAN` for redis flush

kdambekalns:
This should speed up flush operations and lower the load on Redis.

**Review instructions**

The functional tests run `flush()`…

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/RedisBackend.php
+++ b/Neos.Cache/Classes/Backend/RedisBackend.php
@@ -233,10 +233,15 @@ class RedisBackend extends IndependentAbstractBackend implements TaggableBackend
     {
         // language=lua
         $script = "
-        local keys = redis.call('KEYS', ARGV[1] .. '*')
-        for k1,key in ipairs(keys) do
-            redis.call('DEL', key)
-        end
+        local cursor = '0'
+        repeat
+            local result = redis.call('SCAN', cursor, 'MATCH', ARGV[1] .. '*')
+            cursor = result[1]
+            local keys = result[2]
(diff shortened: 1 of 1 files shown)


Sebobo:
Thx sound good! But what does `should speed up` exactly mean in numbers or examples?
What was your use case?

I would like to test this thoroughly with some millions of tags in my projects.

kdambekalns:
We saw high load on a Redis server caused by cache flushes calling `KEYS`. The fact that `KEYS` ist warned about in the documentation made me try this, and so far we saw a lower load. I don't have hard numbers, sadly, as I do not have direct access to the metrics…

Sebobo:
Just to confirm and avoid confusion: This method is only triggered manually in Production correct?
I'm currently trying to come up with a "real" test case for the projects I know.

Sebobo (approved):
I can confirm that with this change, cache entries are still flushed.
Though I can't detect any load or timing changes when testing with some 10.000s of keys.

But the docs mention this to be the better way of doing it, so let's go with it?

kdambekalns:
> Just to confirm and avoid confusion: This method is only triggered manually in Production correct? I'm currently trying to come up with a "real" test case for the projects I know.

In the project this caused issues, we see regular calls to that method. We still need to find out where they come from, to be honest. 🙈

That method is called (only) in `AbstractFrontend->flush()` and that is called automatically only in Develoment context … with the exception of node publishing, which triggers `flush()` on `Flow_Persistence_Doctrine`. Unless something custom in the project causes this, which is probable.

PR #3390, opened 2024-09-03
BUGFIX: Fix support for typo3fluid/fluid 2.15

### LINKED ISSUE #3389: Fluid 2.15 breaks unit tests
The latest [2.15 release of typo3fluid/fluid](https://github.com/TYPO3/Fluid/releases/tag/2.15.0), more specifically this [change set](https://github.com/TYPO3/Fluid/pull/979) breaks our unit tests leading to an error:

> Error: Call to a member function evaluateChildNodes() on null
> Packages/Libraries/typo3fluid/fluid/src/Core/ViewHelper/AbstractViewHelper.php:329

The root cause is the changed implementation of `AbstractViewHelper::buildRenderChildrenClosure()` that no longer invokes `renderChildren()` and expects the `viewHelperNode` field to be set (which isn't the case when mocking the syntax tree)

**Note:** I tagged this with 9.0 because in older versions of Flow we decided to pin Fluid to an older version, so they are not directly affected (see https://github.com/neos/flow-development-collection/pull/3038)

bwaidelich:
Adjusts tests that mocked `AbstractViewHelper::renderChildren()` that is no longer invoked with version 2.15

Fixes: #3389

### DIFF
--- a/Neos.FluidAdaptor/Tests/Unit/ViewHelpers/ViewHelperBaseTestcase.php
+++ b/Neos.FluidAdaptor/Tests/Unit/ViewHelpers/ViewHelperBaseTestcase.php
@@ -17,6 +17,8 @@ use Neos\FluidAdaptor\Core\ViewHelper\AbstractTagBasedViewHelper;
 use Neos\FluidAdaptor\Core\ViewHelper\AbstractViewHelper;
 use Neos\Http\Factories\ServerRequestFactory;
 use Neos\Http\Factories\UriFactory;
+use PHPUnit\Framework\MockObject\MockObject;
+use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\ViewHelperNode;
 use TYPO3Fluid\Fluid\Core\ViewHelper\TagBuilder;
 
 /**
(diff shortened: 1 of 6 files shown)


Neos.FluidAdaptor/Tests/Unit/ViewHelpers/ViewHelperBaseTestcase.php
```diff
@@ -175,4 +177,12 @@ protected function prepareArguments(\TYPO3Fluid\Fluid\Core\ViewHelper\AbstractVi
         $viewHelper->initialize();
         return $viewHelper;
     }
+
+    protected function simulateViewHelperChildNodeContent(AbstractViewHelper&MockObject $viewHelper, string|int $content): void
+    {
+        $viewHelper->method('renderChildren')->willReturn($content);
```
bwaidelich, line 183: Note: This line is no longer required with Fluid 2.15 – but it allows the tests to pass with older versions of Fluid

dlubitz (approved):
Thank you

lolli42:
Quick note.
You may be better off avoiding mocks in VH tests.
Both fluid standalone and typo3 core basically transitioned to "functional" tests: Feed a TemplateView with some fluid template test string that calls the VH, render(), assert result string.
This turns out to be much more easy to read and avoids internal knowledge within tests.
Bonus: It's trivial to do this twice to test if a VH behaves identical with "template not cached and template cached" (we found and fixed quite a few issues due to this).

PR #3391, opened 2024-09-04
Flow: Error message from FileSizeValidator is wrong

Schuppel1:
The Error message larger and smaller are swapped. 

fixed it. 

Code: 
```php
        if ($this->options['minimum'] && $filesize < $this->options['minimum']) {
            $this->addError('The file is smaller than allowed.', 1677934908);
            return;
        }
        if ($this->options['maximum'] && $filesize > $this->options['maximum']) {
            $this->addError('The file is larger than allowed.', 1677934903);
            return;
        }
´´´


dlubitz (approved):
Thank @Schuppel1 you for your contribution! Makes sence.

bwaidelich (approved):

PR #3396, opened 2024-09-26
TASK: Adjust log level of authentication messages

robertlemke:
The level of "successfully authenticated token" messages was lowered to "info" and "wrong credentials" messages to "notice".

Successful authentication of a user is a normal operation and can happen many times within minutes or seconds in highly frequented websites and applications. More often than not, those messages don't require the attention of an administrator. To make it easier for filtering out those messages, the level should really be "info", since it is informational.


kdambekalns (approved):
Absolutely makes sense!

crydotsnake (approved):

PR #3399, opened 2024-10-11
FEATURE: flow query directy field access without get 0

mhsdesign:
The idea was to get rid of the unintuitive `get(0)` calls in cases were now we would like to fetch node fields directly like their identifier:

Currently:

```
q(node).parent().get(0).aggregateId
``` 

Proposed:

```
q(node).parent().aggregateId
``` 

The need for that was created as `.property('_identifier')` does not work anymore.

Further while we thought about this we came to the idea that all final operations - also the `property` operation could be replaced:

Currently:

```
q(node).parent().property('foo-bar')
``` 

Proposed:

```
q(node).parent().properties['foo-bar']
``` 

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Eel/Classes/FlowQuery/FlowQuery.php
+++ b/Neos.Eel/Classes/FlowQuery/FlowQuery.php
@@ -157,12 +157,6 @@ class FlowQuery implements ProtectedContextAwareInterface, \IteratorAggregate, \
      */
     public function __call($operationName, array $arguments)
     {
-        if (str_starts_with($operationName, 'get') && !$this->operationResolver->hasOperation($operationName)) {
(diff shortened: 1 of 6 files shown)


mhsdesign:
i assume we need to discuss this further (see also the ideas from slack https://neos-project.slack.com/archives/C050KKBEB/p1728648485749549)

especially promoting `q(node).properties.title` would make integer arrays / lists like in Neos.Neos in flowquery first level citizen but a real array like `q({foo: 'bar'}).foo` would NOT work, as it would try to access `foo` on the first element:  `q({foo: {foo: 'value'}}).foo == value` this is all a little odd and we have to see if we can live with that

PR #3401, opened 2024-10-12
BUGFIX: Static compile attribute routes

### LINKED ISSUE #3400: BUG: Attribute driven routes are ignored in production context

### Current Behavior

Routes from attributes are not detected

### Expected Behavior

Routes from attributes should be respected and listable via `./flow routing:list`

### Steps To Reproduce

Add `Route` attributes to some MVC action and configure the AttributeRoutesProviderFactory as [described](https://flowframework.readthedocs.io/en/9.0/TheDefinitiveGuide/PartIII/Routing.html#subroutes-from-annotations)

### Environment

```markdown
- Flow: 9.0@dev
```

### Anything else?

I could pin point the issue down to this line https://github.com/neos/flow-development-collection/blob/ee8a8127e1d77908b5f44786a1ba4d6f2500c583/Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php#L81 that returns an empty array in production context, probably because it is invoked before the `ReflectionService` is fully initialized.

Related: #2059

kitsunet:
The necessary reflection data used to build the routes from attributes is not available at (Production) runtime. This is an issue in itself but not trivial to fix, therefore we fix this here by using the `CompileStatic` attribute to bring the necessary data over from compile time.

Fixes: #3400

### DIFF
--- a/Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
+++ b/Neos.Flow/Classes/Mvc/Routing/AttributeRoutesProvider.php
@@ -78,9 +78,7 @@ final class AttributeRoutesProvider implements RoutesProviderInterface
     public function getRoutes(): Routes
     {
         $routes = [];
-        $annotatedClasses = $this->reflectionService->getClassesContainingMethodsAnnotatedWith(Flow\Route::class);
-
-        foreach ($annotatedClasses as $className) {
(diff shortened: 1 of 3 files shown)


mhsdesign (approved):
Looks good so far (not tested) and i think the size of the array that will be precompiled into php is also acceptable as we dont precompile the whole routes but only the ones annotated with attributes.

PR #3413, opened 2024-11-07
TASK: Uniquify exception codes in Abstract(Template)View

kdambekalns:
Exception codes are supposed to be unique, so this makes sure those places get a code of their own.

**Upgrade instructions**

If you rely on that exception code (`1359625876`) for whatever reason, consider taking the new codes into account, too:

- `1359625877`
- `1359625878`
- `1359625879`
- `1359625880`
- `1359625881`
- `1359625882`
- `1359625883`

**Review instructions**

You could check if any of the new codes is in use already, in turn… 😬

**Checklist**


markusguenther (approved):
That definitely makes sense.

PR #3410, opened 2024-10-30
WIP: Extended Privilege syntax

bwaidelich:
Extend privilege syntax to allow for custom, dedicated options:

e.g. instead of:

```yaml
privilegeTargets:
  'Neos\Neos\Security\Authorization\Privilege\EditNodePrivilege':
    'Neos.Demo:EditBlogNodes':
      matcher: 'default:blog'
```

this would allow for:

```yaml
privilegeTargets:
  'Neos\Neos\Security\Authorization\Privilege\EditNodePrivilege':
    'Neos.Demo:EditBlogNodes':
      contentRepository: 'default'
      subtreeTag: 'blog'
```

PR #3415, opened 2024-11-11
FEATURE: Introduce `RoleId` and `RoleIds` value objects

### LINKED ISSUE #3414: Introduce `RoleIds` value object
`Security\Context::getRoles()` currently returns an array of (expanded) `Role` instances.

It is used all over the place even though:
- in most places we're only interested in the _identifier_ of the expanded roles, not about the (mutable) `Role` instances
- arrays in APIs are bad because types.. :)

To make this a b/c change, I would suggest to introduce the _additional_  methods:

```php
class Context {


  /**
   * consider using {@see self::getExpandedRoleIdentifiers()} instead
   */
  public function getRoles(): array {
    // ...
  }

  public function getExpandedRoleIdentifiers(): RoleIdentifiers {
    // ...
  }
}
```

Also the `RoleIds` could have some static constructor like:

```php
final readonly class RoleIds {
  // ...
  public static function forAnonymousUser(): self
  {
    return new self('Neos.Flow:Everybody', 'Neos.Flow:Anonymous');
  }
}
```

to centralize this piece of magic

bwaidelich:
Resolves: #3414

### DIFF
--- a/Neos.Flow/Classes/Security/Policy/Role.php
+++ b/Neos.Flow/Classes/Security/Policy/Role.php
@@ -59,14 +59,17 @@ class Role
     protected $privileges = [];
 
     /**
-     * @param string $identifier The fully qualified identifier of this role (Vendor.Package:Role)
+     * @param RoleId|string $id The fully qualified identifier of this role (Vendor.Package:Role)
      * @param Role[] $parentRoles
      * @param string $label A label for this role
      * @param string $description A description on this role
      */
-    public function __construct(string $identifier, array $parentRoles = [], string $label = '', string $description = '')
+    public function __construct(RoleId|string $id, array $parentRoles = [], string $label = '', string $description = '')
     {
-        $this->id = RoleId::fromString($identifier);
+        if (is_string($id)) {
+            $id = RoleId::fromString($id);
+        }
+        $this->id = $id;
         $this->label = $label ?: $this->id->getName();
         $this->description = $description;
         $this->parentRoles = $parentRoles;
--- a/Neos.Flow/Classes/Security/Policy/Role.php
+++ b/Neos.Flow/Classes/Security/Policy/Role.php
@@ -269,7 +269,7 @@ class Role
      */
     public function __toString()
     {
-        return $this->identifier;
+        return $this->id->value;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Neos.Flow/Classes/Security/Policy/Role.php
```diff
@@ -84,45 +66,48 @@ class Role
...
-        }
-        $this->identifier = $identifier;
-        $this->packageKey = $matches[1];
-        $this->name = $matches[2];
-        $this->label = $label ?: $matches[2];
+        $this->id = RoleId::fromString($identifier);
+        $this->label = $label ?: $this->id->getName();
         $this->description = $description;
         $this->parentRoles = $parentRoles;
     }
```
kitsunet, line 77:
```suggestion

public static function fromRoleId(RoleId $id, array $parentRoles = [], string $label = '', string $description = ''): self
{
    return new ($id->value, array $parentRoles = [], string $label = '', string $description = '');
}

public static function fromIdentifierString(string $identifier, array $parentRoles = [], string $label = '', string $description = ''): self
{
    return new ($identifier, array $parentRoles = [], string $label = '', string $description = '');
}
```

And then deprecate the __construct as it is so we can at least  open a window for refactoring internally?

bwaidelich: @kitsunet mh, `Role::fromRoleId()` and `Role::fromIdentifierString()` sounds wrong to me.. But what do you think of
```php
public function __construct(RoleId|string $id, array $parentRoles = [], string $label = '', string $description = '') {
  if (is_string($id)) {
    $id = RoleId::fromString($id);
  }
  // ...
}
```

kitsunet: Sure, sounds good! My thought train was not finished really. I was thinking / hoping how we can refactor towards something like RoleIds $parentRoles or somthing, but the static constructors would not necessarily make that easier :D

bwaidelich: I agree – there is a lot of potential to improve things, but I wanted to avoid another rabbit hole (like #3410)
Adjusted with 32bbd2dbeeeda3bd5ea7a8daa6205d1298f01580

kitsunet: Yeah, totally fair

kitsunet (approved):

Issue #3417, opened 2024-11-20
BUG: Almost all static resources are published twice


Sebobo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Due to a regression introduced in https://github.com/neos/flow-development-collection/commit/29394e7a135c6736f91d579ca175ac8195b2cc21#diff-4360c37e41d3456986dee193b75552ea7dfa3cb03b0c4161a07fdcbe8e990f07R460
a list of all files in public package folders are collected twice.

The pattern `'*/Resources/Public/'` includes all files, the pattern '*/Resources/Public/*' only the ones in subfolders. So f.e. the first iteration collects 1000 files and the second 990, but those are all duplicates.

### Expected Behavior

Every file in a collection is only published once, even with additional path patterns.

### Steps To Reproduce

Add a var_dump to \Neos\Flow\ResourceManagement\Storage\PackageStorage::getObjectsByPathPattern that shows the number of iterations depending on the given $pattern.

### Environment

```markdown
- Flow: Since TYPO3 times
```

Sebobo:
This could be first fixed by removing the 2nd pattern from the default configuration in `Neos.Flow.resource.collections.static.pathPatterns`, but it would also be good if the Collection would filter duplicate paths or objects.

PR #3418, opened 2024-11-20
BUGFIX: Don’t load the same files twice in the static resource collection

Sebobo:
The second pattern included a 99% subset of the first pattern except the files in the public folders themselves. Therefore all files in subfolders were loaded twice in the `\Neos\Flow\ResourceManagement\Collection::getObjects` method and afterwards also published twice.

In a medium sized project the first pattern loaded 1165 files and the second pattern 1155 which were all duplicates.

An additional change is required to deduplicate paths loaded in the `Collection` class to prevent further configuration regressions for cases like this.

Resolves: #3417

**Review instructions**

* Delete the `Web/_Resources/Static` folder in a demo distribution
* Run `./flow resource:publish --collection static` 
* Reload the Neos backend and all the static resources like JS and CSS should still be there.

### DIFF
--- a/Neos.Flow/Configuration/Settings.Resource.yaml
+++ b/Neos.Flow/Configuration/Settings.Resource.yaml
@@ -46,7 +46,6 @@ Neos:
           target: 'localWebDirectoryStaticResourcesTarget'
           pathPatterns:
             - '*/Resources/Public/'
-            - '*/Resources/Public/*'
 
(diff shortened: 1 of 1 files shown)


kdambekalns (approved):
Looks reasonable… did not test, though. 😬

mhsdesign (approved):

PR #3419, opened 2024-11-28
FEATURE: Add fromClassAndMethodName() to LogEnvironment

kdambekalns:
This does some code cleanup and adds a new method `fromClassAndMethodName($className, $methodName)` to the `LogEnvironment`.

It be used like this:

    LogEnvironment::fromClassAndMethodName(__CLASS__, __METHOD__)

and will return useful data – as opposed to `LogEnvironment::fromMethodName(__METHOD__)`, which only works for static methods (`SomeClass::someMethod` or closures).

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Log/Utility/LogEnvironment.php
+++ b/Neos.Flow/Classes/Log/Utility/LogEnvironment.php
@@ -15,27 +15,20 @@ namespace Neos\Flow\Log\Utility;
 
 use Neos\Flow\Core\Bootstrap;
 use Neos\Flow\ObjectManagement\ObjectManagerInterface;
-use Neos\Flow\Package\PackageInterface;
 use Neos\Flow\Package\PackageKeyAwareInterface;
 use Neos\Flow\Package\PackageManager;
 use Neos\Flow\Annotations as Flow;
 
 abstract class LogEnvironment
 {
-    /**
-     * @var array
-     */
-    protected static $packageKeys = [];
+    protected static array $packageKeys = [];
 
-    /**
-     * @var bool
-     */
-    protected static $initialized = false;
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Log/Utility/LogEnvironment.php
```diff
@@ -61,9 +54,21 @@ public static function fromMethodName(string $methodName): array
     }
 
     /**
-     * @param string $className
-     * @return string
+     * Returns an array containing the log environment variables
+     * under the key FLOW_LOG_ENVIRONMENT to be set as part of the
+     * additional data in an log method call.
      */
+    public static function fromClassAndMethodName(string $className, string $methodName): array
```
robertlemke, line 61: What if we also allow passing an object instead of the class name? That way you could just pass `$this` in most cases.

```php
public static function fromClassAndMethodName(string|object $class, string $methodName): array
{
    $className = is_object($class) ? get_class($object) : $className;
...
```

robertlemke:
Why is this a task and not a new (tiny) feature?

Neos.Flow/Classes/Log/Utility/LogEnvironment.php
kdambekalns: Does that make it easier to use? You can even use `$this::class` if you like…

kdambekalns:
> Why is this a task and not a new (tiny) feature?

So I can sneak it into 8.3… 😎

dlubitz (approved):

markusguenther (approved):
Fine, thank you @kdambekalns 
It is a tiny tiny feature, but not sure if we should merge it in 8.3 or better 8.4 or so

PR #3405, opened 2024-10-25
TASK: Update PhpUnit

kdambekalns:
Allow newer PhpUnit versions and adjust tests as needed.

**Checklist**

- [x] Code follows the PSR-2 coding style
- [ ] Tests have been created, run and adjusted as needed
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [ ] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [ ] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked with `!!!` and have upgrade-instructions

PR #3421, opened 2024-12-09
TASK: Fix override of runBare() in functional test

kdambekalns:
As of PHPUnit 11 `runBare()` is a `final` method.

This was written ages ago by Sebastian Kurfürst, who recently said:

> IMHO we wanted to run each test twice to run it without cache and
> then with cache. But it seems this was broken anyways since a long
> time – so we can drop it

**Checklist**


kdambekalns:
This is an excerpt from #3405 that fixes just the `runBare()` override, as that leads to a fatal error when "doing things" with PhpUnit 11…

mhsdesign (commented):
in case we consider it worth fixing we can get around this restriction of `runBare` and achieve the same result in a similar way.

For that we would put all generic test functions in a trait and then have two TestClasses both using this trait but having different setup methods. That way we test the same interactions with differently configured views.

kdambekalns:
Ping…

mhsdesign (approved):
jup the StandaloneView thing override also looks kinda hacky :)

Sebobo (approved):

Issue #3404, opened 2024-10-21
BUG: `doctrine/migrations` has no more ConsoleLogger


Jan3k3y:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

./flow doctrine:migrate

> Class "Doctrine\Migrations\Tools\Console\ConsoleLogger" not found
> 
>   Type: Error
>   File: Data/Temporary/Development/SubContextBeach/SubContextInstance/Cache/Code/Fl
>         ow_Object_Classes/Neos_Flow_Persistence_Doctrine_Service.php
>   Line: 238
> beach@6bc8157fb20c:/application$ ./flow doctrine:migrate
> Migrating up to Neos\Flow\Persistence\Doctrine\Migrations\Version20240906102606

### Expected Behavior

./flow doctrine:migrate

> Migrating up to ...

### Steps To Reproduce

Updating `neos/flow` the `doctrine/migrations` package is updated to 3.9.x-dev.
After './flow doctrine:migrate' there is an error (see current behavior) that ConsoleLogger was not found.

In there the ConsoleLogger was deleted and the Logger from Symfony is used instead.

See here: https://github.com/doctrine/migrations/commit/2700819827c0748ba33f3a3b0e4f0746cbcf05b6#diff-2cc9c3db4c79b32974d69ac0f6ff0438356f4ce1c39b918b5e0cea02b21dfdd5 

### Environment

```markdown
- Flow: 9.0.x-dev
- PHP: 8.2.24
```

kitsunet:
Are you sure about the PHP version? Because we officially support only PHP 8.3.x  for Flow/Neos 9.0 ? That could already be the root of the issue otherwise.

Jan3k3y:
Pretty sure 🤔 

![image](https://github.com/user-attachments/assets/3b32977a-6b38-40a4-8671-829ff80c6bd5)

kitsunet:
I'd say then our constraints are bad, but we definitely expect 8.3 at least

kitsunet:
I am sorry, too much going on, we are discussing PHP 8.3 as requirement but that is not the case right now.

Which version of the doctrine packages got installed?

Jan3k3y:
Maybe we both have been doing enough Neos stuff the last days…

I just checked the repo and found the change on the 4.0.x branch
Currently i have a working codebase with `neos/flow:9.0.x-dev` requiring `doctrine/migrations:^3.3`

So I don't know what happened with my system
[...]

PR #3422, opened 2024-12-13
TASK: Use Symfony ConsoleLogger instead of Doctrine

dlubitz:
Resolves: https://github.com/neos/flow-development-collection/issues/3404

In next minor release (3.9) doctrine/migration will remove the ConsoleLogger
See: https://github.com/doctrine/migrations/pull/1449

To prevent any issues after release of this we replace the ConsoleLogger of doctrine with ConsoleLogger of Symfony.


mficzel (approved):

mhsdesign (approved):
thanks, it resolves https://github.com/neos/flow-development-collection/issues/3404

dlubitz:
Good catch @mhsdesign

Issue #3423, opened 2024-12-19
BUG: Property mapping fails for partial param annotation coverage


bwaidelich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

An exception
```
SomeController::someAction(): Argument #1 ($a) must be of type TypeA, TypeB given, called in ActionController.php on line 123
```
is thrown when trying to map arguments

### Expected Behavior

no exception should be thrown :)

### Steps To Reproduce

1. Create an MVC action that expects two params and only add a `@param` annotation *for the second* argument: 

```php
class SomeController extends ActionController
{
    /**
     * @param TypeB $b foo
     */
    public function someAction(TypeA $a, TypeB $b): string
    {
        return '...';
    }
}
```

2. Invoke the action with valid arguments

e.g. a HTTP request to http://127.0.0.1:8081/?a=valueA&b=valueB

### Environment

```markdown
- Flow: 8.3+
- PHP: all
```

### Anything else?

The issue lies in our [ReflectionService](https://github.com/neos/flow-development-collection/blob/ecb20f67d260497fc998b0fcee0429fea9fd5fdc/Neos.Flow/Classes/Reflection/ReflectionService.php#L1709-L1714) that extends the parameter information based on the **index** of the corresponding `@param` annotation instead of the **name**.

PR #3424, opened 2024-12-19
BUGFIX: Fix method param type expansion for partial annotation coverage

bwaidelich:
Adjusts the behavior of `ReflectionService::getMethodParameters()` such that `@param` annotations are mapped to the corresponding method argument based in their _name_ instead of the _index_.

Fixes: #3423

### DIFF
--- a/Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
+++ b/Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
@@ -227,7 +227,7 @@ class ReflectionServiceTest extends FunctionalTestCase
                 'array' => true,
                 'byReference' => false,
                 'allowsNull' => false,
-                'defaultValue' => NULL,
+                'defaultValue' => null,
                 'scalarDeclaration' => false,
             ],
             'param3' => [
@@ -238,7 +238,7 @@ class ReflectionServiceTest extends FunctionalTestCase
                 'array' => false,
                 'byReference' => false,
                 'allowsNull' => true,
-                'defaultValue' => NULL,
+                'defaultValue' => null,
                 'scalarDeclaration' => false,
             ],
         ];
--- a/Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
+++ b/Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
@@ -12,6 +12,9 @@ namespace Neos\Flow\Tests\Functional\Reflection;
  */
 
 use Neos\Flow\Reflection\ReflectionService;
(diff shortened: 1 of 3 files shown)


Sebobo (approved):
Thx! Works as expected in my test cases :)

dlubitz (approved):
Looks good.

dlubitz:
StyleCI is complaining about ... style.

Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
```diff
@@ -197,6 +200,51 @@ public function methodParameterTypeExpansionWorksWithNullable()
...
+            ],
+            'param2' => [
+                'position' => 1,
+                'optional' => false,
+                'type' => 'array<' . SubSubEntity::class . '>',
+                'class' => null,
+                'array' => true,
+                'byReference' => false,
+                'allowsNull' => false,
+                'defaultValue' => NULL,
```
bwaidelich:
```suggestion
                'defaultValue' => null,
```

Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
```diff
@@ -197,6 +200,51 @@ public function methodParameterTypeExpansionWorksWithNullable()
...
+            ],
+            'param3' => [
+                'position' => 2,
+                'optional' => true,
+                'type' => SubSubSubEntity::class,
+                'class' => SubSubSubEntity::class,
+                'array' => false,
+                'byReference' => false,
+                'allowsNull' => true,
+                'defaultValue' => NULL,
```
bwaidelich:
```suggestion
                'defaultValue' => null,
```

mficzel:
FYI: This changes behavior when used with union parameters in type hints that contain spaces ... 

before: 
- ` @param AspectRatio | string | null $aspectRatio` -> was interpreted as type `\Neos\Media\Domain\ValueObject\Configuration\AspectRatio` for the first argument since it simply split after the first space
now: 
- ` @param AspectRatio | string | null $aspectRatio` -> is interpreted as `mixed` because of the spaces
even without spaces the classnames are not expanded: 
- ` @param AspectRatio|string|null $aspectRatio` -> is interpreted as `AspectRatio|string|null`

@see: https://github.com/neos/neos-development-collection/blob/796fe8f3eb6cdfbed7a75833725587bcaa59ce66/Neos.Media/Classes/Domain/Model/Adjustment/CropImageAdjustment.php#L162C4-L168C3

mficzel:
The related neos bug: https://github.com/neos/neos-development-collection/issues/5448

Issue #3425, opened 2024-12-21
FEATURE: Setting for encryptionKey


christophlehmann:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

The encryption key is normally a persistent filesystem cache entry. Currently it can't get overwritten with a setting. In environments that should not have local persistent data it would be handy to define it manually.

PR #3426, opened 2024-12-21
FEATURE: Add setting for encryption key

christophlehmann:
With this the encryption key can be defined in a setting. When defined it is not received from cache anymore.

```yaml
Neos:
  Flow:
    security:
      cryptography:
        encryptionKey: 'something-random-usually-40-chars-long'
```

Ideally set this from an environment variable, so it doesn't have to be in your codebase…

Resolves: #3425

**Upgrade instructions**

None – but feel free to use the new feature. For existing projects, set the encryption key to the value found in the cache, `Data/Persistent/Cache/Data/Flow_Security_Cryptography_HashService/encryptionKey` usually.

**Review instructions**

Set the new setting to a non-empty string and see that `\Neos\Flow\Security\Cryptography\HashService::generateHmac()` returns a Hmac generated with your new encryption key.

**Checklist**

### DIFF
--- a/Neos.Flow/Configuration/Settings.Security.yaml
+++ b/Neos.Flow/Configuration/Settings.Security.yaml
@@ -98,6 +98,10 @@ Neos:
 
       cryptography:
 
+        # A private, unique key used for encryption tasks. Normally 40 characters long and received from a persistent
+        # filesystem cache. If set to a non-empty string, the cache is not involved anymore.
+        encryptionKey: ''
+
         hashingStrategies:
 
           # The default strategy will be used to hash or validate passwords if no specific strategy is given
--- a/Neos.Flow/Configuration/Settings.Security.yaml
+++ b/Neos.Flow/Configuration/Settings.Security.yaml
@@ -100,7 +100,7 @@ Neos:
 
         # A private, unique key used for encryption tasks. Normally 40 characters long and received from a persistent
         # filesystem cache. If set to a non-empty string, the cache is not involved anymore.
-        encryptionKey: ''
+        encryptionKey: null
 
         hashingStrategies:
 
--- a/Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.security.schema.yaml
+++ b/Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.security.schema.yaml
@@ -74,6 +74,8 @@ properties:
     additionalProperties: false
     required: true
     properties:
+      'encryptionKey': { type: string, required: true }
+
       'hashingStrategies':
         type: dictionary
         additionalProperties: { type: string, format: class-name }
--- a/Neos.Flow/Classes/Security/Cryptography/HashService.php
+++ b/Neos.Flow/Classes/Security/Cryptography/HashService.php
@@ -63,6 +63,9 @@ class HashService
     public function injectSettings(array $settings)
     {
         $this->strategySettings = $settings['security']['cryptography']['hashingStrategies'];
+        if (!empty($settings['security']['cryptography']['encryptionKey'])) {
+            $this->encryptionKey = $settings['security']['cryptography']['encryptionKey'];
+        }
     }
 
     /**


kdambekalns:
Usually we put the cache for this into the database, in case the filesystem does not suffice (e.g. if it's non-persistent or the setup scales over multiple server)…

kdambekalns (commented):
I am not opposed to this. I wonder, though, if there should be some check for the key, to make sure it's not just `123` when it is set…

christophlehmann:
Plaintext secrets in the database might be a way, but have a bad touch to me.

> wonder, though, if there should be some check for the key

Would keep it simple stupid. Its comment states the default length, which is sufficient in my eyes. Ok?

kdambekalns:
> Plaintext secrets in the database might be a way, but have a bad touch to me.

True. Point taken.

> Would keep it simple stupid. Its comment states the default length, which is sufficient in my eyes. Ok?

Yeah, fine. I guess anyone caring enough to set it, will do it right.

kdambekalns:
I tweaked the PR description a bit. It would be awesome if you could rebase this on Flow 8.3 though, as it's clearly not in scope for the security-only versions before that.

Oh, and – thanks for the PR! I don't mean to sound hostile. 😇

christophlehmann:
>I tweaked the PR description a bit. It would be awesome if you could rebase this on Flow 8.3 though, as it's clearly not in scope for the security-only versions before that.

Okay, I interpreted 7.3 as lowest maintained branch. Is rebased!

>I don't mean to sound hostile.

No, you don't. Questions and discussions are absolutely fine 👍

kdambekalns (approved):
Ok, fine with me.

I requested review from a few people and have one last question (to those): In theory this must go into 8.4, as it is marked as a feature. But… is it? WDYT?

Neos.Flow/Configuration/Settings.Security.yaml
```diff
@@ -98,6 +98,10 @@ Neos:
 
       cryptography:
 
+        # A private, unique key used for encryption tasks. Normally 40 characters long and received from a persistent
+        # filesystem cache. If set to a non-empty string, the cache is not involved anymore.
+        encryptionKey: ''
```
dlubitz: Wouldn't it make sence to set this to `null` instead? Would clearly say, that this property is not set.
```suggestion
        encryptionKey: null
```

dlubitz (commented):
Thank you for your PR. This makes sence. I left one comment to discuss.

I'd rather vote for 8.4 but also fine with sneaking it into 8.3.

Neos.Flow/Configuration/Settings.Security.yaml
kdambekalns: I digress – the schema says it's a required string, so either make it an empty string by default or adjust the schema. 😎

dlubitz: Yeah, sure. But the schema is part of this PR, so we are open for both solutions, right? So question is, what do we expect the value for an not-set encryption key?

Neos.Flow/Resources/Private/Schema/Settings/Neos.Flow.security.schema.yaml
```diff
@@ -74,6 +74,8 @@ properties:
     additionalProperties: false
     required: true
     properties:
+      'encryptionKey': { type: string, required: true }
```
dlubitz, line 77:
```suggestion
      'encryptionKey': { type: [string, 'null'], required: true }
```

I guess the schema sould look like this. (untested)

Neos.Flow/Configuration/Settings.Security.yaml
kdambekalns: Something empty… 🙃 Either way is fine.

kitsunet: I recently use empty string = special value more if I need it, just to avoid union types and really we do not want an empty encryptionKey anyways, so that makes sense to me.

kitsunet (approved):

dlubitz:
@kitsunet This has broken the pipeline in 8.3:
https://github.com/neos/flow-development-collection/actions/runs/27062906451

> Settings.Neos.Flow.security.cryptography.encryptionKey -> expected: type=string found: type=NULL

I guess it's because of the open discussion above.

dlubitz:
Fix: https://github.com/neos/flow-development-collection/pull/3574

PR #3427, opened 2025-01-08
TASK: Correctly mark nullable method parameters as nullable

Benjamin-K:
**Review instructions**

This PR will correctly mark implicit nullable types as nullable, which will remove deprecation warnings in PHP 8.4.

**Checklist**

- [x] Code follows the PSR-2 coding style
- [ ] Tests have been created, run and adjusted as needed
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [ ] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [ ] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked with `!!!` and have upgrade-instructions

PR #3429, opened 2025-01-14
TASK: Correctly mark nullable method parameters as nullable

dlubitz:
Replaces #3427 

This PR will correctly mark implicit nullable types as nullable, which will remove deprecation warnings in PHP 8.4.

I've used rector to migrate all occurrences.
https://getrector.com/rule-detail/explicit-nullable-param-type-rector

See also for Neos: https://github.com/neos/neos-development-collection/pull/5433

Note: We should run the rector migration again on 9.0 after upmerging this PR

Thanks to @Benjamin-K for bringing this up and providing the first PR (#3427).


markusguenther (approved):
It looks great to me, and it seems like the flow is still running smoothly. :P

Issue #3430, opened 2025-01-15
BUG: Validation label for id 1327865764 is missing


markusguenther:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When you use Neos.Form and the Neos.Form.FusionRenderer with the File Upload, there’s a chance that the file extension isn’t valid. In that case, the validation error should be displayed, but the labels are in Flow, and the label for the case is missing.

### Expected Behavior

Validation Error will be rendered!

### Steps To Reproduce

1. Utilize Neos.Form to construct a form and incorporate a File Upload component.
2. Specify the list of acceptable file types.
3. Upload a file of an incompatible type.

### Environment

```markdown
- Flow:8.3
- PHP:8.x
```

markusguenther:
This is somehow related to these PR https://github.com/neos/form-fusionrenderer/pull/35
As the validation message is not visible without the labels.

PR #3431, opened 2025-01-15
BUGFIX: Add missing validation labels

markusguenther:
When you use Neos.Form and the Neos.Form.FusionRenderer with the File Upload, there’s a chance that the file extension isn’t valid. In that case, the validation error should be displayed, but the labels are in Flow, and the label for the case is missing.

This change adds the missing label for the validation error message.

Fixes: #3430

### DIFF
--- a/Neos.Flow/Resources/Private/Translations/de/ValidationErrors.xlf
+++ b/Neos.Flow/Resources/Private/Translations/de/ValidationErrors.xlf
@@ -248,6 +248,11 @@
 				<source>The image orientation "{0}" is not allowed</source>
         <target state="final">Das Bildformat "{0}" ist nicht erlaubt</target>
       </trans-unit>
+      <!-- Package "Neos.Form" -->
+      <trans-unit id="1327865764" xml:space="preserve">
+				<source>The file extension "{0}" is not allowed.</source>
+            <target state="translated">Die Dateiendung "{0}" ist nicht erlaubt.</target>
+      </trans-unit>
     </body>
   </file>
 </xliff>
--- a/Neos.Flow/Resources/Private/Translations/en/ValidationErrors.xlf
+++ b/Neos.Flow/Resources/Private/Translations/en/ValidationErrors.xlf
@@ -219,6 +219,10 @@
 				<source>The image orientation "{0}" is not allowed</source>
 			</trans-unit>
 
+			<!-- Package "Neos.Form" -->
+			<trans-unit id="1327865764" xml:space="preserve">
+				<source>The file extension "{0}" is not allowed.</source>
+			</trans-unit>
 		</body>
 	</file>
 </xliff>


markusguenther:
I know that we have weblate, but I could not login and did not get my mail reset 🙈

dlubitz:
But this need to go into `Neos.Form` instead of `Neos.Flow` right?

markusguenther:
Yeah, for some reason, the labels are set to Neos.Flow as a package.
```Neos.Form/Configuration/Settings.yaml```

We can also adjust `validationErrorTranslationPackage` and copy the needed labels to the form package, but I bet that it was done that way to have one single point for the labels and not mix the location of the labels based on the context.

```
Neos:
  Form:
    presets:
      default:
        formElementTypes:
          'Neos.Form:Base':
            renderingOptions:
              templatePathPattern: 'resource://{@package}/Private/Form/{@type}.html'
              partialPathPattern: 'resource://{@package}/Private/Form/Partials/{@type}.html'
              layoutPathPattern: 'resource://{@package}/Private/Form/Layouts/{@type}.html'
              skipUnknownElements: false
              translationPackage: Neos.Form
              validationErrorTranslationPackage: Neos.Flow
```

And this is used for rendering the validation errors:
```
content = Neos.Fusion:Loop {
        items = ${elementValidationErrors}
        itemName = 'error'
        itemRenderer = ${Translation.translate(error.code, null, error.arguments, 'ValidationErrors', element.renderingOptions.validationErrorTranslationPackage)}
    }
```

**Review by mhsdesign — CHANGES_REQUESTED** [review]
Neos.Form is not even part of this dev collection ... i f weblate doesnt get it i would just do it by hand in the form package but here its the wrong place?!

**Comment by markusguenther** [comment]
> Neos.Form is not even part of this dev collection ... i f weblate doesnt get it i would just do it by hand in the form package but here its the wrong place?!

Maybe it feels wrong but as it is build currently it is the right place!

**Comment by dlubitz** [comment]
IDK if this works with weblate, but usually you would create a translation file in the external package, that points to the source package.

**Comment by markusguenther** [comment]
@mhsdesign Neos.Media is also not part of this collection but Neos.Flow contains the validation labels for that!

**Review by mhsdesign — APPROVED** [review]
Jup i see we have done this with Neos.Media as well and also other things since the introduction of being able to translate the errors at all with https://github.com/neos/flow-development-collection/commit/a6f776bd0aff745fb2fb1dc64d003686969124cc

Now your change made this 13 YEAR old smell visible but yeah then i guess for pragmatic reasons we shall follow the course.

Problem is that translations have to reside in one package and there is no real translation service for validation errors but rather a convention to use `Neos.Flow` and the `ValidationErrors` translations.

**Comment by mhsdesign** [comment]
As alternative which would be more flexible id use


```
 = Match {
          @subject = error.code
          1327865764 = // custom translation source
          @default = ${Translation.translate(error.code, null, error.arguments, 'ValidationErrors', element.renderingOptions.validationErrorTranslationPackage)}
       }
```

PR #3434, opened 2025-01-15
TASK: Fix reflection test

dlubitz:
With 9.0 we get more data from `ReflectionService::getMethodParameters` so the test needs to get adapted.

See: https://github.com/neos/flow-development-collection/pull/3086


mhsdesign (approved):
Makes totally sense thanks!

PR #3435, opened 2025-01-16
Avoid race-condition when reading localeCollection

christianharrington:
The previous implementation was an example of a check-then-act race-condition: There is no guarantee that the cached value still exists when it is subsequently read out.

This would cause errors later when `$this->localeCollection` could suddenly have the value `false`.

See #3188 for more context.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/I18n/Service.php
+++ b/Neos.Flow/Classes/I18n/Service.php
@@ -83,8 +83,10 @@ class Service
         $this->configuration = new Configuration($this->settings['defaultLocale']);
         $this->configuration->setFallbackRule($this->settings['fallbackRule']);
 
-        if ($this->cache->has('availableLocales')) {
-            $this->localeCollection = $this->cache->get('availableLocales');
+        $cachedCollection = $this->cache->get('availableLocales');
+
+        if ($cachedCollection !== false) {
+            $this->localeCollection = $cachedCollection;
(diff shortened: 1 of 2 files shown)


kdambekalns (approved):
Seems edge-casey to me, but totally makes sense.

kdambekalns:
Just the tests need to be adjusted…

christianharrington:
> Seems edge-casey to me, but totally makes sense.

This was happening many times a day in production for us before we implemented a mitigation 🤷🏻 

> Just the tests need to be adjusted…

Yeah, sorry, I had difficulty figuring out how to run them locally. I couldn't find any documentation on it, so I ended up reverse-engineering it from the Github Actions workflow.

I found the error raised in the test quite confusing as well:
```
Expectation failed for method name is "addLocale" when invoked 4 time(s).
```
The issue [was](https://github.com/neos/flow-development-collection/pull/3435/commits/d9fdc38c43b8df9794efb6725406f7c150601caa) in fact the method name being used for the mocked cache 🤔

markusguenther (approved):
Thanks for bringing that up again! Sorry I missed it yesterday with the tests and was too quick with the merge since it worked locally. 😅

Thanks for your contribution, and it’s great to hear that you still use Flow at famly.

kdambekalns:
> This was happening many times a day in production for us before we implemented a mitigation

Doh. Ok, to me it seems that cache should exist rather reliably… 🙈

PR #3437, opened 2025-01-22
BUGFIX: Drop use of `E_STRICT` to fix PHP 8.4 deprecation

kdambekalns:
The use of `E_STRICT` is deprecated as of PHP 8.4, so this fixes deprecation warnings. Furthermore, the constant is no longer useful…

In PHP 5.4, the functionality of `E_STRICT` was incorporated into `E_ALL`, meaning strict standards notices are included in the `E_ALL` error level. As a result, there is no need to use `E_STRICT` separately starting with PHP 5.4. This change is documented in the PHP manual under the migration guide for PHP 7.0, which states:

> All of the E_STRICT notices have been reclassified to other levels.
> The E_STRICT constant is retained, so calls like
> `error_reporting(E_ALL|E_STRICT)` will not cause an error.

(see https://www.php.net/manual/en/migration70.incompatible)

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Error/ErrorHandler.php
+++ b/Neos.Flow/Classes/Error/ErrorHandler.php
@@ -72,7 +72,6 @@ class ErrorHandler
             E_USER_ERROR         => 'User Error',
             E_USER_WARNING       => 'User Warning',
             E_USER_NOTICE        => 'User Notice',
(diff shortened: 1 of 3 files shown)


dlubitz (approved):
👍 Looks good by reading.

PR #3439, opened 2025-01-23
BUGFIX: Union type handling

mficzel:
Demonstrate broken union types from annotations via test.

1. commit - remove rule that tests union types only with php 8 ... all green
2. commit - add a method with union type parameter hints and test thererfore ... still green
3. commit - add a class similar to 2. but with @param and @return annotations instead of type-hints  ... tests verify that the same types are detected as with type hints  ... and fail

Please note tha added case `parameterB1`  is equel to **parameterB** but demonstrates the different handlung of spaces in annotations that lead to detecting `mixed` type.

see: #3440

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
+++ b/Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
@@ -12,12 +12,12 @@ namespace Neos\Flow\Tests\Functional\Reflection;
  */
 
 use Neos\Flow\Reflection\ReflectionService;
+use Neos\Flow\Tests\Functional\Persistence;
+use Neos\Flow\Tests\Functional\Reflection;
 use Neos\Flow\Tests\Functional\Reflection\Fixtures\Model\SubEntity;
 use Neos\Flow\Tests\Functional\Reflection\Fixtures\Model\SubSubEntity;
 use Neos\Flow\Tests\Functional\Reflection\Fixtures\Model\SubSubSubEntity;
 use Neos\Flow\Tests\FunctionalTestCase;
-use Neos\Flow\Tests\Functional\Reflection;
-use Neos\Flow\Tests\Functional\Persistence;
 
 /**
  * Functional tests for the Reflection Service features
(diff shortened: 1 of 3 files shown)


mficzel:
I think ... 
1. the @param annotation should be read with a regex and not explode( ' ',  ...)
2. the expanType in the reflectionService has to learn about UnionTypes and probably intersectioTypes

Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
```diff
@@ -356,12 +356,57 @@ public function annotatedArrayTypeHintsWorkCorrectly()
...
+        $returnTypes = [
+            'returnTypeA' => $this->reflectionService->getMethodDeclaredReturnType(Fixtures\DummyClassWithUnionTypeHints::class, 'methodWithUnionReturnTypeA'),
+            'returnTypeB' => $this->reflectionService->getMethodDeclaredReturnType(Fixtures\DummyClassWithUnionTypeHints::class, 'methodWithUnionReturnTypesB'),
+            'returnTypeC' => $this->reflectionService->getMethodDeclaredReturnType(Fixtures\DummyClassWithUnionTypeHints::class, 'methodWithUnionReturnTypesC'),
+        ];
+
+        self::assertEquals(
+            [
+                'returnTypeA' => 'string|false',
+                'returnTypeB' => '\Neos\Flow\Tests\Functional\Reflection\Fixtures\DummyClassWithUnionTypeHints|false',
```
kdambekalns, line 368: In PHP FQCN in strings never have a leading backslash – should we handle that the same way? At least consistency would be great, seeing the next test expects those without a leading backslash…

Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
```diff
@@ -356,15 +356,121 @@ public function annotatedArrayTypeHintsWorkCorrectly()
...
+        $returnTypes = [
+            'returnTypeA' => $this->reflectionService->getMethodDeclaredReturnType(Fixtures\DummyClassWithUnionTypeAnnotations::class, 'methodWithUnionReturnTypeA'),
+            'returnTypeB' => $this->reflectionService->getMethodDeclaredReturnType(Fixtures\DummyClassWithUnionTypeAnnotations::class, 'methodWithUnionReturnTypesB'),
+            'returnTypeC' => $this->reflectionService->getMethodDeclaredReturnType(Fixtures\DummyClassWithUnionTypeAnnotations::class, 'methodWithUnionReturnTypesC'),
+        ];
+
+        self::assertEquals(
+            [
+                'returnTypeA' => 'string|false',
+                'returnTypeB' => '\Neos\Flow\Tests\Functional\Reflection\Fixtures\DummyClassWithUnionTypeAnnotations|false',
```
kdambekalns, line 427: See earlier comment about leading backslash…

Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
mficzel: The tests just took what the type hints created for granted. May aswell be buggy.

Issue #3402, opened 2024-10-12
BUG: Reflection data unavailable in Production runtime


kitsunet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

This is a complex issue with multiple layers and no easy fix.

The ReflectionService currently has a special case in initialize for Production context specifically that will load a limited subset of cached reflection data compared to the full compiletime set. This distinction is generally very sensible as the amount of data is huge and shouldn't really be used during runtime anyways. BUT there is no limitation to calling the respective methods using this data, thus you unexpectedly get different results in Development and Production on the same codebase when asking the ReflectionService for example for `getClassesContainingMethodsAnnotatedWith`. 
This is dangerous as it can lead to errors that appear only after a production deployment.

### Expected Behavior

Loading all reflection data on production is prohibitive due to the RAM usage it would incur, so alternative solutions have to be found. A workaround exists in the form of CompileStatic annotation/attribute, which will execute the annotated static method at compiletime and store the results of that call as hardcoded return value in the generated proxy class. This works and is great for performance, yet has issues in itself. First of all CompileStatic is inert in Development again, so you get different proxy classes between Production and Development environments. While no known issues exist around this and the Flow core uses CompileStatic since years, it might lead to hard to track down effects as well. 
As these two (CompileStatic and the different reflection data in Production) work hand in hand any steps changing this need to be taken carefully. 

It could be benefial for example to **change `\Neos\Flow\Reflection\ReflectionService::initialize` to not be special for Production but generally divide between compile time and runtime data for all appliction contexts**. This however would immediately lead to compile static failing in Development and Testing as the necessary reflection data (as mentioned above CompileStatic is inert outside of Production) is not available anymore so the runtime call executing the static method will not receive the necessary data anymore.

The tandem step would be to **enable CompileStatic for Development and Testing**. This should generally be fairly safe, the biggest consideration is managing regeneration of the proxies (a quick and safe approach could be to just regenerate any proxy with compile static in it on any change to reflection data). To help debugging compile static behavior and what happens inside the static method we could **provide a command that runs in compiletime and executes a given (compile) static method, so you can xdebug and observe it in isolation**.

This combination of 3 measures together would be one route to take to clean this up a bit and prevent surprises in Production.

Next up to add additional safety would be a split between compiletime reflection and runtime reflection analog to the object manager, such that the runtime reflectionservice doesn't offer any methods using data not available at runtime.
[...]

PR #3443, opened 2025-02-07
!!!TASK: Reduce complexity of ReflectionService

kitsunet:
This change tackles some problems within the reflection service that stem from historically increasing complexity due to various caching mechanisms depending on application context and compile time status.

The aim was to cut down on this complexity, while ensuring that all existing use-cases continue working as intended.

This ultimately also fixes issue #3402 by providing the same reflection data across all possible contexts.

A few features and caches got deprecated with this change and could be breaking in the rare case you used the freeze package api in your code:

The entire concept of freezing a package is deprecated

What remains are the commands in the package controller, which are now all no-ops and deprecated to be removed with 9.0. This is to ensure deployment pipelines possibly calling freeze commands do not break with the 8.4 update.

Additionally the single method `PackageManager::isPackageFrozen` remains, while the rest was removed. None of the methods was ever api and it seems unlikely that someone used them in user-land code. `isPackageFrozen` however is at the very least used in Framework and Neos code and therefore remains until 9.0, but will now return false for every package.

Caches deprecated and unused

With the simplification two caches are no longer needed, both are still declared so that possibly existing cache configuration in user projects doesn't error, but both

`Flow_Reflection_Status`

and

`Flow_Reflection_CompiletimeData`

will no longer be used and any content can be removed.

The only reflection cache is now `Flow_Reflection_RuntimeData`, which makes the name somewhat deceptive as it is also used in compile time. To avoid backwards compatibility issues however it makes sense to keep the name for the foreseeable future.

Quick performance comparisons suggest that especially the initial compile from empty cache benefits from this change. Reflection updates in Development context afterwards seem to be on par with the existing code base.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Command/PackageCommandController.php
+++ b/Neos.Flow/Classes/Command/PackageCommandController.php
@@ -14,8 +14,6 @@ namespace Neos\Flow\Command;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Cli\CommandController;
 use Neos\Flow\Composer\ComposerUtility;
-use Neos\Flow\Core\Booting\Scripts;
-use Neos\Flow\Core\Bootstrap;
 use Neos\Flow\Package\PackageInterface;
 use Neos\Flow\Package\PackageKeyAwareInterface;
 use Neos\Flow\Package\PackageManager;
@@ -33,40 +31,12 @@ class PackageCommandController extends CommandController
     protected $packageManager;
 
     /**
-     * @var array
-     */
-    protected $settings;
-
-    /**
-     * @var Bootstrap
-     */
-    protected $bootstrap;
-
-    /**
-     * @param array $settings The Flow settings
-     * @return void
-     */
-    public function injectSettings(array $settings)
-    {
-        $this->settings = $settings;
-    }
-
-    /**
      * @param PackageManager $packageManager
      * @return void
      */
     public function injectPackageManager(PackageManager $packageManager)
     {
         $this->packageManager = $packageManager;
-    }
-
-    /**
-     * @param Bootstrap $bootstrap
-     * @return void
-     */
-    public function injectBootstrap(Bootstrap $bootstrap)
-    {
-        $this->bootstrap = $bootstrap;
     }
 
     /**
@@ -112,9 +82,7 @@ class PackageCommandController extends CommandController
     public function listCommand(bool $loadingOrder = false)
     {
         $availablePackages = [];
-        $frozenPackages = [];
         $longestPackageKey = 0;
-        $freezeSupported = $this->bootstrap->getContext()->isDevelopment();
 
         foreach ($this->packageManager->getAvailablePackages() as $packageKey => $package) {
             if (strlen($packageKey) > $longestPackageKey) {
@@ -122,10 +90,6 @@ class PackageCommandController extends CommandController
             }
 
             $availablePackages[$packageKey] = $package;
-
-            if ($this->packageManager->isPackageFrozen($packageKey)) {
-                $frozenPackages[$packageKey] = $package;
-            }
         }
 
         if ($loadingOrder === false) {
@@ -135,13 +99,7 @@ class PackageCommandController extends CommandController
         $this->outputLine('PACKAGES:');
         /** @var PackageInterface|PackageKeyAwareInterface $package */
         foreach ($availablePackages as $package) {
-            $frozenState = ($freezeSupported && isset($frozenPackages[$package->getPackageKey()]) ? '* ' : '  ');
-            $this->outputLine(' ' . str_pad($package->getPackageKey(), $longestPackageKey + 3) . $frozenState . str_pad($package->getInstalledVersion(), 15));
-        }
-
-        if (count($frozenPackages) > 0 && $freezeSupported) {
-            $this->outputLine();
-            $this->outputLine(' * frozen package');
+            $this->outputLine(' ' . str_pad($package->getPackageKey(), $longestPackageKey + 3) . str_pad($package->getInstalledVersion(), 15));
         }
     }
 
@@ -165,47 +123,11 @@ class PackageCommandController extends CommandController
      * @return void
      * @see neos.flow:package:unfreeze
      * @see neos.flow:package:refreeze
+     * @deprecated since 8.4
      */
     public function freezeCommand(string $packageKey = 'all')
     {
-        if (!$this->bootstrap->getContext()->isDevelopment()) {
-            $this->outputLine('Package freezing is only supported in Development context.');
-            $this->quit(3);
-        }
-
-        $packagesToFreeze = [];
-
-        if ($packageKey === 'all') {
-            foreach (array_keys($this->packageManager->getAvailablePackages()) as $packageKey) {
-                if (!$this->packageManager->isPackageFrozen($packageKey)) {
-                    $packagesToFreeze[] = $packageKey;
-                }
-            }
-            if ($packagesToFreeze === []) {
-                $this->outputLine('Nothing to do, all packages were already frozen.');
-                $this->quit(0);
-            }
-        } elseif ($packageKey === 'blackberry') {
-            $this->outputLine('http://bit.ly/freeze-blackberry');
-            $this->quit(42);
-        } else {
-            if (!$this->packageManager->isPackageAvailable($packageKey)) {
-                $this->outputLine('Package "%s" is not available.', [$packageKey]);
-                $this->quit(2);
-            }
-
-            if ($this->packageManager->isPackageFrozen($packageKey)) {
-                $this->outputLine('Package "%s" was already frozen.', [$packageKey]);
-                $this->quit(0);
-            }
-
-            $packagesToFreeze = [$packageKey];
-        }
-
-        foreach ($packagesToFreeze as $packageKey) {
-            $this->packageManager->freezePackage($packageKey);
-            $this->outputLine('Froze package "%s".', [$packageKey]);
-        }
+        $this->outputLine('Package freezing is no longer supported, this command is deprecated and will be removed with 9.0.');
     }
 
     /**
@@ -222,47 +144,11 @@ class PackageCommandController extends CommandController
      * @return void
      * @see neos.flow:package:freeze
      * @see neos.flow:cache:flush
+     * @deprecated since 8.4
      */
     public function unfreezeCommand(string $packageKey = 'all')
     {
-        if (!$this->bootstrap->getContext()->isDevelopment()) {
-            $this->outputLine('Package freezing is only supported in Development context.');
-            $this->quit(3);
-        }
-
-        $packagesToUnfreeze = [];
-
-        if ($packageKey === 'all') {
-            foreach (array_keys($this->packageManager->getAvailablePackages()) as $packageKey) {
-                if ($this->packageManager->isPackageFrozen($packageKey)) {
-                    $packagesToUnfreeze[] = $packageKey;
-                }
-            }
-            if ($packagesToUnfreeze === []) {
-                $this->outputLine('Nothing to do, no packages were frozen.');
-                $this->quit(0);
-            }
-        } else {
-            if ($packageKey === null) {
-                $this->outputLine('You must specify a package to unfreeze.');
-                $this->quit(1);
-            }
-
-            if (!$this->packageManager->isPackageAvailable($packageKey)) {
-                $this->outputLine('Package "%s" is not available.', [$packageKey]);
-                $this->quit(2);
-            }
-            if (!$this->packageManager->isPackageFrozen($packageKey)) {
-                $this->outputLine('Package "%s" was not frozen.', [$packageKey]);
-                $this->quit(0);
-            }
-            $packagesToUnfreeze = [$packageKey];
-        }
-
-        foreach ($packagesToUnfreeze as $packageKey) {
-            $this->packageManager->unfreezePackage($packageKey);
-            $this->outputLine('Unfroze package "%s".', [$packageKey]);
-        }
+        $this->outputLine('Package freezing is no longer supported, this command is deprecated and will be removed with 9.0.');
     }
 
     /**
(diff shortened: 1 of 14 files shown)


kitsunet:
Future steps here IMHO could/should be to transfer the huge array structure of reflection data into DTOs and possibly split code that creates the reflection into a Compiletime ReflectionService as we shouldn't need to reflect anything during runtime, therefore making the code easier to reason with. But I resisted from adding those changes here to avoid making it a needlessly large change for 8.4

Neos.Flow/Classes/Command/PackageCommandController.php
```diff
@@ -165,47 +123,11 @@ public function listCommand(bool $loadingOrder = false)
...
-            }
-
-            $packagesToFreeze = [$packageKey];
-        }
-
-        foreach ($packagesToFreeze as $packageKey) {
-            $this->packageManager->freezePackage($packageKey);
-            $this->outputLine('Froze package "%s".', [$packageKey]);
-        }
+        $this->outputLine('Package freezing is no longer supported, this command is deprecated and will be removed with 9.0.');
```
mhsdesign, line 130: juhuu thanks for removing that ❤️

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -762,7 +690,7 @@ public function hasMethod(string $className, string $methodName): bool
      * Returns all tags and their values the specified method is tagged with
      *
      * @throws ReflectionException
-     * @api
+     * @deprecated since 8.4
      */
     public function getMethodTagsValues(string $className, string $methodName): array
```
mhsdesign, line 750: i suppose this is getting deprecated because its unused system wide?

mhsdesign (commented):
thank you sooo much for the change! Ill yet have to actually test it and also fully understand the new reflection logic but from reading it briefly i love it. I remember attempting to fix things already in the reflection but the whole initialisation processes is just soo hard to wrap your head around and every-time i got a grasp later i forgot again. Because of that could you please just shortly describe what the expected difference in cache behaviour/performance during Development/Testing and Production might be? Anyway all this freezing functionally will be finally a thing of the past and it definitely does not justify any of the complexity! So super glad to have it gone.

Neos.Flow/Classes/Reflection/ReflectionService.php
kitsunet: Yep, correct, there is a test for it, but otherwise it's completely unused.

kitsunet:
So what this thing do **before**:

`buildReflectionData` is the starting point for actually doing reflection and is triggered via the CompiletimeObjectManager (only). The whole system is a bit disconnected and might be another candidate for future refactoring. 
`buildReflectionData` gets all known classes from the CompiletimeObjectManager, then goes ahead to  call`forgetChangedClasses` which checks if any of the classes no longer exist in the status cache (now removed). Those cache entries get removed from the cach _directly_ in the `CacheManager` based on possibly modifiedfiles found by the `FileMonitor`. This obviously has to happen beforehand. It would be nicer if these things were handled closer together (eg. filemonitor provides the changes to the ReflectionService or so TBD). 
Afterwards `reflectEmergedClasses` will go through the classes and reflect those not reflected.

Then the second big part is the `saveToCache` method called via signal.
This would do a bunch of different things, `updateReflectionData` would write one big cache entry to the reflection compiletime cache (now removed) if there was changed reflection data, BUT only if two other conditions didn't match, one was ($applicationContext = PRODUCTION && runtimecache frozen), no need to write anything because caches were frozen, the other was ("loadFromClassSchemaRuntimeCache" true), which indirectly was the same condition as above decided in `initialize`. IMHO that second condition should never happen as otherwise we should've early returned at the first already. 

Additionally in production context we would write the runtime cache (the only one now) and freeze it afterwards, BUT this production cache did not contain all data the compiletime cache contained leading to bug #3402 
The runtime cache contains separate entries for all caches unlike the compiletime cache which is one big file, this is important later to understand what I did.
Additionally it stores an array of all known classes as an entry (also important later)

And finally in development we would write an additional big cache file outside of any cache backend, that woudl contain about the same as the compiletime cache. This was called `freezePackageReflection` even though it has nothing to do with packages nor with package freezing or cache freezing. It was simply a separate cache file, but this is gone now too.

`initialize` did then decide which of these many caches should be used.

What happens **WITH THIS CHANGE**:

We will discuss this in a slightly different order to make more sense.
`saveToCache` checks if there is updated reflection data and if so runs  `updateCacheEntries` (renamed from `saveProductionData`), which basically works as before BUT stores all necessary data now, this uses the reflection runtime and schemata caches. It will no longer freeze caches.

`reflectEmergedClasses` is still the entry point for (re)generating caches and the flow is as before, we get all known classes from the compiletimeobjectmanager, then we `forgetChangedClasses` which no longer uses the status cache but directly checks the runtime class for the class entries, this is also now the cache were entries get removed from via the CacheManager. But this cuts out the status cache, which was necessary because in compiletime we had that big single file cache entry so we couldn't remove entries from that, so the status cache was a workaround to do this. 

Finally `reflectEmergedClasses` pretty much works as before, with the added special that we prefill the reflectionData with "empty" keys for all known classes, coming from the addtional cache entry `__classNames` in our runtime cache, which is then reduced in `forgetChangedClsases` if necessray, to have a complete list of currently reflected classes to compare the new list to.

`initialize` now always loads the "compounded" cache entries from the runtime cache (so `__classNames`, `__annotatedClasses`, and `__classesByMethodAnnotations`') and class reflection is loaded as needed via `loadOrReflectClassIfNecessary` which is pretty much used everywhere now to make sure we prefer the cache entries.




The class schemata weren't mentioned before, they are basically a sideline and contain data for doctrine ORM retrieved via reflection, so it's kinda in the right place but kinda not. Might also warrant another refactoring, but another day.

kitsunet:
Note I might have found two little things to imrpove / fix, shouldn't stop you from reviewing, but should be done before merge.

kitsunet:
Here we go, split in separate commits to  make it easier to review. the last commit is a lot of foobar adaptions to  docblocks and code style, nothing functional.

Neos.Flow/Classes/Package/PackageManager.php
```diff
@@ -229,6 +228,7 @@ public function getAvailablePackages(): array
      * precompiled reflection data in order to improve performance.
      *
      * @return array<PackageInterface>
+     * @deprecated since 8.4
```
kdambekalns, line 231: Why is this code still "doing things" instead of being removed or simply returning an empty array?

Neos.Flow/Classes/Package/PackageManager.php
```diff
@@ -403,6 +403,7 @@ public function createPackage($packageKey, array $manifest = [], $packagesPath =
      * @param string $toAbsolutePath
      * @return void
      * @throws FilesException
+     * @deprecated since 8.4
```
kdambekalns, line 394: I think that breaks horribly if actually used these days, no? 🙈

Neos.Flow/Configuration/Caches.yaml
```diff
@@ -93,16 +93,23 @@ Flow_Persistence_Doctrine_Results:
...
-Flow_Reflection_RuntimeData: []
-Flow_Reflection_RuntimeClassSchemata: []
+
+# Flow_Reflection
+#
+#
+Flow_Reflection_RuntimeData:
+  backend: Neos\Cache\Backend\SimpleFileBackend
+Flow_Reflection_RuntimeClassSchemata:
+  backend: Neos\Cache\Backend\SimpleFileBackend
```
kdambekalns, line 112: Why do these get an explicit backend configured now?

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1159,9 +1229,10 @@ public function reflectClassProperty(string $className, PropertyReflection $prop
             $this->classReflectionData[$className][self::DATA_CLASS_PROPERTIES][$propertyName][self::DATA_PROPERTY_TYPE] = trim((string)$property->getType(), '?');
         }
 
-        $visibility = $property->isPublic() ? self::VISIBILITY_PUBLIC : ($property->isProtected() ? self::VISIBILITY_PROTECTED : self::VISIBILITY_PRIVATE);
+        $visibility = $this->extractVisibility($property);
         $this->classReflectionData[$className][self::DATA_CLASS_PROPERTIES][$propertyName][self::DATA_PROPERTY_VISIBILITY] = $visibility;
 
+
```
kdambekalns:
```suggestion
```

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1727,29 +1825,15 @@ protected function convertParameterReflectionToArray(ParameterReflection $parame
...
-            foreach ($classNames as $index => $className) {
-                if (str_starts_with($className, $namespace)) {
-                    unset($classNames[$index]);
-                }
-            }
-        }
-
         foreach ($classNames as $className) {
-            if (!$this->statusCache->has($this->produceCacheIdentifierFromClassName($className))) {
+            if (is_string($className) && !$this->reflectionDataRuntimeCache->has($this->produceCacheIdentifierFromClassName($className))) {
```
kdambekalns, line 1827: What – if not a string – can `$className` be here?

Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php
```diff
@@ -57,6 +57,7 @@ public function classSchemaCanBeBuiltForAggregateRootsWithPlainOldPhpBaseClasses
     /**
      * @test
      * @throws
+     * @deprecated
```
kdambekalns:
```suggestion
     * @deprecated since 8.4
```

kdambekalns (approved):
Had a close look, left some comments. Generally fine, so 👍 already.

Neos.Flow/Classes/Package/PackageManager.php
kitsunet: You are right, total oversight, this should just return [];

Neos.Flow/Classes/Package/PackageManager.php
kitsunet: I am pretty sure it does LOL but I didn't want to do too much in this change, deprecating seemed sensible thouhg.

Neos.Flow/Configuration/Caches.yaml
kitsunet: Because they don't use any features the (more complex) FileBackend offers, and can therefore just use the SimpleFileBackend (should have been the case since ever).

Neos.Flow/Classes/Package/PackageManager.php
kdambekalns: Absolutely. I was just thinking "good this doesn't constantlyexplode into everyones faces…" 🤣

Neos.Flow/Classes/Reflection/ReflectionService.php
kitsunet: I had an example were that was _for some reason I couldn't figure out_ a bool, so I jjust wanted to be sure here....

kdambekalns (approved):

Neos.Flow/Classes/Package/PackageManager.php
mhsdesign: haha jup idk how i never found out or ignored move package 😂

Neos.Flow/Classes/Reflection/ReflectionService.php
mhsdesign: i think that must have been in a previous iteration of the code as `array_keys` is used which can NEVER return a bool as that is not a valid array key.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -17,29 +17,28 @@
...
-use Neos\Utility\Files;
 use Neos\Utility\TypeHandling;
 use Psr\Log\LoggerInterface;
 use Psr\Log\LogLevel;
 use ReflectionException;
+use ReflectionIntersectionType;
+use ReflectionNamedType;
+use ReflectionType;
+use ReflectionUnionType;
+use RuntimeException;
```
mhsdesign: we aggreed that we dont want to import first level things and use `\` - will fix that

mhsdesign: Fixed via 82452e8bd1121949f963b8efd625a9f86866d8f4

mhsdesign (approved):
Okay now im wondering whats with the `FreezableBackendInterface` - do we want to deprecate that as well?

Also two times the word "frozen" still appears in the inline docs of the reflection service. Probably the whole doc block of how it works is outdated - but maybe we can put in there youre nice outline from above?

Tested it by upmerging into 9.0 (my 8.4 setup is broken^^) Anyways there are only little merge conflict and besides an error at start (which is expected) and cache flushing the Neos Demo runs fine.
> Warning: unserialize(): Extra data starting at offset 283 of 307 bytes in Framework/Neos.Cache/Classes/Frontend/VariableFrontend.php line 94

I also prepared a followup for 9.0 to remove the dead apis https://github.com/neos/flow-development-collection/pull/3446
and a neos change to remove the usages as well: https://github.com/neos/neos-development-collection/pull/5467

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1012,14 +1067,17 @@ protected function prepareClassReflectionForUsage(string $className): string
...
     protected function reflectEmergedClasses(): void
     {
-        $classNamesToReflect = [];
-        foreach ($this->availableClassNames as $classNamesInOnePackage) {
-            $classNamesToReflect = array_merge($classNamesToReflect, $classNamesInOnePackage);
+        $availableClassnames = [];
+        foreach ($this->availableClassNames as $classNamesInPackage) {
+            $availableClassnames[] = $classNamesInPackage;
         }
+        $classNamesToReflect = array_merge([], ...$availableClassnames);
```
mhsdesign, line 1080: okay that part got me first confused ^^ clever though :D

it first looked like a `Object.assign({}, copyFrom)` in javascript but its way better, we unpack all arrays and then merge them together to a flat thing

we can make that a little more simple even:

```suggestion
        // flatten nested array structure to a list of classes
        $availableClassnames = array_merge(...array_values($this->availableClassNames));
```

We dont need the `[]` as since 7.4.0 this function can be called without any parameter, and it will return empty array.

And we still need the `array_values` to prevent: Uncaught ArgumentCountError: array_merge() does not accept unknown named parameters

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1873,219 +1901,75 @@ protected function loadClassReflectionCompiletimeCache(): bool
...
      */
     public function saveToCache(): void
     {
-        if ($this->hasFrozenCacheInProduction()) {
+        if (empty($this->updatedReflectionData)) {
             return;
         }
+
         if (!$this->initialized) {
             $this->initialize();
```
mhsdesign, line 1939: initialising here when shutting down doesnt make sense and can never be the case i think - but was so beforehand as well ...

kitsunet:
FreezableBackendInterface yep it's also for shit like this, that said, can be deprecated but didn't really feel like the scope here.

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1092,10 +1151,14 @@ protected function reflectClass(string $className): void
...
 
         $class = new ClassReflection($className);
-        if (!isset($this->classReflectionData[$className])) {
+        if (!isset($this->classReflectionData[$className]) || !is_array($this->classReflectionData[$className])) {
             $this->classReflectionData[$className] = [];
         }
 
+        if (!isset($this->classReflectionData[$className][self::DATA_INTERFACE_IMPLEMENTATIONS]) && $class->isInterface()) {
+            $this->classReflectionData[$className][self::DATA_INTERFACE_IMPLEMENTATIONS] = [];
+        }
```
mhsdesign, line 1160: we dont need that explicitly here but it doesnt hurt either right? should be done automatically in `addImplementedInterface`

kitsunet:
> Warning: unserialize(): Extra data starting at offset 283 of 307 bytes in Framework/Neos.Cache/Classes/Frontend/VariableFrontend.php line 94

Right that's because of the change to simplefilebackend, which doesn't expect stuff after hte payload, but it's alos faster therefore.

mhsdesign (commented):
just some other things i found when looking at the code - id still also like the inline docs to be fixed as i wrote above ^^ -> will make a followup pr to tackle my code findings ;) 


Edit: https://github.com/neos/flow-development-collection/pull/3447

mhsdesign:
Ha funny i found the original refactoring that introduced this production vs development distinction :D

https://github.com/neos/flow-development-collection/commit/ba1f5fb2adc2a274ce230c2d8328735f7d4e0e8b


And i did the least of performance testing one could do in dev mode and ran `time ./flow flow:core:compile` from cold caches and once after a file change and i could not see any immediate performance drawback before or after

PR #3444, opened 2025-02-13
TASK: Add documentation how work with psr responses in controller

mhsdesign:
There was some confusion how to replace `$this->response` when modifying view results.

The suggested pattern now is more explicit instead of providing any form of helper like:

```php
abstract class ResponseInformationHelper
{
    public static function response(Response|StreamInterface $response): Response
    {
        return $response instanceof Response ? $response : new Response(body: $response);
    }
}
```

Instead this will be written explicitly:

```php
$response = $this->view->render();
if (!$response instanceof Response) {
    $response = new Response(body: $response);
}
return $response
    ->withHeader('My-Header', 'foo');
```

Also this change inlines `ActionController::renderView` as the method signature is now really odd and unusable or understandable from the outside. `$this->view` should instead be used instead like when also assigning variables.

Related changes that lead to the new patterns:

- https://github.com/neos/flow-development-collection/pull/3286
- https://github.com/neos/flow-development-collection/pull/3311

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Mvc/ActionResponse.php
+++ b/Neos.Flow/Classes/Mvc/ActionResponse.php
@@ -28,13 +28,44 @@ use GuzzleHttp\Psr7\Response;
  * via `$this->response` {@see AbstractController::$response} and pass it along to places.
  * But this behaviour is deprecated!
  *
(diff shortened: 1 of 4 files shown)


mhsdesign:
ping @kitsunet ;)

kitsunet (approved):

PR #3445, opened 2025-02-13
FEATURE: 3334 Resource targets without space tunnels

### LINKED ISSUE #3334: FEATURE: Dynamic resource uri host (baseUri) for CLI
Building a resource uri with the default configuration will crash on CLI:

> No base URI could be provided.
> This probably means a call was made outside of an HTTP request and a base
> URI was neither configured nor specified as $fallbackRequest.

The entry point [ResourceManager::getPublicPackageResourceUri](https://github.com/neos/flow-development-collection/blob/9766acc894384d53928947f8ee5e4b320280b34e/Neos.Flow/Classes/ResourceManagement/ResourceManager.php#L425)
calls [TargetInterface::getPublicStaticResourceUri](https://github.com/mhsdesign/flow-development-collection/blob/821ac42c81806fcc7f3fc863ed810edf7c2d3cd2/Neos.Flow/Classes/ResourceManagement/Target/TargetInterface.php#L61)
which will use the configured target like [localWebDirectoryStaticResourcesTarget](https://github.com/neos/flow-development-collection/blob/0591a81a80c24b537d37843b629b39c0abb461db/Neos.Flow/Configuration/Settings.Resource.yaml#L60-L65)
which in case of the FileSystemTarget and FileSystemSymlinkTarget
will call the [BaseUriProvider::getConfiguredBaseUriOrFallbackToCurrentRequest](https://github.com/neos/flow-development-collection/blob/26e95ac5102961aff8b2e64e55b7c2973c73cd95/Neos.Flow/Classes/Http/BaseUriProvider.php#L78)
which uses `$bootstrap->getActiveRequestHandler->getHttpRequest` which is not possible on CLI.

Possible workarounds include:
- statically set `localWebDirectoryStaticResourcesTarget.targetOptions.baseUri` to `http://localhost:8081/_Resources/Static/Packages/`
- statically set `Neos.Flow.http.baseUri` to `http://localhost:8081`
- hack the BaseUriProvider (see https://github.com/neos/flow-development-collection/issues/2157#issuecomment-1939443774) for dynamic multi site support
- hack by extending the targets to dynamically override `getResourcesBaseUri` https://github.com/Flowpack/Flowpack.DecoupledContentStore/blob/6adb8e04246f5fc9b8471b656db7ddf131474745/Classes/Transfer/Resource/Target/MultisiteFileSystemSymlinkTarget.php#L16-L38
- use `$this->bootstrap->setActiveRequestHandler($requestHandler);` and provide a custom `HttpRequestHandlerInterface` which provides a mocked `getHttpRequest()` (like during testing)

To _actually_ fix this issue we should pass the current base uri to the target:

```php
interface TargetInterface
{
    public function getPublicStaticResourceUri(string $relativePathAndFilename, UriInterface $baseUri): string;

    public function getPublicPersistentResourceUri(PersistentResource $resource, UriInterface $baseUri): string;
    
    // ...
}
```

The methods are currently only used by the resource manager Neos & Flow.
And there we could allow a backwards compatible nullable parameter which will default to the current request:

```php

class ResourceManager
{
    public function getPublicPackageResourceUri(string $packageKey, string $relativePathAndFilename, ?UriInterface $baseUri = null): string
    {
        $baseUri ??= $this->generateBaseUriFromHttpRequest();
        // ...
    }

    public function getPublicPersistentResourceUri(PersistentResource $resource, ?UriInterface $baseUri = null)
    {
        $baseUri ??= $this->generateBaseUriFromHttpRequest();
        // ...
    }

    private function generateBaseUriFromHttpRequest(): ?UriInterface
    {
        // copied from the base uri provider without `Neos.Flow.http.baseUri` support
        $activeRequestHandler = $this->bootstrap->getActiveRequestHandler();
        if (!$activeRequestHandler instanceof HttpRequestHandlerInterface) {
            return new Uri();
        }
        $request = $activeRequestHandler->getHttpRequest();
        return RequestInformationHelper::generateBaseUri($request);
    }
}
```

(i left out the demonstration for `ResourceManager::getPublicPersistentResourceUriByHash` as its unused)

Meaning the resource manager can be used with fixed base uris

```php
$this->resourceManager->getPublicPersistentResourceUri($resource, new Uri('http://localhost'));
```

or when leaving this parameter out, the previous default behaviour is used:

```php
$this->resourceManager->getPublicPersistentResourceUri($resource);
```

a little more breaking would be do always enforce to specify the baseUri which would definitely be correct.

In Fusion we could start inside our prototypes to call `getPublicPersistentResourceUri` with the base uri of the current 
request and thus make the life a little easier ;)

- related bug cause https://github.com/neos/neos-development-collection/issues/4888
- related https://github.com/neos/flow-development-collection/issues/2157

Alternative fixes
1. Add a context switcher like `BaseUriProvider::applyBaseUriToClosure($newBaseUri, fn () => $resourcemanger->...(...))`
2. Add an environment variable like `FLOW_BASE_URI` see https://github.com/neos/flow-development-collection/pull/3002
    - (like option 1 but less hacky to the eye, as it could be specified from outside flow ^^)
5.  Use empty base uri (host relative) and dont crash https://github.com/neos/flow-development-collection/pull/3314

mhsdesign:
wip resolves: #3334

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ResourceManagement/ResourceManager.php
+++ b/Neos.Flow/Classes/ResourceManagement/ResourceManager.php
@@ -24,6 +24,7 @@ use Neos\Flow\ResourceManagement\Target\TargetInterface;
 use Neos\Flow\Utility\Algorithms;
 use Neos\Flow\Utility\Environment;
 use Neos\Utility\Unicode\Functions as UnicodeFunctions;
+use Psr\Http\Message\UriInterface;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -381,7 +382,7 @@ class ResourceManager
      * @return string|false A URI as a string or false if the collection of the resource is not found
      * @api
      */
-    public function getPublicPersistentResourceUri(PersistentResource $resource)
+    public function getPublicPersistentResourceUri(PersistentResource $resource, ?UriInterface $baseUri = null)
     {
         $this->initialize();
 
@@ -391,7 +392,7 @@ class ResourceManager
         /** @var TargetInterface $target */
         $target = $this->collections[$resource->getCollectionName()]->getTarget();
         if ($target instanceof AbsoluteBaseUriAwareTarget) {
-            $target->setAbsoluteBaseUri($this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest());
+            $target->setAbsoluteBaseUri($baseUri ?? $this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest());
         }
         return $target->getPublicPersistentResourceUri($resource);
     }
@@ -406,7 +407,7 @@ class ResourceManager
      * @throws Exception
      * @api
      */
-    public function getPublicPersistentResourceUriByHash($resourceHash, $collectionName = self::DEFAULT_PERSISTENT_COLLECTION_NAME)
+    public function getPublicPersistentResourceUriByHash($resourceHash, $collectionName = self::DEFAULT_PERSISTENT_COLLECTION_NAME, ?UriInterface $baseUri = null)
     {
         $this->initialize();
 
@@ -420,7 +421,7 @@ class ResourceManager
             throw new Exception(sprintf('Could not determine persistent resource URI for "%s" because no PersistentResource object with that SHA1 hash could be found.', $resourceHash), 1375347691);
         }
         if ($target instanceof AbsoluteBaseUriAwareTarget) {
-            $target->setAbsoluteBaseUri($this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest());
+            $target->setAbsoluteBaseUri($baseUri ?? $this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest());
         }
         return $target->getPublicPersistentResourceUri($resource);
     }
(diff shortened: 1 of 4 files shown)


Neos.Flow/Classes/ResourceManagement/ResourceManager.php
```diff
@@ -422,12 +435,15 @@ public function getPublicPersistentResourceUriByHash($resourceHash, $collectionN
...
      */
-    public function getPublicPackageResourceUri($packageKey, $relativePathAndFilename)
+    public function getPublicPackageResourceUri($packageKey, $relativePathAndFilename, ?UriInterface $baseUri = null)
     {
         $this->initialize();
 
         /** @var TargetInterface $target */
         $target = $this->collections[self::DEFAULT_STATIC_COLLECTION_NAME]->getTarget();
+        if ($target instanceof AbsoluteBaseUriAwareTarget) {
+            $target->setAbsoluteBaseUri($baseUri ?? $this->baseUriProvider->getConfiguredBaseUriOrFallbackToCurrentRequest());
```
mhsdesign, line 445: technically `absoluteBaseUri` is not fully correct, as the api would also allow passing an empty uri: `new Uri('')` 

Now the value returned by `getConfiguredBaseUriOrFallbackToCurrentRequest` is _likely_ just the plain host but might also contain the dir of the request path `$SCRIPT_NAME`.
That would fit with the description of absolute uri.

Now as soon as `Neos.Flow.http.baseUri` is not `null` even an empty string would be returned which does NOT fit again into idea.

PR #3447, opened 2025-02-13
TASK: Cosmetic followup for #3443

mhsdesign:
During the review of #3443 I couldn't help to do a little cleaning myself :D 🧹 

**Checklist**

- [x] Code follows the PSR-2 coding style
- [ ] Tests have been created, run and adjusted as needed
- [ ] ~The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)~
- [x] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [x] Reviewer - The first section explains the change briefly for change-logs
- [ ] ~Reviewer - Breaking Changes are marked with `!!!` and have upgrade-instructions~

PR #3448, opened 2025-02-17
TASK: Remove obsolete sorting of big arrays in reflection service

mhsdesign:
While cleaning up code in the reflection service https://github.com/neos/flow-development-collection/pull/3447 i noticed the sorting calls and suspected that they have become obsolete over the time. Especially as the reflection was refactored now and that some explanations seem dated.

> after reflecting each class we sort classReflectionData which can be avoided
> but we still ensure that getAllClassNames returns all classes sorted.
> For using `array_diff` to get the $newClassNames we dont need sorting

Locally removing the sort calls is just fine but the CI throws this exception.

```
There was 1 failure:

1) Neos\Flow\Tests\Functional\Reflection\ReflectionServiceTest::aggregateRootAssignmentsInHierarchiesAreCorrect
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Neos\Flow\Tests\Functional\Reflection\Fixtures\Repository\SubSubEntityRepository'
+'Neos\Flow\Tests\Functional\Reflection\Fixtures\Repository\SuperEntityRepository'

/home/runner/work/flow-development-collection/flow-development-collection/flow-development-distribution/Packages/Framework/Neos.Flow/Tests/Functional/Reflection/ReflectionServiceTest.php:90
/home/runner/work/flow-development-collection/flow-development-collection/flow-development-distribution/bin/phpunit:122
```

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -265,8 +265,9 @@ class ReflectionService
         if (!$this->initialized) {
             $this->initialize();
         }
-
-        return array_keys($this->classReflectionData);
+        $classNames = array_keys($this->classReflectionData);
+        sort($classNames);
+        return $classNames;
     }
 
     /**
@@ -1093,8 +1094,6 @@ class ReflectionService
         }
         $classNamesToReflect = array_merge([], ...$availableClassnames);
         $reflectedClassNames = array_keys($this->classReflectionData);
-        sort($classNamesToReflect);
-        sort($reflectedClassNames);
         $newClassNames = array_diff($classNamesToReflect, $reflectedClassNames);
         if ($newClassNames === []) {
             return;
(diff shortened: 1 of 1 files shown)


Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -265,8 +265,9 @@ public function getAllClassNames(): array
         if (!$this->initialized) {
             $this->initialize();
         }
-
-        return array_keys($this->classReflectionData);
+        $classNames = array_keys($this->classReflectionData);
+        sort($classNames);
```
mhsdesign, line 269: as karsten wrote

> Why the sorting? Is that the central replacement for the removed sort calls later on?

yes as the structure is not internally sorted anymore but `getAllClassNames` guaranteed A-Z order before i thought we should continue to do so.
On the other hand, `getAllClassNames` can be considered rather low level and is only used once so we can also put sorting on the callers site - maybe no sorting is needed for every usecase.

PR #3455, opened 2025-03-11
BUGFIX: Set property to null via objects yaml

mhsdesign:
This is probably similar to an `autowiring: false` option, but this does not exist as far as i debugged things for separate properties via objects yaml

The idea was to have an optional object, which can be removed via settings

```php
#[Flow\Inject]
protected SomeService|null $someService = null;
```

```yaml
Foo\Bar\Buz:
  properties:
    someService:
      value: null
```

At the current state invalid php code is generated, see `= ;` which causes php to fail during parsing:

```php
if (\Neos\Utility\ObjectAccess::setProperty($this, 'someService', ) === false) { $this->someService = ;}
```

**Upgrade instructions**

**Review instructions**

**Checklist**


kitsunet (approved):

kitsunet:
LOL thx github

Issue #2553, opened 2021-08-31
class(.*\\SpecialController) leads to ProxyBuilding of TYPO3Fluid\\Fluid\\Core\\Variables\\StandardVariableProvider


fcool:
### Description

If you use a matcher in your Policy.yaml like that:
```
privilegeTargets:
  Neos\Flow\Security\Authorization\Privilege\Method\MethodPrivilege:
    MyPrivilegeName:
      matcher: "class(.*\\StartController)"
```
Flow will build a proxy for `TYPO3Fluid\\Fluid\\Core\\Variables\\StandardVariableProvider` which then does not match `TYPO3Fluid\\Fluid\\Core\\Variables\\VariableProviderInterface` because of the constructor definition.

#### Expected behavior

While we probably cannot change the "awkward" decision of `TYPO3Fluid` we probably should not all of sudden build proxy for ever class, if we stumble over such an expression.

I'm still unsure, where the problem exactly arises but believe it comes from here:
`\Neos\Flow\Aop\Pointcut\PointcutClassNameFilter::reduceTargetClassNames` but I cannot get the big picture together.

Flow: 6.3

bwaidelich:
The proxy building shouldn't be affected by your policy, i.e. proxies are always created unless a class has a `Proxy(false)` annotation or isn't a class. => Flow won't create a proxy for `VariableProviderInterface`.

> resulting in a PHP Fatal error.

Could you share that?

fcool:
The error?

```
[Tue Aug 31 19:21:55.347686 2021] [proxy_fcgi:error] [pid 30257:tid 140693842413120] [client 127.0.0.1:44376] AH01071: Got error 'PHP message: PHP Fatal error:  Declaration of TYPO3Fluid\\Fluid\\Core\\Variables\\StandardVariableProvider::__construct() must be compatible with TYPO3Fluid\\Fluid\\Core\\Variables\\VariableProviderInterface::__construct(array $variables = Array) in /projectpath/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/TYPO3Fluid_Fluid_Core_Variables_StandardVariableProvider.php on line 410'
```

All it needs to trigger is an empty project with a Policy.yaml like above

bwaidelich:
Ah, the problem is that the interface declares a constructor (which is considered bad style btw) in this case.
That and the fact that apparently Flow changes the constructor for the `StandardVariableProvider` somehow.
Is it possible that your privilege matches that class somehow?

[...]

PR #3458, opened 2025-03-16
BUGFIX: Enums throw `::__construct() does not exist` on compilation

mhsdesign:
`flow flow:core:compile`

```
Exception [0] in line 666 ReflectionService.php: Method Neos\Demo\Components\Headline\HeadlineType::__construct() does not exist

22 ReflectionMethod::__construct("Neos\Demo\Components\Headline\HeadlineType", "__construct")
21 Neos\Flow\Reflection\ReflectionService::getMethodAnnotations("Neos\Demo\Components\Headline\HeadlineType", "__construct", "Neos\Flow\Annotations\Autowiring")
20 Neos\Flow\Reflection\ReflectionService::getMethodAnnotation("Neos\Demo\Components\Headline\HeadlineType", "__construct", "Neos\Flow\Annotations\Autowiring")
19 Neos\Flow\ObjectManagement\Configuration\ConfigurationBuilder::autowireArguments(array|1681|)
18 Neos\Flow\ObjectManagement\Configuration\ConfigurationBuilder::buildObjectConfigurations(array|28|, array|32|)
17 Neos\Flow\ObjectManagement\CompileTimeObjectManager::initialize(array|165|)
16 Neos\Flow\Core\Booting\Scripts::initializeObjectManagerCompileTimeFinalize(Neos\Flow\Core\Bootstrap)
15 call_user_func(array|2|, Neos\Flow\Core\Bootstrap)
```

Adding the enum check works.

```php
if (enum_exists($classOrInterfaceName)) {
    continue;
}
```

The question is why does it only cause an error in my project? Because of the odd class path `Components` instead of `Classes`?
Anyways enum support was introduced via https://github.com/neos/flow-development-collection/pull/2702 and should not be a problem ...

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationBuilder.php
@@ -87,6 +87,10 @@ class ConfigurationBuilder
                     continue;
                 }
 
+                if (enum_exists($classOrInterfaceName)) {
+                    continue;
+                }
+
                 if (interface_exists($classOrInterfaceName)) {
                     $interfaceName = $classOrInterfaceName;
                     $implementationClassName = $this->reflectionService->getDefaultImplementationClassNameForInterface($interfaceName);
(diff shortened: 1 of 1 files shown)


kitsunet:
What, where is this failing file?

mhsdesign:
its an presentational enum that i placed in `Neos.Demo/Components` ... 

i suspect this happens because the point-cut `class(.*\\....)` attempts to build a proxy class for every class in the system as long as that is still true: https://github.com/neos/flow-development-collection/issues/2553#issuecomment-1184381853

PR #3459, opened 2025-03-19
FEATURE: Add CLDR-based date, time, and datetime formatting methods with "formatLength"

dlubitz:
Adds feature to format dates, times and datetimes with the CLDR "Basic Formats" (known as "Format length" in Flow) to the Date Eel helper. 

This feature was already implemented in fluids format viewhelper before.

See:
* https://cldr.unicode.org/translation/date-time/date-time-patterns#basic-date-formats
* https://cldr.unicode.org/translation/date-time/date-time-patterns#basic-time-formats
* https://neos.readthedocs.io/en/stable/References/ViewHelpers/FluidAdaptor.html#f-format-date

### DIFF
--- a/Neos.Eel/Tests/Unit/Helper/DateHelperTest.php
+++ b/Neos.Eel/Tests/Unit/Helper/DateHelperTest.php
@@ -127,6 +127,72 @@ class DateHelperTest extends \Neos\Flow\Tests\UnitTestCase
     /**
      * @test
      */
+    public function formatCldrDate()
+    {
+        $date = \DateTime::createFromFormat('Y-m-d H:i:s', '2013-07-03 12:34:56');
+        $formatLength = 'long';
+        $localeString = 'en';
+
+        $locale = new Locale('en');
+        $expectedString = '2013-07-03 12:34:56';
+
+        $formatMock = $this->createMock(\Neos\Flow\I18n\Formatter\DatetimeFormatter::class);
+        $formatMock->expects(self::atLeastOnce())->method('formatDate')->with($date, $locale, $formatLength)->willReturn($expectedString);
+
+        $helper = new DateHelper();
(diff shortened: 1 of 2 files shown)


Sebobo (changes requested):
Thx, works great when testing in the Demo site. 

Can you also update the `\Neos\Eel\Tests\Unit\DateHelperTest` with the new methods.

dlubitz:
Done @Sebobo

Neos.Eel/Tests/Unit/Helper/DateHelperTest.php
```diff
@@ -124,6 +124,72 @@ public function formatCldrCallsFormatService()
...
+     * @test
+     */
+    public function formatCldrTime()
+    {
+        $date = \DateTime::createFromFormat('Y-m-d H:i:s', '2013-07-03 12:34:56');
+        $formatLength = 'long';
+        $localeString = 'en';
+
+        $locale = new Locale('en');
+        $expectedString = '2013-07-03 12:34:56';
```
Sebobo, line 159: Shouldn't the expected string be just the time and in the other test just the date, all three methods now have the same output in the test?

Sebobo: Ah now I get it, the mock is just copied.

Sebobo (approved):

Neos.Eel/Tests/Unit/Helper/DateHelperTest.php
dlubitz: Yeah, it doesn't matter as the test, simply checks if the params are passed properly to the service and if the result will be returned.

kdambekalns (commented):
👏

PR #3462, opened 2025-03-28
BUGFIX: #3442 ignore deprecations during binary check

### LINKED ISSUE #3442: BUG: PHP binary check may report false negatives

### Current Behavior

With Xdebug enabled but no listener to connect to, the command used to check the PHP version and SAPI "fails".

https://github.com/neos/flow-development-collection/blob/57bc20b2bc49cee030910338c93fd20d5a86fd38/Neos.Flow/Classes/Core/Booting/Scripts.php#L939-L951

Because when Xdebug cannot connect, the output looks like this:

```
Xdebug: [Step Debug] Could not connect to debugging client. Tried: …:9003 (fallback through xdebug.client_host/xdebug.client_port).
{"sapi":"cli","version":"8.2.27"}
```

Note the expected JSON is now in the **second line** of the output (instead of `$output[0]`), so the check fails, even though `$result` is `0`.

### Expected Behavior

The Xdebug output is a warning, and, well hard to suppress, unfortunately. But it should not stop execution of the code… especially since warming up Flow from the CLI works just fine and also subsequent browser requests do work.

### Steps To Reproduce

1. Activate Xdebug, set `xdebug.start_with_request=yes` (as is default in DDEV)
2. Do **not** listen for debug connections
3. Call Flow from a browser, with empty Flow caches

### Environment

```markdown
- Flow: 8.3
- PHP: 8.3
- Xdebug: 3
```

### Anything else?

- A way to solve this would be to not redirect stderr to stdout? No, that Xdebug output is not sent to stderr.
- Unset the `subRequestEnvironmentVariables` to get rid of `XDEBUG_CONFIG`? No, with `xdebug.start_with_request=yes` that doesn't help.

mhsdesign:
fixes: https://github.com/neos/flow-development-collection/issues/3442

seems kinda hacky i dont know how else we can easily make the output parseable.
I guess we _could_ turn of deprecations and all error handling for that request but i dindt want to start the fun to make these parameters work cross platform (windows) so here is an easy fix.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -944,7 +944,8 @@ class Scripts
 
         exec(join(' ', $command), $output, $result);
 
-        $phpInformation = json_decode($output[0] ?? '{}', true) ?: [];
+        $resultWithJson = end($output); // For multiple lines use the last line, to skip possible emitted deprecations
+        $phpInformation = $resultWithJson ? json_decode($resultWithJson, true) : false;
 
         if ($result !== 0 || ($phpInformation['sapi'] ?? null) !== 'cli') {
             throw new Exception\SubProcessException(sprintf('PHP binary might not exist or is not suitable for cli usage. Command `%s` didnt succeed.', $phpCommand), 1689676967447);
--- a/Neos.Flow/Classes/Core/Booting/Scripts.php
+++ b/Neos.Flow/Classes/Core/Booting/Scripts.php
@@ -874,9 +874,9 @@ class Scripts
         $output = [];
         exec(join(' ', $command), $output, $result);
 
-        if ($result === 0 && count($output) === 1) {
+        if ($result === 0 && $output !== []) {
             // Resolve any wrapper
-            $configuredPhpBinaryPathAndFilename = $output[0];
+            $configuredPhpBinaryPathAndFilename = end($output); // Last entry, skip possible emitted deprecations
         } else {
             // Resolve any symlinks that the configured php might be pointing to
             $configuredPhpBinaryPathAndFilename = realpath($phpBinaryPathAndFilename);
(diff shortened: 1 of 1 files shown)


kdambekalns:
Slightly related, as it eases finding such issues: https://github.com/neos/flow-development-collection/pull/3441

kitsunet (changes requested):
I am not convinced we should do that, there could be other output apart from deprecations that should block using the binary, also like if there are deprecations maybe use a log level that silences them if you want to use that php version?

mhsdesign:
This issue just popped up in [slack](https://neos-project.slack.com/archives/C04V4C6B0/p1762415970257259) when there is an error like `OPcache already loaded`

I think we should reconsider if the check really has to be done at (production) runtime or if this should be part of the (one time) neos setup (`./flow setup` and `/setup` endpoint). Adding this validation to the setup would allow us to give more verbose output to the user. Like if the php binary emits odd stuff i want to know about it ^^. But if this all has to be be encapsulated when the actual runtime code already attempts to start a php subprocess its a little late for this guidance. (Especially if this is deployed already wrongly)

So i think we should try to remove the (expensive) php validation here from the `Scripts` and instead fully put the load to the setup with more verbose information. The only problem is that we need to see if we want people to have the neos setup package installed on the production server to see possible problems ... maybe hidden by login after all...

dlubitz:
But in the setup this doesn't help at all. Mostly I run into this during changing the PHP versions in already productive running systems and you forgot to change the binary path in the config.

skurfuerst:
Hmmm, I think it is problematic to remove this check if we cannot hint to the user that something is wrong. In fact we would need to point the user to the setup in error cases; otherwise they will not find it... I need to think it through more carefully though.

kdambekalns:
> when there is an error like `OPcache already loaded`

With #3441 the cause should be clearly visible. And then you can just fix it, no?

PR #3463, opened 2025-04-01
BUGFIX: Remove unused variable assignment

### LINKED ISSUE #3395: BUG: EntityPrivilegeExpressionEvaluator accesses undefined property

### Current Behavior

When `\Neos\Flow\Security\Authorization\Privilege\Entity\Doctrine\EntityPrivilegeExpressionEvaluator::evaluate` is executed and deprecations are part of the logged errors, this error occrus:

`Creation of dynamic property Neos\Flow\Security\Authorization\Privilege\Entity\Doctrine\EntityPrivilegeExpressionEvaluator::$newExpressions is deprecated in /var/www/html/Data/Temporary/Production/SubContextBehat/Cache/Code/Flow_Object_Classes/Neos_Flow_Security_Authorization_Privilege_Entity_Doctrine_EntityPrivilegeExpressionEvaluator.php line 45`
[](url)

### Expected Behavior

There is no deprecation error message.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Flow: 8.3
- PHP: 8.3
```

### Anything else?

This problem is a regression of [this commit](https://github.com/neos/flow-development-collection/commit/642a11ca489fe508b0ab40ebaa0c16f9a51397b4).
The use of the property `newExpressions` was not renamed in the extending class.

The implementation of the code generation differs between extended and extending class. Since `EntityPrivilegeExpressionEvaluator` does not use the expression cache but instead checks the function existence directly, I think the line
`$this->newExpressions[$functionName] = $functionDeclaration;`
should be removed completely instead of renaming `newExpressions` to `evaluatedExpressions`.

theilm:
Fixes https://github.com/neos/flow-development-collection/issues/3395

Without the fix, there are warnings in current PHP versions.

### DIFF
--- a/Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/EntityPrivilegeExpressionEvaluator.php
+++ b/Neos.Flow/Classes/Security/Authorization/Privilege/Entity/Doctrine/EntityPrivilegeExpressionEvaluator.php
@@ -42,7 +42,6 @@ class EntityPrivilegeExpressionEvaluator extends CompilingEvaluator
         if (!function_exists($functionName)) {
             $code = $this->generateEvaluatorCode($expression);
             $functionDeclaration = 'function ' . $functionName . '($context){return ' . $code . ';}';
-            $this->newExpressions[$functionName] = $functionDeclaration;
             eval($functionDeclaration);
         }
 


kitsunet (approved):
Fine by me, thanks for noticing! 

Note for later: We should probably take care of having cache here again. Haven't looked into the history but I assume way back we used a shutdown object function to turn the array of newExpressions into cache entries but that is no longer the case. We should probably look at \Neos\Eel\CompilingEvaluator::evaluate() on how to deal with cache...

mhsdesign (changes requested):
jup as it seems there was indeed a cache in the `CompilingEvaluator` which was renamed to `$evaluatedExpressions`.

https://github.com/neos/flow-development-collection/commit/642a11ca489fe508b0ab40ebaa0c16f9a51397b4

so the correct fix should be to adjust the naming here instead as bugfix for 8.3

mhsdesign (approved):
Then again as no one worked on this still and i just found out that since the introduction of the `EntityPrivilegeExpressionEvaluator` the cache was never read but only written:

https://github.com/neos/flow-development-collection/commit/5fd0550ee489b32ab1c6d76a0b10eb0cb839f575#diff-905a50aa348c7db8a665387d4cd7358982bca1b904d4ef020bfdaa17daaaf97b

This means its not a regression and just a copy paste error thus indeed it might just be simpler to remove the code.

Issue #3475, opened 2025-05-05
BUG: Eel parser error when node names start with numbers


Sebobo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The Flowquery `q(site).find('/sites/helzleit/abc/123-abc').get()` throws the following error:

`The Selector "/sites/helzleit/abc/123-abc" could not be parsed. Error at character 20.`

The FlowQuery `q(site).find('/sites/helzleit/abc/abc-123').get()` works.

But nodes starting with numbers in their names can be created without issue and the name patterns in the core also allow this. The error is thrown in the `FizzleParser`.

### Expected Behavior

The query should work without throwing an error.

### Steps To Reproduce

See above

### Environment

```markdown
- Flow: 8.3
- Neos: 8.3
- PHP: 8.x
```

mhsdesign:
indeed names are allowed to start with a number or just any allowed char - :O

now the parses `PathFilter` specifies to be `Identifier` separated by `/`

https://github.com/neos/flow-development-collection/blob/19c16698661c9a17a150bab96a912446ff1ad606/Neos.Eel/Resources/Private/Grammar/Fizzle.peg.inc#L65

And an `Identifier` is registered here which explicitly says it must start with a letter.

https://github.com/neos/flow-development-collection/blob/69ede5ad8682e65d2106656f43adee2053ae6861/Neos.Eel/Resources/Private/Grammar/AbstractParser.peg.inc#L44

Now fixing the shared `Identifier` which is used by the eel parser would be a bad idea but we can just introduce a specific `NodeName` in `Fizzle.peg.inc` and use that inside the `PathFilter`

```
NodeName: / [a-z0-9\-]+ /
```

The parser needs to be rebuild for that ofcourse and i never dunnit and idk if the tech allows to still boot up from the dead :D its ancient.

dlubitz:
Ok, I could rebuild Fizzle with the changes you suggested. That fixes the new test, but fails an old one.

```
1) Neos\Eel\Tests\Unit\FlowQuery\FizzleParserTest::pathFilterIsMatched
Assert parser method Filter matches string foo/node_1234-5678
Failed asserting that false is true.
```

The issue is the underscore here, which is not allowed in nodename

mhsdesign:
thanks @dlubitz 

~that seems like a good trade! Because underscores are just invalid - also in Neos 9.0~

[...]

PR #3476, opened 2025-05-05
BUGFIX: Accept node names starting with numbers in fizzle parser

Sebobo:
Resolves: #3475

### DIFF
--- a/Neos.Eel/Classes/FlowQuery/FizzleParser.php
+++ b/Neos.Eel/Classes/FlowQuery/FizzleParser.php
@@ -37,72 +37,60 @@ function match_ObjectIdentifier ($stack = array()) {
 }
 
 
-/* NodeName: / [a-z0-9\-]+ / */
-protected $match_NodeName_typestack = array('NodeName');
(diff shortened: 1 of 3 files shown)


Sebobo:
Added the test that shows the issue, no clue yet how to solve the actual problem.

mhsdesign (commented):
Thanks. I believe that this is what we want as outlined here https://github.com/neos/flow-development-collection/issues/3475#issuecomment-2851873816

Fizzle is generic and doesnt have to be only focused on neos CR nodes so restricting it to noden-names like `[a-z0-9\-]+` would be too strict.

Im not sure in what way the AST will be affected if there is even one for the parser ... in that case this warrants for a neos adjustment and all callsites of the parser ... but it seems the API of the parser is that it returns a string joined by `/` for `PathFilter` so nothing changes.

mhsdesign (approved):

kitsunet (approved):
makes sense, thank you!

kitsunet:
style ci change is unrelated, merging

PR #3479, opened 2025-05-16
TASK: Allow passing sorting option in Array.unique helper

dlubitz:
This allows to pass an expected sorting behavior for array_unique, which enables the possibility to filter an array of objects (like nodes).

```
uniqueNodes = Array.unique(nodes, "SORT_REGULAR")
```

### DIFF
--- a/Neos.Eel/Classes/Helper/ArrayHelper.php
+++ b/Neos.Eel/Classes/Helper/ArrayHelper.php
@@ -315,15 +315,20 @@ class ArrayHelper implements ProtectedContextAwareInterface
     /**
      * Removes duplicate values from an array
      *
+     *  Usage example for options:
+     *
+     *  Array.unique(value, 'SORT_REGULAR')
+     *
      * @param iterable $array The input array
+     * @param string $option The constant name for sorting behavior ('SORT_STRING', 'SORT_REGULAR', 'SORT_NUMERIC', 'SORT_LOCALE_STRING')
      * @return array The filtered array.
      */
-    public function unique(iterable $array): array
+    public function unique(iterable $array, string $option = 'SORT_STRING'): array
     {
         if ($array instanceof \Traversable) {
             $array = iterator_to_array($array);
         }
-        return array_unique($array);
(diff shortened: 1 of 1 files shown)


Sebobo (approved):
LGTM by reading, thx

Sebobo:
You want to target 9.0 with this?

dlubitz:
We can also put it in 8.3. WDYT?

Sebobo:
One might say its a feature -> 9.1
If it's just a task/bugfix -> 8.3

dlubitz:
I'd consider this as a bugfix as it fixes a missing parameter 😁 ... and I need it 9.0

mhsdesign:
i think it was not part of the api before as we try to be ECMA script compliant: 

> The implementation uses the JavaScript specificiation where applicable, including EcmaScript 6 proposals.
> See https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array for a documentation and
> specification of the JavaScript implementation.

Though id say this failed already in most ways - especially that arrays can have keys like in php but not in javascript.

As it seems youre trying to use node objects here, id just recommend to use `\Neos\ContentRepository\NodeAccess\FlowQueryOperations\UniqueOperation` 

`q(nodes).unique().get()` 

Then we can save us this change and putting the php hacks into eel userland :D

dlubitz:
@mhsdesign Nice idea.... bût the flowquery operation doesn't respect this parameter too. I guess this operation will fail always, if you use nodes.

https://github.com/neos/eel/blob/408ee4f207969a8359d8148fcab8f3987e717c89/Classes/FlowQuery/Operations/UniqueOperation.php#L41-L41

dlubitz:
Ok, I've tested it and it works. IDK why or what the difference is compared to the EEL helper, but it works. 🤷‍♂️

mhsdesign:
because the cr overloads the flowquery for nodes `\Neos\ContentRepository\NodeAccess\FlowQueryOperations\UniqueOperation`

PR #3482, opened 2025-05-23
BUGFIX: Handle stand-alone format for weekdays

dlubitz:
Add proper handling for stand-alone format of weekdays.

See: https://cldr.unicode.org/translation/date-time/date-time-symbols

### DIFF
--- a/Neos.Flow/Tests/Unit/I18n/Cldr/Reader/DatesReaderTest.php
+++ b/Neos.Flow/Tests/Unit/I18n/Cldr/Reader/DatesReaderTest.php
@@ -164,9 +164,10 @@ class DatesReaderTest extends UnitTestCase
             ['HH:mm:ss zzz', ['HH', [':'], 'mm', [':'], 'ss', [' '], 'zzz']],
             ['EEE, MMM d, \'\'yy', ['EEE', [','], [' '], 'MMM', [' '], 'd', [','], [' '], ['\''], 'yy']],
             ['hh \'o\'\'clock\' a, zzzz', ['hh', [' '], ['o'], ['\''], ['clock'], [' '], 'a', [','], [' '], 'zzzz']],
-            ['QQyyLLLLDFEEEEccc', ['QQ', 'yy', 'LLLL', 'D', 'F', 'EEEE', 'ccc']],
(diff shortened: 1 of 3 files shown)


mhsdesign:
hi thanks;) i dont know what that really does now but i suspect we need to adjust our documentation for the new case 'c' and also a test to document that this feature exist and is valid and what should it do;)

dlubitz:
I've added the tests. But I'm not aware of any documentation, where we need to add this. AFAIS we simply reference the CLDR docs.

Neos.Flow/Tests/Unit/I18n/Cldr/Reader/DatesReaderTest.php
```diff
@@ -164,7 +164,7 @@ public function formatStringsAndParsedFormats()
             ['HH:mm:ss zzz', ['HH', [':'], 'mm', [':'], 'ss', [' '], 'zzz']],
             ['EEE, MMM d, \'\'yy', ['EEE', [','], [' '], 'MMM', [' '], 'd', [','], [' '], ['\''], 'yy']],
             ['hh \'o\'\'clock\' a, zzzz', ['hh', [' '], ['o'], ['\''], ['clock'], [' '], 'a', [','], [' '], 'zzzz']],
-            ['QQyyLLLLDFEEEE', ['QQ', 'yy', 'LLLL', 'D', 'F', 'EEEE']],
+            ['QQyyLLLLDFEEEEccc', ['QQ', 'yy', 'LLLL', 'D', 'F', 'EEEE', 'ccc']],
```
kdambekalns: Why does this change the test and not add one? Is the previously tested case invalid?

kdambekalns (approved):
👍 by 👀 – but added a question…

Benjamin-K:
Couldn't the date formatting be handled by the [IntlDateFormatter](https://www.php.net/manual/en/class.intldateformatter.php) class? So we could replace lots of code on our side?

Neos.Flow/Tests/Unit/I18n/Cldr/Reader/DatesReaderTest.php
dlubitz: I've put them into new tests.

kitsunet (approved):
Great!

PR #3485, opened 2025-06-12
BUGFIX: Disallow singleton route part handlers

bwaidelich:
Using singleton route part handlers can have severe side effects because they are shared between all route instances.

This change leads to an exception being thrown if a singleton handler is configured

Related: https://github.com/neos/neos-development-collection/issues/5571


bwaidelich:
**DISCLAIMER:** I changed this code in the browser, didn't test it yet!

kitsunet (approved):
I guess that's fine... Haven't tested but should work.
If you _need_ a shared runtime cache that can always be a singleton you inject into the handler or so...

PR #3486, opened 2025-06-21
BUGFIX: out of memory error when using AbsorbingHandler

### LINKED ISSUE #3487: BUG: out of memory error when using AbsorbingHandler for exceptions

### Current Behavior

> Fatal error: Allowed memory size of 4294967296 bytes exhausted (tried to allocate 20480 bytes) in /app/Data/Temporary/Development/SubContextDocker/Cache/Code/Flow_Object_Classes/Neos_Neos_Fusion_NodeUriImplementation.php on line 162

### Expected Behavior

An exception log file instead of the memory error.

### Steps To Reproduce

There are different ways to trigger the error, I use the [following example](https://github.com/Flowpack/Flowpack.Neos.FrontendLogin):

First you need a content element of type `Neos.Neos:Plugin` which uses `Neos.Neos:NodeUri`.

```
prototype(Flowpack.Neos.FrontendLogin:LoginForm) < prototype(Neos.Neos:Plugin) {
        …
	redirectAfterLoginUri = Neos.Neos:NodeUri {
  		node = ${q(node).property('redirectAfterLogin')}
	}
       …
}
```

Note that `redirectAfterLogin` has no default value thus may be `null`.

### Environment

```markdown
- Flow: 9.0 and 9.1
- Neos: 9.0 and 9.1
- PHP: 8.3
```

### Anything else?

When `redirectAfterLogin` is `null`, `NodeUriImplementation::evaluate()` called in `PluginImplementation::evaluate()` throws.

Since the `NodeUri` uses `@exceptionHandler = 'Neos\\Fusion\\Core\\ExceptionHandlers\\AbsorbingHandler'` the exception is ignored and a log is generated by calling `Debugger::getObjectSnippetPlaintext(object)`.

At some point, `object` is an instance of `PluginImplementation` which implement `__toString()`. Hence, `__toString()` is used to create the log, but it looks as follows:

```php
/**
 * @return string
 */
public function __toString()
{
    return $this->evaluate();
}
```

This causes an endless loop rendering the stack trace until the memory limit is reached.

christoph-daehne:
Exceptions while rendering error stack traces no longer exhaust all memory and crash.

Resolves https://github.com/neos/flow-development-collection/issues/3487

**Review instructions**

You find details in https://github.com/neos/flow-development-collection/issues/3487

**Checklist**

- ~~Tests have been created, run and adjusted as needed~~
    - bug was introduced in _9.0_

### DIFF
--- a/Neos.Flow/Classes/Error/Debugger.php
+++ b/Neos.Flow/Classes/Error/Debugger.php
@@ -402,7 +402,6 @@ class Debugger
         $exception = new \Exception();
         $callstack = $exception->getTraceAsString();
         if (substr_count($callstack, __METHOD__) > 3) {
-
             $explanation = <<<'EOT'
                 There seems to be an error while rendering the stack trace, see https://github.com/neos/neos-development-collection/issues/5576 .
                 This might happen due to broken __toString() methods: an error occurs -> we generate the log here -> call __toString() again -> error occurs again -> we generate the log here again -> call __toString() again -> ...
--- a/Neos.Flow/Classes/Error/Debugger.php
+++ b/Neos.Flow/Classes/Error/Debugger.php
@@ -399,6 +399,18 @@ class Debugger
      */
     protected static function getBacktraceCodePlaintext(array $trace, bool $includeCode = true): string
     {
+        $exception = new \Exception();
+        $callstack = $exception->getTraceAsString();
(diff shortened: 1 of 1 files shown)


kitsunet (approved):
Clever idea, I like!

PR #3489, opened 2025-07-15
BUGFIX: Prevent premature connection to database before `PersistenceManager::persistAllowedObjects`

mhsdesign:
For every simple GET request `persistAllowedObjects` is called via the Package.php in flow. This results in building a db connection, even if there is nothing to do.

We first check if the `entityManager` is open and if the entity manager has not been used (something is heavily cached via middle-ware or sth) flows objectmanagement retrieves the entity manger.

Now in this retrieval process - even though doctrines connection is lazy itself - the connection will be made forcefully see `\Neos\Flow\Persistence\Doctrine\EntityManagerFactory::create` line 120 (https://github.com/neos/flow-development-collection/blob/11e2348125dd8286ff9ccc088e5d187dc9143bf5/Neos.Flow/Classes/Persistence/Doctrine/EntityManagerFactory.php#L120) or https://github.com/neos/flow-development-collection/blob/d93b6b09ca2071c87812a9ef4bc120201c44608a/Neos.Flow/Classes/Persistence/Doctrine/EntityManagerConfiguration.php#L229

This is ironic because if there is no connection - or no entity manager in the first place, the current process cannot have made any changes to the transaction.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Package.php
+++ b/Neos.Flow/Classes/Package.php
@@ -66,9 +66,11 @@ class Package extends BasePackage
         $dispatcher = $bootstrap->getSignalSlotDispatcher();
 
         $dispatcher->connect(Mvc\Dispatcher::class, 'afterControllerInvocation', function ($request) use ($bootstrap) {
-            // No auto-persistence if there is no PersistenceManager registered
+            // No auto-persistence if there is no PersistenceManager loaded
+            // This signal will not be fired at compile time, as it's only emitted for web requests which happen at runtime.
             if (
-                $bootstrap->getObjectManager()->has(Persistence\PersistenceManagerInterface::class)
+                /** @phpstan-ignore-next-line the object manager interface doesn't specify this method */
+                $bootstrap->getObjectManager()->hasInstance(Persistence\PersistenceManagerInterface::class)
             ) {
                 if (!$request instanceof Mvc\ActionRequest || SecurityHelper::hasSafeMethod($request->getHttpRequest()) !== true) {
                     $bootstrap->getObjectManager()->get(Persistence\PersistenceManagerInterface::class)->persistAll();


kitsunet:
Please have a look at the Flow Package.php, which runs the persistAll after controller. There is a condition in there on:

`$bootstrap->getObjectManager()->has(Persistence\PersistenceManagerInterface::class)`

this is wrong though, not sure if that has ever changed in the object manager or if the condition was wrong since ever, but flipping that to:

`$bootstrap->getObjectManager()->hasInstance(Persistence\PersistenceManagerInterface::class)`

Should fix this as well?!

kdambekalns:
> Please have a look at the Flow Package.php, which runs the persistAll after controller. There is a condition in there on:

Good catch!

> `$bootstrap->getObjectManager()->has(Persistence\PersistenceManagerInterface::class)`
> 
> this is wrong though, not sure if that has ever changed in the object manager

The `has()` was added with #1163 by @kitsunet 😎

> or if the condition was wrong since ever, but flipping that to:
> 
> `$bootstrap->getObjectManager()->hasInstance(Persistence\PersistenceManagerInterface::class)`

That was changed from `hasinstance()` to `has()` in #1366 (a PR by @kitsunet) – but in this case it was @kdambekalns who made [that particular commit](https://github.com/neos/flow-development-collection/pull/1366/commits/163e204df2b724555af2c2a6fd800bd32eb0b95c) in that PR. 🙈

Whatever made me think `hasInstance()` is deprecated and `has()` was the replacement… 🤷‍♂️

kitsunet:
Thanks for looking that up.

I actually expected it was myself breaking my own code from before 😂. `has` is part of the Psr ContainerInterface so that might have looked sensible in the "change to PSR" methods change. Sadly there is no ContainerInterface replacement for hasInstance

mhsdesign:
> Should fix this as well?!

indeed. Good catch thank you both for digging into that as well.

Adjusted that and tested it:)

kdambekalns (approved):
👍 by 👀 – wow, I didn't realize this was still open… Thanks for bringing it up, @Sebobo!

mhsdesign:
Okay thanks;) Ill like to merge this into 9.0 as planned as this is something between an actual bugfix and a task ... not that urgent imo and past me created it agains 9.0 ... so if you let me lets merge <3

kdambekalns:
I know you are on the 9.x train, but this would make very much sense for 8.3, too, no?

Also, the title no longer properly reflects the PR content. 😎

mhsdesign:
Okay i threw the anchor the train took a full stop and will restart from platform 8.3/1

mhsdesign:
Okay during upmerge to 9.0 we now have to add this below as these lines for CLI didnt exist in 8.3 here at this place. 🤦 
-> see https://github.com/neos/flow-development-collection/pull/3296 why

```php
        $dispatcher->connect(Cli\Dispatcher::class, 'afterControllerInvocation', function () use ($bootstrap) {
            // No auto-persistence if there is no PersistenceManager loaded or during compile time
            if (
                /** @phpstan-ignore-next-line the object manager interface doesn't specify this method */
                $bootstrap->getObjectManager()->hasInstance(Persistence\PersistenceManagerInterface::class)
                && !($bootstrap->getObjectManager() instanceof CompileTimeObjectManager)
            ) {
                $bootstrap->getObjectManager()->get(Persistence\PersistenceManagerInterface::class)->persistAll();
            }
        });
```

PR #3491, opened 2025-08-15
BUGFIX: Support IPv6 in `getTrustedClientIpAddress()`

kdambekalns:
When using IPv6, the method incorrectly returns invalid results. For an IPv6 address like `2001:db8:cafe::17` it would return `2001`, which is wrong.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/Http/Middleware/TrustedProxiesMiddleware.php
+++ b/Neos.Flow/Classes/Http/Middleware/TrustedProxiesMiddleware.php
@@ -274,9 +274,20 @@ class TrustedProxiesMiddleware implements MiddlewareInterface
 
         $ipAddress = false;
         foreach (array_reverse($trustedIpHeader) as $headerIpAddress) {
-            $portPosition = strpos($headerIpAddress, ':');
-            $ipAddress = $portPosition !== false ? substr($headerIpAddress, 0, $portPosition) : $headerIpAddress;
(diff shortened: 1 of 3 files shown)


dlubitz (dismissed):
LGTM. Thank you

BUT... Tests are failing, right?

kdambekalns:
Yep, indeed… But why does it pass on PHP 8.3 and 8.4? 🤔

dlubitz:
🤷‍♂️ It doesn't seem to be your code change in the middleware:
https://3v4l.org/Lsflt

kdambekalns:
It fails earlier in the code, so the code changed in this PR isn't even the culprit. This:
```php
filter_var($ipAddress, FILTER_VALIDATE_IP, FILTER_FLAG_NO_RES_RANGE | FILTER_FLAG_NO_PRIV_RANGE)
```
is supposed to not return false for the expected client IP address, in the test it's `2001:db8:cafe::17`. Checking this in 3v4l.org it becomes clear this is quite unreliable in past versions of PHP…

image
(see https://3v4l.org/RqTcT)

Obviously filtering IP addresses is hard, see https://bugs.php.net/bug.php?id=61700 for an illustrative example…

So I'll "fix" this by skipping the test for PHP 8.2, < PHP 8.3.16 and PHP < 8.4.3

dlubitz (approved):
Thank you!

PR #3492, opened 2025-08-18
!!! FEATURE: Fluid 4.2

kitsunet:
This updates the external Fluid dependency to version 4.2 making it
an upgrade of two major versions.
This includes some deprecations and major changes. Much of which we
cover up in our API but if you extended Fluid or used internals your
code might be affected.


kitsunet:
Now this is is a tricky one, clearly more of a 10.0 thing, due to the unpredictable breakiness possibility. But we kinda need to replace it because it holds pretty much everything back...

kitsunet:
I'll have a look at the static analysis, as for the rest of the tests, they are not _really_ conclusive for Fluid. So that they are green doesn't have to mean much.

nezaniel:
On a related note:

PHP 8.3 will be supported until December 2027
Neos 8.3 and 8.4 are supposed to be supported until August 2028
Neos 9.0 through 9.2 are supposed to be supported until April 2028
Neos 9.3 is supposed to be supported until August 2029

so release-cycle-wise this should even target the 8.3 branch. Do you see any chance to do this or is this completely out of the question? 

Otherwise I'd say let's target 9.1 because otherwise 9.3 would not really be anything close to an LTS version

sorenmalling:
Could we go for the Fluid 5 branch instead - 5.3 just released. Let's take the big jump while we are at it

sorenmalling:
@kitsunet If I was to work on bringing the update to Fluid 5, can you share any required changes to root composer.json or whatever to have it install the version set in the composer.json of the collection package?

mhsdesign:
What is our plan now regarding what bsc said? https://github.com/neos/flow-development-collection/pull/3492#issuecomment-3199952643 Target the upcoming 9.2 because there are no other changes?

Fluid 5 sounds even more breaking but if Neos survives this im fine with it^^

PR #3406, opened 2024-10-25
BUGFIX: Adjust to PHP 8.3 get_parent_class() deprecation

lorenzulrich:
Somehow, https://github.com/neos/flow-development-collection/pull/3351/files only targeted Flow 9.0, but per the docs, Flow 8.3 also supports PHP 8.3.
This is just a backport of these changes.

Issue #3212, opened 2023-11-01
BUG: Dto with Dto in constructor creates proxy


mhsdesign:
Given the following code, i would expect no proxy class to be generated for either `Dto` or `OtherDto`

```php
// no proxy class?
final readonly class Dto
{
     private function __construct(public string $bar) {}

     public static function fromString(string $bar) { return new self($bar); }
}

// no proxy class? -> or maybe yes because of a non straight value?
final readonly class OtherDto
{
     private function __construct(public Dto $dto) {}

     public static function fromString(Dto $dto) { return new self($dto); }
}
```

For `Dto` this is true, but there will be a proxy class for `OtherDto`, which causes problems like not being able to use named arguments: https://github.com/neos/flow-development-collection/issues/3076

```php
<?php 

declare(strict_types=1);

namespace Neos\Neos;

readonly class OtherDto_Original
{
    public function __construct(public Dto $dto) {}

    public static function fromString(Dto $dto) { return new static($dto); }
}

#
# Start of Flow generated Proxy code
#

final readonly class OtherDto extends OtherDto_Original implements \Neos\Flow\ObjectManagement\Proxy\ProxyInterface {


    /**
     * Autogenerated Proxy Method
     */
    public function __construct()
    {
        $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2);
        if (isset($backtrace[1]) &&
            !is_subclass_of($backtrace[1]['class'], \Neos\Neos\OtherDto::class) &&
            !is_subclass_of(\Neos\Neos\OtherDto::class, $backtrace[1]['class']) &&
            $backtrace[1]['class'] !== 'Neos\Neos\OtherDto' &&
            $backtrace[1]['class'] !== 'Neos\Neos\OtherDto_Original'
        ) {
            throw new \Error('Call to private Neos\Neos\OtherDto::__construct() from invalid context', 1686153840);
        }
        $arguments = func_get_args();


        if (!array_key_exists(0, $arguments)) $arguments[0] = \Neos\Flow\Core\Bootstrap::$staticObjectManager->get('Neos\Neos\Dto');
[...]

Issue #3076, opened 2023-06-14
BUG: named arguments with proxied constructor wont work (fx when passing objects as argument)


mhsdesign:
see https://github.com/neos/flow-development-collection/pull/3082/commits/2f7003f34346a8f6b7131acdaed7f9ce818a1386

for a failing test to prove ;)

```php
class PrototypeClassJ
 {
     public function __construct(
         public ValueObjectClassA $classA
     ) {
     }
 }
```

you cant currently instantiate this like

```php
new PrototypeClassJ(classA: $classA);
```

as the constructor of the proxy class will have no arguments.

a proxy constructor looks like:

```php
#
# Start of Flow generated Proxy code
#
/**
 * A class of scope prototype with a value object as constructor argument
 * @codeCoverageIgnore
 */
class PrototypeClassJ extends PrototypeClassJ_Original implements \Neos\Flow\ObjectManagement\Proxy\ProxyInterface {


    /**
     * Autogenerated Proxy Method
     */
    public function __construct()
    {
        $arguments = func_get_args();


        if (!array_key_exists(0, $arguments)) $arguments[0] = \Neos\Flow\Core\Bootstrap::$staticObjectManager->get('Neos\Flow\Tests\Functional\ObjectManagement\Fixtures\ValueObjectClassA');
        if (!array_key_exists(0, $arguments)) throw new \Neos\Flow\ObjectManagement\Exception\UnresolvedDependenciesException('Missing required constructor argument $classA in class ' . __CLASS__ . '. Note that constructor injection is only support for objects of scope singleton (and this is not a singleton) – for other scopes you must pass each required argument to the constructor yourself.', 1296143788);
        parent::__construct(...$arguments);
    }
}
```

the proxy constructor should have all arguments, but nullable

```php
     public function __construct(
         ?ValueObjectClassA $classA = null
     ) {
     }
```

mhsdesign:
Implementing this might make https://github.com/neos/flow-development-collection/issues/2966 easier ^^

mhsdesign:
This would also allow the `Symfony\Component\Serializer\Normalizer\ObjectNormalizer` to work with proxy classes see https://github.com/neos/neos-development-collection/blob/0976d919a6268f58e05f827793cd78fecd12c1ea/Neos.ContentRepositoryRegistry/Classes/Infrastructure/Property/Normalizer/ProxyAwareObjectNormalizer.php#L12

PR #3494, opened 2025-09-09
BUGFIX: ProxyBuilding minor refactor and fixes

### LINKED ISSUE #3493: BUG: Incorrectly missing __sleep for proxy classes with entity properties and/or advices in some cases

### Current Behavior

A simple class like this:

```
class ExampleEvent {

    public function __construct(public SomeDoctrineEntity $doctrineEntity) {}

}
```

Should strip the doctrine entity (and possible AOP advices on `serialize($exampleEventInstance);` but it doesn't which leads to serialize errors due to closures used in advices and sometimes entities.

This worked in < 9.0 and was broken with the reworked proxy building.

`\Neos\Flow\ObjectManagement\DependencyInjection\ProxyClassBuilder::buildSerializeRelatedEntitiesCode` now skips the `Flow_serializeRelatedEntities` under some conditions (see early returns) which wasn't the case previously. Unfortunately the early returns do not check for entity properties nor for advices (both might be difficult to know at proxy building time). 
I think this was done to reduce the amount of proxies we create as we basically generated a __sleep for every class before, however we either need a way to bypass this skip or tighten our conditions to make this waterproof as the resulting error is hard to debug and understand.

A workaround is to add a useless transient property to the respective class ala:

```

class ExampleEvent {

    /**
     * Do not remove, this is a workaround for Flows proxy building
     * @Flow\Transient  	
     * @var null 
     */
    protected null $flowProxyFix = null;

    public function __construct(public SomeDoctrineEntity $doctrineEntity) {}

}
```

Which then triggers the proxy building of the necessary __sleep method.

### Expected Behavior

Proxy gets __sleep method to extract unwanted doctrine proxy before serialization.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Flow: 9.0
- PHP: xx
```

### Anything else?

_No response_

kitsunet:
Props to @lorenzulrich for this correct fix, which always gets the
parent in relation to the method being called and not the parent of
the instance at hand.

Fixes: https://github.com/neos/flow-development-collection/pull/3406

This is an overhaul of how object serialization
is prepared in proxies.

Proxies can skip object serialization code if there is nothing to
serialize, that is, if there are no entity properties, no injected, or
transient properties. We were too eager prior to this patch with
not using the serialization code, the checks are now way more detailed.
Additionally the "Proxy" Annotation now allows to force serialization
code for a class if the checks still fail to detect correctly, this
should be rarely needed.
This fix however broke some code in Neos that should have gotten the
serialization code previously but didn't. Since the class in question
is readonly, injecting a mutable property via trait resulted in
PHP errors.
Therefore we now use a mutable object to hold related entities for
serialization purposes which is declared readonly in the proxy to
avoid errors with readonly classes should they need serialization
code. Other mutable properties were removed as they are not strictly
needed. We should do the same refactoring for AOP as well.
Proxies can use the original constructor argument signature if no
constructor injection is used.
Finally prototype autowiring is now a choice via setting, currently
default enabled to not change behavior, in the future we should plan
a breaking change to disable it and then remove the option altogether.

Fixes: https://github.com/neos/flow-development-collection/issues/3493
Related: https://github.com/neos/flow-development-collection/issues/3212
Related: https://github.com/neos/flow-development-collection/issues/3076


kitsunet:
Interesting test failures... because that did work fine... Investigating...

lorenzulrich:
Thanks @kitsunet. This is "above my pay grade", so thanks for taking care and finding the reason why my code works ;-).

kitsunet:
Ok funny, now this PR shows one of the errors I wanted to fix in the functional tests. Why? Well because I broke something (that is if an object already has a sleep method we do not proxy anything) but intererstingly this scenario is AFAIK not happening in Flow nor Neos, nor have I seen it in any of the projects I checked. Anyways, will fix, seems like a good idea.

robertlemke:
Some documentation for the new flag - I don't know where to put it, maybe into this PR's description? Or the manual?

When Flow generates proxy classes, it automatically detects if your class contains entity properties
(properties typed with classes annotated as ``@Flow\Entity``) or other framework-managed objects
that require special handling during serialization. In such cases, Flow automatically generates
``__sleep()`` and ``__wakeup()`` methods that:

* Convert entity references to metadata (class name and persistence identifier) before serialization
* Remove injected and transient properties
* Restore entity references after deserialization

This detection is automatic and works in most cases. However, in rare edge cases where the automatic
detection fails (e.g., with complex generic types, deeply nested entity structures, or unusual type
declarations), you can force the generation of serialization code using the ``@Flow\Proxy`` annotation::

```php

	/**
	 * @Flow\Proxy(forceSerializationCode=true)
	 */
	class ComplexObjectWithEntities {

		/**
		 * @var ComplexGenericType<SomeEntity>
		 */
		protected $complexProperty;

	}
```

You should rarely need to use ``forceSerializationCode``. If you find yourself needing it
for standard entity properties or injected dependencies, this indicates a bug in Flow's
automatic detection that should be reported.

robertlemke (approved):
Makes sense. I tested this locally based on the functional tests and verified the difference of the generated proxy classes. Fine to merge, I guess?

Neos.Flow/Classes/ObjectManagement/Proxy/RelatedEntitiesContainer.php
```diff
@@ -0,0 +1,52 @@
...
+use Neos\Flow\Persistence\PersistenceManagerInterface;
+use Neos\Utility\ObjectAccess;
+
+/**
+ * This is a mutable container to hold references to entities (class & identifier) for serialization.
+ * Userland code should never (have to) interact with this, it is used in proxy classes only. You might
+ * see references to it in serialized object strings.
+ * @internal
+ */
+#[Flow\Proxy(false)]
```
kdambekalns, line 16: Is this _needed_? Or an optimization / safeguard?

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -113,36 +105,36 @@ private function Flow_serializeRelatedEntities(array $transientProperties, array
      * @param string $path
      * @param mixed $propertyValue
      * @param string $originalPropertyName
-     * @return void
+     * @return bool if an entity was found
      */
-    private function Flow_searchForEntitiesAndStoreIdentifierArray(string $path, mixed $propertyValue, string $originalPropertyName): void
+    private function Flow_searchForEntitiesAndStoreIdentifierArray(string $path, mixed $propertyValue, string $originalPropertyName): bool
     {
+        $foundEntity = false;
```
kdambekalns:
```suggestion
        $entityWasFound = false;
```
like in the code above – and ` $foundEntity` sounds as if it contains the entity that was found.

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -113,36 +105,36 @@ private function Flow_serializeRelatedEntities(array $transientProperties, array
...
         if (is_array($propertyValue) || ($propertyValue instanceof \ArrayObject || $propertyValue instanceof \SplObjectStorage)) {
             foreach ($propertyValue as $key => $value) {
-                $this->Flow_searchForEntitiesAndStoreIdentifierArray($path . '.' . $key, $value, $originalPropertyName);
+                $foundEntity = $foundEntity || $this->Flow_searchForEntitiesAndStoreIdentifierArray($path . '.' . $key, $value, $originalPropertyName);
```
kdambekalns:
```suggestion
                $entityWasFound = $entityWasFound || $this->Flow_searchForEntitiesAndStoreIdentifierArray($path . '.' . $key, $value, $originalPropertyName);
```

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -113,36 +105,36 @@ private function Flow_serializeRelatedEntities(array $transientProperties, array
...
-                $identifier = current(ObjectAccess::getProperty($propertyValue, '_identifier', true));
-            }
-            $this->Flow_Persistence_RelatedEntities[$originalPropertyName . '.' . $path] = [
-                'propertyName' => $originalPropertyName,
-                'entityType' => $className,
-                'identifier' => $identifier,
-                'entityPath' => $path
-            ];
-            $this->$originalPropertyName = Arrays::setValueByPath($this->$originalPropertyName, $path, null);
+            $foundEntity = true;
```
kdambekalns:
```suggestion
            $entityWasFound = true;
```

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
```diff
@@ -113,36 +105,36 @@ private function Flow_serializeRelatedEntities(array $transientProperties, array
...
         }
+
+        return $foundEntity;
```
kdambekalns:
```suggestion
        return $entityWasFound;
```

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -739,4 +763,68 @@ protected function compileStaticMethods(string $className, ProxyClass $proxyClas
             $compiledMethod->setBody('return ' . $compiledResult . ';');
         }
     }
+
+    protected function couldHaveEntityRelations(Configuration $objectConfiguration): bool
+    {
+        $result = false;
+        /** @var class-string $className */
```
kdambekalns: Is `class-string` on a plain `@var` something useful?

Neos.Flow/Configuration/Settings.Object.yaml
```diff
@@ -51,3 +51,13 @@ Neos:
...
+
+        # This is a forward compatibility setting, preventing constructor arguments to be autowired
+        # if they are not declared singleton or session when this setting is false.
+        #
+        # BEWARE of dragons, Flow code should be fine, but Neos might still contain such constructs.
+        # Manual full cache flush is necessary after changing this setting.
+        #
+        # In the future autowiring of prototypes should no longer be an option as it makes no sense.
+        # Use a factory class if you need this otherwise.
+        prototypeAutowiring: true
```
kdambekalns: The name is not as clear to me, as it should be. Does this prevent the autowiring of constructor arguments _to_ prototypes or does it prevent the use of prototypes _as_ constructor arguments?

Neos.Flow/Classes/ObjectManagement/Proxy/ObjectSerializationTrait.php
mhsdesign: fixed via 4da664557d9f9e96619a965b86e6351a5ded00ca

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
mhsdesign: fixed via 4da664557d9f9e96619a965b86e6351a5ded00ca

Neos.Flow/Configuration/Settings.Object.yaml
mhsdesign: fixed via https://github.com/neos/flow-development-collection/pull/3494/commits/7b363e2a3d71df17165cd68a5bcc80c8a6175717

mhsdesign:
Okay i found a bug last minute

### Regression:

During testing i had Neos Debug installed and the proxies generated failed at php compile time:

> Fatal error: Type of Flowpack\Neos\Debug\DataCollector\MessagesCollector::$dataFormatter must be ?Flowpack\Neos\Debug\DataFormatter\DataFormatterInterface (as in class Flowpack\Neos\Debug\DataCollector\AbstractDataCollector) in Data/Temporary/Development/Cache/Code/Flow_Object_Classes/Flowpack_Neos_Debug_DataCollector_MessagesCollector.php on line 40 

The reason for this is the constructor in the abstract class: https://github.com/Flowpack/neos-debug/blob/9c82cb345cb1b786b88f07dee73824882479a21d/Classes/DataCollector/AbstractDataCollector.php#L12

```php
public function __construct(
    protected ?DataFormatterInterface $dataFormatter = null,
) {
}
```

Now the object extending the `AbstractDataCollector` got in 9.x a proxy like

```php
public function __construct()
{
    $arguments = func_get_args();
    if (get_class($this) === 'Flowpack\Neos\Debug\DataCollector\MessagesCollector') \Neos\Flow\Core\Bootstrap::$staticObjectManager->setInstance('Flowpack\Neos\Debug\DataCollector\MessagesCollector', $this);
}
```

But with this change (`prototypeAutowiring` untouched!) the flow proxy gets generated like this by default:

```php
public function __construct(protected \Flowpack\Neos\Debug\DataFormatter\DataFormatterInterface $dataFormatter)
{
    $arguments = func_get_args();
    if (get_class($this) === 'Flowpack\Neos\Debug\DataCollector\MessagesCollector') \Neos\Flow\Core\Bootstrap::$staticObjectManager->setInstance('Flowpack\Neos\Debug\DataCollector\MessagesCollector', $this);
}
```

Where the default value and the nullability is missing.


EDIT: Found a possibly hack fix https://github.com/neos/flow-development-collection/pull/3494/commits/c7e76f9f07fe2d05369f69d2602d9ef0e0c5e111 but it seems we might need https://github.com/laminas/laminas-code/issues/182 in reality?

mhsdesign:
and also another oddity during the depreciation logging - there are much more entries than i would have anticipated in which the user of Neos and Flow cant do anything. Even Flow itself seems to raise the deprecation for its own code. I fear that we cannot distinguish between _needing_ the auto-wiring and that Flow just gave that gratis until now. We should probably specify autowiring off for the respective constructors.

The list shows the prototype first entry where it is used in the constructor ... nested entries.

- `Neos\ContentRepositoryRegistry\Migration\Factory\MigrationFactory` (prototype):
  - `Neos\ContentRepositoryRegistry\Command\NodeMigrationCommandController`

- `Neos\ContentRepositoryRegistry\Service\NodeMigrationGeneratorService` (prototype):
  - `Neos\ContentRepositoryRegistry\Command\NodeMigrationCommandController`

- `Neos\ContentRepository\LegacyNodeMigration\RootNodeTypeMapping` (prototype):
  - `Neos\ContentRepository\LegacyNodeMigration\LegacyExportService`
  - `Neos\ContentRepository\LegacyNodeMigration\LegacyExportServiceFactory`
  - `Neos\ContentRepository\LegacyNodeMigration\Processors\EventExportProcessor`

- `Neos\Flow\Aop\JoinPoint` (prototype):
  - `Neos\Flow\Security\Authorization\Privilege\Method\MethodPrivilegeSubject`

- `Neos\Flow\I18n\Locale` (prototype):
  - `Neos\Flow\I18n\Xliff\Model\FileAdapter`

- `Neos\Flow\Mvc\ActionRequest` (prototype):
  - `Neos\Flow\Mvc\Controller\ControllerContext`
  - `Neos\Form\Core\Runtime\FormRuntime`
  - `Neos\Fusion\Core\LegacyFusionControllerContext`

- `Neos\Flow\Mvc\ActionResponse` (prototype):
  - `Neos\Flow\Mvc\Controller\ControllerContext`
  - `Neos\Form\Core\Runtime\FormRuntime`
  - `Neos\Fusion\Core\LegacyFusionControllerContext`

- `Neos\Flow\Mvc\Controller\Arguments` (prototype):
  - `Neos\Flow\Mvc\Controller\ControllerContext`

- `Neos\Flow\Mvc\Routing\UriBuilder` (prototype):
  - `Neos\Flow\Mvc\Controller\ControllerContext`

- `Neos\Flow\Persistence\Doctrine\Query` (prototype):
  - `Neos\Flow\Persistence\Doctrine\QueryResult`
  - `Neos\Flow\Persistence\EmptyQueryResult`
  - `Neos\Media\Domain\Model\AssetSource\Neos\NeosAssetProxyQuery`

- `Neos\Flow\ResourceManagement\PersistentResource` (prototype):
  - `Neos\Media\Domain\Model\Asset`
  - `Neos\Media\Domain\Model\Audio`
  - `Neos\Media\Domain\Model\Document`
  - `Neos\Media\Domain\Model\Image`
  - `Neos\Media\Domain\Model\Video`

- `Neos\Flow\Security\Authorization\Privilege\PrivilegeTarget` (prototype):
  - `Neos\Media\Security\Authorization\Privilege\ReadAssetCollectionPrivilege`
  - `Neos\Media\Security\Authorization\Privilege\ReadAssetPrivilege`
  - `Neos\Media\Security\Authorization\Privilege\ReadTagPrivilege`
  - `Neos\Neos\Security\Authorization\Privilege\EditNodePrivilege`
  - `Neos\Neos\Security\Authorization\Privilege\ModulePrivilege`
  - `Neos\Neos\Security\Authorization\Privilege\ReadNodePrivilege`

- `Neos\Form\Core\Model\FormDefinition` (prototype):
  - `Neos\Form\Core\Runtime\FormRuntime`

- `Neos\Form\Core\Runtime\FormRuntime` (prototype):
  - `Neos\Form\Core\Model\FinisherContext`

- `Neos\Fusion\Core\Cache\FusionContextSerializer` (prototype):
  - `Neos\Neos\Fusion\Cache\NeosFusionContextSerializer`

- `Neos\Fusion\Core\FusionConfiguration` (prototype):
  - `Neos\Fusion\Core\Runtime`

- `Neos\Fusion\Core\FusionGlobals` (prototype):
  - `Neos\Fusion\Core\Runtime`

- `Neos\Fusion\Core\ObjectTreeParser\Lexer` (prototype):
  - `Neos\Fusion\Core\ObjectTreeParser\ObjectTreeParser`

- `Neos\Fusion\Core\ObjectTreeParser\MergedArrayTree` (prototype):
  - `Neos\Fusion\Core\ObjectTreeParser\MergedArrayTreeVisitor`

- `Neos\Fusion\Core\Runtime` (prototype):
  - `Neos\Fusion\Core\Cache\RuntimeContentCache`
  - all objects of type `AbstractFusionObject` because of the shared constructor

- `Neos\Fusion\FusionObjects\AbstractFusionObject` (prototype):
  - `Neos\Fusion\FusionObjects\Helpers\FluidView`

- `Neos\Fusion\FusionObjects\TemplateImplementation` (prototype):
  - `Neos\Fusion\FusionObjects\Helpers\FusionPathProxy`

- `Neos\Http\Factories\UriFactory` (prototype):
  - `Neos\Http\Factories\PsrHttpFactory`
  - `Neos\Http\Factories\ServerRequestFactory`

- `Neos\Media\Domain\Model\AssetSource\Neos\NeosAssetSource` (prototype):
  - `Neos\Media\Domain\Model\AssetSource\Neos\NeosAssetProxy`
  - `Neos\Media\Domain\Model\AssetSource\Neos\NeosAssetProxyQuery`
  - `Neos\Media\Domain\Model\AssetSource\Neos\NeosAssetProxyRepository`

- `Neos\Media\Domain\Model\Image` (prototype):
  - `Neos\Media\Browser\Domain\ImageMapper`
  - `Neos\Media\Domain\Model\ImageVariant`

- `Neos\Media\Domain\Model\ThumbnailConfiguration` (prototype):
  - `Neos\Media\Domain\Model\Thumbnail`

- `Neos\Media\Domain\ValueObject\Configuration\Label` (prototype):
  - `Neos\Media\Domain\ValueObject\Configuration\VariantPreset`

- `Neos\Neos\AssetUsage\AssetUsageIndexingProcessor` (prototype):
  - `Neos\Neos\AssetUsage\Command\AssetUsageCommandController`

- `Neos\Neos\Domain\SubtreeTagging\SoftRemoval\SoftRemovalGarbageCollector` (prototype):
  - `Neos\Neos\Domain\Service\WorkspacePublishingService`
  - `Neos\Neos\Domain\Service\WorkspaceService`

- `Neos\Neos\FrontendRouting\DimensionResolution\Resolver\UriPathResolver\Segments` (prototype):
  - `Neos\Neos\FrontendRouting\DimensionResolution\Resolver\UriPathResolver`

- `Neos\Neos\FrontendRouting\Projection\DocumentUriPathFinder` (prototype):
  - `Neos\Neos\FrontendRouting\CatchUpHook\RouterCacheHook`
  - `Neos\RedirectHandler\NeosAdapter\CatchUpHook\DocumentUriPathProjectionHook`

- `Neos\Neos\Setup\Infrastructure\ImageHandler\ImageHandlerService` (prototype):
  - `Neos\Neos\Setup\Infrastructure\Healthcheck\ImageHandlerHealthcheck`

- `Neos\Setup\Domain\HealthcheckEnvironment` (prototype):
  - `Neos\Setup\Infrastructure\HealthChecker`

- `Neos\TimeableNodeVisibility\Domain\ChangedVisibilityType` (prototype):
  - `Neos\TimeableNodeVisibility\Domain\ChangedVisibility`

- `TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler` (prototype):
  - `TYPO3Fluid\Fluid\Core\Compiler\NodeConverter`

- `TYPO3Fluid\Fluid\Core\Parser\ParsingState` (prototype):
  - `TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\ViewHelperNode`

kitsunet:
Mmm, technically I think the constructor change doesn't have to live in here, I guess I can extract it and try  to  put it in the runtime constructor injection change, probably more thematic there anyways. Then this  becomes a bit simpler.

kitsunet:
Superseeded by constructor injection change now in 9.2

Issue #3499, opened 2025-10-10
FEATURE: Use native PHP LazyProxy


mficzel:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

PHP Lazy proxies can avoid the currently needed checks for `instanceof DependencyProxy` as they are transparent to the code.

@see https://www.php.net/manual/en/language.oop5.lazy-objects.php

mficzel:
This is less trivial as it looks as we do things that are not supported by PHP lazyProxies yet.

LazyObjects have to be of a concrete class. We inject ClassNames or Interfaces and let the Object Manager decide what we will actually get. Which may be a subclass or an implementation of the interface.

PR #3500, opened 2025-10-10
!!! WIP !!! TASK: Use native LazyProxy Objects instead of DependencyProxy

mficzel:
resolves: #3499

**Upgrade instructions**

**Review instructions**

Not sure how we consider the breakiness. However this requires PHP 8.4 and thus surely needs Flow 10

**Checklist**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.2', '8.3', '8.4']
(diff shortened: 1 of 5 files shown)


mficzel:
This is sadly not feasible as the native php lazy loading does not support cases where the implementation class is decided during hydration. We rely on then when we inject interfaces and virtual objects. Since interfaces are a major part of flow injections it would only be possible to optimize a fraction of the lazy objects for the cost of maintaining two separate pathes.

PR #3504, opened 2025-10-20
BUGFIX: Support PHP 8+ named arguments in proxy constructors

### LINKED ISSUE #3076: BUG: named arguments with proxied constructor wont work (fx when passing objects as argument)

robertlemke:
This change enables Flow proxy classes to support PHP 8+ named argument syntax by including the original constructor parameter signature in generated proxies, fixing compatibility issues with modern PHP code and reflection-based serializers.

The proxy constructor now always includes parameters matching the original constructor signature, with all parameters made nullable and given null default values to maintain dependency injection flexibility. This allows both named and positional argument syntax while preserving backward compatibility with existing DI patterns.

Resolves #3076


robertlemke:
Note: This PR is currently based on https://github.com/neos/flow-development-collection/pull/3494 - so merge that before this one.

kitsunet (changes requested):
Just leaving this here so no one decides to spontanously merge this. 
Neos is broken with this change applied, I will add some more test cases showing problems with this approach tomorrow.

kitsunet (approved):
From my POV this should be a good improvement to the status quo now. Lets see what the tests say. I also ran Neos tests locally.

robertlemke:
As far as I can tell, this is now already solved with #3076 in 9.2, so we can close this – or?

kdambekalns:
> As far as I can tell, this is now already solved with #3076 in 9.2, so we can close this – or?

In fact 8aed324 looks like the same change – minus a ton of improvements to comments and some other tweaks. I guess you'd need to check what is missing from what you intended to do.

robertlemke:
> > As far as I can tell, this is now already solved with #3076 in 9.2, so we can close this – or?
> 
> In fact [8aed324](https://github.com/neos/flow-development-collection/commit/8aed3247e7e38a1ec136ad1adf0b8e4b6df8d39f) looks like the same change – minus a ton of improvements to comments and some other tweaks. I guess you'd need to check what is missing from what you intended to do.

Good catch. I checked this: all 9 commits of this PR are ancestors of the current 9.2 branch. They came in through #3510, which was based on this branch. So nothing of this change is missing in 9.2. I ran `git merge-base --is-ancestor` and that confirms every single commit, including 8aed324 and the follow-up tweaks.
                                                                                                                                                                 
What never happened is the backport to 9.0 / 9.1 this PR was originally targeting. But since the constructor code generation evolved further in 9.2 with #3510, backporting this old state would create a diverging implementation which we'd have to maintain separately. That's not worth it, so I'm closing this PR. The fix is available as of Flow 9.2.

Fine with you?

PR #3505, opened 2025-10-21
BUGFIX: Support never return types in AOP proxies

### LINKED ISSUE #3451: BUG: Using `never` as return type breaks built proxies

### Current Behavior

Imagine an action like this:

```php
public function testAction(): never
{
    // some code here
    exit();
}
```

The proxy method built for this does return, leading to an error:

`PHP Fatal error:  A never-returning function must not return in /…/Data/Temporary/Production/SubContextLive/Cache/Code/Flow_Object_Classes/Acme_Foo_SomeController.php on line 123`

The code looks like this (depending on advices being present):

```php
public function testAction() : never
    {
 
        if (isset($this->Flow_Aop_Proxy_methodIsInAdviceMode['testAction'])) {
            $result = parent::testAction();
 
        } else {
            $this->Flow_Aop_Proxy_methodIsInAdviceMode['testAction'] = true;
            try {
 
                $methodArguments = [];
 
                $adviceChains = $this->Flow_Aop_Proxy_getAdviceChains('testAction');
                $adviceChain = $adviceChains['Neos\Flow\Aop\Advice\AroundAdvice'];
                $adviceChain->rewind();
                $joinPoint = new \Neos\Flow\Aop\JoinPoint($this, 'Acme\Foo\Controller\SomeController', 'testAction', $methodArguments, $adviceChain);
                $result = $adviceChain->proceed($joinPoint);
                $methodArguments = $joinPoint->getMethodArguments();
 
                unset($this->Flow_Aop_Proxy_methodIsInAdviceMode['testAction']);
                throw $exception;
            }
            unset($this->Flow_Aop_Proxy_methodIsInAdviceMode['testAction']);
        }
        return $result;
    }
``` 

### Expected Behavior

The action can be called just fine, the proxy method is built correctly. That means, it should not return, as `never` must not return:

> [never](https://www.php.net/manual/en/language.types.never.php) is a return-only type indicating the function does not terminate. This means that it either calls [exit()](https://www.php.net/manual/en/function.exit.php), throws an exception, or is an infinite loop.

Any AOP "after" this must not be built (useless), "around" could skip anything handling the return value (useless), assigning `$result` is useless and the proxy itself must not return at all. Which might render any AOP useless (unless the proxy changes the return type, which is impossible unless the original is changed, too, which would be a very bad idea).

### Steps To Reproduce

_No response_

### Environment

```markdown
- Flow: 8.3
- PHP: 8.3
```

### Anything else?

Workaround for now: Do not use `never` in your code.

robertlemke:
This adjusts proxy class building for AOP proxies so that the generate code correctly advises methods which have a "never" return type.

Resolves #3451

### DIFF
--- a/Neos.Flow/Classes/Aop/Builder/AbstractMethodInterceptorBuilder.php
+++ b/Neos.Flow/Classes/Aop/Builder/AbstractMethodInterceptorBuilder.php
@@ -156,7 +156,7 @@ abstract class AbstractMethodInterceptorBuilder
      * @param string|null $declaringClassName Name of the declaring class. This is usually the same as the $targetClassName. However, it is the introduction interface for introduced methods.
      * @return string PHP code to be used in the method interceptor
      */
-    protected function buildAdvicesCode(array $groupedAdvices, ?string $methodName = null, ?string $targetClassName = null, ?string $declaringClassName = null): string
+    protected function buildAdvicesCode(array $groupedAdvices, ?string $methodName, ?string $targetClassName, ?string $declaringClassName, ?string $declaredReturnType): string
     {
         $advicesCode = $this->buildMethodArgumentsArrayCode($declaringClassName, $methodName, ($methodName === '__construct'));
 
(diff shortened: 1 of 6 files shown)


Neos.Flow/Classes/Aop/Builder/AbstractMethodInterceptorBuilder.php
```diff
@@ -184,9 +184,17 @@ protected function buildAdvicesCode(array $groupedAdvices, ?string $methodName =
                 $adviceChain = $adviceChains[\'Neos\Flow\Aop\Advice\AroundAdvice\'];
                 $adviceChain->rewind();
                 $joinPoint = new \Neos\Flow\Aop\JoinPoint($this, \'' . $targetClassName . '\', \'' . $methodName . '\', $methodArguments, $adviceChain);
+';
+            if ($declaredReturnType === 'never') {
+                $advicesCode .= '
+                $adviceChain->proceed($joinPoint);
```
kdambekalns, line 190: From reading this I wonder, why `$methodArguments = $joinPoint->getMethodArguments();` is dropped, too…

robertlemke: That's a mistake, indeed.

kdambekalns (approved):

PR #3508, opened 2025-10-21
BUGFIX: Support ORM attributes

### LINKED ISSUE #2602: PHP8 Attributes ignored for entities
### Description
I tried to experimentally make a packet PHP8 only, using only the new attributes without PHPDoc annotations.

But it looks like that is not yet supported, or is not implemented in some places.

As example:

[Neos\Flow\Persistence\Doctrine\Mapping\Driver\FlowAnnotationDriver::loadMetadataForClass](https://github.com/neos/flow-development-collection/blob/a4c625d558d6acabb9ed58a3fdb7559ab0bbda74/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php#L173)

Its only us the PHPDoc annotations.

### Steps to Reproduce

Create an Entity only with Attributes
call ./flow doctrine:migrationgenerate

#### Expected behavior

Attributes should work, also without PHPDoc annotations.

#### Actual behavior

Class "Vendor\Packet\Domain\Model\Entity" sub class of "Vendor\Packet\Domain\Model\Entity_Original" is not a valid entity or mapped super class.

### Affected Versions

Flow: 7.2.1

kdambekalns:
Hopefully fixes #2602 and allows to use PHP attributes to configure Doctrine ORM, not only annotations like before.

**Upgrade instructions**

No change should be needed.

**Checklist**

- [ ] ~~Reviewer - Breaking Changes are marked with `!!!` and have upgrade-instructions~~

### DIFF
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1572,6 +1572,8 @@ class ReflectionService
         if (count($varTagValues) > 1) {
             throw new InvalidPropertyTypeException('More than one @var annotation given for "' . $className . '::$' . $propertyName . '"', 1367334366);
         }
+        // We need var tags for typed arrays or collections as those cannot be
+        // expressed natively so they have to take precedence
         if (count($varTagValues) === 0) {
             $declaredType = $this->getPropertyType($className, $propertyName);
         } else {
--- a/Neos.Flow/Classes/Reflection/ReflectionService.php
+++ b/Neos.Flow/Classes/Reflection/ReflectionService.php
@@ -1568,16 +1568,19 @@ class ReflectionService
             return false;
         }
 
-        if (!$this->isPropertyTaggedWith($className, $propertyName, 'var')) {
-            return false;
-        }
-
         $varTagValues = $this->getPropertyTagValues($className, $propertyName, 'var');
         if (count($varTagValues) > 1) {
             throw new InvalidPropertyTypeException('More than one @var annotation given for "' . $className . '::$' . $propertyName . '"', 1367334366);
         }
+        if (count($varTagValues) === 0) {
+            $declaredType = $this->getPropertyType($className, $propertyName);
+        } else {
+            $declaredType = strtok(trim(current($varTagValues), " \n\t"), " \n\t");
+        }
 
-        $declaredType = strtok(trim(current($varTagValues), " \n\t"), " \n\t");
+        if (!$declaredType) {
+            return false;
+        }
 
         if ($this->isPropertyAnnotatedWith($className, $propertyName, ORM\Id::class)) {
             $skipArtificialIdentity = true;
--- a/Neos.Flow/Tests/Functional/Persistence/FixturesPHP8/Post.php
+++ b/Neos.Flow/Tests/Functional/Persistence/FixturesPHP8/Post.php
@@ -47,13 +47,10 @@ class Post
     protected Collection $tags;
 
     /**
-     * Still annotated because nested Attributes require PHP 8.1
-     * @todo Adjust to attributes for Flow 8.4 when the min PHP version will be 8.1
-     *
      * @var Collection<Post>
-     * @ORM\ManyToMany
-     * @ORM\JoinTable(inverseJoinColumns={@ORM\JoinColumn(name="related_post_id")})
      */
+    #[ORM\ManyToMany]
+    #[ORM\JoinTable(inverseJoinColumns: [new ORM\JoinColumn(name: 'related_post_id')])]
     protected Collection $related;
 
     /**
--- a/Neos.Flow/Tests/Functional/Persistence/FixturesPHP8/Post.php
+++ b/Neos.Flow/Tests/Functional/Persistence/FixturesPHP8/Post.php
@@ -18,58 +18,49 @@ use Neos\Flow\Annotations as Flow;
 
 /**
  * A sample entity for tests
- *
- * @Flow\Entity
- * @ORM\HasLifecycleCallbacks
- * @ORM\InheritanceType("JOINED")
  */
+#[Flow\Entity]
+#[ORM\HasLifecycleCallbacks]
+#[ORM\InheritanceType('JOINED')]
 class Post
 {
-    /**
-     * @var string
-     */
-    protected $title = '';
+    protected string $title = '';
 
-    /**
-     * @var Image
-     * @ORM\OneToOne
-     */
-    protected $image;
+    #[ORM\OneToOne]
+    protected Image $image;
 
-    /**
-     * @var Image
-     * @ORM\OneToOne
-     */
-    protected $thumbnail;
+    #[ORM\OneToOne]
+    protected Image $thumbnail;
 
     /**
      * Yeah, only one comment allowed for a post ;-)
      * But that's the easiest option for our functional test.
-     *
-     * @var Comment
-     * @ORM\OneToOne
-     * @ORM\JoinColumn(onDelete="SET NULL")
      */
-    protected $comment;
+    #[ORM\OneToOne]
+    #[ORM\JoinColumn(onDelete: 'SET NULL')]
+    protected Comment $comment;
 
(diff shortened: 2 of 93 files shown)


kdambekalns:
**This does not yet work** – whether this "minimally invasive" approach is feasible is yet to be proven…

kdambekalns:
@mficzel Using the reflection service in our own reader implementation works. 

If our assumptions are correct, new tests should pass, too.

mficzel:
@kdambekalns something is still fishy ...
```
Doctrine\ORM\Mapping\MappingException: No identifier/primary key specified for Entity "Neos\Flow\Tests\Functional\Persistence\FixturesPHP8\AnnotatedIdEntity" sub class of "Neos\Flow\Tests\Functional\Persistence\FixturesPHP8\AnnotatedIdEntity_Original". Every Entity must have an identifier/primary key.

/var/www/html/Packages/Libraries/doctrine/orm/src/Mapping/MappingException.php:44
/var/www/html/Packages/Libraries/doctrine/orm/src/Mapping/ClassMetadataInfo.php:1194
/var/www/html/Packages/Libraries/doctrine/orm/src/Mapping/ClassMetadataFactory.php:284
/var/www/html/Packages/Libraries/doctrine/orm/src/Mapping/ClassMetadataFactory.php:264
/var/www/html/Packages/Libraries/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php:343
/var/www/html/Packages/Libraries/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php:207
/var/www/html/Packages/Libraries/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php:96
/var/www/html/Data/Temporary/Testing/Cache/Code/Flow_Object_Classes/Neos_Flow_Persistence_Doctrine_PersistenceManager.php:317
/var/www/html/Packages/Framework/Neos.Flow/Tests/FunctionalTestCase.php:159

```

mficzel:
I wonder wether we also have to adjust `\Neos\Flow\Persistence\Doctrine\Mapping\ClassMetadata->_initializeReflection`

kdambekalns:
Making `@Flow\Entity` an attribute in "real code" within the Flow classes allows a clean compile run. And `#[ORM\Id]` and `#[Flow\Identity]` do work, too. So, is only testing affected?

kdambekalns:
Doh… I removed all the adjusted entities except `AnnotatedIdEntity` to limit the scope. Turns out **this is broken**:
```php
#[Flow\Entity]
class AnnotatedIdEntity
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\SequenceGenerator(sequenceName: 'annotatedidentity_seq')]
    protected string $id;
```

But **this does work**:
```php
#[Flow\Entity]
class AnnotatedIdEntity
{
    /** @var string */
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\SequenceGenerator(sequenceName: 'annotatedidentity_seq')]
    protected string $id;
```

The missing `@var` breaks it! On to digging deeper…

mficzel:
We should think about targeting Neos 8.4 + PHP 8.2 as i noticed that in Neos 8.3 with PHP 8.0 we: 

- Cannot declare nested attributes in PHP as those are a PHP 8.1 feature
- Even with PHP 8.2 + Neos 8.3 Nested Attributes are destroyed by the proxy class builder. I hope this is better in 8.4

mficzel:
Checked Neos 8.4 and the proxy builder still destroys nested attributes. Created #3511 for that.

kdambekalns:
I am confused. Locally this PR passes on PHP 8.3 if I add this:
[var-annotations.patch](https://github.com/user-attachments/files/23096789/var-annotations.patch)

How does this relate to broken attribute rendering?

- In the sense it will never work on PHP 8.0 and 8.1, and the solution is to go for Flow 8.4, as that requires PHP 8.2? Fine with me!
- But that will not automagically fix the "missing type issue", or is that something that has changed in Flow 8.4 already?

mficzel:
I tried rebasing this pr to 8.4 and together with the changes from #3512 i could successfully define a repository and entity via attributes. 

However the properties in the entity were not detected unless i added the @var annotation. So this problem persists even in 8.4. 

However it looks like it might really be the last real problem as the proxy classes were successfully created otherwise and flow did not complain about anything.

kdambekalns:
> However it looks like it might really be the last real problem as the proxy classes were successfully created otherwise and flow did not complain about anything.

That's great news already!

kdambekalns:
🎉

Neos.Flow/Tests/Functional/Persistence/FixturesPHP8/Post.php
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Still annotated because nested Attributes require PHP 8.1
+     * @todo Adjust to attributes for Flow 8.4 when the min PHP version will be 8.1
+     *
+     * @var Collection<Post>
+     * @ORM\ManyToMany
+     * @ORM\JoinTable(inverseJoinColumns={@ORM\JoinColumn(name="related_post_id")})
+     */
+    protected Collection $related;
```
mhsdesign, line 54: shouldn't we now be able to use nested attributes here? From https://github.com/neos/flow-development-collection/pull/3512

kitsunet: I think this prevents it: https://github.com/neos/flow-development-collection/issues/3511

kdambekalns: But #5312 fixes #5311…

mficzel: @mhsdesign yes if #3512 is in we should be able to adjust this. Before this caused broken php code in the proxy classes.

kdambekalns: Yay, tests pass even with nested attributes. 🎉

Neos.Flow/Classes/Reflection/ReflectionService.php
```diff
@@ -1568,16 +1568,19 @@ protected function evaluateClassPropertyAnnotationsForSchema(ClassSchema $classS
...
 
-        if (!$this->isPropertyTaggedWith($className, $propertyName, 'var')) {
-            return false;
-        }
-
         $varTagValues = $this->getPropertyTagValues($className, $propertyName, 'var');
         if (count($varTagValues) > 1) {
             throw new InvalidPropertyTypeException('More than one @var annotation given for "' . $className . '::$' . $propertyName . '"', 1367334366);
         }
+        if (count($varTagValues) === 0) {
```
kitsunet, line 1577: Do we want to prefer the var tags over the declaredType if both exist? 
Spontanously would have expected something like this:

$declaredType = $this->getPropertyType($className, $propertyName);
if ($declaredType = null && count($varTagValues) === 1)

kitsunet (approved):

kitsunet:
I left one question but it's not a blocker for merge I guess...

Neos.Flow/Tests/Functional/Persistence/FixturesPHP8/Post.php
mficzel: Adjusted the tests ... still green

Neos.Flow/Classes/Reflection/ReflectionService.php
mficzel: We need var tags for typed arrays or collections as those cannot be expressed natively so i assume they will have to take precedence.

kdambekalns: Ex-act-ly what @mficzel said!

kdambekalns: Added a comment to clarify that for us and future generations…

PR #3510, opened 2025-10-22
FEATURE: Runtime Constructor Injection

kitsunet:
This changes constructor injection to happen at runtime by the ObjectManager. That means the constructor no longer needs to be proxied and loose it's signature (Any other proxying happening, eg. AOP or other injection types will retain the signature now). It therefore also reduces the  amount of proxies necessary in a project.  
Additionally to proof how this works some of the core classes now use exclusively constructor injection.
The change also includes refactorings for type safety around the object configuration and proxy building.

*NOTE*
This obviously breaks instantiating classes with constructor injection via `new MyClass()` omitting any arguments. IMHO that is very bad practice anyhow. As can be seen in this change (and I didn't see anything in Neos either), it was only used in tests as of now. I think therefore we can relatively safely change this in a minor release?


kitsunet:
Obviously broke the tests with my two class changes to make those unproxied. But then, I am not sure I want to keep those changes in there, it was just for show.

kitsunet:
@kdambekalns if you could look at this please. I would want to merge this for a 9.2 beta to have it tested on neos.io as we want tto test the CR layer change equally on neos.io before a release. Both with the option of revert

kdambekalns:
> @kdambekalns if you could look at this please.

Will do either later today or tomorrow…

mficzel (commented):
~~Note: I just found multiple instances of `new UriBuilder()` in actual project code. Not a blocker but this could be more breaky than expected.~~

Correction: Since the change would only affect situations where `new Classname` is called with missing arguments it should be non breaky. That should alert all QS tools anyways.

kitsunet:
@mficzel that is fine though. Attribute/injectMethod works fine, and you can new this. It's not pretty. The only thing that does not work anymore is constructor injection. So something like this:

```
class Foo {
  protected $objectManager;

  public function __construct(ObjectManagerInterface $objectManager) {
     $this->objectManager = $objectManager;
  }
}

$myFoo = new Foo();
```

mficzel (commented):
Looks fine but i wonder wether we really want to go all in with constructor injection as the way to  do autowiring?

+ Pro: each class makes sense on its own and is probably easier to test aswell no question wether those injections happened and no need to do initializing stuff in `initializeObject`  as all dependencies are in the constructor already
- Contra: there is no hint in the code that those arguments will be injected automagically so it is one more thing one has to know

I suggest to eventually add `#[Flow\Inject]` to constructor-methods when autowiring is required to make this visible.

Props: for cleaning up `classesBeeingInstantiated` in the `finally` ... learned a bit about php reading that.

Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
```diff
@@ -416,14 +416,17 @@ public function buildProxyClass(string $targetClassName, array $aspectContainers
         $callBuildMethodsAndAdvicesArrayCode = "\n        \$this->Flow_Aop_Proxy_buildMethodsAndAdvicesArray();\n";
         $proxyClass->getConstructor()->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
         $proxyClass->getMethod('__wakeup')->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode);
+        $proxyClass->getMethod('__wakeup');
```
kdambekalns: Might a comment help to keep this seemingly useless call? And explain it to me at the same time? 😉

(Same for `__clone` below.)

Neos.Flow/Classes/Cli/CommandManager.php
```diff
@@ -27,41 +27,19 @@
 class CommandManager
 {
     /**
-     * @var array<Command>
+     * @var array<Command>|null
      */
-    protected $availableCommands = null;
+    protected array|null $availableCommands = null;
```
kdambekalns: This would be cleaner – is it impossible?
```suggestion
    protected array $availableCommands = [];
```

Same for `$shortCommandIdentifiers` below.

Neos.Flow/Configuration/Objects.yaml
```diff
@@ -278,8 +278,8 @@ Neos\Flow\Mvc\Routing\RouterCachingService:
             value: Flow_Mvc_Routing_Resolve
 
 Neos\Flow\Mvc\ViewConfigurationManager:
-  properties:
-    cache:
+  arguments:
+    3:
```
kdambekalns, line 282: We should add support for named arguments here, `3` seems very brittle.

Neos.Flow/Configuration/Caches.yaml
kdambekalns, line 1: Why are these definitions dropped?

Neos.Flow/Classes/Annotations/Proxy.php
```diff
@@ -27,13 +35,64 @@
...
+    public bool $forceSerializationCode = false;
+
+    public function __construct(bool $enabled = true, bool $forceSerializationCode = false)
     {
+        if ($enabled === false && $forceSerializationCode === true) {
+            throw new ProxyCompilerException('Cannot disable a Proxy but forceSerializationCode at the same time.', 1756813222);
+        }
+
         $this->enabled = $enabled;
+        $this->forceSerializationCode = $forceSerializationCode;
```
mficzel, line 96: Where/how is that actually used? Could not figure that out.

Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
```diff
@@ -385,18 +374,11 @@ public function setProperty(ConfigurationProperty $property)
...
-            $this->arguments = [];
-        } else {
-            foreach ($arguments as $argument) {
-                if ($argument instanceof ConfigurationArgument) {
-                    $this->setArgument($argument);
-                } else {
-                    throw new InvalidConfigurationException(sprintf('Only ConfigurationArgument instances are allowed, "%s" given', is_object($argument) ? get_class($argument) : gettype($argument)), 1449217803);
-                }
-            }
+        $this->arguments = [];
```
kdambekalns, line 379: Funny, this willl remove the potential for a subtly weird behaviour, as calling `setArguments()` will no longer be additive (omething that probably was unintended…?).

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -111,36 +89,24 @@ public function build(): void
             if ($proxyClass === false) {
                 continue;
             }
-            $this->logger->debug(sprintf('Building dependency injection proxy for "%s"', $className), LogEnvironment::fromMethodName(__METHOD__));
+            $this->logger->debug(sprintf('Building dependency injection proxyG:q for "%s"', $className), LogEnvironment::fromMethodName(__METHOD__));
```
kdambekalns: `proxyG:q`, really? Looks like a vi command to me…

mficzel (approved):
Fine by 👀 and 💬

Neos.Flow/Classes/Annotations/Proxy.php
kdambekalns: See https://github.com/neos/flow-development-collection/pull/3510/changes#diff-8515e53b9cee4f4ad7e1fe36ff941a83d25992a174d5643a28ea6d9a6aad8e9bR172

That is in `Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php`, which by default is collapsed in the PR view, as the diff is too large.

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
kitsunet: Ha! indeed....

Neos.Flow/Classes/Cli/CommandManager.php
kitsunet: Yep good catch, I should probably drop this part of the change altogether because it's unrelated...

Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
kitsunet: Not really, there is exactly one single place in ConfigurationBuilder where we call this and at that point we actually getArguments() beforehand and add stuff to that array which is then passed to setArguments().

Neos.Flow/Configuration/Caches.yaml
kitsunet: They are actually not used.

Neos.Flow/Configuration/Objects.yaml
kitsunet: We do have support, I just didn't think to use it...

Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
kitsunet: I strongly assume this was a multi level merge errror where it was 

`$proxyClass->getMethod('__wakeup')->addPreParentCallCode($callBuildMethodsAndAdvicesArrayCode)->setReturnType('void');`

before, then became two lines and setReturnType was removed and it was wrongly put together. I will fix it. In fact I can reintroduce the return type now I guess because we require PHP 8.4

Neos.Flow/Configuration/Objects.yaml
kitsunet: Well, no, we do not have support... yet....
I guess I can define the other injections too, so it doesn't look as weird...

Neos.Flow/Classes/ObjectManagement/Configuration/Configuration.php
kdambekalns: What I meant is that before this change the behaviour was weird. All good!

Neos.Flow/Classes/Cli/CommandManager.php
kitsunet: done

Neos.Flow/Classes/Aop/Builder/ProxyClassBuilder.php
kitsunet: done

Neos.Flow/Configuration/Objects.yaml
kitsunet: I have a change ready to support named arguments from the object Configuration as well, but I would rather not also add it to this change. It's big enough as is.

kitsunet:
Apparently observing autowiring mode is not a good idea according to the tests. I guess TBD, I will remove it for now, as I had it originally...

Issue #3511, opened 2025-10-22
BUG: Proxy class building destroys nested attributes or enums


mficzel:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Proxy Building seems to destroy nested Attributes. This is an additional reason why ORM Attributes cannot yet be used. However it is important that fixing this will not enable ORM Attributes yet.

When the following Attributes are added to a class that is proxied:

```
#[Flow\Entity]
#[ORM\Table(uniqueConstraints: [
    new ORM\UniqueConstraint(name: 'nodeData_name_value', columns: ["nodedata", "name", "value"])
])]
```

the following code is generated in the proxy class 

```
#[Flow\Entity]
#[\Doctrine\ORM\Mapping\Table(uniqueConstraints: array (
    0 =>
        \Doctrine\ORM\Mapping\UniqueConstraint::__set_state(array(
            'name' => 'nodeData_name_value',
            'columns' =>
                array (
                    0 => 'nodedata',
                    1 => 'name',
                    2 => 'value',
                ),
            'fields' => NULL,
            'options' => NULL,
        )),
))]
```

### Expected Behavior

The generated code should be similar to the original
 ```
#[Flow\Entity]
#[ORM\Table(uniqueConstraints: [
    new ORM\UniqueConstraint(name: 'nodeData_name_value', columns: ["nodedata", "name", "value"])
])]
```

### Environment

```markdown
- Flow: 8.3+ 8,4+
- PHP: 8.0+
```

### Anything else?

Nested Attributes are supported from PHP 8.1 so one could say that we should only support this starting from Neos 8.4 that has PHP 8.2 as minimal requirement.

The attributes of proxy classes are rendered here:
- Class Attributes \Neos\Flow\ObjectManagement\Proxy\Compiler::renderAttribute
- Method Attributes \Neos\Flow\ObjectManagement\Proxy\ProxyMethodGenerator::buildAttributesCode

mficzel:
Technical reason is that starting from php 8.1 attributes may contain constant expressions as arguments which may be other attributes aswell. However the code generated by var_export is not a constant expression.

mhsdesign:
The issue of not supporting nested attributes also shows when attempting to proxy a method that contains those either via AOP or security.

[...]

PR #3512, opened 2025-10-22
BUGFIX: Handle enums and attributes (Objects) in attribute arguments for proxy classes

mficzel:
Handle Enums nested Attributes (actually any Objects) in attribute arguments for proxy classes.

Currently the Attributes for generated proxy classes have some limitations:
- Class Attributes: attribute values are exported via var_export which creates invalid code for attributes containing Enums or other Attributes as arguments
- Method Attributes: attributes of type Object will yield an UnsupportedAttribute Exception

This change addresses that by:
- creating Enums in generated code
- creating Objects via new () by inferring the constructor arguments from properties and getters
- handling this recursively in arrays

!!! This on itself does not add attribute support for persistence. See #3508 for that !!!

resolves: #3511
relates: #3075 

**Upgrade instructions**

No changes are needed only cases that previously yielded errors should work now.

**Review instructions**

The inference of constructor arguments from values obtained via ObjectAccess is close but not perfect. However it works in the cases we want to support like Doctrine Table Attributes and probably most other cases aswell. 

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -17,7 +17,6 @@ use Neos\Cache\Frontend\PhpFrontend;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\ObjectManagement\CompileTimeObjectManager;
 use Neos\Flow\ObjectManagement\Exception\ProxyCompilerException;
-use Neos\Flow\ObjectManagement\Exception\UnsupportedAttributeException;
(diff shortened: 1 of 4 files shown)


kdambekalns (approved):
👍 by 👀

mficzel:
Please Note: This change reverts a change from the proxyMethod builder that used the Compiler::renderAttribute in 8.3 but got its own methods for that in 8.4. I am not sure why as the logic is a bit tricky and it should be beneficial to only have one source of issues.

robertlemke (approved):
Fine by reading, but to be honest, I can't tell without further testing, if it works as intended. The test looks good at least ;-)

PR #3438, opened 2025-01-23
BUGFIX: Silence warning in `readCacheFile()`

kdambekalns:
readCacheFile() in SimpleFileBackend does fopen(). It wraps it into a try-catch clause and checks the result, but it still produces a warning if the file does not exist:

`Warning: fopen(/application/Data/Temporary/…): Failed to open stream: No such file or directory`

The only way to suppress that warning is to use the shut-up operator (`@`) in this place. Given that everything that can go wrong here is taken care of, I think this is fine.

**Checklist**

- [x] Code follows the PSR-2 coding style
- [x] Tests have been created, run and adjusted as needed
- [x] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [x] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [x] Reviewer - The first section explains the change briefly for change-logs
- [ ] ~Reviewer - Breaking Changes are marked with `!!!` and have upgrade-instructions~

PR #3513, opened 2025-10-23
BUGFIX: Silence warnings in SimpleFileBackend

kdambekalns:
`SimpleFileBackend` does `fopen()` in a few places. It wraps it into a try-catch clause and checks the result, but it still produces a warning if the file does not exist:

`Warning: fopen(/application/Data/Temporary/…): Failed to open stream: No such file or directory`

The only way to suppress that warning is to use the shut-up operator (`@`) in these places.

See neos/flow-development-collection#3438

**Checklist**

- [ ] ~~Reviewer - Breaking Changes are marked with `!!!` and have upgrade-instructions~~

### DIFF
--- a/Neos.Cache/Classes/Backend/SimpleFileBackend.php
+++ b/Neos.Cache/Classes/Backend/SimpleFileBackend.php
@@ -221,7 +221,7 @@ class SimpleFileBackend extends IndependentAbstractBackend implements PhpCapable
             // In the worst case, the unlink will just fail due to concurrent access and the caller needs to deal with that.
             return unlink($fileName);
         }
-        $file = fopen($fileName, 'rb');
+        $file = @fopen($fileName, 'rb');
         if ($file === false) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


kdambekalns:
The linked PR https://github.com/neos/flow-development-collection/pull/3438 fixed only the "simple read case", but we still see warnings a lot, so…

dlubitz (approved):
I'm fine with that. 

But wouldn't a `file_exists` also solve that?

kdambekalns:
> But wouldn't a `file_exists` also solve that?

Given that e.g. this
```php
if (!file_exists($pathAndFilename)) {
    return false;
}

return $this->readCacheFile($pathAndFilename);
```
is in place, or this:
```php
for ($directoryIterator = new \DirectoryIterator($this->cacheDirectory); $directoryIterator->valid(); $directoryIterator->next()) {
    if ($directoryIterator->isDot()) {
        continue;
    }

    $cacheEntryPathAndFilename = $directoryIterator->getPathname();
    $allData = $this->readCacheFile($cacheEntryPathAndFilename);
```
but we still see those errors, this "solution" seems better. Plus, we check the return value, so…

robertlemke (approved):

PR #3474, opened 2025-05-03
DRAFT: Raise PhpStan to level 8

nezaniel:
**Upgrade instructions**

None hopefully

**Review instructions**

Please check whether
-  all  types have been applied correctly
-  all fallback values make sense
-  this can be applied to 9.0 instead of 10

**Checklist**

- [ ] Code follows the PSR-2 coding style
- [ ] Tests have been created, run and adjusted as needed
- [ ] The PR is created against the [lowest maintained branch](https://www.neos.io/features/release-roadmap.html)
- [ ] Reviewer - PR Title is brief but complete and starts with `FEATURE|TASK|BUGFIX`
- [ ] Reviewer - The first section explains the change briefly for change-logs
- [ ] Reviewer - Breaking Changes are marked with `!!!` and have upgrade-instructions

PR #3515, opened 2025-10-24
TASK: Raise PhpStan to level 8 - Part 1

mficzel:
Raise PhpStan to level 8 for: 

- `Neos.Flow.Log`
- `Neos.FluidAdaptor`
- `Neos.Http.Factories`
- `Neos.Kickstarter`
- `Neos.Utility.Arrays`
- `Neos.Utility.Files`
- `Neos.Utility.MediaTypes`
- `Neos.Utility.ObjectHandling`
- `Neos.Utility.OpcodeCache`
- `Neos.Utility.Pdo`
- `Neos.Utility.Schema`
- `Neos.Utility.Unicode`

**Upgrade instructions**

No changes needed.

**Review instructions**

This the first of multiple PRs to cherry pick the changes from https://github.com/neos/flow-development-collection/pull/3474 from @nezaniel and @mhsdesign step by step and fix issues and breakiness. 

This is intended to be unbreaky but be the foundation for future refactorings that can at least partly be automated once we have really solid type informations. 

**Checklist**


mhsdesign:
wow thanks a lot! If you need help here let me know;) Also lets ignore the style CI failures as we want to move to php cs either way and they might have different expectations again;)

mficzel:
@mhsdesign is suggest to merge this in parts as otherwise the reviews will get really overwhelming.

mhsdesign:
oke i assume my commits will be part of another change then? idk why i did those but i probably did them to get flow back running

image

mficzel:
@mhsdesign yes your commits were mostly regarding flow. I cherry pick commits for each section one by one and afterwards remove breakiness and make tests green + verify neos still works.

mhsdesign:
Okay so this is the final set that should be tested and go into Flow 9.1 then ill review and test that ;) Thanks for splitting this mammoth!

mficzel:
I will try to prepare follow ups that may or may not be ready for 9.1 but imho this is about the maximal size for a pr that is chewable. So imho this can be reviewed and merged

mhsdesign (approved):
Thanks for reviving this change and battling with it to make it run :D

I did a full review locally and pushed some small adjustment changes - just mostly syntactical stuff.

I see we now use a lot of times the pattern `?: ''` to convert a `false` to a string. That is often okay as this the `false` case should never happen for constant preg expressions but in some cases ignoring this type juggling disguises bugs. So if there is real reason to believe it _could_ happen im definitely for having a sane fallback or telling whats wrong.

Thats exactly why i added a new exception to `\Neos\Utility\Files::getRecursiveDirectoryGenerator` see https://github.com/neos/flow-development-collection/pull/3515/commits/5769192c81c8450c10f9ed2c1744f8d343343e82:

> BUGFIX: Throw exception when `Files::getRecursiveDirectoryGenerator()` is used with `$returnRealPath` but realpaths cannot be calculated
>
> This is the case when using a virtual file system for testing in example (`vfs://`) or really any stream wrapper.
>
> Silently swallowing all the `realpath` misses disguises bugs and hiders developer experience when using a vfs.

PR #3520, opened 2025-11-11
BUGFIX: Check also for SkipCsrfProtection annotation

kitsunet:
We currently check only for the "tag" `@skipcsrfprotection`, any (action) method annotated with  `@Flow\SkipCsrfProtection` is only part of the times handled correctly.

Attributes `#[Flow\SkipCsrfProtection]` or an annotation with comment on the same line `@Flow\SkipCsrfProtection Some explanation` are not handled currently.

Obviously this is not intended, rather using the tag should be avoided.

This adds the check for reflected annotations/attributes.

**Review instructions**

The root cause for this behaviour is that we just never used `isAnnotatedWith` but only relied on `isTaggedWith` to interpret if a method is tagged. The tagging parsing is custom done via `DocCommentParser` and utterly broken https://github.com/neos/flow-development-collection/pull/3520#issuecomment-3518076068 in that it attempts to convert annotations to tags by dropping the namespace but only if no comments exist on that line.


mhsdesign:
The problem is rooted in the DocBlock parser of ours - and i could validate what you wrote in [slack](https://neos-project.slack.com/archives/C050KKBEB/p1747671750453739?thread_ts=1747669833.582169&cid=C050KKBEB).

The `Neos\Flow\Reflection\DocCommentParser` has a habit of making everything into flat non-namespaced tags except if there is a comment.

The following unit tests hold true:

```php
/**
 * @test
 */
public function annotationAsTagValue()
{
    $parser = new DocCommentParser();
    $parser->parseDocComment('/** @Flow\SkipCsrfProtection */');

    $this->assertEquals(['skipcsrfprotection' => []], $parser->getTagsValues());
}

/**
 * @test
 */
public function annotationAsTagValueOmitsNamespace()
{
    $parser = new DocCommentParser();
    $parser->parseDocComment('/** @SomeOtherNameSpace\SkipCsrfProtection */');

    $this->assertEquals(['skipcsrfprotection' => []], $parser->getTagsValues());
}

/**
 * @test
 */
public function annotationAsTagValueWithComments()
{
    $parser = new DocCommentParser();
    $parser->parseDocComment('/** @Flow\SkipCsrfProtection Some comment in this line */');

    $this->assertEquals(['flow\skipcsrfprotection' => [
        'Some comment in this line'
    ]], $parser->getTagsValues());
}
```

kitsunet:
Yep but that only teslls me we need to deprecate using tags ASAP and get rid of this LOL but for now checking both seems the sensible thing to do.

mhsdesign (approved):
Pushed the tests showing whats wrong with the tags ... tough we probably cannot bugfix that or rather want that.

Approving as discussed in [slack](https://neos-project.slack.com/archives/C04V4C6B0/p1762875203888329) thanks! Good to have that problem understood and solved.

mhsdesign:
Im actually confused is there no actual full test for either the `skipcsrfprotection` tag or as annotation? There is a `MockActionController` but that is dead code and not used (can be removed with this pr as well) 
I adjusted the one unit test to use the new approach.

PR #3521, opened 2025-11-15
BUGFIX: Fix race condition during log rotation -> the `.lock` file is not properly locked

mhsdesign:
See https://discuss.neos.io/t/unlink-system-development-log-10-no-such-file/7140/2

Multiple processes writing logs can easily run into the following error when its time to rotate the logs:

```
Warning: rename(Data/Logs/System_Development.log.3,Data/Logs/System_Development.log.4): No such file or directory in Packages/Framework/Neos.Flow.Log/Classes/Backend/FileBackend.php line 217

  Code: 1
  File: Packages/Framework/Neos.Flow/Classes/Error/ErrorHandler.php
  Line: 80
```

During log rotation we attempt to use a lock via `file_exists` but its not an atomic operation how we use it https://github.com/neos/flow-development-collection/blob/54cea0a2cafb9ee1475ae2fdbc1fd3a1830ddc86/Neos.Flow.Log/Classes/Backend/FileBackend.php#L194-L198 `file_exists` could be true for two processes at the same time in a race condition and both processes would `touch` (create) the lock file.

Instead, we need to use a atomic locking via `flock`

```
$lockResource = fopen($this->logFileUrl . '.lock', 'w+');

$exclusiveNonBlockingLockResult = flock($lockResource, LOCK_EX | LOCK_NB);
if ($exclusiveNonBlockingLockResult === false) {
   // someone else is on it
   return;
}

// do something only this process is supposed to do...

if (!flock($setupLockResource, LOCK_UN)) {
    throw new \RuntimeException('failed to release lock');
}
```

Tested on MacOs with 3 simultaneous shells running each a flow command that writes logs in a shell while loop.

I hope windows supports `flock($lockResource, LOCK_EX | LOCK_NB);` to exclusively claim a lock without waiting for it if its claimed.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Flow.Log/Classes/Backend/FileBackend.php
+++ b/Neos.Flow.Log/Classes/Backend/FileBackend.php
@@ -191,10 +191,11 @@ class FileBackend extends AbstractBackend
      */
     protected function rotateLogFile(): void
     {
-        if (file_exists($this->logFileUrl . '.lock')) {
+        $lockResource = fopen($this->logFileUrl . '.lock', 'w+');
+
+        $exclusiveNonBlockingLockResult = flock($lockResource, LOCK_EX | LOCK_NB);
+        if ($exclusiveNonBlockingLockResult === false) {
             return;
-        } else {
-            touch($this->logFileUrl . '.lock');
         }
 
         if ($this->logFilesToKeep === 0) {


bwaidelich (approved):
by reading

dlubitz (approved):

mhsdesign:
The docs say windows supports this kind of locking so lets go ;) Style CI is unrelated.

kdambekalns (commented):
post-merge 👍

PR #3522, opened 2025-11-17
FEATURE: Command to list Aspect targets

kitsunet:
The new command "flow:aop:aspecttargets" will provide information about found aspects in the system as well as the advices declared and what they target. Additionally introduced interfaces, properties and traits can also be shown.

There is a filter to show only a single aspect via

-- only-aspect "My\Aspect\ClassAspect"

and the various outputs can be switched on or off as required. By default only advices are shown.

Note that this is a compiletime command and does not have property mapping available, therefore the boolean arguments are best given as 0 or 1.

Some refactoring happened along the way to make enough of the functionality accessible by the command controller.


kitsunet:
ja, ja, ja, I know...

PR #3527, opened 2025-11-24
TASK: Run pipeline with PHP8.5

dlubitz:


Neos.Eel/Tests/Unit/Helper/MathHelperTest.php
```diff
@@ -211,7 +212,7 @@ public function variousFunctionsWork($method, $expected)
         ]);
         $result = $evaluator->evaluate($method, $context);
         if ($expected === static::NAN) {
-            self::assertTrue(is_nan($result), 'Expected NAN, got value "' . $result . '"');
+            self::assertTrue(is_nan($result), 'Expected NAN, got value "' . @(string)$result . '"');
```
kdambekalns, line 215: Wait, what? String concatenation now needs an explicit type cast?

kdambekalns: Ah, in case of `NAN`, PHP now emits `unexpected NAN value was coerced to string` as a warning. ✅

kdambekalns (approved):
👏

dlubitz:
Cherry-Picked into 8.3

PR #3531, opened 2025-12-11
TASK: Add missing symfony polyfill replacements due to php82 version

mhsdesign:
This change might need adjustments if one has a composer `replacement` in ones distribution as flow failed to provide it earlier. Updating could then cause a conflict as composer - i believe - only respects one replace directive to exist per package.

**Upgrade instructions**

Remove these replacements in your own composer.json

**Review instructions**

**Checklist**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,8 @@
         "symfony/polyfill-php73": "*",
         "symfony/polyfill-php74": "*",
         "symfony/polyfill-php80": "*",
+        "symfony/polyfill-php81": "*",
+        "symfony/polyfill-php82": "*",
         "symfony/polyfill-mbstring": "*",
         "neos/cache": "self.version",
         "neos/eel": "self.version",


composer.json
```diff
@@ -57,6 +57,8 @@
         "symfony/polyfill-php73": "*",
         "symfony/polyfill-php74": "*",
         "symfony/polyfill-php80": "*",
+        "symfony/polyfill-php81": "*",
+        "symfony/polyfill-php82": "*",
```
dlubitz, line 61: Why do we need the polyfill for 8.2? 8.2 is the min required version of Flow 9.x.

mhsdesign: we _replace_ the polyfill ... e.g. we dont need that dependency which will be installed otherwise by some random `symfony` to use the newest  8.2 functions which already exist in the runtime

mhsdesign: that replace pattern is the way we get rid of them in all installations

dlubitz: Ahh, I see. Didn't know about that.

If someone also didn't: https://php.watch/articles/composer-replace-polyfills

dlubitz (approved):

PR #3533, opened 2026-01-21
BUGFIX: Pass through enum mapping to ORM

kitsunet:
This change allows to use the enumType property for a `Column` attribute and thus use enums as types in entities with the mapping being handled by doctrine ORM. The special case for the "type" is no longer needed as it has been nullable in the attribute/annotation for some years now, so we can rely on a default null if you didn't configure the type, and therefore triggering our auto detection in this case. We need this special case removed to allow string backed enums to be mapped correctly, as we need to fix the type to string and bypass auto detection.

In 8.3 we could remove the special handling for type already, but enumType was not available in the lowest accepted ORM version, therefore I targeted 8.4 as the required ORM version contains enumType.


kitsunet:
Probably should add a behat test with an enum in an entity

dlubitz (approved):
Looks good to me

kdambekalns (approved):

kitsunet:
Well behat in Flow still needs a major refactoring so functional test it is...

Issue #3480, opened 2025-05-22
FEATURE: Support Xdebug Native Path Mapping


jrenggli:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

Flow relies heavily on Proxy Classes, which complicates debugging and intercepting traffic using traditional tools. To address this, developers have previously used the [Flow Debug Proxy](https://github.com/dfeyer/flow-debugproxy) as a workaround.

In February 2024, [Robert Lemke initiated a funding request](https://discuss.neos.io/t/request-for-funding-xdebug-support-for-flow-and-neos/6512) to support Derick Rethans, the author of Xdebug, in developing native path mapping support. The goal: significantly improve the debugging experience for Flow and Neos developers.

> This is one of the most requested improvements related to testing and debugging Flow. A long time after I first discussed this with Derick, we now finally have the chance to make it happen.

As of May 2025, Derick has released an update titled ["Finishing the Initial Preview"](https://xdebug.org/funding/001-native-path-mapping/updates/2025-05-13), indicating substantial progress on this feature.

As far as I see, there have been no visible efforts from the Neos community to prepare Flow for adoption of this feature once it's officially released.

After experimenting with recent Xdebug builds and the Flow Proxy Compiler, I can confirm that the native path mapping feature works as expected in my local setup.

IMHO we need to do the following steps:
* Implement a builder for path maps directly in the Neos Flow core.
* Update documentation, write tutorials
* Testing, Testing, Testing...

If this initiative is already being tracked elsewhere—apologies for the redundancy. At the very least, I hope this post helps centralize the conversation and points interested developers toward the right direction.

jrenggli:
The following content is just a dump of all the steps necessary to make Xdebug Native Path Mapping available on "my machine".

# Build and install xdebug development version

I'm using DDEV here

```Dockerfile
ENV XDEBUG_REPOSITORY=https://github.com/derickr/xdebug.git
ENV XDEBUG_REF=001-native-path-mapping
RUN cd /opt \
    && apt-get update \
    && apt-get install --yes php${DDEV_PHP_VERSION}-dev build-essential \
    && git clone --depth=1 ${XDEBUG_REPOSITORY} --branch ${XDEBUG_REF} \
    && cd /opt/xdebug \
    && sed -i '/fprintf/d' ./src/lib/maps/maps.c \
    && phpize${DDEV_PHP_VERSION} \
    && ./configure --enable-xdebug \
    && make \
    && make install \
[...]

PR #3535, opened 2026-01-29
TASK: Configurable Flow support for Xdebug native path mapping

dlubitz:
**Adds Flow support for Xdebug native path mapping (>= 3.5).**

If enabled (disabled by default for now), the XdebugPathMappingBuilder connects to the "afterCompile" signal and builds a `./xdebug/flow.map` to provide the path mapping information for xdebug.

Example (.xdebug/flow.map)
```
# Created by Flow Framework during compile time proxy generation.
#
# Ensure you are using xdebug >= v3.5 with enabled path mapping.
# Configuration in your php.ini:
# xdebug.mode = develop
# xdebug.path_mapping = 1
#
# ----------------------------------------------------------------
# Last update: 2026-03-03 22:49:57
remote_prefix:/var/www/html/Data/Temporary/Development/SubContextDdev/Cache/Code/Flow_Object_Classes/
local_prefix:/var/www/html/
Neos_Diff_Diff.php = Packages/Neos/Neos.Diff/Classes/Diff.php
Neos_Diff_SequenceMatcher.php = Packages/Neos/Neos.Diff/Classes/SequenceMatcher.php
Neos_Diff_Renderer_Html_HtmlInlineRenderer.php = Packages/Neos/Neos.Diff/Classes/Renderer/Html/HtmlInlineRenderer.php
Neos_Diff_Renderer_Html_HtmlSideBySideRenderer.php = Packages/Neos/Neos.Diff/Classes/Renderer/Html/HtmlSideBySideRenderer.php
...
```

**Configuration**
```
Neos:
  Flow:
    object:
      proxy:
        enableXdebugPathMapping: true
```
_Should be only enabled in development context._

**Why targeting 8.3?**
Even if this is a new feature, we wanted to provide this all developers as soon as possible. So we add this already to 8.3 but disabled by default, to be non-breaking.
We will enable the path mapping file generation by default in development context beginning with Flow 9.2. 

**How to use**
You need xdebug in v3.5+ with enabled path mapping. Also the xdebug mode needs to be not "OFF". The php.ini configuration could be:
```
xdebug.mode = develop
xdebug.path_mapping = 1
```

See: https://xdebug.org/funding/001-native-path-mapping

Depends on: #3538
Fixes: #3480

### DIFF
--- a/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
+++ b/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
@@ -90,11 +90,11 @@ final class AnnotationsCacheFlusher
     }
 
     /**
-     * A slot that flushes caches as needed if classes with specific annotations have changed @param array<string> $classNames The full class names of the classes that got compiled
+     * A slot that flushes caches as needed if classes with specific annotations have changed @see registerAnnotation()
+     *
+     * @param array<string> $classNames The full class names of the classes that got compiled
      * @return void
      * @throws NoSuchCacheException
-     * @see registerAnnotation()
-     *
      */
     public function flushConfigurationCachesByCompiledClass(array $classNames): void
     {
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
@@ -72,7 +72,7 @@ class XdebugPathMappingBuilder
             '#',
             '# Ensure you are using xdebug >= v3.5 with enabled path mapping.',
             '# Configuration in your php.ini:',
-            '# xdebug.mode = develop',
+            '# xdebug.mode = debug',
             '# xdebug.path_mapping = 1',
             '#',
             '# ----------------------------------------------------------------',
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
@@ -1,0 +1,107 @@
+<?php
+
+namespace Neos\Flow\ObjectManagement\Proxy;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Cache\Backend\SimpleFileBackend;
+use Neos\Flow\Annotations as Flow;
+use Neos\Flow\Cache\CacheManager;
+use Neos\Utility\Files;
+
+/**
+ * @Flow\Scope("singleton")
+ */
+class XdebugPathMappingBuilder
+{
+    /**
+     * @Flow\Inject
+     * @var CacheManager
+     */
+    protected $cacheManager;
+
+    /**
+     * @var array
+     */
+    protected $settings;
+
+    /**
+     * @param array $settings
+     * @return void
+     */
+    public function injectSettings(array $settings): void
+    {
+        $this->settings = $settings;
+    }
+
+    public function injectCacheManager(CacheManager $cacheManager): void
+    {
+        $this->cacheManager = $cacheManager;
+    }
+
+    /**
+     * @param array<string, array{path: string, proxyClassIdentifier: string}> $compiledClasses
+     * @return void
+     */
+    public function buildFromCompiledClasses(array $compiledClasses): void
+    {
+        if ($compiledClasses === []) {
+            return;
+        }
+
+        if (!$this->isXdebugPathMappingEnabled()) {
+            return;
+        }
+
+        $cacheBackend = $this->cacheManager->getCache('Flow_Object_Classes')->getBackend();
+        if (!$cacheBackend instanceof SimpleFileBackend) {
+            return;
+        }
+        $cacheDirectory = $cacheBackend->getCacheDirectory();
+
+        $mappingFileLines = [
+            '# Created by Flow Framework during compile time proxy generation.',
+            '#',
+            '# Ensure you are using xdebug >= v3.5 with enabled path mapping.',
+            '# Configuration in your php.ini:',
+            '# xdebug.mode = develop',
+            '# xdebug.path_mapping = 1',
+            '#',
+            '# ----------------------------------------------------------------',
+            sprintf('# Last update: %s', date('Y-m-d H:i:s')),
+            sprintf('remote_prefix:%s', $cacheDirectory),
+            sprintf('local_prefix:%s', FLOW_PATH_ROOT),
+        ];
+        foreach ($compiledClasses as $data) {
+            $localPath = str_replace(FLOW_PATH_ROOT, '', $data['path']);
+            $remotePath = $data['proxyClassIdentifier'];
+            $mappingFileLines[] = sprintf("%s.php = %s", $remotePath, $localPath);
+        }
+        // Add an empty line to the end of the file
+        $mappingFileLines[] = '';
+
+        Files::createDirectoryRecursively($this->getXdebugMappingFilePath());
+        file_put_contents(
+            Files::concatenatePaths([$this->getXdebugMappingFilePath(), 'flow.map']),
+            implode("\n", $mappingFileLines)
+        );
+    }
+
+    private function isXdebugPathMappingEnabled(): bool
+    {
+        return isset($this->settings['object']['proxy']['enableXdebugPathMapping']) && $this->settings['object']['proxy']['enableXdebugPathMapping'] === true;
+    }
+
+    private function getXdebugMappingFilePath(): string
+    {
+        return Files::concatenatePaths([FLOW_PATH_ROOT, '.xdebug']);
+    }
+}
--- a/Neos.Flow/Classes/Package.php
+++ b/Neos.Flow/Classes/Package.php
@@ -17,6 +17,7 @@ use Neos\Flow\Configuration\Source\YamlSource;
 use Neos\Flow\Core\Booting\Step;
 use Neos\Flow\Http\Helper\SecurityHelper;
 use Neos\Flow\ObjectManagement\Proxy\Compiler;
+use Neos\Flow\ObjectManagement\Proxy\XdebugPathMappingBuilder;
 use Neos\Flow\Package\Package as BasePackage;
 use Neos\Flow\Package\PackageManager;
 use Neos\Flow\ResourceManagement\ResourceManager;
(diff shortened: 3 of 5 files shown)


skurfuerst (dismissed):
I love it, great feature :) I'd suggest to enable it by default in for Development context; as otherwise people will likely not profit from it if they explicitly need to enable it?

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
```diff
@@ -1,107 +0,0 @@
...
-use Neos\Flow\Annotations as Flow;
-use Neos\Flow\Configuration\ConfigurationManager;
-use Neos\Flow\Reflection\ReflectionService;
-use Neos\Flow\Log\Utility\LogEnvironment;
-use Psr\Log\LoggerInterface;
-
-/**
- * @Flow\Scope("singleton")
- */
-final class AnnotationsCacheFlusher
```
skurfuerst, line 26: BTW why is it safe to remove this class?

skurfuerst (changes requested):
@dlubitz I tested it, and found the following issues:

- on incremental file changes (i.e. not full rebuilds but partial rebuilds, i.e. when the Compiletime cache is filled but you add a single new file), then the `flow.map` only contains the single file -> so IMHO we need to incrementally merge updates in this file.

- additionally, the file needs to end with newline: `[21916] [Path Mapping] WARN: Parse error in path mapping file '/Users/sebastian/src/neos-8.3/.xdebug/flow.map' on line 1111: Line does not end in a new line`

Currently trying to make it work with xdebug 3.5.0 :) 
EDIT: Xdebug 3.5.0/3.5.1 will NOT WORK, due to https://github.com/xdebug/xdebug/commit/33fbfe67d3d213f947d991fc734aa8a309e54e24 - we need this fixed, as this breaks our PHP Path checking


TEST PROTOCOL on master:

```bash

pie install xdebug/xdebug:dev-master

cat /opt/homebrew/etc/php/8.4/conf.d/xdebug.ini
xdebug.mode = debug
xdebug.log = /tmp/xdebug.log
xdebug.path_mapping = 1

```


All the best,
Sebastian

dlubitz:
Thanks @skurfuerst for testing.

The partial compilation is not covered in fact. I was not aware of that and assumed it always compiles everything. I'll have a look into this.

I don't see, why it shouldn't work with 3.5.0, as I tested it with this version and it worked pretty well.

skurfuerst:
@dlubitz for me 3.5.0/3.5.1 breaks because of debug output in running "./php" in CLI mode (if not on windows); and then the Flow compilation fails with a line like:

```
Flow has been run with "/opt/homebrew/Cellar/php/8.4.14/bin/php", while
the PHP version Flow is configured to use for subrequests is
"/opt/homebrew/Cellar/php/8.4.14/bin/phpMap directory
/Users/sebastian/src/neos-8.3/.xdebug is a directory: yes"
```

This was fixed on xdebug master with the commit linked above :)

skurfuerst:
@dlubitz re. partial compilation: IMHO loading the map file, applying the changed lines, and dumping it should work pretty well. We just need to somehow notice if the cache is completely empty, because then we want to rebuild completely :) 

BTW: did you try out `remote_prefix` and `local_prefix`? This could make the whole file a lot shorter and thus quicker to parse etc: https://xdebug.org/funding/001-native-path-mapping#file-format

I still need to test whether this works in Docker properly or not; it might be that the user then needs a way to adjust the remote_prefix accordingly  :)

skurfuerst:
@dlubitz BTW: as you can see from my comments - I am REALLY excited about this feature, and if you like we can collaborate in the upcoming days to get this done-done :) THANKS for your initiative :)

dlubitz:
Good catch. remote_prefix should work easily as we prefix all proxies with the cache path anyway.

For local_prefix this might work, but I'm not sure if we should do this. This would AFAIK require to extract the root path from all class file paths.

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
dlubitz: Yes, I discussed it with @kitsunet.

This class was never called as the signal was never dispatched. 
See: https://github.com/neos/flow-development-collection/pull/3538

dlubitz:
@skurfuerst 
> I love it, great feature :) I'd suggest to enable it by default in for Development context; as otherwise people will likely not profit from it if they explicitly need to enable it?

Idea is to enable it by default with next feature release (9.2), but not in the bugfix releases (8.3 - 9.1).


I also fixed the bug when single classes have been changed and added the prefixes for local and remote.

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private function isXdebugPathMappingEnabled(): bool
+    {
+        return isset($this->settings['object']['proxy']['enableXdebugPathMapping']) && $this->settings['object']['proxy']['enableXdebugPathMapping'] === true;
+    }
+
+    private function getXdebugMappingFilePath(): string
+    {
+        return FLOW_PATH_ROOT . '.xdebug';
```
kdambekalns: We have a utility class to do path operations. In this case I see no reason why this would break, but… lead by example?

kdambekalns (approved):
Aside from a nitpick, 👍 by 👀

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
dlubitz: Thank you. I changed it.

dlubitz:
@skurfuerst Anything you want to add or does this PR work for you now?

skurfuerst:
@dlubitz Hey, for me it still does not work...

```
[93868] [Path Mapping] WARN: Parse error in path mapping file '/Users/sebastian/src/neos-8.3/.xdebug/flow.map' on line 3: Remote mapping part ('remote_prefix') type (file) must match local mapping part ('/Users/sebastian/src/neos-8.3/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/') type (directory)
```

Checking the source code, I believe the format is still subtilely wrong:

**Wrong, as it is generated**:

```
# Created by Flow Framework during compile time proxy generation.
# Last update: 2026-03-02 20:38:12
remote_prefix=/Users/sebastian/src/neos-8.3/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/
local_prefix=/Users/sebastian/src/neos-8.3/
Neos_Diff_SequenceMatcher.php = Packages/Neos/Neos.Diff/Classes/SequenceMatcher.php
```

**correct**:

```
remote_prefix:/Users/sebastian/src/neos-8.3/Data/Temporary/Development/Cache/Code/Flow_Object_Classes/
local_prefix:/Users/sebastian/src/neos-8.3/
```

Additionally, the file must end in a newline. With these two changes it works for me :)

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+        $cacheBackend = $this->cacheManager->getCache('Flow_Object_Classes')->getBackend();
+        if (!$cacheBackend instanceof SimpleFileBackend) {
+            return;
+        }
+        $cacheDirectory = $cacheBackend->getCacheDirectory();
+
+        $mapping = [
+            '# Created by Flow Framework during compile time proxy generation.',
+            sprintf('# Last update: %s', date('Y-m-d H:i:s')),
+            sprintf('remote_prefix=%s', $cacheDirectory),
```
skurfuerst:
```suggestion
            sprintf('remote_prefix: %s', $cacheDirectory),
```

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+            sprintf('local_prefix=%s', FLOW_PATH_ROOT),
```
skurfuerst:
```suggestion
            sprintf('local_prefix: %s', FLOW_PATH_ROOT),
```

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+        foreach ($compiledClasses as $data) {
+            $localPath = str_replace(FLOW_PATH_ROOT, '', $data['path']);
+            $remotePath = $data['proxyClassIdentifier'];
+            $mapping[] = sprintf("%s.php = %s", $remotePath, $localPath);
+        }
+
+        Files::createDirectoryRecursively($this->getXdebugMappingFilePath());
+        file_put_contents(
+            Files::concatenatePaths([$this->getXdebugMappingFilePath(), 'flow.map']),
+            implode("\n", $mapping)
```
skurfuerst:
```suggestion
            implode("\n", $mapping) . "\n"
```

skurfuerst (commented):
@dlubitz with the three adjustments it works nicely for me :) Wondering how it could work for you though :)

dlubitz:
Thank you @skurfuerst! Good catch.
I only checked the file generation worked after latest changes 🙈 

Will change that. Where did you found the info about the newline at the end?

skurfuerst:
I found an error in my log if the file did not end with newline :)

dlubitz:
@skurfuerst I've fixed the bugs and tested the changes this time. This works now fine for me.

I also changed the preparation PR #3538 , so it contains all compiler and signal changes. This PR here is depending on that preparation PR and only contains the xdebug integration.

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
```diff
@@ -0,0 +1,107 @@
...
+            return;
+        }
+        $cacheDirectory = $cacheBackend->getCacheDirectory();
+
+        $mappingFileLines = [
+            '# Created by Flow Framework during compile time proxy generation.',
+            '#',
+            '# Ensure you are using xdebug >= v3.5 with enabled path mapping.',
+            '# Configuration in your php.ini:',
+            '# xdebug.mode = develop',
```
skurfuerst: debug :D

skurfuerst (approved):
+1 by reading and testing! See my mini nitpick (can be done on merge :) )

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
dlubitz: Is changed.

mhsdesign:
Thank you! I adjusted the title to be `TASK` as we dont merge `FEATURE` into old releases;) :D
And yes for Flow 9.2 we can set it by default.

baschny:
> Thank you! I adjusted the title to be `TASK` as we dont merge `FEATURE` into old releases;) 

Who are you trying to fool?? This is a FEATURE with all capital letters, and a GREAT ONE and I am greatful that this is in 8.3/8.4 also! Thanks a lot for that!



> Thank you! I adjusted the title to be `TASK` as we dont merge `FEATURE` into old releases;) :D And yes for Flow 9.2 we can set it by default.

PR #3536, opened 2026-02-02
FEATURE: Add MCP support for Flow's request handling cycle

nezaniel:
This adjusts the RequestBodyParsingMiddleware so that the request body can be processed by an MCP server.

**Upgrade instructions**

none

**Review instructions**

this should not have any side effects on anything, but I'm not sure

**Checklist**

- [X] Code follows the PSR-12 coding style
- [X] The PR is created against the 8.3 branch (https://www.neos.io/features/release-roadmap.html)

### DIFF
--- a/Neos.Flow/Classes/Http/Middleware/RequestBodyParsingMiddleware.php
+++ b/Neos.Flow/Classes/Http/Middleware/RequestBodyParsingMiddleware.php
@@ -51,6 +51,9 @@ class RequestBodyParsingMiddleware implements MiddlewareInterface
     protected function parseRequestBody(ServerRequestInterface $httpRequest)
     {
         $requestBody = $httpRequest->getBody()->getContents();
+        if ($httpRequest->getBody()->isSeekable()) {
+            $httpRequest->getBody()->rewind();
+        }
         if ($requestBody === null || $requestBody === '') {
             return $requestBody;
         }


kdambekalns (commented):
To be honest: The explanation does not explain why this is needed. Could you elaborate on that a bit more?

Maybe even in a way that is more generic? Because, why should we do this only for MCP? Surely it's not MCP-specific?

nezaniel:
Converting to draft as there are more things to be considered as well.

As far as I understood it, the [mcp/sdk](https://github.com/modelcontextprotocol/php-sdk) server uses the request  body in a way that also supports streaming, meaning it cannot consume an already resolved request body.  My interpretation (not knowledge) would be that this affects everything of (potential) content-type "text/event-stream", whether MCP or not.

I'm not exactly sure whether this belongs on framework level as I can do the rewind in my MCP base controller as well with the same working result.

nezaniel:
I've managed to get streaming working with our base MCP controller (see https://github.com/sitegeist/Sitegeist.Pandora). The handling of "text/event-stream" must imho be done by Flow's request handler (we currently use a custom one). I'd like to discuss a possible implementation (it might be enough to just check the response content type) and target version. As this is a new feature, I guess it would go to 9.2?

kdambekalns:
Looking at the `SseRequestHandler` in Sitegeist.Pandora it seems there is more code to bring into Flow. I'd be fine with 9.2 (still), if that works for you. 😎

nezaniel:
As we have a working custom solution now (it's on Neos 8 anyway), I'd be fine with 9.3 as well as this might need some discussion and has no need for haste

Neos.Flow/Classes/Http/Middleware/RequestBodyParsingMiddleware.php
```diff
@@ -51,6 +51,9 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
     protected function parseRequestBody(ServerRequestInterface $httpRequest)
     {
         $requestBody = $httpRequest->getBody()->getContents();
+        if ($httpRequest->getBody()->isSeekable()) {
```
kitsunet, line 54: I wonder if we shouldn't actually NOT get the contents here at all, if the body isn't seekable. Because that will certainly a problem later on, if you can't retrieve the body content again.

PR #3537, opened 2026-02-11
BUGFIX: Remove singleton annotation on validators

### LINKED ISSUE #2879: BUG: EmailAddressValidator is of scope singleton, but supports options

### Current Behavior

When resolving the `Neos\Flow\Validation\Validator\EmailAddressValidator` with options, I get the following exception:

`The validator "Neos\Flow\Validation\Validator\EmailAddressValidator" is of scope singleton, but configured to be used with options.
A validator with options must be of scope prototype.`

Exception Code: 1358958575
Exception Type: Neos\Flow\Validation\Exception\InvalidValidationConfigurationException
Original File: Packages/Framework/Neos.Flow/Classes/Validation/ValidatorResolver.php
Line: 108

### Expected Behavior

Successfully resolving the EmailAddressValidator with options.

### Steps To Reproduce

Resolve the Validator through the `Neos\Flow\Validation\ValidationResolver`:

`$emailAddressValidator = $this->validatorResolver->getValidator(EmailAddressValidator::class, ['strict' => true]);`

### Environment

```markdown
- Flow:
- PHP: 8.1.1
```

### Anything else?

Unsure if we just should change the scope to prototype, or if there is a better solution.

dlubitz:
Fixes #2879 

The validators are not meant to be singleton. The singleton annotations causes issues if you try to set options via `Flow\Validate` annotation or try to create a validator with `ValidatorResolver->createValidator()`.

Technically a singleton would work for validators without options. But the advantage in not constructing them every time, seems to be negligible.

### DIFF
--- a/Neos.Flow/Classes/Validation/Validator/AlphanumericValidator.php
+++ b/Neos.Flow/Classes/Validation/Validator/AlphanumericValidator.php
@@ -17,7 +17,6 @@ use Neos\Flow\Annotations as Flow;
  * Validator for alphanumeric strings.
  *
  * @api
- * @Flow\Scope("singleton")
  */
 class AlphanumericValidator extends AbstractValidator
 {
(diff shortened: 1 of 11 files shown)


robertlemke (approved):
Makes sense! 👍🏻

kdambekalns (approved):

PR #3538, opened 2026-02-19
TASK: Replace broken signal "compiledClasses" with "afterCompile"

dlubitz:
This replaces the broken signal `compiledClasses`, which has never worked as it was missing the manual signal dispatch, because the Compiler is not proxied.

This is a preparation for https://github.com/neos/flow-development-collection/issues/3480, which needs a new slighlty different signal.

This also fixes the AnnotationCacheFlusher dependency injection, to get this working again.

### DIFF
--- a/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
+++ b/Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
@@ -16,8 +16,8 @@ namespace Neos\Flow\Cache;
 use Neos\Cache\Exception\NoSuchCacheException;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Configuration\ConfigurationManager;
-use Neos\Flow\Reflection\ReflectionService;
 use Neos\Flow\Log\Utility\LogEnvironment;
+use Neos\Flow\Reflection\ReflectionService;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -58,6 +58,26 @@ final class AnnotationsCacheFlusher
      */
     private $annotationToCachesMap = [];
 
+    public function injectLogger(LoggerInterface $logger): void
+    {
+        $this->logger = $logger;
+    }
+
+    public function injectConfigurationManager(ConfigurationManager $configurationManager): void
+    {
+        $this->configurationManager = $configurationManager;
+    }
+
+    public function injectCacheManager(CacheManager $cacheManager): void
+    {
+        $this->cacheManager = $cacheManager;
+    }
+
+    public function injectReflectionService(ReflectionService $reflectionService): void
+    {
+        $this->reflectionService = $reflectionService;
+    }
+
     /**
      * Register an annotation that should trigger a cache flush
      *
@@ -70,11 +90,11 @@ final class AnnotationsCacheFlusher
     }
 
     /**
-     * A slot that flushes caches as needed if classes with specific annotations have changed @see registerAnnotation()
-     *
-     * @param array<string> $classNames The full class names of the classes that got compiled
+     * A slot that flushes caches as needed if classes with specific annotations have changed @param array<string> $classNames The full class names of the classes that got compiled
      * @return void
      * @throws NoSuchCacheException
+     * @see registerAnnotation()
+     *
      */
     public function flushConfigurationCachesByCompiledClass(array $classNames): void
     {
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -15,6 +15,7 @@ use Neos\Cache\Frontend\PhpFrontend;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\ObjectManagement\CompileTimeObjectManager;
 use Neos\Flow\Reflection\ReflectionService;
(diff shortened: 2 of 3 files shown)


kdambekalns (approved):
👍 by 👀

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
```diff
@@ -1,107 +0,0 @@
...
-     */
-    private $annotationToCachesMap = [];
-
-    /**
-     * Register an annotation that should trigger a cache flush
-     *
-     * @param string $annotationClassName fully qualified class name of the annotation
-     * @param string[] $cacheNames Cache names to flush if a class containing the given annotation is compiled (e.g. ["Flow_Mvc_Routing_Route", Flow_Mvc_Routing_Resolve"])
-     */
-    public function registerAnnotation(string $annotationClassName, array $cacheNames): void
```
mhsdesign, line 67: as it seems in 8.3 there are no usages though Neos 9.0 introduced a feature which requires this flushing:

```php
$annotationsCacheFlusher->registerAnnotation(Flow\Route::class, ['Flow_Mvc_Routing_Route', 'Flow_Mvc_Routing_Resolve']);
```

https://github.com/neos/flow-development-collection/pull/3325

So we rather need to repair the `AnnotationsCacheFlusher` at least for Neos version 9.0 but then it does not make sense to go through the hassle of removing this here first.

Im not aware that anyone else uses the `AnnotationsCacheFlusher` though removing classes and methods in a patch level release is always nasty as it will let your Flow application die instantly. Even if the functionality - like here - was a noop, its always good to leave an empty deprecated facade.

Under these circumstances id rather like to see the introduction of a new `afterCompile` in this pr and also the fix of the `AnnotationsCacheFlusher`. Doesnt this make sense the most and introduces the least hassle - also during up-merges and for people updating?

I know i was absent in your team meetings so if you decided something else and all agree fell free to also ignore my proposal, though Neos 9.0 needs to have its new `Flow\Route` feature fixed at some point.

mhsdesign: Actually even though the `Route` annotation feature was only lately released the whole annotation flushing existed years before hand via https://github.com/neos/flow-development-collection/pull/2461 and was especially created for this _with_ the broken signal. And as the annotation came only lately we never noticed it being broken - though i recall having to manually flush caches indeed when using the route annotation.

So i a full +1 for renaming the signal to  `afterCompile` as the previous name was never working, but please keep the `AnnotationsCacheFlusher` intact.

kdambekalns: Ultimately this is needed for https://github.com/neos/flow-development-collection/pull/3535 which is targeted at 8.3, that's why the removal is planned.

Repair would thus rather make use of the new signal!?

mhsdesign (changes requested):

Neos.Flow/Classes/Cache/AnnotationsCacheFlusher.php
mhsdesign: Yes in my head that would be the best. Add the new signal here. Use it with the `AnnotationsCacheFlusher`. Everyone is happy. The xdebug pr would just use the new signal then.

dlubitz: TBH I don't want to repair the AnnotationsCacheFlusher as it has never been working. The signal never was dispatched and there is no code which register annotations yet.

I'm fine with keeping the class, so someone can fix it and connect it then to the new signal. But I'd like to get the xdebug feature merged soonish, as this provides a better development experience as the broken AnnotationsCacheFlusher.

mhsdesign: > and there is no code which register annotations yet

not in 8.3 but in 9.0. And yes i will help fix the `AnnotationsCacheFlusher`:)
So if this pr keeps the class alive ill prepare a bugfix once the new signal is merged with the other pr:)

dlubitz:
As discussed, I've changed the PR to replace the signal and fix the AnnotationCacheFlusher.

mhsdesign (approved):
Thank you so much! Will test it with the upmerge to 9.0 before the patch release:)

kitsunet (approved):
<3

Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
```diff
@@ -188,32 +203,36 @@ public function compile()
         $compiledClasses = [];
         foreach ($this->objectManager->getRegisteredClassNames() as $fullOriginalClassNames) {
             foreach ($fullOriginalClassNames as $fullOriginalClassName) {
+                $proxyClassIdentifier = str_replace('\\', '_', $fullOriginalClassName);
+                $class = new \ReflectionClass($fullOriginalClassName);
```
mhsdesign, line 207: during upmerge i only noticed the change of this line.

Now we reflect _more_ than beforehand as also the `else` case of `isset($this->proxyClasses[$fullOriginalClassName])` needs `$class->getFileName();`

I cant say how many classes are now additionally reflected but we should keep this in mind in case we notice the proxy building to be slower? But then again it should be fast enough.

PR #3539, opened 2026-02-20
BUGFIX: Replace outdated postgresql-action with docker command

dlubitz:
The github action `harmon758/postgresql-action@v1` is outdated and doesn't work with the current docker version anymore. So it needs to get replaced.


mhsdesign (approved):
Looks legit?

markusguenther (approved):

PR #3540, opened 2026-02-24
BUGFIX: `#[Flow\Route]` Annotation unset `@subpackage` instead of `null`

mhsdesign:
There is an inconsistency with defined routes which can lead to not properly working resolving, so we herby leave out the `@subpackage` declaration as it should be instead of specifying the key without value.

If a route is defined manually configuration following things are logged in routes:match

```
Name: My.Api :: Configuration
Pattern: api/news

Results:
  @package: My.Api
  @controller: News
  @action: getNews

Matched Controller: ...
```

But the result is different when using the RouteAnnotation, then the `@subpackage` is apparently set which again can confuse the resolving.

```
Name: My.Api :: News :: getNews
Pattern: api/news

Results:
  @package: My.Api
  @subpackage:
  @controller: News
  @action: getNews
  @format: html

Matched Controller: ...
```

I found out that during resolving the `@subpackage` must be set explicitly to an empty string and cannot be left out or set to null.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -152,9 +152,16 @@ jobs:
                   user: 'root'
                   password: 'neos'
                   dbname: 'flow_functional_testing'
+
+              # see Neos\Flow\Tests\Functional\Mvc\RoutingTest
               mvc:
                 routes:
-                  'Neos.Flow': TRUE
+                  "Neos.Flow:TestingAttributes":
+                    providerFactory: Neos\Flow\Mvc\Routing\AttributeRoutesProviderFactory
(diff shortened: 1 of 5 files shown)


mhsdesign:
In case youre wondering which maniac builds uris for routes created via annotation: Its not me but rather the package https://github.com/sitegeist/Sitegeist.SchemeOnYou to generate open api specs and thats legit;)

sorenmalling (approved):

sorenmalling:
Confirmed 👌

PR #3541, opened 2026-03-07
BUGFIX: Explicitly normalize FK constraint names after table renames for MariaDB 12 compatibility

dlubitz:
MariaDB 12.x stopped auto-renaming InnoDB FK constraint names during RENAME TABLE. Two migrations that rename the resource_resource table were followed by downstream migrations expecting the new constraint name — which never materialized on MariaDB 12.x.                                       

To get both versions working we now drop both possible index-name versions directly after the rename and recreate the index with the new - expected name.

For better visibility, consitency and support I also changed the pipeline:
* Add build configurations for PHP 8.5 with MariaDB 12.3 and Postgres 18
* Change default version for MariaDB to latest LTS 11.8

Fixes: #3543

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -183,7 +183,7 @@ jobs:
         #if: env.BEHAT == true
         run: |
           FLOW_CONTEXT=Testing/Behat ./flow doctrine:migrate
-          FLOW_CONTEXT=Testing/Behat ./flow behat:setup && ./flow doctrine:migrationversion --add --version all
+          FLOW_CONTEXT=Testing/Behat ./flow behat:setup
           bin/behat --stop-on-failure -f progress -c Packages/Framework/Neos.Flow/Tests/Behavior/behat.yml.dist
 
       - name: Setup Flow configuration (PGSQL)
@@ -221,7 +221,7 @@ jobs:
         #if: env.BEHAT == true
         run: |
           FLOW_CONTEXT=Testing/Behat ./flow doctrine:migrate
-          FLOW_CONTEXT=Testing/Behat ./flow behat:setup && ./flow doctrine:migrationversion --add --version all
+          FLOW_CONTEXT=Testing/Behat ./flow behat:setup
           bin/behat --stop-on-failure -f progress -c Packages/Framework/Neos.Flow/Tests/Behavior/behat.yml.dist
 
   buildall:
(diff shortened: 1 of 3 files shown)


mhsdesign (changes requested):
Sorry just something about our principles first: https://neos-project.slack.com/archives/C04Q0TC15/p1774082500005349

.github/workflows/build.yml
```diff
@@ -39,6 +39,13 @@ jobs:
             experimental: true
             dependencies: 'highest'
 
+          - php-versions: '8.5'
+            mariadb-versions: '12.2'
+            postgresql-versions: '18-alpine' # skip postgresql setup
```
mhsdesign: this comment ` # skip postgresql setup` is not correct as we _run_ the tests as its not empty string;)

... fixed it, edited the commit

.github/workflows/build.yml
```diff
@@ -175,7 +182,8 @@ jobs:
         if: ${{ matrix.static-analysis == 'no' && matrix.dependencies != 'lowest' }}
         #if: env.BEHAT == true
         run: |
-          FLOW_CONTEXT=Testing/Behat ./flow behat:setup && ./flow doctrine:create && ./flow doctrine:migrationversion --add --version all
+          FLOW_CONTEXT=Testing/Behat ./flow doctrine:migrate
+          FLOW_CONTEXT=Testing/Behat ./flow behat:setup && ./flow doctrine:migrationversion --add --version all
```
mhsdesign: `doctrine:migrationversion --add --version all` is also no longer needed imo when we use proper doctrine:migrate as this marks the versions correctly as migrated.

mhsdesign (approved):
Thanks for also using the CI to proof that this is fixed:

> SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP FOREIGN KEY `typo3_flow3_resource_resource_ibfk_1`; check that it exists

@Sebobo you have the new version of maria db? You can test it locally :D

I added my adjustments and during that _$something_ might have went missing ... or not 😎...

kdambekalns (commented):
👍 by 👀

kdambekalns (approved):

Sebobo:
Thx <3 ! Will check soon. Didn't want to switch back and forth my db versions while working on important stuff ;)

PR #3542, opened 2026-03-12
BUGFIX: Add debug info to exception if backend creation fails

mhsdesign:
we use `setProperties` in the AbstractBackend to map all options to the backend class. Now there might be a setter for that option but the value is not guaranteed to match the type.

Now a type error is thrown which itself doesnt give a hint where something is wrong:

> Neos\Cache\Backend\RedisBackend::setHostname(): Argument #1 ($hostname) must be of type string, false given

I have encountered this many times especially in combination with Caches configured with env variables that are not set and thus `false`.

**Upgrade instructions**

**Review instructions**

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/CacheFactory.php
+++ b/Neos.Cache/Classes/CacheFactory.php
@@ -57,7 +57,11 @@ class CacheFactory implements CacheFactoryInterface
      */
     public function create(string $cacheIdentifier, string $cacheObjectName, string $backendObjectName, array $backendOptions = []): FrontendInterface
     {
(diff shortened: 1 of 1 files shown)


mhsdesign:
I dont like that we always have to catch `Throwable` and not only `Exception` but thats because a `TypeError` is emitted - thats the downside from the strict types in php ... previously one would have used a manual check an thrown an `InvalidArgumentException` ... and with the straight mapping of settings to php objects this goes wrong easily and we do that in so many places not a alone configuration injection ourselves ... right way would be to check via reflection beforehand i guess?

kitsunet (approved):

PR #3547, opened 2026-04-11
FEATURE: Enable Xdebug Path Mapping in Development context per default

dlubitz:
Enables the Xdebug Path Mapping per default in Development context.

See: https://github.com/neos/flow-development-collection/pull/3535


mhsdesign (approved):
THX

kdambekalns (approved):

PR #3559, opened 2026-05-21
TASK: Also read property names from annotations

### LINKED ISSUE #3558: BUG: FlowAnnotationDriver does not support ID attribute

### Current Behavior

Using Doctrine ORM in a model we run into limitations.

The FlowAnnotationDriver does not support reading the `Id` annotation - but only the `@Id` tag

Meaning that a property marked up as this (using other annotations)

```
    #[ORM\Id]
    #[ORM\Column(name: 'id', length: 40)]
    protected ?string $id = null;

```

is not being found in `FlowAnnotationDriver::buildJoinColumnsIfNeeded` as it only looks for `getPropertyNamesByTag`

### Expected Behavior

expected behavior is to use annotation in configuring a property of a model

### Steps To Reproduce

_No response_

### Environment

```markdown
- Flow:
- PHP:
```

### Anything else?

_No response_

sorenmalling:
When collecting id properties for join columns
also read properies annotated with the Id
mapping from Doctrine ORM

resolved: #3558

**Review instructions**

Create a model `A` with a relation to another model `B`.

Before the patch:
The model `B` must have the `@Id` tag to a property.

After the patch:
The model `B` can now use `[ORM\Id]` annotation

The property is now considered a `ID Property`

### DIFF
--- a/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
+++ b/Neos.Flow/Classes/Persistence/Doctrine/Mapping/Driver/FlowAnnotationDriver.php
@@ -522,11 +522,17 @@ class FlowAnnotationDriver implements DoctrineMappingDriverInterface, PointcutFi
         foreach ($joinColumns as &$joinColumn) {
             if ($joinColumn['referencedColumnName'] === null || $joinColumn['referencedColumnName'] === 'id') {
                 if ($direction === self::MAPPING_REGULAR) {
-                    $idProperties = $this->reflectionService->getPropertyNamesByTag($mapping['targetEntity'], 'id');
+                    $idProperties = array_unique(array_merge(
+                        $this->reflectionService->getPropertyNamesByAnnotation($mapping['targetEntity'], ORM\Id::class),
+                        $this->reflectionService->getPropertyNamesByTag($mapping['targetEntity'], 'id')
+                    ));
                     $joinColumnName = $this->buildJoinTableColumnName($mapping['targetEntity']);
                 } else {
                     $className = $this->getUnproxiedClassName($property->getDeclaringClass()->getName());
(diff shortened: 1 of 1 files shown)


mhsdesign:
Hej, thanks for your fix - Im not deep enough in that but i think even if i were i would have following questions:

- i though we had `Neos\Flow\Annotations\Identity` as replacement for `ORM\ID`
- i cannot really fully grasp the implications without a new tests ^^
- isnt this a bugfix and should target the lowest version?

sorenmalling:
> - i though we had `Neos\Flow\Annotations\Identity` as replacement for `ORM\ID`
> 

The used annotation is not really relevant here. F
If you use annotation instead of var tags for that it would not be read either

> - i cannot really fully grasp the implications without a new tests ^^

Is it a wish or requirement? Where should it be placed?

> - isnt this a bugfix and should target the lowest version?

Agree, its more of s bugfix

mhsdesign:
Hi, i found that there are several tests already using `@ORM\Id` annotation but there have been none for `@id` or `#[ORM\ID]`

The reason `@ORM\Id` works even if we are only reading tags, is that the namespace is trimmed and we could have been writing `@Nudelsuppe\Id`  see test https://github.com/neos/flow-development-collection/pull/3520/changes/0323dfbd059e8fbecf6d390ba5d27939c5be9b82

id say add extend one in `Neos.Flow/Tests/Functional/Persistence/Fixtures` and copy a `@ORM\Id` definition and add a `#[ORM\ID]`  :)

And this now definitely qualifies as bugfix to the lowest version which attempts to actually provide orm attributes in flow :)

mficzel:
> And this now definitely qualifies as bugfix to the lowest version which attempts to actually provide orm attributes in flow :) 

That would be 8.4 i think

PR #3560, opened 2026-05-25
BUGFIX: Retain original PDO exceptions in Query

kitsunet:
This helps debugging, otherwise important debugging information can get lost with the original
exception.


kdambekalns (approved):
👏

PR #3562, opened 2026-05-26
TASK: fix xdebug path mapping file generation with multiple contexts

t-heuser:
Fixes two issues with the Xdebug path mapping file introduced in #3535:

1. **Multiple contexts overwrote each other.** Compiling in `Development` and then in `Testing` (or vice versa) produced a single `flow.map` containing only the last context's proxies, so breakpoints in proxy classes broke in whichever context was compiled second. The map file is now written per-context as `flow-<Context>.map` (e.g. `flow-Development.map`, `flow-Testing.map`), so each context can be debugged independently.

2. **Map file lost across container restarts.** The map directory often lives only in an short lasting container layer, while the proxy class cache is mounted from the host. After a restart the proxy cache was still warm — so no `afterCompile` signal fired — but the map file was gone, leaving Xdebug without a path mapping until the cache was flushed. The compiled-class input is now also serialized next to the proxy class cache as `.xdebug-pathmap-data.serialized`, and the map file is regenerated from that data on boot whenever it's missing.

related: #3535 

**Upgrade instructions**

If you previously pointed Xdebug at `flow.map`, update your Xdebug configuration to load the per-context file, e.g. `flow-Development.map`. The old `flow.map` is no longer produced.

**Review instructions**

To verify the multi-context fix:
1. Enable Xdebug path mapping as described in the [release notes from Flow 8.4.3](https://flowframework.readthedocs.io/en/8.4/TheDefinitiveGuide/PartV/ChangeLogs/843.html#id8)
2. Run `./flow flow:cache:flush --force` and then trigger a compile in `FLOW_CONTEXT=Development` (e.g. open the app).
3. Trigger a compile in `FLOW_CONTEXT=Testing` (e.g. run the test suite).
4. Confirm both `Data/Temporary/.../<...>/flow-Development.map` and `flow-Testing.map` exist and contain entries for their respective proxy classes.

To verify the restart-recovery fix:
1. With a warm proxy cache, delete the `.xdebug/` map directory (simulating a container restart).
2. Boot Flow again (any CLI command). The map file should be re-created from the persisted `.xdebug-pathmap-data.serialized` without recompiling proxies.

Note on the `Package.php` wiring: `XdebugPathMappingBuilder` lives under `Neos\Flow\ObjectManagement`, which is on `Compiler::$excludedSubPackages` and therefore has no DI proxy at runtime. The `afterInvokeStep` handler resolves the instance via `ObjectManager::get()` and invokes the inject* setters manually — this mirrors the pattern used for other early-boot services and is required for the call to work.
I'm not a big fan of how that looks, if anyone has another idea how to handle that, I'd be happy to change that part of the code.

**Checklist**

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
@@ -15,6 +15,7 @@ namespace Neos\Flow\ObjectManagement\Proxy;
 use Neos\Cache\Backend\SimpleFileBackend;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Cache\CacheManager;
+use Neos\Flow\Core\Bootstrap;
 use Neos\Utility\Files;
 
 /**
@@ -22,11 +23,19 @@ use Neos\Utility\Files;
  */
 class XdebugPathMappingBuilder
 {
+    private const PERSISTED_DATA_FILENAME = '.xdebug-pathmap-data.serialized';
+
     /**
      * @Flow\Inject
      * @var CacheManager
      */
     protected $cacheManager;
+
+    /**
+     * @Flow\Inject
+     * @var Bootstrap
+     */
+    protected $bootstrap;
 
     /**
      * @var array
@@ -47,6 +56,11 @@ class XdebugPathMappingBuilder
         $this->cacheManager = $cacheManager;
     }
 
+    public function injectBootstrap(Bootstrap $bootstrap): void
+    {
+        $this->bootstrap = $bootstrap;
+    }
+
     /**
      * @param array<string, array{path: string, proxyClassIdentifier: string}> $compiledClasses
      * @return void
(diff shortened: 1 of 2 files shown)


mhsdesign:
Hi thanks for this change. Huhm thats definitely an oversight that we forgot development context :O Thats unexpected / unfortunate ^^
So i agree on the idea to emit multiple maps.

But regarding what you said 2. i think this pr introduces now a lot of complexity which we wouldnt have if we just write the proxy map into the `Data` folder next-to the proxy objects ... that also solves problem 1 as in `Data` we separate by context the caches. I think there is no requirement for the map to be in the root .xdebug directory.

CC @dlubitz

dlubitz:
Thank you @t-heuser. From my POV this adds more complexity with less advantage.
1) I don't see the point for multiple context support. As this is a dev tool you would probably only use it in development context. But maybe we should clear the map on context switches or only build the map for a configured (development) context.

2) This is rather a infrastructure problem than an application problem.

Nevertheless, Derik has already a feature request, to allow multiple directories for the map sources. This would solve both issues. I think we should rather wait for this.
https://bugs.xdebug.org/view.php?id=2399


>  I think there is no requirement for the map to be in the root .xdebug directory.

@mhsdesign Actually it is:

> Xdebug will check for the .xdebug directory in the grand-parent directory of the file that is being requested first, then in the parent directory, and then in the directory of the file that is being requested.

https://xdebug.org/funding/001-native-path-mapping#proposal

t-heuser:
Hey, thank you for your answers :)
1. Number 1 is a real problem that I have.
In our docker container entrypoint.sh we execute some flow commands. We do that without giving any flow context (so it uses the standard Development one) but also some flow commands like this:
`FLOW_CONTEXT=Development/Docker/Testing ./flow site:import --packageKey Vendor.AwesomePackage`
This leads to the overwrite of the flow.map file and its generated for the Testing context. It will not be re-generated for my normal Development context in following requests as the Proxy classes are already compiled. I cannot use xdebug then and I have to manually clean the Data/Temporary directory and then have to trigger the Development context compilation again.

And also if I want to debug something with xdebug in another context besides my normal Development context (so something in my tests) then I would need the flow.map for that context.
Yes, I could also clear the Data/Temporary directory, trigger a compilation again but I think this will lead to confusion, especially as nothing of the behavior is yet documented.

2. I see the point that this is an infrastructure problem. Just had this bothering me and wanted to fix it for everyone with the same setup :D
But yeah, then that part if definitely obsolete.

One question: How would the permit of multiple directories for the map sources solve that problem? Because we could then write the proxy map directly to the Data directory?

mhsdesign:
> One question: How would the permit of multiple directories for the map sources solve that problem? Because we could then write the proxy map directly to the Data directory?

Yes that was the idea but it seems xdebug would not detect if we write the map to `Data/Temporary/...` ... though it was mentioned that there is a function .. maybe this works

> An application can additionally configure path maps by calling xdebug_set_source_map() with a file name as argument. Rules in the parsed file(s) override earlier path mappings of the same type.

> The file scanning based approach is necessary so that the [fileurl element](https://xdebug.org/docs/dbgp#connection-initialization) in the debug initiation request can be mapped, as when an application gets to call xdebug_set_source_map() during script execution, it is already too late to apply that mapping to the debug initiation.

I dont know ^^

t-heuser:
Okay, so what do we do with this PR?
Close it and wait for https://bugs.xdebug.org/view.php?id=2399?
Or merge my solution for problem number 1 and remove the solution for problem number 2 (before merging)?
Or something else? :)

mhsdesign:
Thanks for pushing this.

Im a bit undecided and ill cant look to deep into this now because of the neos con:D
Maybe ask in #flow-core-dev or bring this topic to our next dev meeting ... at the sprint next week!

dlubitz:
I'm working on a less complex solution for "multi-context" support. Will provide the PR later.

And as I already stated out, I think the missmatch of cache and mapping is rather a infrastructure issue. I happy to fix this as soon as xdebug supports the configuration of folders.

dlubitz:
Please check this PR: https://github.com/neos/flow-development-collection/pull/3568

I would like to close this PR here in favor of #3568

kdambekalns:
And a thank you to @t-heuser for opening that PR! Even if it didn't go in! 🚀

Issue #3543, opened 2026-03-21
BUG: Doctrine migration is broken with MariaDB 12


dlubitz:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When running doctrine migrations with an empty database on MariaDB 12.2.2, the migration fails.

```
./flow doctrine:migrate
An exception occurred while executing 'ALTER TABLE typo3_flow3_resource_resource DROP FOREIGN KEY typo3_flow3_resource_resource_ibfk_1':

SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP FOREIGN KEY `typo3_flow3_resource_resource_ibfk_1`; check that it exists

The exception details have been logged to the Flow system log.
Exception [0] in line 182 of /home/m/Sites/neos-example/Packages/Libraries/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php: An exception occurred while executing 'ALTER TABLE typo3_flow3_resource_resource DROP FOREIGN KEY typo3_flow3_resource_resource_ibfk_1':

SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP FOREIGN KEY `typo3_flow3_resource_resource_ibfk_1`; check that it exists - See also: 20260307153136d90ae5.txt
```

See on slack: https://neos-project.slack.com/archives/C050C8FEK/p1772897763683099

See also: https://github.com/neos/flow-development-collection/pull/3541

I assume this also effects Flow 8.x

### Expected Behavior

Migration is running without any erros

### Steps To Reproduce

* New Flow/Neos Setup
* run `./flow doctrine:migrate`

### Environment

```markdown
- Flow: 9.1 
- PHP: 8.5
```

PR #3565, opened 2026-05-31
BUGFIX: Check foreign keys explicit before removal

dlubitz:
After fixing #3543 we introduced an incompatibility with MySQL 8.x. This replaces the proprietary "IF EXISTS" with explit check of FK existence to be compatible with both databases.

It also added the MySQL database to the ci pipeline.


kitsunet (approved):

kdambekalns (approved):
👍 by 👀

PR #3566, opened 2026-06-05
TASK: Overhaul CI jobs to support multiple databases and parallel execution

dlubitz:
This splits the build job in different actions, do have reusable actions and allow run multiple independent jobs in parallel. Also I added the support for defining the database used as service per job.

This allows us running matrix jobs for all supported PHP Versions in combination with supported database systems and versions.

I also updated the deprecated checkout actions.


kdambekalns:
Awesome! 🫶

But…

- Why do some tests seems to run twice or even thrice with the same setup? MySQL version not being shown in he name?
- Do we really need / want to test from PHP 8.0 upwards?

dlubitz:
> Why do some tests seems to run twice or even thrice with the same setup? MySQL version not being shown in he name?

I already changed that.

> Do we really need / want to test from PHP 8.0 upwards?

For 8.3 yes, as we support that range. I will reduce the set in higher Flow versions.

kitsunet (approved):

PR #3567, opened 2026-06-05
TASK: Harmonize semantics of Neos\Utility\TypeHandling with PHP 8 typ…

### LINKED ISSUE #2945: TASK: Harmonize semantics of `Neos\Utility\TypeHandling` with PHP 8 type system semantics
I suggest to adjust the `TypeHandling` class to operate closer to PHP's type system definitions (as per https://www.php.net/manual/en/language.types.type-system.php). 

The idea would be to let `TypeHandling` provide the following new methods:

* `isScalarType`
  * covers `bool`, `int`, `float` and `string`
* `isBuiltInType`
  * covers `null`, `array`, `object`, `resource`, `never`, `void`, `self`, `parent` and `static`
  * falls back to `isScalarType`
* `isLiteralType`
  * Same as it is right now in this PR
* `isCallableType`
  * covers `callable`
* `isBuiltInTypeAlias`
  * covers `mixed` and `iterable`
* `isUserDefinedType`
  * calculated from: `!isBuiltInType && !isLiteralType && !isCallableType && !isBuiltInTypeAlias`
  * can be used in `ReflectionService` to determine whether the type name needs to be prefixed with `\` or not

For backwards compatibility `isLiteral` and `isSimpleType` need to continue to work, but can use the new methods. Given my suggestion, `isLiteral` and `isSimpleType` should (probably) also be deprecated.

_Originally posted by @grebaldi in https://github.com/neos/flow-development-collection/pull/2944#discussion_r1051429432_

vcg-development:
FIXES: #2945

### DIFF
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -28,7 +28,7 @@ abstract class TypeHandling
     /**
      * A type pattern to detect literal types.
      */
-    const LITERAL_TYPE_PATTERN = '/^(?:integer|int|float|double|boolean|bool|string)$/';
+    const LITERAL_TYPE_PATTERN = '/^(?:integer|int|float|double|boolean|bool|string|true|false|null)$/';
 
     /**
      * A type pattern to detect scalar types.
--- a/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
+++ b/Neos.Utility.ObjectHandling/Classes/TypeHandling.php
@@ -31,6 +31,21 @@ abstract class TypeHandling
     const LITERAL_TYPE_PATTERN = '/^(?:integer|int|float|double|boolean|bool|string)$/';
 
     /**
+     * A type pattern to detect scalar types.
+     */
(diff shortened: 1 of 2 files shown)


kdambekalns:
This looks like the wrong target branch was selected – too many commits in the PR.

Neos.Utility.ObjectHandling/Classes/TypeHandling.php
```diff
@@ -30,6 +30,21 @@ abstract class TypeHandling
...
+    /**
+     * @var array<string>
+     */
+    protected static $builtInTypes = ['null', 'array', 'object', 'resource', 'never', 'void', 'self', 'parent', 'static'];
+
+    /**
+     * @var array<string>
+     */
+    protected static $builtInTypeAliases = ['mixed', 'iterable'];
+
```
dlubitz, line 47: I wonder about the mix here between regex and array configuration.

vcg-development:
I think I am not informed enough to decide how the code should function here. Could someone else please take over this PR that has more theoretical knowledge of how php works here and knows how the reflections should be handled here.

PR #3568, opened 2026-06-05
BUGFIX: Context aware path mapping for xdebug

dlubitz:
This PR generates the mapping files per FlowContext. This ensures the mapping works even after context switching.

Replaces #3562

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
@@ -52,7 +52,7 @@ class XdebugPathMappingBuilder
      * @param string|null $flowContextName
      * @return void
      */
-    public function buildFromCompiledClasses(array $compiledClasses, ?string $flowContextName = null): void
+    public function buildFromCompiledClasses(array $compiledClasses, string $flowContextName): void
     {
         if ($compiledClasses === []) {
             return;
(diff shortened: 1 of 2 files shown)


Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
```diff
@@ -49,9 +49,10 @@ public function injectCacheManager(CacheManager $cacheManager): void
 
     /**
      * @param array<string, array{path: string, proxyClassIdentifier: string}> $compiledClasses
+     * @param string|null $flowContextName
      * @return void
      */
-    public function buildFromCompiledClasses(array $compiledClasses): void
+    public function buildFromCompiledClasses(array $compiledClasses, ?string $flowContextName = null): void
```
kdambekalns:
```suggestion
    public function buildFromCompiledClasses(array $compiledClasses, string $flowContextName): void
```
since `$bootstrap->getContext()` never returns `null`.

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
```diff
@@ -89,8 +90,9 @@ public function buildFromCompiledClasses(array $compiledClasses): void
         $mappingFileLines[] = '';
 
         Files::createDirectoryRecursively($this->getXdebugMappingFilePath());
+        $mappingFileName = $flowContextName !== null ? sprintf("%s.flow.map", str_replace("/", "_", $flowContextName)) : 'flow.map';
```
kdambekalns:
```suggestion
        $mappingFileName = sprintf("%s.flow.map", str_replace("/", "_", $flowContextName));
```
since `$bootstrap->getContext()` never returns `null`.

Neos.Flow/Classes/ObjectManagement/Proxy/XdebugPathMappingBuilder.php
dlubitz: Yes, but I wanted to make this parameter optional, to not add a breaking change in a bugfix.

kdambekalns: That is very honorable, but… no interface, no `@api`, … I'd be fine with "breaking" it.

dlubitz:
@kdambekalns Is changed.

kdambekalns (approved):

mficzel (approved):
fine by 👀  ... makes sense to me

PR #3571, opened 2026-06-05
TASK: Use migrations in pipeline and fixed migrations for MariaDB 12.x

dlubitz:
**What I did**

Some old Flow migrations dropped foreign keys by their hardcoded auto-generated names (e.g. `..._ibfk_1`). These names are no longer reliable: on MariaDB 12+, renaming a table no longer renames its auto-generated foreign key/index names along with it, so the constraint names the migrations expect may not exist and the migrations fail.

Instead of relying on hardcoded constraint names, the affected migrations now look up the actual foreign keys on the table via the schema manager and drop the one defined on the relevant column. This works regardless of what the constraint is actually named, making the migrations compatible with MySQL, MariaDB < 12 and MariaDB ≥ 12 alike.

**Why the CI change**

Until now this breakage went unnoticed because the pipeline never executed the Doctrine migrations. The build workflow now runs `./flow doctrine:migrate` before the functional tests, so the full migration chain is verified on every build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -156,6 +156,9 @@ jobs:
           working-directory: ${{ env.FLOW_DIST_FOLDER }}
           flow-context: ${{ env.FLOW_CONTEXT }}
 
+      - name: Run migrations to test if they work
+        run: FLOW_CONTEXT=Testing ./flow doctrine:migrate
+
       - name: Run functional tests
         run: composer test-func -- --verbose
 
--- a/Neos.Flow/Migrations/Mysql/Version20110923125535.php
+++ b/Neos.Flow/Migrations/Mysql/Version20110923125535.php
@@ -24,7 +24,11 @@ class Version20110923125535 extends AbstractMigration
         $this->addSql("CREATE INDEX IDX_B4D45B323CB65D1 ON typo3_flow3_resource_resource (resourcepointer)");
 
         if ($this->isPartyPackageInstalled()) {
-            $this->addSql("ALTER TABLE typo3_flow3_security_account DROP FOREIGN KEY typo3_flow3_security_account_ibfk_1");
+            foreach ($this->sm->listTableForeignKeys('typo3_flow3_security_account') as $foreignKey) {
+                if (in_array('party_abstractparty', array_map('strtolower', $foreignKey->getLocalColumns()), true)) {
+                    $this->addSql("ALTER TABLE typo3_flow3_security_account DROP FOREIGN KEY " . $foreignKey->getName());
+                }
+            }
             $this->addSql("DROP INDEX IDX_44D0753B38110E12 ON typo3_flow3_security_account");
         }
         $this->addSql("ALTER TABLE typo3_flow3_security_account CHANGE party_abstractparty party VARCHAR(40) DEFAULT NULL");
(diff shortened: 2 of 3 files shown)


Neos.Flow/Migrations/Mysql/Version20110923125535.php
```diff
@@ -24,7 +24,11 @@ public function up(Schema $schema): void
         $this->addSql("CREATE INDEX IDX_B4D45B323CB65D1 ON typo3_flow3_resource_resource (resourcepointer)");
 
         if ($this->isPartyPackageInstalled()) {
-            $this->addSql("ALTER TABLE typo3_flow3_security_account DROP FOREIGN KEY typo3_flow3_security_account_ibfk_1");
+            foreach ($this->sm->listTableForeignKeys('typo3_flow3_security_account') as $foreignKey) {
+                if (in_array('party_abstractparty', array_map('strtolower', $foreignKey->getLocalColumns()), true)) {
+                    $this->addSql("ALTER TABLE typo3_flow3_security_account DROP FOREIGN KEY " . $foreignKey->getName());
+                }
+            }
```
bwaidelich, line 31: @dlubitz can you provide some context? why is this needed?

bwaidelich (commented):
Thanks for the description.
So I guess we accept the fact that a down migration would not re-add the FK constraints, right?

dlubitz:
Yes and no. If you only go down one version, this will not be an issue, as the table has not been renamed in meanwhile.

But if you would go down many versions until the renaming, this might have an effect. But TBH, this shouldn't be a valid use case.

AND target should be anyways to get rid of these old old migrations by compression.

bwaidelich (approved):
I agree, thanks!
+1 by reading

.github/workflows/build.yml
```diff
@@ -156,6 +156,9 @@ jobs:
           working-directory: ${{ env.FLOW_DIST_FOLDER }}
           flow-context: ${{ env.FLOW_CONTEXT }}
 
+      - name: Run migrations to test if they work
+        run: FLOW_CONTEXT=Testing ./flow doctrine:migrate
```
mhsdesign, line 160: +1 def makes sense to include that ... even though the functional test will use sqlite and thus not use the prepared database ...

Neos.Flow/Migrations/Mysql/Version20110923125535.php
```diff
@@ -24,7 +24,11 @@ public function up(Schema $schema): void
         $this->addSql("CREATE INDEX IDX_B4D45B323CB65D1 ON typo3_flow3_resource_resource (resourcepointer)");
 
         if ($this->isPartyPackageInstalled()) {
-            $this->addSql("ALTER TABLE typo3_flow3_security_account DROP FOREIGN KEY typo3_flow3_security_account_ibfk_1");
+            foreach ($this->sm->listTableForeignKeys('typo3_flow3_security_account') as $foreignKey) {
```
mhsdesign, line 27: ill find these migrations super hard to read now ... i think we discussed multiple times to have a merged migration state with Neos 10.0 https://github.com/neos/neos-development-collection/issues/5656 can you say if we also need to use these kind of checks when the migrations are compacted or is this then an issue of the past?

TBH i dont understand why we not just define the names of the foreign keys during the creation and thus never have the problem of anything being "random" or autogenerated...

dlubitz: This question is to late. As we already have this situation now and we need to handle that.

mhsdesign: Fully understood. I meant i am _more_ willing to accept this state if we know its definitely just $temporary

mhsdesign: but i dont understand the problem well enough still because

> why we not just define the names of the foreign keys during the creation and thus never have the problem of anything being "random" or autogenerated...

PR #3573, opened 2026-06-06
BUGFIX: Always serialize the RelatedEntitesContainer in proxies

kitsunet:
fixes regression from #3510


so-grimm (approved):
Don't entirely understand how/why this fixes the issue, but tested it locally and it does seem to fix the issue.

PR #3575, opened 2026-06-07
BUGFIX: Reuse singleton instance when requested by interface

kitsunet:
Regression from #3510 

When a singleton was first requested by its implementation class name, subsequent requests by its interface name would create a new instance instead of reusing the existing one.

The ObjectManager now detects this case and returns the already-existing instance, preserving singleton semantics for both class name and interface name lookups.

Also removes the restriction that prevented building proxy classes for classes implementing interfaces with a constructor declaration, and guards ProxyClassBuilder's parent-advice lookup against base classes without a parent.


kitsunet:
Bit more comprehensive bugfix but adding the fix and trying to test it then unravelled the other weird and broken cases...

kitsunet:
ok weird test failures... investigating

kitsunet:
ah it's the new test clearing the routes provider I think^^

Neos.Flow/Tests/Functional/ObjectManagement/ObjectManagerTest.php
```diff
@@ -32,6 +34,22 @@ public function ifOnlyOneImplementationExistsGetReturnsTheImplementationByTheSpe
...
+    public function requestingTheImplementationAndThenTheInterfaceWorks()
+    {
+        $this->objectManager->forgetInstance(TestingRoutesProvider::class);
+        $this->objectManager->forgetInstance(RoutesProviderInterface::class);
+        $objectByClassName = $this->objectManager->get(TestingRoutesProvider::class);
+        $objectByInterface = $this->objectManager->get(RoutesProviderInterface::class);
+
+        self::assertInstanceOf(RoutesProviderInterface::class, $objectByInterface);
+        self::assertInstanceOf(RoutesProviderInterface::class, $objectByClassName);
+        self::assertEquals($objectByClassName, $objectByInterface);
```
mhsdesign:
```suggestion
        self::assertSame($objectByClassName, $objectByInterface, sprintf('Instance %d does not equal %d', spl_object_id($objectByClassName), spl_object_id($objectByInterface)));
```

not just equals ... the objects must have the same identity 

otherwise the test still passed without this pr

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -209,6 +209,18 @@ public function get($objectName, ...$constructorArguments): object
             throw new Exception\UnknownObjectException('Object "' . $objectName . '" is not registered.' . $hint, 1264589155);
         }
 
+        // Someone might have requested the implementation class directly, in that case we want to reuse that instance when requesting the object.
+        if (
+            $objectName !== $className
+            && isset($this->objects[$className])
```
mhsdesign, line 215:
```suggestion
```

my IDE says isset can check nested arrays directly

Neos.Flow/Tests/Functional/ObjectManagement/ObjectManagerTest.php
```diff
@@ -32,6 +34,22 @@ public function ifOnlyOneImplementationExistsGetReturnsTheImplementationByTheSpe
 
         self::assertInstanceOf(Fixtures\InterfaceAImplementation::class, $objectByInterface);
         self::assertInstanceOf(Fixtures\InterfaceAImplementation::class, $objectByClassName);
+        self::assertEquals($objectByClassName, $objectByInterface);
```
mhsdesign:
```suggestion
        self::assertSame($objectByClassName, $objectByInterface, sprintf('Instance %d does not equal %d', spl_object_id($objectByClassName), spl_object_id($objectByInterface)));
```

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/InterfaceWithConstructor.php
```diff
@@ -1,17 +0,0 @@
...
- * (c) Contributors of the Neos Project - www.neos.io
- *
- * This package is Open Source Software. For the full copyright and license
- * information, please view the LICENSE file which was distributed with this
- * source code.
- */
-
-interface InterfaceWithConstructor
-{
-    public function __construct(array $someArgument = []);
```
mhsdesign, line 16: wtf this works in php :O

Neos.Flow/Tests/Functional/ObjectManagement/ProxyCompilerTest.php
```diff
@@ -231,20 +227,6 @@ public function attributesArePreserved(): void
         self::assertEquals(Fixtures\ClassWithPhpAttributes::class, $attributes[0]->getArguments()[0]);
     }
 
-    /**
-     * @test
-     * @see https://github.com/neos/flow-development-collection/issues/2554
-     * @throws
-     */
-    public function proxyingClassImplementingInterfacesWithParametrizedConstructorsLeadsToException(): void
```
mhsdesign, line 239: yes the test is weird so +1 removing it but should we add a proper one?

mhsdesign (approved):
i guesss

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
kitsunet: I still need to check if we have objects configuration for the class before checking the scope (scope is implicitly required in this data). Just checking if an instance is already there won't help that. I could check `$this->objects[$className][self::KEY_SCOPE]` first but same same

Neos.Flow/Tests/Functional/ObjectManagement/ProxyCompilerTest.php
kitsunet: Probably, can be separate though. Need to think about what to actually test then.

PR #3576, opened 2026-06-07
BUGFIX: Handle objects without name in argument configuration parsing

kitsunet:
Instead of throwing an exception when an argument configuration contains no name, create a synthetic virtual object name derived from a configuration hash. This handles edge cases encountered with global Objects.yaml configurations.

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationParser.php
+++ b/Neos.Flow/Classes/ObjectManagement/Configuration/ConfigurationParser.php
@@ -169,7 +169,9 @@ readonly class ConfigurationParser
             }
 
             if ($objectName === null) {
-                throw new InvalidObjectConfigurationException('Object configuration for argument "' . $argumentName . '" contains neither object name nor factory object or method name in ' . $parentObjectConfiguration->getConfigurationSourceHint(), 1417431742);
+                // This is a weird object configuration, we create a magic virtual object for it, throwing might be stricter,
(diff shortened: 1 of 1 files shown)


kitsunet:
This PR is mostly for having all tests run in the pipeline and visibility. I would merge immediately if green.

mhsdesign (commented):
Post merge +1 as it was not forbidden to define ad-hoc objects 

basically this works also before, but with this fix it allowed to remove `name: 'Neos.ContentGraph:Virtual.Connection'` which is kinda obsolete either way as this is an inline ad-hoc object and has no name

```
Neos\ContentGraph\DoctrineDbalAdapter\DoctrineDbalContentGraphProjectionFactory:
  scope: singleton
  factoryObjectName: 'Neos\ContentRepositoryRegistry\Infrastructure\GenericObjectFactory'
  arguments:
    1:
      value: 'Neos\ContentGraph\DoctrineDbalAdapter\DoctrineDbalContentGraphProjectionFactory'
    2:
      object:
        name: 'Neos.ContentGraph:Virtual.Connection'
        factoryObjectName: 'Neos\ContentRepository\BehavioralTests\DBALConnectionFactory'
        arguments:
          1:
            value:
              driver: pdo_mysql
              host: 127.0.0.1
              dbname: neos9-behat
              user: neos
              password: '123456'
```


see also slack https://neos-project.slack.com/archives/C04PYL8H3/p1780905034494249?thread_ts=1780856123.461329&cid=C04PYL8H3

PR #3579, opened 2026-06-12
TASK: Sort out use of cleanClassName vs cleanTypeName

kitsunet:


mficzel (approved):
Fine by :eyes ... restores previous behavior that broke from adding checks to cleanClassName

PR #3588, opened 2026-07-22
TASK: PSR12 & PHP-CS-Fixer

mficzel:
The codestyle is adjusted from psr2 to psr12 and the linters are adjusted accordingly.

PHP-CS-Fixer shall replace the StyleCi that caused lots of unexpected test failures. The fixer is included in the `composer lint` command and thus can be executed locally before pushing. The `composer fix` command allows to adjust the coding style to the requirements quickly.
 
also:
- changes the codestyle from psr2 to psr12 
- cs fixer in check mode to the `composer lint` command
- a general `composer fix` command is added that for now only runs the cs fixed
- style ci configuration is removed

**Upgrade instructions**

**Review instructions**

!!! Style CI will fail as we remove the configuration and change the settings. This is intended and will be fixed by removeing style ci after the merge !!!

**Checklist**


mficzel:
close -> reopen

mficzel:
This can be merged imho and afterwards the style ci can be disabled for flow. It will cause minor merge conflicts with the Stan L8  PR but those are kindof trivial.

The only failing test is style ci which cannot be green as it not configured anymore and was configured for psr2 instead of psr12

mhsdesign (approved):
Thanks for giving this a shot!

During reviewing i noticed that the exclusions of the parsers did not work the paths must be relative see (https://github.com/neos/flow-development-collection/pull/3588/changes/d2f5536be292d248cbf75552b3994e5e354f1486)

Also some packages were excluded from php-cs-fixer, i now added them as well: https://github.com/neos/flow-development-collection/pull/3588/changes/7f4c679b88db01f00be70703852f092b1ed742d6

Thus i re-ran the fixer so the parsers are not touched. Doing that i split up the fixing into Two commits to ease reviewing.

Once the commit to sort imports and adds spacing around the php file opening (https://github.com/neos/flow-development-collection/pull/3588/changes/2faa6992f9976fc969c688379e0084c07198638a) this is safe to skip during review as all follows the same pattern

And then there is the commit (https://github.com/neos/flow-development-collection/pull/3588/changes/8a9eb7e6742503604564bf15f00c9e51aa2a94d5) to do the remaining PSR12 changes. I skimmed through them all and they look good.

.composer.json
```diff
@@ -8,9 +8,17 @@
...
+    "lint:phpstan-legacy": "../../bin/phpstan analyse -c phpstan-legacy.neon.dist -v",
+    "lint:phpstan": "../../bin/phpstan analyse -c phpstan.neon.dist -v --memory-limit 256M",
     "lint": [
-      "@lint:phpstan"
+      "@lint:php-cs-fixer",
+      "@lint:phpstan",
+      "@lint:phpstan-legacy"
+    ],
+    "fix:php-cs-fixer": "../../bin/php-cs-fixer fix",
+    "fix": [
```
bwaidelich, line 20: is that a common naming scheme?
IMO it should be clear that this only fixes code _style_ and we should never extend it to fix _implementation_. I usually call these `test:cs` and `test:cs:fix` – We could go for `lint:php-cs-fixer:fix`.

But I'm also fine with keeping it as is

bwaidelich (approved):
A 1700+ files change is not really possible to review, but I skimmed over the PR and could not find any oddities (apart from the new composer script names, but that's mostly a nitpick).

Apart from that, this will probably affect most if not all of the 87(!) open Pull Requests.. I don't think, that there is a way around this, but we should rebase the active ones as fast as possible to avoid a dependency hell

.composer.json
mhsdesign: Yes i also wondered but martin seems to like it for his projects :) Now its uniform in Neos and Flow and Martin wont respond because he's on tour 🚲 so i think lest merge it like this and if we really dislike it adjust it later?

note: ill be using `fix:php-cs-fixer` only so i dont need the `fix` as well but 🤷

mhsdesign: You proposed 2024 https://github.com/neos/neos-development-collection/pull/4949 these ... i could also get familar with them ...

> It makes those checks part of our `composer:lint` script and adds:

* `composer lint:php-cs-fixer` to check them and
* `composer lint:php-cs-fixer:fix` to fix violations
* `composer lint:fix` to fix any linter violations (currently the same)

mficzel: I usually only ever call „composer fix“ when a linter complains. That way every project can adjust what happens behind.

PR #3595, opened 2026-08-03
TASK: Add PHP language feature reference

robertlemke:
This adds a new reference to the appendix of the Definitive Guide which documents the support status of PHP language constructs in proxied classes, separated by subsystem (Dependency Injection, AOP and object serialization) and including the Flow version which introduced support.

The reference also documents known limitations with links to the corresponding issues, so that users can look up whether a specific language feature can be used in classes managed by the object framework.


bwaidelich (approved):
Lovely, thanks!

Issue #3594, opened 2026-08-03
Attributes on method parameters are lost in proxy classes


robertlemke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Attributes on method parameters are not reproduced when a method is rendered into a proxy class. This affects userland attributes as well as PHP's built-in `#[\SensitiveParameter]`:

```php
#[Flow\Scope('singleton')]
class AuthenticationService
{
    public function authenticate(string $username, #[\SensitiveParameter] string $password): bool
    {
        // ...
    }
}
```

If `authenticate()` is advised, the proxy class declares the method without the attribute. As a consequence, reflection on the parameter no longer reports the attribute – and, more importantly, the password is **no longer masked in stack traces**: an exception thrown inside the method exposes the plain text password, although the developer explicitly marked the parameter as sensitive. This turns a proxy building limitation into a security issue for anything that logs stack traces.

Attributes on promoted constructor properties and on class properties are affected in the same way.

### Expected Behavior

Attributes on parameters (including promoted constructor properties) and on properties are reproduced in generated proxy classes, like Flow already does for class and method attributes since #3075 / #3265.

### Steps To Reproduce

Create the class above, advise `authenticate()` with any advice, then:

1. Reflect the parameters of `authenticate()` – `getAttributes()` returns an empty array for `$password`.
2. Let the method throw an exception and inspect the stack trace – the password appears in plain text. Without the advice in place, PHP replaces it with `Object(SensitiveParameterValue)`.

### Environment

```markdown
- Flow: 9.2-dev (9d5c82d0a)
- PHP: 8.4.24
```

### Anything else?

The parameter rendering is currently delegated to `Laminas\Code\Generator\ParameterGenerator`, which has no support for attributes.

Issue #3593, opened 2026-08-03
Classes implementing __serialize() bypass Flow's generated serialization logic


robertlemke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Flow's proxy classes implement dependency re-injection and entity reference handling for serialization through a generated `__sleep()` method (see `ObjectSerializationTrait`). PHP however prefers `__serialize()` / `__unserialize()` over `__sleep()` / `__wakeup()` if they exist.

The proxy class builder only checks for an existing `__sleep()` method in the original class. If a class implements `__serialize()` instead, the generated `__sleep()` is silently ignored: injected properties are serialized along with the object, entity references are not converted into identifiers, and dependencies are not re-injected after deserialization.

Consider the following session-scoped class:

```php
#[Flow\Scope('session')]
class SessionServiceWithSerialize
{
    #[Flow\Inject]
    protected VariadicService $variadicService;

    public string $state = 'initial';

    public function __serialize(): array
    {
        return get_object_vars($this);
    }

    public function __unserialize(array $data): void
    {
        foreach ($data as $propertyName => $value) {
            $this->$propertyName = $value;
        }
    }
}
```

Serializing an instance of this class includes the injected singleton in the serialized data. The same class without `__serialize()` / `__unserialize()` correctly strips the dependency and re-injects it on wakeup.

### Expected Behavior

Flow either supports `__serialize()` / `__unserialize()` in proxied classes (analogous to the existing `__sleep()` handling), or proxy building fails with a meaningful exception explaining that these methods conflict with the generated serialization code.

Independently of this bug, it's worth noting that PHP 8.5 soft-deprecates `__sleep()` and `__wakeup()`, so in the long run the generated serialization code should move to `__serialize()` / `__unserialize()` anyway.

### Steps To Reproduce

See the example above; compare `strlen(serialize($instance))` and the serialized content with and without the `__serialize()` method.

### Environment

```markdown
- Flow: 9.2-dev (9d5c82d0a)
- PHP: 8.4.24
```

Issue #3592, opened 2026-08-03
final public static methods with CompileStatic break in Production context


robertlemke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The proxy compiler comments out the `final` keyword of methods which are overridden in the proxy class. The regular expression used for this only matches the modifier combinations `final public function` and `public final function` – it does not match `final public static function`.

Methods annotated with `Flow\CompileStatic` are overridden in the proxy class when compiling in `Production` context. If such a method is also declared `final`, the `final` keyword remains in the renamed original class and loading the proxy fails:

```
PHP Fatal error: Cannot override final method
Acme\ProxyRepro\StaticCompiledService_Original::compiledConfiguration()
in .../Flow_Object_Classes/Acme_ProxyRepro_StaticCompiledService.php on line 29
```

Since `CompileStatic` results are only compiled in `Production` context, the application works fine in `Development` context and only breaks in `Production` – on the first request which loads the affected class.

### Expected Behavior

The `final` keyword is commented out for all modifier combinations, including `static` methods.

### Steps To Reproduce

Create the following class:

```php
#[Flow\Scope('singleton')]
class StaticCompiledService
{
    #[Flow\CompileStatic]
    final public static function compiledConfiguration(ObjectManagerInterface $objectManager): array
    {
        return ['answer' => 42];
    }
}
```

Run `FLOW_CONTEXT=Production ./flow core:compile` and load the class. In `Development` context everything works as expected.

### Environment

```markdown
- Flow: 9.2-dev (9d5c82d0a)
- PHP: 8.4.24
```

### Anything else?

The regular expression lives in `Compiler::commentOutFinalKeywordForMethods()`. It also requires method names to be at least two characters long and the indentation to match the generated proxy code exactly – a token based implementation would probably be more robust than extending the regular expression.

Issue #3591, opened 2026-08-03
AOP advices on readonly classes produce invalid proxy code


robertlemke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When an advice matches a method of a `readonly` class, the AOP proxy class builder adds its internal untyped properties (`Flow_Aop_Proxy_targetMethodsAndGroupedAdvices` etc.) to the proxy class. Since the proxy of a readonly class must be readonly itself, PHP rejects the generated code:

```
Fatal error: Readonly property Acme\ProxyRepro\ReadonlyGreeter::$Flow_Aop_Proxy_targetMethodsAndGroupedAdvices
must have type in .../Flow_Object_Classes/Acme_ProxyRepro_ReadonlyGreeter.php on line 26
```

Note that `./flow core:compile` finishes without any error – the fatal error only occurs when the generated class is loaded for the first time, which makes this hard to debug.

Even with a type declaration the generated code would not work, because the proxy code writes to these properties outside of the constructor, which readonly properties don't allow.

### Expected Behavior

Advices on readonly classes either work, or – if the internal proxy state cannot be reconciled with readonly semantics – the proxy class builder refuses to advise readonly classes with a meaningful exception at compile time.

### Steps To Reproduce

Create the following class:

```php
final readonly class ReadonlyGreeter
{
    public function __construct(private string $name = 'world')
    {
    }

    public function greet(): string
    {
        return 'Hello ' . $this->name;
    }
}
```

advise `greet()` with any advice, run `./flow core:compile` (which succeeds) and then use the class. You should see the fatal error.

### Environment

```markdown
- Flow: 9.2-dev (9d5c82d0a)
- PHP: 8.4.24
```

### Anything else?

The properties are added in `Aop\Builder\ProxyClassBuilder` without considering the readonly flag of the target class. Support for readonly classes was introduced with #3025 / #3051, but only covers the dependency injection part.

robertlemke:
While adding test coverage for readonly classes (#3596), I found that this bug does not only affect classes which are advised by userland aspects: a readonly class annotated with `#[Flow\Scope('session')]` triggers the same fatal error, because the session scope introduces Flow's lazy loading aspect. In other words, readonly classes currently cannot be session-scoped at all.

Also note that `./flow core:compile` finishes with exit code 0 in this case – the fatal error only occurs when the generated class is loaded for the first time, which makes the problem easy to miss in CI setups which only compile.

[...]

PR #3596, opened 2026-08-03
TASK: Extend test coverage for proxy class building

robertlemke:
This adds functional and unit tests which verify that PHP language constructs work as expected in proxied classes: generator methods, first-class callable syntax, pure intersection and DNF types, typed class constants, constants declared in traits, the Override attribute, readonly properties and classes, new without parentheses and dynamic class constant fetch — covering both dependency injection and AOP advices.

It also adds unit tests for ProxyMethodGenerator, ProxyConstructorGenerator, RelatedEntitiesContainer and ObjectSerializationTrait, which previously had no test coverage at all. The unit tests document the current behavior of the code generators, including a few oddities which should be addressed separately.

Related to #3589, #3590, #3591, #3592, #3593, #3594


Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP84/ClassWithNewWithoutParentheses.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public function describeNewStatic(): string
+    {
+        return new static()->describe();
+    }
+
+    public function describeNewClassName(): string
+    {
+        return new ClassWithNewWithoutParentheses()->describe();
```
mhsdesign, line 54: According to the name of it shouldnt it be?

```suggestion
        return new ClassWithNewWithoutParentheses->describe();
```

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyConstructorGeneratorTest.php
```diff
@@ -0,0 +1,279 @@
...
+     * Constructor property promotion must not be repeated in the proxy constructor, because
+     * the promoted properties are already declared by the original class.
+     *
+     * @test
+     */
+    public function promotedParametersOfTheOriginalConstructorAreRenderedAsPlainParameters(): void
+    {
+        $proxyConstructor = $this->createProxyConstructorFor(ClassWithPublicConstructor::class);
+        $proxyConstructor->addPreParentCallCode('$foo = 1;');
+
```
mhsdesign, line 62: i dont understand that line - maybe its obsolete will try to run the tests without ^^ 

```suggestion

```

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyConstructorGeneratorTest.php
```diff
@@ -0,0 +1,279 @@
...
+class ProxyConstructorGeneratorTest extends UnitTestCase
+{
+    /**
+     * @param class-string $className
+     */
+    private function createProxyConstructorFor(string $className): ProxyConstructorGenerator
+    {
+        return ProxyConstructorGenerator::fromReflection(new MethodReflection($className, '__construct'));
+    }
+
```
mhsdesign, line 37: ha is the idea using the actually real reflected classes and inspecting their source-code?

Though i guess `ProxyConstructorGenerator::fromReflection()` would have been never called in a real flow compilation _with_ the proxied $className reflection but obviously with the not yet proxied class ...

EDIT: I was wrong, we are in `Unit` testing context and thus there are no proxies so we actually test the real proxy generation!

mhsdesign (commented):
Thanks a lot. I had a quick look.

Am i correct that as all tests pass this is very much already the baseline? I could not find commented out tests or skipped ones? So this is really just documenting how it is today?

robertlemke:
> Am i correct that as all tests pass this is very much already the baseline? I could not find commented out tests or skipped ones? So this is really just documenting how it is today?

Exactly, that's the baseline. I created these tests to be well prepared for further bug fixes and refactorings.

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP84/ClassWithNewWithoutParentheses.php
robertlemke: Well, no … the name refers to the parentheses which had to surround the whole expression before PHP 8.4:

```php
(new ClassWithNewWithoutParentheses())->describe()
```

The parentheses of the constructor call itself are still required when chaining, so your suggestion actually produces a parse error:

```                                           
PHP Parse error: syntax error, unexpected token "->", expecting ")"                                                                                                                                                            
```

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyConstructorGeneratorTest.php
robertlemke: That line is what makes the constructor being rendered at all: a proxy constructor (like any proxy method) is only rendered if some pre or post parent call code was added. Otherwise generate() returns an empty string and the assertions on the generated signature would fail.

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyConstructorGeneratorTest.php
robertlemke: Exactly. In the unit testing context no proxy classes exist, so the reflection is based on the original class, just like during an actual compile run.

Neos.Flow/Tests/Functional/ObjectManagement/Fixtures/PHP84/ClassWithNewWithoutParentheses.php
mhsdesign: 🤦 forgot how php works thanks for reminding me

robertlemke:
So, @mhsdesign, is that a +1? ;-)

kitsunet (approved):
Just adopted to phpunit 11. Looks good to me, having a solid test foundation will be good for the future.

PR #3598, opened 2026-08-04
BUGFIX: Call original method in void proxy methods

### LINKED ISSUE #3597: Proxy methods with a void return type can silently skip the original method

### Current Behavior

When proxy building adds only "pre parent call code" to a method – without any post code – and the method declares a `void` or `never` return type, the generated proxy method contains the pre code but **no call to the original method at all**. The original implementation is silently skipped.

The culprit is the last branch in `ProxyMethodGenerator::renderBodyCode()`:

```php
} elseif (!$returnTypeIsVoidOrNever && $callParentMethodCode !== '') {
    $code .= 'return ' . $callParentMethodCode . ";\n";
}
```

The branch for pre *and* post code a few lines above handles the same case correctly by rendering the parent call without a `return` statement – the pre-code-only branch just drops it.

### Expected Behavior

The parent call is rendered (without a `return` statement) for `void` and `never` return types, like the pre-and-post-code branch already does.

### Steps To Reproduce

The behavior is pinned down by the unit test `noParentCallIsRenderedForVoidAndNeverReturnTypesIfOnlyPreParentCallCodeExists` introduced with #3596, which documents the current (incorrect) behavior and should be inverted along with the fix.

A user-facing example would be a proxied class implementing `public function __wakeup(): void` – depending on which code the proxy builders add to that method, the original wakeup logic may never be executed.

### Environment

```markdown
- Flow: 9.2-dev (9d5c82d0a)
- PHP: 8.4.24
```

### Anything else?

_No response_

robertlemke:
When proxy building adds only pre parent call code to a method – without any post code – and the method declares a `void` or `never` return type, the generated proxy method contains the pre code but no call to the original method: the original implementation is silently skipped.

This change renders the parent call (without a return statement, which is not allowed for these return types) like the branch for pre and post parent call code already does.

Note for the upmerge to 9.2: the unit test `noParentCallIsRenderedForVoidAndNeverReturnTypesIfOnlyPreParentCallCodeExists` introduced with #3596 documents the old behavior and needs to be inverted – the `ProxyMethodGeneratorTest` introduced with this change covers the corrected behavior and will conflict with the file of the same name from #3596.

Fixes #3597

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/ProxyMethodGenerator.php
@@ -175,8 +175,12 @@ class ProxyMethodGenerator extends MethodGenerator
             if (!$returnTypeIsVoidOrNever) {
                 $code .= "return \$result;\n";
             }
-        } elseif (!$returnTypeIsVoidOrNever && $callParentMethodCode !== '') {
-            $code .= 'return ' . $callParentMethodCode . ";\n";
+        } elseif ($callParentMethodCode !== '') {
+            if ($returnTypeIsVoidOrNever) {
+                $code .= '    ' . $callParentMethodCode;
+            } else {
+                $code .= 'return ' . $callParentMethodCode . ";\n";
+            }
         }
         return $code;
     }
--- a/Neos.Flow/Tests/Unit/ObjectManagement/Fixture/ClassWithVoidAndNeverMethods.php
+++ b/Neos.Flow/Tests/Unit/ObjectManagement/Fixture/ClassWithVoidAndNeverMethods.php
@@ -1,0 +1,33 @@
+<?php
+namespace Neos\Flow\Tests\Unit\ObjectManagement\Fixture;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+/**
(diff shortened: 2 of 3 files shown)


robertlemke:
Since upmerging is usually done manually, here is what to do when this change is merged upwards into 9.2:

Both this change and #3596 (which went directly into 9.2) introduce a file `Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyMethodGeneratorTest.php`, so the merge from 9.1 into 9.2 will result in an add/add conflict (the merge from 9.0 into 9.1 is conflict-free).

To resolve it:

1. Keep the 9.2 version of the file – it is the more comprehensive one and covers everything the version from this change covers. The fixture `Tests/Unit/ObjectManagement/Fixture/ClassWithVoidAndNeverMethods.php` coming from this change can be discarded as well, it is only used by the discarded test file.
2. In the 9.2 version, replace the test method `noParentCallIsRenderedForVoidAndNeverReturnTypesIfOnlyPreParentCallCodeExists()` – which documents the old, broken behavior, see the note in its doc block – with the following one:

```php
/**
 * @test
 * @dataProvider voidAndNeverReturnTypesDataProvider
 */
public function parentCallWithoutReturnStatementIsRenderedForVoidAndNeverReturnTypesIfOnlyPreParentCallCodeExists(string $methodName): void
{
    $proxyMethod = $this->createProxyMethodFor($methodName);
    $proxyMethod->addPreParentCallCode('$foo = 1;');

    $bodyCode = $proxyMethod->renderBodyCode();

    self::assertStringContainsString('parent::' . $methodName . '($argument);', $bodyCode);
    self::assertStringNotContainsString('return', $bodyCode);
}
```

After that, `composer test-unit` should be green again.

kitsunet (approved):

PR #3599, opened 2026-08-04
BUGFIX: Comment out final keyword of static methods

### LINKED ISSUE #3592: final public static methods with CompileStatic break in Production context

robertlemke:
The regular expression which comments out the final keyword of methods that are overridden in a proxy class did not match methods declared as `final public static function` or similar combinations including the static keyword. As a consequence, a final static method annotated with `Flow\CompileStatic` resulted in a "Cannot override final method" fatal error – but only in Production context, where CompileStatic methods are compiled into the proxy class.

The new expression covers all combinations of public, protected, static and final, methods returning by reference and method names consisting of a single character.

Fixes #3592

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
+++ b/Neos.Flow/Classes/ObjectManagement/Proxy/Compiler.php
@@ -503,12 +503,12 @@ return ' . var_export($this->storedProxyClasses, true) . ';';
     {
         // comment out "final" keyword, if the method is final and if it is advised (= part of the $proxyClassCode)
         // Note: Method name regex according to http://php.net/manual/en/language.oop5.basic.php
-        $classCode = preg_replace_callback('/^(\s*)((public|protected)\s+)?final(\s+(public|protected))?(\s+function\s+)([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+\s*\()/m', static function ($matches) use ($proxyClassCode) {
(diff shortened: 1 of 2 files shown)


kdambekalns (approved):
👍 by 👀

PR #3600, opened 2026-08-04
BUGFIX: Unpack variadic parameters in proxies

### LINKED ISSUE #3589: Advised methods with variadic parameters fail with a TypeError

### Current Behavior

When an advice is applied to a method with a variadic parameter, calling that method fails with a fatal error, because the generated proxy code passes the collected arguments as a single array instead of unpacking them:

```
Acme\ProxyRepro\VariadicService::sum(): Argument #1 must be of type int, array given,
called in .../Neos.Flow/Classes/Aop/AdvicesTrait.php on line 59
```

The generated proxy method contains a call like this:

```php
$result = parent::sum($numbers);
```

where it should be `parent::sum(...$numbers);`. The same pattern is generated by `ProxyMethodGenerator::buildCallParentMethodCode()`, so all proxied methods with variadic parameters are affected, not only advised ones. Proxy constructors are fine, because they forward via `...func_get_args()`.

### Expected Behavior

Methods with variadic parameters can be advised like any other method.

### Steps To Reproduce

Create the following class:

```php
#[Flow\Scope('singleton')]
class VariadicService
{
    public function sum(int ...$numbers): int
    {
        return array_sum($numbers);
    }
}
```

and an aspect which advises the method:

```php
#[Flow\Aspect]
class LoggingAspect
{
    #[Flow\Before('method(Acme\Repro\VariadicService->sum())')]
    public function beforeSum(JoinPointInterface $joinPoint): void
    {
    }
}
```

Then call `$variadicService->sum(1, 2, 3)` and you should see the fatal error.

### Environment

```markdown
- Flow: 9.2-dev (9d5c82d0a)
- PHP: 8.4.24
```

### Anything else?

The unpacking is missing in `ProxyMethodGenerator::buildCallParentMethodCode()` as well as in the method arguments handling of the AOP interceptor builders. Older Flow versions are affected as well.

robertlemke:
A generated proxy method passes the parameters of the original method on to the parent method one by one. For a variadic parameter this means that the collected arguments are handed over as a single array, which results in a TypeError as soon as a proxied or advised method declares a variadic parameter.

The generated parent call now unpacks variadic parameters again. The same applies to the invocation of a join point at the end of an advice chain: a join point stores the arguments of a variadic parameter as one array, mapped to the name of that parameter, so that array needs to be unpacked before the method is called with the arguments taken from the join point.

Fixes #3589

### DIFF
--- a/Neos.Flow/Tests/Functional/Aop/VariadicParametersTest.php
+++ b/Neos.Flow/Tests/Functional/Aop/VariadicParametersTest.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Neos\Flow\Tests\Functional\Aop;
+
+/*
+ * This file is part of the Neos.Flow package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Flow\Tests\Functional\Aop\Fixtures\TargetClassWithVariadicParameters;
+use Neos\Flow\Tests\FunctionalTestCase;
+use PHPUnit\Framework\Attributes\Test;
+
+/**
+ * Testcase for advices on methods with variadic parameters
+ */
+class VariadicParametersTest extends FunctionalTestCase
+{
+    #[Test]
+    public function argumentsOfAVariadicParameterArePassedOnToTheAdvisedMethod(): void
+    {
+        $targetClass = new TargetClassWithVariadicParameters();
+        self::assertSame(6, $targetClass->sum(1, 2, 3));
+    }
+
+    #[Test]
+    public function anEmptyVariadicParameterIsPassedOnToTheAdvisedMethod(): void
+    {
+        $targetClass = new TargetClassWithVariadicParameters();
+        self::assertSame(0, $targetClass->sum());
+    }
+
+    #[Test]
+    public function joinPointContainsTheArgumentsOfAVariadicParameterAsArray(): void
+    {
+        $targetClass = new TargetClassWithVariadicParameters();
+        $targetClass->sum(1, 2, 3);
+
+        self::assertSame(['numbers' => [1, 2, 3]], $targetClass->beforeAdviceArguments);
+        self::assertSame(['numbers' => [1, 2, 3]], $targetClass->aroundAdviceArguments);
+    }
+
+    #[Test]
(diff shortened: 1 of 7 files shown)


Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyMethodVariadicsTest.php
```diff
@@ -0,0 +1,90 @@
...
+                'methodName' => 'collect',
+                'expectedArgumentsCode' => '$collection, ...$items',
+                'expectedSignatureCode' => 'array &$collection, string ... $items',
+            ],
+        ];
+    }
+
+    /**
+     * @test
+     * @dataProvider variadicMethodsDataProvider
```
kdambekalns: These must be attributes now…

Neos.Flow/Tests/Unit/ObjectManagement/Proxy/ProxyMethodVariadicsTest.php
```diff
@@ -0,0 +1,90 @@
...
+use Neos\Flow\ObjectManagement\Proxy\ProxyMethodGenerator;
+use Neos\Flow\Tests\Unit\ObjectManagement\Fixture\ClassWithVariadicMethods;
+use Neos\Flow\Tests\UnitTestCase;
+
+/**
+ * Tests for the handling of variadic parameters in generated proxy methods
+ */
+class ProxyMethodVariadicsTest extends UnitTestCase
+{
+    public function variadicMethodsDataProvider(): array
```
kdambekalns: Data providers must be static (now).

Neos.Flow/Tests/Functional/Aop/VariadicParametersTest.php
```diff
@@ -0,0 +1,73 @@
...
+use Neos\Flow\Tests\Functional\Aop\Fixtures\TargetClassWithVariadicParameters;
+use Neos\Flow\Tests\FunctionalTestCase;
+
+/**
+ * Testcase for advices on methods with variadic parameters
+ */
+class VariadicParametersTest extends FunctionalTestCase
+{
+    /**
+     * @test
```
kdambekalns: Must be attributes (now).

kdambekalns (changes requested):

kdambekalns:
That last test run is three weeks old, it would not pass today. 😬

kdambekalns (approved):
👍 by 👀

PR #3602, opened 2026-08-12
BUGFIX: Do not restart cache iteration when entries change

kdambekalns:
The iterable cache backends reset their shared iterator whenever an entry was written or removed. As a result a running iteration started over from the beginning after every change: entries were visited repeatedly and, with the file backends, some were not visited at all.

SessionManager::collectGarbage() removes entries while iterating the session metadata cache (backed by FileBackend by default) and was affected by this.

Instead of resetting the iterator on write, rewind() now always builds a fresh one. That still yields an up-to-date view on every rewind - which is what the reset was meant to achieve - while leaving a running iteration alone.

**Checklist**

### DIFF
--- a/Neos.Cache/Classes/Backend/ApcuBackend.php
+++ b/Neos.Cache/Classes/Backend/ApcuBackend.php
@@ -133,8 +133,6 @@ class ApcuBackend extends IndependentAbstractBackend implements TaggableBackendI
         } else {
             throw new Exception('Could not set value.', 1232986877);
         }
-
-        $this->cacheEntriesIterator = null;
     }
 
     /**
@@ -177,7 +175,6 @@ class ApcuBackend extends IndependentAbstractBackend implements TaggableBackendI
     public function remove(string $entryIdentifier): bool
     {
         $this->removeIdentifierFromAllTags($entryIdentifier);
-        $this->cacheEntriesIterator = null;
         return apcu_delete($this->getPrefixedIdentifier($entryIdentifier));
     }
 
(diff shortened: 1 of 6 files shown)


kdambekalns:
A review here would be nice…

robertlemke:
Thanks, good catch! Now that you discovered it, I can think of a bunch of situations where the old behavior could cause trouble.

One gap though: `RedisBackend` has the same mechanism and isn't covered here. `remove()` and `flush()` still set `$this->entryIterator = null`, and `rewind()` only calls `$this->getEntryIterator()->rewind()`. So when an entry is removed during the iteration, the next `next()` rebuilds the iterator at position 0 and then advances to 1, i.e. the same restart.

On top of that, the doc block you added now promises that `rewind()` always yields an up-to-date view, which Redis does not do yet: it rewinds the *existing* iterator.

That matters for the very case you're fixing: As you know, `Flow_Session_MetaData` is commonly backed by Redis in multi-node setups, and `SessionManager::collectGarbage()` removes while iterating.

Maybe add the same thing for Redis?
  
```php
  public function rewind(): void
  {
      $this->entryIterator = null;
      $this->getEntryIterator()->rewind();
  }
```

And drop the two $this->entryIterator = null; lines in remove() and flush().

A final nitpick: The existing iterationResetsWhenDataIsSet / iterationResetsWhenDataGetsRemoved tests keep their names although nothing "resets" anymore.

kdambekalns:
Thanks for the review! I'll take care of the changes.

I need to check my notes, something about Redis doing the right thing anyway keeps nagging me…

PR #3611, opened 2026-08-24
BUGFIX: Preserve by-reference returns in proxies

### LINKED ISSUE #3590: Proxied methods returning by reference silently break the reference

### Current Behavior

If a class contains a method which returns by reference and that method ends up in the proxy class (for example, because it is advised), the reference is silently broken: the proxy method's signature correctly declares the `&`, but the generated body assigns the parent's return value to a local variable and returns that copy. Modifications through the returned reference then have no effect on the object's state.

### Expected Behavior

Either the reference survives proxy building, or – if that is not feasible – proxy building fails with a meaningful exception which explains that return-by-reference is not supported for advised methods. Silently changing the behavior is the worst of the possible outcomes.

### Steps To Reproduce

Create the following class:

```php
#[Flow\Scope('singleton')]
class RefService
{
    protected array $items = ['initial'];

    public function &itemsReference(): array
    {
        return $this->items;
    }

    public function getItems(): array
    {
        return $this->items;
    }
}
```

advise `itemsReference()` with any advice, and run:

```php
$items = &$refService->itemsReference();
$items[] = 'added via reference';
var_dump($refService->getItems());
```

Without the advice, the dumped array contains both entries. With the advice in place, the modification is lost and the array only contains `initial`.

### Environment

```markdown
- Flow: 9.2-dev (9d5c82d0a)
- PHP: 8.4.24
```

### Anything else?

_No response_

robertlemke:
Methods which are declared to return by reference (`function &foo()`) lost their reference semantics when the generated proxy method contained code before and after the parent call: the parent call was assigned to a local variable by value, therefore "return $result" returned a copy.

The proxy method generator now assigns the parent call by reference if the method itself returns by reference.

For advised methods this cannot be repaired: the return value passes through the join point and the advice chain by value, so the reference is lost by design. Instead of silently changing the semantics of such a method, the AdvisedMethodInterceptorBuilder now refuses to build the interceptor code and throws an exception while the proxy classes are compiled. The exception names the affected class and method and explains the possible ways out. Methods returning by reference which are introduced by an interface are covered as well.

Note that this is a change in behavior which should be mentioned in the release notes: a project with a pointcut which happens to match a method returning by reference will not compile anymore until the pointcut is adjusted. None of the Flow and Neos packages declare such methods.

Fixes #3590

### DIFF
--- a/Neos.Flow/Classes/Aop/Builder/AdvisedMethodInterceptorBuilder.php
+++ b/Neos.Flow/Classes/Aop/Builder/AdvisedMethodInterceptorBuilder.php
@@ -48,6 +48,12 @@ class AdvisedMethodInterceptorBuilder extends AbstractMethodInterceptorBuilder
         if ($declaringClassName !== $targetClassName) {
             $originalMethod = MethodGenerator::copyMethodSignature(new \Laminas\Code\Reflection\MethodReflection($declaringClassName, $methodName));
             $proxyMethod->setParameters($originalMethod->getParameters());
+            $methodReturnsReference = $originalMethod->returnsReference();
+        } else {
+            $methodReturnsReference = $proxyMethod->returnsReference();
+        }
+        if ($methodReturnsReference) {
+            throw new Exception(sprintf('The %s cannot build interceptor code for method %s::%s() because it returns by reference: references cannot be preserved through an advice chain, so the advised method would silently return a copy instead of the original reference. Please remove the advice for this method, or exclude the class from proxying by adding a #[Flow\Proxy(false)] attribute.', __CLASS__, $targetClassName, $methodName), 1785837971);
         }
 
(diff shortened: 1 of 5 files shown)


kitsunet:
Is there a good argument for returning by reference? This seems to me like a "better don't do this" kind of thing that we normally wouldn't want to support?

robertlemke:
> Is there a good argument for returning by reference? This seems to me like a "better don't do this" kind of thing that we normally wouldn't want to support?

I think we shouldn't judge if that's a good idea or not. If someone decides to do it, it's unexpected if Flow returns a copy instead of a reference. So, even if we need to support some dirty hacks a user decided to go for, I'd rather try to support that instead of silently turning their code into something else.

kitsunet:
I would throw an exception and say that we do not want to support it 😆 But fine I guess.

kdambekalns (approved):
👍 by 👀

Good catch!

PR #3617, opened 2026-09-01
TASK: Extract Neos.Flow.Testing package that provides test base classes as dev dependency.

mficzel:
This change removes the need for special autoloading in Tests by:

- moving the test base-classes into a separate `Neos.Flow.Testing` package that can be used as dev-dependency
- requiring all the fixtures in tests explicitly via require_once

The old classes from the `Neos\Flow\Tests` namespace are moved aswell for backwards compatibility but as they were deprecated already a note was added that they will be removed with flow 10.

TODOS:
- [x] Extract Testcase base classes to Neos.Flow.Testing
- [x] Provide new base classes without deprecated methods
- [x] Require Fixtures in tests explicitly ... as autoloading will no longer work

Follow up prs should address:

* all tests that currently extend `Neos\Flow\Tests\UnitTestCase` and `Neos\Flow\Tests\FunctionalTestCase` should be  changed to `Neos\Flow\Testing\UnitTestCase` resp. `Neos\Flow\Testing\FunctionalTestCase`
* all usages of the deprecated methods from  `Neos\Flow\Tests\BaseTestCase` like `$this->inject` should be adjusted with non deprecated methods
* the testing autoloader can be removed from the unit test bootstrap in the neos|flow-development-distribution

**Upgrade instructions**

**Review instructions**

To test this you should disable the lines here in the PhpUnit bootstrap:
https://github.com/neos/BuildEssentials/blob/9.2/PhpUnit/UnitTestBootstrap.php#L24
https://github.com/neos/BuildEssentials/blob/9.2/PhpUnit/UnitTestBootstrap.php#L30-L31

The follow up adjustments are not included to keep this pr reviewable. 

It should also be considered wether this is too breaky as this will require all tests that are built on top of the Neos Base Testcases to only extend classes that are autoloadable via composer or that are required manually.

However it is tests ... not sure they are covered by our semantic promise.

Prior art that were not successful:

* Attempt 1 https://github.com/neos/flow-development-collection/pull/2683
* Attempt 2 https://github.com/neos/flow-development-collection/pull/3105

**Checklist**

### DIFF
--- a/Neos.Eel/Tests/Unit/InterpretedEvaluatorTest.php
+++ b/Neos.Eel/Tests/Unit/InterpretedEvaluatorTest.php
@@ -17,6 +17,9 @@ namespace Neos\Eel\Tests\Unit;
 use Neos\Eel\Context;
 use Neos\Eel\InterpretedEvaluator;
 
+require_once "AbstractEvaluatorTestcase.php";
+
+
 /**
  * Interpreted evaluator test
  */
--- a/Neos.Eel/Tests/Unit/Validation/ExpressionSyntaxValidatorTest.php
+++ b/Neos.Eel/Tests/Unit/Validation/ExpressionSyntaxValidatorTest.php
@@ -5,18 +5,40 @@ declare(strict_types=1);
 namespace Neos\Eel\Tests\Unit\Validation;
 
 use Neos\Eel\Validation\ExpressionSyntaxValidator;
-use Neos\Flow\Tests\Unit\Validation\Validator\AbstractValidatorTestcase;
+use Neos\Flow\Tests\UnitTestCase;
+use Neos\Flow\Validation\Validator\ValidatorInterface;
 use PHPUnit\Framework\Attributes\Test;
 
 /**
  * Tests for the ExpressionSyntaxValidator
  */
-final class ExpressionSyntaxValidatorTest extends AbstractValidatorTestcase
+final class ExpressionSyntaxValidatorTest extends UnitTestCase
 {
     /**
      * @var string
      */
     protected $validatorClassName = ExpressionSyntaxValidator::class;
+
+    /**
+     *
+     * @var ValidatorInterface
+     */
+    protected $validator;
+
+    protected function setUp(): void
+    {
+        $this->validator = $this->getValidator();
+    }
+
+    protected function getValidator($options = [])
+    {
+        return $this->getAccessibleMock($this->validatorClassName, [], [$options], '', true);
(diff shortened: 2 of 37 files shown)


kitsunet (commented):
I like this a lot. My thinking was that we might want to extract these directly into BuildEssentials where also the phpunit config is? But I guess this works too.

Neos.Eel/Tests/Unit/InterpretedEvaluatorTest.php
```diff
@@ -17,6 +17,9 @@
 use Neos\Eel\Context;
 use Neos\Eel\InterpretedEvaluator;
 
+require_once "AbstractEvaluatorTestcase.php";
+
+
```
kdambekalns, line 22:
```suggestion
```

Neos.Eel/Tests/Unit/Validation/ExpressionSyntaxValidatorTest.php
```diff
@@ -5,19 +5,41 @@
...
-final class ExpressionSyntaxValidatorTest extends AbstractValidatorTestcase
+final class ExpressionSyntaxValidatorTest extends UnitTestCase
 {
     /**
      * @var string
      */
     protected $validatorClassName = ExpressionSyntaxValidator::class;
 
+    /**
+     *
```
kdambekalns, line 23:
```suggestion
```

Neos.Eel/Tests/Unit/Validation/ExpressionSyntaxValidatorTest.php
```diff
@@ -5,19 +5,41 @@
...
+    protected function getValidator($options = [])
+    {
+        return $this->getAccessibleMock($this->validatorClassName, [], [$options], '', true);
+    }
+
+    protected function validatorOptions($options)
+    {
+        $this->validator = $this->getValidator($options);
+    }
+
```
kdambekalns, line 42:
```suggestion
```

this is unused in this test.

kdambekalns (changes requested):
There is a mix of `require_once('Fixtures/…');` and `require_once(__DIR__ . '/Fixtures/…');` in this PR. That should be cleaned up.

kdambekalns:
> I like this a lot.

Me too.

> My thinking was that we might want to extract these directly into BuildEssentials where also the phpunit config is? But I guess this works too.

Both would work, but keeping things together might be better? Then again… 🤷‍♂️

Issue #3615, opened 2026-08-27
BUG: New constructor argument wiring not up to date with actual constructor signature


mhsdesign:
Probably a regression from #3510 
Tested with Flow 9.2-beta2
FLOW_CONTEXT=Testing/Behat

When developing in testing context i regularly create new factories or add new constructor injections to factories.

Im my case i wanted to add the `ImageFactory` which is also just another final class with constructor injection and of implicit type prototype like this factory

```php
final class MyThingFactory
{
    public function __construct(
        private OneFactory $oneFactory,
        private OtherFactory $otherFactory,
        private ImageFactory $imageFactory,
    ) {
    }
```

> Uncaught Exception in Flow Too few arguments to function MyThingFactory::__construct(), 2 passed in Testing/SubContextBehat/Cache/Code/Flow_Object_Classes/Vendor_Endpoints_Controller_SomeController.php on line 371 and exactly 3 expected
> thrown in file Packages/.../MyThingFactory.php

Which means the place where `MyThingFactory` was injected - a controller - via

```php
    #[Flow\Inject()]
    protected MyThingFactory $myThingFactory;
```

did not instantiate the class properly and missed out on the new `$imageFactory`.

mhsdesign:
Workaround either flush the cache (sadly fully) or comment out the root injection

```php
#[Flow\Inject()]
protected MyThingFactory $myThingFactory;
```

and run flow, and then comment it back in.

So the root cause seems to be that the place instantiating the class seems stale, but flow overall understood the changes to the file in principle.

CC @kitsunet

kitsunet:
Alright this is a tricky one.

You do something you yourself said we can possibly remove in the future of Flow ;) You rely on auto injection for non singleton classes. As you noted yourself. The trick here is that a prototype must have a new instance for this injection. What happens when that is done via auto injection is that the proxy class contains a new call for the class. That was fine as empty `new MyThingFactory()` call before and the proxy constructor would sort out the necessary arguments to inject. However without proxy constructor we must sort this beforehand, so now the calling side (in your example the Controller that wants `MyThingFactory` injected) needs to sort this out, which means now the proxies contain a full constructor call filling in the arguments ala

```
new MyThingFactory(
  Bootstrap::staticObjectManager->get('OneFactory'), //  or some other call, depending on argument type etc.
  Bootstrap::staticObjectManager->get('OtherFactory'),
)
```

[...]

PR #3619, opened 2026-09-05
BUGFIX: Support circular singleton dependencies

### LINKED ISSUE #3618: Circular dependencies between singletons fail since 9.2

### Current Behavior

Since Flow 9.2 (beta3 as well as the current 9.2 branch), singletons which depend on each other – directly or through a longer chain – cannot be instantiated anymore. The Object Manager throws

```
Neos\Flow\ObjectManagement\Exception\CannotBuildObjectException:
Circular dependency detected while trying to instantiate class "…" (1168505928)
```

This affects mutual dependencies through typed `#[Flow\Inject]` properties (which are injected eagerly) as well as chains which involve constructor injection. Such object graphs are common in Flow applications and worked fine up to Flow 9.1 – the same application boots without any changes on 9.1 and fails with a 500 on every request on 9.2.0-beta3.

The regression was introduced with #3510: the Object Manager now resolves constructor arguments itself in `get()`, before the instance exists, and `instantiateClass()` marks the class as "being instantiated". At the same time, the generated proxy constructors lost the `setInstance()` call which used to register the half-constructed singleton as their very first statement. That early registration was what actually closed the circle in 9.1 – a re-entrant `get()` found the already registered instance. The "circular dependency" guard in `instantiateClass()`, by contrast, was never armed in the 9.1.x releases: the flag was only checked and unset, but never set. #3510 removed the working mechanism and armed the dormant guard in one go.

### Expected Behavior

Singletons depending on each other can be instantiated like in Flow 9.1: the instance is registered before its dependencies are resolved, so a circular reference resolves to the instance currently being built.

### Steps To Reproduce

Two singletons with mutual typed property injection:

```php
#[Flow\Scope('singleton')]
class CircularSingletonA
{
    #[Flow\Inject]
    public CircularSingletonB $singletonB;
}

#[Flow\Scope('singleton')]
class CircularSingletonB
{
    #[Flow\Inject]
    public CircularSingletonA $singletonA;
}
```

`$objectManager->get(CircularSingletonA::class)` throws the exception above. The same happens if one side uses constructor injection instead (`public function __construct(public CircularSingletonC $singletonC)`).

### Environment

```markdown
- Flow: 9.2.0-beta3, also 9.2-dev (b4dd6fd43)
- PHP: 8.4.24
```

### Anything else?

A functional test covering both variants passes on 9.1 and fails on 9.2. A pull request with a fix – registering the singleton instance in the Object Manager before its constructor arguments are resolved and its constructor is called – follows.

Note that mutual dependencies between *prototypes* are a different story: they lead to infinite recursion in 9.1 already, so the guard is right to reject them.

Related to #3615, another regression caused by #3510.

robertlemke:
Since #3510, the Object Manager resolves the constructor arguments of an object itself, before the instance exists, and marks the class as being instantiated. At the same time, the generated proxy constructors lost the `setInstance()` call which registered a half-constructed singleton as their very first statement. That early registration was what allowed singletons to depend on each other: a re-entrant `get()` for an object which is currently being built returned that very instance. Without it, such object graphs – which are common in Flow applications and worked up to Flow 9.1 – fail with a "Circular dependency detected" exception while their dependencies are resolved.

The Object Manager now creates the instance of a singleton (or session scoped object) without calling its constructor, registers it, and only then resolves the constructor arguments and calls the constructor. If that fails, the registration is rolled back. This restores the 9.1 behavior at the place where dependencies are resolved today.

Prototypes cannot be registered up front, because each request must result in a new instance. Two prototypes depending on each other lead to an infinite recursion in 9.1 already, so the circular dependency guard keeps rejecting them – now with a meaningful exception instead of a stack overflow.

Fixes #3618

### DIFF
--- a/Neos.Flow/Classes/ObjectManagement/ObjectManager.php
+++ b/Neos.Flow/Classes/ObjectManagement/ObjectManager.php
@@ -222,25 +222,105 @@ class ObjectManager implements ObjectManagerInterface
             return $this->objects[$objectName][self::KEY_INSTANCE];
         }
 
-        // by object name
+        if (!isset($this->objects[$objectName]) || $this->objects[$objectName][self::KEY_SCOPE] === ObjectConfiguration::SCOPE_PROTOTYPE) {
+            return $this->instantiateClass($className, $this->autowireConstructorArguments($objectName, $className, $constructorArguments));
+        }
+
+        return $this->instantiateRegisteredObject($objectName, $className);
+    }
+
+    /**
+     * Creates the instance of a singleton or session scoped object and registers it *before* its
+     * constructor arguments are resolved and its constructor is called.
+     *
+     * Objects of these scopes may depend on each other, directly or through a chain of other
+     * objects. Registering the instance first makes sure that a re-entrant call to get() for the
+     * same object – triggered while resolving its dependencies – returns the instance which is
+     * currently being built, instead of trying to build a second one and running into the
+     * circular dependency guard of instantiateClass().
+     *
+     * @param string $objectName Name of the object to instantiate
+     * @param class-string $className Name of the class implementing the object
+     * @return object The object
+     * @throws Exception\CannotBuildObjectException
+     * @throws \Throwable
+     */
+    protected function instantiateRegisteredObject(string $objectName, string $className): object
+    {
+        $classReflection = new \ReflectionClass($className);
+        if ($classReflection->isInternal()) {
+            $instance = $this->instantiateClass($className, $this->autowireConstructorArguments($objectName, $className, []));
+            $this->registerInstance($objectName, $className, $instance);
+            return $instance;
+        }
+
+        $instance = $classReflection->newInstanceWithoutConstructor();
+        $this->registerInstance($objectName, $className, $instance);
+        try {
+            $constructorArguments = $this->autowireConstructorArguments($objectName, $className, []);
+            if ($classReflection->hasMethod('__construct')) {
+                $instance->__construct(...$constructorArguments);
+            }
+        } catch (\Throwable $throwable) {
+            $this->unregisterInstance($objectName, $className);
+            throw $throwable;
+        }
+        return $instance;
+    }
+
+    /**
+     * Registers the given instance for the object name and – if the implementation class is
+     * registered as a singleton of its own – for the class name as well, so that requesting
+     * the class directly returns the same instance.
+     *
+     * @param class-string $className
+     */
+    protected function registerInstance(string $objectName, string $className, object $instance): void
+    {
+        $this->objects[$objectName][self::KEY_INSTANCE] = $instance;
+        if ($this->classIsRegisteredAsSingleton($objectName, $className)) {
+            $this->objects[$className][self::KEY_INSTANCE] = $instance;
+        }
+    }
+
+    /**
+     * @param class-string $className
+     */
+    protected function unregisterInstance(string $objectName, string $className): void
+    {
+        unset($this->objects[$objectName][self::KEY_INSTANCE]);
+        if ($this->classIsRegisteredAsSingleton($objectName, $className)) {
+            unset($this->objects[$className][self::KEY_INSTANCE]);
+        }
+    }
+
+    /**
+     * @param class-string $className
+     */
+    protected function classIsRegisteredAsSingleton(string $objectName, string $className): bool
+    {
+        return $objectName !== $className
+            && isset($this->objects[$className])
+            && $this->objects[$className][self::KEY_SCOPE] === ObjectConfiguration::SCOPE_SINGLETON;
+    }
+
+    /**
+     * Resolves the constructor arguments configured for the given object – by object name and,
+     * if different, by class name – which were not passed explicitly.
+     *
+     * @param class-string $className
+     * @param array<mixed> $constructorArguments Arguments which were passed explicitly
+     * @return array<mixed> The complete list of constructor arguments
+     */
+    protected function autowireConstructorArguments(string $objectName, string $className, array $constructorArguments): array
+    {
         if (isset($this->objects[$objectName][self::KEY_CONSTRUCTOR_ARGUMENTS])) {
             $constructorArguments = $this->autowireArguments($this->objects[$objectName][self::KEY_CONSTRUCTOR_ARGUMENTS], $constructorArguments);
         }
-
-        // by class name
         if ($objectName !== $className && isset($this->objects[$className][self::KEY_CONSTRUCTOR_ARGUMENTS])) {
             $constructorArguments = $this->autowireArguments($this->objects[$className][self::KEY_CONSTRUCTOR_ARGUMENTS], $constructorArguments);
         }
-
-        if (!isset($this->objects[$objectName]) || $this->objects[$objectName][self::KEY_SCOPE] === ObjectConfiguration::SCOPE_PROTOTYPE) {
-            return $this->instantiateClass($className, $constructorArguments);
-        }
-
-        $this->objects[$objectName][self::KEY_INSTANCE] = $this->instantiateClass($className, $constructorArguments);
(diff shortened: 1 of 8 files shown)


kitsunet:
As noted in the issue, I do not consider this a bug but a fixed weirdness and I don't like the whole Reflection creation introduced by this.

robertlemke:
> As noted in the issue, I do not consider this a bug but a fixed weirdness and I don't like the whole Reflection creation introduced by this.

The fixed weirdness would, however, force us to do an extensive rework of our own application (Beach). I am sure that we are not the only ones, and I would not expect such a breaking change in a minor release.

robertlemke:
> don't like the whole Reflection creation introduced by this.

Can you elaborate? I can try to use the ReflectionService for some metadata it might have. But we do need some runtime information and probably won't get around creating a Reflection class when we instantiate singletons.

I know that you dislike this fix, but I consider it a fix of a breaking change which shouldn't happen in a minor release. That's the legacy, call it a wrong decision from 15 years ago, but I think we need to support this unless we decide to drop support for it in the next major.

kitsunet:
`$classReflection->newInstanceWithoutConstructor();` this seems pretty dirty magic and might/will fail if the injected instance is immediately (tried to be) used in the receiving object. As in, with this, one side of the circle will receive a "unfinished" object in the constructor. Sounds like a big foot gun. I think it boils down to, is this a feature that was removed or a bug that was fixed? I consider it a bug that it ever worked. But of course I do not know how many people exploited the bug.
Neither Flow nor Neos use this and there was no test for it prior. And I think we can agree that circular dependencies are a bad idea.
I am at least not convinced that this is the right direction. We can immediately add a deprecation log and remove this whole code path again in 10 🤷 But I am still not entirely in agreement.

robertlemke:
Allright, a few points on the "feature or bug" question.

Circular dependencies via setter or property injection are documented behavior. The manual forbids circular constructor dependencies. For such cycles it explicitly recommends setter injection as the workaround: "As a workaround you can use setter injection instead for either one or both of the objects causing the trouble." See https://flowframework.readthedocs.io/en/9.2/TheDefinitiveGuide/PartIII/ObjectManagement.html#dependency-resolution

The buildSetInstanceCode() method which #3510 removed existed for exactly this purpose. The classesBeingInstantiated flag I introduced back then documents the same intention. There was no test for it, that is why this could break unnoticed (but there are so many other features we don't have tests for).

The "unfinished object" is not new. Until 9.1, setInstance($this) was the very first statement of every proxy constructor. It ran before constructor injection and before property injection. A re-entrant get() therefore always returned a half-constructed instance. This PR restores exactly that behavior at the place where #3510 moved the dependency resolution to. newInstanceWithoutConstructor() is simply the only way to have an instance before its constructor arguments are resolved.

If you want to remove the foot gun for the one case the manual declares unsupported, I can add a check. It would keep throwing when a not yet constructed object is requested as a constructor argument. Setter and property cycles would still pass.

You remember that I mostly modeled Flow's DI part after Spring Framework, which was brand new back then. So, I lookt it up: This is also how Spring handles it. Constructor cycles throw a BeanCurrentlyInCreationException. Setter cycles resolve through early singleton references. The reference documentation states that this "forces one of the beans to be injected into the other prior to being fully initialized itself" (https://docs.spring.io/spring-framework/reference/core/beans/dependencies/factory-collaborators.html). Spring Boot 2.6 later prohibited circular references by default and added an opt-in property to restore the previous behavior (https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.6-Release-Notes). We can do the same for Flow 10: a setting which defaults to "not allowed", a deprecation log in 9.3, and the compatible behavior in 9.2. Dropping it silently in a minor release is the one option I don't see.

Regarding the reflection: I measured it. Creating the ReflectionClass costs about 0.4 microseconds per singleton. That is less than 0.1 ms per request, even if all 205 singletons of Beach Control Panel are instantiated. The ReflectionService cannot help here. It is a compile time tool, and the Object
Manager must not depend on an object it manages itself.

kitsunet:
Alright I see. In that case, agreed, we should fix this. However I would like to suggest an alternative approach. We only need early registration in case the object has more than constructor injection, which also means it will have a proxy constructor as of now. We could re-instate setting the instance in the proxy constructor for such cases, before the injections are triggered. That should allow the documented behaviour again, which has a deterministic order and therefore is fine, also probably the smaller change. I can take care of that, but not today.

robertlemke:
Thanks, I appreciate that.

Regarding you idea: I did consider registering the instance in the proxy constructor, since that is what 9.1 did, and I have now also tried it: the setter and property cycles indeed work again with that approach, but it only covers half of the problem.

Since #3510, the constructor arguments are resolved before new is called, so a registration inside the constructor comes too late as soon as a constructor-injected object is part of the cycle. 

Take our fixture pair: C gets D injected as a property, D gets C injected through its constructor. If D is requested first, resolving its arguments creates C, whose property injection requests D again. D is not registered yet, so a second D is created. The first one ends up in C, the second one becomes the registered singleton. Two instances of a singleton and the dev won't see any error message. In 9.1 this could not happen, because constructor injection also ran after setInstance() inside the constructor.

I have pushed another test, I tried to write it exactly for this case (requesting D first). Here locally, it passes with the Object Manager approach but fails with the proxy constructor approach. So, I'd rather stick with the former. 

One more arguments: it does not bring back the mechanism which caused #3078, where a new Singleton() in application code overwrites the registered instance.

kitsunet:
right, I guess we will to this then. Thanks for walking through this!

kitsunet (approved):
Will approve given there are no alternatives right now to make this non breaking.

mhsdesign (dismissed):
Im with christian, the impossible is correctly impossible and should stay so :D

https://github.com/neos/flow-development-collection/issues/3618#issuecomment-5557676916

robertlemke:
@mhsdesign Before 9.2.0 goes out, I'd like to explain the release impact here, because shipping the final with this regression would, I believe, cause real damage.

As it stands, #3510 makes any application whose singletons depend on each other, directly or through a chain, fail to boot on 9.2. These graphs are common in larger Flow applications, they are the documented and supported setter/property-injection pattern, and they have worked since the very early days. Someone upgrading from 9.1 to a 9.2 minor would hit this immediately, without changing a single line on their side. The full reasoning is in my comment in #3618.

And this is not the only fallout from #3510. #3615, which you filed yourself, is a second regression from the same change, in the testing context. That makes me all the more cautious about releasing 9.2 as it stands.

On the direction itself, I agree: Christian has approved this PR ("given there are no alternatives right now to make this non breaking"), and the two of us walked through the details and agreed on the path in the thread. I fully share the long-term goal of not encouraging circular dependencies. My concern is purely timing and the semver contract, because removing a documented, decade-old behaviour in a minor breaks working applications silently. The clean way to the same destination is to keep it working in 9.2, add a deprecation notice in 9.3, and make "disallowed" the default in 10 with an opt-in, exactly as Spring did in Boot 2.6.

Let's not ship 9.2 with a known boot-breaking regression. Let's merge a compatible fix first, whether this one or Christian's proxy-constructor variant, and then remove circular dependencies properly in 10.

mhsdesign:
I just think we strive for fundamental different goals but thats fine. First and foremost thought THANK YOU for bringing up this regression - i forgot to mention that;)

Neos releases always were okay with explicitly dropping niche features that got harder to maintain during minor releases, so i think thats not an issue per-se. It rather depends on if there are real world projects relying on this and how many and at what scale? A plugin can be fixed via patch and is quickly discovered to have faulty code.
But if you say and believe:

>  I believe, cause real damage.

Then i have to take your word;) Maybe the compromise is then to DEPRECATE (and document that) in Flow 9.2 NOW and REMOVE it with 10.

 > We can do the same for Flow 10: a setting which defaults to "not allowed", a deprecation log in 9.3, and the compatible behavior in 9.2. Dropping it silently in a minor release is the one option I don't see.

Anyway ill remove my hard deny here for now;)

PR #3620, opened 2026-09-11
Lazy PHP objects in Flow

kitsunet:
Singletons and session scoped objects are now created as native PHP lazy proxies. The ObjectManager registers the proxy right away and only builds the actual object, together with its constructor arguments, once the proxy is used for the first time. Objects built by a factory are handled the same way if their class is known and can be made lazy. This replaces Flow's own DependencyProxy and the lazy property injection code in the generated proxy classes.

Previously, a DependencyProxy stood in for a lazy dependency. It forwarded method calls through __call() and had to be resolved by hand wherever a type check or property access got in the way. A PHP lazy proxy is an instance of the actual class, so these special cases go away. Registering the proxy before anything is built also keeps circular dependencies between singletons working, no matter through which of them the cycle is entered.

Once a singleton has been built, its instance is registered for the class name and for every interface the class provides, so requesting any of them returns the same object. Prototypes are still built right away: their dependencies have to be resolved on every request anyway, and errors while doing so should surface where the object is requested. Configuring a differing class name for a class in Objects.yaml now only logs a warning instead of failing.


kitsunet:
I think the PHPstan error is due to the old stan version, PHP8.4 clearly has this method... Will try to fix/update.
Due to the new constructor injection in objectManager, the objectmanager has way more insight into our object reflection and can resolve interface -> class relations. Which was one of the issues of previous changes.
This change is battle tested in a large flow application (not beach) as I wrote it for their 9.0 system as a patch. It's running there for almost a year now without problems.

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
```diff
@@ -284,9 +284,8 @@ protected function buildPropertyInjectionCode(Configuration $objectConfiguration
     protected function buildPropertyInjectionCodeByConfiguration(Configuration $objectConfiguration, $propertyName, Configuration $propertyConfiguration): array
     {
         $className = $objectConfiguration->getClassName();
-        $propertyObjectName = $propertyConfiguration->getObjectName();
         $propertyClassName = $propertyConfiguration->getClassName();
```
mficzel, line 287: Is the switch from `getObjectName` to `getClassName` really safe?

I remember running into issues because the objectName can be an interface or even a virtual-object we invented in configuration.

Will take a deeper look soon as this would be really great.

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -626,10 +573,49 @@ protected function buildObjectByFactory(string $objectName): object
...
 
-        return $factoryMethodName(...$factoryMethodArguments);
+    /**
+     * @param class-string $className
+     * @param \Closure(): object $builder Creates the actual instance once the proxy is used for the first time
+     * @throws \ReflectionException
+     */
+    protected function buildLazyProxy(string $className, \Closure $builder): object
+    {
+        return (new \ReflectionClass($className))->newLazyProxy($builder);
```
mficzel, line 619: This should be simply ignored via `/** @phpstan-ignore method.notFound */` ... the update to phpstan 2.2 will detect this and raise an error for an unnecessary ignore which will ensure we will remove the ignore again once we can.

Neos.Flow/Classes/ObjectManagement/DependencyInjection/DependencyProxy.php
```diff
@@ -1,93 +0,0 @@
-<?php
```
mficzel, line 1: Wonder wether this should stay as deprecated and unused until Flow 10. Since this is always used in a typeCheck PHP will complain if the class is missing but for developers it would be hard to write code that passes static analysis in multiple Neos versions.

kitsunet: I am happy to leave it around, however I hadn't found the scenario yet, because either you do `instanceof` which is safe for non existing classes or you do `get_class($object) === (importedOrNot)DependencyProxy::class` which is also fine with non existing classes.

mficzel: Yes php will not complain but PhpStan will ... i think

Neos.Flow/Classes/ObjectManagement/DependencyInjection/ProxyClassBuilder.php
kitsunet: AFAIK I purged the code of these instances previously in the constructor refactoring. The only way an object configuration cannot have a class name assigned at this stage is a factory is assigned and we had no way to know the exact class that comes out of it. That case is covered 2 lines below, and then we don't care about the object name nor the class name.

I would really at some point like to refactor the object configuration into value objects to ensure/show these relationships better.

mficzel:
I noticed that i now get a lazy proxy even when i set `#[Flow\Inject(lazy:false]` .. is that intended?

Can confirm that this works for interfaces which makes me very happy. Still have to fully understand it.

kitsunet:
Yes the annotation is ignored (and should then change for next major). There is no real point in disabling this feature IMHO.

Neos.Flow/Classes/ObjectManagement/DependencyInjection/DependencyProxy.php
kitsunet: reinstated

mficzel (commented):
As i understand this i see two side effects:
1. Objects where the class can be determined beforehand are now always lazy injected . Regardless what the annotation says. ... This is probably fine but should be documented.
2. Objects with configured factories will no longer be injected lazy at all. Regardless what the annotation says. ... I am not sure this is safe, this is used quite a bit even in Flow itself.

kitsunet:
AFAIK we never guaranteed that you get a lazy injection. In fact if the object is already constituted you would definitely not. That said, this change would shift the burden to the factories to provide lazy objects, which should be fine. Flows factories (like cache for example) could pick that up and do that.

mficzel:
I agree that factories could/should care about lazyness. We should discuss shortly wether we consider this too breaky or not. For a major this is fine anyways.

kitsunet:
Yeah, we can push this to 10, I just wanted to put this out there as it is tested in a bigger Flow application for a while now and seems like the cleaner forward way than the bugfix for the cyclic dependencies (yet still fixing that bug, because these lazy objects allow for cyclic dependencies)

mficzel:
Not against releasing it in 9.3, just discuss the implications before. Maybe at a team meeting/sprint. The change regarding factories is bold but makes sense. Making all other injections lazy would imho be totally fine.

kitsunet:
@mficzel I was actually hoping for 9.2 still 😆 But I guess that might be a bit risky.

mficzel:
As you are RM I suggest discussing this with the other RMs. I  fine with whatever the RMs decide.

To be complete I would like to see the configured object factories create lazy proxies aswell. Could be done in a follow up but should be in the same release.

Neos.Flow/Classes/Annotations/Inject.php
```diff
@@ -32,6 +32,7 @@ final class Inject
      * proxy should be injected instead
      *
      * @var boolean
+     * @deprecated No longer observed, we use PHP 8.4 lazy objects everywhere by default if possible.
```
dlubitz: Can we add a removal target version (10.0)? Would make it easier to search for and to decide on the removal in future.

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
```diff
@@ -626,10 +573,50 @@ protected function buildObjectByFactory(string $objectName): object
...
-        return $factoryMethodName(...$factoryMethodArguments);
+    /**
+     * @param class-string $className
+     * @param \Closure(): object $builder Creates the actual instance once the proxy is used for the first time
+     * @throws \ReflectionException
+     */
+    protected function buildLazyProxy(string $className, \Closure $builder): object
+    {
+        /** @phpstan-ignore method.notFound */
+        return (new \ReflectionClass($className))->newLazyProxy($builder);
```
dlubitz, line 644: Why do we use the LazyProxy and not the LazyGhost here?

dlubitz (approved):
I like. Thank you very much! 

I left two comments, but nothing blocking.

Neos.Flow/Classes/ObjectManagement/ObjectManager.php
kitsunet: To be fair because it was easier to implement that way. The LazyGhost expects me to call the constructor correctly on the provided (LazyGhost) $object, while for the LazyProxy I can just return a (newly) constructed object. I guess we could refactor this to LazyGhosts as well, but it might not be straight forward for all cases. IMHO an improvement for later.

dlubitz: I see. I was just wondering. What about comparison of the object before and after initialization? This is not an issue for us?

kitsunet: Currently don't see where is would make a difference. As soon as you observe the object it gets initialised. So whatever you do to compare it, should trigger the builder already. At least in all my tests I haven't observed any case where it was noticeable.

mficzel (approved):
Fine by 👀 and 🧪  ... adjusting the logger and cache factories to create lazy objects would be great as those are very often injected but not always used

